Issue #16, opened 2019-05-22
linearizeNumerically bug


dellaert:
Brice says (moving here from Bitbucket issue #460):

in linearizeNumerically I think dX should be created insice the factor loop instead of outside

https://bitbucket.org/gtborg/gtsam/src/357e739127deb7d9bec5dad2a5d117d9f4e4e1be/gtsam/nonlinear/factorTesting.h#lines-49

I am not too sure how this didn't affect current unit tests, but my unit test on my custom factor was failing because of that. After checking my jacobians for most of the day and banging my head against the walls, I figured thanks to gdb that this was happening. I moved dX in the loop and my unit test is now passing.

dellaert:
Fixed in PR #17

PR #17, opened 2019-05-22
Fix linearizeNumerically (#16)

bricerebsamen:
Fixes issue #16 by moving variable `dX` in the `key` loop. Added a unit test to repro the issue.


dellaert:
Thanks Brice !!!!

Issue #13, opened 2019-05-20
Expression problem


dellaert:
Address Brice's compile problems (reported in Google groups) by patching slam/expressions.h like so:
```
namespace internal {
Rot3 rotation(const Pose3& pose, OptionalJacobian<3, 6> H) {
  return pose.rotation(H);
}
}  // namespace internal

inline Rot3_ rotation(const Pose3_& pose) { return Rot3_(internal::rotation, pose); }

TEST(SlamExpressions, rotation) {
  Pose3_ T_(0);
  const Rot3_ R_ = rotation(T_);
}

/* ************************************************************************* */
TEST(SlamExpressions, unrotate) {
  Rot3_ R_(0);
  Point3_ p_(1);
  const Point3_ q_ = unrotate(R_, p_);
}

```

dellaert:
Brice, submit a PR with your solution?

dellaert:
Fixed in PR #18

PR #18, opened 2019-05-22
Expression wrapper for rotation

bricerebsamen:
Fixes issue #13 by adding a wrapper for `rotation()` that returns a value and not a const reference. Also added a unit test.


dellaert:
And thanks again !!!!

PR #26, opened 2019-05-29
Release tag free wrapper

varunagrawal:
Updated the cmake files so that the non-release versions of the `gtsam.so` and other similar files do not have the release type tag attached to them. This also ensures the release specific information (e.g. `debug` info) is properly captured rather than circumvented.


dellaert:
Should we close after our off-line discussion?

varunagrawal:
Yes let's close this.

For future context, we've decided to modify the actual cython wrapper to
link to the correct version of the .so files based on the postfix. This
will prevent confusion when making updates to the cython wrapper in the
future.

dellaert:
Ok

PR #27, opened 2019-05-29
C++11 and "-march=native" flags done via "modern" CMake target-based properties

jlblancoc:
I think the title says it all :-)

The PUBLIC flags ensures that "-march=native" & "c++11" propagates to all dependent projects automatically.


dellaert:
We just had an issue with "modern" cmake... Is this supported by cmake that is installed with Ubuntu 16.04?

dellaert:
For record, all on board it if it works with that cmake :-)

dellaert:
Modulo comment above, looks good to me, but on cmake PR's I will typically defer to @chrisbeall to do final check and merge...

jlblancoc:
Ok, let's wait for @chrisbeall opinion!

> We just had an issue with "modern" cmake... Is this supported by cmake that is installed with Ubuntu 16.04?

Hmm... it seems you Jenkins instance uses Ubuntu 18.04 (CMake 3.10), and I don't have any 16.04 (with cmake 3.5) around to test. So this PR is not tested against that older version, though "modern" here normally means cmake >=3.0 so it should not be a problem, but it would be better to be sure. 

Probably the best way to do this is by adding several docker images to CI?

chrisbeall:
I'll test this on Ubuntu 16.04, but might not have time until Thursday evening or Friday.

jlblancoc:
Rebased on top of develop so Travis con run on this PR too.

dellaert:
Woohoo, travis is working :-) Seems that maybe the c++11 flags do not work for all configs?

gtsam/CMakeLists.txt
```diff
@@ -101,6 +101,7 @@ message(STATUS "Building GTSAM - shared: ${BUILD_SHARED_LIBS}")
 
 # BUILD_SHARED_LIBS automatically defines static/shared libs:
 add_library(gtsam ${gtsam_srcs})
+target_compile_features(gtsam PUBLIC cxx_std_11)
```
chrisbeall: The `cxx_std_11` feature was introduced in [CMake 3.8](https://cmake.org/cmake/help/v3.8/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html). Ubuntu 16.04 has CMake 3.5.1 by default, so unfortunately we can't use this.
Looks like Travis has CMake 3.12, but the property still wasn't set. Don't know why.

jlblancoc: You're right! Let's see if https://github.com/borglab/gtsam/pull/27/commits/53cf33b045426bd1b443c0297d1d6f9d062a70c8 makes it...

gtsam/CMakeLists.txt
```diff
@@ -101,6 +101,15 @@ message(STATUS "Building GTSAM - shared: ${BUILD_SHARED_LIBS}")
 
 # BUILD_SHARED_LIBS automatically defines static/shared libs:
 add_library(gtsam ${gtsam_srcs})
+
+if(NOT CMAKE_VERSION VERSION_LESS 3.8)
+	target_compile_features(gtsam PUBLIC cxx_std_11)
+else()
+	if (NOT MSVC)
+		target_compile_options(gtsam PUBLIC -std=c++11)
```
chrisbeall: Unfortunately this also doesn't seem to work. Failed in Travis, and for me, too. I have CMake 3.5.1, but `-std=c++11` does not appear in my compiler args. I checked that we *do* reach this line, so there's some problem with actually setting the property on the target?  First error:
```
[  2%] Building CXX object wrap/CMakeFiles/wrap_lib.dir/Method.cpp.o
cd /home/cbeall/git/gtsam/build-scratch/wrap && ccache /usr/bin/c++    -I/home/cbeall/git/gtsam  -Wall -O3 -DNDEBUG    -O3 -DNDEBUG   -o CMakeFiles/wrap_lib.dir/Method.cpp.o -c /home/cbeall/git/gtsam/wrap/Method.cpp
In file included from /usr/include/c++/5/unordered_set:35:0,
                 from /home/cbeall/git/gtsam/wrap/OverloadedFunction.h:23,
                 from /home/cbeall/git/gtsam/wrap/FullyOverloadedFunction.h:21,
                 from /home/cbeall/git/gtsam/wrap/MethodBase.h:22,
                 from /home/cbeall/git/gtsam/wrap/Method.h:21,
                 from /home/cbeall/git/gtsam/wrap/Method.cpp:18:
/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support \
  ^
```
It also doesn't show up when we print config options (`march=native` is also missing):
```
--   C compilation flags            :  -Wall -O3 -DNDEBUG    -O3 -DNDEBUG
--   C++ compilation flags          :  -Wall -O3 -DNDEBUG    -O3 -DNDEBUG
```

jlblancoc: Ok, I had to log in into a u16.04 docker image to debug this one, but I think it should be fixed now ;-)

All commits in this PR have been rebased for ease of history inspection.

PS: CXX_FLAGS won't display all the flags now, since most flags are now set via cmake target properties. So, those properties are now also shown in the console in the final cmake summary. There are "scary" cmake expressions over there (`$<condition:...>`), but that's right, and it's one of the main reasons of why using this system (+exporting cmake targets) makes a system easier to maintain in the long term (hope so!).

dellaert:
Chris, I’ll let you merge…

dellaert:
@chrisbeall are you good now?

dellaert:
@jlblancoc can you paste an example of new output?

jlblancoc:
(I just solved the merge conflict)

Sure: 

```
$ cmake .
-- GTSAM_SOURCE_ROOT_DIR: [/home/jlblanco/code/gtsam-jlblancoc]
-- Boost version: 1.65.1
-- Found the following Boost libraries:
--   serialization
--   system
--   filesystem
--   thread
--   program_options
--   date_time
--   timer
--   chrono
--   regex
--   atomic
-- GTSAM_BOOST_LIBRARIES: optimized;/usr/lib/x86_64-linux-gnu/libboost_serialization.so;/usr/lib/x86_64-linux-gnu/libboost_system.so;/usr/lib/x86_64-linux-gnu/libboost_filesystem.so;/usr/lib/x86_64-linux-gnu/libboost_thread.so;/usr/lib/x86_64-linux-gnu/libboost_date_time.so;/usr/lib/x86_64-linux-gnu/libboost_regex.so;debug;/usr/lib/x86_64-linux-gnu/libboost_serialization.so;/usr/lib/x86_64-linux-gnu/libboost_system.so;/usr/lib/x86_64-linux-gnu/libboost_filesystem.so;/usr/lib/x86_64-linux-gnu/libboost_thread.so;/usr/lib/x86_64-linux-gnu/libboost_date_time.so;/usr/lib/x86_64-linux-gnu/libboost_regex.so
Ignoring Boost restriction on optional lvalue assignment from rvalues
-- Could NOT find MKL (missing: MKL_INCLUDE_DIR MKL_LIBRARIES) 
-- Found Eigen version: 3.3.7
-- Building 3rdparty
-- checking for thread-local storage - found
-- Building base
-- Building geometry
-- Building inference
-- Building symbolic
-- Building discrete
-- Building linear
-- Building nonlinear
-- Building sam
-- Building sfm
-- Building slam
-- Building smart
-- Building navigation
-- GTSAM Version: 4.0.0
-- Install prefix: /usr/local
-- Building GTSAM - shared: ON
-- Building base_unstable
-- Building geometry_unstable
-- Building linear_unstable
-- Building discrete_unstable
-- Building dynamics_unstable
-- Building nonlinear_unstable
-- Building slam_unstable
-- Building partition_unstable
-- GTSAM_UNSTABLE Version: 4.0.0
-- Install prefix: /usr/local
-- Wrote /home/jlblanco/code/gtsam-jlblancoc/build/GTSAMConfig.cmake
-- ===============================================================
-- ================  Configuration Options  ======================
--   CMAKE_CXX_COMPILER_ID type     : GNU
--   CMAKE_CXX_COMPILER_VERSION     : 7.4.0
--   CMake version                  : 3.10.2
--   CMake generator                : Unix Makefiles
--   CMake build tool               : /usr/bin/make
-- Build flags                                               
--   Build Tests                    : Enabled
--   Build examples with 'make all' : Enabled
--   Build timing scripts with 'make all': Disabled
--   Build Docs                     : Enabled
--   Build shared GTSAM libraries   : Enabled
--   Put build type in library name : Enabled
--   Build libgtsam_unstable        : Enabled
--   Build for native architecture  : Enabled
--   Build type                     : Release
--   C++ compilation flags          :  -O3 -DNDEBUG
--   gtsam CXX_STANDARD             : 11
--   gtsam COMPILE_OPTIONS          : -march=native;-Wall;$<$<CONFIG:Debug>:-g;-fno-inline;-DEIGEN_INITIALIZE_MATRICES_BY_NAN>;$<$<CONFIG:RelWithDebInfo>:-g;-O3;-DNDEBUG>;$<$<OR:$<CONFIG:Release>,$<CONFIG:Profiling>>:-O3;-DNDEBUG>;$<$<CONFIG:Timing>:-g;-O3;-DNDEBUG;-DENABLE_TIMING>;-Wno-unused-local-typedefs
--   gtsam INCLUDE_DIRECTORIES      : $<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/metis/include>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/metis/libmetis>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/metis/GKlib>;$<INSTALL_INTERFACE:/usr/local/include/gtsam/3rdparty/metis/>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/SuiteSparse_config>;$<INSTALL_INTERFACE:/usr/local/include/gtsam/3rdparty/SuiteSparse_config>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/CCOLAMD/Include>;$<INSTALL_INTERFACE:/usr/local/include/gtsam/3rdparty/CCOLAMD>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc>;$<INSTALL_INTERFACE:/usr/local/include/>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/build>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/CppUnitLite>;/usr/include;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/Eigen/>;$<INSTALL_INTERFACE:include/gtsam/3rdparty/Eigen/>;/usr/include
--   gtsam COMPILE_DEFINITIONS      : BOOST_OPTIONAL_ALLOW_BINDING_TO_RVALUES;BOOST_OPTIONAL_CONFIG_ALLOW_BINDING_TO_RVALUES
--   Use System Eigen               : OFF (Using version: 3.3.7)
--   Use Intel TBB                  : Yes
--   Eigen will use MKL             : MKL not found
--   Eigen will use MKL and OpenMP  : OpenMP found but GTSAM_WITH_EIGEN_MKL is disabled
--   Default allocator              : TBB
--   Build with ccache              : No
-- Packaging flags                                               
--   CPack Source Generator         : TGZ
--   CPack Generator                : TGZ
-- GTSAM flags                                               
--   Quaternions as default Rot3     : Disabled
--   Runtime consistency checking    : Disabled
--   Rot3 retract is full ExpMap     : Disabled
--   Pose3 retract is full ExpMap    : Disabled
--   Deprecated in GTSAM 4 allowed   : Enabled
--   Point3 is typedef to Vector3    : Disabled
--   Metis-based Nested Dissection   : Enabled
--   Use tangent-space preintegration: Enabled
--   Build Wrap                     : Enabled
-- MATLAB toolbox flags                                      
--   Install matlab toolbox         : Disabled
-- Cython toolbox flags                                      
--   Install Cython toolbox         : Disabled
-- ===============================================================
-- Configuring done
-- Generating done
```

dellaert:
That's super-cool :-)

chrisbeall:
I'm not a huge fan of:
```
--   gtsam CXX_STANDARD             : 11
--   gtsam COMPILE_OPTIONS          : -march=native;-Wall;$<$<CONFIG:Debug>:-g;-fno-inline;-DEIGEN_INITIALIZE_MATRICES_BY_NAN>;$<$<CONFIG:RelWithDebInfo>:-g;-O3;-DNDEBUG>;$<$<OR:$<CONFIG:Release>,$<CONFIG:Profiling>>:-O3;-DNDEBUG>;$<$<CONFIG:Timing>:-g;-O3;-DNDEBUG;-DENABLE_TIMING>;-Wno-unused-local-typedefs
--   gtsam INCLUDE_DIRECTORIES      : $<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/metis/include>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/metis/libmetis>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/metis/GKlib>;$<INSTALL_INTERFACE:/usr/local/include/gtsam/3rdparty/metis/>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/SuiteSparse_config>;$<INSTALL_INTERFACE:/usr/local/include/gtsam/3rdparty/SuiteSparse_config>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/CCOLAMD/Include>;$<INSTALL_INTERFACE:/usr/local/include/gtsam/3rdparty/CCOLAMD>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc>;$<INSTALL_INTERFACE:/usr/local/include/>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/build>;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/CppUnitLite>;/usr/include;$<BUILD_INTERFACE:/home/jlblanco/code/gtsam-jlblancoc/gtsam/3rdparty/Eigen/>;$<INSTALL_INTERFACE:include/gtsam/3rdparty/Eigen/>;/usr/include
--   gtsam COMPILE_DEFINITIONS      : BOOST_OPTIONAL_ALLOW_BINDING_TO_RVALUES;BOOST_OPTIONAL_CONFIG_ALLOW_BINDING_TO_RVALUES
```
That doesn't look very user friendly. This was easy to understand:
```
--   C compilation flags            :  -std=c11   -Wall -march=native -O3 -DNDEBUG
--   C++ compilation flags          :  -std=c++11 -Wall -march=native -O3 -DNDEBUG
```

dellaert:
Two thoughts:
- "gtsam " is definitely redundant on those lines, so I'd omit it.
- can we have both?

chrisbeall:
Can these options be shown & configured in `cmake-gui` and `cmake-curses-gui` as before? I don't see any fields there which show `c++11`, debug flags, etc. Should there be a field `gtsam CXX_STANDARD`? 
New on the left, old on the right. 
image

jlblancoc:
Hmm... I guess it could be done such that the `GTSAM_CMAKE_CXX_*` variables are back, but then integrated into targets (gtsam, gtsam_unstable,...) via the generator expressions instead of the `CMAKE_CXX_*` variables. 
That would be a trade-off solution, right?

There are some pros and cons in both the "old" and the "new" cmake ways, see for example: 
* https://stackoverflow.com/questions/45955272/modern-way-to-set-compiler-flags-in-cross-platform-cmake-project
* https://stackoverflow.com/questions/5096881/does-set-target-properties-in-cmake-override-cmake-cxx-flags

But for library authors, I would strongly advise using the "modern" way to make exported targets as self-contained, and automatically-configured as possible. If all/most dependencies and users adhere to the "new" (cmake >=3.0)  way, things are *way* simpler and more robust to maintain in the long term  :-)

cmake/GtsamBuildTypes.cmake
```diff
@@ -54,26 +53,11 @@ set(GTSAM_CMAKE_SHARED_LINKER_FLAGS_PROFILING "${CMAKE_SHARED_LINKER_FLAGS_RELEA
 set(GTSAM_CMAKE_MODULE_LINKER_FLAGS_PROFILING "${CMAKE_MODULE_LINKER_FLAGS_RELEASE}" CACHE STRING "Linker flags during profiling builds.")
 set(GTSAM_CMAKE_EXE_LINKER_FLAGS_PROFILING    "${CMAKE_EXE_LINKER_FLAGS_RELEASE}" CACHE STRING "Linker flags during profiling builds.")
 
-mark_as_advanced(GTSAM_CMAKE_C_FLAGS_TIMING GTSAM_CMAKE_CXX_FLAGS_TIMING GTSAM_CMAKE_EXE_LINKER_FLAGS_TIMING 
+mark_as_advanced(GTSAM_CMAKE_EXE_LINKER_FLAGS_TIMING
                  GTSAM_CMAKE_SHARED_LINKER_FLAGS_TIMING GTSAM_CMAKE_MODULE_LINKER_FLAGS_TIMING
-                 GTSAM_CMAKE_C_FLAGS_PROFILING GTSAM_CMAKE_CXX_FLAGS_PROFILING GTSAM_CMAKE_EXE_LINKER_FLAGS_PROFILING 
+                 GTSAM_CMAKE_C_FLAGS_PROFILING GTSAM_ GTSAM_CMAKE_EXE_LINKER_FLAGS_PROFILING
```
dellaert: Is this a typo?

cmake/GtsamBuildTypes.cmake
```diff
@@ -86,15 +70,16 @@ set(CMAKE_EXE_LINKER_FLAGS_PROFILING ${GTSAM_CMAKE_EXE_LINKER_FLAGS_PROFILING})
...
+        list(APPEND GTSAM_COMPILE_OPTIONS_PUBLIC "-ftemplate-depth=1024")
     endif()
 endif()
 
 if (NOT MSVC)
   option(GTSAM_BUILD_WITH_MARCH_NATIVE  "Enable/Disable building with all instructions supported by native architecture (binary may not be portable!)" ON)
   if(GTSAM_BUILD_WITH_MARCH_NATIVE)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=native")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
+    # Add as public flag so all dependant projects also use it, as required
```
dellaert: dependent

cmake/GtsamBuildTypes.cmake
```diff
@@ -86,15 +70,16 @@ set(CMAKE_EXE_LINKER_FLAGS_PROFILING ${GTSAM_CMAKE_EXE_LINKER_FLAGS_PROFILING})
...
+    # by Eigen to avid crashes due to SIMD vectorization:
```
dellaert: avoid

cmake/GtsamBuildTypes.cmake
```diff
@@ -161,3 +146,23 @@ function(gtsam_assign_all_source_folders)
...
+  target_compile_definitions(${target_name_} PRIVATE ${GTSAM_COMPILE_DEFINITIONS_PRIVATE})
+  target_compile_definitions(${target_name_} PUBLIC ${GTSAM_COMPILE_DEFINITIONS_PUBLIC})
+  if (NOT "${GTSAM_COMPILE_OPTIONS_PUBLIC}" STREQUAL "")
+    target_compile_options(${target_name_} PUBLIC ${GTSAM_COMPILE_OPTIONS_PUBLIC})
+  endif()
+  target_compile_options(${target_name_} PRIVATE ${GTSAM_COMPILE_OPTIONS_PRIVATE})
+
+  # Also, ensure the compiler uses at least C++11.
+  # The use of target_compile_features() is preferred since it will be not in
+  # conflict with a more modern C++ standard, if used in a client program.
```
dellaert: I like this!

cmake/GtsamBuildTypes.cmake
```diff
@@ -161,3 +146,23 @@ function(gtsam_assign_all_source_folders)
...
+  if(NOT CMAKE_VERSION VERSION_LESS 3.8)
```
dellaert: Can we replace double negative with one positive :-) ?

cmake/GtsamBuildTypes.cmake
```diff
@@ -161,3 +146,23 @@ function(gtsam_assign_all_source_folders)
...
+  	target_compile_features(${target_name_} PUBLIC cxx_std_11)
+  else()
+  	if (NOT MSVC)
+  		target_compile_options(${target_name_} PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-std=c++11>)
```
dellaert: Add comment ?

gtsam/CMakeLists.txt
```diff
@@ -101,17 +101,16 @@ message(STATUS "Building GTSAM - shared: ${BUILD_SHARED_LIBS}")
...
 # Other deps:
 target_link_libraries(gtsam PUBLIC ${GTSAM_ADDITIONAL_LIBRARIES})
-target_compile_definitions(gtsam PRIVATE ${GTSAM_COMPILE_DEFINITIONS_PRIVATE})
-target_compile_definitions(gtsam PUBLIC ${GTSAM_COMPILE_DEFINITIONS_PUBLIC})
-if (NOT "${GTSAM_COMPILE_OPTIONS_PUBLIC}" STREQUAL "")
-  target_compile_options(gtsam PUBLIC ${GTSAM_COMPILE_OPTIONS_PUBLIC})
-endif()
-target_compile_options(gtsam PRIVATE ${GTSAM_COMPILE_OPTIONS_PRIVATE})
+
+# Apply build flags:
```
dellaert: in comment, say where that function is defined...

cmake/GtsamBuildTypes.cmake
jlblancoc: Sure, I would prefer that too! :-)
The fact is that CMake ["if"](https://cmake.org/cmake/help/v3.5/command/if.html) doesn't provide a  greater-or-equal operator... There is the option of the double negative... or the trick of using `VERSION_GREATER 3.7.9` with the fictional version number (sigh). 

Preference here?

jlblancoc:
PS: I'll come back to this issue soon and improve things. Thanks for the patience!

cmake/GtsamBuildTypes.cmake
dellaert: just keep as is...

jlblancoc:
Please, checkout the latest version. Options for each configuration are now editable from the cache, see attached screenshot. As CMake does, we have separate variables for preprocessor macros, and for all other compiler flags, and for "private" and "public" properties. 
* private: flags do only affect to gtsam projects
* public: if they *must* be propagated to user projects (e.g. "-Wall" should not be propagated, the requirement for >=C++11 should be).

Note that the "final option variables" (those **without** the suffix `_${build_type}`) are not editable (it's mentioned in the cache "helpstring"), so users should change the ones with the per-configuration suffix. Even with this limitation, it's been far from straightforward to set this up, but I think the result is much better now.... 
Thoughts?

PS: Still *not* tested on Windows.

Screenshot from 2019-06-07 00-51-35

jlblancoc:
It seems there are tests [failing](https://travis-ci.com/borglab/gtsam/jobs/206178946) now, will look at it...

Apparently, there are errors after not finding test datasets (?). Any wild guess of possible reasons would be appreciated... otherwise, will dig into the code. I'm not familiar (yet) with the so many tests in gtsam ;-)

dellaert:
Yeah, I saw that. Maybe an option problem screwed up something in `dataset.h`

jlblancoc:
All tests pass now. Let's see if Travis agrees...

dellaert:
Segfaults? Restarted last build to confirm.

dellaert:
Still segfauls :-) I recommend checking detailed build commands between develop and this branch. Seems serialization-related.

jlblancoc:
The only difference I found (for both, gcc and clang) was `-std=g++11` instead of `-std=c++11`. 
Let's see if that change makes a difference...

PS: If it works and someone wants to merge, feel free of using "squash merge" to simplify all the changes into a single commit... or wait until I rebase the commits.

gtsam/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+project(gtsam LANGUAGES CXX)
```
chrisbeall: Just want to make sure I understand what you did here: We now have a top-level project `GTSAM`, and two sub-projects `gtsam` and `gtsam_unstable`, and the main benefit of having the sub-projects is that we don't have to write `/gtsam/`, or is there something else less obvious that I'm missing?

dellaert:
Hold up: I’m getting the same segfaults on another branch: https://travis-ci.com/borglab/gtsam/jobs/206627376. Might not be related to this PR at all! Only happens on Linux/clang...

gtsam/CMakeLists.txt
jlblancoc: Exactly, the top-level was named all upper case just to make it different from the actual library. 

I normally add these `project()`s for many subdirectories because it's handy: for some IDEs (e.g. MSVC) CMake generates "solutions" that only include the dependent projects, so for large projects it's much faster to load only that part of the project.

Anyway, that's **not** why I added the two projects: it was the only way I found to prevent MSVC to fail, it was done here: https://github.com/borglab/gtsam/commit/65f442ce53b4dd7dd29435e19eb1089895a1b235

Unfortunately, I didn't documented very well (at all, actually...) **which** was the error and can't remember it. Could checkout the former version just to test, will try to do it if time permits. 

Bottom line anyway: something in the targets configuration broke with the new variables and the only way to make CMake to get things right was to add these project() definitions.

@chrisbeall : What do you think of the new build CMake variables? I'm afraid you will think there are too many cache entries now, but in this way things should be much easier when packaging GTSAM as a Debian packages with CMake exported projects and such, for example.

jlblancoc:
> Hold up: I’m getting the same segfaults on another branch: https://travis-ci.com/borglab/gtsam/jobs/206627376. Might not be related to this PR at all! Only happens on Linux/clang...

wow... well, I don't know if that's good news. There is something we can try: to run valgrind at Travis. 
I have that set-up for another project, see: https://github.com/MRPT/mrpt/blob/master/tests/CMakeLists.txt#L111-L125

Could be done in a separate PR to see if it detects something weird. Having `make check_valgrind` is useful in itself anyway.

dellaert:
Not sure it's a memory issue, but adding a valgrind cmake target would be cool.
For now, I will push a travis branch with Debug on, and if that does provide a clue on how to resolve the issue, I will (atempt to) turn off linux-clang-tests and create a new issue.

gtsam/CMakeLists.txt
chrisbeall: Ok, that makes sense. Definitely useful to add a comment for those kinds of workarounds in the future.
I think the new CMake variables are good, although I haven't yet tested building another project against it. 
Have you had a chance to test this on Windows? Seems we always get surprises there.

jlblancoc:
Closing, there are conflicts that have been more easily solved by cherry-picking and rebasing on a fork. 
New PR is #65.

PR #28, opened 2019-05-29
Ensure users do not mix up Eigen versions

jlblancoc:
Since GTSAM uses, by default, an embedded copy of Eigen, there exists the possibility of user programs (also depending on Eigen) to include a different version of it (for example, from `/usr/include`).

To ensure this does not happen, this PR:
* Detects Eigen version parsing the corresponding .h file (if it fails to parse it, nothing of the following will take effect)
* Prints the detected Eigen version in the CMake final summary. 
* Defines the detected Eigen version in gtsam/config.h
* Uses an `static_assert()` in `base/Vector.h` to detect for potential version mismatches.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,29 @@ else()
 	set(GTSAM_EIGEN_INCLUDE_FOR_BUILD "${CMAKE_SOURCE_DIR}/gtsam/3rdparty/Eigen/")
 endif()
 
+# Detect Eigen version:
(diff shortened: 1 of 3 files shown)


dellaert:
Looks good to me, on cmake PR's I will typically defer to @chrisbeall to do final check and merge (or, if Chris does a cmake PR, to you :-))

chrisbeall:
I'll test, but might not have time until Thursday evening or Friday.

jlblancoc:
Rebased on top of develop so Travis con run on this PR too.

PR #29, opened 2019-05-29
Updated python package information

varunagrawal:
Only text based updated. No code functionality change, hence no unit test.


varunagrawal:
@dellaert this should be a quick PR. I noticed this when I was trying to make sure I had the correct `gtsam` package being installed and it took all of 2 minutes, so why not?
P.S. makes it look more professional!

dellaert:
Thanks!!

Issue #31, opened 2019-05-30
Make navigation expressions easier


dellaert:
We should make expression overloads for NavState related things just like Brice's PR #18

dellaert:
Addressed with PR #32

PR #32, opened 2019-05-30
Feature/navigation expressions

dellaert:
Closes issue #31 

Added a new file navigation/expressions.h that creates some handy expression overloads. see testNavExpressions.cpp for two examples.


dellaert:
Tagging @bricerebsamen as FYI

PR #33, opened 2019-05-30
DoglegOptimizer: Throw exception if infeasible test

varunagrawal:


dellaert:
Note, will create new branch for related issue #34

PR #35, opened 2019-05-31
Add Travis CI

jlblancoc:
This adds Travis CI support. It builds gtsam, gtsam_unstable, all examples, all tests, and run them, for: 
* Ubuntu Xenial, with default gcc version.
* Ubuntu Xenial, with default clang version.
* Ubuntu Xenial, with gcc-8
* OSX with gcc.
* OSX with clang.

Due to build time limitations in the free version of Travis, it has been split into different jobs, some build only examples, other tests. 

Example output (from my fork): https://travis-ci.org/jlblancoc/gtsam/builds/539620367

Please, enable the GitHub App "Travis CI" to allow TravisCI to actually run the tests here (admins should have received a request email earlier today).

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -24,14 +24,14 @@ function build_and_test ()
 
   # gcc is too slow and we have a time limit in Travis CI: selective builds.
   if [ "$BUILD_EXAMPLES" == "1" ]; then
-    GTSAM_BUILD_EXAMPLES_ALWAYS=FALSE
+    GTSAM_BUILD_EXAMPLES_ALWAYS=ON
   else
-    GTSAM_BUILD_EXAMPLES_ALWAYS=TRUE
+    GTSAM_BUILD_EXAMPLES_ALWAYS=OFF
   fi
   if [ "$RUN_TESTS" == "1" ]; then
-    GTSAM_BUILD_TESTS=FALSE
+    GTSAM_BUILD_TESTS=ON
   else
-    GTSAM_BUILD_TESTS=TRUE
+    GTSAM_BUILD_TESTS=OFF
   fi
 
   cmake $SOURCE_DIR \
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+set -e   # Make sure any error makes the script to return an error code
+set -x   # echo
+
+SOURCE_DIR=`pwd`
+BUILD_DIR=build
+
+#CMAKE_C_FLAGS="-Wall -Wextra -Wabi -O2"
+#CMAKE_CXX_FLAGS="-Wall -Wextra -Wabi -O2"
+
+function build_and_test ()
+{
+  #env
+  git clean -fd || true
+  rm -fr $BUILD_DIR || true
+  mkdir $BUILD_DIR && cd $BUILD_DIR
+
+  if [ ! -z "$GCC_VERSION" ]; then
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-$GCC_VERSION 60 \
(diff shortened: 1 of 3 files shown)


dellaert:
Jose, that's awesome! By coincidence, I had already authorized travis yesterday so I think we should be good. 

The most important and problematic flag to test I think is GTSAM_TYPEDEF_POINTS_TO_VECTOR, if we were to add any. By default off, so it might be broken by now.

chrisbeall:
-I think we should run tests for every config we build. Most serious errors happen here; I don't consider them optional. Also, gtsam users are very likely to report build errors (because they block), but we're far less likely to hear about test errors. 
- If build time is limited, drop the mac os + gcc config. Hardly anybody uses that, and ubuntu + gcc should catch vast majority of similar build issues.
- On Jenkins, I had taken the approach of running just one or two configs on PRs, and many combinations on develop. From http://hawkertempest.cc.gatech.edu:8080/job/gtsam-ubuntu/
image

dellaert:
Agree with all of this. But I'm anxious to see *some* travis CI and we could add those options in a next PR...

jlblancoc:
Good! 

There are several ways to use Travis CI... using Docker images to test on many different OS versions, etc. 
But the one in this PR is one of the simplest we managed to set up for another project (MRPT). This is the "old" configuration without docker, but IMO it's much simpler to understand and to extend. 

So, perhaps you could continue building @chrisbeall 's matrix on top of this PR by adding lines to `.travis.yml` and sending new variables to be forwarded to cmake in `.travis.sh`. Or if there is something else already half baked, feel free of discarding this PR and use something else.

.travis.sh
```diff
@@ -0,0 +1,52 @@
...
+  else
+    GTSAM_BUILD_TESTS=OFF
+  fi
+
+  cmake $SOURCE_DIR \
+      -DGTSAM_BUILD_EXAMPLES_ALWAYS=$GTSAM_BUILD_EXAMPLES_ALWAYS \
+      -DGTSAM_BUILD_TESTS=$GTSAM_BUILD_TESTS
+
+  # Actual build:
+  make -j2
```
varunagrawal: Can we increase this to 4? I don't know how much of a difference it will make, but my general experience is that `-j4` is a great middle ground between performance and resource efficiency.

.travis.yml
```diff
@@ -0,0 +1,49 @@
+language: cpp
+cache: ccache
+sudo: required
+dist: xenial
+
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
```
varunagrawal: Very cool! I didn't know this PPA existed.

.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+  - bash .travis.sh
+
+env:
+  global:
+    - MAKEFLAGS="-j 2"
+    - CCACHE_SLOPPINESS=pch_defines,time_macros
+
+matrix:
+  include:
+  - compiler: gcc
```
varunagrawal: As per the travis docs, the default gcc version is 5.4. Should we increase this gcc 6.3 since it seems to be a bit faster as per [this](https://www.phoronix.com/scan.php?page=article&item=gcc-7-january&num=2). We can always add gcc 5.4 with examples off as an additional build matrix element.

.travis.sh
jlblancoc: Actually travis only provides 2 cores per job, that's the reason of the "-j2" ;-)

dellaert:
Jose, please resolve conflict and I'll merge. I tried doing it with web interface but it hangs...

jlblancoc:
Done! :+1:

PR #40, opened 2019-06-01
Factor graph update

dellaert:
Re-factored std::enable_if use, and added a method `add_factors` that will be used to refactor iSAM2.

Some of the push_back methods are complex, even using my new templates. If someone wants to take a look whether some of these can be simplified using C++11 please do :-) It will be in another PR.


dellaert:
Checks only fail because of time limit, will merge.

PR #42, opened 2019-06-02
iSAM2 refactor

dellaert:
This is a pretty substantial refactor of iSAM2, which mainly did:
- broke up `recalculate` into a batch and incremental method
- broke up `update` into manageable chunks, now in `ISAM2-impl.h`

This was mostly cosmetic. There are only two functional changes I made:
- I discovered `unusedIndices` == `unusedKeys`, so I deleted obsolete `unusedIndices`
- I discovered `markedRelinMask` == `relinKeys`, so I deleted obsolete `markedRelinMask`

would be nice if Michael and Jose could concur.


varunagrawal:
Yes! Guess the frustration with that function was real.

PR #44, opened 2019-06-02
Minor MSVC fixes

jlblancoc:
* `-march=native` is not for MSVC
* Nicer grouping of targets in IDE "folders"

### DIFF
--- a/cmake/GtsamBuildTypes.cmake
+++ b/cmake/GtsamBuildTypes.cmake
@@ -90,10 +90,12 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
     endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


chrisbeall:
Oops! But we don't want Windows users to miss out on the huge speedup completely, so if there's no direct equivalent flag, then we should try to enable `AVX`, `FMA`, etc. directly? [Something like this](https://stackoverflow.com/a/42755665) maybe?
```
CHECK_CXX_COMPILER_FLAG("/arch:AVX" COMPILER_OPT_ARCH_AVX_SUPPORTED)
if(COMPILER_OPT_ARCH_AVX_SUPPORTED)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX")
endif()
```

dellaert:
I will merge, I propose we leave the windows improvements for another PR.

PR #45, opened 2019-06-02
Fix/travis timeout

dellaert:
I refactored slightly, to use more direct configuration of cmake.
I also disabled unstable for gcc tests build, it comes in right under 50 mins ....
 
One possible speedup is using the gold linker, see http://www.bitsnbites.eu/faster-c-builds/


chrisbeall-nuro:
Btw, is `ccache` actually working? Looking at the last few builds on develop of gcc + examples, they all took 24+ minutes, which is not what I would expect. For example, the build after you merged PR #40 should have gone much faster because it didn't have to rebuild `base`, `geometry`, etc.

dellaert:
I don’t think we enabled ccache yet, is described here. I can add it to this PR. Otherwise ok?

dellaert:
https://docs.travis-ci.com/user/caching/

dellaert:
By the way, caching will only help for small changes, right? I think we still have to make sure that a full rebuild fits within 50 minutes.

dellaert:
OK, 
- I saw ccache was enable by Jose, nice
- I added Mac OS install
- it seems ccache also needs "Build pushed branches" ON, which I did.

chrisbeall:
Hmm, I thought it was enabled because we have:
https://github.com/borglab/gtsam/blob/823c8bd5e848892a2cc6a7186243aee7f556dbe4/.travis.yml#L1-L2
That's exactly what https://docs.travis-ci.com/user/caching says to do. Our cmake config *does* pick it up, and says it will use it.
Correct, it'll only help with small changes. As long as you're not touching `base` and `geometry`, speedup is huge.

chrisbeall:
Just fyi: As far as gtsam is concerned, ccache is turned on by default if we detect it (unless we're on Windows or generate an Xcode project):
https://github.com/borglab/gtsam/blob/823c8bd5e848892a2cc6a7186243aee7f556dbe4/CMakeLists.txt#L71-L73

chrisbeall:
Ah, there's an easy way to check whether cache was used. At the end of shell script, run `ccache -s` to print statistics.

dellaert:
It seems to be used after I enabled branches. First two builds were 3/4 mins only!
I'll add `ccache -s` and remove the forcing flag, making sure to check whether it actually works.

dellaert:
PS, `brew update` takes almost 5 mins all by itself. Are we sure it's needed, @jlblancoc  ?

jlblancoc:
Hmm... we normally use another setup (CircleCI+docker), so can't recall the
details in Travis, but it looks like a problem in its file cache across
matrix configurations, doesn't it?

Worth a careful read:

https://docs.travis-ci.com/user/caching/#caches-and-build-matrices

chrisbeall:
Looks like it definitely worked on Linux, but mac is still slow.
If I understand the documentation correctly, I don't think we run the risk of poisoning the cache between configurations, because we have unique OS, compiler, and environment variables in each build.

chrisbeall:
But we get two sets of builds now; that's not good :-( 
image

jlblancoc:
> But we get two sets of builds now; that's not good :-(


That's actually normal in travis: to build each branch, and also prs. You
can disable branch builds if so desired in travis project settings.

dellaert:
Finally the clang one passed on Mac and wrote the cache. This story is not yet over I think, but I will merge for now.

PR #47, opened 2019-06-04
Fix/bayes tree

dellaert:
Fixed a number of deprecated instances, and made sure travis catches them.


dellaert:
Travis is timing out again but I'll merge this as Jenkins is happy at least.

PR #49, opened 2019-06-06
Feature/stage travis

dellaert:
Added a compile stage before the matrix of features is built, to populate the ccache. Works well for Linux, does *still* not seem to work for OSX. Somehow, even though the cache is written by the compile stage, the filenames do not match when the test stage tries to read.

I will cancel the branch build for now, let's see how the CI fares on this PR. But if it still does not work, I have no idea how to fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,16 @@ stages:
 # Compile stage without building examples/tests to populate the caches.
 jobs:
   include:
+# on Mac, GCC
+  - stage: compile
+    os: osx
+    compiler: gcc
+    env: GTSAM_ALLOW_DEPRECATED_SINCE_V4=OFF GTSAM_BUILD_TESTS=OFF GTSAM_BUILD_UNSTABLE=ON GTSAM_BUILD_EXAMPLES_ALWAYS=OFF
+# on Mac, CLANG
+  - stage: compile
+    os: osx
+    compiler: clang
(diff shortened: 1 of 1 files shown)


jlblancoc:
Awesome! Adding stages is the best way to avoid timeouts. If it's needed, another intermediary stage could be created: 1) gtsam 2) gtsam_unstable 3) examples / tests. 
Let's wait for Travis' output...

.travis.yml
```diff
@@ -26,32 +26,52 @@ install:
...
+compiler:
+  - gcc
+  - clang
+env:
+  global:
+    - MAKEFLAGS="-j 2"
+    - CCACHE_SLOPPINESS=pch_defines,time_macros
+  matrix:
+    - GTSAM_ALLOW_DEPRECATED_SINCE_V4=OFF GTSAM_BUILD_TESTS=OFF GTSAM_BUILD_UNSTABLE=ON GTSAM_BUILD_EXAMPLES_ALWAYS=ON
+    - GTSAM_ALLOW_DEPRECATED_SINCE_V4=OFF GTSAM_BUILD_TESTS=ON GTSAM_BUILD_UNSTABLE=ON GTSAM_BUILD_EXAMPLES_ALWAYS=OFF
```
jlblancoc: I *think* that this "matrix" block could (and should?) be removed.
At least, it could be removed to check whether the list of generated jobs changes; if not, it would be clear it has no effects.

See: https://github.com/travis-ci/docs-travis-ci-com/issues/1500

dellaert: No, I think it works as expected: it generates the eight builds: 2 platforms, 2 compilers, 2 environment settings

PR #56, opened 2019-06-10
Feature/travis global

dellaert:
This PR does two things:
- runs travis in debug mode and release mode
- uses global env variables to make matrix a bit easier to read
- compiles examples in first stage to reduce matrix size in second stage


dellaert:
Travis works as expected although cache still not working well. Build use clang Linux in Release has segfaults. Will merge this but create an issue.

Issue #57, opened 2019-06-11
Unit tests Segfault w clang on Linux 


dellaert:
clang Linux in Release has segfaults in 4 unit tests. Seems serialization related. I am not running clang on Linux so would appreciate some help from someone who does or has a docker image handy.

jlblancoc:
This is what valgrind says: 

```
make testSerializationLinear.valgrind
[  4%] Built target CppUnitLite
[ 36%] Built target metis
[100%] Built target gtsam
[100%] Built target testSerializationLinear
Scanning dependencies of target testSerializationLinear.valgrind
==13410== Memcheck, a memory error detector
==13410== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==13410== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==13410== Command: /home/jlblanco/code/gtsam-jlblancoc/build-clang/gtsam/linear/tests/testSerializationLinear
==13410== 
==13410== Invalid read of size 8
==13410==    at 0x539BCC: void boost::archive::detail::save_pointer_type<boost::archive::text_oarchive>::polymorphic::save<gtsam::noiseModel::Diagonal>(boost::archive::text_oarchive&, gtsam::noiseModel::Diagonal&) (in /home/jlblanco/code/gtsam-jlblancoc/build-clang/gtsam/linear/tests/testSerializationLinear)
==13410==    by 0x539B5F: boost::archive::detail::oserializer<boost::archive::text_oarchive, boost::shared_ptr<gtsam::noiseModel::Diagonal> >::save_object_data(boost::archive::detail::basic_oarchive&, void const*) const (in /home/jlblanco/code/gtsam-jlblancoc/build-clang/gtsam/linear/tests/testSerializationLinear)
==13410==    by 0x549BA42: ??? (in /usr/lib/x86_64-linux-gnu/libboost_serialization.so.1.65.1)
==13410==    by 0x547EF0: boost::archive::detail::oserializer<boost::archive::text_oarchive, gtsam::JacobianFactor>::save_object_data(boost::archive::detail::basic_oarchive&, void const*) const (in /home/jlblanco/code/gtsam-jlblancoc/build-clang/gtsam/linear/tests/testSerializationLinear)
==13410==    by 0x549BA42: ??? (in /usr/lib/x86_64-linux-gnu/libboost_serialization.so.1.65.1)
[...]

PR #61, opened 2019-06-12
Proper cacheing for travis

dellaert:
Use command line argument for script so that cache names are not polluted by environment variables. This should allow ccache to use caches from build stage for testing stage.

We'll see from the travis builds here whether it worked :-)


dellaert:
@jlblancoc you like? I'm also turning off clang on Linux until we solved issue #57

PR #63, opened 2019-06-12
Cython wrapper for all release modes

varunagrawal:
This PR upates the CMake files so that the generated cython wrapper only appends the build type postfix to the top-level `cython` directory in the install path, rather than appending it to the sub-directories or `.so` files since that would cause import issues when using the cython wrapper in any mode other than `Release`.

This was a doozy, but it works and should be a huge step forward with respect to improvements in the python wrapper.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/63)

### DIFF
--- a/gtsam/base/debug.h
+++ b/gtsam/base/debug.h
@@ -47,6 +47,7 @@ namespace gtsam {
   // Non-guarded use led to crashes, and solved in commit cd35db2
   bool GTSAM_EXPORT guardedIsDebug(const std::string& s);
   void GTSAM_EXPORT guardedSetDebug(const std::string& s, const bool v);
+  bool GTSAM_EXPORT isDebugVersion();
 }
 
 #undef ISDEBUG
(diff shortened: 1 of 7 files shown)


dellaert:
Took off Mandy and Jose as reviewers, suggest you add Duy and Matthew

gtsam/base/debug.h
```diff
@@ -47,6 +47,7 @@ namespace gtsam {
   // Non-guarded use led to crashes, and solved in commit cd35db2
   bool GTSAM_EXPORT guardedIsDebug(const std::string& s);
   void GTSAM_EXPORT guardedSetDebug(const std::string& s, const bool v);
+  bool GTSAM_EXPORT isDebugVersion();
```
dellaert: Separate from two other functions and add comment

dellaert:
Adding @jlblancoc again because *maybe* you should be merging this into his PR, not straight into develop? Lots of cmake changes there... I'll let him comment: if the two PR's are not conflicting, you can just target develop as is.

gtsam/base/debug.h
varunagrawal: Done.

varunagrawal:
I can rebase off the branch once it is approved/merged and make the needed changes, if that works?

dellaert:
Resolve conflicts so we can merge?

varunagrawal:
Rebase done. It was pretty quick. :smile:

dellaert:
Ok let’s wait for travis

Issue #38, opened 2019-06-01
Warning in SmartRangeExample_plaza1


dellaert:
Recent travis run revealed: 
```
[ 96%] Building CXX object gtsam_unstable/examples/CMakeFiles/SmartRangeExample_plaza1.dir/SmartRangeExample_plaza1.cpp.o
In file included from /home/travis/build/borglab/gtsam/gtsam/3rdparty/Eigen/Eigen/Core:450:0,
                 from /home/travis/build/borglab/gtsam/gtsam/3rdparty/Eigen/Eigen/Dense:1,
                 from /home/travis/build/borglab/gtsam/gtsam/base/OptionalJacobian.h:22,
                 from /home/travis/build/borglab/gtsam/gtsam/base/Matrix.h:25,
                 from /home/travis/build/borglab/gtsam/gtsam/base/Manifold.h:22,
                 from /home/travis/build/borglab/gtsam/gtsam/geometry/BearingRange.h:21,
                 from /home/travis/build/borglab/gtsam/gtsam/geometry/Pose2.h:23,
                 from /home/travis/build/borglab/gtsam/gtsam_unstable/examples/SmartRangeExample_plaza1.cpp:20:
/home/travis/build/borglab/gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/DenseStorage.h: In member function ‘gtsam::Point2 gtsam::SmartRangeFactor::triangulate(const gtsam::Values&) const’:
/home/travis/build/borglab/gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/DenseStorage.h:194:66: warning: ‘*((void*)& bestCircle2 +16)’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     DenseStorage(const DenseStorage& other) : m_data(other.m_data) {
                                                                  ^
In file included from /home/travis/build/borglab/gtsam/gtsam_unstable/examples/SmartRangeExample_plaza1.cpp:43:0:
/home/travis/build/borglab/gtsam/gtsam_unstable/slam/SmartRangeFactor.h:103:30: note: ‘*((void*)& bestCircle2 +16)’ was declared here
     boost::optional<Circle2> bestCircle2;
                              ^
```

dellaert:
Fixed in PR #64

PR #64, opened 2019-06-12
Fix/38 smart range examples 2

dellaert:
Fixing issue #38, also fixed file paths and ILS due to too loose BetweenFactors, and an issue where duplicate range measurements were added.


dellaert:
Branch CI passed except one time out. Uncontroversial and fixes issue #38 so will merge.

PR #66, opened 2019-06-13
[Do not review] pybind wrapper

thduynguyen:
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/66)


dellaert:
We need to undo commit 037c0cf for 4.1

thduynguyen:
> We need to undo commit [037c0cf](https://github.com/borglab/gtsam/commit/037c0cf43145d997b6eb2f8d977f2343c5503401) for 4.1

Already did: bc85765

thduynguyen:
FYI, a better setup.py script is in place. I changed pybind package structure and haven't fixed the tests so they will fail. But you can try:

```
mkdir build
cd build
cmake ..
python setup.py develop
python -c "import gtsam_py as gtsam; gtsam.Pose3().print_('')"
```
Still need to fix the py util scripts and tests folder structure.

The setup.py script largely follows `https://github.com/pybind/cmake_example`. It basically just runs cmake with relevant options to enable the python build. I made some changes so that we can run it inside the build folder instead of the main source. A caveat of this strategy is that some cmake options are silently changed.

dellaert:
Hey, compile fails:
```
/var/lib/jenkins/workspace/gtsam-ubuntu-github-pullrequest-default/tests/testNonlinearOptimizer.cpp:437:10: fatal error: wrap/utilities.h: No such file or directory
 #include <wrap/utilities.h>
          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```
I'm not sure why that test would access anything in wrap, anyway.

thduynguyen:
The pybind build/setup/install system is fully functional now. If you have time to check, there are several ways to run the tests:
1. Build with `GTSAM_BUILD_PYTHON=ON` as usual.
2. While in the build folder, run `python -m unittest discover`
  - You can edit the source `.py` file and simply rerun the test. I symlinked them to the build folder to facilitate fast development. Might need to run `cmake ..` if files are deleted or added.
  - Or just run `python` then `from gtsam_py import gtsam` and play around.
3. Do `make install` and `cd` to the install folder.
4. Here, you can:
  - Run `python setup.py test` to see what happen.
  - Run `python setup.py install` to install `gtsam_py` to the current python environment. (It's a good idea to create a virtual environment first.)

Next, I'll clean up the obsolete stuff and add documentation.

dellaert:
Duy, could you do another PR that just adds the pybind stuff? Also, should we do that with a github submodule rather than copying the pybind files?

thduynguyen:
> Duy, could you do another PR that just adds the pybind stuff? Also, should we do that with a github submodule rather than copying the pybind files?

How about doing this in the current PR? I was debating between copy vs submodule and realized we have lots of copies in 3rdparty.

dellaert:
Doing it in the current PR makes it very cumbersome to review. There are 200+ files that were added

dellaert:
I agree on the copies. @varunagrawal suggested we copy how pytorch does this, which apparently has many submodules.

CMakeLists.txt
```diff
@@ -112,8 +99,11 @@ if(GTSAM_INSTALL_MATLAB_TOOLBOX AND GTSAM_TYPEDEF_POINTS_TO_VECTORS)
     message(FATAL_ERROR "GTSAM_INSTALL_MATLAB_TOOLBOX and GTSAM_TYPEDEF_POINTS_TO_VECTORS are both enabled. For now, the MATLAB toolbox cannot deal with this yet.  Please turn one of the two options off.")
 endif()
 
-if(GTSAM_INSTALL_CYTHON_TOOLBOX AND GTSAM_TYPEDEF_POINTS_TO_VECTORS)
-    message(FATAL_ERROR "GTSAM_INSTALL_CYTHON_TOOLBOX and GTSAM_TYPEDEF_POINTS_TO_VECTORS are both enabled. For now, the CYTHON toolbox cannot deal with this yet.  Please turn one of the two options off.")
+if(GTSAM_BUILD_PYTHON)
+	if (NOT GTSAM_TYPEDEF_POINTS_TO_VECTORS)
+		message(FATAL_ERROR "GTSAM_BUILD_PYTHON requires GTSAM_TYPEDEF_POINTS_TO_VECTORS to be enabled but it is not.")
```
dellaert: I will modify Travis to add an extra compile stage in “special” to make sure GTSAM_TYPEDEF_POINTS_TO_VECTORS actually still compiles. Or even make this the default. This has to succeed before this can merge.

CMakeLists.txt
```diff
@@ -584,17 +556,20 @@ print_config_flag(${GTSAM_ALLOW_DEPRECATED_SINCE_V4}   "Deprecated in GTSAM 4 al
...
 
 message(STATUS "MATLAB toolbox flags                                      ")
 print_config_flag(${GTSAM_INSTALL_MATLAB_TOOLBOX}      "Install matlab toolbox         ")
 
-message(STATUS "Cython toolbox flags                                      ")
-print_config_flag(${GTSAM_INSTALL_CYTHON_TOOLBOX}      "Install Cython toolbox         ")
-if(GTSAM_INSTALL_CYTHON_TOOLBOX)
+message(STATUS "Python module flags                                       ")
+
+if(GTSAM_PYTHON_WARNINGS)
```
dellaert: Just always print?

cmake/CMakeLists.txt
```diff
@@ -17,11 +17,8 @@ install(FILES
   GtsamBuildTypes.cmake
   GtsamMakeConfigFile.cmake
   GtsamMatlabWrap.cmake
-  GtsamPythonWrap.cmake
-  GtsamCythonWrap.cmake
+  GtsamPybindWrap.cmake
   GtsamTesting.cmake
```
dellaert: How are python and jumpy found now?

gtsam.h
```diff
@@ -2558,8 +2558,8 @@ pair<gtsam::NonlinearFactorGraph*, gtsam::Values*> load2D(string filename,
 pair<gtsam::NonlinearFactorGraph*, gtsam::Values*> load2D(string filename,
     gtsam::noiseModel::Diagonal* model);
 pair<gtsam::NonlinearFactorGraph*, gtsam::Values*> load2D(string filename);
-pair<gtsam::NonlinearFactorGraph*, gtsam::Values*> load2D_robust(string filename,
-    gtsam::noiseModel::Base* model);
+// pair<gtsam::NonlinearFactorGraph*, gtsam::Values*> load2D_robust(string filename,
```
dellaert: Oh?

gtsam/CMakeLists.txt
```diff
@@ -116,7 +116,10 @@ set_target_properties(gtsam PROPERTIES
     OUTPUT_NAME         gtsam
     CLEAN_DIRECT_OUTPUT 1
     VERSION             ${gtsam_version}
-    SOVERSION           ${gtsam_soversion})
+    SOVERSION           ${gtsam_soversion}
+    INSTALL_RPATH       "$ORIGIN"
```
dellaert: No idea - should have Varun and Jose look at this as well

thduynguyen:
Closed to move to a [new branch](https://github.com/borglab/gtsam/tree/feature/py_wrapper_boost_shared_ptr), which should be PRed to develop after Matlab stuff is done.

ProfFan:
This is awesome!

PR #67, opened 2019-06-13
Housekeeping

varunagrawal:
A bunch of niceness regarding README files as well as an ISSUE template for Github to start out with.


varunagrawal:
I wanted to get your opinion on what was a good issue template format before committing too much time to it. I can start off with the default message from Bitbucket.

varunagrawal:
@dellaert updated the issue templates.

dellaert:
Cool! Where are the two markdown files referenced?

varunagrawal:
They should be in `.github/ISSUE_TEMPLATE`. There are different templates for the kind of issue to be reported (bug report vs feature request vs general support).

dellaert:
Ah, but they're not... They're in `.github`

varunagrawal:
Good catch. I've updated them.

PR #68, opened 2019-06-14
Feature/vector value print

MandyXie:
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/68)


MandyXie:
> Mandy, thanks !!! But I don't see how a simple print change should change so many things? Are you sure you have the latest develop? And can you restrict the PR to just the print?
> Finally, I suggest you make a unit test. To make that work, make a stream operator `<<` that you call from Print, and do the test with stringstream...

Sorry, I will update my pr to the latest develop. However, I have some questions.
First, what do you mean by restricting the PR to just the print?
Second, I tried to overload a stream operator <<, but I don't know how to have it taken two extra arguments which are needed in the print function, const string& str, const KeyFormatter& formatter.

dellaert:
If you look at the "changed files" you see that there are differences all over the file. Did you by chance re-format? I'd prefer if that is not done in GTSAM, to make sure the PR and commits are limited to the new functionality.

You are right about the KeyFormatter - that is an issue that we will have to confront when we switch to streams. I think there is a solution with stream manipulators, let me investigate a bit...

dellaert:
Could you use the `DefaultKeyformatter` for now in the stream operator and ignore the argument in `Print`? Then commit with [ci-skip] in the commit message. I will then add the formatting thing - I know how to do it now...

gtsam/linear/tests/testVectorValues.cpp
```diff
@@ -228,6 +230,24 @@ TEST(VectorValues, vector_sub)
...
+{
+  VectorValues vv;
+  vv.insert(0, (Vector(1) << 1).finished());
+  vv.insert(1, Vector2(2, 3));
+  vv.insert(2, Vector2(4, 5));
+  vv.insert(5, Vector2(6, 7));
+  vv.insert(7, Vector2(8, 9));
+
+  string expected =
+      "VectorValues: : 5 elements\n  0: 1\n  1: 2 3\n  2: 4 5\n  5: 6 7\n  7: 8 9\n";
```
dellaert: Why 2 colons?

gtsam/linear/tests/testVectorValues.cpp
```diff
@@ -228,6 +230,24 @@ TEST(VectorValues, vector_sub)
...
+  stringstream actual;
+  actual << vv;
+
```
dellaert: kill extra return

gtsam/linear/VectorValues.h
```diff
@@ -29,6 +29,7 @@
 
 #include <map>
 #include <string>
+#include <iostream>
```
dellaert: Oh no ! 
Please use `<iosfwd>`

gtsam/linear/VectorValues.h
```diff
@@ -228,6 +229,11 @@ namespace gtsam {
      */
     const_iterator find(Key j) const { return values_.find(j); }
 
+    /**
```
dellaert: Use /// for single line comments (three / is for doxygen)

gtsam/linear/VectorValues.h
```diff
@@ -228,6 +229,11 @@ namespace gtsam {
...
+     * overload operator << to print to stringstream
+     */
+    friend std::ostream& operator<<(std::ostream& ss, const VectorValues& v);
```
dellaert: Omit argument names like so:
```
friend std::ostream& operator<<(std::ostream&, const VectorValues&);
```

gtsam/linear/VectorValues.cpp
```diff
@@ -129,13 +129,39 @@ namespace gtsam {
...
+  bool compare(std::pair<Key, Vector>& lhs, std::pair<Key, Vector>& rhs) {
+    return lhs.first < rhs.first;
+  }
+
+  ostream& operator<<(ostream& ss, const VectorValues& v) {
+    ss << "VectorValues: "
+       << ": " << v.size() << " elements\n";
+    // Change print depending on whether we are using TBB
+#ifdef GTSAM_USE_TBB
+    std::vector<std::pair<Key, Vector>> vec;
```
dellaert: Think about this more: why not use a std::map? Will be automatically sorted.
```
map<Key,Vector> sorted;
for (const auto& key_value : sorted) {
```

gtsam/linear/VectorValues.cpp
```diff
@@ -129,13 +129,39 @@ namespace gtsam {
...
+    for (const auto& key_value : v) {
+      vec.push_back(std::make_pair(key_value.first, key_value.second));
+    }
+    sort(vec.begin(), vec.end(), compare);
+    for (const auto& key_value : vec)
+      ss << "  " << key_value.first << ": " << key_value.second.transpose()
+         << "\n";
+#else
+    for (const auto& key_value : v)
+      ss << "  " << key_value.first << ": " << key_value.second.transpose()
```
dellaert: copy-pasta: keep this after the #endif. Also, use {} for loop.

gtsam/linear/VectorValues.cpp
```diff
@@ -129,13 +129,39 @@ namespace gtsam {
...
-  void VectorValues::print(const string& str, const KeyFormatter& formatter) const {
-    cout << str << ": " << size() << " elements\n";
-    for(const value_type& key_value: *this)
-      cout << "  " << formatter(key_value.first) << ": " << key_value.second.transpose() << "\n";
-    cout.flush();
+  bool compare(std::pair<Key, Vector>& lhs, std::pair<Key, Vector>& rhs) {
+    return lhs.first < rhs.first;
+  }
+
+  ostream& operator<<(ostream& ss, const VectorValues& v) {
```
dellaert: This is an ostream, not stringstream, so use `os`

gtsam/linear/VectorValues.cpp
```diff
@@ -129,13 +129,39 @@ namespace gtsam {
   }
 
   /* ************************************************************************* */
-  void VectorValues::print(const string& str, const KeyFormatter& formatter) const {
-    cout << str << ": " << size() << " elements\n";
-    for(const value_type& key_value: *this)
-      cout << "  " << formatter(key_value.first) << ": " << key_value.second.transpose() << "\n";
-    cout.flush();
+  bool compare(std::pair<Key, Vector>& lhs, std::pair<Key, Vector>& rhs) {
```
dellaert: Will not be needed if you use map

gtsam/linear/VectorValues.cpp
```diff
@@ -129,13 +129,31 @@ namespace gtsam {
...
-      cout << "  " << formatter(key_value.first) << ": " << key_value.second.transpose() << "\n";
-    cout.flush();
+  ostream& operator<<(ostream& os, const VectorValues& v) {
+    os << "VectorValues"
+       << ": " << v.size() << " elements\n";
+    // Change print depending on whether we are using TBB
+#ifdef GTSAM_USE_TBB
+    map<Key, Vector> sorted;
+    for (const auto& key_value : v) {
+      sorted.insert(std::make_pair(key_value.first, key_value.second));
```
dellaert: Use map::emplace

gtsam/linear/VectorValues.cpp
```diff
@@ -129,13 +129,31 @@ namespace gtsam {
...
+    }
+    for (const auto& key_value : sorted)
```
dellaert: Use {}

dellaert:
OK, cool, PR #69 is merged which you can use to send the formatter into the stream:
```
cout << key_formatter(formatter) << blah...
```

PR #69, opened 2019-06-14
Feature/streaming keys

dellaert:
Added streaming for Keys/Symbols and special `key_formatter` stream manipulator

### DIFF
--- a/gtsam/inference/Key.h
+++ b/gtsam/inference/Key.h
@@ -27,6 +27,8 @@
 
 #include <boost/function.hpp>
 
+#include <iosfwd>
+
 namespace gtsam {
 
 /// Typedef for a function to format a key, i.e. to convert it to a string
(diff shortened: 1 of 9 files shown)


gtsam/inference/Key.h
```diff
@@ -52,6 +54,34 @@ GTSAM_EXPORT std::string _multirobotKeyFormatter(gtsam::Key key);
...
+
+/**
+ * Output stream manipulator that will format gtsam::Keys according to the given
+ * KeyFormatter, as long as Key values are wrapped in a gtsam::StreamedKey.
+ * LabeledSymbol and Symbol values do not have to be wrapped.
+ * usage:
+ *   Key key = LabeledSymbol('x', 'A', 5); // cast to key type
+ *   cout << key_formatter(MultiRobotKeyFormatter) << StreamedKey(key);
+ */
+class key_formatter {
```
varunagrawal: Class naming is inconsistent and potentially confusing with `KeyFormatter`. Unless this is more of a `interface` class which is what it seems to be, then the name could be something like `KeyFormatterFactory`.

gtsam/inference/Key.h
```diff
@@ -52,6 +54,34 @@ GTSAM_EXPORT std::string _multirobotKeyFormatter(gtsam::Key key);
...
+  explicit key_formatter(KeyFormatter v) : formatter_(v) {}
+  friend std::ostream &operator<<(std::ostream &, const key_formatter &);
+  friend std::ostream &operator<<(std::ostream &, const StreamedKey &);
+
+ private:
+  KeyFormatter formatter_;
+  static void *&property(std::ios_base &s);
+  static void set_property(std::ios_base &s, const KeyFormatter &f);
+  static KeyFormatter *get_property(std::ios_base &s);
+};
```
varunagrawal: Any reason for not updating the Boost typedef to a full-blown class with all these properties. The `boost::function` function pointer can be a member of the class. Though this may be a significant engineering effort.

gtsam/inference/Key.h
dellaert: This is because it is a stream manipulator

gtsam/inference/Key.h
dellaert: The properties are not part of `KeyFormatter`, but static methods belonging to the stream manipulator class `key_formatter`, whose only purpose in life is to store that property in the stream.

dellaert:
Thanks for the comments. BTW, unless you feel very strongly about things (I don't think you did in this case - comments were more questions) or you're dealing with newbies (I'm not) just approve with comments in future PRs :-)

PR #78, opened 2019-06-22
Integrate Debian & Ubuntu PPA scripts

jlblancoc:
I don't know if you would like to integrate all of this in the main repository, but it might make sense. 
This PR includes: 
* Scripts to prepare the generation of .deb packages. 
* Scripts to prepare and upload to a PPA Ubuntu packages for gtsam. 
* A example CMake project to show the minimal set up for a user to build a project using gtsam, exploiting all the new cmake-exported targets (a single cmake command is required).

All this was done to help integrating gtsam-based apps with Continuous Integration, since building gtsam alone already takes too much time, so it makes sense to offer gtsam "precompiled" ;-)

For now I'm publishing nightly gtsam packages [here](https://launchpad.net/~joseluisblancoc/+archive/ubuntu/gtsam-develop), for Ubuntu Xenial, Bionic, Cosmic and Disco, for all main architectures (amd64, armhf, arm64, i386).

These scripts are largely based on previous works by [Bernd Pfrommer](https://launchpad.net/~bernd-pfrommer/+archive/ubuntu/gtsam) and myself (for the MRPT project). 

Note that the packages do not 100% comply yet with Debian standards, but they are 100% usable. 
For example, we should separate all files into, at least, two packages: `libgtsam` and `libgtsam-dev`. Right now everything is in the latter, for backwards compatibility if in the future things are done right.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/78)

PR #79, opened 2019-06-22
Allow Eigen versions of different patch number

jlblancoc:
Different patch numbers **shouldn't** be a problem for binary compatibility. 
This fixes problems caused by PPA packages included in #78 .

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/79)


dellaert:
Deferring to @chrisbeall on this and other PR

PR #84, opened 2019-07-11
Changes to get gtsam to compile in Windows

cntaylor:
These changes were required on my Windows machine to get the gtsam library to compile.  Using Boost 1.67, and VS 2017 (Community)

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/84)


jlblancoc:
Thanks! I think all cmake changes are fine. 

However, I changed your PR so it's merged into a temporary branch `msvc-fixes`, to edit your changes, basically to:
- remove whitespace changes, 
- remove the definition of NOMINMAX in the gtsam .h file, and add it via cmake definitions, which is a much cleaner way ;-)

PR #87, opened 2019-07-13
Resolved several MSVC compiling errors

cntaylor:
* The check.base unit tests all pass now.
* The gstam_matlab_wrapper class compiles with now errors now.
* Note that I had to remove all LieMatrix, LieVector, and LieScalar stuff
to get this to work...

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/87)


cntaylor:
Several of the "check." libraries seem to have multiple defined test functions.  However, I was able to --using this branch and commit -- compile gtsam_toolbox and the examples seem to run in MATLAB, so I think it is all compiling properly.  This happens despite the check.geometry and check.inference  projects, for example, _not_ compiling properly.  Happy that I have been able to compile a MATLAB toolbox from source though.  Now to add my own factors... :)

cntaylor:
It didn't seem like the optimal fix.  However, all of the errors are of the type that "you have declared a function, but there is no definition for it", very similar to what would happen if you have the .h files but not the .cpp files in a normal circumstance.  (Honestly, I was wondering if someone had removed the .cpp files when they deprecated, but not fully removed the requirement for that code from all sections of the repository.)  Are the .h files supposed to be the full implementation of the Lie* classes?

dellaert:
You can answer that question :-) Just check which methods VS complains about have no body in the .h file?

dellaert:
BTW: the fact it compiles *and* links on mac/linux makes me think: none...

cntaylor:
Unfortunately... I think you are right.  I can poke at this some more on
Monday.

cntaylor:
Well, I think I finally understand how this dllexport and dllimport stuff works with MSVC.  Made the fix rather simple, actually.  Take a look at this one and see what you think...

jlblancoc:
Sure, @dellaert !

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Usage Rules
+1.  Put `GTSAM_EXPORT` in front of any function that you want exported in the DLL _if and only if_ that function is defined in a .cpp file, not just a .h file.
+2.  Use `GTSAM_EXPORT` in a class definition (i.e. `class GSTAM_EXPORT MyClass {...}`) only if:
+    * At least one of the functions inside that class is defined in a .cpp file and not just the .h file.
+    * You can `GTSAM_EXPORT` any class it inherits from as well.  (Note that this implictly requires the class does not derive from a "header-only" class.  Note that Eigen is a "header-only" library, so if your class derives from Eigen, _do not_ use `GTSAM_EXPORT` in the class definition!) 
+3.  If you have defined a class using `GTSAM_EXPORT`, do not use `GTSAM_EXPORT` in any of its individual function declarations.  (Note that you _can_ put `GTSAM_EXPORT` in the definition of individual functions within a class as long as you don't put `GTSAM_EXPORT` in the class definition.)
```
jlblancoc: 100% agree with this one!

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
+# Using GTSAM_EXPORT
+
+To create a DLL in windows, the `GTSAM_EXPORT` keyword has been created and needs to be applied to different methods and classes in the code to expose this code outside of the DLL.  However, there are several intricacies that make this more difficult than it sounds.  In general, if you follow the following three rules, GTSAM_EXPORT should work properly.  The rest of the document also describes (1) the common error message encountered when you are not following these rules and (2) the reasoning behind these usage rules.
+
+## Usage Rules
+1.  Put `GTSAM_EXPORT` in front of any function that you want exported in the DLL _if and only if_ that function is defined in a .cpp file, not just a .h file.
```
jlblancoc: I don't totally get this one... perhaps you refer here to inlined functions defined in headers?
Also, this rule may get confusing when working with templates, explicitly instantiated templates, etc. which may be declared in the .h, and defined (and "exported") in the .cpp, etc. 

What's a clear use example you have in your mind when writing this rule?

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
+# Using GTSAM_EXPORT
+
```
jlblancoc: wow, this doc is like a large Blog entry on the topic... good idea, @dellaert , @cntaylor ! :+1:

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
...
+To create a DLL in windows, the `GTSAM_EXPORT` keyword has been created and needs to be applied to different methods and classes in the code to expose this code outside of the DLL.  However, there are several intricacies that make this more difficult than it sounds.  In general, if you follow the following three rules, GTSAM_EXPORT should work properly.  The rest of the document also describes (1) the common error message encountered when you are not following these rules and (2) the reasoning behind these usage rules.
```
jlblancoc: > the `GTSAM_EXPORT` keyword has been created 

Should it be clearer mentioning that it is CMake the one who automatically defines this symbol (as far as I know) when building dynamic libs?

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+Sorry for the long error message there, but I want to quickly decode it to help understand what is going on.  First, there is an unresolved symbol `gtsam::SO3::print`.  This can occur because _either_ `GTSAM_EXPORT` was not added to the print function definition when it should have been, _OR_ because `GTSAM_EXPORT` was added to the print function definition when it is fully declared in the header.  This error was detected while compiling `check_geometry_program` and pulling in `...\testSO3.obj`.  Specifically, within the function call `gtsam::Testabl<class gtsam::SO3>::Print (...)`.  Note that this error did _not_ occur when compiling the library that actually has SO3 inside of it.
```
jlblancoc: I would remove the "sorry for ..." opening sentence ;-) Just go straight to the point...

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
...
+
+## But Why?
+To me, the behavior of GTSAM_EXPORT was very baffling.  However, I believe it can be explained by understanding the rules that MSVC operates under.
```
jlblancoc: dllimport/dllexport has been like that since... literally decades.
I would remove this first sentence, which you may seem superfluous if you read it again 10 years from now ;-)

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
...
+## But Why?
+To me, the behavior of GTSAM_EXPORT was very baffling.  However, I believe it can be explained by understanding the rules that MSVC operates under.  
+
+First, we need to understand exactly what `GTSAM_EXPORT` does.  When you use `GTSAM_EXPORT` on a function (or class) definition, it actually does two things.  First, when you are building the library, it turns into a "dllexport" command, telling the compiler that this function should go into the DLL and be visible externally.  When you pull in that same header for a different library, it switches to a "dllimport" command, telling the compiler it should find this function in a DLL somewhere.  This leads to the first rule the compiler uses.  (Note that I say "compiler rules" when the rules may actually be in the linker, but I am conflating the two terms here when I speak of the "compiler rules".) 
+
+***Compiler Rule #1*** If a `dllimport` command is used in defining a function or class, that function or class _must_ be found in a DLL.
+
+Rule #1 doesn't seem very bad, until you combine it with rule #2
+
+***Compiler Rule #2*** Anything defined in a header file is not included in a DLL.
```
jlblancoc: > Anything defined in a header file is not included in a DLL.

I'm not 100% sure of this... it is NOT exported if you mark it explicitly with `inline`. Otherwise, I'm not sure if there is a well-defined, standardized expected behavior (?). 

I'll kindly request the knowledge of @jolting here just in case he can help us to be accurate in the description...
@jolting

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
...
+
+When these two rules are combined, you get some very confusing results.  For example, a class which is completely defined in a header (e.g. LieMatrix) cannot use `GTSAM_EXPORT` in its definition.  If LieMatrix is defined with `GTSAM_EXPORT`, then the compiler _must_ find LieMatrix in a DLL.  Because LieMatrix is a header-only class, however, it can't find it, leading to a very confusing "I can't find this symbol" type of error.  Note that the linker says it can't find the symbol even though the compiler found the header file that completely defines the class.
+
+This leads to another compiler rule that finishes explaining the usage rules proposed above.
+
+***Compiler Rule #3*** If a class is defined with `GTSAM_EXPORT`, then any classes it inherits from must be definable with a dllexport command as well.
```
jlblancoc: Again, I'm not 100% sure of this... (but may be wrong!). 
For example, [this MSVC warning](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4275?view=vs-2019), which I found many times, talks specifically about that situation, and it's not forbidden, it "only" **could** lead to data corruption. However, I have been using DLL-exported classes that inheriting from header-only classes for years, without problems. Nowadays, I avoid it just to be safe, but I wouldn't say it's carved in stone that it cannot be done...

Using-GTSAM-EXPORT.md
```diff
@@ -0,0 +1,41 @@
...
+
+One of the most immediate results of this compiler rule is that any class that inherits from an Eigen class _cannot_ be defined with `GTSAM_EXPORT`.  Because Eigen is a header-only class, compiler rule #2 prohibits any Eigen class from being defined with a "dllexport" command.  Therefore, no Eigen inherited classes can use `GTSAM_EXPORT` either.  (Note that individual functions with an inherited class can be "exported."  Just not the entire class.)
```
jlblancoc: It's a dirty solution, but if that's the unique way you found to solve the linking problems, let is be!  

Just let's iterate once over this text, and in the meanwhile give me some time to check it on a local machine before merging. 

Awesome work!

Using-GTSAM-EXPORT.md
cntaylor: A couple of examples:

1.  inline_QR did not have the GTSAM_EXPORT keyword in front of it.  It was declared in a .cpp file, but not in the .h file, so there was a linker error when another library called inline_QR.
2.  The Lie* classes did not have any functions declared in a .cpp file.  If you added GTSAM_EXPORT to the class definition, you got errors when calling any of those functions because they were _not_ in the DLL.

My understanding is basically if it is completely declared in a header, there is no reason to put it in the DLL.

I am wondering if the confusion may be that I put the wrong word in.  It should really be "that function is _declared_" not _defined_.

Does that make sense?  I have changed the defined to declared in the .md file.

Using-GTSAM-EXPORT.md
cntaylor: Changing "defined" to "declared" in the first sub-bullet of usage rule 2 as well.

Using-GTSAM-EXPORT.md
cntaylor: Done...

Using-GTSAM-EXPORT.md
cntaylor: Thank you!  Rewritten.

Using-GTSAM-EXPORT.md
cntaylor: Please note that my expertise on DLLimport/export is what I have learned playing with GTSAM for the last week or so getting it to compile and looking up things online.  So, I am theorizing to the best of my knowledge, but I certainly am not an expert.

That said, I have seen several website mention that if you declare a function in a header inside a class definition, the compiler automatically makes it an "inline" function.  If it was not inline and you included this class definition in more than one file, then you would get a "multiple definitions of the function 'doit'" error.  So this one is essentially marked inline and will not be put in the DLL.

Once again, I should probably change "define" to declare.  My im-precision may be causing confusion. :(

I guess it may be possible that a non-inline function is declared in a header file and included in the DLL, but that seems like a corner case (due to the "multiple instances" compilation problem inherent with non-inline functions) so I don't think we need to worry about that too much.

Using-GTSAM-EXPORT.md
cntaylor: I'm not 100% sure. I know that for the "Point?" classes in gtsam/geometry, because they derive from Eigen classes, DLLimport problems were popping up until I moved the GTSAM_EXPORT inside the class to the individual functions.  It may be that it violates on the conditions discussed at the [MSVC warning](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4275?view=vs-2019) page you kindly pointed towards.

In the meantime, I will weaken the verbiage to say that things could go bad and it is generally not recommended.

Using-GTSAM-EXPORT.md
cntaylor: Thanks.  Please note that there are still compiler errors in several of the "check" classes that I am slowly working my way through.  The matlab wrappers seem to work, which is my main concern, but while I am at it, I would like to get the "check" classes all working as well.

Also, @dellaert I compiled the libraries with the TYPEDEF_POINT_AS_VECTOR option enabled and did not see any new errors pop up.

jlblancoc:
I verified it builds, thanks @cntaylor ! I'll merge, please open a new PR if you want to add new fixes :+1:

jlblancoc:
@cntaylor I noticed that there is an inconsistent usage of export macros between GTSAM and GTSAM_UNSTABLE. I'll fix it in the branch msvc_fixes. Please, update your local copy of that branch if you also keep working on that...

PR #107, opened 2019-08-29
Fixed spelling for Welsch and added to gtsam.h

fishyai:
Welsch is spelled inconsistently with the citation: https://members.loria.fr/MOBerger/Enseignement/Master2/Documents/ZhangIVC-97-01.pdf

I am also hoping to use this from Python, so I added it to gtsam.h

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/107)

### DIFF
--- a/gtsam.h
+++ b/gtsam.h
@@ -252,9 +252,9 @@ class FactorIndices {
 bool isDebugVersion();
 
 #include <gtsam/base/DSFMap.h>
-class IndexPair { 
-  IndexPair(); 
-  IndexPair(size_t i, size_t j); 
+class IndexPair {
+  IndexPair();
+  IndexPair(size_t i, size_t j);
   size_t i() const;
   size_t j() const;
 };
@@ -1387,6 +1387,15 @@ virtual class Tukey: gtsam::noiseModel::mEstimator::Base {
   // enabling serialization functionality
   void serializable() const;
 };
+
+virtual class Welsch: gtsam::noiseModel::mEstimator::Base {
+  Welsch(double k);
+  static gtsam::noiseModel::mEstimator::Welsch* Create(double k);
+
+  // enabling serialization functionality
+  void serializable() const;
+};
+
 
 }///\namespace mEstimator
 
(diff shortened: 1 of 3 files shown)


fishyai:
Oops... I must have forgotten to add the change to gtsam.h. I'll fix that now. 

Yeah--that all makes sense. Is there already a deprecated section you all have been using for this purpose or are you suggesting a new one?

fishyai:
Or rather, is there a particular deprecation pattern you'd like this to follow? If you have an example, I can copy it. 

I think it would be good if the Welsh::Create() {} message prints a message saying that this method is deprecated and will be removed in a future version.

mikesheffler:
> If you have an example, I can copy it.

Take a look at 20736b6f14ad0245d36efdc7784b72955dd1a468. Usually there is just a little block near where the code 'used to' live.

fishyai:
I also was unable to find the citation for the Welsch model in the citations file. I posed in the Google Group and someone sent me the citation. I'd like to put it in the right bibtex file. Where would be best?

mikesheffler:
>  I'd like to put it in the right bibtex file. Where would be best?

I'm not really sure. There isn't a bibtex file for the code base _per se_ (at least, as far as I know) -- just for specific concepts written up as independent documents.

One possible model would be how it was done for the (Combined) IMU Factor:

[gtsam/navigation/CombinedImuFactor.cpp](https://github.com/borglab/gtsam/blob/develop/gtsam/navigation/CombinedImuFactor.cpp)
[gtsam/navigation/CombinedImuFactor.h](https://github.com/borglab/gtsam/blob/develop/gtsam/navigation/CombinedImuFactor.h)

Near the top of the `.h`, there is a list of four references, and, at times, there are statements [such as](https://github.com/borglab/gtsam/blob/develop/gtsam/navigation/CombinedImuFactor.cpp#L76):

`In this implementation, in contrast to [2], we consider the uncertainty of the bias selection and we keep correlation between biases and preintegrated measurements`

That seems reasonable. Basically, keep a plain text comment of the reference nearby, without turning it into a book report 😁 @dellaert ?

dellaert:
Not loving all the copy/paste, but also don't see an easy way around it. Will merge.

varunagrawal:
@fishyai why is your commit showing up with the name Alice Anderson? This will make future back-referencing hard.

Moreover, @dellaert this PR has a bug in it. The constructors for the mEstimators in `gtsam.h` are all wrong due to all the copy-paste.

dellaert:
yeah, that's weird. Adam, you might want to fix that :-)

PR #110, opened 2019-09-07
Several fixes in legacy code

ProfFan:
This fixes segfaults of `testSubgraphSolver` on my machine (Arch Linux with latest gcc 9.1.0).

Also fixes compilation of `gtsam_unstable`.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/110)

### DIFF
--- a/gtsam/linear/SubgraphBuilder.cpp
+++ b/gtsam/linear/SubgraphBuilder.cpp
@@ -127,7 +127,8 @@ static vector<size_t> UniqueSampler(const vector<double> &weight,
 
     /* sampling and cache results */
     vector<size_t> samples = iidSampler(localWeights, n - count);
-    for (const size_t &index : samples) {
+    const auto samplesSize = samples.size();
+    for (size_t index = 0; index < samplesSize; index++) {
       if (touched[index] == false) {
         touched[index] = true;
         samples.push_back(index);
--- a/gtsam/linear/SubgraphBuilder.cpp
+++ b/gtsam/linear/SubgraphBuilder.cpp
@@ -107,7 +107,7 @@ static vector<size_t> UniqueSampler(const vector<double> &weight,
   const size_t m = weight.size();
   if (n > m) throw std::invalid_argument("UniqueSampler: invalid input size");
 
-  vector<size_t> samples;
+  vector<size_t> results;
 
   size_t count = 0;
   vector<bool> touched(m, false);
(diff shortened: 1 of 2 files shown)


dellaert:
Not sure why you call it "legacy code", SubgraphBuilder is pretty recent refactor :-)

gtsam/linear/SubgraphBuilder.cpp
```diff
@@ -127,7 +127,8 @@ static vector<size_t> UniqueSampler(const vector<double> &weight,
 
     /* sampling and cache results */
     vector<size_t> samples = iidSampler(localWeights, n - count);
-    for (const size_t &index : samples) {
+    const auto samplesSize = samples.size();
+    for (size_t index = 0; index < samplesSize; index++) {
```
dellaert: This is suspicious. To me (a) line 130 should not be an issue for any c++11 compiler, and (b) you are actually changing the behavior, from the *content* of samples to the indices 0...sampleSize-1.

ProfFan: The `push_back` inside the for loop sometimes will trigger a reallocation of the vector, and subsequently invalidating all existing iterators.

https://stackoverflow.com/questions/35938329/c-push-back-in-stdvector-while-iterating-it

ProfFan: So the `SEGSEGV` will be on line 131 because `index` is invalid.

dellaert: Yikes, the problem is much bigger! This function is completely wrong, as we're shadowing samples on line 110 with new definition on line 129. I will look at PR where this happened to find correct behavior.

dellaert: For reference, PR is on Bitbucket: https://bitbucket.org/gtborg/gtsam/pull-requests/415/split-off-subgraphbuilder/diff

dellaert: Could you restore to `for ( const size_t &index: samples ) {` but change first `samples` to `result` ? push back should happen to `result` and function should return `result`...

ProfFan: What is the purpose of `UniqueSampler`? Also the `samples` defined on line 110 is never used.

dellaert: Clearly we need a unit test here to fail, but unit tests were not abundant when this code was created, though I added some in refactor.

dellaert: Note, I have a branch out to make IID sampling better and replace this code, but CI tests failed for some config and I did not get to fixing it yet.

dellaert: > What is the purpose of `UniqueSampler`? Also the `samples` defined on line 110 is never used.

Yeah, that's what I'm saying. In the refactor I renamed result to samples without seeing the shadowing. It should be restored to it's original name.

ProfFan: Done :)

dellaert:
Lgtm

PR #115, opened 2019-09-18
C++11 Absolute Value

varunagrawal:
Replaced `fabs` with `std::abs` to be more C++11 compliant.

I have used `std::abs` everywhere instead of simply `abs` since I didn't want accidental linking against the C version of `abs` which only works for integers. Explicit is better than implicit.

https://stackoverflow.com/questions/33738509/whats-the-difference-between-abs-and-fabs

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/115)

### DIFF
--- a/gtsam/base/Matrix.h
+++ b/gtsam/base/Matrix.h
@@ -90,7 +90,7 @@ bool equal_with_abs_tol(const Eigen::DenseBase<MATRIX>& A, const Eigen::DenseBas
     for(size_t j=0; j<n1; j++) {
       if(boost::math::isnan(A(i,j)) ^ boost::math::isnan(B(i,j)))
         return false;
-      else if(fabs(A(i,j) - B(i,j)) > tol)
+      else if(std::abs(A(i,j) - B(i,j)) > tol)
         return false;
     }
   return true;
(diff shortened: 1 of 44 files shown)


varunagrawal:
@dellaert you may want to reconcile the `bugfix/Huber` branch which was last looked at in 2016.

dellaert:
On your recommendation :-) will you also create an issue to track this?

dellaert:
I think this should be split into two PRs. One is easy (std::abs). Other one should correspond to issue about Huber.

varunagrawal:
@dellaert this PR now only deals with the absolute value computation.

gtsam/linear/tests/testNoiseModel.cpp
```diff
@@ -452,19 +452,24 @@ TEST(NoiseModel, WhitenInPlace)
 /*
  * These tests are responsible for testing the weight functions for the m-estimators in GTSAM.
  * The weight function is related to the analytic derivative of the residual function. See
- *  http://research.microsoft.com/en-us/um/people/zhang/INRIA/Publis/Tutorial-Estim/node24.html
+ *  https://members.loria.fr/MOBerger/Enseignement/Master2/Documents/ZhangIVC-97-01.pdf
```
dellaert: Changes in this file (except std::abs) belong in other PR

varunagrawal: I just force-pushed so the PRs should now be truly separate.

PR #116, opened 2019-09-18
Fix Huber mEstimator

varunagrawal:
1. Fixed `weight` for Huber mEstimator in `NoiseModel.h`
2. Added corresponding unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/116)

### DIFF
--- a/gtsam/linear/NoiseModel.h
+++ b/gtsam/linear/NoiseModel.h
@@ -776,7 +776,8 @@ namespace gtsam {
 
         Huber(double k = 1.345, const ReweightScheme reweight = Block);
         double weight(double error) const {
-          return (std::abs(error) < k_) ? (1.0) : (k_ / fabs(error));
+          double absError = std::abs(error);
+          return (absError < k_) ? (1.0) : (k_ / absError);
         }
         void print(const std::string &s) const;
         bool equals(const Base& expected, double tol=1e-8) const;
--- a/gtsam/linear/NoiseModel.h
+++ b/gtsam/linear/NoiseModel.h
@@ -629,7 +629,7 @@ namespace gtsam {
     /**
      * The mEstimator name space contains all robust error functions.
      * It mirrors the exposition at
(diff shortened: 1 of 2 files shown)


gtsam/linear/tests/testNoiseModel.cpp
```diff
@@ -452,19 +452,24 @@ TEST(NoiseModel, WhitenInPlace)
...
   const mEstimator::Huber::shared_ptr huber = mEstimator::Huber::Create(k);
   const double weight1 = huber->weight(error1),
-               weight2 = huber->weight(error2);
+               weight2 = huber->weight(error2),
+               weight3 = huber->weight(error3),
+               weight4 = huber->weight(error4);
   DOUBLES_EQUAL(1.0, weight1, 1e-8);
   DOUBLES_EQUAL(0.5, weight2, 1e-8);
+  // Test negative value to ensure we take absolute value of error.
+  DOUBLES_EQUAL(0.5, weight3, 1e-8);
```
dellaert: just inline calls, like so: `DOUBLES_EQUAL(0.5, huber->weight(error3), 1e-8) `

gtsam/linear/NoiseModel.h
```diff
@@ -776,7 +776,7 @@ namespace gtsam {
 
         Huber(double k = 1.345, const ReweightScheme reweight = Block);
         double weight(double error) const {
-          return (error < k_) ? (1.0) : (k_ / fabs(error));
+          return (std::abs(error) < k_) ? (1.0) : (k_ / fabs(error));
```
dellaert: I propose
```
const double absError = std::abs(error);
return (absError < k_) ? (1.0) : (k_ / absError);
```

varunagrawal: Updated both.

ProfFan: This reminded me again the beauty of functional programming :)

PR #117, opened 2019-09-18
Wrap mEstimators

varunagrawal:
1. Wrap more mEstimators.
2. Also wrap the `weight` function of each of the mEstimators.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/117)

### DIFF
--- a/gtsam.h
+++ b/gtsam.h
@@ -1364,6 +1364,7 @@ virtual class Null: gtsam::noiseModel::mEstimator::Base {
   void serializable() const;
 
   double weight(double error) const;
+  double residual(double error) const;
 };
 
 virtual class Fair: gtsam::noiseModel::mEstimator::Base {
@@ -1384,6 +1385,7 @@ virtual class Huber: gtsam::noiseModel::mEstimator::Base {
   void serializable() const;
 
   double weight(double error) const;
+  double residual(double error) const;
 };
 
 virtual class Cauchy: gtsam::noiseModel::mEstimator::Base {
@@ -1394,6 +1396,7 @@ virtual class Cauchy: gtsam::noiseModel::mEstimator::Base {
   void serializable() const;
 
   double weight(double error) const;
+  double residual(double error) const;
 };
 
 virtual class Tukey: gtsam::noiseModel::mEstimator::Base {
@@ -1404,6 +1407,7 @@ virtual class Tukey: gtsam::noiseModel::mEstimator::Base {
   void serializable() const;
 
   double weight(double error) const;
+  double residual(double error) const;
 };
 
 virtual class Welsch: gtsam::noiseModel::mEstimator::Base {
@@ -1414,6 +1418,7 @@ virtual class Welsch: gtsam::noiseModel::mEstimator::Base {
   void serializable() const;
 
   double weight(double error) const;
+  double residual(double error) const;
 };
 
 virtual class GemanMcClure: gtsam::noiseModel::mEstimator::Base {
@@ -1424,6 +1429,7 @@ virtual class GemanMcClure: gtsam::noiseModel::mEstimator::Base {
   void serializable() const;
 
   double weight(double error) const;
+  double residual(double error) const;
 };
 
 //TODO DCS and L2WithDeadZone mEstimators
--- a/gtsam.h
+++ b/gtsam.h
@@ -1362,6 +1362,8 @@ virtual class Null: gtsam::noiseModel::mEstimator::Base {
 
   // enabling serialization functionality
   void serializable() const;
+
+  double weight(double error) const;
 };
 
 virtual class Fair: gtsam::noiseModel::mEstimator::Base {
@@ -1370,6 +1372,8 @@ virtual class Fair: gtsam::noiseModel::mEstimator::Base {
 
   // enabling serialization functionality
   void serializable() const;
+
+  double weight(double error) const;
 };
 
 virtual class Huber: gtsam::noiseModel::mEstimator::Base {
@@ -1378,6 +1382,18 @@ virtual class Huber: gtsam::noiseModel::mEstimator::Base {
 
   // enabling serialization functionality
   void serializable() const;
+
+  double weight(double error) const;
+};
+
+virtual class Cauchy: gtsam::noiseModel::mEstimator::Base {
+  Cauchy(double k);
+  static gtsam::noiseModel::mEstimator::Cauchy* Create(double k);
+
+  // enabling serialization functionality
+  void serializable() const;
+
+  double weight(double error) const;
 };
 
 virtual class Tukey: gtsam::noiseModel::mEstimator::Base {
@@ -1386,6 +1402,8 @@ virtual class Tukey: gtsam::noiseModel::mEstimator::Base {
 
   // enabling serialization functionality
   void serializable() const;
+
+  double weight(double error) const;
 };
 
 virtual class Welsch: gtsam::noiseModel::mEstimator::Base {
@@ -1394,8 +1412,21 @@ virtual class Welsch: gtsam::noiseModel::mEstimator::Base {
 
   // enabling serialization functionality
   void serializable() const;
+
+  double weight(double error) const;
 };
 
+virtual class GemanMcClure: gtsam::noiseModel::mEstimator::Base {
+  GemanMcClure(double k);
+  static gtsam::noiseModel::mEstimator::GemanMcClure* Create(double k);
+
+  // enabling serialization functionality
+  void serializable() const;
+
+  double weight(double error) const;
+};
+
+//TODO DCS and L2WithDeadZone mEstimators
 
 }///\namespace mEstimator
 
--- a/gtsam.h
+++ b/gtsam.h
@@ -1375,6 +1375,7 @@ virtual class Fair: gtsam::noiseModel::mEstimator::Base {
   void serializable() const;
 
   double weight(double error) const;
+  double residual(double error) const;
 };
 
 virtual class Huber: gtsam::noiseModel::mEstimator::Base {
--- a/gtsam.h
+++ b/gtsam.h
@@ -1422,8 +1422,8 @@ virtual class Welsch: gtsam::noiseModel::mEstimator::Base {
 };
 
 virtual class GemanMcClure: gtsam::noiseModel::mEstimator::Base {
-  GemanMcClure(double k);
-  static gtsam::noiseModel::mEstimator::GemanMcClure* Create(double k);
+  GemanMcClure(double c);
+  static gtsam::noiseModel::mEstimator::GemanMcClure* Create(double c);
 
   // enabling serialization functionality
   void serializable() const;
--- a/gtsam.h
+++ b/gtsam.h
@@ -1432,7 +1432,27 @@ virtual class GemanMcClure: gtsam::noiseModel::mEstimator::Base {
   double residual(double error) const;
 };
 
-//TODO DCS and L2WithDeadZone mEstimators
+virtual class DCS: gtsam::noiseModel::mEstimator::Base {
+  DCS(double c);
+  static gtsam::noiseModel::mEstimator::DCS* Create(double c);
+
+  // enabling serialization functionality
+  void serializable() const;
+
+  double weight(double error) const;
+  double residual(double error) const;
+};
+
+virtual class L2WithDeadZone: gtsam::noiseModel::mEstimator::Base {
+  L2WithDeadZone(double k);
+  static gtsam::noiseModel::mEstimator::L2WithDeadZone* Create(double k);
+
+  // enabling serialization functionality
+  void serializable() const;
+
+  double weight(double error) const;
+  double residual(double error) const;
+};
 
 }///\namespace mEstimator
 
--- a/gtsam/linear/NoiseModel.cpp
+++ b/gtsam/linear/NoiseModel.cpp
@@ -858,6 +858,12 @@ double GemanMcClure::weight(double error) const {
   return c4/(c2error*c2error);
 }
 
+double GemanMcClure::residual(double error) const {
+  const double c2 = c_*c_;
+  const double error2 = error*error;
+  return 0.5 * (c2 * error2) / (c2 + error2);
+}
+
 void GemanMcClure::print(const std::string &s="") const {
   std::cout << s << ": Geman-McClure (" << c_ << ")" << std::endl;
 }
--- a/gtsam/linear/NoiseModel.cpp
+++ b/gtsam/linear/NoiseModel.cpp
@@ -718,15 +718,25 @@ void Null::print(const std::string &s="") const
 Null::shared_ptr Null::Create()
 { return shared_ptr(new Null()); }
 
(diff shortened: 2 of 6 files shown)


dellaert:
Can we add the gtsam.h changes and the matlab script in here, as well?

varunagrawal:
This PR builds upon #116. Once 116 is approved, this one will only show changes to `gtsam.h`.

varunagrawal:
This PR no longer depends on #116.

gtsam.h
```diff
@@ -1378,6 +1382,18 @@ virtual class Huber: gtsam::noiseModel::mEstimator::Base {
 
   // enabling serialization functionality
   void serializable() const;
+
+  double weight(double error) const;
+};
+
+virtual class Cauchy: gtsam::noiseModel::mEstimator::Base {
```
dellaert: You know, inheritance works in gtsam.h. Can you not just add weight to the Base?

matlab/gtsam_examples/VisualizeMEstimators.m
```diff
@@ -0,0 +1,172 @@
+import gtsam.*
```
dellaert: Please add documentation and comments? Take a look at LocalizationExample.m to see the documentation standards. In addition, could we scale all figures the same way so the models are directly comparable?

matlab/gtsam_examples/VisualizeMEstimators.m
```diff
@@ -0,0 +1,172 @@
...
+end
+
+function [rho, psi, w] = fair(x)
+    import gtsam.*
+    c = 1.3998;
+
+    rho = c^2 * ( (abs(x) ./ c) - log(1 + (abs(x)./c)) );
+    est = noiseModel.mEstimator.Fair(c);
+    
+    w = zeros(size(x));
```
dellaert: Copy/paste

matlab/gtsam_examples/VisualizeMEstimators.m
```diff
@@ -0,0 +1,172 @@
...
+    plot(x, y);
+    xlim([-15, 15]);
+    ylim([-yl, yl]);
+end
+
+function [rho, psi, w] = fair(x)
+    import gtsam.*
+    c = 1.3998;
+
+    rho = c^2 * ( (abs(x) ./ c) - log(1 + (abs(x)./c)) );
```
dellaert: So, this is crazy. We should, as Mike Bosse suggested, put the residuals into the noiseModel header. Let's talk if we can then kill two birds with one stone and immediately understand and fix the problem he is talking about.

matlab/gtsam_examples/VisualizeMEstimators.m
```diff
@@ -0,0 +1,172 @@
...
+
+[rho, psi, w] = tukey(x);
+figure(6);
+plot_rho(x, rho, 1, -5, 10);
+title('Tukey');
+plot_psi(x, psi, 2, -2, 2);
+plot_w(x, w, 3, 2);
+saveas(figure(6), 'tukey.png');
+
+function plot_rho(x, y, idx, yll, ylu)
```
dellaert: Can we just combine those three functions? Docs...

ProfFan:
#121

varunagrawal:
tukey
welsch
gemanmcclure
cauchy
huber
fair

varunagrawal:
I need to regenerate Huber with the fixed code. :(

dellaert:
I guess rebase this on develop should fix it?

dellaert:
Do we now have the residuals into the noiseModel header?

varunagrawal:
We don't. Should I prioritize that over other tasks?

dellaert:
The blog post is priority. I thought this PR was part of it. I guess you can do without, but we can't land this PR without addressing this. You could fix this after the blog post.

varunagrawal:
I have added the residuals. However, I can't find the code that calls the weight function to compute the total penalty. That would be the last thing to update (along with the Matlab example) before landing this PR.

gtsam.h
dellaert: Limitation of wrap, understood.

gtsam.h
```diff
@@ -1370,6 +1373,8 @@ virtual class Fair: gtsam::noiseModel::mEstimator::Base {
 
   // enabling serialization functionality
   void serializable() const;
+
+  double weight(double error) const;
```
dellaert: No residual?

gtsam.h
```diff
@@ -1394,8 +1416,23 @@ virtual class Welsch: gtsam::noiseModel::mEstimator::Base {
...
+  static gtsam::noiseModel::mEstimator::GemanMcClure* Create(double k);
+
+  // enabling serialization functionality
+  void serializable() const;
+
+  double weight(double error) const;
+  double residual(double error) const;
 };
 
+//TODO DCS and L2WithDeadZone mEstimators
```
dellaert: Do it?

gtsam/linear/NoiseModel.h
```diff
@@ -18,6 +18,7 @@
 
 #pragma once
 
+#include<iostream>
```
dellaert: Nonono! Iosfwd is *reluctantly* tolerated, iostream is not. We want as few headers included in .h files as possible, to cut down on compile time. Especially a header like noiseModel.

Also, header order in GTSAM is near to far, so iosfwd (IF needed) would come last.

gtsam/linear/NoiseModel.h
```diff
@@ -752,6 +754,12 @@ namespace gtsam {
         double weight(double error) const {
           return 1.0 / (1.0 + std::abs(error) / c_);
         }
+        double residual(double error) const {
+          double absError = std::abs(error);
```
dellaert: Probably compiler will figure it out, but add const to all these?

gtsam/linear/NoiseModel.h
```diff
@@ -779,6 +787,14 @@ namespace gtsam {
           double absError = std::abs(error);
           return (absError < k_) ? (1.0) : (k_ / absError);
         }
+        double residual(double error) const {
+          double absError = std::abs(error);
+          if (absError <= k_) {  // |x| <= k
```
dellaert: Not: above it is < not <=

gtsam/linear/NoiseModel.h
```diff
@@ -839,6 +860,16 @@ namespace gtsam {
           }
           return 0.0;
         }
+        double residual(double error) const {
+          double absError = std::abs(error);
+          if (absError <= c_) {
+            double t = csquared_ - (error*error);
+            double val = t * t * t;
+            return ((csquared_*csquared_*csquared_) - val) / (6.0 * csquared_ * csquared_);
```
dellaert: So, c^6 ?? Seems weird. Is this validated by a test? Also, should be computed in constructor then.

gtsam/linear/NoiseModel.h
```diff
@@ -897,6 +932,10 @@ namespace gtsam {
           double xc2 = (error*error)/csquared_;
           return std::exp(-xc2);
         }
+        double residual(double error) const {
```
dellaert: Is pretty horrible copy/paste. can we not just say `using Welsh = Welsch’ ??

gtsam.h
varunagrawal: Yup working on that.

gtsam/linear/NoiseModel.h
varunagrawal: Woops. I was using that for debugging and accidentally left it in. :disappointed:

gtsam/linear/NoiseModel.h
varunagrawal: Actually, as per the definition of Huber loss, it should be `<=` for both.
https://en.wikipedia.org/wiki/Huber_loss#Definition

gtsam/linear/NoiseModel.h
varunagrawal: Yup added to these as well as function definitions for other M-estimators.

gtsam/linear/NoiseModel.h
varunagrawal: I modified this to make this easier to read and correlate with the actual expression.

gtsam/linear/NoiseModel.h
```diff
@@ -751,15 +750,8 @@ namespace gtsam {
...
-        double weight(double error) const {
-          return 1.0 / (1.0 + std::abs(error) / c_);
-        }
-        double residual(double error) const {
-          double absError = std::abs(error);
-          double normalizedError = absError / c_;
-          double val = normalizedError - std::log(1 + normalizedError);
-          return c_ * c_ * val;
-        }
+        double weight(double error) const;
```
dellaert: Const then? But, I normally consider const redundant on arguments passed by value...

gtsam/linear/NoiseModel.h
varunagrawal: Yeah I would actually prefer that. This is jarring.

gtsam/linear/NoiseModel.h
varunagrawal: Yeah I was wondering what your policy on that was? Some people say const in the declaration is redundant since it doesn't do anything. Should I remove the const for the function arguments then? I felt it was a nice built-in check since the values should not change.

dellaert: I would say undo these changes. We don’t do it anywhere else in GTSAM, and it is orthogonal to the goal of the PR

varunagrawal: Understood.

gtsam/linear/NoiseModel.h
```diff
@@ -924,7 +890,8 @@ namespace gtsam {
 
         GemanMcClure(double c = 1.0, const ReweightScheme reweight = Block);
         virtual ~GemanMcClure() {}
-        virtual double weight(double error) const;
+        virtual double weight(const double error) const;
```
varunagrawal: @dellaert what do you have to say about these `virtual` identifiers? Would we need to subclass the estimators?

dellaert: Good point: the base class should have virtual, but derived classes should have override in c++11

varunagrawal: Doesn't this mean all the estimators should use `override`?

dellaert: Yes

varunagrawal: All done.

gtsam/linear/NoiseModel.cpp
```diff
@@ -718,15 +718,25 @@ void Null::print(const std::string &s="") const
...
     throw runtime_error("mEstimator Fair takes only positive double in constructor.");
   }
 }
 
-/* ************************************************************************* */
-// Fair
-/* ************************************************************************* */
+double Fair::weight(double error) const {
+  return 1.0 / (1.0 + std::abs(error) / c_);
+}
```
ProfFan: Newline?

gtsam/linear/NoiseModel.cpp
```diff
@@ -791,7 +825,30 @@ Cauchy::shared_ptr Cauchy::Create(double c, const ReweightScheme reweight) {
...
+  }
+}
+
+double Tukey::weight(double error) const {
+  if (std::abs(error) <= c_) {
+    const double xc2 = error*error/csquared_;
+    return (1.0-xc2)*(1.0-xc2);
+  }
+  return 0.0;
+}
```
ProfFan: Same

matlab/gtsam_examples/VisualizeMEstimators.m
ProfFan: Is this resolved?

gtsam/linear/NoiseModel.h
```diff
@@ -727,6 +727,7 @@ namespace gtsam {
         Null(const ReweightScheme reweight = Block) : Base(reweight) {}
         virtual ~Null() {}
```
ProfFan: Why some of the destructors are virtual and some not?

varunagrawal: Oops I missed that. Fixed.

varunagrawal:
@dellaert @ProfFan the CI should now pass and I have fixed a few more things (such as adding `residual` wrapper to Fair).

I have also gone ahead and addressed Mike Bosse's TODOs, and hence I am asking for a review once again to ensure I am doing the right thing.

gtsam/linear/NoiseModel.h
```diff
@@ -1069,8 +1018,7 @@ namespace gtsam {
       inline virtual Vector unwhiten(const Vector& /*v*/) const
       { throw std::invalid_argument("unwhiten is not currently supported for robust noise models."); }
       inline virtual double distance(const Vector& v) const
-      { return this->whiten(v).squaredNorm(); }
-      // TODO(mike): fold the use of the m-estimator residual(...) function into distance(...)
+      { return robust_->residual(v.norm()); }
```
dellaert: This is suspect. Where did the whiten go. If unit tests did not catch this, it means a unit test is missing :-) We should probably add unit tests for regular and some robust model, e.g. Huber.

ProfFan: 2nd

varunagrawal: So there was a unit test which was commented out. It had the same TODO note as this issue, so I uncommented it and made the changes, and it passed without issues. :slightly_smiling_face: 

But yes, I was on the fence wrt whether this should be a new PR. Thanks for clarifying this.

gtsam.h
varunagrawal: Done!

varunagrawal:
@dellaert undid the last commit. Will make a new PR with the changes. Mike Bosse is not a collaborator on Github, so we'll have to ask him separately to help out on this.

matlab/gtsam_examples/VisualizeMEstimators.m
varunagrawal: Yup

matlab/gtsam_examples/VisualizeMEstimators.m
varunagrawal: This is done. Residual functions are implemented and wrapped.

dellaert:
Merge at will

gtsam/linear/NoiseModel.cpp
```diff
@@ -774,6 +799,16 @@ Cauchy::Cauchy(double k, const ReweightScheme reweight) : Base(reweight), k_(k),
   }
 }
 
+double Cauchy::weight(double error) const {
+  return ksquared_ / (ksquared_ + error*error);
+}
+
+double Cauchy::residual(double error) const {
+  const double xc2 = error / k_;
```
michaelbosse: minor nit: `xc2` is usually `error*error/ksquared_`, and then can do `std::log(1 + xc2)` below.

dellaert: @varunagrawal I know you can't finish this yet because of other priorities, but could you incorporate this change into the new PR?

PR #132, opened 2019-10-03
boost libraries fix

mbway:
I encountered a problem when building gtsam and linking with boost installed through anaconda, because it seems that none of the debug libraries were available in that package. This was OK because I was building in release mode, however because the `Boost_*_LIBRARY_DEBUG` symbols were all empty, the `GTSAM_BOOST_LIBRARIES` variable ended up being `optimized;<library>;optimized;<library>......;debug;debug;debug;debug` which is invalid syntax because `debug` and `optimized` has to immediately be followed by a path. It is also invalid to have a dangling `debug` or `optimized` at the end of the string.

I don't think my fix is optimal, but it does work.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/132)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,32 +178,49 @@ option(GTSAM_DISABLE_NEW_TIMERS "Disables using Boost.chrono for timing" OFF)
 # problems with Boost versions newer than FindBoost.cmake was prepared to handle,
 # so we downgraded this to classic filenames-based variables, and manually adding
 # the target_include_directories(xxx ${Boost_INCLUDE_DIR})
-set(GTSAM_BOOST_LIBRARIES
-  optimized ${Boost_SERIALIZATION_LIBRARY_RELEASE}
-  optimized ${Boost_SYSTEM_LIBRARY_RELEASE}
-  optimized ${Boost_FILESYSTEM_LIBRARY_RELEASE}
-  optimized ${Boost_THREAD_LIBRARY_RELEASE}
-  optimized ${Boost_DATE_TIME_LIBRARY_RELEASE}
-  optimized ${Boost_REGEX_LIBRARY_RELEASE}
-  debug ${Boost_SERIALIZATION_LIBRARY_DEBUG}
-  debug ${Boost_SYSTEM_LIBRARY_DEBUG}
-  debug ${Boost_FILESYSTEM_LIBRARY_DEBUG}
-  debug ${Boost_THREAD_LIBRARY_DEBUG}
-  debug ${Boost_DATE_TIME_LIBRARY_DEBUG}
-  debug ${Boost_REGEX_LIBRARY_DEBUG}
-)
+set(GTSAM_BOOST_LIBRARIES "")
+foreach(TMP
+        ${Boost_SERIALIZATION_LIBRARY_RELEASE}
+        ${Boost_SYSTEM_LIBRARY_RELEASE}
+        ${Boost_FILESYSTEM_LIBRARY_RELEASE}
+        ${Boost_THREAD_LIBRARY_RELEASE}
+        ${Boost_DATE_TIME_LIBRARY_RELEASE}
+        ${Boost_REGEX_LIBRARY_RELEASE})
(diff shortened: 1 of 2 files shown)


jlblancoc:
@mbway I'm curious if the former version of that messy part of the CMake script would work in your case. 
Please, give it a try, changing manually back to what was removed here: 

https://github.com/borglab/gtsam/commit/a0fce4257fe4e5fe483ec9d048ee62bac803df8f#diff-af3b638bc2a3e6c650974192a53c7291R148-R166

that is, using: 

```
# Circa line 147
set(GTSAM_BOOST_LIBRARIES
  Boost::serialization
  Boost::system
  Boost::filesystem
  Boost::thread
  Boost::date_time
  Boost::regex
)
```
and:
```
# Circa line 170
if(Boost_TIMER_LIBRARY)
      list(APPEND GTSAM_BOOST_LIBRARIES Boost::timer Boost::chrono)
```
and so on (check the diff in the link above). 

It's a **way** shorter and more readable solution, but we had to roll it back for the reasons explained in the code comments.

Whatever is your result, please also report what's your version of Boost and CMake, since it all depends on new-enough versions of both of them.

ProfFan:
Any idea why are we including debug libraries in a release build?

jlblancoc:
@ProfFan :

> Any idea why are we including debug libraries in a release build?

What do you mean? The list of "debug" and "optimized" libraries used in cmake actually is interpreted by cmake such that only one of the sets is used depending on CMAKE_BUILD_TYPE.

ProfFan:
I see. Seems that we have not checked the nullity of these variables.

ProfFan:
LGTM. But still I can't see a problem with IMPORTED targets like `Boost::<C>` @jlblancoc mentioned. FindBoost is a cmake component, it should work I think?

jlblancoc:
> But still I can't see a problem with IMPORTED targets like

The problem was that if a user has a version of Boost newer than the latest version "known" by its CMake's `FindBoost.cmake` module, imported targets seemed not to work (sigh). A too modern Boost or a too old CMake raised this problem.

An alternative could be, perhaps, importing the latest FindBoost.cmake from CMake master branch into the gtsam repo. More work to maintain and keep that file up-to-date, but the alternative is actually not much better... having to fight against all kind of problems like the one you tried to solve with this PR. 

What's your opinion about that alternative solution? @dellaert , @varunagrawal ??

PS: Bottom line problem is, actually IMHO, that Boost doesn't want to use CMake, so they don't provide proper exported cmake targets in their packages. But that's out of our control.

mbway:
The versions I'm using:
```
    boost:                  1.70.0-py37h9de70de_1      conda-forge
    boost-cpp:              1.70.0-h8e57a91_2          conda-forge
    cmake:                  3.15.4-hf94ab9c_0          conda-forge
```
using `Boost::*` rather than `${Boost_*_LIBRARY_*}` did work for me

jlblancoc:
Ok, I'll open a new PR reverting a0fce4257fe4e5fe483ec9d048ee62bac803df8f and including the latest FindBoost.cmake to prevent problems in users with older cmake versions, and let's move the discussion there...  I will like @chrisbeall to test that version, since he had problems with the `Boost::xxx` targets some time ago in his Mac .

So, let's close this PR...

cynosure4sure:
@jlblancoc sorry to comment on an old thread but i didn't want to open a new issue.

When moving from GTSAM 4.0.2 to 4.0.3, I could not compile my ROS code due to following error:

>  /usr/bin/ld: cannot find -lBoost::timer  
collect2: error: ld returned 1 exit status


I am on:
>     ubuntu:                  18.04
>     ROS:                     melodic
>     cmake:                  3.10.2
>     boost:                    1.65.1.0ubuntu1

I applied all the changes mentioned in this PR to the CMakeLists.txt of 4.0.3 and the code compiles again. I understand the issue of cmake vs boost version finding each other but I am not sure what is the correct solution here. Any suggestion/advice how to resolve this would be appreciated. Thank you

PR #134, opened 2019-10-06
[WIP] New Python Wrapper

ProfFan:
This giant pile of commits definitely needs to be cleaned up. Since this originates from a fork, we probably need to cherry-pick commits this time.

@matthewsklar We should meet in the coming week or so to figure this out :)

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/134)


octopuscabbage:
Hi, We are planning to use the python API for a project. Do you know likely when this will be merged? Also does this wrap factors like the GPS factor or allow declaration of new factors in python? (We would like to use a GPS factor in our project) 

If this is an inappropriate place to ask this, I can delete this and ask somewhere else.

dellaert:
This is a modernization of the existing wrapper. In theory you can already do python wrapping (with the cython wrapper machinery) of a project, but we are planning to add an example of that soon.

dellaert:
Does this supersede PR #66 ?

ProfFan:
@dellaert It should, but need confirmation.

matthewsklar:
@dellaert It does

ProfFan:
Closing in lieu of my branch

PR #130, opened 2019-10-03
Bump version to 4.0.2

Ellon:
Hi,

I noticed the version was not bumped after 4.0.2 release, so I bumped it on this PR.

It was also not bumped for 4.0.1... was this on purpose ?

Cheers!

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/130)

PR #136, opened 2019-10-07
Update outdated GTSAM version number

jlblancoc:
`GTSAMConfigVersion.cmake` was still reporting gtsam to be version 4.0.0 due to outdated cmake commands. 

Since 4.0.2 was already released, I think it makes sense to bump it to 4.0.3 and keep it until 4.0.3 (or 4.1.0, etc.) is released, and so on. 

Thoughts?

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/136)


dellaert:
So, I like the 4.0.3 idea: it makes so we don’t have to move a tag. But there is already a PR etc with a bunch of changes on versioning, right? How does this relate and in what order do things need to happen now? Adding @varunagrawal and @Ellon

jlblancoc:
It seems the other PR https://github.com/borglab/gtsam/pull/130 was already merged, but into "master". This one goes against "develop" (!).

dellaert:
Aha, yes, if those changes make sense, should you cherry pick them here as well?

jlblancoc:
Thanks for the link to the "contributing" page, it **really** explains it all! :-)

> Aha, yes, if those changes make sense, should you cherry pick them here as well?

In fact, this PR bumps to 4.0.3, which is in conflict with the other PR which bumped to 4.0.2 :-)

So, if you are happy with the numbering scheme, this one can be merged into develop as is.

dellaert:
Nono, there is no mess. The only weird thing is that several commits seemed to have happened straight on develop *without* a pull request. That's a nono :-)
Next comment will say what I think is happening, should be happening.

dellaert:
Looping in @varunagrawal @ProfFan @chrisbeall 
As far as I can tell:

1. develop has three raw commits. Bad idea, and I just made develop a "protected branch" to prevent that.
2. master now has the correct version tag, 4.0.2, but the "4.0.2" tag still points to the commit before that. It probably also needs the version script that was added to develop? Proposed solution: cherry-pick what is needed from develop, do a PR to make sure we agree, then force-push the tag. I we agree, I would like @ProfFan to do this.
3. This PR should just update the version tag to **4.0.2** in develop, to match the current release.

Thoughts?

dellaert:
BTW @ProfFan is coordinating a 4.1 release. So this 4.0.2 version here will be very fleeting.

Ellon:
>     1. develop has three raw commits. Bad idea, and I just made develop a "protected branch" to prevent that.

Shouldn't master be made a "protected branch" too (if it isn't the case already) ?

>     2. master now has the correct version tag, 4.0.2, but the "4.0.2" tag still points to the commit before that. It probably also needs the version script that was added to develop? Proposed solution: cherry-pick what is needed from develop, do a PR to make sure we agree, then force-push the tag. I we agree, I would like @ProfFan to do this.

I think this is a good idea.

>     3. This PR should just update the version tag to **4.0.2** in develop, to match the current release.

So develop would just match master in content, but it would be merged only later, on a 4.1 version for example ? If that's the case I also think it's a good solution.

varunagrawal:
I agree with Frank's proposal.

1. Cherry-pick the commits from #130, #133 (maybe? to be consistent), #136 onto a new branch.
2. Undo the changes on develop via either soft or hard resets.
3. Create new PR for the new changes, maintaining version 4.0.2.
4. Move 4.0.2 tag to the correct location (and possibly update the release on `Releases` as well).
5. Post to Google Group (I can do this).

Sounds like a plan?

dellaert:
@varunagrawal No, that is *not* the plan I suggested. Just agree or disagree with the plan I put in the comment. Don't come up with a new plan.

varunagrawal:
Oh sorry about that. I was just spec'ing out your original plan to remove ambiguity.

I agree with your plan.

dellaert:
OK. For everyone else: my plan does not involve undoing changes on develop and master via either soft or hard resets. Action items:
1. José will just amend this PR to 4.0.2 to **develop**, I'll approve and merge.
2. Ellon (if he agrees) cherry-picks the changes from develop that are needed for proper versioning into a PR that goes on **master**.
3. @varunagrawal approves, merges, and pushes the tag. Bonus: post on Google group.
:-)

Ellon:
I agree with 2. I'll do another PR later today with the commit that was cherry picked on develop.

ProfFan:
Will do this tonight, as @dellaert suggested. Protecting "develop" is a good idea I think.

For the plan, let me reconcile:
1. 3 raw commits in develop will be cherry-picked into master as a PR
2. 4.0.2 tag will now be pointed to latest master

Is that correct? @dellaert

dellaert:
@ProfFan nope, in latest plan you are out. Task 3 was assigned to @varunagrawal

jlblancoc:
> José will just amend this PR to 4.0.2 to develop, I'll approve and merge.

@dellaert Done.

Ellon:
Step 2. of the plan is on #137

PR #137, opened 2019-10-07
Fix GTSAM version detection for 4.0.2

Ellon:
As discussed on #136 this PR cherry-picks commits from develop that are needed to let CMake detect the version of GTSAM.

I think only two commits are needed:
* One that installs GTSAMConfigVersion.cmake
* One that remove obsolete FindGTSAM*.cmake files

Note that this PR proposes a merge into master, as I see it as a (hot?) fix that can be added to 4.0.2 if the tag is moved properly (again, as discussed on #136).

Let me know if I should add any other commits.

Best,

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/137)


varunagrawal:
@Ellon I'm slightly confused. You've taken the version bump commit that you pushed into master, then cherry picked commits from develop without the commit from @jlblancoc that bumps the version.

I am not sure how that would affect a future release in terms of merge conflicts from develop to master. Can you make this PR have only cherry picked commits from develop? (specifically b3dccc6ef3e24eb11b60f03f857f177feff5714f)

dellaert:
But, master already has the 4.0.2 version....

varunagrawal:
But doesn't that same argument apply for commit b3dccc6? When develop next gets merged into master, master will already have 4.0.2 as the version (when this PR gets merged to master).

dellaert:
@varunagrawal you can now merge, move tag

PR #144, opened 2019-10-12
Use -fPIC flag when building gtsam libraries (gcc/clang)

jlblancoc:
This flag is required to ensure proper code generation for shared libraries in all possible scenarios. 
I've tested everything builds and all tests pass after the change. 

Refer to discussions, e.g.:
- https://github.com/borglab/gtsam/pull/new/add-fPIC-flag
- https://stackoverflow.com/questions/5311515/gcc-fpic-option

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/144)

### DIFF
--- a/cmake/GtsamBuildTypes.cmake
+++ b/cmake/GtsamBuildTypes.cmake
@@ -94,7 +94,8 @@ if(MSVC)
   set(GTSAM_COMPILE_OPTIONS_PRIVATE_TIMING          /MD /O2  CACHE STRING "(User editable) Private compiler flags for Timing configuration.")
 else()
   # Common to all configurations, next for each configuration:
-  set(GTSAM_COMPILE_OPTIONS_PRIVATE_COMMON          -Wall CACHE STRING "(User editable) Private compiler flags for all configurations.")
+  # "-fPIC" is to ensure proper code generation for shared libraries
+  set(GTSAM_COMPILE_OPTIONS_PRIVATE_COMMON          -Wall -fPIC CACHE STRING "(User editable) Private compiler flags for all configurations.")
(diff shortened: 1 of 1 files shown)


dellaert:
Perhaps a good one fro @mikesheffler to approve - if I remember correctly he also compiles on Windows, right?

jlblancoc:
ping? :-) 

PS: I rebased the PR on top of current `develop`.

dellaert:
@mikesheffler is this something you could review/check on your end?

mikesheffler:
Hey, guys. I didn't see _any_ of the earlier messages in this thread. Hmm ...

I do indeed build on Windows sometimes, but I don't have a Windows build setup right this moment. I'm deploying it on my Ubuntu 16.04 and 18.04 x64 machines and a Jetson TX2 , which is aarch64. All gcc.

I think  @jlblancoc is right; if you don't specify `-fPIC`, successful dynamic linking of GTSAM would be basically a matter of luck. I usually statically link GTSAM, so it's not a problem. **BUT**, if I write custom factors that are MEX functions for MATLAB (_e.g._, if you follow [Jing  Dong's GTSAM tutorial examples](https://github.com/dongjing3309/gtsam-examples)), the MEX functions (usually) won't work without `-fPIC` because they are dynamically linking GTSAM, which was probably not set up with `-fPIC`.

jlblancoc:
> if you don't specify -fPIC, successful dynamic linking of GTSAM would be basically a matter of luck.

Exactly. I can't remember what I was doing, but I ended up stuck with a cryptic link error in my application, hence this PR to fix it.

PR #146, opened 2019-10-12
Fix/clang warnings

dellaert:
Fix warnings on clang. Most are missing overrides from @varunagrawal 's last PR. Probably warnings did not show up on Linux. But I also resolved calls to `ptr_fun` which was deprecated in c++11.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/146)

### DIFF
--- a/gtsam/linear/GaussianBayesNet.cpp
+++ b/gtsam/linear/GaussianBayesNet.cpp
@@ -187,16 +187,16 @@ namespace gtsam {
   }
 
   /* ************************************************************************* */
-  double GaussianBayesNet::logDeterminant() const
-  {
+  double GaussianBayesNet::logDeterminant() const {
     double logDet = 0.0;
-    for(const sharedConditional& cg: *this) {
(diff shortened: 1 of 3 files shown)


gtsam/linear/GaussianBayesNet.cpp
```diff
@@ -187,16 +187,16 @@ namespace gtsam {
...
+  double GaussianBayesNet::logDeterminant() const {
     double logDet = 0.0;
-    for(const sharedConditional& cg: *this) {
-      if(cg->get_model()) {
+    for (const sharedConditional& cg : *this) {
+      if (cg->get_model()) {
         Vector diag = cg->R().diagonal();
         cg->get_model()->whitenInPlace(diag);
-        logDet += diag.unaryExpr(ptr_fun<double,double>(log)).sum();
+        logDet += diag.unaryExpr([](double x) { return log(x); }).sum();
```
ProfFan: Lambdas! :+1:

PR #151, opened 2019-10-17
fix default parameters in constructor of IncrementalFixedLagSmoother.h

### LINKED ISSUE #150: findUnusedFactorSlots should be true
## Description

The incremental fixed lag smoother creates an iSAM instance, but does not set the `findUnusedFactorSlots` flag to true, even though the docs in iSAM2Params.h say: "When you will be removing many factors, e.g. when using ISAM2 as a fixed-lag smoother, enable this option to add factors in the first available factor slots, to avoid accumulating NULL factor slots, at the cost of having to search for slots every time a factor is added."

## Steps to reproduce

just run incremental fixed lag smoother and you will see number of factors grow...

## Expected behavior

Should be approx constant.

## Proposed fix

Change construction of iSAM2 instance in incremental fixed lag smoother...

evanrobot:
Fixes #150

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/151)

### DIFF
--- a/gtsam_unstable/nonlinear/IncrementalFixedLagSmoother.h
+++ b/gtsam_unstable/nonlinear/IncrementalFixedLagSmoother.h
@@ -39,8 +39,8 @@ public:
 
   /** default constructor */
   IncrementalFixedLagSmoother(double smootherLag = 0.0,
-      const ISAM2Params& parameters = ISAM2Params()) :
-      FixedLagSmoother(smootherLag), isam_(parameters) {
+      const boost::optional<ISAM2Params>& parameters = boost::none) :
+      FixedLagSmoother(smootherLag), isam_(parameters ? (*parameters) : getDefaultParams()) {
   }
 
   /** destructor */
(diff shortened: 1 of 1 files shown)


gtsam_unstable/nonlinear/IncrementalFixedLagSmoother.h
```diff
@@ -114,6 +114,14 @@ class GTSAM_UNSTABLE_EXPORT IncrementalFixedLagSmoother: public FixedLagSmoother
   const ISAM2Result& getISAM2Result() const{ return isamResult_; }
 
 protected:
+
+  /* Create default parameters */
+  ISAM2Params getDefaultParams() const {
```
dellaert: Should be static, and renamed to `DefaultISAM2Params'

dellaert:
Added a comment, but PR seems to be closed?

gtsam_unstable/nonlinear/IncrementalFixedLagSmoother.h
evanrobot: Done

PR #153, opened 2019-10-18
Fix python wrappers for BetweenFactorVector and PreintegratedCombined…

tardifjp:
…Measurements and

added Vector type to the BetweenFactor generation
Created PreintegrationCombinedParams to add constructor for PreintegratedCombinedMeasurements

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/153)

### DIFF
--- a/gtsam.h
+++ b/gtsam.h
@@ -2812,7 +2812,6 @@ virtual class ImuFactor: gtsam::NonlinearFactor {
 };
 
 #include <gtsam/navigation/CombinedImuFactor.h>
-//typedef PreintegratedCombinedMeasurements::Params PreintegrationCombinedParams;
 virtual class PreintegrationCombinedParams : gtsam::PreintegrationParams {
   PreintegrationCombinedParams(Vector n_gravity);
 
--- a/gtsam.h
+++ b/gtsam.h
@@ -2405,7 +2405,7 @@ virtual class PriorFactor : gtsam::NoiseModelFactor {
 
 
 #include <gtsam/slam/BetweenFactor.h>
-template<T = {gtsam::Point2, gtsam::Point3, gtsam::Rot2, gtsam::Rot3, gtsam::Pose2, gtsam::Pose3, gtsam::imuBias::ConstantBias}>
+template<T = {Vector,gtsam::Point2, gtsam::Point3, gtsam::Rot2, gtsam::Rot3, gtsam::Pose2, gtsam::Pose3, gtsam::imuBias::ConstantBias}>
 virtual class BetweenFactor : gtsam::NoiseModelFactor {
   BetweenFactor(size_t key1, size_t key2, const T& relativePose, const gtsam::noiseModel::Base* noiseModel);
   T measured() const;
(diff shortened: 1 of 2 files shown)


gtsam/navigation/CombinedImuFactor.h
```diff
@@ -54,6 +48,61 @@ typedef ManifoldPreintegration PreintegrationType;
...
+    ar& BOOST_SERIALIZATION_NVP(biasOmegaCovariance);
+    ar& BOOST_SERIALIZATION_NVP(biasAccOmegaInt);
+  }
+
+public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+};
+
+  
+#ifdef GTSAM_TANGENT_PREINTEGRATION
```
dellaert: Is there a reason this moved? Maybe move it back as it seems orthogonal

gtsam.h
```diff
@@ -2812,7 +2812,34 @@ virtual class ImuFactor: gtsam::NonlinearFactor {
 };
 
 #include <gtsam/navigation/CombinedImuFactor.h>
+//typedef PreintegratedCombinedMeasurements::Params PreintegrationCombinedParams;
```
dellaert: kill :-)

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -54,6 +48,61 @@ typedef ManifoldPreintegration PreintegrationType;
  *     Robotics: Science and Systems (RSS), 2015.
  */
 
+struct GTSAM_EXPORT PreintegrationCombinedParams : PreintegrationParams {
```
dellaert: Add doxygen /** */ block above?

gtsam.h
tardifjp: Yes, that's from when I was trying to leave CombinedImuFactor alone!

gtsam/navigation/CombinedImuFactor.h
tardifjp: Ok, I moved it back to the top

gtsam/navigation/CombinedImuFactor.h
tardifjp: ok, added a quick description of the new struct

dellaert:
Awesome. Will merge when check passes.
Thanks again !!!!

tardifjp:
One last thing I haven't pushed. In CombinedImuFactor.h, I tried changing:
    ar& BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegratedRotation::Params);
for
    ar& BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegrationParams);
in the serialize function.

It would seem logical since that's the actual superclass. It compiles just fine. However, I don't have a way to validate that change.

dellaert:
> One last thing I haven't pushed. In CombinedImuFactor.h, I tried changing:
> ar& BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegratedRotation::Params);
> for
> ar& BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegrationParams);
> in the serialize function.
> 
> It would seem logical since that's the actual superclass. It compiles just fine. However, I don't have a way to validate that change.

@tardifjp I will merge, but could you do one last thing for us and open up an issue for the serialization bug. It is a bug, and you found it. 

To verify it you would use a serialization test to and from a string. So, adding a unit test...

tardifjp:
Sure, I'll create the issue.

PR #157, opened 2019-10-20
        upgrade colamd to suite-sparse-5.4.0

MandyXie:
Replaced CCOLADM and SuiteSparse_config in 3rdparty with folders from suite-sparse-5.4.0, and passed all the unittests.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/157)


dellaert:
One of the things Tim mentioned is that the license changed, but I see no change to the LICENSE files. Puzzling, investigate?

chrisbeall:
License was changed to BSD about two years ago? https://bitbucket.org/gtborg/gtsam/pull-requests/301/ccolamd-now-bsd-3/diff

dellaert:
Ah, ok, we’ll, then, never mind :-)

PR #163, opened 2019-11-10
Fix bug when constructing VariableSlots

michaelbosse:
VariableSlots were not constructable from GaussianFactorGraphs that contained null factors.
This caused seg faults when generating Jacobian factors from such graphs.

This bug was discovered by Guillermo Dueas Arana <gdueasar@hawk.iit.edu>:
https://groups.google.com/d/msgid/gtsam-users/5d7b05a9-2200-43e2-8f1d-ddc878d6ab5e%40googlegroups.com.

This fix was developed by myself and Francesco Papi <fpapi@zoox.com>

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/163)

### DIFF
--- a/gtsam_unstable/examples/FixedLagSmootherExample.cpp
+++ b/gtsam_unstable/examples/FixedLagSmootherExample.cpp
@@ -145,5 +145,18 @@ int main(int argc, char** argv) {
     cout << setprecision(5) << "    Key: " << key_timestamp.first << "  Time: " << key_timestamp.second << endl;
   }
 
+  // get the linearization point
+  Values result = smootherISAM2.calculateEstimate();
+
+  // create the factor graph
+  auto &factor_graph = smootherISAM2.getFactors();
+
+  // linearize to a Gaussian factor graph
+  boost::shared_ptr<GaussianFactorGraph> linear_graph = factor_graph.linearize(result);
+
+  // this is where the segmentation fault occurs
+  Matrix A = linear_graph->jacobian().first;
+  cout << " A = " << A << endl;
+
   return 0;
 }
--- a/gtsam_unstable/examples/FixedLagSmootherExample.cpp
+++ b/gtsam_unstable/examples/FixedLagSmootherExample.cpp
@@ -145,18 +145,19 @@ int main(int argc, char** argv) {
     cout << setprecision(5) << "    Key: " << key_timestamp.first << "  Time: " << key_timestamp.second << endl;
   }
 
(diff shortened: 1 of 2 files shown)


gtsam_unstable/examples/FixedLagSmootherExample.cpp
```diff
@@ -145,5 +145,18 @@ int main(int argc, char** argv) {
     cout << setprecision(5) << "    Key: " << key_timestamp.first << "  Time: " << key_timestamp.second << endl;
   }
 
+  // get the linearization point
```
dellaert: I understand this is the code that diagnoses the problem. However, this file is an example. If we leave the code below, comment the block to say what this is an example of :-)

gtsam_unstable/examples/FixedLagSmootherExample.cpp
```diff
@@ -145,5 +145,18 @@ int main(int argc, char** argv) {
...
+  Values result = smootherISAM2.calculateEstimate();
+
+  // create the factor graph
+  auto &factor_graph = smootherISAM2.getFactors();
+
+  // linearize to a Gaussian factor graph
+  boost::shared_ptr<GaussianFactorGraph> linear_graph = factor_graph.linearize(result);
+
+  // this is where the segmentation fault occurs
```
dellaert: Hopefully not anymore?

gtsam_unstable/examples/FixedLagSmootherExample.cpp
```diff
@@ -145,5 +145,18 @@ int main(int argc, char** argv) {
...
+  Matrix A = linear_graph->jacobian().first;
+  cout << " A = " << A << endl;
```
dellaert: Print out “Jacobian =“ ?

gtsam_unstable/examples/FixedLagSmootherExample.cpp
```diff
@@ -145,5 +145,18 @@ int main(int argc, char** argv) {
     cout << setprecision(5) << "    Key: " << key_timestamp.first << "  Time: " << key_timestamp.second << endl;
   }
 
+  // get the linearization point
+  Values result = smootherISAM2.calculateEstimate();
+
+  // create the factor graph
+  auto &factor_graph = smootherISAM2.getFactors();
```
dellaert: Nit: in c++ we now try to follow the Google conventions, so factor graph and linear graph should become camelCase.

PR #169, opened 2019-11-19
Make include_directories relative for install interface

lsolanka:
This patch removes `${CMAKE_INSTALL_PREFIX}` from in front of gtsam target include directories in order to make these paths relative to the install prefix. Otherwise the paths are hard-coded and thus the whole installation is not relocatable to another folder.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/169)


jlblancoc:
:+1: Thanks! 
Let me check if this works for generating .deb packages, and I'll approve for someone else to merge.

jlblancoc:
ooops... totally forgot about this one, sorry. 
The patched version is now building [here](https://launchpad.net/~joseluisblancoc/+archive/ubuntu/gtsam-develop/+build/18171255), when it finished let me check if the resulting .deb is suitable for building a user program and I'll be confident everything is right ;-)

PR #170, opened 2019-11-22
Fix potential infinite loop in GTSAM's nonlinear optimizer.

erik-nelson:
This PR fixes a possible infinite loop within the NonlinearOptimizer class.

To reproduce the issue:
1) Build a factor graph with at least one factor whose error evaluates to inf or NaN on a given set of values (for example, I made a fake factor that always returns inf error).
2) Build a `LevenbergMarquardtOptimizer`
3) Run `LevenbergMarquardtOptimizer::optimize()`

The `optimize()` function will get stuck in [this loop](https://github.com/borglab/gtsam/blob/develop/gtsam/nonlinear/NonlinearOptimizer.cpp#L91) indefinitely.

While this is not a common use case and should be avoided at the user level, it is better to break out of `optimize()` and still return that the optimization did not converge than to hang the program in an infinite loop.

I have only verified that this bug occurs when using `LevenbergMarquardOptimizer`, but I would imagine that this fix is generally a good thing to have. The reason that the issue occurs in `LevenbergMarquardtOptimizer` specifically is that we do not increase the state's `iteration` field unless `decreaseLambda` succeeds ([this](https://github.com/borglab/gtsam/blob/develop/gtsam/nonlinear/internal/LevenbergMarquardtState.h#L93) is the only location in the LM optimizer where iteration number is increased). 

There are likely other ways to fix this issue specifically for LM, such as increasing `state_.iteration` in at least one other location, but this PR blanket fixes the issue for all optimizers.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/170)

### DIFF
--- a/gtsam/nonlinear/NonlinearOptimizer.cpp
+++ b/gtsam/nonlinear/NonlinearOptimizer.cpp
@@ -101,7 +101,7 @@ void NonlinearOptimizer::defaultOptimize() {
       cout << "newError: " << error() << endl;
   } while (iterations() < params.maxIterations &&
            !checkConvergence(params.relativeErrorTol, params.absoluteErrorTol, params.errorTol,
-                             currentError, error(), params.verbosity));
+                             currentError, error(), params.verbosity) && std::isfinite(currentError));
 
   // Printing if verbose
   if (params.verbosity >= NonlinearOptimizerParams::TERMINATION) {


varunagrawal:
Hi Erik, thank you for the PR!

We are already aware of such an issue in the codebase. However I am wondering what would be a case where a factor returns a `NaN` or `Inf`. It seems like that would be a poor model of the problem and no optimizer would do a good job on that. I believe if you use GaussNewton, you'll end up getting an Indeterminate Linear System error when the optimizer tries to perform the Cholesky decomposition.

If you could please explain the use case some more, that would help us understand the various issues in a broader context.

erik-nelson:
I completely agree -- it's certainly a user error when a factor returns an `inf` or `NaN` error for a given problem, and that should be corrected on the user end. The fix here is simply to make it such that in such error cases, we do not hang the program, and instead terminate the optimization (`checkConvergence` will still return `false`), allowing the user to continue and potentially recover by themselves.

An error of `inf` can happen in many different ways, e.g. mistakenly dividing by zero inside of a factor. Happy to close this if we feel it is more desirable to loop forever than to exit from it when the user makes an error.

dellaert:
I think Erik’s idea is good. But, I don’t understand how the change fixes it. Should it be `!isinfinite`? I might not be parsing the code right...

erik-nelson:
We are using `std::isfinite` since it checks against both `inf` and `NaN`. The idea is that we continue looping so long as:

1) We haven't reached max iterations
2) We haven't converged
3) We have finite error (if error ever evalutes to inf or nan, conditions (1) and (2) will always evaluate to true, and we will continue looping)

erik-nelson:
Oh haha I think I see what you mean @dellaert. This is `std::isfinite`. "isfinite" looks a lot like "infinite" :)

dellaert:
Right!!!!

PR #183, opened 2019-12-07
Enable and fix unit tests in different archs

### LINKED ISSUE #21: Needs C++17 over-aligned new operator
While building tests: 

```
/home/jlblanco/code/gtsam-jlblancoc/gtsam/nonlinear/Expression-inl.h: In instantiation of ‘gtsam::Expression<T>::Expression(const T&) [with T = gtsam::Unit3]’:
/home/jlblanco/code/gtsam-jlblancoc/gtsam/slam/tests/testEssentialMatrixFactor.cpp:149:39:   required from here
/home/jlblanco/code/gtsam-jlblancoc/gtsam/nonlinear/Expression-inl.h:32:11: warning: ‘new’ of type ‘gtsam::internal::ConstantExpression<gtsam::Unit3>’ with extended alignment 32 [-Waligned-new=]
     root_(new internal::ConstantExpression<T>(value)) {
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jlblanco/code/gtsam-jlblancoc/gtsam/nonlinear/Expression-inl.h:32:11: note: uses ‘void* operator new(std::size_t)’, which does not have an alignment parameter
/home/jlblanco/code/gtsam-jlblancoc/gtsam/nonlinear/Expression-inl.h:32:11: note: use ‘-faligned-new’ to enable C++17 over-aligned new support
```

Need to investigate a bit more, but it's likely due to Eigen memory constraints.

Are you ok with adding `-faligned-new` as a private/public flag to gtsam libraries to solve this? Otherwise, I guess this may lead to random crashes if "optimizing for native" is enabled.

### LINKED ISSUE #168: Unit Test Tolerances
##  Description

When running the Unit tests in some 32bit architectures, some of the unit test `assert_eaual` assertions fail with deltas in the order of 1e-6.

Looking into this further, the `expected` results in many of the tests (all of the failing tests, but many of the non-failing ones too) are calculated using functions from `numericalDerivative.h`. By default, the increment for numerical derivatives is 1e-5. All of the tests for the `numericalDerivative.h` functions have a tolerance of 1e-5. However throughout the rest of the code base, there seems to be an expectation of 1e-9 tolerances.

Is it OK that tests based on code that has only been proven to generate results within a tolerance of 1e-5 are expected to pass with tolerances of 1e-9?

This is not going to be easy to reproduce without our build environment - but I can test any suggestions in our builds.

However, I am more interested in answering whether or not the currently used tolerances make sense than a specific fix.

## Failing Tests
```
gtsam/geometry/tests/testOrientedPlane3.cpp:174: Failure: "assert_equal(H_expected_numerical, H_actual, 1e-9)"
gtsam/geometry/tests/testOrientedPlane3.cpp:180: Failure: "assert_equal(H_expected_numerical, H_actual, 1e-9)"
gtsam/geometry/tests/testPose3.cpp:663: Failure: "assert_equal(expectedH1, actualH1)"
gtsam/geometry/tests/testPose3.cpp:664: Failure: "assert_equal(expectedH2, actualH2)" 
gtsam/geometry/tests/testUnit3.cpp:384: Failure: "assert_equal(H_expected_numerical, H, 1e-9)"
```

### LINKED ISSUE #181: testSymbolicFactorGraph unit test failure
As part of debugging #168, I found a new issue running the unit tests for i386 (for some reason it works right on amd64): 

```
 68/237 Test  #68: testSymbolicFactorGraph ................***Failed    0.08 sec
Not equal:
expected:
: cliques: 4, variables: 6
expected:
- P( e0 l0 b0)
expected:
| - P( s0 | b0 l0)
expected:
| - P( t0 | e0 l0)
expected:
| - P( x0 | e0)
actual:
: cliques: 4, variables: 6
actual:
- P( e0 l0 b0)
actual:
| - P( t0 | e0 l0)
actual:
| - P( x0 | e0)
actual:
| - P( s0 | b0 l0)
/home/jlblanco/code/gtsam/gtsam/symbolic/tests/testSymbolicFactorGraph.cpp:100: Failure: "assert_equal(asiaBayesTree, actual2)" 
There were 1 failures
```

@dellaert  Any idea where to start looking at to debug?

### LINKED ISSUE #182: testSubgraphPreconditioner bad_alloc exception under i386
Another spin off of #168 . 

Building for i386, the unit test testSubgraphPreconditioner fails. 
Here's a gdb stack trace: 

```
Thread 1 "testSubgraphPre" hit Catchpoint 1 (exception thrown), 0xb7a0d426 in __cxa_throw () from /usr/lib/i386-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0xb7a0d426 in __cxa_throw () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#1  0x0044eaf6 in Eigen::internal::throw_std_bad_alloc() ()
#2  0xb7e56d70 in gtsam::getSubvector(Eigen::Matrix<double, -1, 1, 0, -1, 1> const&, gtsam::KeyInfo const&, std::vector<unsigned long long, tbb::tbb_allocator<unsigned long long> > const&) ()
   from /home/jlblanco/code/gtsam/build/gtsam/libgtsam.so.4
#3  0xb7e59c22 in gtsam::SubgraphPreconditioner::solve(Eigen::Matrix<double, -1, 1, 0, -1, 1> const&, Eigen::Matrix<double, -1, 1, 0, -1, 1>&) const ()
   from /home/jlblanco/code/gtsam/build/gtsam/libgtsam.so.4
#4  0x004432f8 in SubgraphSolverSolvesTest::run(TestResult&) ()
#5  0x0046ee4c in TestRegistry::run(TestResult&) ()
#6  0x0046f45d in TestRegistry::runAllTests(TestResult&) ()
#7  0x004398da in main ()
```

Will have to re-run with debug info to investigate what's wrong...

jlblancoc:
This PR does: 
- [X] Enable running unit tests in Debian builds.
- [X] Relax numerical thresholds in unit tests. Fixes #168 
- [x] Fixes #182 
- [x] Fixes #181 
- [x] Fixes #21 ==> Done in #185 

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/183)


cmake/GtsamBuildTypes.cmake
```diff
@@ -9,6 +9,11 @@ function(list_append_cache var)
   set(${var} "${cur_value}" CACHE STRING "${MYVAR_DOCSTRING}" FORCE)
 endfunction()
 
+# Check for optional compiler flags, if supported:
+include(CheckCXXCompilerFlag)
+# See https://github.com/borglab/gtsam/issues/21
+check_cxx_compiler_flag(-faligned-new COMPILER_HAS_FALIGNED_NEW)
```
chrisbeall: What are the minimum compiler versions of gcc, clang, and VS which support this? If it's only very new compilers, I think we should still prefer to address Eigen alignment issues by appropriate placement of `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` so we don't abandon users who have no control over the compiler version in their build environment.

dellaert: Agreed. Please note GTSAM is used by *many* companies on a variety of platforms, including embedded ones where compiler versions might be lagging. Incidentally, This is also why C++17 and std:shared_ptr are not suited for 4.X versions...

dellaert:
Many thanks BTW! I’ll wait to approve before the discussion above is resolved, but this allows GTSAM to build on more platforms, so awesome!

cmake/GtsamBuildTypes.cmake
jlblancoc: FYI: I have removed this commit from the PR, since it was finally solved in a better way in #185

ProfFan:
@dellaert  I am not that comfortable with lowering tolerances that much... As in #168, I think the best way is to diagnose the problem in the numerical solutions, especially those implemented with `a * b`, where `a` can be very small so the result goes to zero...

ProfFan:
For example, in the SO4 PoC here:

Even we define `FLOAT_TYPE` to `float`,
```
==== TESTING a!=0 && b == 0 ====
a = 0.0000001192,   c2 = 0.0000000000,   c3 = 0.0000000000
    0.0000001192, c2_s = 0.5000000000, c3_s = 0.1666666716
a = 0.0000011921,   c2 = 0.0000000000,   c3 = 0.0000000000
    0.0000011921, c2_s = 0.4999999404, c3_s = 0.1666666716
a = 0.0000119209,   c2 = 0.0000000000,   c3 = 0.0000000000
    0.0000119209, c2_s = 0.4999994934, c3_s = 0.1666666716
a = 0.0001192093,   c2 = 0.0000000000,   c3 = 0.0000000000
    0.0001192093, c2_s = 0.4999950230, c3_s = 0.1666666716
a = 0.0011920929,   c2 = 0.5033164620,   c3 = 0.1374389529
    0.0011920929, c2_s = 0.4999503195, c3_s = 0.1666666567
a = 0.0119209290,   c2 = 0.4999610484,   c3 = 0.1665760130
    0.0119209290, c2_s = 0.4995032847, c3_s = 0.1666654795
a = 0.1192092896,   c2 = 0.4994073808,   c3 = 0.1665496230
    0.1192092896, c2_s = 0.4950332344, c3_s = 0.1665482819
a = 1.1920928955,   c2 = 0.4435229003,   c3 = 0.1552171558
    1.1920928955, c2_s = 0.4530631304, c3_s = 0.1552170664
```
The result with Taylor series is still accurate to `10^(-7)`

ProfFan:
(misclicked on close...)

jlblancoc:
Any advise on how to deal with this PR? I don't actually have much more time to address a more detailed analysis on optimal unit test thresholds... ;-)
I can't remember why I changed most 1e-9 to 1e-5 but it was probably because 1e-8 or 1e-7 was still not coarse enough for some of the architectures to pass the tests.

ProfFan:
@jlblancoc Have you looked at the PoC I posted? I think some  routines in GTSAM are actually not numerically optimal, so we can definitely first test on 32-bit architectures like i386 and arm and see what unit tests fail, and trace the code path to improve the algorithms. It is on my list for quite a while but I haven't got time...

dellaert:
@ProfFan if your last comment is a "nice to have" then add an issue and merge this PR? If it is a blocker please refresh us on what the way to proceed is.

ProfFan:
@jlblancoc My apologies for leaving this for a long time! My two cents on this PR are:
1. Merge this PR so we don't block the packaging
2. Add a macro for the precision (`#define GTSAM_UNITTEST_TOLERANCE 1e-5`) in the future

jlblancoc:
Hi, 
On a second thought, I found most unit tests used 1e-9 or 1e-5 as thresholds. So, I just added them as two CMake-configurable preprocessor variables: `GTSAM_UNITTEST_TOLERANCE_TIGHT` and `GTSAM_UNITTEST_TOLERANCE_LOOSE`. 

Will wait until Travis goes over this last commit before merging.

jlblancoc:
Update: after rebasing on top of develop, we now have a failing test: 

```
 48/241 Test  #48: testRot3 ...............................***Failed    0.00 sec
not equal:
expected[-7.99999999; 0; 0];
actual[-1.71681468; 0; 0];
/home/jlblanco/code/gtsam/gtsam/geometry/tests/testRot3.cpp:532: Failure: "assert_equal(w, actualw, 1e-15)" 
There were 1 failures
```

ProfFan:
@dellaert On second thoughts, I think the core cause of this is that some of the exotic architectures supported by Debian lacks proper float operations? Because I think `double` is an IEEE standard and should always be 64 bit wide...

jlblancoc:
> Hate to be a pain in the butt, but I don't understand what the use is of those two defines. They make the code uglier (IMHO) and I'm not getting under what circumstances anyone would change them. I don't see any logic to set them differently for different platforms, but maybe I'm missing it? I'm just afraid we're adding complexity that will never be used by anyone - but I'm happy to be convinced otherwise.

Hmm... it's totally true that the macros add one "level of indirection" to the reader. 

They *might* have a potential utility, though, although the gain may be marginal so I don't have a strong point defending them: this PR started when I detected tests that passed in x64, but failed in other archs like x32 (see #168 and 54b739f69585c1c723a859df17168ae4199a55e2 ). In those cases, it would have been faster to change the value in one place instead of changing it in many test points. 
OTOH... this is something that we won't be doing every day :-)

So, that was my motivation, but let me know what's your preference. It's 99% a style thing, not a big deal.

dellaert:
> > Hate to be a pain in the butt, but I don't understand what the use is of those two defines. They make the code uglier (IMHO) and I'm not getting under what circumstances anyone would change them. I don't see any logic to set them differently for different platforms, but maybe I'm missing it? I'm just afraid we're adding complexity that will never be used by anyone - but I'm happy to be convinced otherwise.
> 
> Hmm... it's totally true that the macros add one "level of indirection" to the reader.
> 
> They _might_ have a potential utility, though, although the gain may be marginal so I don't have a strong point defending them: this PR started when I detected tests that passed in x64, but failed in other archs like x32 (see #168 and [54b739f](https://github.com/borglab/gtsam/commit/54b739f69585c1c723a859df17168ae4199a55e2) ). In those cases, it would have been faster to change the value in one place instead of changing it in many test points.
> OTOH... this is something that we won't be doing every day :-)
> 
> So, that was my motivation, but let me know what's your preference. It's 99% a style thing, not a big deal.

I'd prefer we did not do it. I'd even go the opposite way: wherever there isa 1e-9, we remove it, as it is the default value for the parameter. Then, if there *is* a different value, we suspect there was a reason for it.

jlblancoc:
@dellaert : rolled back to good old plain numbers ;-) 

Approve and merge at your convenience.

debian/rules
```diff
@@ -13,13 +15,15 @@ export DH_VERBOSE = 1
...
-
 %:
 	dh $@ --parallel
 
-
 # dh_make generated override targets
 # This is example for Cmake (See https://bugs.debian.org/641051 )
 override_dh_auto_configure:
-	dh_auto_configure -- -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr -DGTSAM_BUILD_EXAMPLES_ALWAYS=OFF -DGTSAM_BUILD_TESTS=OFF -DGTSAM_BUILD_WRAP=OFF -DGTSAM_BUILD_DOCS=OFF -DGTSAM_INSTALL_CPPUNITLITE=OFF -DGTSAM_INSTALL_GEOGRAPHICLIB=OFF -DGTSAM_BUILD_TYPE_POSTFIXES=OFF -DGTSAM_BUILD_WITH_MARCH_NATIVE=OFF
+	dh_auto_configure -- -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr -DGTSAM_BUILD_EXAMPLES_ALWAYS=OFF -DGTSAM_BUILD_TESTS=ON -DGTSAM_BUILD_WRAP=OFF -DGTSAM_BUILD_DOCS=OFF -DGTSAM_INSTALL_CPPUNITLITE=OFF -DGTSAM_INSTALL_GEOGRAPHICLIB=OFF -DGTSAM_BUILD_TYPE_POSTFIXES=OFF -DGTSAM_BUILD_WITH_MARCH_NATIVE=OFF
```
dellaert, line 24: behavior change?

jlblancoc: this is one of the main reasons of this PR: to run unit tests on the PPA server to have extra feedback on other architectures, additional to Travis ;-)

jlblancoc: PS: We didn't enable tests from the beginning due to some errors, if I recall correctly, but this should be the preferred state

PR #189, opened 2019-12-11
Travis: upgrade to use clang-9

jlblancoc:
This is done to workaround apparent compiler internal errors, e.g. [here](https://travis-ci.com/borglab/gtsam/jobs/265834977).

Also, it seems it would be beneficial to #73. cc: @ProfFan . 

If you prefer an older clang version, let me know...

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/189)

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # common tasks before either build or test
-function prepare ()
+function configure()
 {
   set -e   # Make sure any error makes the script to return an error code
   set -x   # echo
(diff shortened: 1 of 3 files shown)


ProfFan:
Due to this sudden influx of SIGILLs, I suspect Travis is cutting down their cost by increasing the usage of VMs with older-gen processors (*unfounded speculation)

varunagrawal:
LGTM!

.travis.sh
```diff
@@ -44,11 +35,12 @@ function build ()
   prepare
 
   cmake $SOURCE_DIR \
-      -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE \
+      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE:-Debug} \
```
varunagrawal: My CMake is still a work in progress but I'm guessing this is akin to a default argument if `CMAKE_BUILD_TYPE` is not defined?

ProfFan: This is bash variable substitution

.travis.sh
```diff
@@ -1,34 +1,38 @@
...
-  if [ -z "$GTSAM_ALLOW_DEPRECATED_SINCE_V4" ]; then
-    GTSAM_ALLOW_DEPRECATED_SINCE_V4=OFF
-  fi
-
-  if [ ! -z "$GCC_VERSION" ]; then
-    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-$GCC_VERSION 60 \
-                         --slave /usr/bin/g++ g++ /usr/bin/g++-$GCC_VERSION
-    sudo update-alternatives --set gcc /usr/bin/gcc-$GCC_VERSION
-  fi
+	set -e   # Make sure any error makes the script to return an error code
```
ProfFan: indentation

.travis.sh
```diff
@@ -41,32 +45,23 @@ function finish ()
...
+  make -j2
 
   finish
 }
 
 # run the tests
 function test ()
 {
-  prepare
+  export GTSAM_BUILD_EXAMPLES_ALWAYS=Off
```
ProfFan: I think it is better to use `ON` and `OFF`?

.travis.yml
```diff
@@ -28,8 +30,15 @@ stages:
   - compile
   - test
 
+env:
+  global:
+    - MAKEFLAGS="-j2"
+    - CCACHE_SLOPPINESS=pch_defines,time_macros
+    -
```
ProfFan: Is this empty entry intended?

.travis.sh
jlblancoc: Oops! I've been playing these days switching forth and back between new IDEs, it seems in some of them I forgot to enable "show invisibles" Thanks! :+1:

jlblancoc:
Thanks for the feedback, @ProfFan . 
New `push --force` done, let's wait for Travis... once again. 

I have prepared a new build system based on CircleCI, will PR when this one is over. It's **way** faster than Travis and it's easy to use docker images with preloaded apt packages (saves even more time). The only downside is that they don't allow using macos builds in the free version... But Travis is really fast on MacOS so a good solution might be using both CI systems. 

From [tests on my fork](https://app.circleci.com/jobs/github/jlblancoc/gtsam/20), build times could cut from current ~ 1h40min to ~45min. 
For a future PR to work here, an admin should [login into CircleCI](https://circleci.com/dashboard) and enable this repo to work with CircleCI (Don't waste time following their wizard to install a config.yml, since it's already done in the fork). 
cc: @dellaert

PR #190, opened 2019-12-12
Fix GenericValue copy

ProfFan:
`GenericValue` correctly handles alignment when constructing objects but fails to do so when copying objects.

This is currently based on `feature/rotation_group`, but can also rebase on `develop` if that is preferred :)

CI passes for branch, let's hope that it pass for PR as well :smile:

Link to `GenericValue.h`: https://github.com/borglab/gtsam/pull/190/files#diff-12d727d01bcb063ba1996d3d58ed26c4

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/190)

### DIFF
--- a/gtsam/base/GenericValue.h
+++ b/gtsam/base/GenericValue.h
@@ -89,12 +89,9 @@ public:
 
     /**
      * Create a duplicate object returned as a pointer to the generic Value interface.
-     * For the sake of performance, this function use singleton pool allocator instead of the normal heap allocator.
-     * The result must be deleted with Value::deallocate_, not with the 'delete' operator.
      */
     virtual Value* clone_() const {
-      void *place = boost::singleton_pool<PoolTag, sizeof(GenericValue)>::malloc();
-      GenericValue* ptr = new (place) GenericValue(*this); // calls copy constructor to fill in
+      GenericValue* ptr = new GenericValue(*this); // calls copy constructor to fill in
       return ptr;
     }
 
@@ -102,8 +99,7 @@ public:
      * Destroy and deallocate this object, only if it was originally allocated using clone_().
      */
     virtual void deallocate_() const {
-      this->~GenericValue(); // Virtual destructor cleans up the derived object
-      boost::singleton_pool<PoolTag, sizeof(GenericValue)>::free((void*) this); // Release memory from pool
+      delete this;
     }
 
     /**
@@ -118,10 +114,7 @@ public:
       // Call retract on the derived class using the retract trait function
       const T retractResult = traits<T>::Retract(GenericValue<T>::value(), delta);
 
-      // Create a Value pointer copy of the result
-      void* resultAsValuePlace =
-          boost::singleton_pool<PoolTag, sizeof(GenericValue)>::malloc();
-      Value* resultAsValue = new (resultAsValuePlace) GenericValue(retractResult);
+      Value* resultAsValue = new GenericValue(retractResult);
 
       // Return the pointer to the Value base class
       return resultAsValue;
@@ -171,12 +164,6 @@ public:
       value_ = rhs.value_;
       return *this;
     }
-
-  private:
-
-    /// Fake Tag struct for singleton pool allocator. In fact, it is never used!
-    struct PoolTag {
-    };
 
   private:
 


gtsam/base/GenericValue.h
```diff
@@ -89,21 +89,17 @@ class GenericValue: public Value {
 
     /**
      * Create a duplicate object returned as a pointer to the generic Value interface.
-     * For the sake of performance, this function use singleton pool allocator instead of the normal heap allocator.
-     * The result must be deleted with Value::deallocate_, not with the 'delete' operator.
      */
     virtual Value* clone_() const {
-      void *place = boost::singleton_pool<PoolTag, sizeof(GenericValue)>::malloc();
-      GenericValue* ptr = new (place) GenericValue(*this); // calls copy constructor to fill in
+      GenericValue* ptr = new GenericValue(*this); // calls copy constructor to fill in
```
jlblancoc: Perhaps there is a way to expand `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` to cope with this additional overload of `new`, but it would probably be a mess and not worth the alleged performance gain of this `singleton_pool`... 
Unless you profiled that the singleton_pool is **really** worth in some application (?).

ProfFan: This is a placement new so the object does not have any control over alignment... also the pool only have `malloc` so no guarantees on alignment whatsoever. I can do a manual alignment here but I don't think this is worth it. And there is no proof of the alleged "performance gain" either :)

chrisbeall: > there is no proof of the alleged "performance gain" either :)

Not true. One of Frank's former students profiled the pool allocator when it was first introduced, and it made a significant difference. Please profile before you remove.

jlblancoc: I meant "performance gain" due to lack of continuous malloc/free vs. using the pool. It all depends on how many times you need to call this, I guess...

> This is a placement new so the object does not have any control over alignment...

Oh! Then the alternative would be checking if singleton_pool<> allows for aligned memory blocks, not changing the Eigen macros, you're right.

ProfFan: > Oh! Then the alternative would be checking if singleton_pool<> allows for aligned memory blocks, not changing the Eigen macros, you're right.

I have investigated this before. It does not support aligned allocation in itself, so we need to manually allocate `sizeof(object)+alignment` of memory and shift the pointer a bit to the aligned position.

> Not true. One of Frank's former students profiled the pool allocator when it was first introduced, and it made a significant difference. Please profile before you remove.

You are right that using an arena allocator makes performance gains in some workloads. However, I think the proper way of doing this would be using stuff like `jemalloc` which have been field proven and have proper `aligned_alloc()` support. `boost::singleton_pool` is just another arena allocator, but it is not widely used (google only have 3000 results).

dellaert:
@ProfFan it looks like you are merging in the entire rotation group branch? I don't think that's the intention, right? Please close this PR and make a *small* PR on develop, and add profiling results as @chrisbeall asked?

ProfFan:
## `timeLago`, trials = 10:

With Patch:
```
-Total: 0 CPU (0 times, 0 wall, 17.83 children, min: 0 max: 0)
|   -lago: 10.28 CPU (10 times, 4.96581 wall, 10.26 children, min: 1.01 max: 1.05)
|   |   -init: 6.44 CPU (10 times, 3.34908 wall, 6.44 children, min: 0.63 max: 0.66)
|   |   -refine: 3.82 CPU (10 times, 1.59171 wall, 3.82 children, min: 0.37 max: 0.4)
|   -optimize: 7.57 CPU (10 times, 3.01666 wall, 7.57 children, min: 0.74 max: 0.77)
fan@arx7:pts/2->/home/fan/Projects/CV/SLAM/gtsam_build (0)
```

Before (`develop`):
```
-Total: 0 CPU (0 times, 0 wall, 17.84 children, min: 0 max: 0)
|   -lago: 10.35 CPU (10 times, 4.98973 wall, 10.31 children, min: 1.03 max: 1.06)
|   |   -init: 6.49 CPU (10 times, 3.38008 wall, 6.49 children, min: 0.64 max: 0.66)
|   |   -refine: 3.82 CPU (10 times, 1.58463 wall, 3.82 children, min: 0.37 max: 0.39)
|   -optimize: 7.53 CPU (10 times, 2.99391 wall, 7.53 children, min: 0.74 max: 0.76)
```

## `timeOneCameraExpression`

After: `timing:0.345 musecs/call`

Before: `timing:0.346 musecs/call`

## `timeSFMBAL`

After:
```
-Total: 0 CPU (0 times, 0 wall, 7.46 children, min: 0 max: 0)
|   -optimize: 7.46 CPU (1 times, 3.25498 wall, 7.46 children, min: 7.46 max: 7.46)
```

Before:
```
-Total: 0 CPU (0 times, 0 wall, 7.49 children, min: 0 max: 0)
|   -optimize: 7.49 CPU (1 times, 3.28958 wall, 7.49 children, min: 7.49 max: 7.49)
```

ProfFan:
@chrisbeall Could you point out a specific timing case where a significant performance difference can be observed?

ProfFan:
@dellaert In my benchmarks using the `timing` scripts I cannot observe a major change in timing... Do we have other benchmarks involving `GenericValue` that I missed?

chrisbeall:
Sorry, I've been super busy. While at Tech, I used to run somewhat larger benchmark problems than these listed here. Unfortunately my benchmark lived outside of gtsam, and it would take me some time to resurrect this. The smallest took about 30s total, and some were much larger. The config scenarios I would cover:
- with/without TBB 
- Linux vs. Mac

 At the time we found that choice of allocator has different impact on different platforms. Iirc, standard malloc was terrible on Linux, but ok on Mac. Also, std allocator had terrible concurrent performance on Linux. I assume you ran your benchmark on Linux, so your result seems ok.

I think it would be worthwhile for someone to add a proper benchmark suite to gtsam

I think it's ok to merge. I'll run a benchmark the next time I pull upstream into our repo at work, and I'll report back if I see a performance issue there.

dellaert:
Thanks Chris, appreciate the feedback. I will merge.

ProfFan:
Just as a side note, I think the result is on par because that the allocator in `glibc` has improved a lot through the years: `glibc` 2.26 introduced a thread-local cache (similar to jemalloc and tcmalloc)

Ref: https://www.phoronix.com/scan.php?page=news_item&px=glibc-malloc-thread-cache

PR #192, opened 2019-12-13
Test using AdjointMap for transforming Pose2 and Pose3 covariance matrices from body to world frames

ptrmu:
1. Pose2 tests are created with simple covariance matrices and transformations. These tests produce covariances that can be validated with simple computations.
2. The first set of Pose3 tests are created by crafting scenarios where the 3d transformation is simply a 2d transformation about a primary axis. These tests are validated by comparing the results with Pose2 results.
3. The second set of Pose3 tests are created with relatively simple covariance matrices and transformations. These tests can be validated with simple computations.

Questions:
1. Supporting templated functions are shared between the Pose2 and Pose3 tests. There is potential for sharing additional code but I could not figure out how to instantiate Rot2 or Rot3 from an array of angles in a way that could be shared in a templated function.
2. The code for these shared functions is replicated in the two test files. What is the recommended file name and location for creating a new file to contain this shared test code?

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/192)

### DIFF
--- a/gtsam/geometry/tests/testPose2.cpp
+++ b/gtsam/geometry/tests/testPose2.cpp
@@ -894,6 +894,8 @@ namespace transform_covariance3 {
   }
 }
 TEST(Pose2 , TransformCovariance3) {
+  // Use simple covariance matrices and transforms to create tests that can be
+  // validated with simple computations.
   using namespace transform_covariance3;
 
   // rotate
--- a/gtsam/geometry/tests/testPose2.cpp
+++ b/gtsam/geometry/tests/testPose2.cpp
@@ -845,71 +845,68 @@ TEST(Pose2 , TransformCovariance3) {
 
   // rotate
   {
-    auto cov = GenerateFullCovariance<Pose2>({{0.1, 0.3, 0.7}});
-    auto cov_trans = RotateTranslate<Pose2>({{90.}}, {{}}, cov);
+    auto cov = FullCovarianceFromSigmas<Pose2>({0.1, 0.3, 0.7});
+    auto transformed = Pose2::TransformCovariance{{0., 0., 90 * degree}}(cov);
     // interchange x and y axes
-    EXPECT_DOUBLES_EQUAL(cov_trans(1, 1), cov(0, 0), 1e-9);
-    EXPECT_DOUBLES_EQUAL(cov_trans(0, 0), cov(1, 1), 1e-9);
-    EXPECT_DOUBLES_EQUAL(cov_trans(2, 2), cov(2, 2), 1e-9);
-
-    EXPECT_DOUBLES_EQUAL(cov_trans(1, 0), -cov(1, 0), 1e-9);
-    EXPECT_DOUBLES_EQUAL(cov_trans(2, 0), -cov(2, 1), 1e-9);
-    EXPECT_DOUBLES_EQUAL(cov_trans(2, 1), cov(2, 0), 1e-9);
+    EXPECT(assert_equal(
+      Vector3{cov(1, 1), cov(0, 0), cov(2, 2)},
+      Vector3{transformed.diagonal()}));
+    EXPECT(assert_equal(
+      Vector3{-cov(1, 0), -cov(2, 1), cov(2, 0)},
+      Vector3{transformed(1, 0), transformed(2, 0), transformed(2, 1)}));
   }
 
   // translate along x with uncertainty in x
   {
-    auto cov = GenerateOneVariableCovariance<Pose2>(0, 0.1);
-    auto cov_trans = RotateTranslate<Pose2>({{}}, {{20., 0.}}, cov);
-    EXPECT_DOUBLES_EQUAL(cov_trans(0, 0), 0.1 * 0.1, 1e-9);
-    EXPECT_DOUBLES_EQUAL(cov_trans(1, 1), 0., 1e-9);
-    EXPECT_DOUBLES_EQUAL(cov_trans(2, 2), 0., 1e-9);
-
-    EXPECT_DOUBLES_EQUAL(cov_trans(1, 0), 0., 1e-9);
-    EXPECT_DOUBLES_EQUAL(cov_trans(2, 0), 0., 1e-9);
-    EXPECT_DOUBLES_EQUAL(cov_trans(2, 1), 0., 1e-9);
+    auto cov = SingleVariableCovarianceFromSigma<Pose2>(0, 0.1);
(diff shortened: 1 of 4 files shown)


gtsam/geometry/tests/testPose2.cpp
```diff
@@ -835,6 +835,139 @@ TEST(Pose2 , ChartDerivatives) {
   CHECK_CHART_DERIVATIVES(T2,T1);
 }
 
+//******************************************************************************
+namespace transform_covariance3 {
+  const double degree = M_PI / 180;
+
+  template<class TPose>
+  static typename TPose::Jacobian generate_full_covariance(
```
dellaert: naming, pls use Google style guide: `GenerateFullCovariance` (capitalized since static). Also, add at least a one-line comment? Goes for all functions below...

gtsam/geometry/tests/testPose2.cpp
```diff
@@ -835,6 +835,139 @@ TEST(Pose2 , ChartDerivatives) {
...
+    std::array<double, TPose::dimension> sigma_values)
```
dellaert: const & ?

gtsam/geometry/tests/testPose2.cpp
ptrmu: Changed names and added function comments.

gtsam/geometry/tests/testPose2.cpp
ptrmu: Good! Thanks.

ptrmu:
Think the comments are addressed. Happy to work on other issues.

gtsam/geometry/tests/testPose2.cpp
```diff
@@ -835,6 +835,84 @@ TEST(Pose2 , ChartDerivatives) {
...
+
+TEST(Pose2 , TransformCovariance3) {
+  // Use simple covariance matrices and transforms to create tests that can be
+  // validated with simple computations.
+  using namespace test_pose_adjoint_map;
+
+  // rotate
+  {
+    auto cov = GenerateFullCovariance<Pose2>({{0.1, 0.3, 0.7}});
+    auto cov_trans = RotateTranslate<Pose2>({{90.}}, {{}}, cov);
```
dellaert: I think we are now ready to move this to Lie.h :-) Might I suggest this should read
```
auto transformed = TransformCovariance(Pose2(0,0,90*degree))(cov);
```
where `TransformCovariance` is a more generic version of your RotateTranslate that works for any Lie group, and is defined as
```
template <class T>
class TransformCovariance {
private:
T::Jacobian adjointMap_;
public:
TransformCovariance(const T& X):adjointMap_(X.AdjointMap()) {}
operator ()(const T::Jacobian& covariance) {return adjointMap_ * covariance * adjointMap_.transpose();}
};
```
The "functor" form above makes it possible to transform many covariances without always having to re-compute the Adjoint map. In addition, we might want to consider templating on the Eigen matrix type so it works efficiently, say, with DiagonalMatrices.

gtsam/geometry/tests/testPoseAdjointMap.h
```diff
@@ -0,0 +1,87 @@
...
+  {
+    typename T::Jacobian cov = T::Jacobian::Zero();
+    cov(idx0, idx0) = sigma0 * sigma0;
+    cov(idx1, idx1) = sigma1 * sigma1;
+    cov(idx0, idx1) = cov(idx1, idx0) = sigma0 * sigma1;
+    return cov;
+  }
+
+  // Overloaded function to create a Rot2 from one angle.
+  Rot2 RotFromArray(const std::array<double, Rot2::dimension> &r)
```
dellaert: I don't see much need for these over the default Pose2 and Pose3 constructors, esp. when we switch to the proposed functor above.

ptrmu:
I'd be happy to keep working on this. I would like to understand more about Lie groups in general and in particular about your techniques for coding them. Learning more about Eigen would also be good. I've avoided Eigen because of the painful compile times and debugging opaqueness. Let me review your comments and see if I can implement some of your ideas.

dellaert:
Excellent. Note my prototype code might not compile as is... A nice discussion on functors is here: https://stackoverflow.com/questions/356950/what-are-c-functors-and-their-uses

dellaert:
Wow, this is starting to look pretty cool. One change still requested: you put the functor inside the LieGroup class, but that is actually a convenience class and is not mandatory to create classes that satisfy the Lie group concept. Hence, the functor should really be its own, free functor templated on `T` rather than class. Documentation should reflect that T needs to satisfy the Lie group concept.

ptrmu:
Fun.

Do additional tests need to be created for other subclasses of LieGroup?

Are there special techniques for transforming vector covariances? For example the covariance of a Vector3 when transformed by a Pose3 or the covariance of a Vector2 when transformed by a Pose2?

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -878,6 +878,105 @@ TEST(Pose3 , ChartDerivatives) {
...
+      Vector3{cov3(spatial_axis0, spatial_axis0), cov3(spatial_axis1, spatial_axis1), cov3(r_axis, r_axis)}));
+    EXPECT(assert_equal(
+      Vector3{cov2(1, 0), cov2(2, 0), cov2(2, 1)},
+      Vector3{cov3(spatial_axis1, spatial_axis0), cov3(r_axis, spatial_axis0), cov3(r_axis, spatial_axis1)}));
+  };
+
+  // rotate around x axis
+  {
+    auto cov3 = FullCovarianceFromSigmas<Pose3>((Vector6{} << s2(2), 0., 0., 0., s2(0), s2(1)).finished());
+    auto transformed3 = Pose3::TransformCovariance{{Rot3::RzRyRx(p2.theta(), 0., 0.), {0., p2.x(), p2.y()}}}(cov3);
```
ptrmu: Here the argument to FullCovarianceFromSigmas<Pose3>() needs to be constructed with a Vector6 comma initializer. Yet FullCovarianceFromSigmas<Pose2>() can use 3 comma separated doubles. It would be nice to use 6 comma separated doubles in this case as well.

dellaert: Not super-critical.

dellaert:
I love what you did here ! I think having Pose2 and Pose3 tests is sufficient, as they are the most common use cases and it proves the new functor maps over different types. 
I will merge, thanks a bunch for your contribution! 
Maybe a post in the GTSAM user group with an example is useful to disseminate this new functionality and show off what you did :-)

PR #195, opened 2019-12-19
Remove ugly preprocessor directives

ToniRV:
Using combined imu preintegration is still hardcoded though.
If there is any interest in not having preprocessor directives, I'll go ahead and make it a parameter to avoid recompilations

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/195)

### DIFF
--- a/examples/ImuFactorsExample.cpp
+++ b/examples/ImuFactorsExample.cpp
@@ -17,8 +17,8 @@
 
 /**
  * Example of use of the imuFactors (imuFactor and combinedImuFactor) in conjunction with GPS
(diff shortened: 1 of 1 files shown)


dellaert:
Nice. A command-line parameter (with the same default as before) would be nice. It will also trigger a travis re-build which seems to have stuck on an altogether different file. We've had travis issues before after recent changes they made.

ToniRV:
@dellaert Done, unless tests fail and/or comments.

@andretag you could play with this example and compare the errors between ImuFactor and CombinedImuFactor.

ProfFan:
@ToniRV  Can you rebase to the latest `develop`? This would probably fix the CI error.

PR #198, opened 2019-12-20
CMake include improvements for macOS

varunagrawal:
- Install gtsam_unstable correctly on APPLE devices.
- ~~Add `CMAKE_INSTALL_PREFIX` for generator expressions of different 3rd party libraries bundled with GTSAM.~~

This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/198)


dellaert:
Interesting. Why is this an issue only on Macs??

varunagrawal:
CMake needs a specific property for build targets on macs which is `INSTALL_NAME_DIR`.
Here's the doc page https://cmake.org/cmake/help/latest/prop_tgt/INSTALL_NAME_DIR.html

varunagrawal:
Funnily enough, this is done for `gtsam` but not for `gtsam_unstable` which would explain why you're having issues on macs.

dellaert:
Ah. Should we use “ This property is initialized by the value of the variable CMAKE_INSTALL_NAME_DIR if it is set when a target is created.” ?

varunagrawal:
That would be better, it would get rid of a bunch of boilerplate.

varunagrawal:
Actually, I'd like to have @jlblancoc's thoughts as well. It seems like this property is primarily used for building shared libraries on Macs. A shared library is the default option when building GTSAM but I'm not sure how the build would work if a user sets the option to OFF.

dellaert:
It definitely is an issue on macs... that recent commit may have caused it to go wrong...

jlblancoc:
@lsolanka as author of #169: what exact problem was that PR trying to solve?
We are pondering unrolling it...

varunagrawal:
That makes sense. I'll undo the last commit.

varunagrawal:
@dellaert I would need a quick test on a mac machine to truly know if this works.

dellaert:
I just tried it and it will not work. I did `cmake ..` and then `ccmake..` to change the prefix, but I see that `CMAKE_INSTALL_NAME_DIR` is unchanged as `/usr/local/lib`:
```
CMAKE_INSTALL_NAME_DIR           /usr/local/lib                                                                                                                           
 CMAKE_INSTALL_PREFIX             /Users/dellaert                                                                                                                          
```
so that's an issue.

dellaert:
Other than the above issue, I was able to link successfully now in an external project

ProfFan:
I am working on mac so I can test. `install_name_dir` is related to the `rpath` of an macOS executable. What is the exact problem here?

dellaert:
@ProfFan The issue is that when linking with gtsam in an external project, gtsam_unstablewas not linked properly if gtsam is installed with a custom prefix. At the moment @varunagrawal 's fix does not work as is as when using ccmake to change the prefix the change is not reflected.

varunagrawal:
Interesting. It seems that ccmake does not respect the variable dependencies.
If I run `cmake -DCMAKE_INSTALL_PREFIX=/home/varun/projects/borglab/gtsam/install ..` (with the appropriate changes), all the variables are set correctly. My belief is that `cmake ..` sets the variables correctly, but ccmake does not run the same assignment logic. We need something like a "set reference".

varunagrawal:
Figured it out. We need to **not**  add `CMAKE_INSTALL_NAME_DIR` to the cmake cache. It works fine then. I'll update the code.

varunagrawal:
@dellaert @ProfFan can you please check this on mac devices?

ProfFan:
@varunagrawal Yes that is the default behavior (and the correct one) for CMake.

dellaert:
@varunagrawal it works flawlessly on my mac!

dellaert:
@jlblancoc does this really undo anything in #169, it seems to only change the definition of the variable not its usage? 
Also, when looking at that PR, I see the changes are only in gtsam/CMakeLists.txt, but I see many more mentions of `CMAKE_INSTALL_PREFIX` all over the project. Was #169 actually thorough or did it only solve a tiny part of the problem? I'm not the cmake expert, but it seems we might be in some weird half-baked state now?

jlblancoc:
> @jlblancoc does this really undo anything in #169, it seems to only change the definition of the variable not its usage?

@dellaert : The PR *unrolled* the changes, but a latter commit re-added it, apparently :-)

> but I see many more mentions of CMAKE_INSTALL_PREFIX all over the project. 
> ...
> Was #169 actually thorough or did it only solve a tiny part of the problem? 

You are right! I have found and removed a couple of additional usages of that variable in "INSTALL" generator expressions. 
I think #169 was complete, though, since gtsam_unstable didn't need any further `target_include_directories()` command (which is where the cmake generator expressions are mostly used), since it inherits the include directories from `gtsam`. 

Please, give this a try in Mac to be sure it still works.

dellaert:
@jlblancoc still works.

lsolanka:
> @lsolanka as author of #169: what exact problem was that PR trying to solve?
> We are pondering unrolling it...

Sorry for a delay...It was a fix to enable the whole installation to be movable into another folder. If `CMAKE_INSTALL_PREFIX` is used then the paths to the shared objects, etc. are hard-coded and the whole installation folder cannot be moved. It seems to be a cleaner way to use the `INSTALL_INTERFACE` but also it is relevant for package managers like https://github.com/cpp-pm/hunter which cache the build products into a tarball and then extract them potentially to different folders than the build originated from.

However if it breaks something on MacOS then it's not a big deal...

At the same time looking at this PR I'm not sure if I follow, it doesn't really undo #169, rather it removes even more occurrences of `CMAKE_INSTALL_PREFIX`?

jlblancoc:
@lsolanka : Thanks for the explanations! I guessed it would be related to package managers...

PS: It is normal you are confused about the question: the first commit did readd `CMAKE_INSTALL_PREFIX`, but latter it was removed again.

PR #215, opened 2020-01-21
Point3 from Unit3 in wrapper code

varunagrawal:
Simple update to allow users to get `Point3` from a `Unit3` in wrapped languages aka Python.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/215)

### DIFF
--- a/gtsam.h
+++ b/gtsam.h
@@ -814,6 +814,7 @@ class Unit3 {
   // Other functionality
   Matrix basis() const;
   Matrix skew() const;
+  gtsam::Point3 point3() const;
 
   // Manifold
   static size_t Dim();


gtsam.h
```diff
@@ -814,6 +814,7 @@ class Unit3 {
   // Other functionality
   Matrix basis() const;
   Matrix skew() const;
+  gtsam::Point3 point3() const;
```
ProfFan: as_point?

varunagrawal: Wouldn't that break backwards compatibility?

dellaert: @ProfFan gtsam.h is just a wrapper. Maybe the original function is badly named, but that's orthogonal to this PR. And, indeed, would be an API change.

dellaert:
@ProfFan if you agree, please approve, merge, and delete branch.

PR #219, opened 2020-02-05
Minor updates to wrapper

varunagrawal:
Replace Cal3_S2 deprecated matrix() with K(), add Cal3DS2 support to triangulation.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/219)

### DIFF
--- a/gtsam.h
+++ b/gtsam.h
@@ -875,6 +875,7 @@ class Cal3_S2 {
   gtsam::Point2 principalPoint() const;
   Vector vector() const;
   Matrix K() const;
+  Matrix matrix() const;
   Matrix matrix_inverse() const;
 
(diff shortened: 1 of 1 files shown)


gtsam.h
```diff
@@ -874,7 +874,7 @@ class Cal3_S2 {
   double py() const;
   gtsam::Point2 principalPoint() const;
   Vector vector() const;
-  Matrix matrix() const;
+  Matrix K() const;
```
dellaert: Deprecated but can’t break existing python code in GTSAM 4. Feel free to add K, but leave matrix?

varunagrawal: All done.

PR #223, opened 2020-02-11
Add two required string for version detection

ProfFan:
For pip to detect proper package for python 2.

@dellaert

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/223)


dellaert:
So, I do *not* see any wrapper CI. where should I look?

ProfFan:
@dellaert These two changes are necessary for pushing Python 2 packages to PyPI.

Issue #204, opened 2020-01-02
Modern wrapper plan


dellaert:
Desired end-state: python scripts generate both python and MATLAB wrappers. Because there is an API change in python, this will bump our version to 4.1.

- [x] PR 0 to package *existing* cython wrapper for manylinux, README, so loads from colab (this is done in a separate repo, https://github.com/borglab/gtsam-manylinux-build)
- [x] Add pybind to feature/new_wrapper
- [x] Add pybind wrapper version (purely additive, duplicates all python scripts/tests into gtsam_py, includes GTSAM changes), ~fix Values~ (done), PR to feature/new_wrapper
- [x] Add travis CI for cython wrapper, PR to develop
- [x] Add travis CI for new wrapper, PR to feature/new_wrapper
- [x] Do PR to develop with new wrapper and we ask everyone to test/switch
- [x] @ProfFan tests MATLAB wrapper 
- [x] Switch MATLAB wrapper to be created by python, includes Matthew unit tests, PR to feature/new_wrapper
- [x] Fix most Bugs before 4.1
- [x] Remove cython wrapper and old c++ wrap, PR to develop (only file deletions)
- [ ] Create release branch and do PR to master, add release tag
- [ ] Create packaging in gtsam-manylinux-build for 4.1 so we can pip install, think about versioning

ProfFan:
@dellaert Now gtsam works in Colab with some patches :)

Demo: https://colab.research.google.com/drive/1j2HgxVUDUfQEseLmke4Tk_LKiyJ87QjQ

One blocker: pypi got a file size limit of 60MB, but GTSAM packaged is 272M... I filed a request for increase the quota, in the meantime can load from Google Drive or an URL instead.

ProfFan:
Code to build the wheels:

https://github.com/ProfFan/gtsam-manylinux-build

ProfFan:
@dellaert Now GTSAM lives on PyPI with a size of 7.3MB :)

I changed the Colab notebook to just use `pip install gtsam` :smile:

dellaert:
This is *AMAZING*

dellaert:
How did you get it down to 7.3MB??

ProfFan:
I stripped out the debug symbols and used a custom boost build instead of the system provided package...

ProfFan:
@dellaert Note that macOS build is complete and tested (Python 3.7).

dellaert:
[...]

PR #224, opened 2020-02-11
Add Python CI

ProfFan:
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/224)

### DIFF
--- a/.travis.python.sh
+++ b/.travis.python.sh
@@ -12,7 +12,7 @@ if [[ $(uname) == "Darwin" ]]; then
     brew install wget
 else
     # Install a system package required by our library
-    sudo apt-get install wget libicu-dev python3-pip
+    sudo apt-get install wget libicu-dev python3-pip python3-setuptools
 fi
 
 CURRDIR=$(pwd)
--- a/.travis.python.sh
+++ b/.travis.python.sh
@@ -1,8 +1,12 @@
 #!/bin/bash
 set -x -e
 
-# Install a system package required by our library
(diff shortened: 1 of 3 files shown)


dellaert:
python CI seems to fail. You might want to temporarily suspend all other CI in this PR so you can iterate more quickly, then re-enable when everything works.

ProfFan:
@dellaert 
image

dellaert:
So, I’m lost. is this building and doing CI on *bothu cython and python wrappers? 

Also: 
1) python 2.7 EOL warnings. Are we checking 2.7 rather than python 3?
2) these warnings have been going in on my Mac as well:
CMake Warning at /usr/local/lib/cmake/boost_timer-1.71.0/libboost_timer-variant-shared.cmake:59 (message):
922  Target Boost::timer already has an imported location

ProfFan:
@dellaert 
1. I have changed the tests to build & test against Python 3.
2. I think these warnings are not introduced by the Python CI?

ProfFan:
On the 2nd issue, see: https://stackoverflow.com/questions/58081084/target-boostlibrary-already-has-an-imported-location-link-errors

ProfFan:
Ping @dellaert

dellaert:
I'm still lost :-) is this building and doing CI on *both* cython and python wrappers? I don't see the answer in comments above.

ProfFan:
This is the CI for the old wrapper, as we have discussed in #204 :)

PR #226, opened 2020-02-21
Printing updates

varunagrawal:
Some niceties w.r.t. printing of matrices and poses.

Separating row values with commas is supported by MATLAB so no effect there,
but this also helps when printing with Python and C++ since the output can be
directly copied without modification.

As for `Pose3`, we simply use the `ostream` support from `Point3` directly so that we are DRY.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/226)

### DIFF
--- a/gtsam/base/Matrix.cpp
+++ b/gtsam/base/Matrix.cpp
@@ -142,7 +142,7 @@ void print(const Matrix& A, const string &s, ostream& stream) {
   static const Eigen::IOFormat matlab(
       Eigen::StreamPrecision, // precision
       0, // flags
-      " ", // coeffSeparator
+      ", ", // coeffSeparator
       ";\n", // rowSeparator
       " \t",  // rowPrefix
       "", // rowSuffix
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -106,7 +106,7 @@ Vector6 Pose3::adjointTranspose(const Vector6& xi, const Vector6& y,
 void Pose3::print(const string& s) const {
   cout << s;
   R_.print("R:\n");
-  cout << '[' << t_.x() << ", " << t_.y() << ", " << t_.z() << "]\';";
+  cout << t_ << ";" << endl;
 }
 
 /* ************************************************************************* */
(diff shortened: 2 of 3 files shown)


gtsam/base/Matrix.cpp
```diff
@@ -142,7 +142,7 @@ void print(const Matrix& A, const string &s, ostream& stream) {
   static const Eigen::IOFormat matlab(
       Eigen::StreamPrecision, // precision
       0, // flags
-      " ", // coeffSeparator
+      ", ", // coeffSeparator
```
dellaert: Why? MATLAB does not need it (or do this, I think).

gtsam/geometry/Pose3.cpp
```diff
@@ -106,7 +106,7 @@ Vector6 Pose3::adjointTranspose(const Vector6& xi, const Vector6& y,
 void Pose3::print(const string& s) const {
   cout << s;
   R_.print("R:\n");
-  cout << '[' << t_.x() << ", " << t_.y() << ", " << t_.z() << "]\';";
+  cout << t_ << ";" << endl;
```
dellaert: Why, what's the reason?

gtsam/base/Matrix.cpp
varunagrawal: Whoops. I put the explanation in the commit message, should have put it here as well.

Separating row values with commas is supported by MATLAB so no effect there,
but this also helps when printing with Python and C++ since the output can be
directly copied without modification.

gtsam/geometry/Pose3.cpp
varunagrawal: Point3 already support ostream so it didn't make much sense to have this duplication of code. I like keep it DRY.

dellaert: Meaning, output is exactly the same before and after?

varunagrawal: Yes. Verifiably.

varunagrawal: Good news is, if we change the printing of Point3 in the future, this will be picked up py Pose3 automatically now.

dellaert:
Oops, pressed wrong button. Feel free to merge yourself after travis builds again :-/

varunagrawal:
@dellaert added a unit test to verify that `Pose3` output for translation is the same as before. Will merge once travis passes.

PR #227, opened 2020-02-21
Wrap SfM_data struct

varunagrawal:
Wrap the `SfM_data` struct so we can use it in other projects and properly unit test.

We also wrap `SfM_Track` which is a struct to hold tracks of measurements from a single landmark across various images/poses. Each measurement has an associated `SIFT_index`, and the track holds the 3D landmark point and its RGB information.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/227)

### DIFF
--- a/gtsam.h
+++ b/gtsam.h
@@ -2724,6 +2724,11 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 };
 
 #include <gtsam/slam/dataset.h>
+class SfM_data {
+  size_t number_cameras() const;
+  size_t number_tracks() const;
+};
+
 string findExampleDataFile(string name);
 pair<gtsam::NonlinearFactorGraph*, gtsam::Values*> load2D(string filename,
     gtsam::noiseModel::Diagonal* model, int maxID, bool addNoise, bool smart);
--- a/gtsam.h
+++ b/gtsam.h
@@ -2724,9 +2724,18 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 };
 
 #include <gtsam/slam/dataset.h>
+class SfM_Track {
+  size_t number_measurements() const;
+  pair<size_t, gtsam::Point2> measurement(size_t idx) const;
+  pair<size_t, size_t> SIFT_index(size_t idx) const;
+};
+
 class SfM_data {
   size_t number_cameras() const;
   size_t number_tracks() const;
+  //TODO(Varun) Need to fix issue #237 first before this can work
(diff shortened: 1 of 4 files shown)


gtsam.h
```diff
@@ -2723,6 +2723,11 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 };
 
 #include <gtsam/slam/dataset.h>
+class SfM_data {
```
dellaert: Seems rather limited usefulness if you can't actually get any data from it...

varunagrawal: Yeah I was wondering about that but I just needed something quick to get my A->B of testing for the projections.

dellaert: Yes, but that means it will need to be added before we can merge this PR

dellaert: Yes, but that means it will need to be added before we can merge this PR

varunagrawal:
I want to update the name to `SfM_Data` so that it is consistent with `SfM_Track` and `SfM_Measurement`, as well as wrap more things before moving forward with this PR, so changing it to `WIP`.

varunagrawal:
I added some additional methods and wrapped them, so that this feature is more useful.

dellaert:
What is SfM_Track? Where does it come from? Give more context in your PR comment.

varunagrawal:
Updated the description. `SfM_Track` is a struct to encapsulate measurement tracks and is used by `SfM_data`. It makes sense to wrap `SfM_Track` as well since it has useful data.

gtsam/slam/dataset.h
```diff
@@ -173,14 +173,20 @@ typedef std::pair<size_t, size_t> SIFT_Index;
...
-  SfM_Track():p(0,0,0) {}
+  SfM_Track(): p(0,0,0) {}
   Point3 p; ///< 3D position of the point
   float r, g, b; ///< RGB color of the 3D point
   std::vector<SfM_Measurement> measurements; ///< The 2D image projections (id,(u,v))
   std::vector<SIFT_Index> siftIndices;
   size_t number_measurements() const {
     return measurements.size();
   }
+  SfM_Measurement measurement(size_t idx) const {
```
dellaert: docstrings needed. Also add them for constructors while you're at it. Can be
```
/// Construct empty track
```

I must say I don't understand the point of `point` anymore. I seems that can only work in an imperative context :-(

varunagrawal:
The new naming convention (including in the wrapper) will happen in #239. I wanted to keep the changes separated for a clear history in git.

dellaert:
OK, if both are merged at the same time

PR #228, opened 2020-02-21
Added override everywhere where it was needed

dellaert:
Fixed warnings on clang: added override in GaussianFactor derived classes.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/228)


dellaert:
@ProfFan @yetongumich should be easy review?

ProfFan:
LGTM

PR #231, opened 2020-02-26
Fix/dimension

dellaert:
Fixed broken GetDimension for Lie groups with variable dimension.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/231)


ProfFan:
Ah, I see the same issue again (for the Python CI)... Travis just timed out for that platform...

PR #233, opened 2020-02-29
Feature/random std

dellaert:
Switched from boost random sampling to std <random>, since it is now in c++11.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/233)

### DIFF
--- a/gtsam/base/WeightedSampler.h
+++ b/gtsam/base/WeightedSampler.h
@@ -1,0 +1,116 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file    WeightedSampler.h
+ * @brief   Fast sampling without replacement.
+ * @author  Frank Dellaert
+ * @date    May 2019
+ **/
+
+#include <queue>
+#include <random>
+#include <stdexcept>
+#include <utility>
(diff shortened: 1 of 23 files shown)


dellaert:
Thanks @jlblancoc . I don't have the time w ECCV to add such a unit test, and, as you say, it's a bit tricky with randomness, so I'll merge now. Someone should read up on how to *guarantee* such tests - must be a solved problem ;-)

varunagrawal:
Why wouldn't a set of predetermined random seeds work for testing this? The randomness is after all just pseudo-random, so having a set of say 5 random seeds should provide the kind of variability for robust tests we seek.

jlblancoc:
@varunagrawal it looks a good idea at first sight... but in practice the std c++11 random classes do NOT give the same sequence of pseudo random numbers for all platforms/ c++ library vendors/architectures, for a given seed.

varunagrawal:
I see. It makes sense that PRNGs vary according to platform. I'll do some investigating on Monday to see what the right way to handle this is.

peteroupc:
> 
> 
> @varunagrawal it looks a good idea at first sight... but in practice the std c++11 random classes do NOT give the same sequence of pseudo random numbers for all platforms/ c++ library vendors/architectures, for a given seed.

This is the case for the C++ _distributions_ (e.g., `uniform_int_distribution` and `uniform_real_distribution`), but not the C++ random engines (e.g., `discard_block_engine`, `mt19937`), which work the same way in all C++ library implementations.

jlblancoc:
Thanks for the clarification, @peteroupc ! :+1:

varunagrawal:
https://softwareengineering.stackexchange.com/questions/147134/how-should-i-test-randomness

Great Stackexchange answer on this issue. The go-to way seems to be following the Battery tests, which as @jlblancoc mentioned, is a bunch of statistical tests.

PR #239, opened 2020-03-06
Rename SfM_data to SfM_Data to be consistent

varunagrawal:
`SfM_data` does not follow a consistent naming scheme to `SfM_Track`, `SfM_Measurement`, etc.
This PR updates the name to enforce consistency aka a cleaner codebase.

To allow backwards compatibility, the older name is typedef'd.

This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/239)

### DIFF
--- a/examples/CreateSFMExampleData.cpp
+++ b/examples/CreateSFMExampleData.cpp
@@ -31,7 +31,7 @@ void createExampleBALFile(const string& filename, const vector<Point3>& P,
         Cal3Bundler()) {
 
   // Class that will gather all data
-  SfM_data data;
+  SfM_Data data;
 
   // Create two cameras
   Rot3 aRb = Rot3::Yaw(M_PI_2);
--- a/examples/CreateSFMExampleData.cpp
+++ b/examples/CreateSFMExampleData.cpp
@@ -31,19 +31,19 @@ void createExampleBALFile(const string& filename, const vector<Point3>& P,
         Cal3Bundler()) {
 
   // Class that will gather all data
-  SfM_Data data;
+  SfmData data;
 
   // Create two cameras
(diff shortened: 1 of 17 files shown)


varunagrawal:
This PR depends on #227. Let's get that in first before we can merge this.

dellaert:
1) You really need better PR explanations. Just the title is not enough.
2) We *cannot* change names/APIs without a version bump. You can deprecate the old name so all existing code out there still works, if that's what you want to do.

varunagrawal:
I thought the title was pretty self-explanatory. I'll update the PR description.

Also, on line 213 of `dataset.h` I have added this for backwards compatibility:

```cpp
/// Alias for backwards compatibility
typedef SfM_Data SfM_data;
```

This should prevent breakage of any kind. :slightly_smiling_face:

dellaert:
That's great, but you should bracket it with DEPRECATED etc :-)
Frank

dellaert:
And, since we are doing a name change, per @Alescontrela's comment: what *is* the Google style guide for classes? I think maybe this has to become SfmData, SfmTrack etc...

varunagrawal:
Yes. As per [this](https://google.github.io/styleguide/cppguide.html#Type_Names), the correct naming convention is `SfmData`, `SfmTrack`, etc.

Thank you @Alescontrela.

varunagrawal:
I'm not sure I understand what you mean by "bracket with DEPRECATED". Should it be wrapped in a `#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V4` or something else? I can't find anything else.

dellaert:
Yes, exactly. 
```
#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V4
typedef...
#endif
```

PR #243, opened 2020-03-06
Axis-Angle (in radians) representation for Rot3 matrices

varunagrawal:
New functions `pair<Unit3, double> Rot3::ToAxisAngle(Rot3)` and `pair<Unit3, double> axisAngle()` in `Rot3` to compute the angle-axis representation of a rotation matrix.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/243)


gtsam/geometry/tests/testRot3.cpp
```diff
@@ -661,6 +661,21 @@ TEST(Rot3, ClosestTo) {
   EXPECT(assert_equal(expected, actual.matrix(), 1e-6));
 }
 
+/* ************************************************************************* */
+TEST(Rot3, angle) {
```
dellaert: Hmmm. This is not a good test case. How do we know its correct? Simpler case?

gtsam/geometry/Rot3.h
```diff
@@ -479,6 +479,12 @@ namespace gtsam {
      */
     Rot3 slerp(double t, const Rot3& other) const;
 
+    /**
+     * @brief Compute the angle (in degrees) between *this and other
```
dellaert: We did in degrees in python but I propose as part of Rot3 it should be radians. Also, we might as well make this axis and angle? Return `pair<Unit3,double>` pair? Call it `axisAngle`?

PS, naming: GTSAM has consistently used camelCase for functions, not snake_case. We should stick to that knowing it departs from Google style guide.

varunagrawal: I thought I followed camel case here? It's just single word and it is an object method (in contrast to a static method). Am I missing something?

I'll update the signature.

gtsam/geometry/tests/testRot3.cpp
varunagrawal: That's a good point. I'll come up with a more contrived example which we know is correct.

gtsam/geometry/Rot3.h
dellaert: Nothing wrong :-) It was just an FYI, should be 'axisAngle' if we do both.

gtsam.h
```diff
@@ -677,6 +677,7 @@ class Rot3 {
 //  Vector toQuaternion() const;  // FIXME: Can't cast to Vector properly
   Vector quaternion() const;
   gtsam::Rot3 slerp(double t, const gtsam::Rot3& other) const;
+  double angle(const gtsam::Rot3& other) const;
```
dellaert: needs to be axisAngle now, with correct return type.

gtsam/geometry/Rot3.cpp
```diff
@@ -228,6 +228,12 @@ Rot3 Rot3::slerp(double t, const Rot3& other) const {
   return interpolate(*this, other, t);
 }
 
+/* ************************************************************************* */
+pair<Unit3, double> Rot3::axisAngle(const Rot3& other) const {
+  Vector3 rot = Rot3::Logmap(this->between(other));
```
dellaert: rot -> omega

gtsam/geometry/Rot3.cpp
```diff
@@ -228,6 +228,12 @@ Rot3 Rot3::slerp(double t, const Rot3& other) const {
   return interpolate(*this, other, t);
 }
 
+/* ************************************************************************* */
+pair<Unit3, double> Rot3::axisAngle(const Rot3& other) const {
```
dellaert: Actually, there should be no other :-(
I think this should be a Rot3 static method. We can always do
axisAngle =Rot3::AxisAngle(R1.between(R2));
Which presents a problem , as AxisAngle already exists as a "named constructor". Maybe Rot3::ToAxisAngle ?

gtsam/geometry/tests/testRot3.cpp
```diff
@@ -661,6 +661,21 @@ TEST(Rot3, ClosestTo) {
...
+  Rot3 R1 = Rot3::Ypr(0, 0, 0);
```
dellaert: this is identity. Which is good, as your test is almost good for static method. Just kill R1, rename R2 to R1. Or, better, just test angle for all rotations defined at top of this file...

gtsam/geometry/Rot3.cpp
varunagrawal: Would having both make sense?

`Rot3::ToAxisAngle` as well as `R1.axisAngle(R2)`?

dellaert: No, I don't think so.

gtsam/geometry/tests/testRot3.cpp
```diff
@@ -663,14 +664,12 @@ TEST(Rot3, ClosestTo) {
...
 /* ************************************************************************* */
 TEST(Rot3, angle) {
-  Rot3 R1 = Rot3::Ypr(0, 0, 0);
-  
-  Rot3 R2 = Rot3::Ypr(M_PI/4, 0, 0);
+  Rot3 R1 = Rot3::Ypr(M_PI/4, 0, 0);
 
-  Unit3 expectedAxis(0, 0, 1);
-  double expectedAngle = M_PI/4;
+  Unit3 expectedAxis(-0.350067, -0.472463, 0.808846);
```
dellaert: Again, not a great test. How about just adding three tests for each of yaw, pitch, roll? It’ll be interesting in itself as, thinking about it, there is a sign ambiguity that we probably need to resolve. For example, guarantee the angle is in `(-pi, pi]`

dellaert:
Update title, no longer in degrees

gtsam/geometry/tests/testRot3.cpp
varunagrawal: I'm confused. Am I still supposed to use the rotation defined at the top of the file?

If I understand correctly, a good set of tests would be those that check for roll, pitch and yaw individually using numbers that make it easy to verify manually that what is happening is indeed correct. Isn't this the same as my original test?
After that, using the rotation defined at the top would involve another, non-trivial test(?). I am not sure where the sign ambiguity fits in here. Wouldn't that need a specific rotation matrix that is obtuse and which would require `ToAxisAngle` to correctly return an acute angle?

dellaert: more tests are better, so do the three yaw/pitch/roll ones.
The one at the top , yes, but structure the test that the answer is obvious, rather than copy from output:
```
Unit3 axis; double angle;
std::tie(axis,angle) = R.axisAngle(0);
EXPECT(assert_equal(Unit3(0.1, 0.4, 0.2)));
EXPECT_DOUBLES_EQUAL(Vector3(0.1, 0.4, 0.2).norm());
```
Sign ambiguity: you can construct a test case with AxisAngle that has an angle `theta` larger than pi, and axisAngle should then return `theta-2*pi`. And vice-versa, angle < -pi.

dellaert:
Two comments:
1. please update title of PR ;-)
2. I was dense before: this is not a static method, but just a method: `R.axisAngle()`;

dellaert:
Notice build is failing because of gtsam.h not matching the code. You can change to method and immediately change gtsam.h in same commit.

varunagrawal:
1. Title updated.
2. I just added both class method and the static function. I am confused about the decision switch, but okay.
3. Build was failing because apparently I have to use `pair<M,N>` and not `std::pair<M,N>`. :roll_eyes:

varunagrawal:
FYI: I'll be doing a squash and merge on this PR since all the commits in this PR perform one update, so it make sense to have it being defined by 1 commit rather than pollute history with our indecisiveness.

Plus, this would be a good segue into our previous discussion on the merits of squash-and-merge.

dellaert:
Please lose the static function. It's redundant and inconsistent with other single-argument methods on Rot3. About confusion: reviewing is a process, typing is not the bottleneck.

varunagrawal:
> Please lose the static function. It's redundant and inconsistent with other single-argument methods on Rot3. About confusion: reviewing is a process, typing is not the bottleneck.

Agreed, but it need not show in the git history

varunagrawal:
Okay, everything should be fixed. A review would be nice.

gtsam.h
```diff
@@ -674,6 +674,7 @@ class Rot3 {
   double roll() const;
   double pitch() const;
   double yaw() const;
+  pair<gtsam::Unit3, double> axisAngle() const;
```
jlblancoc: What about defining: 

```
using AxisAngle = std::pair<gtsam::Unit3, double>;
``` 
?

I'm aware of it probably being a matter of taste, since both options have pros and cons...

varunagrawal: `AxisAngle` is already a function in this class, so it might make it confusing (if it compiles). I had similar thoughts too (e.g. a struct) but like you said, it's a matter of taste.

gtsam/geometry/Rot3.cpp
```diff
@@ -188,14 +188,18 @@ Vector Rot3::quaternion() const {
...
-    direction = -1;
+  Vector3 omega = Rot3::Logmap(*this);
+  // Get the rotation angle.
+  double theta = omega.norm();
+
+  // Check if angle `theta` belongs to (-pi, pi].
+  // If yes, rotate in opposite direction to maintain range.
+  // Since omega = theta * u, if all coefficients are negative,
+  // then theta is outside the expected range.
+  if ((omega.array() < 0).all()) {
```
dellaert: Can you give some more explanation as to *why* this is the right way to normalize the angle? And why does just taking -theta work>? Because Unit3 constructor will normalize omega?

varunagrawal:
After a deep dive into the math and the behavior of Axis-Angle, turns out that the angle θ ∈ [0, π]. Using the right hand rotation rule, below I elaborate on the 3 cases:

- The way the axis-angle handles π < θ < 2π, is that it flips the rotation axis to point in the opposite direction and then rotates by (2π - θ). E.g. if we try to rotate by 3π/2 aka 270 degrees, it will flip the axis and rotate by π/2, causing the same rotation effect.

- Similarly, for -π < θ < 0, it will flip the axis and rotate by θ.

- Finally for θ < -π, it will directly rotate by 2π+θ, thus always keeping the angle in [0, π].

Basically, we don't need to change anything since we are guaranteed by the math (θ is the result of an `arrcos` which is between 0 and π), that we get the angle domain as [0, π]. I have added tests to verify the same, so that should give us confidence in the code.

dellaert:
I pushed a refactor of the tests that works 100% on my Mac. But it disagreed with your tests for yaw, so I'm curious as to what travis will do.

dellaert:
> After a deep dive into the math and the behavior of Axis-Angle, turns out that the angle θ ∈ [0, π]. Using the right hand rotation rule, below I elaborate on the 3 cases:
> 
> * The way the axis-angle handles π < θ < 2π, is that it flips the rotation axis to point in the opposite direction and then rotates by (2π - θ). E.g. if we try to rotate by 3π/2 aka 270 degrees, it will flip the axis and rotate by π/2, causing the same rotation effect.
> * Similarly, for -π < θ < 0, it will flip the axis and rotate by θ.
> * Finally for θ < -π, it will directly rotate by 2π+θ, thus always keeping the angle in [0, π].
> 
> I wonder what is the best way to resolve the sign ambiguity if we are guaranteed by the math (θ is the result of an `arrcos` which is between 0 and π), that we get the angle domain as [0, π]. Or is it even necessary to do so since it is dependent on the axis? I have added tests to verify the same, so that should give us confidence in the code.

Still don't fully understand this. When you say "The way the axis-angle handles" what code are you talking about?

varunagrawal:
> I pushed a refactor of the tests that works 100% on my Mac. But it disagreed with your tests for yaw, so I'm curious as to what travis will do.

Tests look incredible. I learned something new about writing good tests today.

varunagrawal:
> > After a deep dive into the math and the behavior of Axis-Angle, turns out that the angle θ ∈ [0, π]. Using the right hand rotation rule, below I elaborate on the 3 cases:
> > 
> > * The way the axis-angle handles π < θ < 2π, is that it flips the rotation axis to point in the opposite direction and then rotates by (2π - θ). E.g. if we try to rotate by 3π/2 aka 270 degrees, it will flip the axis and rotate by π/2, causing the same rotation effect.
> > * Similarly, for -π < θ < 0, it will flip the axis and rotate by θ.
> > * Finally for θ < -π, it will directly rotate by 2π+θ, thus always keeping the angle in [0, π].
> > 
> > I wonder what is the best way to resolve the sign ambiguity if we are guaranteed by the math (θ is the result of an `arrcos` which is between 0 and π), that we get the angle domain as [0, π]. Or is it even necessary to do so since it is dependent on the axis? I have added tests to verify the same, so that should give us confidence in the code.
> 
> Still don't fully understand this. When you say "The way the axis-angle handles" what code are you talking about?

By `axis-angle` I mean the Logmap mechanism for going from Rotation Matrices to AxisAngle in GTSAM.

PR #250, opened 2020-03-16
Add TBB to Travis CI (simple)

acxz:
This is an alternative to #210. It is not as clean of a solution, but does work.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/250)


dellaert:
Is #210 not going to work? I saw you added a commit there with the last outstanding issue?

acxz:
I'm afraid I can't get it building with Travis on Ubuntu 16 w/ TBB 4 and Clang.

varunagrawal:
@acxz thanks for the great PR. I am a bit confused, don't you have to update `.travis.sh` as well to use the `GTSAM_WITH_TBB` flag in the cmake step?

acxz:
The TBB part of GTSAM gets built automatically when `intel-tbb` is installed. The CMake flags in `.travis.yml` get passed to `.travis.sh`.

varunagrawal:
`.travis.sh` is using the flag values from `travis.yml`. Yes, if TBB isn't installed, cmake will just complain that TBB isn't installed, so the addition of the `GTSAM_WITH_TBB=OFF` in `travis.yml` doesn't have any effect.

acxz:
Ah sorry my edit came through pretty late. I'll go ahead and add `GTSAM_WITH_TBB` in the `.travis.sh` file.

varunagrawal:
Cool! I am trying your PR in a docker container to see if we can figure this out. This is a valuable PR and we'd like to land it as soon as possible.

varunagrawal:
So I have found an issue with the `tbb` package itself. In DEBUG mode, we can't find TBB because the debug versions of the shared object files are not included in the `libtbb-dev` package. Installing `tbb2-dbg` doesn't help either since we need the `tbb_debug` and `tbbmalloc_debug` shared object files installed for Cmake to correctly pick them up.

Any ideas on how to fix that beyond installing from source?

acxz:
Looking into the file list for `libtbb-dev` and `libtbb2-dbg` on both xenial and bionic it seems as tho the debug *so are not packaged (https://packages.ubuntu.com/bionic/amd64/libtbb2-dbg/filelist). 
The `-dbg` packages are debug symbol packages which contain the necessary files for when using valgrind or gdb. For us to use the `_debug.so` files we would need to compile it ourselves, since Ubuntu does not package them (unlike other distributions like Gentoo).
We can create a PPA for `libtbb-debug` if we want to use a binary solution, instead of compiling it everytime in our CI.

varunagrawal:
We don't need to create a PPA for that since
1. We can just use the binaries from the Github repo.
2. Maintaining the PPA is something I don't want to do long term.

The good news is that all the tests pass on my docker container with Clang. My set up is GTSAM+Clang9+TBB4.4. Is there anything else I should test for before committing in the changes?

.travis.sh
```diff
@@ -24,6 +24,7 @@ function configure()
       -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE:-Debug} \
       -DGTSAM_BUILD_TESTS=${GTSAM_BUILD_TESTS:-OFF} \
       -DGTSAM_BUILD_UNSTABLE=${GTSAM_BUILD_UNSTABLE:-ON} \
+      -DGTSAM_WITH_TBB=${GTSAM_WITH_TBB:-ON} \
```
varunagrawal: Would it be better to set the default to `OFF` and then set the flag to `ON` where required for the sake of brevity and readability?

dellaert:
So, what's implication for #210? There was other stuff in there besides CI I think.

.travis.yml
```diff
@@ -43,7 +44,7 @@ jobs:
   - stage: compile
     os: osx
     compiler: gcc
-    env: CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF
+    env: CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF GTSAM_WITH_TBB=OFF
```
varunagrawal: Why is TBB off for only debug and not for release versions? What is the rationale behind that?

Wouldn't it be better to have a separate stage which compiles with TBB for both Debug and Release and runs the test, in the interest of time?

.travis.yml
```diff
@@ -54,7 +55,7 @@ jobs:
   - stage: compile
     os: osx
     compiler: clang
-    env: CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF
+    env: CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF GTSAM_WITH_TBB=OFF
```
varunagrawal: Same

.travis.yml
```diff
@@ -65,7 +66,7 @@ jobs:
   - stage: compile
     os: linux
     compiler: gcc
-    env: CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF
+    env: CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF GTSAM_WITH_TBB=OFF
```
varunagrawal: Same

.travis.yml
```diff
@@ -76,7 +77,7 @@ jobs:
   - stage: compile
     os: linux
     compiler: clang
-    env: CC=clang-9 CXX=clang++-9 CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF
+    env: CC=clang-9 CXX=clang++-9 CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF GTSAM_WITH_TBB=OFF
```
varunagrawal: Same

.travis.yml
```diff
@@ -87,7 +88,13 @@ jobs:
   - stage: special
     os: linux
     compiler: clang
-    env: CC=clang-9 CXX=clang++-9 CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF GTSAM_ALLOW_DEPRECATED_SINCE_V4=ON
+    env: CC=clang-9 CXX=clang++-9 CMAKE_BUILD_TYPE=Debug GTSAM_BUILD_UNSTABLE=OFF GTSAM_ALLOW_DEPRECATED_SINCE_V4=ON GTSAM_WITH_TBB=OFF
+    script: bash .travis.sh -b
+# on Linux, with GTSAM_WITH_TBB not off to make sure GTSAM still compiles/tests
```
varunagrawal: `with TBB ON to make sure`

.travis.yml
```diff
@@ -87,7 +88,13 @@ jobs:
...
+  - stage: special
```
varunagrawal: Wouldn't it make more sense to use the special stage to run with TBB ON on OSX and Linux for both Release and Debug builds? TBB is optional after all.

CMakeLists.txt
```diff
@@ -200,6 +200,11 @@ find_package(TBB 4.4 COMPONENTS tbb tbbmalloc)
 # Set up variables if we're using TBB
 if(TBB_FOUND AND GTSAM_WITH_TBB)
 	set(GTSAM_USE_TBB 1)  # This will go into config.h
+       if (${TBB_VERSION_MAJOR} GREATER_EQUAL 2020)
```
varunagrawal: Please fix the indentation here.

varunagrawal:
Apart from `ConcurrentMap.h`, the two PRs are similar in their changes to the CI.

varunagrawal:
I would figure that we land this, strip out the duplicates in #210 and land that, then focus back onto #209.

.travis.yml
varunagrawal: Ignore this. I just saw Frank's comment.

.travis.sh
acxz: I was thinking that but then I saw that GTSAM_BUILD_UNSTABLE was also like that so I didn't. I'll go ahead and make the change.

.travis.yml
acxz: I was going off @dellaert comments [here]()

CMakeLists.txt
acxz: So line 202 is indented with 5 spaces, should I indent my additions 5 spaces each or fix the indentation for line 202 to 4?

acxz:
Can you point me to the binaries you are talking about?
Also by not offering a PPA do you intend for users to obtain GTSAM via source compilation on Ubuntu?

.travis.sh
acxz: Let me know if you want me to make GTSAM_BUILD_UNSTABLE to be default to OFF and I can change that as well. Or I can make another PR for it later.

.travis.yml
varunagrawal: Cool!

CMakeLists.txt
acxz: I made the lines throughout the TBB block in the CMakeLists to be indented 4 spaces.

varunagrawal: Fix to 4.

varunagrawal:
For Ubuntu we already have a PPA for GTSAM thank some great work by @jlblancoc.
Supporting debug packages for TBB is not our prerogative and honestly, isn't required for production use either. If someone needs the debug packages, they would have to install it manually.

We can add a note to the README about this though.

As for the binaries, [this gist](https://gist.github.com/varunagrawal/96544d9a935bdf20f43ab269a8803a25) should give you everything. You can paste it into `.travis.sh` and then run a `git diff` to see the changes I've made.

.travis.sh
varunagrawal: Let's keep `GTSAM_BUILD_UNSTABLE` as is for now.

acxz:
Ah I see, thx for pointing that out to me!

varunagrawal:
Note that I haven't tested that function, but it is basically what I use in the docker container so let's see how Travis responds.

acxz:
Are there any other changes that I should commit before this gets merged?

varunagrawal:
Just the shell script update to install tbb from the binaries so it is picked up in Debug mode. That script still needs some work to be generic though, but I'll take care of it in another PR.

varunagrawal:
@acxz I cancelled the current travis build since I want to run it after you add the `install_tbb` function.

acxz:
Is this is not the same condition: https://github.com/borglab/gtsam/pull/250/commits/80fc06cad7fbb31089e4e11709fd7c81bc98327b#diff-b4553af39acae935d505c76e139d4aa3R9

acxz:
Also we are pulling in a different binary version of TBB than what Xenial uses: https://packages.ubuntu.com/source/xenial/tbb

acxz:
@varunagrawal added a commit that should fix the CI errors regarding the tbb function in .travis.sh

varunagrawal:
Ah good catches. Thanks for fixing those! :slightly_smiling_face:

varunagrawal:
@acxz seems like the `tbbvars.sh` script in `v4.4.2` has some issues and so TBB is still not being picked up. I updated the `install_tbb` function to fix this.

varunagrawal:
@acxz I finally figured out the TBB install on Travis. TBB should be picked up correctly now and Travis should pass no problem.

acxz:
Let’s go!

varunagrawal:
Travis passed. Merging in!

PR #251, opened 2020-03-17
Bunch of fixes for Rot3

varunagrawal:
- Ensure axis in `AxisAngle` constructor is a unit vector. Added test.
- Improved rotation inverse with support for quaternions.
- Use `Eigen::Transpose` for transpose return type.
- Roll/Pitch/Yaw is more efficient.
- Fix/Remove old TODOs.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/251)

### DIFF
--- a/gtsam/geometry/Point3.cpp
+++ b/gtsam/geometry/Point3.cpp
@@ -29,10 +29,7 @@ bool Point3::equals(const Point3 &q, double tol) const {
 }
 
 void Point3::print(const string& s) const {
-  if (s.size() != 0) {
-    cout << s << " ";
-  }
-  cout << *this << endl;
+  cout << s << *this << endl;
 }
 
 /* ************************************************************************* */
--- a/gtsam/geometry/Point3.cpp
+++ b/gtsam/geometry/Point3.cpp
@@ -29,7 +29,10 @@ bool Point3::equals(const Point3 &q, double tol) const {
 }
 
 void Point3::print(const string& s) const {
-  cout << s << *this << endl;
+  if (s.size() != 0) {
+    cout << s << " ";
+  }
+  cout << *this << endl;
 }
 
 /* ************************************************************************* */
--- a/gtsam/geometry/Rot3.cpp
+++ b/gtsam/geometry/Rot3.cpp
@@ -36,7 +36,6 @@ void Rot3::print(const std::string& s) const {
 
 /* ************************************************************************* */
 Rot3 Rot3::Random(std::mt19937& rng) {
-  // TODO allow any engine without including all of boost :-(
   Unit3 axis = Unit3::Random(rng);
   uniform_real_distribution<double> randomAngle(-M_PI, M_PI);
   double angle = randomAngle(rng);
--- a/gtsam/geometry/Rot3.cpp
+++ b/gtsam/geometry/Rot3.cpp
@@ -16,6 +16,7 @@
  * @author  Christian Potthast
  * @author  Frank Dellaert
  * @author  Richard Roberts
+ * @author  Varun Agrawal
  */
 
 #include <gtsam/geometry/Rot3.h>
@@ -182,6 +183,12 @@ Vector Rot3::quaternion() const {
   v(2) = q.y();
   v(3) = q.z();
   return v;
+}
+
+/* ************************************************************************* */
+pair<Unit3, double> Rot3::axisAngle() const {
+  const Vector3 omega = Rot3::Logmap(*this);
+  return std::pair<Unit3, double>(Unit3(omega), omega.norm());
 }
 
 /* ************************************************************************* */
--- a/gtsam/geometry/Rot3.cpp
+++ b/gtsam/geometry/Rot3.cpp
@@ -197,27 +197,15 @@ Matrix3 Rot3::LogmapDerivative(const Vector3& x)    {
 /* ************************************************************************* */
 pair<Matrix3, Vector3> RQ(const Matrix3& A) {
 
-  // atan2 has curious/unstable behavior near 0.
-  // If either x or y is close to zero, the results can vary depending on
-  // what the sign of either x or y is.
-  // E.g. for x = 1e-15, y = -0.08, atan2(x, y) != atan2(-x, y),
-  // even though arithmetically, they are both atan2(0, y).
-  // Suppressing small numbers to 0.0 doesn't work since the floating point
-  // representation still causes the issue due to a delta difference.
-  // The only fix is to convert to an int so we are guaranteed the value is 0.
-  // This lambda function checks if a number is close to 0.
-  // If yes, then return the integer representation, else do nothing.
-  auto suppress = [](auto x) { return abs(x) > 1e-15 ? x : int(x); };
-
-  double x = -atan2(-suppress(A(2, 1)), suppress(A(2, 2)));
+  double x = -atan2(-A(2, 1), A(2, 2));
   Rot3 Qx = Rot3::Rx(-x);
   Matrix3 B = A * Qx.matrix();
 
-  double y = -atan2(suppress(B(2, 0)), suppress(B(2, 2)));
+  double y = -atan2(B(2, 0), B(2, 2));
   Rot3 Qy = Rot3::Ry(-y);
   Matrix3 C = B * Qy.matrix();
 
-  double z = -atan2(-suppress(C(1, 0)), suppress(C(1, 1)));
+  double z = -atan2(-C(1, 0), C(1, 1));
   Rot3 Qz = Rot3::Rz(-z);
   Matrix3 R = C * Qz.matrix();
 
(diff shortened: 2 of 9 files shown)


gtsam/geometry/Point3.cpp
```diff
@@ -29,7 +29,10 @@ bool Point3::equals(const Point3 &q, double tol) const {
 }
 
 void Point3::print(const string& s) const {
-  cout << s << *this << endl;
+  if (s.size() != 0) {
+    cout << s << " ";
```
dellaert: Changes behavior, should revert. People might be relying on old behavior.

gtsam/geometry/Rot3.h
```diff
@@ -194,22 +194,24 @@ namespace gtsam {
...
     /**
      * Convert from axis/angle representation
-     * @param  axisw is the rotation axis, unit length
-     * @param   angle rotation angle
+     * @param  axis is the rotation axis, unit length
+     * @param  angle rotation angle
      * @return incremental rotation
      */
-    static Rot3 AxisAngle(const Point3& axis, double angle) {
+    static Rot3 AxisAngle(const Vector3& axis, double angle) {
```
dellaert: Changing type: not a good idea: Point3 and Vector3 are not always the same. See Compilation flags.

gtsam/geometry/Rot3.h
```diff
@@ -268,9 +270,13 @@ namespace gtsam {
     /// Syntatic sugar for composing two rotations
     Rot3 operator*(const Rot3& R2) const;
 
-    /// inverse of a rotation, TODO should be different for M/Q
+    /// inverse of a rotation
     Rot3 inverse() const {
-      return Rot3(Matrix3(transpose()));
+#ifdef GTSAM_USE_QUATERNIONS
+      return quaternion_.inverse();
```
dellaert: be explicit about calling Rot3 constructor. Should be made explicit probably.

gtsam/geometry/Rot3.h
```diff
@@ -268,9 +270,13 @@ namespace gtsam {
...
+#else
+      return Rot3(transpose());
```
dellaert: Since you're changing it, why not Rot3(rot_.transpose()) ?

gtsam/geometry/Rot3.h
```diff
@@ -405,8 +411,7 @@ namespace gtsam {
     /**
      * Return 3*3 transpose (inverse) rotation matrix
      */
-    Matrix3 transpose() const;
-    // TODO: const Eigen::Transpose<const Matrix3> transpose() const;
+    const Eigen::Transpose<const Matrix3> transpose() const;
```
dellaert: I like it.

gtsam/geometry/Rot3.h
```diff
@@ -435,27 +440,18 @@ namespace gtsam {
 
     /**
      * Accessor to get to component of angle representations
-     * NOTE: these are not efficient to get to multiple separate parts,
-     * you should instead use xyz() or ypr()
-     * TODO: make this more efficient
      */
-    inline double roll() const  { return ypr()(2); }
+    inline double roll() const  { return xyz()(0); }
```
dellaert: TODO of making more efficient is gone, but *is this more efficient?

gtsam/geometry/Point3.cpp
varunagrawal: Hmm I figured the printing would not affect actual behavior, but I guess for logging/metrics.

gtsam/geometry/Rot3.h
varunagrawal: Good point. Sorry for overlooking that.

gtsam/geometry/Rot3.h
varunagrawal: Done.

gtsam/geometry/Rot3.h
varunagrawal: So be more explicit? I figured I'd make minimal changes to the code.

dellaert: More efficient. And Rot3 constructor with one argument should be explicit.

gtsam/geometry/Rot3.h
varunagrawal: The TODO note said to use either `ypr` or `xyz`

```
* NOTE: these are not efficient to get to multiple separate parts,
* you should instead use xyz() or ypr()
* TODO: make this more efficient
```

I figured it was a leftover comment from a previous commit (similar to the boost TODO for rng). `ypr` was internally calling `xyz` so all I did was remove the extra layer. Internally it is still calling the `RQ` algorithm.

dellaert: So, add the TODO again :-)

varunagrawal:
@dellaert there's a bunch of issues with the quaternion representation. I've opened #253 to track the same. Can't merge this yet. :cry:

gtsam/geometry/Rot3.cpp
```diff
@@ -197,27 +197,15 @@ Matrix3 Rot3::LogmapDerivative(const Vector3& x)    {
...
-  // even though arithmetically, they are both atan2(0, y).
-  // Suppressing small numbers to 0.0 doesn't work since the floating point
-  // representation still causes the issue due to a delta difference.
-  // The only fix is to convert to an int so we are guaranteed the value is 0.
-  // This lambda function checks if a number is close to 0.
-  // If yes, then return the integer representation, else do nothing.
-  auto suppress = [](auto x) { return abs(x) > 1e-15 ? x : int(x); };
-
-  double x = -atan2(-suppress(A(2, 1)), suppress(A(2, 2)));
+  double x = -atan2(-A(2, 1), A(2, 2));
```
dellaert: Uhm, please add more justification than “I removed it’ ;-)

varunagrawal: This is an undo of a previous commit I had introduced in this PR. The stability is less of a problem compared to the behavior of `Eigen::Transpose`.

varunagrawal:
`matrix().eval().transpose()` gives incorrect results which leads to test failures. Half the values end up being `1e-200` for some reason. This is something definitely to do with `Eigen::Tranpose`. The docs aren't exactly helpful and I can't figure out what I'm doing wrong. :/

varunagrawal:
So after a LOT of investigation, there is something weird about the Eigen expression templates for transpose aka `Eigen::Transpose<const Matrix3>` when using the quaternion representation. It is  hard to figure out why since `eval()` doesn't work as it should (the transpose works except for `mat(2, 2)` which is some random `1e-10`) and I can't find anything in the docs or online to help with this.

This is clearly some issue with either how we're using the expression template or with how it is managed for quaternions. The odd thing is, if we print the value before returning, Eigen understands how to evaluate the expression correctly. The only other option is to use `eval()` in all the downstream code which would mean something like `R.transpose().eval()` which is ugly.

Any Eigen experts we can consult to figure this out? @thduynguyen?

ProfFan:
Note that `matrix()` is a temporary and its lifetime is *the function itself*

varunagrawal:
Even when we override it? I did `matrix().eval()` but that didn't help.

dellaert:
@varunagrawal can't we just go back to the previous implementation? And, is quaternion one of the paths in Travis? I don't think you replied to that question.

varunagrawal:
My last commit undid the Eigen::Transpose change while also marking some
methods as const.

Travis does have a path for Quaternion rotations which is why the template
expression behaving so strangely is a mystery to me.

gtsam/geometry/Rot3Q.cpp
```diff
@@ -79,10 +79,14 @@ namespace gtsam {
...
-  // const Eigen::Transpose<const Matrix3> Rot3::transpose() const {
-  Matrix3 Rot3::transpose() const {
+  // TODO: Maybe use return type `const Eigen::Transpose<const Matrix3>`?
+  // It works in Rot3M but not here, because of some weird behavior
+  // due to Eigen's expression templates which needs more investigation.
+  // For example, if we use matrix(), the value returned has a 1e-10,
+  // and if we use quaternion_.toRotationMatrix(), the matrix is arbitrary.
+  // Using eval() here doesn't help, it only helps if we use it in
+  // the downstream code.
+  const Matrix3 Rot3::transpose() const {
```
dellaert: remove const

gtsam/geometry/Rot3Q.cpp
```diff
@@ -116,7 +120,7 @@ namespace gtsam {
   }
 
   /* ************************************************************************* */
-  Matrix3 Rot3::matrix() const {return quaternion_.toRotationMatrix();}
+  const Matrix3 Rot3::matrix() const {return quaternion_.toRotationMatrix();}
```
dellaert: remove const

PR #255, opened 2020-03-20
Python plotting upgrades

varunagrawal:
Added functions for

- Plotting trajectory
- Plotting marginal covariances in 3D as ellipsoids.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/255)


varunagrawal:
Shouldn't that be fine since more than 80% of the code is using pure python or matplotlib. Only certain functions to get the data from GTSAM structures would need to be updated for the new wrapper.

dellaert:
Well, the issue is we have two wrappers now, and we will soon delete the old one.

ProfFan:
@dellaert Just a side note that the new wrapper is still blocked by the CI timeout :(

varunagrawal:
Wouldn't the transition process involve moving these examples and updating them?
I am not privy to the discussions regarding the new wrapper but I assume that a migration plan is in the works.

dellaert:
That already happened, so new examples/changes need to be mirrored, hence the comment.

varunagrawal:
@dellaert @ProfFan made a PR for the new wrapper which mirrors these changes. If you're happy with this PR, we can add this to the Cython wrapper until the PyBind11 wrapper is ready.

PR #256, opened 2020-03-21
Various improvements

varunagrawal:
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/256)


gtsam/geometry/Cal3DS2.h
```diff
@@ -33,9 +33,13 @@ class GTSAM_EXPORT Cal3DS2 : public Cal3DS2_Base {
 
   typedef Cal3DS2_Base Base;
 
+private:
+
+  static const size_t dimension_ = 9;
```
dellaert: Why?

gtsam/linear/NoiseModel.cpp
```diff
@@ -311,10 +312,9 @@ void Diagonal::WhitenInPlace(Eigen::Block<Matrix> H) const {
 
 namespace internal {
 // switch precisions and invsigmas to finite value
-// TODO: why?? And, why not just ask s==0.0 below ?
 static void fix(const Vector& sigmas, Vector& precisions, Vector& invsigmas) {
   for (Vector::Index i = 0; i < sigmas.size(); ++i)
-    if (!std::isfinite(1. / sigmas[i])) {
+    if (sigmas[i] <= 1e-12) {
```
ProfFan: Is there a reason for this change?

BTW we are in progress of reviewing the entire Constrained model...

gtsam/linear/NoiseModel.cpp
```diff
@@ -341,8 +341,8 @@ Constrained::shared_ptr Constrained::MixedSigmas(const Vector& mu,
 
 /* ************************************************************************* */
 bool Constrained::constrained(size_t i) const {
-  // TODO why not just check sigmas_[i]==0.0 ?
-  return !std::isfinite(1./sigmas_[i]);
+  // numerically stable way, rather than comparing to 0.0 directly.
```
ProfFan: Same

gtsam/geometry/Cal3DS2.h
varunagrawal: This was a TODO and it made sense from a future-proofing perspective to make the variable static.

gtsam/linear/NoiseModel.cpp
varunagrawal: This was a TODO item and seemed cheaper and better than checking for infinite. Made sense to me and the change was easy so cleared that technical debt.

gtsam/linear/NoiseModel.cpp
varunagrawal: Clearing technically debt.

gtsam/linear/NoiseModel.cpp
ProfFan: I think the intention here is to explicitly check for 0.

Ping @dellaert

gtsam/geometry/Cal3DS2.h
dellaert: Sorry, no, this change seems to *add* to the plethora of places where we define dimension. the `enum { dimension = 9 }` sees to already do the trick. Just use that in the functions.

gtsam/linear/NoiseModel.cpp
dellaert: Indeed, this is a behavior change. Please undo or replace with a better test that has the *same* behavior, and explain how.

gtsam/linear/NoiseModel.cpp
```diff
@@ -341,8 +341,8 @@ Constrained::shared_ptr Constrained::MixedSigmas(const Vector& mu,
...
+  return sigmas_[i] <= 1e-12;
```
dellaert: Same comment

gtsam/geometry/Cal3DS2.h
varunagrawal: I am a bit confused. I am not sure how it adds, since we are simply replacing `enum { dimension = 9 }` with `enum {dimension = dimension_}`. Also, I believe the enum version doesn't compile, the reason for which I don't know off the top of my head.

gtsam/linear/NoiseModel.cpp
varunagrawal: Wouldn't numerical stability require that we don't explicitly check for 0? A value like `6e-30` (which is considered 0) would not pass that check and that would explain why the original code was checking for infinity instead.

dellaert: Comments still stands as is. I have no time to engage in long dialog :-(

gtsam/linear/NoiseModel.cpp
```diff
@@ -136,14 +136,18 @@ bool Gaussian::equals(const Base& expected, double tol) const {
...
+  //TODO(Alex);
+  //if (!sqrt_information_) return true;
   return equal_with_abs_tol(R(), p->R(), sqrt(tol));
 }
 
 /* ************************************************************************* */
 Vector Gaussian::sigmas() const {
-  // TODO(frank): can this be done faster?
-  return Vector((thisR().transpose() * thisR()).inverse().diagonal()).cwiseSqrt();
+  Matrix R = thisR();
```
dellaert: Explain why this is faster?

varunagrawal: Since we know R is upper triangular, using `triangularView` lets us use `.solve(m)` which effectively performs `this.inverse()*m` but in an efficient way using backsubstitution with the triangular matrix. Thus `n.solve(I) = n.inverse()*I = n.inverse()` which is cheaper than a full inverse aka `< O(n^3)`.

varunagrawal: https://eigen.tuxfamily.org/dox/classEigen_1_1TriangularViewImpl_3_01__MatrixType_00_01__Mode_00_01Dense_01_4.html#a911664ccf5522c778ffc5405247e8a59

varunagrawal:
@ProfFan I undid the sigma checks, saved it for another day.

gtsam/linear/NoiseModel.cpp
dellaert: Actually, you are right, but in that case we should do that saving in `covariance`, moved to the .cpp file. And here just say `covariance(). diagonal()).cwiseSqrt()`

varunagrawal: Done. Running tests locally and then will commit.

varunagrawal:
@dellaert @ProfFan bump

gtsam/linear/NoiseModel.cpp
```diff
@@ -135,14 +135,25 @@ bool Gaussian::equals(const Base& expected, double tol) const {
   const Gaussian* p = dynamic_cast<const Gaussian*> (&expected);
   if (p == nullptr) return false;
   if (typeid(*this) != typeid(*p)) return false;
-  //if (!sqrt_information_) return true; // ALEX todo;
+  //TODO(Alex);
```
dellaert, line 138: Just kill these two commented out lines.

gtsam/linear/NoiseModel.cpp
```diff
@@ -135,14 +135,25 @@ bool Gaussian::equals(const Base& expected, double tol) const {
...
+  //if (!sqrt_information_) return true;
   return equal_with_abs_tol(R(), p->R(), sqrt(tol));
 }
 
+/* ************************************************************************* */
+Matrix Gaussian::covariance() const {
+  // Uses a fast version of `covariance = information().inverse();`
+  Matrix R = this->R();
```
dellaert, line 146: You are making an unnecessary copy here. `const Matrix&`

gtsam/linear/NoiseModel.cpp
```diff
@@ -135,14 +135,25 @@ bool Gaussian::equals(const Base& expected, double tol) const {
...
+  Matrix I = Matrix::Identity(R.rows(), R.cols());
+  // Fast inverse of upper-triangular matrix R using forward-substitution
+  Matrix Rinv = R.triangularView<Eigen::Upper>().solve(I);
+  // (R' * R)^{-1} = R^{-1} * R^{-1}'
+  return (Rinv * Rinv.transpose());
```
dellaert, line 151: lose outer parens

gtsam/linear/NoiseModel.cpp
```diff
@@ -135,14 +135,25 @@ bool Gaussian::equals(const Base& expected, double tol) const {
...
   return equal_with_abs_tol(R(), p->R(), sqrt(tol));
 }
 
+/* ************************************************************************* */
+Matrix Gaussian::covariance() const {
+  // Uses a fast version of `covariance = information().inverse();`
+  Matrix R = this->R();
+  Matrix I = Matrix::Identity(R.rows(), R.cols());
+  // Fast inverse of upper-triangular matrix R using forward-substitution
+  Matrix Rinv = R.triangularView<Eigen::Upper>().solve(I);
```
dellaert, line 149: Actually, https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html says we should use LLT. as does https://stackoverflow.com/questions/38640563/eigen-efficient-inverse-of-symmetric-positive-definite-matrix

gtsam/linear/NoiseModel.cpp
varunagrawal: Nice catch!

gtsam/linear/NoiseModel.cpp
varunagrawal: Umm I don't understand. `R` is already upper triangular, so why decompose again? Since this is my call, I am happy sticking to the current implementation.

varunagrawal:
@dellaert made the requested changes. Will merge once travis passes.

Issue #237, opened 2020-03-04
Mismatch between calibrate functions for Cal3_S2 and Cal3Bundler


varunagrawal:
## Description

`PinholeCamera` can accept both `Cal3_S2` and `Cal3Bundler` as calibrations, but the `calibrate` function is defined differently for each, such that when `camera.backproject(pn, depth)` is called, `Cal3Bundler` errors out.

## Steps to reproduce

This unit test when added to `testPinholeCamera.cpp` raises the error:

```cpp
TEST( PinholeCamera, Cal3Bundler) {
  Cal3Bundler calibration;
  Pose3 wTc;
  PinholeCamera<Cal3Bundler> camera(wTc, calibration);
  Point2 p(50, 100);
  Point3 point = camera.backproject(p, 10);
}
```

## Expected behavior

`PinholeCamera` (and primarily it's base class `PinholeBase`) should be able to call `calibration().calibrate` without any errors for all calibration types.

## Environment

N/A

varunagrawal:
This issue is easy to fix once we figure out how to handle the tolerance argument in `Cal3Bundler` so that the method signature is consistent.

johnwlambert:
@varunagrawal It looks like we can't access the `camera` parameter of `SfmData` in the wrapper until we resolve this (https://github.com/borglab/gtsam/blob/develop/gtsam/gtsam.i#L2865). Any idea for a path forward for resolving the tolerance argument issue?

varunagrawal:
This is something @dellaert and I need to discuss at a design level so I can't comment on that *sigh*.

I've had a PR ready for months actually, but it breaks the API signature which is what is causing the delay.

dellaert:
I’m sorry, what’s the issue?

varunagrawal:
`Cal3_S2` has `calibrate` defined as

```cpp
Point2 uncalibrate(const Point2& p, OptionalJacobian<2,5> Dcal = boost::none,
      OptionalJacobian<2,2> Dp = boost::none) const;
```

while `Cal3Bundler` has it defined as
```cpp
Point2 calibrate(const Point2& pi, const double tol = 1e-5) const;
```

aka no `Dcal` and `Dp`, so this causes issues with wrapping and use cases in optimization.

varunagrawal:
@dellaert the original PR I made is here: https://github.com/borglab/gtsam/pull/257

dellaert:
“Causes issues” ?

varunagrawal:
[...]

PR #257, opened 2020-03-21
Derivatives for Distortion models

varunagrawal:
This PR tackles issue #237 

This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/257)


dellaert:
We cannot just change the parameterization on an existing calibration model. If you want/need a new calibration model with more parameters, modify this PR accordingly and re-open. Although, yo have a sane git history, this is probably better done by starting from scratch and copying your work over to a derived class, so that code duplication is minimized.

PR #259, opened 2020-03-21
GitHub Actions CI

ProfFan:
Example output and build result:
https://github.com/ProfFan/gtsam/actions

This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/259)

### DIFF
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -59,8 +59,9 @@ jobs:
           if [ "${{ matrix.compiler }}" = "clang" ] && [ "${{ matrix.version }}" = "9" ]; then
             sudo add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main"
           fi
-          sudo apt-get update
-
+          sudo apt-get -y update
+          sudo apt-get -y upgrade
+          
           sudo apt install cmake g++-9 clang-9 build-essential pkg-config libpython-dev python-numpy libboost-all-dev
           
           if [ "${{ matrix.compiler }}" = "gcc" ]; then
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -119,14 +119,3 @@ jobs:
         if: runner.os == 'macOS'
         run: |
           bash .travis.python.sh
-      - name: Build (Windows)
-        if: runner.os == 'Windows'
-        run: |
-          cmake -E remove_directory build
-          cmake -B build -S . -DGTSAM_BUILD_EXAMPLES_ALWAYS=OFF
-          cmake --build build --config ${{ matrix.build_type }} --target gtsam
-          cmake --build build --config ${{ matrix.build_type }} --target gtsam_unstable
-          cmake --build build --config ${{ matrix.build_type }} --target wrap
-          cmake --build build --config ${{ matrix.build_type }} --target check.base
-          cmake --build build --config ${{ matrix.build_type }} --target check.base_unstable
-          cmake --build build --config ${{ matrix.build_type }} --target check.linear
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -1,17 +1,18 @@
-name: CI
+name: Python CI
 
 on: [push]
 
 jobs:
   build:
-    name: ${{ matrix.name }} ${{ matrix.build_type }} ${{ matrix.build_unstable }}
+    name: ${{ matrix.name }} ${{ matrix.build_type }} ${{ matrix.wrapper }} ${{ matrix.python_version }}
     runs-on: ${{ matrix.os }}
 
     env:
       CTEST_OUTPUT_ON_FAILURE: ON
       CTEST_PARALLEL_LEVEL: 2
       CMAKE_BUILD_TYPE: ${{ matrix.build_type }}
-      GTSAM_BUILD_UNSTABLE: ${{ matrix.build_unstable }}
+      PYTHON_VERSION: ${{ matrix.python_version }}
+      WRAPPER: ${{ matrix.wrapper }}
     strategy:
       fail-fast: false
       matrix:
@@ -21,13 +22,12 @@ jobs:
           ubuntu-18.04-gcc-5,
           ubuntu-18.04-gcc-9,
           ubuntu-18.04-clang-9,
-          windows-2016-cl,
-          windows-2019-cl,
           macOS-10.15-xcode-11.3.1,
         ]
 
         build_type: [Debug, Release]
-        build_unstable: [ON, OFF]
+        python_version: 3
+        wrapper: [cython, pybind]
         include:
           - name: ubuntu-18.04-gcc-5
             os: ubuntu-18.04
@@ -43,14 +43,6 @@ jobs:
             os: ubuntu-18.04
             compiler: clang
             version: "9"
-
-          - name: windows-2016-cl
-            os: windows-2016
-            compiler: cl
-
-          - name: windows-2019-cl
-            os: windows-2019
-            compiler: cl
 
           - name: macOS-10.15-xcode-11.3.1
             os: macOS-10.15
@@ -121,11 +113,11 @@ jobs:
       - name: Build (Linux)
         if: runner.os == 'Linux'
         run: |
-          bash .travis.sh -t
+          bash .travis.python.sh
       - name: Build (macOS)
         if: runner.os == 'macOS'
         run: |
-          bash .travis.sh -t
+          bash .travis.python.sh
       - name: Build (Windows)
         if: runner.os == 'Windows'
         run: |
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -26,7 +26,7 @@ jobs:
         ]
 
         build_type: [Debug, Release]
-        python_version: 3
+        python_version: [3]
         wrapper: [cython, pybind]
         include:
           - name: ubuntu-18.04-gcc-5
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -1,6 +1,6 @@
 name: Python CI
 
-on: [push]
+on: [pull_request]
 
 jobs:
   build:
--- a/gtsam/base/GenericValue.h
+++ b/gtsam/base/GenericValue.h
@@ -29,13 +29,19 @@
 #include <iostream>
 #include <typeinfo> // operator typeid
 
+#ifdef _WIN32
+#define GENERICVALUE_VISIBILITY
+#else
+#define GENERICVALUE_VISIBILITY GTSAM_EXPORT
+#endif
+
 namespace gtsam {
 
 /**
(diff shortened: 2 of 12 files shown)


.github/workflows/build-python.yml
```diff
@@ -0,0 +1,131 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        # Github Actions requires a single row to be added to the build matrix.
+        # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
+        name: [
+          ubuntu-18.04-gcc-5,
+          ubuntu-18.04-gcc-9,
+          ubuntu-18.04-clang-9,
+          macOS-10.15-xcode-11.3.1,
```
varunagrawal: Doesn't Github actions also support Windows?

.github/workflows/build-python.yml
```diff
@@ -0,0 +1,131 @@
...
+      - name: Install (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          # LLVM 9 is not in Bionic's repositories so we add the official LLVM repository.
+          if [ "${{ matrix.compiler }}" = "clang" ] && [ "${{ matrix.version }}" = "9" ]; then
+            sudo add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main"
+          fi
+          sudo apt-get update
+
+          sudo apt install cmake g++-9 clang-9 build-essential pkg-config libpython-dev python-numpy libboost-all-dev
```
varunagrawal: Won't installing `clang-9` fail when the compiler is `gcc`, since the above if statement would be skipped?

varunagrawal: Same for gcc. We are explicitly installing GCC below.

gtsam/base/GenericValue.h
```diff
@@ -29,13 +29,19 @@
 #include <iostream>
 #include <typeinfo> // operator typeid
 
+#ifdef _WIN32
+#define GENERICVALUE_VISIBILITY
+#else
+#define GENERICVALUE_VISIBILITY GTSAM_EXPORT
+#endif
+
```
varunagrawal: If this is only in one place, wouldn't it be better to just conditionally compile the function signature?

varunagrawal:
Looking at all the steps still required for initial setup, using Docker images as bases to clone and build in still seems like a valid plan. Save that extra bit of time to instead actually run the tests.

.github/workflows/build-python.yml
ProfFan: The Python wrapper does not, unfortuately :)

.github/workflows/build-python.yml
ProfFan: I didn't have this issue, all tests pass on my fork, but it is indeed better to be more specific.

gtsam/base/GenericValue.h
ProfFan: This is for a class, I do not think it is good to ifdef on class definitions.

chrisbeall: Why is this workaround needed now, when Windows compiler was ok with `GTSAM_EXPORT` in the past? Might also be good to add a comment for future reference.

varunagrawal:
Only problem with Docker is the lack of OSX support *sigh*

ProfFan:
@varunagrawal I think it is good to add one or two Docker-based build envs, in addition to the normal builds.

varunagrawal:
Another thing is that perhaps we should merge the compile and test stages since they follow from one another? If compilation fails, we'll get the message that it did, similarly for testing (especially since Make is verbose).

gtsam/base/GenericValue.h
ProfFan: @chrisbeall This will trigger a `LNKxxx` error on Windows, the reason is in https://github.com/borglab/gtsam/blob/develop/Using-GTSAM-EXPORT.md, I will add a ref to that :)

ProfFan: Also, the sad truth is that while we used to have a Windows CI, it only runs a very small subset of the test suite, and this problem is caught by the Actions CI (which runs more tests). Anyway, we need someone with adequate Windows experience to solve this in a **final** way.

CC @dellaert @jlblancoc

ProfFan:
My current plan:

1. Push this directly to a branch in `borglab/gtsam` so we can have it actually built
2. Merge if that works

@dellaert Are you comfortable with this?

dellaert:
Yes, if (2) is “PR if that works”

ProfFan:
For `Python CI / ubuntu-18.04-gcc-9 Debug cython 3 (push) `:
```python3
ERROR: test_VisualISAMExample (test_VisualISAMExample.TestVisualISAMExample)
  File "/home/runner/work/gtsam/gtsam/cython/gtsam/tests/test_VisualISAMExample.py", line 39, in test_VisualISAMExample
    data, truth = generator.generate_data(options)
  File "/usr/local/lib/python3.6/dist-packages/gtsam-4.0.2-py3.6.egg/gtsam/utils/visual_data_generator.py", line 109, in generate_data
    data.Z[i][j] = truth.cameras[i].project(truth.points[j])
  File "gtsam/gtsam.pyx", line 3450, in gtsam.gtsam.SimpleCamera.project
RuntimeError: CheiralityException
```

ProfFan:
@dellaert  @varunagrawal  Any prior experience with this exception?

ProfFan:
Also, #183 will need this to be merged first.

ProfFan:
CI / ubuntu-18.04-gcc-9 Release ON (push):
```c++
136/240 Test #136: testAHRSFactor .........................***Failed    0.00 sec
not equal:
expected = [
 	  1.18993e-98             0   4.40466e-99;
    	-3.17812e-212             0  1.66916e-212;
    	 -4.40466e-99             0   1.18993e-98
  ]
actual = [
 	 8.01711e-99  1.24871e-99 -3.79549e-99;
    	 1.26286e-99        -1.01 -1.13679e-99;
    	 3.79549e-99 -1.17784e-99  8.01711e-99
  ]
actual - expected = [
 	-3.88215e-99  1.24871e-99 -8.20015e-99;
    	 1.26286e-99        -1.01 -1.13679e-99;
    	 8.20015e-99 -1.17784e-99 -3.88215e-99
  ]
/home/runner/work/gtsam/gtsam/gtsam/navigation/tests/testAHRSFactor.cpp:374: Failure: "assert_equal(expectedDelRdelBiasOmega, preintegrated.delRdelBiasOmega(), 1e-3)" 
```

ProfFan:
catching exception by value:
```
/home/runner/work/gtsam/gtsam/gtsam/geometry/tests/testEssentialMatrix.cpp:162:22: warning: catching polymorphic type ‘class std::exception’ by value [-Wcatch-value=]
  162 |   } catch (exception e) {
      |                      ^
```

ProfFan:
@jlblancoc @dellaert  I think I reproduced the problem with `testAHRSFactor` in #168, and it is not a problem with 32-bit machines.

Here:
```python
actual - expected = [
 	-3.88215e-99  1.24871e-99 -8.20015e-99;
    	 1.26286e-99        -1.01 -1.13679e-99;
    	 8.20015e-99 -1.17784e-99 -3.88215e-99
  ]
```

#168:

```python
actual - expected = [
 	        -1.01 -5.15619e-310 -3.09371e-309;
    	-4.62736e-310  3.10108e-310 -8.38192e-310;
    	-3.15341e-309  8.38192e-310  3.10108e-310
  ]
```

dellaert:
Both issues are bizarre:
1) cheirality is feature appearing behind camera, but test is supposed to be deterministic, not? Why would this happen here?
2) the testAHRSFactor issue is mystifying, as well. Some unallocated memory issue?

ProfFan:
@dellaert Based on my prior experience upgrading GCC versions they both should be optimization related, as GCC is being increasingly stringent on undefined behaviors.

I'll spin up a test environment and try debugging.

ProfFan:
@dellaert I did more experiments today, and the `CheiralityError` is spurious: It happens randomly between executions, so I suspect it's the data generator that is acting funky.

ProfFan:
Some more debugging: when this error occurs, the camera is:
```matlab
truth.points[j] =  [10, 10, 10]'

truth.cameras[i] =  .poseR:
[
 	-nan -nan   -0;
    	-nan -nan    0;
    	-nan -nan   -0
  ]
[40, 0, 10]';.calibration[
 	500   0 320;
    	  0 500 240;
    	  0   0   1
  ]
```

dellaert:
Why, that seems bad :-)

ProfFan:
@dellaert No idea... I am debugging now...

ProfFan:
OK, I think I am making progress now...

This is definitely a race condition somewhere, as if I turn on buffering of stdout (so tests become very fast), I can 100% reproduce the issue...

ProfFan:
I figured it out by attaching a debugger on the Python process and inspect the data. It's because the file is using the **default constructor** of `gtsam.Point3()` as if it is zero initializing the Point3, but actually it is **NOT**:

The uninitialized Point3 contains:
```
 [2.06513e-316, 1.68324e+212, 1.58101e-322]'
```

Thus the spurious error as the memory contents between each execution is unpredictable.

ProfFan:
BTW, I think the default constructor is only meant to be used for deserialization...

ProfFan:
Actually there is this piece of code from `geometry/Point3.h`

```c++
    // Deprecated default constructor initializes to zero, in contrast to new behavior below
#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V4
    Point3() { setZero(); }
#endif
```
So in the past we zeroized the Point3, but as the CI disables this flag the default constructor becomes the constructor of `Eigen::Vector3`, which does not zeroize.

dellaert:
> Actually there is this piece of code from `geometry/Point3.h`
> 
> ```c++
>     // Deprecated default constructor initializes to zero, in contrast to new behavior below
> #ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V4
>     Point3() { setZero(); }
> #endif
> ```
> 
> So in the past we zeroized the Point3, but as the CI disables this flag the default constructor becomes the constructor of `Eigen::Vector3`, which does not zeroize.

Yes, I know

dellaert:
> BTW, I think the default constructor is only meant to be used for deserialization...

No, you also need it to make std::vector

ProfFan:
Since all tests pass (sans one on Windows that timed out), I think this PR is ready for review :)

@dellaert @jlblancoc Thanks in advance!

dellaert:
> wow! amazing job!
> 
> I would only suggest, only if @dellaert approves that policy, to do a git rebase to reduce the number of commits from 32 to something like 1-5?

I discourage any rebasing on branches that have been pushed - but you could squash-merge this? Let me do a quick review.

ProfFan:
@dellaert I vote to first let the two run in parallel - this would help us identify the issues with Actions and familiarize ourselves with the new system. Also, currently the Actions-based CI is only in the Pybind branch, so we are not changing the behavior concerning PRs made to `develop` anyway.

PR #264, opened 2020-04-01
Fix Unit Test when GTSAM_DEFINE_POINTS_AS_VECTORS is defined

ProfFan:
When `GTSAM_TYPEDEF_POINTS_TO_VECTORS` is defined the output format will change because it is now an Eigen vector.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/264)


ProfFan:
@dellaert because we do not have this in the test matrix...

PR #274, opened 2020-04-07
printError(): allow custom factor filter

jlblancoc:
I found this feature really useful while debugging whether one or a few factors were not honored as well as the rest, i.e. outliers. 

Please, check and let me know if you would prefer alternative signatures... or defining a named type for the optional functor, etc. There are just too many ways of doing even the simplest thing ;-)

At the beginning I tried implementing a "relative filter", with two passes: one to collect stats on all factor errors, then decide based on some statistic. But that requires dynamic memory allocation (for the stats), doing the two passes, etc. so it's probably not worth. Having the option to use individual whitened errors to decide whether to print or not seems to me a good trade-off.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/274)

### DIFF
--- a/gtsam/nonlinear/NonlinearFactorGraph.cpp
+++ b/gtsam/nonlinear/NonlinearFactorGraph.cpp
@@ -61,19 +61,26 @@ void NonlinearFactorGraph::print(const std::string& str, const KeyFormatter& key
 
 /* ************************************************************************* */
 void NonlinearFactorGraph::printErrors(const Values& values, const std::string& str,
-                                       const KeyFormatter& keyFormatter) const {
+    const KeyFormatter& keyFormatter,
+    const std::function<bool(const Factor* /*factor*/, double /*whitenedError*/, size_t /*index*/)>& printCondition) const
+{
   cout << str << "size: " << size() << endl
        << endl;
   for (size_t i = 0; i < factors_.size(); i++) {
+    const sharedFactor& factor = factors_[i];
+    const double errorValue = (factor != nullptr ? factors_[i]->error(values) : .0);
+    if (!printCondition(factor.get(),errorValue,i))
+      continue; // User-provided filter did not pass
+
     stringstream ss;
     ss << "Factor " << i << ": ";
-    if (factors_[i] == nullptr) {
-      cout << "nullptr" << endl;
+    if (factor == nullptr) {
+      cout << "nullptr" << "\n";
     } else {
-      factors_[i]->print(ss.str(), keyFormatter);
-      cout << "error = " << factors_[i]->error(values) << endl;
+      factor->print(ss.str(), keyFormatter);
+      cout << "error = " << errorValue << "\n";
     }
-    cout << endl;
+    cout << endl; // only one "endl" at end might be faster, \n for each factor
   }
 }
 
--- a/gtsam/nonlinear/NonlinearFactorGraph.h
+++ b/gtsam/nonlinear/NonlinearFactorGraph.h
@@ -103,7 +103,9 @@ namespace gtsam {
 
     /** print errors along with factors*/
     void printErrors(const Values& values, const std::string& str = "NonlinearFactorGraph: ",
-                     const KeyFormatter& keyFormatter = DefaultKeyFormatter) const;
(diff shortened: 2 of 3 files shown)


gtsam/nonlinear/NonlinearFactorGraph.cpp
```diff
@@ -61,19 +61,26 @@ void NonlinearFactorGraph::print(const std::string& str, const KeyFormatter& key
 
 /* ************************************************************************* */
 void NonlinearFactorGraph::printErrors(const Values& values, const std::string& str,
-                                       const KeyFormatter& keyFormatter) const {
+    const KeyFormatter& keyFormatter,
+    const std::function<bool(const Factor* /*factor*/, double /*whitenedError*/, size_t /*index*/)>& printCondition) const
```
ProfFan, line 65: This should be optional

gtsam/nonlinear/NonlinearFactorGraph.cpp
```diff
@@ -61,19 +61,26 @@ void NonlinearFactorGraph::print(const std::string& str, const KeyFormatter& key
...
+{
   cout << str << "size: " << size() << endl
        << endl;
   for (size_t i = 0; i < factors_.size(); i++) {
+    const sharedFactor& factor = factors_[i];
+    const double errorValue = (factor != nullptr ? factors_[i]->error(values) : .0);
+    if (!printCondition(factor.get(),errorValue,i))
```
ProfFan, line 72: Check if this function exists

gtsam/nonlinear/NonlinearFactorGraph.cpp
jlblancoc: Actually, that was my first implementation, using `boost::optional`. But it had problems to automatically bind to a lambda as input.
Will give it another try...

Thanks!

ProfFan: Actually I think simple overloading here makes perfect sense :)

ProfFan: BTW, this https://godbolt.org/z/aSio9A also work

jlblancoc:
@ProfFan I added a unit test and rewrite the interface to use `boost:optional`.
Note that I also had to define a type for the `std::function<>` so callers can use it to first convert lambdas to that type. Otherwise, it seems `boost::optional<>` fails to do the conversion (as far as I tested it, at least!).

ProfFan:
@jlblancoc Have you looked at https://godbolt.org/z/aSio9A ? I think because we are targeting C++11 a simple default lambda would work perfectly

jlblancoc:
@ProfFan 
>  Have you looked at https://godbolt.org/z/aSio9A ? I think because we are targeting C++11 a simple default lambda would work perfectly

Yes... But I can't see the difference with my first proposal in:

https://github.com/borglab/gtsam/commit/07b66eae38fcbfd190792d9a6829244fa07536be#diff-f1690c2d78faec90bcbbbbd239b87347R107-R108 

It was a `std::function` as argument, with a lambda in the header as default parameter. And I understood that "optional" was preferred? Perhaps [your comment](https://github.com/borglab/gtsam/pull/274#discussion_r405098302) missed that the default parameter was already present in the header? :-)

ProfFan:
@jlblancoc Yep I apologize...

jlblancoc:
:+1: No problem. 
I just updated it again to use `std::function<>`. I like it more, too ;-)

Feel free of merging when Travis goes over it...

jlblancoc:
Travis... (sigh)

``` 
+wget https://github.com/oneapi-src/oneTBB/releases/download/4.4.2/tbb44_20151115oss_lin.tgz -O /tmp/tbb.tgz
--2020-04-10 14:28:37--  https://github.com/oneapi-src/oneTBB/releases/download/4.4.2/tbb44_20151115oss_lin.tgz
Resolving github.com (github.com)... 140.82.114.4
Connecting to github.com (github.com)|140.82.114.4|:443... connected.
HTTP request sent, awaiting response... 429 too many requests
2020-04-10 14:28:37 ERROR 429: too many requests.
```

PR #277, opened 2020-04-08
Auto-install Python Package

varunagrawal:
New `make python-install` command to automatically install the python wrapper.

The installation flow is as simple as

```sh
cmake ..
make -j4
make python-install
```

The major update is that instead of copying all the build artifacts to a 3rd party location (e.g. `/usr/loca/cython`), we instead store the build artifacts under the cmake `build/cython` directory (with build tags getting added as postfixes by default).

Running `make` now builds all the wrapper artifacts directly and in place, removing the need to run `make install`.

The `make python-install` command then installs the build artifacts by running the `setup.py` making installation a one-line command. Since we no longer rely on `make install`, the `make python-install` command can directly handle all the dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/277)


varunagrawal:
@jlblancoc I may need your help on this PR. I need to add corresponding python wrapper installation instructions in the `cython/scripts/install.bat` file for Windows and you are the expert on this.

jlblancoc:
> @jlblancoc I may need your help on this PR. I need to add corresponding python wrapper installation instructions in the `cython/scripts/install.bat` file for Windows and you are the expert on this.

Hmmm... what about a cmake script instead, so it's common to both Windows and Linux?

varunagrawal:
I actually tried a cmake script first but running with superuser became problematic and it wouldn't work consistently.
However, I've actually learned some new tricks since then so let me try again.

varunagrawal:
Okay that worked beautifully for me on Linux. Thank you @jlblancoc!

varunagrawal:
This should be ready for review now.

ProfFan:
@varunagrawal have you tried installing and test on all normal configurations (Debug, Release, RelWithDebInfo)?

varunagrawal:
I tested with Release and Debug. Testing with RelWithDebugInfo now.

varunagrawal:
So I discovered a cleaner way to do this that only requires `make install`.

ProfFan:
@varunagrawal That is awesome! But definitely make that optional (off-by-default), so we don't get accidental installs

varunagrawal:
@ProfFan the python wrapper install only happens if `GTSAM_INSTALL_CYTHON_TOOLBOX=ON`, since the install function is defined inside the that conditional. Thus, it should be off by default.

Another thing, I noticed that the install of the `gtsam_eigency` happens after the `setup.py` is run, so I converted that into a function that we have more control over. I couldn't find another way of doing it unfortunately because CMake doesn't support setting up dependencies on install targets.

varunagrawal:
As a follow up, this works with Release, Debug and RelWithDebugInfo. I plan to update the CI with another PR as soon as this is merged in.

ProfFan:
@varunagrawal I think that flag controls whether to "build" the package, not installing it. I'd rather have a separate option that says `AUTO_INSTALL` or something, as this is indeed a breaking change.

cython/gtsam_eigency/CMakeLists.txt
```diff
@@ -38,12 +38,14 @@ add_custom_target(cythonize_eigency)
...
+  install(DIRECTORY ${source_directory}
+          DESTINATION "${GTSAM_CYTHON_INSTALL_FULLPATH}"
+          PATTERN "CMakeLists.txt" EXCLUDE
+          PATTERN "__init__.py.in" EXCLUDE)
+  install(TARGETS cythonize_eigency_core cythonize_eigency_conversions
+          DESTINATION "${GTSAM_CYTHON_INSTALL_FULLPATH}/gtsam_eigency")
+  install(FILES ${source_directory}/conversions_api.h DESTINATION ${GTSAM_CYTHON_INSTALL_FULLPATH}/gtsam_eigency)
+  install(FILES ${source_directory}/__init__.py DESTINATION ${GTSAM_CYTHON_INSTALL_FULLPATH}/gtsam_eigency)
+
+endfunction()
```
ProfFan, line 51: newline

cython/CMakeLists.txt
```diff
@@ -45,4 +44,12 @@ if (GTSAM_INSTALL_CYTHON_TOOLBOX)
   install_cython_scripts("${PROJECT_SOURCE_DIR}/cython/gtsam" "${GTSAM_CYTHON_INSTALL_PATH}" "*.py")
   install_cython_scripts("${PROJECT_SOURCE_DIR}/cython/gtsam_unstable" "${GTSAM_CYTHON_INSTALL_PATH}" "*.py")
 
+  # Adding custom function here so that gtsam_eigency is installed before
```
ProfFan, line 47: I really think that this should be a separate target than the normal cython build target, so to install people can do `make cython_install` or something, and it's easier and clearer if you want to add a option to auto install.

varunagrawal: So this only runs if people do `make install`. The build step is completely handled via `make`.

dellaert: So, I thought that we agreed per our discussion that this should *not* run if people make install. But install should run if `make cython_install` is called and build is out of date.

varunagrawal:
@ProfFan I updated the PR description to reflect the new style of how the installation works. This should address all the concerns with respect to prefixes and packaging requirements.

cython/README.md
```diff
@@ -1,33 +1,36 @@
...
+
+  ```bash
+  export PYTHONPATH=$PYTHONPATH:<GTSAM_CYTHON_INSTALL_PATH>
+  ```
+
+- Build GTSAM and the wrapper with `make`.
+
+- To install system-wide, simply run `make install`.
+    - The same command can be used to install into a virtual environment if it is active.
+    - **NOTE**: if you don't want gtsam to install to a system directory such as `/usr/local`, pass `-DCMAKE_INSTALL_PREFIX="./install"` to cmake to install gtsam to a subdirectory of the build directory.
```
ProfFan, line 32: Awesome, but still the concern: this PR will always **globally** install the Cython wrapper, regardless of the prefix specified, when running `make install`?

cython/CMakeLists.txt
varunagrawal: Currently that is not possible since CMake doesn't support adding dependencies on install commands. That's an open issue on the CMake repository.

Using make install allows CMake to manage the outdated dependencies directly this way, while also truly installing the python package. It didn't make sense to me to have make install copy to an install path and then a separate install package step.

If you would rather have the separate install command, then we can't have CMake manage the build dependencies at the moment.

cython/README.md
varunagrawal: Actually that's a mistake in the docs. The concept of the install prefix for the python wrapper is no longer valid since it builds everything in the CMake build directory. This makes everything self contained, and the install command installs the python package to the python namespace.

ProfFan: I still insist that `make install` should not touch anything beyond the prefix, including the current `site-packages` directory. This is a common practice and we should not break it.

cython/CMakeLists.txt
dellaert: Yes, I don’t think we should change behavior, and I think there should be a separate command for invoking python setup

cython/README.md
dellaert: First of all, I always use a prefix, and it works like a charm, so I understand the documentation to be correct. And indeed, the fact that cython is installed outside the prefix as Fan mentions, does seem like a breach of contract. We should probably change the default to something inside the prefix.

dellaert: By the way, Varun, your use of prefix seems unconventional. I always use my home directory as the prefix, so I have an include and lib folder there.

varunagrawal: I was unclear about what I meant in my previous comment. That was *my* mistake in the docs. The way this PR does this is that cython is no longer installed to the prefix location, so the notion of a prefix is no longer applicable. Everything happens inside the CMake build directory.

And when I mentioned the package, I mean the result of running python setup.py install which is technically what installs the GTSAM package and makes it available via python.

Thanks to this PR, the only things that depend on the prefix are the C++ libraries and the Matlab wrapper.

cython/CMakeLists.txt
varunagrawal: In that case then we have no way of enforcing CMake to rebuild  when something happens.

I really don't understand why this is a problem. Currently the status quo is that we do make install and then we have to manually go to the cython folder to run the setup file. All I did was automate that last step as part of the make install process, so there's nothing changed from the previous way of doing things. We didn't have an extra command then and we don't have it now.

dellaert: Maybe we should chat in person - python install is frought with issue, and I think making it part of "make install" will create more problems than it will solve.

cython/README.md
ProfFan: I do think in line with Varun that the cython wrapper should not be installed to the prefix in its current subdirectory - it should be inside `$PREFIX/share/gtsam/cython` or something, or **NOT** installed at all (only reside in build directory), but I do **NOT** think that `make install` should install anything to the **system** (`site-packages`, etc.), either.

varunagrawal: I actually prefer the **NOT** installed at all option. I am still being mindful of build postfixes so that shouldn't be a problem.

cython/CMakeLists.txt
varunagrawal: So after a good(?) night's sleep, I actually figured out a way to accomplish the exact requirement. Have a separate build target yet have it automatically managed by cmake.

The idea is to use the `${PROJECT_BINARY_DIR}/cython` folder directly to hold everything and then `make python-install` will just use the `setup.py` there. This bypasses the whole "install" step, allowing us to achieve what we want. :grin:

varunagrawal:
@ProfFan @dellaert updated this PR and the description. This should satisfy all the requirements. :slightly_smiling_face:

varunagrawal:
Sure thing. Yeah this PR went through a lot of different cycles due to different requirements being pushed in at different times.

It wasn't a problem per se, but I have not been a fan of having to go into a separate directory just to install the wrapper to the python environment.

dellaert:
@varunagrawal this is cython so we should either give up or put in 45.0.3. Please summarize what this PR now does.

dellaert:
@varunagrawal this is the only PR left in 4.0.3

varunagrawal:
This should still be useful for generating packages with the new Pybind11 wrapper.

Moreover, it has a good number of nice fixes to the setup.py and CMake files.

ProfFan:
@varunagrawal I have implemented `python-install` in the pybind branch already (because in the pybind branch and possibly 4.1 `make install` will not copy the python files to the prefix, resulting in a much cleaner install).

I think this can be picked into 4.0.3 after branching out `release_4.0.3`

dellaert:
> This should still be useful for generating packages with the new Pybind11 wrapper.
> 
> Moreover, it has a good number of nice fixes to the setup.py and CMake files.

Please summarize what this PR now does.

varunagrawal:
This PR

1. Gets rid of the `install` step for the python wrapper by building everything into the `cython` directory under the cmake build directory.
2. Adds a new make command `python-install` which generates the python package and installs it via `python setup.py install`.
3. Introduces a bunch of clean ups to the Cython Cmake files and the `setup.py.in` template file.

dellaert:
OK, thanks! @ProfFan this seems identical to what you did in pybind. Please confirm or tell us how it is different, and what needs to change in this PR - if anything. It should be identical - no sense in changing gears twice in a month.

ProfFan:
@dellaert  I believe this could go into the release branch after branching out (since 4.1 will not have Cython stuff)

dellaert:
@ProfFan can't we just merge into develop and then merge all recent changes from develop into the *already existing* 4.0.3 branch?

ProfFan:
@dellaert Yes

PR #279, opened 2020-04-09
Fisheye Calibration

ghaggin:
I created a fisheye camera calibration based off CalDS2 and CalDS2_base.  The fisheye model i used is consistent [this opencv model](https://docs.opencv.org/3.4/db/d58/group__calib3d__fisheye.html).  I also added a test case for the calibration and an example using the structure from motion data.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/279)


gtsam/geometry/Cal3Fisheye_Base.cpp
```diff
@@ -0,0 +1,184 @@
...
+static Matrix2 D2dintrinsic(const double xi, const double yi, const double r, const double td, const double t, const double tt, const double t4, const double t6, const double t8, const double k1, const double k2, const double k3, const double k4,  const Matrix2& DK) {
+  
+  const double dr_dxi = xi/sqrt(xi*xi + yi*yi);
+  const double dr_dyi = yi/sqrt(xi*xi + yi*yi);
+  const double dt_dr = 1/(1 + r*r);
+  const double dtd_dt = 1 + 3*k1*tt + 5*k2*t4 + 7*k3*t6 + 9*k4*t8;
+  const double dtd_dxi = dtd_dt * dt_dr * dr_dxi;
+  const double dtd_dyi = dtd_dt * dt_dr * dr_dyi;
+
+  const double rr = r*r;
```
dellaert, line 84: I would cache (1/r) and (1/rr), alternatively cache xy/r and yi/r, xy/rr, yi/rr

gtsam/geometry/Cal3Fisheye.h
```diff
@@ -0,0 +1,121 @@
...
+namespace gtsam {
+
+/**
+ * @brief Calibration of a fisheye camera that also supports
+ * Lie-group behaviors for optimization.
+ * \sa Cal3Fisheye_Base
+ * @addtogroup geometry
+ * \nosubgrouping
+ */
+class GTSAM_EXPORT Cal3Fisheye : public Cal3Fisheye_Base {
```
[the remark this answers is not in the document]
ghaggin: The Cal3DS2 class was split this way, base class implemented the math and the other class looked like it was just an interface.  I can combine the classes into one if that would be better. Now that I look at it it seems like Cal3DS2 is the only calibration model split up this way, so I could combine those too while I'm at it.  Just let me know what would be better.

ghaggin:
Also, is there a specific formatting tool I should use (like clang-format).  I saw somewhere you said something about a modified K&R style with 2 space tabs but I didn't see anything about autoformatting

gtsam/geometry/Cal3Fisheye.h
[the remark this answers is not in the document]
dellaert: Cal3DS2_Base has two derived classes: Cal3DS2 and Cal3Unified. That's the only reason it's split up :-)

dellaert:
We now do clangformat with Google style as much as possible for new code.

gtsam/geometry/Cal3Fisheye.h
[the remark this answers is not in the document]
ghaggin: That makes sense, I didn't see that.

gtsam/geometry/Cal3Fisheye.h
```diff
@@ -0,0 +1,211 @@
...
+
+  /// @name Standard Constructors
+  /// @{
+
+  /// Default Constructor with only unit focal length
+  Cal3Fisheye()
+      : fx_(1), fy_(1), s_(0), u0_(0), v0_(0), k1_(0), k2_(0), k3_(0), k4_(0) {}
+
+  Cal3Fisheye(const double fx, const double fy, const double s, const double u0,
+              const double v0, const double k1, const double k2,
```
dellaert, line 63: I'm not sure about your appetite on working more on this PR, but I'm wondering whether we really need scalar access, or whether we could just work with a Vector4 all over the place, and get rid of the k1...k4 access etc below.

gtsam/geometry/Cal3Fisheye.h
```diff
@@ -0,0 +1,211 @@
...
+   */
+  Point2 uncalibrate(const Point2& p, OptionalJacobian<2, 9> Dcal = boost::none,
+                     OptionalJacobian<2, 2> Dp = boost::none) const;
+
+  /// Convert (distorted) image coordinates [u;v] to intrinsic coordinates [x_i,
+  /// y_i]
+  Point2 calibrate(const Point2& p, const double tol = 1e-5) const;
+
+  /// Derivative of uncalibrate wrpt intrinsic coordinates [x_i; y_i]
+  Matrix2 D2d_intrinsic(const Point2& p) const;
```
dellaert, line 140: Why do we need this? I think this might be another pattern that had its use in our old code, but it seems is not needed here. It leads to a lot of copy/paste in the cpp code.

gtsam/geometry/Cal3Fisheye.h
```diff
@@ -0,0 +1,211 @@
...
+  /** Serialization function */
+  friend class boost::serialization::access;
+  template <class Archive>
+  void serialize(Archive& ar, const unsigned int /*version*/) {
+    ar& BOOST_SERIALIZATION_NVP(fx_);
+    ar& BOOST_SERIALIZATION_NVP(fy_);
+    ar& BOOST_SERIALIZATION_NVP(s_);
+    ar& BOOST_SERIALIZATION_NVP(u0_);
+    ar& BOOST_SERIALIZATION_NVP(v0_);
+    ar& BOOST_SERIALIZATION_NVP(k1_);
```
dellaert, line 196: I think we can serialize vectors but it would have to be tested

gtsam/geometry/Cal3Fisheye.cpp
```diff
@@ -0,0 +1,216 @@
...
+  const double td = t * (1 + k1_ * tt + k2_ * t4 + k3_ * t6 + k4_ * t8);
+  const double xd = td / r * xi, yd = td / r * yi;
+  Point2 uv(fx_ * xd + s_ * yd + u0_, fy_ * yd + v0_);
+
+  Matrix2 DK;
+  if (H1 || H2) DK << fx_, s_, 0.0, fy_;
+
+  // Derivative for calibration parameters (2 by 9)
+  if (H1)
+    *H1 = D2dcalibration(xd, yd, xi, yi, t * tt, t * t4, t * t6, t * t8, r, DK);
```
dellaert, line 116: If we don't need the separate Jacobian methods I'd inline them here, which would eliminate a lot of copy/paste, and will also allow to identify which variables are needed for evaluation only, and which *extra* things are needed for the derivatives.

dellaert:
I approved but see my comments on the Jacobians above.

PR #285, opened 2020-04-29
Maintenance: Merge develop into new_wrapper

ProfFan:
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/285)


ProfFan:
Judging from the recent activities, I think I need to add a TBB build into the Actions build matrix.

ProfFan:
@dellaert @jlblancoc It seems that this commit 76b29b7 breaks VS2016 (ICE if use nullptr instead of NULL). Should we revert or just drop VS2016 support? I vote for dropping VS2016 and only support 2019 :)

ProfFan:
CI now all pass except Windows 2016 - ready for review

ProfFan:
@dellaert  If no opposition I will remove VS 2016...

dellaert:
No opposition. Also change the readme specifying VC version needed

dellaert:
Why is this a PR?

ProfFan:
@dellaert To make sure that we do not break the Pybind branch

dellaert:
> @dellaert To make sure that we do not break the Pybind branch

You mean, just to run CI? Travis runs on branch changes the moment you push them (remember, no force-push or rebase on public branches!). I think maybe Github actions should be same?

ProfFan:
@dellaert Actually this is the problem I am trying to solve: pushes to feature branches on the repo only runs a core set of the tests, and PRs will run a whole testset of 40 tests. This is to both reduce time for real-time feedback on pushing and increase test coverage for PRs.

PR #288, opened 2020-05-03
Representation for lines in 3D - Line3

akshay-krishnan:
Introducing the Line3 class in geometry. 

It can be used to represent lines in 3D. The representation used is from [Schindler et al](https://www.cc.gatech.edu/projects/4d-cities/dhtml/pubs/Schindler06-3dpvt.pdf), originally introduced by [Taylor and Kriegman](https://www.cis.upenn.edu/~cjtaylor/PUBLICATIONS/pdfs/TaylorTR94b.pdf). 

Each line is represented as a Rotation matrix and 2 scalars (R, a, b) with a 4-dimensional incremental update. This representation of a line can also be transformed from one coordinate frame to another and projected to the image plane. 

Math.pdf in gtsam/doc has been updated with more details about the line representation, the equations for transforming and projecting a line, along with the jacobians for these functions.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/288)


gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+ * @file    Line3.h
+ * @brief   4 dimensional manifold of 3D lines
+ * @author  Akshay Krishnan
+ * @author  Frank Dellaert
+ */
+// \callgraph
+
+#pragma once
+
+#include <gtsam/base/concepts.h>
```
dellaert, line 19: Please make sure all headers are *really* necessary. Which can be omitted? Which can be moved to .cpp?

Also: geometry subdir cannot refer to either nonlinear or slam directories. Think of geometry as living closer to the metal.

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * A 3D line (R,a,b) : (Rot3,Scalar,Scalar)
+ * @addtogroup geometry
+ * \nosubgrouping
+ */
+class Line3 {
+private:
+    Rot3 R_;    // Rotation of line about x and y in world frame
+    double a_, b_;  // Intersection of line with x-y world plane
```
dellaert, line 39: Is this true? I thought it was with *rotated* x-y plane

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+
+public:
+    enum {dimension = 4};
+
+    /** Default constructor is the Z axis **/
+    Line3() :
```
dellaert, line 45: Did you format with clangformat/Google style?

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+            a_(0), b_(0)
+    {}
+
+    /** Constructor:
+     * Parallel to z axis, intersecting x-y plane at (a,b) **/
+    Line3(const double a, const double b):
```
dellaert, line 51: Why do we have this?

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+
+    /** Constructor for general line from (R, a, b) **/
+    Line3(const Rot3 &R, const double a, const double b) :
+            R_(R), a_(a), b_(b)
+    {}
+
+    /** Sample two points lying on the line
+     * @return Vector6 where first 3 elements are one point, and
+     * next 3 elements are another point **/
+    Vector6 points();
```
dellaert, line 63: Make all const methods `const`. Are you sure this method works for *all* lines? It's a bit of a weird method to have, BTW: why do we need it?

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+     * @param H OptionalJacobian of localCoordinates with respect to line
+     * @return Vector4 difference in the tangent space
+     */
+    Vector4 localCoordinates(const Line3 &q, OptionalJacobian<4,4> H = boost::none) const;
+
+    /**
+     * Rotation of line accessor
+     * @return Rot3
+     */
+    Rot3 R() const
```
dellaert, line 88: bad name, and breaks encapsulation. Do we really need this?

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+    {
+        return R_;
+    }
+
+    /**
+     * Accessor for a, b
+     * @return Vector2(a, b)
+     */
+    Vector2 V() const
```
dellaert, line 97: same

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+     * @param l2 - line to be compared
+     * @param tol : optional tolerance
+     * @return boolean - true if lines are equal
+     */
+    bool equals(const Line3 &l2, double tol = 10e-9) const;
+
+    /**
+     * Projecting a line to an image plane
+     * @param Dline: OptionalJacobian of projected line with respect to this line
+     * @return Point3 - projected line in image plane
```
dellaert, line 119: I don't understand. Why is this a Point3? Properly comment :-)

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+};
+
+template <>
+struct traits<Line3> : public internal::Manifold<Line3> {};
+
+template <>
+struct traits<const Line3> : public internal::Manifold<Line3> {};
+
+// For expression factor graphs
+typedef Expression<Line3> Line3_;
```
dellaert, line 131: Not here. Maybe in #include <gtsam/geometry/expressions.h>

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Function projects input line to image plane
+ * @param L input line
+ * @param Dline OptionalJacbian of projected line with respect to input line
+ * @return Point3 - projected line
+ */
+Point3 projectLine(const Line3 &L,
```
dellaert, line 139: Why ? We already have Line3::project?

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Transform a line from world to camera frame
+ * @param p - Pose3 of camera in world frame
+ * @param l - Line3 in world frame
+ * @param Dpose - OptionalJacobian of transformed line with respect to p
+ * @param Dline -  OptionalJacobian of transformed line with respect to l
+ * @return Transformed line in camera frame
+ */
+Line3 transformTo(const Pose3 &p, const Line3 &l,
```
dellaert, line 150: Can you use dTs naming convention for all poses T and rotation matrices R? Where d = destination and s = source, such that dPoint = dTs * sPoint.

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+                    OptionalJacobian<4, 6> Dpose = boost::none,
+                    OptionalJacobian<4, 4> Dline = boost::none);
+
+/**
+ * Expression method for projection
+ * @param line Line3_ expression
+ * @return Point3_ projected line
+ */
+inline Point3_ projectLineExpression(const Line3_& line)
```
dellaert, line 159: Not here. It's easy to make an expression straight from the method declaration. I think something akin to

```
Point3_ project_(line_, &Line3::project);
```

should work, hence this stuff here is redundant. If it goes anywhere, probably in geometry/expressions.h

gtsam/geometry/Line3.h
```diff
@@ -0,0 +1,178 @@
...
+    return Point3_(f, line);
+}
+
+/**
+ * Expression method for transforming line
+ * @param p - Pose3_ expression
+ * @param line - Line3 expression
+ * @return transformed Line3_ expression
+ */
+inline Line3_ transformToExpression(const Pose3_& p, const Line3_& line)
```
dellaert, line 171: same

gtsam/geometry/Line3.h
akshay-krishnan: It is the 2D line in the image plane in homogenous coordinates. It could be a Vec3, Point3 or Unit3. Could you tell me which one is to be preferred and why?

dellaert: It should be between Vector3 and Unit3, as Point3 really has the "I'm a point in 3D" name :-) Unit3 is in a sense more correct as it has the correct manifold dimension, but it will incur a little bit more computation (as it normalizes). Then again, the sphere and P^3 (the projective space of lines are *not* the same. Still, Unit3 is a bit like a Quaternion representation for 2D lines, so I think it's the best we have. 

Part of the comment is also about documentation, though: say "2D line in the image plane in homogenous coordinates. We use Unit3 as the return type as that is a manifold of the right dimension."

gtsam/geometry/Line3.h
akshay-krishnan: It was a convenience function that was helpful for some tests. We could also simply use the general constructor. Would you recommend removing it?

gtsam/geometry/Line3.h
akshay-krishnan: Yes, it should work for all lines. It is helpful for visualization. But I can move it to the visualization code, as it is not directly related to the line itself.

gtsam/geometry/Line3.h
dellaert: Yes

gtsam/geometry/Line3.h
akshay-krishnan: Yes, it is a necessary accessor. It is used in functions like transformTo, visualization, computing metrics for estimation etc. Would rotation() be a good name?

gtsam/geometry/Line3.h
dellaert: We are defining an API here :-) resist convenience for testing in the API.  However, recovering points from a line might be handy, but we should think about what the API should be - related to the math. For example, I can see a method “point” with a single scalar argument that returns the point where the line pierces the rotated XY plane, and positive scalars go towards positive Z. Easy to implement as R*(0,0,s) ?

gtsam/geometry/Line3.h
dellaert: Maybe make those methods friends.

gtsam/geometry/Line3.h
akshay-krishnan: Same as above, the accessor is necessary. 
Alternative names: translation() might be deceiving, but maybe closestPoint() or just point() is better?

dellaert: Some, make some friends. I really want to get rid of those accessors, as they break encapsulation. We should be able to change the representation of the line without anyone (except friends, who asked for it), noticing.

gtsam/geometry/Line3.h
akshay-krishnan: This might have some effect on my reconstruction. The returned optionalJacobian would change from having a shape <3,4> to <2,4>. I will try it out.

dellaert: Indeed! And very much correctly so :-) 2D lines *are* only 2-dimensional.

dellaert:
@akshay-krishnan request a new review whenever you're ready.

gtsam/geometry/Line3.h
akshay-krishnan: Changed my formatting to Google style, but with 4-space indents instead of 2. Should I keep 2 space indents?

dellaert: What is default for Google?
GTSAM uses 2-space indents everywhere else.

gtsam/geometry/Line3.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+    Rot3 Rt = R_ * eps;
+    return Line3(Rt, a_ + v[2], b_ + v[3]);
+}
+
+Vector4 Line3::localCoordinates(const Line3 &q, OptionalJacobian<4, 4> H) const {
+    Vector3 local_rot;
+    Vector4 local;
+    if (H) {
+        Eigen::Matrix3d Dw_mat = Eigen::Matrix3d::Zero();
```
dellaert: same comments as above

gtsam/geometry/Line3.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+Unit3 Line3::project(OptionalJacobian<2, 4> Dline) const {
+    Vector3 V_0;
+    V_0 << -b_, a_, 0.0;
+
+    Unit3 l;
+    if (Dline) {
+        // Jacobian of the normalized Unit3 projected line with respect to
+        // un-normalized Vector3 projected line in homogeneous coordinates.
+        Eigen::Matrix<double, 2, 3> D_mat = Eigen::Matrix<double, 2, 3>::Zero();
```
dellaert: Similar comments. Just say `Matrix23 D_unit_line`, no need to init.

gtsam/geometry/Line3.cpp
```diff
@@ -0,0 +1,118 @@
...
+        l = Unit3::FromPoint3(Point3(R_ * V_0), D_unit_line);
+        // Jacobian of the un-normalized Vector3 line with respect to
+        // input 3D line
+        Eigen::Matrix<double, 3, 4> D_vec_line = Eigen::Matrix<double, 3, 4>::Zero();
+        D_vec_line.col(0) = a_ * R_.r3();
+        D_vec_line.col(1) = b_ * R_.r3();
+        D_vec_line.col(2) = R_.r2();
+        D_vec_line.col(3) = -R_.r1();
+        // Jacobian of output wrt input is the product of the two.
+        Eigen::Matrix<double, 2, 4> Dline_mat = (*D_unit_line) * D_vec_line;
```
dellaert: Just 

```
        *Dline = (*D_unit_line) * D_vec_line;
```

dellaert:
Hmmm, you seemed to have pushed some changes while I was reviewing, but the comments still apply. Basically, OptionalJacobian is a a very magical type.

H is either false, or you can use *H just as any other dense matrix. You need to write all its elements. 

When you call a method, just allocate but do not initialize a matrix of right size. It should come back fully filled, no need to init.

Finally, passing blocks to methods for the Jacobian argument should work as well.

gtsam/geometry/Line3.cpp
```diff
@@ -0,0 +1,118 @@
...
+    Vector3 w;
+    w << v[0], v[1], 0;
+    Rot3 eps;
+    if (H) {
+        Eigen::Matrix3d Dw_mat = Eigen::Matrix3d::Zero();
+        OptionalJacobian<3, 3> Dw(Dw_mat);
+        Dw->setZero();
+        eps = Rot3::Expmap(w, Dw);
+        H->block<2, 2>(0, 0) = Dw->block<2, 2>(0, 0);
+        (*H)(2, 2) = 1;
```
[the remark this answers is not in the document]
akshay-krishnan: I could not do H->topLeft<2,2>(), because Eigen::Map has no member topLeft. 
I also tried `eps = Rot3::Expmap(w, H->block<2, 2>(0, 0));`. It complains that there's no viable conversion from Eigen::block to OptionalJacobian. 
Apart from this, doesn't Rot3::Expmap expect Matrix3 for the Jacobian argument? 
What I am trying to do is use the <2,2> topLeft of this Matrix3. Would it be possible to directly pass a <2,2> matrix or block to Rot3::Expmap? 
So I'm considering doing this: 
`Matrix3 Dw;
eps = Rot3::Expmap(w, Dw);
H->setIdentity();
H->block<2, 2>(0, 0) = Dw.block<2, 2>(0, 0);`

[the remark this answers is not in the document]
dellaert: Yes, that works. Ps, when I see `eps` I think scalar. I propose you use `incR`.

gtsam/geometry/Line3.cpp
```diff
@@ -5,37 +5,31 @@ namespace gtsam {
...
+    Matrix3 Dw;
+    localR = Rot3::Logmap(R_.inverse() * q.R_, Dw);
+    H->setIdentity();
+    H->block<2, 2>(0, 0) = Dw.block<2, 2>(0, 0);
   } else {
-    local_rot = Rot3::Logmap(R_.inverse() * q.R_);
+    localR = Rot3::Logmap(R_.inverse() * q.R_);
   }
-  local << local_rot[0], local_rot[1], q.a_ - a_, q.b_ - b_;
+  local << localR[0], localR[1], q.a_ - a_, q.b_ - b_;
```
dellaert, line 32: Declare local here.

gtsam/geometry/Line3.cpp
```diff
@@ -5,37 +5,31 @@ namespace gtsam {
...
+    incR = Rot3::Expmap(w);
   }
-  Rot3 Rt = R_ * eps;
+  Rot3 Rt = R_ * incR;
   return Line3(Rt, a_ + v[2], b_ + v[3]);
 }
 
 Vector4 Line3::localCoordinates(const Line3 &q, OptionalJacobian<4, 4> H) const {
-  Vector3 local_rot;
+  Vector3 localR;
```
dellaert, line 22: `localR` -> `omega`, agrees more with convention.

akshay-krishnan:
> Awesome!
> Except, still need jacobians of retract and localCoordinate tested :-)

I think I will need to change my retract and localCoordinate definitions to include the Jacobians with respect to *this line. I realized this when I was trying to create an expression factor with these methods but the [definition for the unary method expression](https://github.com/borglab/gtsam/blob/fbece35715d6c6eed8d98fd5f62fdbb205dbdcfa/gtsam/nonlinear/Expression.h#L127) requires the unary method to have 2 OptionalJacobian arguments.

akshay-krishnan:
> Awesome!
> Except, still need jacobians of retract and localCoordinate tested :-)

I think I will need to change the definitions of retract and localCoordinates to include the OptionalJacobians of *this line. I realized this when I was trying to create an expression factor using these methods but the [definition of the unary method expression](https://github.com/borglab/gtsam/blob/fbece35715d6c6eed8d98fd5f62fdbb205dbdcfa/gtsam/nonlinear/Expression.h#L127) requires the method to have two OptionalJacobians.

akshay-krishnan:
> Awesome. Do you want to also add an example in this PR, or in next PR?

Thank you @dellaert. 
I think it might be better to address the example in a separate PR.

PR #305, opened 2020-05-09
Feature/examples

dellaert:
After merging Fisheye PR, I cleaned up FisheyeExample and many other examples. Mainly, I used `auto` for the shared noiseModel pointers, and did some formatting to get rid of many lint errors.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/305)

### DIFF
--- a/examples/FisheyeExample.cpp
+++ b/examples/FisheyeExample.cpp
@@ -12,7 +12,7 @@
 /**
  * @file    FisheyeExample.cpp
  * @brief   A visualSLAM example for the structure-from-motion problem on a
- * simulated dataset. This version uses a fisheye camera model and a  GaussNewton
+ * simulated dataset. This version uses a fisheye camera model and a GaussNewton
  * solver to solve the graph in one batch
  * @author  ghaggin
  * @Date    Apr 9,2020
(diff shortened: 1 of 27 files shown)


examples/FisheyeExample.cpp
```diff
@@ -55,75 +55,76 @@
...
-        (Vector(6) << Vector3::Constant(0.1), Vector3::Constant(0.3)).finished());
-    graph.emplace_shared<PriorFactor<Pose3>>(Symbol('x', 0), poses[0], kPosePrior);
-
-    // Add a prior on landmark l0
-    static auto kPointPrior = noiseModel::Isotropic::Sigma(3, 0.1);
-    graph.emplace_shared<PriorFactor<Point3>>(Symbol('l', 0), points[0], kPointPrior);
+/* ************************************************************************* */
+int main(int argc, char *argv[]) {
+  // Define the camera calibration parameters
+  auto K = boost::make_shared<Cal3Fisheye>(
```
ProfFan, line 64: Can we use std::make_shared?

dellaert:
Thanks ! I opted to only format the main function in most files to keep the disruption to the repo to a minimum. I in fact only changed files where the "auto" change was made. Hence, the "using" change is for another PR.

PR #312, opened 2020-05-17
Include corrupting noise param to parse3DFactors

binitshah:
This PR modifies the parse3DFactors method in dataset.cpp to include a `corruptingNoise` parameter that adds a noise to the parsed factors.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/312)

### DIFF
--- a/gtsam/slam/dataset.cpp
+++ b/gtsam/slam/dataset.cpp
@@ -546,13 +546,6 @@ BetweenFactorPose3s parse3DFactors(const string& filename,
   ifstream is(filename.c_str());
   if (!is) throw invalid_argument("parse3DFactors: can not find file " + filename);
 
-  // If asked, create a sampler with random number generator
-  Sampler sampler;
-  if (corruptingNoise) {
-    sampler = Sampler(corruptingNoise);
-  }
-
-
   std::vector<BetweenFactor<Pose3>::shared_ptr> factors;
   while (!is.eof()) {
     char buf[LINESIZE];
@@ -595,6 +588,7 @@ BetweenFactorPose3s parse3DFactors(const string& filename,
       SharedNoiseModel model = noiseModel::Gaussian::Information(mgtsam);
       auto R12 = Rot3::Quaternion(qw, qx, qy, qz);
       if (corruptingNoise) {
+        Sampler sampler(corruptingNoise);
         R12 = R12.retract(sampler.sample());
       }
 
--- a/gtsam/slam/dataset.cpp
+++ b/gtsam/slam/dataset.cpp
@@ -37,6 +37,7 @@
 #include <boost/assign/list_inserter.hpp>
 #include <boost/filesystem/operations.hpp>
 #include <boost/filesystem/path.hpp>
+#include <boost/optional.hpp>
 
 #include <cmath>
 #include <fstream>
@@ -546,6 +547,11 @@ BetweenFactorPose3s parse3DFactors(const string& filename,
   ifstream is(filename.c_str());
   if (!is) throw invalid_argument("parse3DFactors: can not find file " + filename);
 
+  boost::optional<Sampler> sampler;
+  if (corruptingNoise) {
+    sampler = Sampler(corruptingNoise);
+  }
+
   std::vector<BetweenFactor<Pose3>::shared_ptr> factors;
   while (!is.eof()) {
     char buf[LINESIZE];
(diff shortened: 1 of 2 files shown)


gtsam/slam/dataset.cpp
```diff
@@ -541,10 +541,18 @@ std::map<Key, Pose3> parse3DPoses(const string& filename) {
...
+BetweenFactorPose3s parse3DFactors(const string& filename, 
+    const noiseModel::Diagonal::shared_ptr& corruptingNoise) {
   ifstream is(filename.c_str());
   if (!is) throw invalid_argument("parse3DFactors: can not find file " + filename);
 
+  // If asked, create a sampler with random number generator
+  Sampler sampler;
+  if (corruptingNoise) {
+    sampler = Sampler(corruptingNoise);
+  }
```
akshay-krishnan, line 553: @binitshah This is first constructing a sampler object, and then copying a new sampler object `Sampler(corruptingNoise)` to it. Build fails because there is no default constructor defined for Sampler, and so we cannot do `Sampler sampler;`. Could you consider removing these lines and instantiating the Sampler object before line 598 where it is being used?

gtsam/slam/dataset.cpp
```diff
@@ -585,8 +593,13 @@ BetweenFactorPose3s parse3DFactors(const string& filename) {
       mgtsam.block<3, 3>(3, 0) = m.block<3, 3>(3, 0);  // off diagonal
 
       SharedNoiseModel model = noiseModel::Gaussian::Information(mgtsam);
+      auto R12 = Rot3::Quaternion(qw, qx, qy, qz);
+      if (corruptingNoise) {
+        R12 = R12.retract(sampler.sample());
```
akshay-krishnan, line 598: We can just do `Sampler sampler(corruptingNoise);` immediately before this line, right?

gtsam/slam/dataset.h
```diff
@@ -159,7 +159,8 @@ GTSAM_EXPORT void writeG2o(const NonlinearFactorGraph& graph,
 
 /// Parse edges in 3D TORO graph file into a set of BetweenFactors.
 using BetweenFactorPose3s = std::vector<gtsam::BetweenFactor<Pose3>::shared_ptr>;
-GTSAM_EXPORT BetweenFactorPose3s parse3DFactors(const std::string& filename);
+GTSAM_EXPORT BetweenFactorPose3s parse3DFactors(const std::string& filename, 
+    const noiseModel::Diagonal::shared_ptr& corruptingNoise=nullptr);
```
akshay-krishnan, line 163: This is not related to the build failure but other functions on GTSAM use `boost::none` instead of `nullptr`. Is there a reason why we prefer `nullptr` here?

dellaert:
Please note that in Jings branch it does compile, so there must be something else going on as well. I prefer if you first to resolve it by making sure we have the correct file from Jing’s branch, and then see what is still wrong. I will comment more deeply after that is done. Pleasse rerequest a review by me at the time. You can just push to this PR with updates

akshay-krishnan:
> Please note that in Jings branch it does compile, so there must be something else going on as well. I prefer if you first to resolve it by making sure we have the correct file from Jing’s branch, and then see what is still wrong. I will comment more deeply after that is done. Pleasse rerequest a review by me at the time. You can just push to this PR with updates

Did Jing's branch compile on Travis CI? 
There is a default constructor defined for Sampler if GTSAM_ALLOW_DEPRECATED_SINCE_V4 is ON. @binitshah 's branch builds locally for me since this is ON by default. It is, however, set to OFF on Travis CI, which I believe causes the compilation error.

dellaert:
> > Please note that in Jings branch it does compile, so there must be something else going on as well. I prefer if you first to resolve it by making sure we have the correct file from Jing’s branch, and then see what is still wrong. I will comment more deeply after that is done. Pleasse rerequest a review by me at the time. You can just push to this PR with updates
> 
> Did Jing's branch compile on Travis CI?
> There is a default constructor defined for Sampler if GTSAM_ALLOW_DEPRECATED_SINCE_V4 is ON. @binitshah 's branch builds locally for me since this is ON by default. It is, however, set to OFF on Travis CI, which I believe causes the compilation error.

Aha. Yes, then fix it :-)

gtsam/slam/dataset.cpp
```diff
@@ -585,8 +586,14 @@ BetweenFactorPose3s parse3DFactors(const string& filename) {
       mgtsam.block<3, 3>(3, 0) = m.block<3, 3>(3, 0);  // off diagonal
 
       SharedNoiseModel model = noiseModel::Gaussian::Information(mgtsam);
+      auto R12 = Rot3::Quaternion(qw, qx, qy, qz);
+      if (corruptingNoise) {
+        Sampler sampler(corruptingNoise);
```
dellaert, line 591: No, please move it out of the loop at least?? You can make an optional Sampler, and only sample if it evaluates to True here

gtsam/slam/dataset.h
dellaert: boost::none is only used for OptionalJacobian

PR #313, opened 2020-05-18
Demangle the type when printing

### LINKED ISSUE #311: Demangle Type Info
## Feature

Currently GTSAM outputs type information like this

```
N5gtsam12GenericValueIN5Eigen6MatrixIdLi4ELi1ELi0ELi4ELi1EEEEE
```

which makes debugging hard.

It would be great to have GTSAM print type info like this instead

```
gtsam::GenericValue<Eigen::Matrix<double, 4, 1, 0, 4, 1> >
```

which is actually quite easily doable.

## Motivation

Make understanding and debugging much, much easier.

## Pitch

This feature is very straightforward. The code to enact this change can be found [here](https://stackoverflow.com/a/4541470/1236990). I have tried it out on a local version of GTSAM and it works great.

I was originally getting this error message:

```
Failure: "Exception: Attempting to retrieve value with key "X0", 
type stored in Values is N5gtsam12GenericValueIN5Eigen6MatrixIdLi4ELi1ELi0ELi4ELi1EEEEE 
but requested type was N5Eigen6MatrixIdLin1ELin1ELi0ELin1ELin1EEE" 
```

and now I get this

```
Failure: "Exception: Attempting to retrieve value with key "X0", 
type stored in Values is gtsam::GenericValue<Eigen::Matrix<double, 4, 1, 0, 4, 1> > 
but requested type was Eigen::Matrix<double, -1, -1, 0, -1, -1>" 
```

which makes it clear immediately what the issue is.

## Alternatives

N/A

varunagrawal:
This PR fixes #311.

I expect to get quite a bit of feedback regarding this.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/313)

### DIFF
--- a/gtsam/base/GenericValue.h
+++ b/gtsam/base/GenericValue.h
@@ -20,6 +20,7 @@
 #pragma once
 
 #include <gtsam/base/Manifold.h>
+#include <gtsam/base/types.h>
 #include <gtsam/base/Value.h>
 
 #include <boost/make_shared.hpp>
@@ -30,33 +31,6 @@
 #include <typeinfo> // operator typeid
 
 namespace gtsam {
-
-#ifdef __GNUG__
-#include <cstdlib>
-#include <cxxabi.h>
-#include <string>
-
-/// Pretty print Value type name
-static std::string demangle(const char* name) {
-  int status = -4; // some arbitrary value to eliminate the compiler warning
-  // enable c++11 by passing the flag -std=c++11 to g++
-  std::unique_ptr<char, void(*)(void*)> res {
-      abi::__cxa_demangle(name, NULL, NULL, &status),
-      std::free
-  };
-  return (status==0) ? res.get() : name ;
-}
-
-#else
-
-// does nothing if not g++
-static std::string demangle(const char* name) {
-    return name;
-}
-
-#endif
-
-
 
 /**
  * Wraps any type T so it can play as a Value
--- a/gtsam/base/GenericValue.h
+++ b/gtsam/base/GenericValue.h
@@ -31,6 +31,33 @@
 
 namespace gtsam {
 
+#ifdef __GNUG__
+#include <cstdlib>
+#include <cxxabi.h>
+#include <string>
+
+/// Pretty print Value type name
+static std::string demangle(const char* name) {
+  int status = -4; // some arbitrary value to eliminate the compiler warning
+  // enable c++11 by passing the flag -std=c++11 to g++
+  std::unique_ptr<char, void(*)(void*)> res {
+      abi::__cxa_demangle(name, NULL, NULL, &status),
+      std::free
+  };
+  return (status==0) ? res.get() : name ;
+}
+
+#else
+
+// does nothing if not g++
+static std::string demangle(const char* name) {
+    return name;
+}
+
+#endif
+
+
+
 /**
  * Wraps any type T so it can play as a Value
  */
@@ -83,7 +110,7 @@ public:
 
   /// Virtual print function, uses traits
   virtual void print(const std::string& str) const {
-    std::cout << "(" << typeid(T).name() << ") ";
+    std::cout << "(" << demangle(typeid(T).name()) << ") ";
     traits<T>::Print(value_, str);
   }
 
--- a/gtsam/base/types.cpp
+++ b/gtsam/base/types.cpp
@@ -48,8 +48,7 @@ std::string demangle(const char* name) {
 
   std::free(demangled);
 
-#endif
-#ifdef _WIN32
+#elif _WIN32
     char undecorated_name[1024];
     
     if (UnDecorateSymbolName(
--- a/gtsam/base/types.cpp
+++ b/gtsam/base/types.cpp
@@ -44,7 +44,7 @@ std::string demangle(const char* name) {
   int status = -1; // some arbitrary value to eliminate the compiler warning
   demangled = abi::__cxa_demangle(name, nullptr, nullptr, &status),
   
-  demangled_name = (status == 0) ? std::string(demangled) : name;
+  demangled_name = (status == 0) ? std::string(demangled) : std::string(name);
 
   std::free(demangled);
 
--- a/gtsam/base/types.cpp
+++ b/gtsam/base/types.cpp
@@ -1,0 +1,68 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file     types.cpp
+ * @brief    Functions for handling type information
+ * @author   Varun Agrawal
+ * @date     May 18, 2020
+ * @addtogroup base
+ */
+
+#include <gtsam/base/types.h>
+
+#ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include <DbgHelp.h>
+#endif
+
+#ifdef __GNUG__
+#include <cstdlib>
+#include <cxxabi.h>
+#include <string>
+#endif
+
+namespace gtsam {
+
+/// Pretty print Value type name
+std::string demangle(const char* name) {
+  // by default set to the original mangled name
+  std::string demangled_name = std::string(name);
+
+#ifdef __GNUG__
+  // g++ version of demangle
+  char* demangled = nullptr;
+  int status = -1; // some arbitrary value to eliminate the compiler warning
+  demangled = abi::__cxa_demangle(name, nullptr, nullptr, &status),
+  
+  demangled_name = (status == 0) ? std::string(demangled) : name;
+
+  std::free(demangled);
+
+#endif
+#ifdef _WIN32
+    char undecorated_name[1024];
+    
+    if (UnDecorateSymbolName(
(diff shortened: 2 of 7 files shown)


jlblancoc:
:+1: 
I would love having this feature, unless it's not recommended for backward-compatibility with existing code (which shouldn't have relied on compiler-dependent names, IMO). 

But I would strongly recommend moving the implementation (and all the required `#includes`) to a .cpp file, not leaving them in the header.

Please, add the MSVC equivalent version too:

```
#ifdef _WIN32
#define WIN32_LEAN_AND_MEAN
#include <windows.h>
#include <DbgHelp.h>
#endif

...

#ifdef _WIN32
    char *mangledName; // input

		char undecorated_name[1024];
		if (UnDecorateSymbolName(
				mangledName, undecorated_name, sizeof(undecorated_name),
				UNDNAME_COMPLETE))
		{
                       // successful conversion: 
                      // take value from: undecorated_name			
		}
#endif
```

and in gtsam cmake:

```
if(WIN32)
  target_link_libraries(gtsam PRIVATE Dbghelp) # For mrpt::system::getCallStackBackTrace()
endif()
```

jlblancoc:
Also, I don't understand the need for using std::unique_ptr :-) 

Please, check, and feel free to copy/adapt, another way to do it w/o unique_ptr from: 
https://github.com/MRPT/mrpt/blob/develop/libs/system/src/backtrace.cpp#L104-L114

varunagrawal:
@jlblancoc I picked up that piece of code from StackOverflow and it seemed fine on first glance, especially for freeing up the demangled string memory. Though I see your point when we are explicitly calling `free`.

varunagrawal:
Also regarding `.cpp` file for this, I agree. I wanted @dellaert's design feedback on the best way to go about this.

dellaert:
I agree with all of @jlblancoc 's comments. I propose you guys iterate until you converge and then request a review by me, but it seems pretty uncontroversial and a pretty nice improvement!

There is one possible issue which @ProfFan can weigh in on: we had errors where even the mangled types completely agreed, because of some intricate c++ issue that I now forget. This means users could get "you requested `abc` but Values has `abc`." Not sure how to solve - and might be orthogonal. But if we see that the mangled types and the demangled types are the same, we could print out a verbose message saying what the likely issue is. @ProfFan can help construct that message.

varunagrawal:
@jlblancoc updated the code. Can you please take a look? I'll keep updating based on your feedback until you are satisfied. :slightly_smiling_face:

gtsam/base/types.cpp
```diff
@@ -0,0 +1,68 @@
...
+  char* demangled = nullptr;
+  int status = -1; // some arbitrary value to eliminate the compiler warning
+  demangled = abi::__cxa_demangle(name, nullptr, nullptr, &status),
+  
+  demangled_name = (status == 0) ? std::string(demangled) : name;
+
+  std::free(demangled);
+
+#endif
+#ifdef _WIN32
```
jlblancoc, line 52: Probably a pedantic issue... :-)
If this was ever compiled by MinGW, it will go over BOTH #if's... Perhaps replace the #endif + #ifdef _WIN32 with #elif ?

gtsam/nonlinear/tests/testValues.cpp
```diff
@@ -589,6 +589,23 @@ TEST(Values, MatrixDynamicInsertFixedRead) {
   CHECK_EXCEPTION(values.at<Matrix23>(key1), exception);
 }
 
+TEST(Values, Demangle) {
+  Values values;
+  Matrix13 v; v << 5.0, 6.0, 7.0;
+  values.insert(key1, v);
+  string expected = "Values with 1 values:\nValue v1: (Eigen::Matrix<double, 1, 3, 1, 1, 3>) [\n	5, 6, 7\n]\n\n";
```
jlblancoc, line 596: I'm not sure relying on "\n" being universally captured from cout as "\r" vs "\r\n", etc. 
But let's see if it passes all tests on CI for Linux/Mac/Windows !

gtsam/nonlinear/tests/testValues.cpp
```diff
@@ -589,6 +589,23 @@ TEST(Values, MatrixDynamicInsertFixedRead) {
...
+
+  stringstream buffer;
+  streambuf * old = cout.rdbuf(buffer.rdbuf());
```
jlblancoc, line 599: Great idea for a test!

I've been thinking for a long time to add an additional optional std::ostream parameter to all print() functions to simplify this kind of things. But yours is the kind of hack that is required as things are now :+1:

gtsam/nonlinear/tests/testValues.cpp
```diff
@@ -589,6 +589,23 @@ TEST(Values, MatrixDynamicInsertFixedRead) {
...
+
+  values.print();
+
+  string actual = buffer.str();
+  cout.rdbuf(old);
+
+  EXPECT(assert_equal(expected, actual));
```
jlblancoc, line 606: `EXPECT(assert_equal())`  sounds weird to me, but I'm not so familiar with this unit test suite. From GTest, I would expect something like EQUAL_STR(), but if you are sure that's the correct way, forget my comment

gtsam/base/types.cpp
varunagrawal: Yes I was wondering about that. A quick search gave me #elseif which didn't work so I figured it was better to be explicit. Thank you for the suggestion.

gtsam/nonlinear/tests/testValues.cpp
varunagrawal: I looked through other tests in GTSAM which do string comparisons and this seemed to be the universal way of doing it. But I know what you mean! Frank is very well aware of some of my frustrations with CppUnitLite.

gtsam/nonlinear/tests/testValues.cpp
varunagrawal: That's an excellent point. I guess I've been living in Linux land for far too long.

gtsam/base/GenericValue.h
```diff
@@ -83,7 +84,7 @@ class GenericValue: public Value {
 
   /// Virtual print function, uses traits
   virtual void print(const std::string& str) const {
-    std::cout << "(" << typeid(T).name() << ") ";
+    std::cout << "(" << demangle(typeid(T).name()) << ") ";
```
varunagrawal, line 87: @jlblancoc I did change it. :) I was thinking this line looked familiar.

varunagrawal:
@ProfFan already did that in the last commit. :slightly_smiling_face: 

(P.S. I may have missed some, so if you find any please do let me know)

varunagrawal:
I am definitely going to need more information regarding the `type == type` phenomenon. For now I can simply check if the demangled names are the same, and if yes, print the original mangled names. Would that work @dellaert?

dellaert:
> I am definitely going to need more information regarding the `type == type` phenomenon. For now I can simply check if the demangled names are the same, and if yes, print the original mangled names. Would that work @dellaert?

Nope. They will be the same as well. I'd like @ProfFan to actually read my comment and respond to you with a draft for a meaningful message to the user :-)

jlblancoc:
Maybe related ([Source](https://stackoverflow.com/a/40675143/1631514)):

  ===========

http://en.cppreference.com/w/cpp/language/typeid. 
The same source goes on to say:

> There is no guarantee that the same std::type_info instance will be referred to by all evaluations of the typeid expression on the same
> type, although std::type_info::hash_code of those type_info objects
> would be identical, as would be their std::type_index.

That means that comparing `&typeid(impl)` to something else, could return false even when the object has the same dynamic type. So it's not such a good choice to compare their addresses. You should compare the objects themselves directly instead, i.e. just remove the `&` operator from both sides, because `std::type_info` (returned by `typeid`) has `operator==` defined.

jlblancoc:
PD: So, perhaps it's safer to change the references in: 

https://github.com/borglab/gtsam/blob/b6c5bceae95d4512a1e839288b22ac73c3546131/gtsam/nonlinear/Values.h#L474-L475

to non-reference values `const std::type_index`, which seem to be safely constructible from `std::type_info`?
I have not even tested if it compiles, but seems correct...thoughts?

ProfFan:
@varunagrawal @dellaert  I propose to do the following:
1. Keep the current logic, print the same error message;
2. When the two names are identical, print the following:
```
WARNING: Detected types with same name but different `typeid`.
This is usually caused by incorrect linking/inlining settings when compiling libraries using GTSAM.
If you are a user, please report to the author of the library using GTSAM.
If you are a package maintainer, please consult `cmake/GtsamPybindWrap.cmake`, line 74 for details.
```

varunagrawal:
@jlblancoc that is an excellent find but I don't know if it is related. From the looks of it, if the type names are the same for the stored and requested types, this means that the `hash_code` and `type_index` should also be the same?

To paraphrase for my own understanding, in the example provided, the answer is basically saying that there is no type interning (similar to string interning in Java) happening when you run `typeid` so it is important to compare the hash codes and not the objects themselves since the `type_info` instances might be different.

varunagrawal:
@ProfFan I added your error message. It would be nice to have a test that reproduces this error though.

jlblancoc:
@varunagrawal :
> To paraphrase for my own understanding, in the example provided, the answer is basically saying that there is no type interning (similar to string interning in Java) happening when you run typeid so it is important to compare the hash codes and not the objects themselves since the type_info instances might be different.

Yes. Or more precisely: it's important to compare the hash codes or the objects themselves, *not* **pointers** to `typeid` objects, since instances may be different despite representing the same type. 
I commented because I noticed that *references* to `typeid` types were stored (which are the same than pointers). 
But later, I realized the `operator ==` of the class (*not the pointers*) is used to make comparisons, so: everything seems correct :+1:

PR #310, opened 2020-05-13
Fix 32-bit unit test failures

jlblancoc:
Closes #308 
Closes #306 

Don't merge until [these PPA builds](https://launchpad.net/~joseluisblancoc/+archive/ubuntu/sandbox/+packages) end, to confirm that the test errors have been fixed.

Update: It can be merged now, PPA builds pass!

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/310)

PR #314, opened 2020-05-18
More precise use of setZero() and add comments

jlblancoc:
This is a follow up of #310 , trying to minimize the changes wrt former behavior and improve the docs. 
cc: @dellaert 

Mainly, the idea is that constructors taking lists of blocks **could** fill with zeros so if a block is not provided, it's taken for granted that it will be filled with zeros (this is now documented, and also fixes the original x32 bug, by the way), while the third constructor modified in the former PR, taking as input a `Matrix`, doesn't actually need the `setZero()` since it will be immediately after that overwritten, so it's a potential waste of time.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/314)

### DIFF
--- a/gtsam/base/SymmetricBlockMatrix.h
+++ b/gtsam/base/SymmetricBlockMatrix.h
@@ -71,6 +71,7 @@ namespace gtsam {
     }
 
     /// Construct from a container of the sizes of each block.
+    /// Uninitialized blocks are filled with zeros.
     template<typename CONTAINER>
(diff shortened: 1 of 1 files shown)


dellaert:
Hmmmm. Since it is a matrix, and we use Eigen for matrices, I actually think more and more its confusing to have different semantics from Eigen. Is it so much harder to fix the original bug, now that we know what the cause is? If needed, a setZero method could be added if it's not already there. To flush out current and future problems we could ad a CI check that initializes to Nan.

PR #316, opened 2020-05-19
added email parameter to ubuntu/debian packaging scripts

berndpfrommer:
This PR adds a parameter to the packaging scripts such that they can be used without having to be edited.
The documentation is adjusted accordingly, and also expanded to cover the handling of keys used for signing.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/316)

### DIFF
--- a/package_scripts/prepare_ubuntu_pkgs_for_ppa.sh
+++ b/package_scripts/prepare_ubuntu_pkgs_for_ppa.sh
@@ -1,20 +1,55 @@
 #!/bin/bash
 # Creates a set of packages for each different Ubuntu distribution, with the
-# intention of uploading them to:
-#   https://launchpad.net/~joseluisblancoc/
+# intention of uploading them to a PPA on launchpad
 #
 # JLBC, 2010
 # [Addition 2012:]
 #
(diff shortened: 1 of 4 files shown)


berndpfrommer:
Maybe @jlblancoc can have a look at the commits? Thanks!

package_scripts/prepare_ubuntu_pkgs_for_ppa.sh
```diff
@@ -1,20 +1,55 @@
...
+    show_help
+    exit -1
+fi
+
+
 set -e
 
 # List of distributions to create PPA packages for:
-LST_DISTROS=(xenial bionic eoan focal)
+LST_DISTROS=(xenial bionic disco eoan focal)
```
jlblancoc, line 51: disco should be removed, since it's deprecated (EOL) and won't be accepted in PPAs. 
groovy (u20.10) could be added, by the way ;-)

berndpfrommer: I found another bug, introduced at the last minute. I'll close this and will resubmit. Very embarrassing ...

PR #317, opened 2020-05-21
Fix Cython build on Windows/MSVC

tuwuhs:
This fixes two issues with the Cython build on Windows/MSVC:
1. Fail compiling `gtsam.pyx` due to error [`C1128: number of sections exceeded object file format limit`](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1128?view=vs-2019). This is solved by adding the compile option `/bigobj`, as suggested in the above link.
2. Fail loading the module `gtsam` in Python after successful install. The error was:
```
Python 3.7.3 (default, Mar 27 2019, 17:13:21) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import gtsam
  File "<stdin>", line 1, in <module>
  File "C:\ProgramData\Anaconda3\lib\site-packages\gtsam-4.0.2-py3.7.egg\gtsam\__init__.py", line 1, in <module>
    from .gtsam import *
ModuleNotFoundError: No module named 'gtsam.gtsam'
>>>
```
This is solved by renaming the compiled file extension from `.dll` to `.pyd`. 

Tested on MSVC 2019 x64 compiler (cl.exe v. 19.25.28614), Python 3.7.3, CMake 3.16.4, Ninja 1.9.0, Windows 10 Pro (1909).

### DIFF
--- a/cmake/GtsamCythonWrap.cmake
+++ b/cmake/GtsamCythonWrap.cmake
@@ -87,6 +87,12 @@ endfunction()
 #    - output_dir:   The output directory
 function(build_cythonized_cpp target cpp_file output_lib_we output_dir)
   add_library(${target} MODULE ${cpp_file})
+  
+  # Use .pyd extension instead of .dll on Windows
+  if(WIN32)
+    set_target_properties(${target} PROPERTIES SUFFIX ".pyd")
+  endif()
+  
   if(APPLE)
     set(link_flags "-undefined dynamic_lookup")
   endif()
--- a/cython/CMakeLists.txt
+++ b/cython/CMakeLists.txt
@@ -7,6 +7,11 @@ if (GTSAM_INSTALL_CYTHON_TOOLBOX)
   add_subdirectory(gtsam_eigency)
   include_directories(${PROJECT_BINARY_DIR}/cython/gtsam_eigency)
 
+  # Fix for error "C1128: number of sections exceeded object file format limit"
+  if(MSVC)
+    add_compile_options(/bigobj)
+  endif()
+
   # wrap gtsam
   add_custom_target(gtsam_header DEPENDS "../gtsam.h")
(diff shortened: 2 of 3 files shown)


ProfFan:
Wow this is very interesting! Can you try running the tests in `cython/gtsam/tests` and see if there are errors?

tuwuhs:
Sure, here you go:
```
C:\Program Files\GTSAM\cython
(base) λ python -m unittest discover
.........................................C:\Program Files\GTSAM\cython\gtsam\tests\test_dsf_map.py:36: DeprecationWarning: Please use assertEqual instead.
  self.assertEquals(key(dsf.find(pair1)), key(dsf.find(pair2)))
.....
Ran 46 tests in 0.102s

OK
```

I added a new commit adding the full path of Python library for linking. This helps Visual Studio generator to find `python37.lib`, so the user doesn't need to modify the `.vcxproj` project file. Building with Visual Studio now works out-of-the-box (previously I only tried nmake and Ninja, setting the path manually as environment variables).

fangxu622:
hi · I got this problem 

```
>Generating gtsam/gtsam.cpp
7>Compiling G:\PHD-Project\3rdparty\gtsam-develop\build\cython\gtsam\gtsam.pyx
7>
7>Error compiling Cython file:
7>------------------------------------------------------------
7>...
7>
7>from gtsam_eigency.core cimport *
7>^
7>------------------------------------------------------------
7>
7>gtsam\gtsam.pxd:2:0: 'gtsam_eigency\core.pxd' not found
7>E:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(231,5): error MSB6006: “cmd.exe”
```

why didn't found core.pxd， in fact .it exist.

PR #323, opened 2020-05-26
Optional jacobian computation in PreintegrationBase predict function

varunagrawal:
Speed up of `predict` in `PreintegrationBase.cpp` by asking for jacobians for `predict_state` and `predict_delta` optionally.

Also removed the TODO since `testIMUFactor.cpp` has tests for correctness.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/323)

### DIFF
--- a/gtsam/navigation/PreintegrationBase.cpp
+++ b/gtsam/navigation/PreintegrationBase.cpp
@@ -115,20 +115,21 @@ void PreintegrationBase::integrateMeasurement(const Vector3& measuredAcc,
 NavState PreintegrationBase::predict(const NavState& state_i,
     const imuBias::ConstantBias& bias_i, OptionalJacobian<9, 9> H1,
     OptionalJacobian<9, 6> H2) const {
-  // TODO(frank): make sure this stuff is still correct
   Matrix96 D_biasCorrected_bias;
   Vector9 biasCorrected = biasCorrectedDelta(bias_i,
-      H2 ? &D_biasCorrected_bias : 0);
+                                             H2 ? &D_biasCorrected_bias : nullptr);
 
   // Correct for initial velocity and gravity
   Matrix9 D_delta_state, D_delta_biasCorrected;
   Vector9 xi = state_i.correctPIM(biasCorrected, deltaTij_, p().n_gravity,
-      p().omegaCoriolis, p().use2ndOrderCoriolis, H1 ? &D_delta_state : 0,
-      H2 ? &D_delta_biasCorrected : 0);
+                                  p().omegaCoriolis, p().use2ndOrderCoriolis, H1 ? &D_delta_state : nullptr,
+                                  H2 ? &D_delta_biasCorrected : nullptr);
 
   // Use retract to get back to NavState manifold
   Matrix9 D_predict_state, D_predict_delta;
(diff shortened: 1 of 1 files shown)


mikesheffler:
> 
> 
> That should result in a nice speedup for anyone using predict.

Guess who's got two thumbs and uses `predict`? THIS GUY. 😄 Excited to see this roll out.

dellaert:
@mikesheffler would be nice if you could validate that there is indeed a speedup :-) We just saw this small optimization as were were looking through the code in advance of using it. We have not actually  used it yet in this new project!

mikesheffler:
> 
> 
> @mikesheffler would be nice if you could validate that there is indeed a speedup :-) 

The codebase I would use is on a machine that I can't reach via VPN (have to physically go in to the office, which I will probably do in the next few days), but, let me see if I have something on a machine I can reach. If I do, I will give it a try and report back!

mikesheffler:
Btw, Could one of you point me toward a good example of some code in the repo doing this basic sort of timing? There are lots of ways to run a timer, but I may as well do it the preferred way, if such a way exists.

dellaert:
Anything time*.cpp
Good example would use timing.h

varunagrawal:
Recommendation to use the `std::chrono` library. It takes care of a lot of the unnecessary hassles.

dellaert:
> Recommendation to use the `std::chrono` library. It takes care of a lot of the unnecessary hassles.

My recommendation to use timing.h stands... If we don't use chrono in there (we might!) feel free to PR

varunagrawal:
> > Recommendation to use the `std::chrono` library. It takes care of a lot of the unnecessary hassles.
> 
> My recommendation to use timing.h stands... If we don't use chrono in there (we might!) feel free to PR

I mentioned `std::chrono` before I saw your comment on `timing.h` (guess github didn't refresh in time). From what I've seen, `timing.h` uses Boost timers and optionally TBB timers if TBB is available.

mikesheffler:
**Short:** It's faster. By a lot. Nice job, @varunagrawal 

**Longer:** 

I won't be able to get on my computer I use for this stuff for a little while, but I did a quick test in a VM. I modified `ImuFactorsExample.cpp` (it uses `predict`) to include timing tests, by wrapping the call to `predict` in `gttic_(Predict)` and `gttoc_(Predict)` with the toc call being followed by `tictoc_finishedIteration();` and, finally, `tictoc_print_();` being called at the end.

The call to `predict` doesn't happen nearly enough with that dataset (only 100 times), so I created a loop wherein I assigned the output of predict to a dummy variable and did so 1 million times. So, with the 100 times the loop is called, that should be 100 million calls to `predict`.

Over several runs, the average without the change was 27.03 seconds (this is the first number reported by `tictoc_print()` -- it's stated as `27.xx CPU`). The average _with_ the change was 15.48 seconds.

So, that's a pretty substantial change. There's a lot of asterisks (virtual machine, tight loop, not a lot of power in the statistics ....), but I'm at least confident in saying it's clearly beneficial 😄

dellaert:
That's awesome. Will merge !

varunagrawal:
Nice!!! Thank you for the analysis @mikesheffler

PR #325, opened 2020-05-28
Alternate signs on Hat/Vee and update tests

binitshah:
This PR makes the following changes:

 - Exposed `checkConvergence` with NonlinearOptimizerParams argument to cython wrapper
 - Exposed `SOn` versions of PriorFactor and NonlinearEquality to cython wrapper
 - Added default constructor and `push_back` to BetweenFactorPose3s in cython wrapper
 - Renamed `ManifoldImpl` to `GetDimensionImpl`
 - `LieGroupTraits` now inherits from `GetDimensionImpl`
 - Alternate Hat/Vee signs in SO3, SO4, and SOn
 - Updates related unit tests.

PR was tested locally with all unit tests passing.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/325)

### DIFF
--- a/gtsam/geometry/SO3.cpp
+++ b/gtsam/geometry/SO3.cpp
@@ -261,13 +261,13 @@ Vector3 SO3::Logmap(const SO3& Q, ChartJacobian H) {
 
   // when trace == -1, i.e., when theta = +-pi, +-3pi, +-5pi, etc.
   // we do something special
-  if (std::abs(tr + 1.0) < 1e-10) {
-    if (std::abs(R33 + 1.0) > 1e-10)
+  if (tr + 1.0 < 1e-10) {
+    if (std::abs(R33 + 1.0) > 1e-5)
       omega = (M_PI / sqrt(2.0 + 2.0 * R33)) * Vector3(R13, R23, 1.0 + R33);
-    else if (std::abs(R22 + 1.0) > 1e-10)
+    else if (std::abs(R22 + 1.0) > 1e-5)
       omega = (M_PI / sqrt(2.0 + 2.0 * R22)) * Vector3(R12, 1.0 + R22, R32);
     else
-      // if(std::abs(R.r1_.x()+1.0) > 1e-10)  This is implicit
+      // if(std::abs(R.r1_.x()+1.0) > 1e-5)  This is implicit
       omega = (M_PI / sqrt(2.0 + 2.0 * R11)) * Vector3(1.0 + R11, R21, R31);
   } else {
     double magnitude;
--- a/gtsam/geometry/SO3.cpp
+++ b/gtsam/geometry/SO3.cpp
@@ -262,12 +262,12 @@ Vector3 SO3::Logmap(const SO3& Q, ChartJacobian H) {
   // when trace == -1, i.e., when theta = +-pi, +-3pi, +-5pi, etc.
   // we do something special
   if (tr + 1.0 < 1e-10) {
-    if (std::abs(R33 + 1.0) > 1e-5)
+    if (std::abs(R33 + 1.0) > 1e-10)
       omega = (M_PI / sqrt(2.0 + 2.0 * R33)) * Vector3(R13, R23, 1.0 + R33);
-    else if (std::abs(R22 + 1.0) > 1e-5)
+    else if (std::abs(R22 + 1.0) > 1e-10)
       omega = (M_PI / sqrt(2.0 + 2.0 * R22)) * Vector3(R12, 1.0 + R22, R32);
     else
-      // if(std::abs(R.r1_.x()+1.0) > 1e-5)  This is implicit
+      // if(std::abs(R.r1_.x()+1.0) > 1e-10)  This is implicit
       omega = (M_PI / sqrt(2.0 + 2.0 * R11)) * Vector3(1.0 + R11, R21, R31);
   } else {
     double magnitude;
--- a/gtsam/geometry/SO3.cpp
+++ b/gtsam/geometry/SO3.cpp
@@ -262,12 +262,12 @@ Vector3 SO3::Logmap(const SO3& Q, ChartJacobian H) {
   // when trace == -1, i.e., when theta = +-pi, +-3pi, +-5pi, etc.
   // we do something special
   if (tr + 1.0 < 1e-10) {
-    if (std::abs(R33 + 1.0) > 1e-10)
+    if (std::abs(R33 + 1.0) > 1e-5)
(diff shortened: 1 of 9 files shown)


dellaert:
LGTM, except the failing test:

```
not equal:
expected[-0.264544406; 0.742217405; 3.04117314];
actual[0.264451979; -0.742197651; -3.04098211];
/home/travis/build/borglab/gtsam/gtsam/geometry/tests/testRot3.cpp:247: Failure: "assert_equal(Vector3(-0.264544406, 0.742217405, 3.04117314), (Vector)Rot3::Logmap(Rlund), 1e-8)" 
There were 1 failures
```

On *one* platform there is apparently a sign change.  This is probably due to the change in SO3.cpp - Can you revert that code (by commenting out) and confirm that hypothesis? If that's the case, we need to examine that more closely and possibly put that change in a different PR.

gtsam/geometry/SO3.cpp
```diff
@@ -261,13 +261,13 @@ Vector3 SO3::Logmap(const SO3& Q, ChartJacobian H) {
 
   // when trace == -1, i.e., when theta = +-pi, +-3pi, +-5pi, etc.
   // we do something special
-  if (std::abs(tr + 1.0) < 1e-10) {
-    if (std::abs(R33 + 1.0) > 1e-10)
+  if (tr + 1.0 < 1e-10) {
+    if (std::abs(R33 + 1.0) > 1e-5)
```
ProfFan, line 265: Why this change from `1e-10` to `1e-5`?

dellaert:
@binitshah I agree, the description is lacking :-) It says one thing but then there are other things happening besides that.

binitshah:
Updated PR description to better document the changes included in this PR. See edit.

dellaert:
Still a failure in the build. Let me know when ready by re-requesting a review...

dellaert:
@akshay-krishnan work with Binit to get this in by meeting?

akshay-krishnan:
> LGTM, except the failing test:
> 
> ```
> not equal:
> expected[-0.264544406; 0.742217405; 3.04117314];
> actual[0.264451979; -0.742197651; -3.04098211];
> /home/travis/build/borglab/gtsam/gtsam/geometry/tests/testRot3.cpp:247: Failure: "assert_equal(Vector3(-0.264544406, 0.742217405, 3.04117314), (Vector)Rot3::Logmap(Rlund), 1e-8)" 
> There were 1 failures
> ```
> 
> On _one_ platform there is apparently a sign change. This is probably due to the change in SO3.cpp - Can you revert that code (by commenting out) and confirm that hypothesis? If that's the case, we need to examine that more closely and possibly put that change in a different PR.

The unit test is failing when using Quaternions for rotations. Both the expected and the actual values there correspond to the same rotation matrix (at least up to the 5th decimal place). The sign swap must be happening [here](https://github.com/borglab/gtsam/blob/d0d80f55a6391e6811ea337cf1e33a319115c2cc/gtsam/geometry/Quaternion.h#L122) in Quaternion.cpp. Essentially, the rotations for expected and actual differ by 2*pi.

akshay-krishnan:
Adding to the previous comment, the expected value has a rotation angle greater than pi (although in the 8th decimal place). The code in Quaternion.cpp limits the rotation angle to lie between [-pi, pi]. If this is the convention, should we not be expecting that in the unit test, and in SO3.cpp as well?

dellaert:
You guys are close to figuring it out?

akshay-krishnan:
> You guys are close to figuring it out?

Yes @dellaert , working on it. I'll post updates here.

akshay-krishnan:
This [rotation logmap unit test](https://github.com/borglab/gtsam/blob/d0d80f55a6391e6811ea337cf1e33a319115c2cc/gtsam/geometry/tests/testRot3.cpp#L246) from the Lund dataset is weird because the trace of the rotation matrix is slightly less than -1 (supposed to be in the range [-1, 3]). The expected rotation vector has an angle slightly greater than pi. 

SO3: for well-conditioned rotation matrices, returns axis and angle between [0, pi]. For [cases when angle is close to pi](https://github.com/borglab/gtsam/blob/d0d80f55a6391e6811ea337cf1e33a319115c2cc/gtsam/geometry/SO3.cpp#L264) (as in this case), the angle can be greater than pi. Hence it returns the expected value. 

Quaternion: If the computed angle is greater than pi, it is shifted to lie in [-pi, pi] for reasons mentioned [here](https://github.com/borglab/gtsam/blob/dbcb3226dd4191d133b9f2b2643d90d5f85775fc/gtsam/geometry/Quaternion.h#L121). 

Hence the difference in the returned value between SO3 and Quaternion representations for this unit test. Although the returned values are different they represent the same rotation (omega, axis) vs (2*pi - omega, -axis). In order to fix this without breaking anything, shall we restrict the unit test to SO3 representations alone? We can provide another expected return value for Quaternion representations, if necessary.

binitshah:
For SO3, why is the angle bound between [0, pi]. We could match SO3 to Quaternion and change to the unittest's expected value.

dellaert:
So, the Lund example is real: the essential matrix calculation yielded this “rotation matrix”. Now, when I made this unit test - to figure out what was going on - I updated the SO3 log map to deal with this. Why? I could have fit the closest SO3 by using “ClosestTo”. However, I chose to slightly fuzz the log map method instead, so it could deal with this. Do you think this is possible in the quarternion path as well, or impossible because we don’t control that code?

akshay-krishnan:
> Do you think this is possible in the quaternion path as well, or impossible because we don’t control that code?

For SO3, there is a special case to deal with the singularity at pi. I believe the modification @dellaert made uses the same case for this rotation matrix. However, there is no singularity at pi for quaternions. So the execution flows through the general case where angle is limited to [-pi, pi]. Changing this to [0, 2*pi] would affect quaternions everywhere else. 

@binitshah In the general SO3 case, the [0, pi] limitation is because of an arccos() function that is used to compute an angle. This is quite standard. You may want to check the math for SO3 logmap. However, that does not happen here as I mentioned in my previous comment.

dellaert:
If what I'm hearing is correct, the quaternion gives the same answer as SO3, just with a different sign. I guess we can just change the test case based on an ifdef then...

akshay-krishnan:
> If what I'm hearing is correct, the quaternion gives the same answer as SO3, just with a different sign. I guess we can just change the test case based on an ifdef then...

Yes, we can do that. To illustrate what is happening - we get an angle around 181 degrees with SO3 and -179 degrees with quaternion. The two are equivalent. But the difference is not purely the sign. There's also a small magnitude difference.

akshay-krishnan:
I think it would help to add a comment to explain what is going on in that unit test. We can merge after that.

PR #326, opened 2020-05-29
Fix problem introduced in #255.

zwn:
The api plot_pose3 and plot_pose3_on_axes has been changed, breaking
backward compatibility by adding arguments in the middle. The newly
introduced arguments have been moved to the end of the argument list to
mitigate the problem.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/326)


varunagrawal:
I admittedly did a bad job on this review. We seem to have a regression, since while the functions were updated, the function calls were not. I'll get on it.

PR #329, opened 2020-06-01
Functorized Factor

varunagrawal:
This PR adds a new type of factor which allows for more dynamic computation.

Consider the case that the input type to a factor's `evaluateError` method is different from the `measurement`, and the factor requires some intermediate processing of the input to match the type (both syntactically and semantically) of the measurement. The `FunctorizedFactor` accepts the functor argument type as a template argument and operates on the input, allowing for considerable reuse.

E.g. if we wish to compute the error from a simple linear matrix operation `Ax = b`, we could define a factor like:

```cpp
Key key = Symbol('X', 0);
auto model = noiseModel::Isotropic::Sigma(9, 1);

/// Functor that takes a matrix and multiplies every element by m
class MultiplyFunctor {
  double m_; ///< simple multiplier

public:
  MultiplyFunctor(double m) : m_(m) {}

  Matrix operator()(const Matrix &X,
                    OptionalJacobian<-1, -1> H = boost::none) const {
    if (H)
      *H = m_ * Matrix::Identity(X.rows() * X.cols(), X.rows() * X.cols());
    return m_ * X;
  }
};

Matrix X = Matrix::Identity(3, 3), measurement = Matrix::Identity(3, 3);
double multiplier = 1.0;

// `multiplier` will be passed in to the functor constructor
auto factor = MakeFunctorizedFactor<Matrix>(key, measurement, model, multiplier);

// This factor now evaluates `m*A` and computes the error to `b` -> `mA - b`
Matrix error = factor.evaluateError(X);
```

Thus the factor is completely determined by the functor passed in, thus providing a great amount of simplicity.

The factor also supports `std::function` and C++ lambda types.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/329)


gtsam/linear/NoiseModel.cpp
```diff
@@ -619,6 +619,7 @@ void Isotropic::WhitenInPlace(Eigen::Block<Matrix> H) const {
 // Unit
 /* ************************************************************************* */
 void Unit::print(const std::string& name) const {
+  //TODO(Varun): Do we need that space at the end?
```
dellaert, line 622: remove from this PR

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,123 @@
...
+#include <gtsam/nonlinear/NonlinearFactor.h>
+
+#include <cmath>
+
+namespace gtsam {
+
+/**
+ * Factor which evaluates functor and uses the result to compute
+ * error on provided measurement.
+ * The provided FUNCTOR should provide two definitions: `argument_type` which
```
dellaert, line 29: type aliases

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,123 @@
...
+ * corresponds to the type of input it accepts and `return_type` which indicates
+ * the type of the return value. This factor uses those type values to construct
+ * the functor.
+ *
```
dellaert, line 33: Add the example from the description *after* you fix it (constructor has no key, jacobian is ignored although it would be simple in your example)

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,123 @@
...
+    typename FUNCTOR::return_type x = func_(params, H);
+    Vector error = traits<typename FUNCTOR::return_type>::Local(measured_, x);
+    return error;
+  }
+
+  /// @name Testable
+  /// @{
+  GTSAM_EXPORT friend std::ostream &
+  operator<<(std::ostream &os, const FunctorizedFactor<FUNCTOR> &f) {
+    os << "  noise model sigmas: " << f.noiseModel_->sigmas().transpose();
```
dellaert, line 87: Stream operator is not part of Testable. Move it somewhere else? Or, in fact remove it. Adding streaming is a big design exercise, large refactor we might not want to tackle now

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,123 @@
...
+              << keyFormatter(this->key()) << ")" << std::endl;
+    traits<typename FUNCTOR::return_type>::Print(measured_, "  measurement: ");
+    std::cout << *this << std::endl;
+  }
+
+    virtual bool equals(const NonlinearFactor &other, double tol = 1e-9)
+    const {
+      const FunctorizedFactor<FUNCTOR> *e =
+          dynamic_cast<const FunctorizedFactor<FUNCTOR>*>(&other);
+      const bool base = Base::equals(*e, tol);
```
dellaert, line 103: will fail if e == nullptr

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,123 @@
...
+  template <class ARCHIVE>
+  void serialize(ARCHIVE &ar, const unsigned int /*version*/) {
+    ar &boost::serialization::make_nvp(
+        "NoiseModelFactor1", boost::serialization::base_object<Base>(*this));
+    ar &BOOST_SERIALIZATION_NVP(measured_);
+    ar &BOOST_SERIALIZATION_NVP(func_);
+  }
+};
+
+// TODO(Varun): Include or kill?
```
dellaert, line 120: if factors have it, include. But do they?

gtsam/nonlinear/tests/testFunctorizedFactor.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+public:
+  using argument_type = Matrix;
+  using return_type = Matrix;
+
+  MultiplyFunctor(double m) : m_(m) {}
+
+  Matrix operator()(const Matrix &X,
+                    OptionalJacobian<-1, -1> H = boost::none) const {
+    return m_ * X;
```
dellaert, line 44: jacobian? Remember, this is a copy/paste source

dellaert:
Awesome. Some comments...

gtsam/nonlinear/FunctorizedFactor.h
varunagrawal: Just as a note, there are factors that have the stream operator as a part of Testable.

gtsam/nonlinear/FunctorizedFactor.h
varunagrawal: They do. Included.

gtsam/nonlinear/tests/testFunctorizedFactor.cpp
varunagrawal: Great point.

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,123 @@
...
+ * corresponds to the type of input it accepts and `return_type` which indicates
+ * the type of the return value. This factor uses those type values to construct
+ * the functor.
+ *
+ * Template parameters are
+ * @param FUNCTOR: A class which operates as a functor.
+ */
+template <typename FUNCTOR>
+class GTSAM_EXPORT FunctorizedFactor
+    : public NoiseModelFactor1<typename FUNCTOR::argument_type> {
```
gchenfc, line 39: Is it possible to allow this to accept more variables?  It seems like it should be reasonably common to have

|h(x1, x2, ...) - z|

where we define a functor for h(x1, x2, ...).

gtsam/nonlinear/tests/testFunctorizedFactor.cpp
```diff
@@ -0,0 +1,127 @@
...
+  double multiplier = 1.0;
+
+  FunctorizedFactor<MultiplyFunctor> factor(keyX, X, model, multiplier);
+
+  Values values;
+  values.insert<Matrix>(keyX, X);
+
+  Matrix error = factor.evaluateError(X);
+
+  EXPECT(assert_equal(Vector::Zero(9), error, 1e-9));
```
gchenfc, line 61: `EXPECT_CORRECT_FACTOR_JACOBIANS` ?

gtsam/nonlinear/FunctorizedFactor.h
varunagrawal: Yup it takes any number of arguments. That's why you have the `template <typename... Args>` on the constructor.

gtsam/nonlinear/tests/testFunctorizedFactor.cpp
varunagrawal: Funny coincidence: I was just working on this.

gtsam/nonlinear/FunctorizedFactor.h
gchenfc: Sorry to clarify, I meant rather than NoiseModelFactor1, also be able to have NoiseModelFactor2, 3, etc

So right now from the constructor templating it's able to do something like

h(x, theta1, theta2, ...)

where theta are fixed parameters, but I was wondering about multiple variables

varunagrawal: Ah I see. Yeah that's phase 2. Now that we have this factor working, it should be straightforward to extend it to multiple keys.

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,145 @@
...
+#include <gtsam/nonlinear/NonlinearFactor.h>
+
+#include <cmath>
+
+namespace gtsam {
+
+/**
+ * Factor which evaluates functor and uses the result to compute
+ * error on provided measurement.
+ * The provided FUNCTOR should provide two type aliases: `argument_type` which
```
dellaert, line 30: Now I’m wondering whether this is just any `std::function`? And whether we could use it with a lambda? https://en.cppreference.com/w/cpp/utility/functional/function

I propose you add two unit tests that exercise those cases and we can see if and how it leads to a change in your requirements. It could also be much simpler to just say that the Functor has to satisfy the standard function concept

In fact, reading more, we probably have to:

https://stackoverflow.com/questions/35907454/why-stdfunctionargument-type-has-been-deprecated

“Now that we have decltype that can be used to deduce a callable's return type, variadic templates and perfect forwarding to forward arbitrary number of arguments to functions, etc. the pre-C++11 mechanisms are way too cumbersome to use.”


https://en.cppreference.com/w/cpp/language/decltype

The following post seems to hint that we might not *want* to know the argument types though: https://stackoverflow.com/questions/6512019/can-we-get-the-type-of-a-lambda-argument

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,145 @@
...
+ *       if (H) *H = m_ * Matrix::Identity(X.rows()*X.cols(), X.rows()*X.cols());
+ *       return m_ * X;
+ *     }
+ *   };
+ *
+ *   Matrix measurement = Matrix::Identity(3, 3);
+ *   double multiplier = 2.0;
+ *   FunctorizedFactor<MultiplyFunctor> factor(keyX, measurement, model, multiplier);
+ */
+template <typename FUNCTOR>
```
dellaert, line 60: To address my other comment, perhaps we should instead template on T. And then a Functor/lambda/std::function passed needs to take (T, jacobian). Not necessarily dynamic jacobian, btw...

gtsam/nonlinear/FunctorizedFactor.h
```diff
@@ -0,0 +1,145 @@
...
+  /** Construct with given x and the parameters of the basis
+   *
+   * @param Args: Variadic template parameter for functor arguments.
+   *
+   * @param key: Factor key
+   * @param z: Measurement object of type FUNCTOR::return_type
+   * @param model: Noise model
+   * @param args: Variable number of arguments used to instantiate functor
+   */
+  template <typename... Args>
```
dellaert, line 85: This would only be one constructor. The other takes a lambda or function object...

gtsam/nonlinear/FunctorizedFactor.h
varunagrawal: I actually thought about this at first and had some initial difficulty which I don't remember right now. Then again, my understanding of functionals in C++ has improved quite a bit since then so I'll tackle this today.

gtsam/nonlinear/FunctorizedFactor.h
varunagrawal: The problem here is that we won't know what the return type is since we don't have anything to call `decltype` on. We need the return type to define the measurement type.

gtsam/nonlinear/FunctorizedFactor.h
varunagrawal: So I've not had a lot of success with this without over-complicating the code. Something to consider would be just removing the need for `argument_type` and `return_type` from the `FUNCTOR` and making it match the traits of `std::function` so that they can be used interchangeably.

dellaert: Yes. And it should also do lambdas. Did you make the two unit tests I proposed?

varunagrawal: I did, but I've been having trouble getting the code to compile.

varunagrawal:
So lots of good stuff thanks to my discussion with @dellaert. Declaring the `FunctorizedFactor` is crazy simple now (with lots of nice function template deduction happening behind the scenes), and we have eliminated the need for functors to define `argument_type` and `return_type` in their definitions.

varunagrawal:
Merging since this PR doesn't introduce any breaking changes

PR #331, opened 2020-06-01
remove unused post order traversal

acxz:
the post order task is not used and as such I thought it would be okay to remove this code for reducing the code base

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/331)

### DIFF
--- a/gtsam/base/treeTraversal/parallelTraversalTasks.h
+++ b/gtsam/base/treeTraversal/parallelTraversalTasks.h
@@ -33,28 +33,6 @@ namespace gtsam {
     namespace internal {
 
       /* ************************************************************************* */
-      template<typename NODE, typename DATA, typename VISITOR_POST>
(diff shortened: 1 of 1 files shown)


varunagrawal:
Hi @acxz. I disagree, I don't think it is wise to remove code just because it isn't being used. It may be useful somewhere in the future. Also this seems to be TBB specific code, are you sure it isn't used somewhere when TBB is enabled?

acxz:
no prob, if you guys want to keep it thats fine. Just thought if we need it in the future, it will be in the commit history, so thought it would be appropriate to clean up the code.

Yeah I did a thorough search couldn't find it being used anywhere.

dellaert:
So, I always say “code for the present not the future.” So if it really is not being used, I’d be ok with removing this if it is non-API breaking. Let me do a quick review first though

PR #335, opened 2020-06-02
Safer cmake default options

jlblancoc:
- Eigen: Default to system's version (only if it's first silently found). 
- Disable GTSAM_BUILD_WITH_MARCH_NATIVE by default. 

Rationale: these two settings may (and do) lead to memory crashes and obscure errors on "free'd memory that was not alloc'd", etc. if mixing different Eigen versions between gtsam and user code/other libraries.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/335)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,12 +275,18 @@ endif()
 ### use our patched version of Eigen
 ### See:  http://eigen.tuxfamily.org/bz/show_bug.cgi?id=704 (Householder QR MKL selection)
 ###       http://eigen.tuxfamily.org/bz/show_bug.cgi?id=705 (Fix MKL LLT return code)
-option(GTSAM_USE_SYSTEM_EIGEN "Find and use system-installed Eigen. If 'off', use the one bundled with GTSAM" OFF)
+
+# Default: Use system's Eigen if found automatically:
+find_package(Eigen3 QUIET)
+set(USE_SYSTEM_EIGEN_INITIAL_VALUE ${Eigen3_FOUND})
+option(GTSAM_USE_SYSTEM_EIGEN "Find and use system-installed Eigen. If 'off', use the one bundled with GTSAM" ${USE_SYSTEM_EIGEN_INITIAL_VALUE})
+unset(USE_SYSTEM_EIGEN_INITIAL_VALUE)
+
 option(GTSAM_WITH_EIGEN_UNSUPPORTED "Install Eigen's unsupported modules" OFF)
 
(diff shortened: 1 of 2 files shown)


dellaert:
I'm wondering whether we could hold off on this until the same can be done for all libraries as you and @berndpfrommer have been working on. Then we can do - if you both are willing - a quick blog post on gtsam.org outlining the change and the rationale for it. At the same time we should update the build instructions on gtsam.org (and INSTALL) to allow people to do third-part installs themselves. What do you (both) think?

jlblancoc:
Hmm... we could create a branch and make all PRs related to this topic against it, then make a PR against develop once at the end. 
Do you agree, @berndpfrommer ?

berndpfrommer:
The other option would be to create a dedicated repo under borg/ just for packaging. It looks like there is one for python packaging already. We could create another one for Debian/Ubuntu/Whatever packaging.

As suggested by  @jlblancoc we could use something like [git buildpackage](https://honk.sigxcpu.org/piki/projects/git-buildpackage/) for the new repo.

The main reason for a separate repo rather than a branch is that we may have to commit a tarball of the entire repo sources with every release, which could bloat the original repo unnecessarily. It could also make maintaining access permissions easier, because it separates developer and packager roles along repos.

dellaert:
> The other option would be to create a dedicated repo under borg/ just for packaging. It looks like there is one for python packaging already. We could create another one for Debian/Ubuntu/Whatever packaging.

Tell me what you want to do. If that, I'll create the repo, make one of you admin :-)

jlblancoc:
Ok, Frank. Yes, please, just create a new empty repo named... gtsam-debian? :-)
Give both of us write permissions to that repo, that is enough. Thanks!

berndpfrommer:
If we call it gtsam-packaging we can potentially put RPMs or snap etc in there as well.

dellaert:
> If we call it gtsam-packaging we can potentially put RPMs or snap etc in there as well.

Done

ameysutavani:
Hello @jlblancoc and @dellaert.
Thanks for the above work.
Will the above changes be eventually applied to borglab/gtsam repo as well? 
Rationale: Package managers such as conan currently build gtsam directly from source. To make these packages portable, currently, the build recipes should set the options as above anyways.
Making the above changes default will flow them nicely downstream which will avoid needs of PR such as this: https://github.com/conan-io/conan-center-index/pull/4686

Also, please advise if there is a simpler or a better way to do this.

PR #346, opened 2020-06-09
Improved Python IMU example 

varunagrawal:
As part of a longer term plan of improving the examples for `ImuFactor` and `CombinedImuFactor`, this PR updates the wrapper and adds random gaussian noise to the initial estimates to better simulate "real" scenarios.

Output visualizations looks sane and pretty much like the previous version.
imufactors


varunagrawal:
`ImuFactorExample2` already has a `DELTA` perturbation (which isn't random), so I am wondering if that example should also be updated.

PR #347, opened 2020-06-09
Wrapper Updates for Global Functions

varunagrawal:
- Wrapped Values to allow for using NavState
- Wrapped the symbol shorthands e.g. `gtsam.symbol_shorthand_X(0)`
- Updated all the tests and examples to use the new shorthand.

So I got super frustrated having to redefine the symbols each and every time, so I just wrapped all the symbol shorthands. Now we can just do

```python
from gtsam import symbol_shorthand_X as X, Values, Pose3

values = Values()
values.insert(X(0), Pose3())
```

I also want to update all the examples, so may be hold off on merging this? I'll add the examples one at a time in different commits on the same PR.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/347)

### DIFF
--- a/cython/gtsam/examples/ImuFactorExample.py
+++ b/cython/gtsam/examples/ImuFactorExample.py
@@ -12,25 +12,18 @@ from __future__ import print_function
 
 import math
 
+import gtsam
 import matplotlib.pyplot as plt
 import numpy as np
+from gtsam import symbol_shorthand_B as B
+from gtsam import symbol_shorthand_V as V
+from gtsam import symbol_shorthand_X as X
+from gtsam.utils.plot import plot_pose3
 from mpl_toolkits.mplot3d import Axes3D
 
-import gtsam
-from gtsam.utils.plot import plot_pose3
 from PreintegrationExample import POSES_FIG, PreintegrationExample
 
-BIAS_KEY = int(gtsam.symbol(ord('b'), 0))
-
-
-def X(key):
-    """Create symbol for pose key."""
-    return gtsam.symbol(ord('x'), key)
-
-
-def V(key):
-    """Create symbol for velocity key."""
-    return gtsam.symbol(ord('v'), key)
+BIAS_KEY = B(0)
 
 
 np.set_printoptions(precision=3, suppress=True)
--- a/gtsam.h
+++ b/gtsam.h
@@ -120,12 +120,6 @@
 
 namespace gtsam {
 
-// Include Key typedef
-#include <gtsam/base/types.h>
-class Key {
-  Key();
-};
-
 // Actually a FastList<Key>
 #include <gtsam/inference/Key.h>
 class KeyList {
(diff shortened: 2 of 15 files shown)


varunagrawal:
Also, your thoughts on replacing `size_t` with `Key` at various places would be great.

gtsam.h
```diff
@@ -120,6 +120,12 @@
 
 namespace gtsam {
 
+// Include Key typedef
+#include <gtsam/base/types.h>
+class Key {
```
dellaert, line 125: Let's not do that in this PR - I think it's actually not a good idea. I'll explain why in person

gtsam.h
```diff
@@ -1977,6 +1983,35 @@ size_t symbol(char chr, size_t index);
 char symbolChr(size_t key);
 size_t symbolIndex(size_t key);
 
+namespace symbol_shorthand {
```
dellaert, line 1986: was this tested? If so where?

varunagrawal: I tested this in an example. I want to port all the examples, so this PR is still technically WIP.

ProfFan:
LGTM except that I think we should not be putting something like an `X` in `gtsam`.

dellaert:
> LGTM except that I think we should not be putting something like an `X` in `gtsam`.

@varunagrawal confirm how it is used in python? qualified namespace I guess, right?

varunagrawal:
It's used exactly like the snippet in the description. That would mean that there is no qualified namespace, unfortunately.

I followed your advice and put up a cohesive, working example. :slightly_smiling_face: 
Any recommendations for how we should tackle this? I am just tired of duplicating the following everywhere

```python
def X(i):
    return gtsam.symbol(ord('x'), i)
```

ProfFan:
Maybe the sympy approach? `from gtsam.symbols import X`.

dellaert:
@varunagrawal Including X is good, as long as it's not a naked symbol. So that's why I am asking how you used it in your example.

dellaert:
I see now. yeah, unfortunately gtsam.X does not work. But I thought it *was* in a namespace. I'm confused.

@ProfFan will this be solved by pybind? Otherwise let's wait until that lands? When ? :-)

ProfFan:
@dellaert In Pybind we still need the `gtsam.symbol('x', 0)` thingy (no `ord` anymore). But I think it is good to add in `gtsam.utils`, like `gtsam.utils.symbols.X`.

ProfFan:
Since `utils` is pure Python~

dellaert:
In c++ it already *has* a namespace. Can't we use the same name?

cython/gtsam/examples/ImuFactorExample.py
```diff
@@ -12,25 +12,18 @@
 
 import math
 
+import gtsam
 import matplotlib.pyplot as plt
 import numpy as np
+from gtsam import symbol_shorthand_B as B
```
dellaert, line 18: beautiful!

PR #349, opened 2020-06-11
[RFC, WIP] Stabilize the New Pybind Wrapper

ProfFan:
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/349)


dellaert:
Can we add pybind11 in a different branch/PR, and make that branch the target of this PR?
Also, does this PR have the latest changes to the cython wrapper?

ProfFan:
> Can we add pybind11 in a different branch/PR, and make that branch the target of this PR?

Yes

> Also, does this PR have the latest changes to the cython wrapper?

No, I plan to add shortly

ProfFan:
May need to open another one, as the diff is not updated, idk why.

ProfFan:
@dellaert I think it is a bit hard to do that as I merged a lot of changes from develop, so GitHub's diff will show a lot of stuff from develop if I change the base to the one with only pybind.

dellaert:
your pybind branch should be branched from develop.

ProfFan:
@dellaert Yeah I tried that but GH is not updating the diff. I will try a different branch name and see if that works.

PR #358, opened 2020-06-18
Consistent Rot3 printing

varunagrawal:
This PR unifies the printing functionality of Rot3 to use the `Eigen::IOFormat` formatting, thus ensuring consistency.

The aim would be to help the user visually identify more easily what the rotation matrices are, as well as copy them into another language such as python.

I've also
1. updated the printing of `Pose3`, `NavState` and `Rot3` to provide default spacing when using the custom prefix string
2. removed extra newlines by removing newlines in the `<<` overload
3. removed `endl` in `<<` so as to avoid unnecessary buffer flushes.

For the below snippet,

```cpp
Pose3 p;
cout << p << endl;
p.print();
```

the previous format looked like this:

```sh
# result of cout
|1, 0, 0|
|0, 1, 0|
|0, 0, 1|

[0, 0, 0]';

# result of .print()
R:
[
	1, 0, 0;
 	0, 1, 0;
 	0, 0, 1
]
[0, 0, 0]';

```

The new format would give the output:

```sh
# result of cout
R: [
	1, 0, 0;
	0, 1, 0;
	0, 0, 1
]
t: [0, 0, 0]'
# result of .print()
R: [
	1, 0, 0;
	0, 1, 0;
	0, 0, 1
]
t: [0, 0, 0]'
```

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/358)


varunagrawal:
Additionally, this is how `Pose3` now prints

```
R: [
	1, 0, 0;
	0, 1, 0;
	0, 0, 1
]
t: [0, 0, 0]'
```

and this is for `NavState`

```
R: [
	1, 0, 0;
	0, 1, 0;
	0, 0, 1
]
p: [0, 0, 0]'
v: [0, 0, 0]'
```

dellaert:
Ask other people for reviews ?

dellaert:
> Ask other people for reviews ?

Dude. I don’t mean ask everyone at once :-)

gtsam/base/Matrix.h
```diff
@@ -76,6 +76,18 @@ GTSAM_MAKE_MATRIX_DEFS(9);
...
+  Eigen::StreamPrecision, // precision
+  Eigen::DontAlignCols, // flags set such that rowSpacers are not added
+  ", ", // coeffSeparator
+  ";\n", // rowSeparator
+  "\t",  // rowPrefix
+  "", // rowSuffix
+  "[\n", // matPrefix
+  "\n]" // matSuffix
+);
+
```
jlblancoc, line 90: Hmmm... I'm not comfortable with this symbol defined here in the header. It was safer as it was before, as a static var inside a smaller scope :-) 
My understanding (and experience) is that it *might* lead to link errors in some compilers/build types. 

Why not defining it as a static variable in the corresponding .cpp file, and make this a getter?

```
// In the .h

// Matrix formatting arguments when printing. Akin to Matlab style.
const Eigen::IOFormat &matlabFormat(); 
```

```
// in the .cpp

static const Eigen::IOFormat matlab(
      Eigen::StreamPrecision, // precision
      0, // flags
      ", ", // coeffSeparator
      ";\n", // rowSeparator
      "\t",  // rowPrefix
      "", // rowSuffix
      "[\n", // matPrefix
      "\n]" // matSuffix
      );

const Eigen::IOFormat &gtsam::matlabFormat() { return matlab; }
```

gtsam/geometry/Rot3.cpp
```diff
@@ -32,7 +32,8 @@ namespace gtsam {
 
 /* ************************************************************************* */
 void Rot3::print(const std::string& s) const {
-  gtsam::print((Matrix)matrix(), s);
+  cout << (s.empty() ? "R: " : s + " ");
+  gtsam::print((Matrix)matrix());
```
jlblancoc, line 36: prefer `static_cast<>()` instead of old C cast?

gtsam/navigation/NavState.cpp
```diff
@@ -88,15 +88,15 @@ Matrix7 NavState::matrix() const {
 
 //------------------------------------------------------------------------------
 ostream& operator<<(ostream& os, const NavState& state) {
-  os << "R:" << state.attitude();
-  os << "p:" << state.position() << endl;
-  os << "v:" << Point3(state.velocity()) << endl;
+  os << "R: " << state.attitude() << "\n";
+  os << "p: " << state.position() << "\n";
```
jlblancoc, line 92: :+1:

gtsam/base/Matrix.h
varunagrawal: That's an excellent point I had not considered, and the proposed solution is elegant! I'll get on it.

gtsam/geometry/Rot3.cpp
varunagrawal: Yes! I was trying not to introduce too many changes, but this is a very good change to introduce.

varunagrawal:
@dellaert sorry about that. I figured casting a wide net would yield faster results. :sweat_smile:

varunagrawal:
@dellaert @jlblancoc now that 4.0.3 is out, I would really like to see this PR merged in. :slightly_smiling_face:

varunagrawal:
Whoops. Yeah the other PR is based off this one hence you're seeing the issue repeat itself. Give me a few quick minutes to update this.

varunagrawal:
So I am a bit confused about the `endl` to `\n` conversion left here. From my understanding, we want to do a buffer flush when calling `print()` and that is the only place where we use `endl`. :-)

jlblancoc:
> So I am a bit confused about the `endl` to `\n` conversion left here. From my understanding, we want to do a buffer flush when calling `print()` and that is the only place where we use `endl`. :-)

Hmm... I'm thinking of, say, we call print() on a Values or an entire FactorGraph. One flush at the end is ok... but 1000s of them, after printing (recursively) each value/matrix/vector/factor/etc. seems not necessary :-) 

Still, it's probably a minor point, so feel free of leaving it as you think it's better.

varunagrawal:
I completely agree with your point. The issue is that the Values/FactorGraph print method just calls the internal print method of the various values/factors, so we need to add the `endl` over there. I think that should be left to a future PR?

jlblancoc:
Ok. I don't want to stop you doing anything more productive! 
Merge it when you want to.

varunagrawal:
Merging because Travis keeps timing out and we will be moving to Github Actions.

PR #361, opened 2020-06-21
Fixed all alignment problems

ProfFan:
This is a **BIG** change so I choose to make a separate branch:

## Pitch

We have been fighting with alignment problems for so long, especially when we do not have access to C++17. Normally we just add `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` and these stuff will go away. However for the Cython wrapper (and similar things using `make_shared`) this is not enough-it will use the unaligned version regardless of the existence of an aligned version. This PR is my answer to this long-hanging issue.

## Explanation

I borrowed the idea from PCL (https://github.com/PointCloudLibrary/pcl/pull/3163). This is basically a polyfill for not have the C++17 alignment-aware new, using very simple templated metaprogramming.

## API impact

None, to switch to the aligned version, one only need to change `boost::make_shared` to `gtsam::make_shared` and add `GTSAM_MAKE_ALIGNED_OPERATOR_NEW` to his classes.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/361)

### DIFF
--- a/gtsam.h
+++ b/gtsam.h
@@ -281,7 +281,7 @@ virtual class Value {
 };
 
 #include <gtsam/base/GenericValue.h>
-template<T = {Vector, gtsam::Point2, gtsam::Point3, gtsam::Rot2, gtsam::Rot3, gtsam::Pose2, gtsam::Pose3, gtsam::StereoPoint2, gtsam::Cal3_S2, gtsam::CalibratedCamera, gtsam::SimpleCamera, gtsam::imuBias::ConstantBias}>
+template<T = {Vector, Matrix, gtsam::Point2, gtsam::Point3, gtsam::Rot2, gtsam::Rot3, gtsam::Pose2, gtsam::Pose3, gtsam::StereoPoint2, gtsam::Cal3_S2, gtsam::Cal3DS2, gtsam::Cal3Bundler, gtsam::EssentialMatrix, gtsam::CalibratedCamera, gtsam::SimpleCamera, gtsam::imuBias::ConstantBias}>
 virtual class GenericValue : gtsam::Value {
   void serializable() const;
 };
--- a/gtsam/base/make_shared.h
+++ b/gtsam/base/make_shared.h
@@ -20,7 +20,7 @@
 #include <boost/make_shared.hpp>
 #include <gtsam/base/types.h>
 
-#include <type_traits>
+#include <boost/type_traits.hpp>
 
 #include <Eigen/Core>
 
--- a/gtsam/base/make_shared.h
+++ b/gtsam/base/make_shared.h
@@ -17,24 +17,24 @@
 
 #pragma once
 
-#include <boost/make_shared.hpp>
 #include <gtsam/base/types.h>
-
-#include <boost/type_traits.hpp>
 
 #include <Eigen/Core>
 
+#include <boost/make_shared.hpp>
+#include <boost/utility/enable_if.hpp>
+
+
 namespace gtsam {
 
-  /**
-   * And our own `make_shared` as a layer of wrapping on `boost::make_shared`
-   */
+  /// Add our own `make_shared` as a layer of wrapping on `boost::make_shared`
   template<typename T, typename ... Args>
   boost::enable_if_t<has_custom_allocator<T>::value, boost::shared_ptr<T>> make_shared(Args&&... args)
   {
     return boost::allocate_shared<T>(Eigen::aligned_allocator<T>(), std::forward<Args> (args)...);
   }
 
+  /// Fall back to the boost version if no need for alignment
   template<typename T, typename ... Args>
   boost::enable_if_t<!has_custom_allocator<T>::value, boost::shared_ptr<T>> make_shared(Args&&... args)
   {
--- a/gtsam/base/make_shared.h
+++ b/gtsam/base/make_shared.h
@@ -1,0 +1,44 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2020, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file make_shared.h
+ * @brief make_shared trampoline function to ensure proper alignment
+ * @author Fan Jiang
+ */
+
+#pragma once
+
+#include <boost/make_shared.hpp>
+#include <gtsam/base/types.h>
+
+#include <type_traits>
+
+#include <Eigen/Core>
+
+namespace gtsam {
+
+  /**
+   * And our own `make_shared` as a layer of wrapping on `boost::make_shared`
+   */
+  template<typename T, typename ... Args>
+  boost::enable_if_t<has_custom_allocator<T>::value, boost::shared_ptr<T>> make_shared(Args&&... args)
+  {
+    return boost::allocate_shared<T>(Eigen::aligned_allocator<T>(), std::forward<Args> (args)...);
+  }
+
+  template<typename T, typename ... Args>
+  boost::enable_if_t<!has_custom_allocator<T>::value, boost::shared_ptr<T>> make_shared(Args&&... args)
+  {
+    return boost::make_shared<T>(std::forward<Args> (args)...);
+  }
+
+}
--- a/gtsam/base/types.h
+++ b/gtsam/base/types.h
@@ -230,3 +230,29 @@ namespace std {
 #ifdef ERROR
 #undef ERROR
 #endif
+
+namespace gtsam {
+
+  /**
+   * A trait to mark classes that need special alignment.
+   * Please refer to https://github.com/PointCloudLibrary/pcl/pull/3163
+   */
+#if __cplusplus < 201703L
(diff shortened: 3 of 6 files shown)


ProfFan:
The most important thing this PR fixes is the Cython wrapper. As the cython wrapper uses `make_shared` for all functions returning a concrete object (like `SO4 SO4::Expmap(...)`, the alignment error is inevitable without this PR.

gtsam.h
```diff
@@ -281,7 +281,7 @@ virtual class Value {
 };
 
 #include <gtsam/base/GenericValue.h>
-template<T = {Vector, gtsam::Point2, gtsam::Point3, gtsam::Rot2, gtsam::Rot3, gtsam::Pose2, gtsam::Pose3, gtsam::StereoPoint2, gtsam::Cal3_S2, gtsam::CalibratedCamera, gtsam::SimpleCamera, gtsam::imuBias::ConstantBias}>
+template<T = {Vector, Matrix, gtsam::Point2, gtsam::Point3, gtsam::Rot2, gtsam::Rot3, gtsam::Pose2, gtsam::Pose3, gtsam::StereoPoint2, gtsam::Cal3_S2, gtsam::Cal3DS2, gtsam::Cal3Bundler, gtsam::EssentialMatrix, gtsam::CalibratedCamera, gtsam::SimpleCamera, gtsam::imuBias::ConstantBias}>
```
dellaert, line 284: rationale for this change?

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,44 @@
...
+#include <boost/make_shared.hpp>
+#include <gtsam/base/types.h>
+
+#include <boost/type_traits.hpp>
+
+#include <Eigen/Core>
+
+namespace gtsam {
+
+  /**
```
dellaert, line 29: use /// for single-liners

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @file make_shared.h
+ * @brief make_shared trampoline function to ensure proper alignment
+ * @author Fan Jiang
+ */
+
+#pragma once
+
+#include <boost/make_shared.hpp>
```
dellaert, line 20: re-order includes the gtsam way: gtsam, then Eigen, then boost, then std

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * And our own `make_shared` as a layer of wrapping on `boost::make_shared`
+   */
+  template<typename T, typename ... Args>
+  boost::enable_if_t<has_custom_allocator<T>::value, boost::shared_ptr<T>> make_shared(Args&&... args)
+  {
+    return boost::allocate_shared<T>(Eigen::aligned_allocator<T>(), std::forward<Args> (args)...);
+  }
+
+  template<typename T, typename ... Args>
```
dellaert, line 38: comment

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,44 @@
...
+  boost::enable_if_t<!has_custom_allocator<T>::value, boost::shared_ptr<T>> make_shared(Args&&... args)
+  {
+    return boost::make_shared<T>(std::forward<Args> (args)...);
+  }
+
+}
```
dellaert, line 44: add missing newline

gtsam/base/types.h
```diff
@@ -230,3 +230,29 @@ namespace std {
 #ifdef ERROR
 #undef ERROR
 #endif
+
+namespace gtsam {
+
+  /**
+   * A trait to mark classes that need special alignment.
+   * Please refer to https://github.com/PointCloudLibrary/pcl/pull/3163
```
dellaert, line 238: how about adding self-contained documentation here. Also say when and how we should use it.

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,44 @@
...
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file make_shared.h
+ * @brief make_shared trampoline function to ensure proper alignment
```
dellaert, line 14: need more than brief - can't expect readers of code to refer to PR in which this was proposed.

gtsam/base/types.h
```diff
@@ -230,3 +230,29 @@ namespace std {
...
+  };
+
+}
+
+/**
+ * This is necessary for the Cython wrapper to work properly, and possibly reduce future misalignment problems.
+ */
+#define GTSAM_MAKE_ALIGNED_OPERATOR_NEW \
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW \
+  using _custom_allocator_type_trait = void;
```
dellaert, line 258: add missing newline

gtsam/base/types.h
```diff
@@ -230,3 +230,29 @@ namespace std {
...
+  template<typename T>
+  struct has_custom_allocator<T, void_t<typename T::_custom_allocator_type_trait>> : std::true_type {
+  };
+
+}
+
+/**
+ * This is necessary for the Cython wrapper to work properly, and possibly reduce future misalignment problems.
+ */
+#define GTSAM_MAKE_ALIGNED_OPERATOR_NEW \
```
dellaert, line 256: I see ! So, that's how gtsam::make_shared does the right thing :-)

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,44 @@
...
+#include <gtsam/base/types.h>
+
+#include <boost/type_traits.hpp>
+
+#include <Eigen/Core>
+
+namespace gtsam {
+
+  /**
+   * And our own `make_shared` as a layer of wrapping on `boost::make_shared`
```
dellaert, line 30: And -> Add

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,44 @@
...
+   */
+  template<typename T, typename ... Args>
+  boost::enable_if_t<has_custom_allocator<T>::value, boost::shared_ptr<T>> make_shared(Args&&... args)
```
dellaert, line 33: in c++11, we have enable_if no?  No need for boost or its header then...

gtsam/geometry/SOn.h
```diff
@@ -54,7 +54,7 @@ class SO : public LieGroup<SO<N>, internal::DimensionSO(N)> {
   using VectorN2 = Eigen::Matrix<double, internal::NSquaredSO(N), 1>;
   using MatrixDD = Eigen::Matrix<double, dimension, dimension>;
 
-  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+  GTSAM_MAKE_ALIGNED_OPERATOR_NEW
```
dellaert, line 57: Should we then replace *all* these??

wrap/Module.cpp
```diff
@@ -350,7 +350,10 @@ void Module::emit_cython_pxd(FileWriter& pxdFile) const {
                  "        T* get()\n"
                  "        long use_count() const\n"
                  "        T& operator*()\n\n"
-                 "    cdef shared_ptr[T] dynamic_pointer_cast[T,U](const shared_ptr[U]& r)\n"
+                 "    cdef shared_ptr[T] dynamic_pointer_cast[T,U](const shared_ptr[U]& r)\n\n";
+
+  // gtsam alignment-friendly shared_ptr
```
dellaert, line 355: how does this guarantee the correct make_shared is used everywhere?

dellaert:
PS this should have been branched from *develop*, right? I propose you do *not* force-push, as it's against our rule, but address my comments, then start a new branch and cherry-pick, then create a new PR. Then this PR and branch can be abandoned...

gtsam.h
ProfFan: These are used in the Python code implicitly as `GenericValue`s, but the templates are not instantiated. This leads to an error when loading the Python module on my Mac.

gtsam/base/make_shared.h
ProfFan: Since we are still on boost I think it's better to use the boost version for now, and switch when we switch all to std.

gtsam/geometry/SOn.h
ProfFan: Yeah, this is still RFC so I didn't replace everything. I can make another PR that does this replacement after this is merged.

wrap/Module.cpp
ProfFan: If you take a look at the generated `gtsam.cpp` you will see that `boost::make_shared` is replace by this everywhere.

gtsam/base/make_shared.h
dellaert: I disagree. Let's wean ourselves from boost. I think all other uses of enable_if are already std::

gtsam/geometry/SOn.h
dellaert: See my comment below about branch from develop. I propose that in the new PR you replace all...

ProfFan: Sure, will do.

gtsam/base/make_shared.h
ProfFan: Copy

ProfFan:
I'll close this and deal with the comments in #363 .

Issue #283, opened 2020-04-28
Question about OrientedPlaneFactor implementation


clams-casino:
Hi, I'm trying to implement a slam program based on plane landmarks and the OrientedPlaneFactor. I have a question about the way the jacobians are computed in the evaluateError() method in the factor. 

```
virtual Vector evaluateError(const Pose3& pose, const OrientedPlane3& plane,
      boost::optional<Matrix&> H1 = boost::none, boost::optional<Matrix&> H2 =
          boost::none) const {
    OrientedPlane3 predicted_plane = OrientedPlane3::Transform(plane, pose, H1, H2);
    Vector err(3);
    err << predicted_plane.error(measured_p_);
    return (err);
}
```
So the jacobians for the predicted plane with respect to the pose and plane variables are computed, but the jacobians for the error with respect to the predicted plane is not.

There are two methods defined for differencing two OrientedPlane3s, error() and errorVector() which also computes jacobians.

```
  /** Computes the error between two planes.
   *  The error is a norm 1 difference in tangent space.
   * @param the other plane
   */
  Vector3 error(const OrientedPlane3& plane) const;

  /** Computes the error between the two planes, with derivatives.
   *  This uses Unit3::errorVector, as opposed to the other .error() in this class, which uses
   *  Unit3::localCoordinates. This one has correct derivatives.
   *  NOTE(hayk): The derivatives are zero when normals are exactly orthogonal.
   * @param the other plane
   */
  Vector3 errorVector(const OrientedPlane3& other, OptionalJacobian<3, 3> H1 = boost::none, //
                      OptionalJacobian<3, 3> H2 = boost::none) const;
```

Why does the OrientedPlaneFactor's evaluateError() not use the errorVector() which also gives the jacobian of the error with respect to the predicted plane?

dellaert:
Yes, *if* the comment is correct and .error has incorrect derivatives this should probably be changed. I don't want to break any existing code someone might have though, so this should probably only be changed if the unit tests still work and the convergence behavior is the same (or better). Feel free to investigate and do a PR, if you're up for it.

jlblancoc:
Maybe related: in #183 , unit test thresholds for  `OrientedPlane3` had to be raised from 1e-9 to 1e-5 for them to pass in x32 builds. I thought it was just a low-level numerical accuracy thing, but *might* be something else if you found an inconsistency in the formulation...

dwisth:
[...]

PR #362, opened 2020-06-22
Fix OrientedPlane3Factor jacobian

dwisth:
See #283.

This PR fixes the derivative of the `OrientedPlane3Factor`. Previously, the calculated Jacobians didn't account for the second function in `evaluateError` which meant that the Jacobian was not correct.

I have added a `numericalDerivative` to calculate the Jacobian of the second calculation and then applied the chain rule. I have also added a unit test to check that this is correct.

TODO: Ideally `OrientedPlane3::error()` should calculate analytical derivatives itself instead of using `numericalDerivatives`. Would anyone be able to help with this?

@mcamurri

### DIFF
--- a/gtsam/geometry/OrientedPlane3.h
+++ b/gtsam/geometry/OrientedPlane3.h
@@ -114,6 +114,10 @@ public:
    */
   Vector3 error(const OrientedPlane3& plane) const;
 
+  static Vector3 Error(const OrientedPlane3& plane1, const OrientedPlane3& plane2) {
+    return plane1.error(plane2);
+  }
+
   /** Computes the error between the two planes, with derivatives.
    *  This uses Unit3::errorVector, as opposed to the other .error() in this class, which uses
    *  Unit3::localCoordinates. This one has correct derivatives.
--- a/gtsam/slam/OrientedPlane3Factor.h
+++ b/gtsam/slam/OrientedPlane3Factor.h
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <gtsam/base/numericalDerivative.h>
 #include <gtsam/geometry/OrientedPlane3.h>
 #include <gtsam/nonlinear/NonlinearFactor.h>
 
@@ -48,10 +49,27 @@ public:
   virtual Vector evaluateError(const Pose3& pose, const OrientedPlane3& plane,
       boost::optional<Matrix&> H1 = boost::none, boost::optional<Matrix&> H2 =
           boost::none) const {
-    OrientedPlane3 predicted_plane = OrientedPlane3::Transform(plane, pose, H1,
-        H2);
     Vector err(3);
-    err << predicted_plane.error(measured_p_);
+
+    if (H1 || H2) {
+      Matrix H1_1, H2_1;
+      OrientedPlane3 predicted_plane = OrientedPlane3::Transform(plane, pose, H1_1, H2_1);
+      err << predicted_plane.error(measured_p_);
+      // Numerically calculate the derivative since this function doesn't provide one.
+      auto f = boost::bind(&OrientedPlane3::Error, _1, _2);
+      Matrix H1_2 = numericalDerivative21<Vector3, OrientedPlane3, OrientedPlane3>(f, predicted_plane, measured_p_);
+
+      // Apply the chain rule to calculate the derivatives.
+      if (H1) {
+        *H1 = H1_2 * H1_1;
+      }
+      if (H2) {
+        *H2 = H1_2 * H2_1;
+      }
+    } else {
+      OrientedPlane3 predicted_plane = OrientedPlane3::Transform(plane, pose, H1, H2);
+      err << predicted_plane.error(measured_p_);
+    }
     return (err);
   }
   ;
--- a/gtsam/slam/tests/testOrientedPlane3Factor.cpp
+++ b/gtsam/slam/tests/testOrientedPlane3Factor.cpp
@@ -122,6 +122,42 @@ TEST (OrientedPlane3Factor, lm_rotation_error) {
   EXPECT(assert_equal(optimized_plane_landmark, expected_plane_landmark));
 }
 
+double randDouble(double max = 1) {
+  return static_cast<double>(rand()) / RAND_MAX * max;
+}
+
+TEST( OrientedPlane3Factor, Derivatives ) {
+  for (int i=0; i<100; i++) {
+    // Random measurement
+    OrientedPlane3 p(randDouble(), randDouble(), randDouble(), randDouble());
+
+    // Random linearisation point
+    OrientedPlane3 pLin(randDouble(), randDouble(), randDouble(), randDouble());
+    gtsam::Point3 pointLin  = gtsam::Point3(randDouble(100), randDouble(100), randDouble(100));
+    gtsam::Rot3 rotationLin = gtsam::Rot3::RzRyRx(randDouble(2*M_PI), randDouble(2*M_PI), randDouble(2*M_PI));
+    Pose3 poseLin(rotationLin, pointLin);
+
+    // Factor
+    Key planeKey(1), poseKey(2);
+    SharedGaussian noise = noiseModel::Diagonal::Sigmas(Vector3(0.1, 0.1, 0.1));
+    OrientedPlane3Factor factor(p.planeCoefficients(), noise, poseKey, planeKey);
+
+    // Calculate numerical derivatives
+    boost::function<Vector(const Pose3&, const OrientedPlane3&)> f = boost::bind(
+        &OrientedPlane3Factor::evaluateError, factor, _1, _2, boost::none, boost::none);
+    Matrix numericalH1 = numericalDerivative21<Vector, Pose3, OrientedPlane3>(f, poseLin, pLin);
+    Matrix numericalH2 = numericalDerivative22<Vector, Pose3, OrientedPlane3>(f, poseLin, pLin);
+
+    // Use the factor to calculate the derivative
+    Matrix actualH1, actualH2;
+    factor.evaluateError(poseLin, pLin, actualH1, actualH2);
+
+    // Verify we get the expected error
+    EXPECT(assert_equal(numericalH1, actualH1, 1e-8));
+    EXPECT(assert_equal(numericalH2, actualH2, 1e-8));
+  }
+}
+
 // *************************************************************************
 TEST( OrientedPlane3DirectionPrior, Constructor ) {
 
--- a/gtsam/slam/tests/testOrientedPlane3Factor.cpp
+++ b/gtsam/slam/tests/testOrientedPlane3Factor.cpp
@@ -83,7 +83,7 @@ TEST (OrientedPlane3Factor, lm_rotation_error) {
   // Tests one pose, two measurements of the landmark that differ in angle only.
   // Normal along -x, 3m away
   Symbol lm_sym('p', 0);
-  OrientedPlane3 test_lm0(-1.0, 0.0, 0.0, 3.0);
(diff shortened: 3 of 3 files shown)


gtsam/geometry/OrientedPlane3.h
```diff
@@ -114,6 +114,10 @@ class GTSAM_EXPORT OrientedPlane3 {
    */
   Vector3 error(const OrientedPlane3& plane) const;
 
+  static Vector3 Error(const OrientedPlane3& plane1, const OrientedPlane3& plane2) {
```
dellaert, line 117: I don't think we should add this. Instead, add 2 OptionalJacobian arguments to method above (H1 for this, H2 for plane)

gtsam/slam/OrientedPlane3Factor.h
```diff
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <gtsam/base/numericalDerivative.h>
```
dellaert, line 10: Let's move this to OrientedPlane.cpp

gtsam/slam/OrientedPlane3Factor.h
```diff
@@ -48,10 +49,27 @@ class OrientedPlane3Factor: public NoiseModelFactor2<Pose3, OrientedPlane3> {
...
-        H2);
     Vector err(3);
-    err << predicted_plane.error(measured_p_);
+
+    if (H1 || H2) {
+      Matrix H1_1, H2_1;
+      OrientedPlane3 predicted_plane = OrientedPlane3::Transform(plane, pose, H1_1, H2_1);
+      err << predicted_plane.error(measured_p_);
+      // Numerically calculate the derivative since this function doesn't provide one.
+      auto f = boost::bind(&OrientedPlane3::Error, _1, _2);
```
dellaert, line 59: move this logic into OrientedPlane.cpp, to calculate the optional derivatives of the error method, then call

```
Matrix33 error_H_predicted, error_H_measured;
err << predicted.error(measured_p_, error_H_predicted, error_H_measured);
```

gtsam/slam/OrientedPlane3Factor.h
```diff
@@ -48,10 +49,27 @@ class OrientedPlane3Factor: public NoiseModelFactor2<Pose3, OrientedPlane3> {
   virtual Vector evaluateError(const Pose3& pose, const OrientedPlane3& plane,
       boost::optional<Matrix&> H1 = boost::none, boost::optional<Matrix&> H2 =
           boost::none) const {
-    OrientedPlane3 predicted_plane = OrientedPlane3::Transform(plane, pose, H1,
-        H2);
     Vector err(3);
-    err << predicted_plane.error(measured_p_);
+
+    if (H1 || H2) {
+      Matrix H1_1, H2_1;
```
dellaert, line 55: So, the static method is deprecated (wrong Jacobian order), so please use the method. Give these better names and make them fixed-size matrices, like

```
Matrix33 predicted_H_plane;
Matrix36 predicted_H_pose;
auto predicted = plane.transform(pose, predicted_H_plane, predicted_H_pose);
```

gtsam/slam/tests/testOrientedPlane3Factor.cpp
```diff
@@ -122,6 +122,42 @@ TEST (OrientedPlane3Factor, lm_rotation_error) {
   EXPECT(assert_equal(optimized_plane_landmark, expected_plane_landmark));
 }
 
+double randDouble(double max = 1) {
```
dellaert, line 125: We really shouldn't have random unit tests. Just find a simple case that fails.

akshay-krishnan:
Yes, I can help with the analytical derivative. Looks like we will need to define the derivatives for  Unit3’s localCoordinates, since OrientedPlane3’s error function makes use of it. 

About the [error method](https://github.com/borglab/gtsam/blob/75eb859ee7dc2ea0ff44ef7b26431f115a0d2e7e/gtsam/geometry/OrientedPlane3.h#L115) in OrientedPlane3, I’m not sure what the norm 1 error is, but if this is the L1 norm, should it not be the absolute value? Are we missing an std::abs [here](https://github.com/borglab/gtsam/blob/75eb859ee7dc2ea0ff44ef7b26431f115a0d2e7e/gtsam/geometry/OrientedPlane3.cpp#L63)?

dwisth:
Thanks for all the detailed feedback @dellaert! I've incorporated almost all the changes you mentioned into the latest commit. Once we fix the derivatives we can remove `OrientedPlane3::Error`
 method.

@akshay-krishnan - Thanks for your help :) The call to `numericalDerivative` is now in `OrientedPlane3::error`, which we will remove once `Unit3::localCoordinates` can output it's derivatives.

Also, I'm not sure about the comment about the L1 norm..

gtsam/slam/OrientedPlane3Factor.cpp
```diff
@@ -19,6 +19,33 @@ void OrientedPlane3Factor::print(const string& s,
...
+Vector OrientedPlane3Factor::evaluateError(const Pose3& pose,
+    const OrientedPlane3& plane, boost::optional<Matrix&> H1,
+    boost::optional<Matrix&> H2) const {
+  Vector err(3);
+
+  if (H1 || H2) {
+    Matrix36 predicted_H_pose;
+    Matrix33 predicted_H_plane, error_H_predicted;
+
+    OrientedPlane3 predicted_plane = plane.transform(pose, predicted_H_plane, predicted_H_pose);
```
dellaert, line 32: There is another trick here you can use to only compute derivatives when needed:

```
plane.transform(pose, H2? &predicted_H_plane : nullptr, H1 ? &predicted_H_pose : nullptr);
```

And similarly below.

Then you don’t need to branch on H1 || H2

dwisth:
I updated the style (as suggested by Frank) and changed from `OrientedPlane3::error` to `OrientedPlane3::errorVector`. If I change this the `OrientedPlane3Factor, lm_rotation_error` unit test fails. It seems that the two functions give very different error values and derivatives.

E.g. Here is the error and derivatives for `OrientedPlane3::error()`:
```
Error:
-0.0996687
        -0
         0
H1:
         0          0          1          0          0          0
 -0.099174    0.99174          0          0          0          0
         0          0          0  -0.995037 -0.0995037          0
H2:
       1        0        0
       0 0.996687        0
       0        0        1
```
And here is the error and derivatives for `OrientedPlane3::errorVector()` with the same inputs:
```
Error:
0.0995037
        0
        0
H1:
         0          0  -0.995037          0          0          0
 0.0990099  -0.990099          0          0          0          0
         0          0          0  -0.995037 -0.0995037          0
H2:
-0.995037         0         0
        0 -0.995037         0
        0         0         1
```

I'm not sure where to go from here - @akshay-krishnan any ideas?

akshay-krishnan:
The error and errorVector functions do not do the same thing. One is the negative of the other. I believe this was designed to be this way. If you look at the source, errorVector computes (this - other) where as error computes (other - this) in tangential space (like localCoordinates). 

I am still confused about the fact that neither of them have an absolute value.

dellaert:
So, any factor calculates a vector of doubles, which GTSAM then takes the norm of to calculate the error. So, the sign of the individual error entries does not matter.

jlblancoc:
My two cents: negative numbers are perfectly OK in an error vector. But I think both formulas, the one used to error evaluation and the one used to derive the Jacobians, should be the same (with same sign: both "a-b", not "a-b" vs "b-a"), or the nonlinear optimizer may be misled...

dellaert:
Hmmm. The way you used errorVector seems correct. Can you confirm that (a) the errorVector derivative unit test succeeds, and (b) print out the output of the failing evaluateError test?

dellaert:
> My two cents: negative numbers are perfectly OK in an error vector. But I think both formulas, the one used to error evaluation and the one used to derive the Jacobians, should be the same (with same sign: both "a-b", not "a-b" vs "b-a"), or the nonlinear optimizer may be misled...

I think David is doing this. Is there a specific line you think is wrong?

jlblancoc:
> > My two cents: negative numbers are perfectly OK in an error vector. But I think both formulas, the one used to error evaluation and the one used to derive the Jacobians, should be the same (with same sign: both "a-b", not "a-b" vs "b-a"), or the nonlinear optimizer may be misled...
> 
> I think David is doing this. Is there a specific line you think is wrong?

@dellaert : Honestly I didn't reviewed the code, I was just answering to: 

> The error and errorVector functions do not do the same thing. One is the negative of the other. I believe this was designed to be this way. If you look at the source, errorVector computes (this - other) where as error computes (other - this) in tangential space (like localCoordinates).

which seemed suspicious at first sight :-)

dellaert:
So, what's the status on this now? @dwisth @akshay-krishnan

akshay-krishnan:
> So, what's the status on this now? @dwisth @akshay-krishnan

Sorry, I have not found time to work on the Jacobians for Unit3::error. Could @dwisth confirm if the errorVector unit tests worked?

dellaert:
If you want this in 4.0.3, last checkpoint before 4.1, we need to clean up this PR now. It seems that switching to the other error made things work @dwisth ?

dwisth:
Yes, seemed that switching error terms made it work. I'm going to take another look at this today / tomorrow and try to finish it off.

dellaert:
Thanks for your latest commit @dwisth . But now that you switched error functions, do we still *need* the analytical Jacobian for localCoordinates? If not I'd rather remove the numerical versions and their unit tests in this PR.

dwisth:
To clarify the status is:

dellaert:
> * The only remaining thing to solve is the analytical derivative for `Unit3::localCoordinates()`.
> I'm a bit stuck on this last point and if anyone can help or give me some pointers it would be greatly appreciated.

Awesome. But, again:  now that you switched error functions, do we still need the analytical Jacobian for localCoordinates? If not I'd rather remove the numerical versions and their unit tests in this PR.

dwisth:
> Awesome. But, again: now that you switched error functions, do we still need the analytical Jacobian for localCoordinates? If not I'd rather remove the numerical versions and their unit tests in this PR.

Yes we do. In the end I found the the original error function worked best (`OrientedPlane3::error()`) but we just need to fix the Jacobian. This requires the derivative for `Unit3::localCoordinates`.

varunagrawal:
So here are my thoughts from looking into this issue:

1. There is clearly a bug in `errorVector` for `OrientedPlane3`. There should be a negative sign for the value of `n_error`. This is because `error` and `errorVector` are equivalent, just the later provides us with jacobians.
2. The `Unit3` class has different definitions for `error`/`errorVector` and `localCoordinates`. I am not sure why this is the case since `localCoordinates` effectively computes `other - this` in the tangent space. The unit tests for `localCoordinates` make sense e.g. `localCoordinates((1, 0, 0), (-1, 0, 0))` gives `pi, 0` which is the angle and magnitude difference. `error((1, 0, 0), (-1, 0, 0))` gives us `0, 0` which doesn't make sense to me because these are not the same unit vectors.

varunagrawal:
I believe if we can figure out `Unit3`'s localCoordinate jacobians (which are actually easy to compute) and resolve the ambiguity with `error`/`errorVector` we should be pretty much done.

Also, `Unit3::error` has been marked deprecated, but IMHO we should just have the one `error` function which has the signature and functionality of `errorVector`.

dellaert:
This was never merged. I propose I merge it into the other branch and do all plane-related stuff at once.

dellaert (approved):
Merging into other branch to consolidate

varunagrawal:
@dellaert be sure to keep track of my comments. We need the Unit3 jacobians implemented. :)

dellaert:
> @dellaert be sure to keep track of my comments. We need the Unit3 jacobians implemented. :)

Maybe. But I'll keep your comments in mind.

PR #363, opened 2020-06-22
[v2] Fix all alignment issues when using make_shared

ProfFan:
#361 but based on develop

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/363)


ProfFan:
Travis is failing because numpy now requires Python 3.6...

ProfFan:
@dellaert @varunagrawal Could you do a quick review on this? I would still like to convert other classes to use `GTSAM_MAKE_ALIGNED` later, as we need this to get Frob PR #360  merged...

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,51 @@
...
+#if __cplusplus <= 201103L
+namespace gtsam {
+  template<bool B, class T = void>
+  using enable_if_t = typename std::enable_if<B, T>::type;
+}
+#endif
+
+namespace gtsam {
+
+  /// Add our own `make_shared` as a layer of wrapping on `boost::make_shared`
```
dellaert, line 37: Explain why, where this will be called, and what problem it is solving

gtsam/base/types.h
```diff
@@ -230,3 +230,30 @@ namespace std {
...
+  };
+
+}
+
+/**
+ * This is necessary for the Cython wrapper to work properly, and possibly reduce future misalignment problems.
+ */
+#define GTSAM_MAKE_ALIGNED_OPERATOR_NEW \
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW \
+  using _custom_allocator_type_trait = void;
```
dellaert, line 259: add newline

gtsam/base/types.h
```diff
@@ -230,3 +230,30 @@ namespace std {
...
 #endif
+
+namespace gtsam {
+
+  /**
+   * A trait to mark classes that need special alignment.
+   * This is required to make boost::make_shared and etc respect alignment, which is essential for the Python
+   * wrappers to work properly.
+   */
+#if __cplusplus < 201703L
```
dellaert, line 241: Is c++17 not backwards compatible with c++11? If so, is this not superfluous? If not, we would need to add CI to ensure compatibility with c++17, which we are not doing at the moment, and so this is probably not needed, either.

gtsam/base/types.h
```diff
@@ -230,3 +230,30 @@ namespace std {
...
+  template<typename ...> using void_t = void;
+#endif
+
+  template<typename, typename = void_t<>>
+  struct has_custom_allocator : std::false_type {
```
dellaert, line 246: The name `has_custom_allocator` is misleading. Really, above you are using `Eigen::aligned_allocator`.  I propose 

has_custom_allocator -> needs_eigen_aligned_allocator

_custom_allocator_type_trait -> _eigen_aligned_allocator_trait

gtsam/base/types.h
```diff
@@ -230,3 +230,30 @@ namespace std {
...
+  /**
+   * A trait to mark classes that need special alignment.
+   * This is required to make boost::make_shared and etc respect alignment, which is essential for the Python
+   * wrappers to work properly.
+   */
+#if __cplusplus < 201703L
+  template<typename ...> using void_t = void;
+#endif
+
+  template<typename, typename = void_t<>>
```
dellaert, line 245: Explain how it works, move the macro below *above* this so you can refer to it in your explanation

gtsam/base/types.h
```diff
@@ -230,3 +230,30 @@ namespace std {
...
+  struct has_custom_allocator : std::false_type {
+  };
+  template<typename T>
+  struct has_custom_allocator<T, void_t<typename T::_custom_allocator_type_trait>> : std::true_type {
+  };
+
+}
+
+/**
+ * This is necessary for the Cython wrapper to work properly, and possibly reduce future misalignment problems.
```
dellaert, line 255: This says why it is needed, but not how it works, and who is supposed to use it when. A link to the Eigen page on alignment would be nice, and finally, make sure you clang-format.

dellaert:
if you want to merge it, you need to do it right :-) I just left a more extensive review.
GitHub <notifications@github.com> wrote:
“@dellaert@varunagrawal Could you do a quick review on this? I would still like to convert other classes to use GTSAM_MAKE_ALIGNED later, as we need this to get Frob PR #360 merged...”

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <gtsam/base/types.h>
+
+#include <Eigen/Core>
+
+#include <boost/make_shared.hpp>
+
+#include <type_traits>
+
+#if __cplusplus <= 201103L
```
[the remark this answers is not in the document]
ProfFan: I removed the ifdef, since we are doing this in our own namespace there is no need to.

gtsam/base/make_shared.h
ProfFan: Done

gtsam/base/types.h
ProfFan: done

gtsam/base/types.h
ProfFan: I removed the ifdef, since we are doing this in our own namespace there is no need to.

gtsam/base/types.h
ProfFan: done

gtsam/base/types.h
ProfFan: I added docs below

gtsam/base/types.h
ProfFan: done

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <gtsam/base/types.h>
+
+#include <Eigen/Core>
+
+#include <boost/make_shared.hpp>
+
+#include <type_traits>
+
+namespace gtsam {
```
dellaert, line 28: document

gtsam/base/make_shared.h
```diff
@@ -0,0 +1,55 @@
...
+  template<bool B, class T = void>
+  using enable_if_t = typename std::enable_if<B, T>::type;
+}
+
+namespace gtsam {
+
+  /** Add our own `make_shared` as a layer of wrapping on `boost::make_shared`
+   * This solves the problem with the stock `make_shared` that custom alignment is not respected, causing SEGFAULTs
+   * at runtime, which is notoriously hard to debug.
+   *
```
dellaert, line 38: still does not make connection with type trait.  Say that `needs_eigen_aligned_allocator<T>::value` will evaluate to True if `using _eigen_aligned_allocator_trait = void` is present in class, and that this is automatically added by the `GTSAM_MAKE_ALIGNED_OPERATOR_NEW` macro.

gtsam/base/types.h
```diff
@@ -230,3 +230,31 @@ namespace std {
...
 #undef ERROR
 #endif
+
+namespace gtsam {
+
+  /// Convenience void_t as we assume C++11, it will not conflict the std one in C++17 as this is in `gtsam::`
+  template<typename ...> using void_t = void;
+
+  /**
+   * A SFINAE trait to mark classes that need special alignment.
```
dellaert, line 240: Explain :-)

gtsam/base/make_shared.h
ProfFan: done

gtsam/base/make_shared.h
ProfFan: done

gtsam/base/types.h
ProfFan: done

ProfFan:
@dellaert  done :)

PR #364, opened 2020-06-22
Export Cython Install Path

varunagrawal:
Export cython install path so it can be picked up by other cmake projects.

Simple 1 line change.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/364)


ProfFan:
Quoting #363, Travis is failing because numpy now requires Python 3.6, a suprising move...

PR #365, opened 2020-06-22
Reworked IMU examples

varunagrawal:
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/365)

- Formatted and updated all the IMU examples to make them cleaner.
- Renamed `ImuFactorExample2.py` to `ImuFactorISAM2Example.py`.
- Added separate example for `CombinedImuFactor`.
- Added better command line option parsing to both Python (`argparse`) and C++ (`boost::program_options`) examples.


varunagrawal:
I've already updated the pybind version of this.

examples/CombinedImuFactorsExample.cpp
```diff
@@ -0,0 +1,287 @@
...
+  double output_time = 0.0;
+  double dt = 0.005;  // The real system has noise, but here, results are nearly
+                      // exactly the same, so keeping this for simplicity.
+
+  // All priors have been set up, now iterate through the data file.
+  while (file.good()) {
+    // Parse out first value
+    string value;
+    getline(file, value, ',');
+    int type = atoi(value.c_str());
```
jlblancoc, line 207: Probably being picky but... it catches my eye the use of "old C" atoi() atof()... there are the alternatives std::stoi(), std::stof(), etc.

varunagrawal: Yes, you're right. I wasn't paying too much attention to detail in the code itself since it works well, but this is the kind of technical debt this PR is aiming to address. :slightly_smiling_face:

ProfFan:
@varunagrawal Ping

PR #366, opened 2020-06-23
Set minimum supported numpy version

varunagrawal:
Set the minimum numpy version to `1.11.0` aka the version provided via the Xenial debian package to provide a baseline for minimum support.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/366)

### DIFF
--- a/.travis.python.sh
+++ b/.travis.python.sh
@@ -17,8 +17,7 @@ fi
 
 CURRDIR=$(pwd)
 
-# Explicitly install packages so we are consistent with the system installed python version
-sudo $PYTHON -m pip install Cython>=0.25.2 backports_abc>=0.5
+sudo $PYTHON -m pip install -r ./cython/requirements.txt
 
 mkdir $CURRDIR/build
 cd $CURRDIR/build
--- a/.travis.python.sh
+++ b/.travis.python.sh
@@ -17,7 +17,8 @@ fi
 
 CURRDIR=$(pwd)
 
-sudo $PYTHON -m pip install -r ./cython/requirements.txt
+# Explicitly install packages so we are consistent with the system installed python version
+sudo $PYTHON -m pip install Cython>=0.25.2 backports_abc>=0.5
 
 mkdir $CURRDIR/build
 cd $CURRDIR/build
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,12 @@ addons:
     - ubuntu-toolchain-r-test
     - sourceline: 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-9 main'
       key_url: 'https://apt.llvm.org/llvm-snapshot.gpg.key'
-    - deadsnakes
     packages:
     - g++-9
     - clang-9
     - build-essential pkg-config
     - cmake
-    - python3.6 python3.6-dev
+    - python3-dev libpython-dev
     - libboost-all-dev
 
 # before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ addons:
     - build-essential pkg-config
     - cmake
     - python3-dev libpython-dev
+    - python3-numpy
     - libboost-all-dev
 
 # before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,20 @@
 language: cpp
 cache: ccache
 sudo: required
-dist: xenial
+dist: bionic
 
 addons:
   apt:
     sources:
     - ubuntu-toolchain-r-test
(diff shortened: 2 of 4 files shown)


ProfFan:
I am not sure if we should do this, or update our base to 18.04, or install numpy from the 16.04 official source `python3-numpy`?

varunagrawal:
I am not a fan of updating the base to `18.04` since a lot of users are presumably still on `16.04`.  Since 16.04 is technically supported till April 2021, I don't foresee people updating any time soon.

As for using `python3-numpy`, again not a fan since that is not platform independent. Installing via pip and the `requirements.txt` is the most general form.

varunagrawal:
I am sure you know I would love to use 18.04 instead of 16.04. Xenial has caused me enough headaches to last a while.

ProfFan:
Yeah I totally understand :) My argument here is that we may want to keep compatibility with 16.04, as some of our clients may be still using some embedded robot system that is stuck in 16.04...

varunagrawal:
That's a great point. In that case I guess we should figure out a way to pin a specific version of numpy.

varunagrawal:
@dellaert @ProfFan after sleeping on it, I feel the best mode of action is a simple travis image update to Ubuntu 18.04. Given that it's less than a year till 16.04 is no longer supported, we should leave the onus of managing that dependency to downstream users.

dellaert:
Uhm, no, we can’t in general just push the problem to users. So, that’s not a good argument.

We have to decide what version of numpy and Ubuntu we support with GTSAM. I think indeed 16.04 should continue to work. It seems it works just fine, so I see no reason to bump. And, before you say numpy,  I don’t actually understand why numpy matters for Travis as I thought we did python build/checking with GitHub actions now?

varunagrawal:
It's not really a problem per se. The `requirements.txt` in the `cython` directory just pulls the latest numpy and numpy 1.19.0 drops support for Python 3.5 and below. Users of Xenial will just have to use numpy 1.18.4, which I imagine their requirements.txt should already account for despite [this comic](https://xkcd.com/1987/).

Regarding Github Actions, I don't know what the status on that is, and what is the migration plan thus far. I don't want to stretch @ProfFan too thin since he's already overloaded with the wrapper and SwiftFusion, and Github Actions is low hanging fruit.

This issue though is blocking all our other PRs so we should resolve it ASAP.

It seems like we have 3 options:

1. Update Travis to Bionic.
2. Use the `deadsnakes` PPA and use Python 3.6 by default in the test environment.
3. Restrict testing to a numpy version < 1.19.0.

Thoughts?

dellaert:
We should be testing what our minimum requirement is. I suggest it is Ubuntu 16.04, and the lowest version of python and numpy that work with our current code, as I am assuming later versions of python and numpy are backwards compatible. We should find out out, use that for testing, and document in GTSAM documentation :-)

dellaert:
And, my question about github actions?

ProfFan:
I think after the discussion here the solution should be use the `python3-pip` package in the distribution. @dellaert @varunagrawal What do you think? We need to move forward fast as this is blocking CI for other PRs.

dellaert:
@ProfFan I already said what I think above. Explain how your proposal is compatible with that.

ProfFan:
> @ProfFan I already said what I think above. Explain how your proposal is compatible with that.

Yeah in that spirit, we do the following:
1. Stay on `xenial`;
2. add `sudo apt-get install python3-numpy` (numpy included in the official repo) to the build script before `pip3 install -r requirements.txt`
3. Done.

dellaert:
OK, this is not quite what I said, as the python and numpy version that *actually* work might be lower. So you are proposing to use whatever comes with 16.04 as the minimum version requirement?

varunagrawal:
@ProfFan's suggestion seems reasonable. It uses numpy v1.11 (which is ancient at this point). The only thing I am concerned about is that there have been a lot of bug fixes ever since, which made me want to suggest v1.14.5.

I can run a quick docker container and run the failing test config to verify.

dellaert:
bug fixes are irrelevant to us if they don't trigger any of our unit tests :-)

varunagrawal:
Yup, agreed. Hence the quick check on docker before I make the necessary changes. :)

varunagrawal:
@ProfFan @dellaert unittests pass for 1.11.0. I have made the necessary updates to this PR.

varunagrawal:
@dellaert @ProfFan can you please review?

dellaert:
Delegating @ProfFan to do *thorough* review :-)

ProfFan:
@varunagrawal

PR #370, opened 2020-06-25
Minor Updates

varunagrawal:
1. Fixed a CMake status message.
2. Wrapped the `biasHat` function.
3. ~~Added a `.gitignore` file for the `cmake` directory to help when building the sample projects.~~

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/370)


cmake/.gitignore
```diff
@@ -0,0 +1 @@
+**/build
```
dellaert, line 1: Should not be necessary

varunagrawal: Yeah I just thought it was a nice to have to help keep workspaces clean. Doesn't do much otherwise.

PR #371, opened 2020-06-30
Capture stdout in python test

varunagrawal:
Add Python3 based code to capture the output from the test so prevent printing to terminal and potentially corrupting test metrics.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/371)

### DIFF
--- a/cython/gtsam/tests/test_logging_optimizer.py
+++ b/cython/gtsam/tests/test_logging_optimizer.py
@@ -4,6 +4,8 @@ Author: Jing Wu and Frank Dellaert
 """
 # pylint: disable=invalid-name
 
+import io
+import sys
 import unittest
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


ProfFan:
Could you add a before/after? Thanks :)

varunagrawal:
Before:

```
..............................................s0.02013409284498978
0.010016753752400492
0.010016690308967725
..
Ran 49 tests in 0.066s

OK (skipped=1)
```

After:

```
..............................................s.
Ran 48 tests in 1.226s

OK (skipped=1)
```

dellaert:
In GTSAM 4 we are, in 4.1 we are not. So, when the pybind PR is split (@ProfFan ?) and both halves are reviewed and certified to work with the python example project, we can move on...

This is definitely not a 4.0 PR

varunagrawal:
Adding Python2 support is simple enough. I wasn't sure we were supporting it anymore since Python2 has reached EOL.

varunagrawal:
Huh, what's funny/weird is that the test I'm trying to make quiet is not Python2 compliant. Guess I'll have to update that.

https://github.com/borglab/gtsam/pull/371/files#diff-d92a2d735f7acc7b98cd44632e056028R54

varunagrawal:
Updated the code to be both python 2 and 3 compliant.

Issue #377, opened 2020-07-07
MATLAB Wrapper Build Is Broken


mikesheffler:
The MATLAB wrapper build is broken. I'm without a Linux box right this second, so I can only confirm that that is the case on Windows. It appears to be just stuff that was either not updated or (not sure about this one) showed up early from another branch.

1) Serialization in `SOn.h`. There' some serialization code needed. I plucked some from `origin/feature/new_wrapper` and that seemed to do the trick.

2) Something to do with Frobenius factor. Possibly a CMake issue? There are unresolved external symbols related to FrobeniusFactor, the first of which is `ConvertPose3NoiseModel`. If you look at `gtsam.h`, there is

```
#include <gtsam/slam/FrobeniusFactor.h>
gtsam::noiseModel::Isotropic* ConvertPose3NoiseModel(
    gtsam::noiseModel::Base* model, size_t d);
```

I went to `FrobeniusFactor.h` to make sure that the function existed, but, what I found was that the file (and the accompanying `.cpp`) was in my file system but not in my solution, which makes me think 'CMake problem'. The `#include` statement in `gtsam.h` is pulling the `.h` in directly but not the `.cpp`. Locally, I just commented the `FrobeniusFactor` stuff out of `gtsam.h` which is almost certainly the wrong way to do it, but I wasn't sure what the real story is.

I need to get set up to do a pull request on github (I haven't done one since bitbucket), but, if anyone wants a preview of the `SOn.h` thing, the added lines are just:

```
// For save/load
#include <boost/serialization/split_member.hpp>

// This block is in
// class SO : public LieGroup<SO<N>, internal::DimensionSO(N)> {
// public :
  template <class Archive>
  friend void save(Archive&, SO&, const unsigned int);
  template <class Archive>
  friend void load(Archive&, SO&, const unsigned int);

/** Serialization function */
template<class Archive>
void serialize(
    Archive &ar, SOn &Q,
    const unsigned int file_version
){
  Matrix& M = Q.matrix_;
  ar & M;
}

```

I'll update when I actually have  a PR ready to go.

PR #378, opened 2020-07-07
Fix/matlab wrapper

mikesheffler:
Quick changes to get the MATLAB wrapper building.

@jlblancoc Per #377 , I think something is up with CMake and the FrobeniusFactor. You seem to be the CMake wizard, do you know why the Frobenius factor files might not turn up in a solution? I would have assumed that all of the files in `slam` were just added together automatically.

@ProfFan Also per #377 , the changes to `SOn.h` were just grabbed from `feature/new_wrapper`. Could you confirm that porting them over like this is appropriate?

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/378)

### DIFF
--- a/gtsam.h
+++ b/gtsam.h
@@ -2836,33 +2836,33 @@ virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
   KarcherMeanFactor(const gtsam::KeyVector& keys);
 };
 
-#include <gtsam/slam/FrobeniusFactor.h>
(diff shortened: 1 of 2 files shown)


mikesheffler:
I built on Windows with the MATLAB toolbox enabled. My kluge for `FrobeniusFactor` could have Python side effects that I couldn't see.

mikesheffler:
I think I see what the problem is here. I'm going to withdraw this one and submit a new one.

Issue #20, opened 2019-05-23
Smartfactor body_P_sensor


dellaert:
Not respected: fix it.

mikesheffler:
What's the deal with this? I see `body_P_sensor_` mentioned in the `unwhitenedError` method in `SmartFactorBase.h`, but I think that's it. Where is it missed?

dellaert:
Needs a unit test whether if we pass `body_P_sensor`, it is actually used in projection factors and the like. It seems it is not, so I expect unit test to fail, and then we have to fix that :-)

mikesheffler:
I swear I'm not trying to be dense, but is this not covered by

`TEST( SmartProjectionPoseFactor, smartFactorWithSensorBodyTransform )`

in `testSmartProjectionPoseFactor.cpp`?

I'm suddenly interested because I'm using `SmartProjectionPoseFactor<Cal3_S2>` with a `sensor_to_body` transform, and I'm having a hell of a time.

I'm sure I'm just putting some garbage data in somewhere, but I figure it's worth asking ... :grin:

mikesheffler:
Ah. Did you have [this](https://bitbucket.org/gtborg/gtsam/issues/411/smartposeprojfactor-error-with-sensor) (issue number 41 from the bitbucket site) in mind? What kind of test did you need?

dellaert:
Just a test that checks that the projections are being done correctly if the camera is offset. Including derivatives

ProfFan:
Closed by #379, please reopen if not :)

PR #379, opened 2020-07-07
Test body_P_sensor in SmartFactor

varunagrawal:
Addresses #20.

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/379)

### DIFF
--- a/gtsam/slam/SmartFactorBase.h
+++ b/gtsam/slam/SmartFactorBase.h
@@ -206,12 +206,12 @@ protected:
       boost::optional<Matrix&> E = boost::none) const {
     Vector ue = cameras.reprojectionError(point, measured_, Fs, E);
     if (body_P_sensor_ && Fs) {
-      const Pose3 sensor_T_body = body_P_sensor_->inverse();
+      const Pose3 sensor_P_body = body_P_sensor_->inverse();
       for (size_t i = 0; i < Fs->size(); i++) {
-        const Pose3 world_T_body = cameras[i].pose() * sensor_T_body;
+        const Pose3 world_P_body = cameras[i].pose() * sensor_P_body;
         Matrix J(6, 6);
         // Call compose to compute Jacobian
-        world_T_body.compose(*body_P_sensor_, J);
+        world_P_body.compose(*body_P_sensor_, J);
         Fs->at(i) = Fs->at(i) * J;
       }
     }
--- a/gtsam/slam/SmartFactorBase.h
+++ b/gtsam/slam/SmartFactorBase.h
@@ -206,11 +206,12 @@ protected:
       boost::optional<Matrix&> E = boost::none) const {
     Vector ue = cameras.reprojectionError(point, measured_, Fs, E);
     if (body_P_sensor_ && Fs) {
-      const Pose3 sensor_P_body = body_P_sensor_->inverse();
+      const Pose3 sensor_T_body = body_P_sensor_->inverse();
       for (size_t i = 0; i < Fs->size(); i++) {
-        const Pose3 w_Pose_body = cameras[i].pose() * sensor_P_body;
+        const Pose3 world_T_body = cameras[i].pose() * sensor_T_body;
         Matrix J(6, 6);
-        const Pose3 world_P_body = w_Pose_body.compose(*body_P_sensor_, J);
+        // Call compose to compute Jacobian
+        world_T_body.compose(*body_P_sensor_, J);
         Fs->at(i) = Fs->at(i) * J;
       }
     }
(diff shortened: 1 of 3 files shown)


gtsam/slam/SmartFactorBase.h
```diff
@@ -206,11 +206,12 @@ class SmartFactorBase: public NonlinearFactor {
       boost::optional<Matrix&> E = boost::none) const {
     Vector ue = cameras.reprojectionError(point, measured_, Fs, E);
     if (body_P_sensor_ && Fs) {
-      const Pose3 sensor_P_body = body_P_sensor_->inverse();
+      const Pose3 sensor_T_body = body_P_sensor_->inverse();
```
dellaert, line 209: Please keep P, as the field uses that convention

varunagrawal: Gotcha. I was trying to follow the convention as per Samarth's post. Reverted.

varunagrawal:
@dellaert pretty much. The original issue asked for a test so decided it was easy enough to do so (aka unit tests are my friends).

varunagrawal:
So unit testing won this round. Discovered a bug in the handling of the jacobian for cases such as `PinholeCamera<Cal3Bundler>` where the number of parameters is 9 instead of the usual 6.

gtsam/slam/SmartFactorBase.h
```diff
@@ -207,10 +207,17 @@ class SmartFactorBase: public NonlinearFactor {
     Vector ue = cameras.reprojectionError(point, measured_, Fs, E);
     if (body_P_sensor_ && Fs) {
       const Pose3 sensor_P_body = body_P_sensor_->inverse();
+      size_t camera_dim = size_t(traits<CAMERA>::dimension);
```
dellaert, line 210: Use constexpr int, no casting

gtsam/slam/SmartFactorBase.h
```diff
@@ -207,10 +207,17 @@ class SmartFactorBase: public NonlinearFactor {
...
+      size_t pose_dim = size_t(traits<Pose3>::dimension);
+
       for (size_t i = 0; i < Fs->size(); i++) {
-        const Pose3 w_Pose_body = cameras[i].pose() * sensor_P_body;
-        Matrix J(6, 6);
-        const Pose3 world_P_body = w_Pose_body.compose(*body_P_sensor_, J);
+        const Pose3 world_P_body = cameras[i].pose() * sensor_P_body;
+        Matrix J = Matrix::Zero(camera_dim, camera_dim);
```
dellaert, line 215: Why switch to dynamic? Make static, use setZero(), save a malloc

gtsam/slam/SmartFactorBase.h
```diff
@@ -207,10 +207,17 @@ class SmartFactorBase: public NonlinearFactor {
...
+        // Call compose to compute Jacobian for camera extrinsics
+        Matrix H(pose_dim, pose_dim);
+        world_P_body.compose(*body_P_sensor_, H);
+        // Assign extrinsics
+        J.block(0, 0, pose_dim, pose_dim) = H;
```
dellaert, line 220: You can just pass the block into the optionalJacobian slot. Amazing, right?

gtsam/slam/SmartFactorBase.h
varunagrawal: Gotcha. It was already dynamic for some reason so was trying not to inadvertently introduce any bugs.

gtsam/slam/SmartFactorBase.h
varunagrawal: I tried doing that but it was complaining about something (which I don't remember right now). I am guessing it will fix itself with static matrices.

varunagrawal: So as per [this link](https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html), casting an `Eigen::Block` expression to a non-const `Matrix` is not possible.

Quoting the specific line:

> When trying to execute the following code
>```
>MatrixXf C = MatrixXf::Zero(3,6);
>cov(x,y, C.block(0,0,3,3));
>```
> 
> the compiler will fail, because it is not possible to convert the expression returned by MatrixXf::block() into a non-const MatrixXf&. This is the case because the compiler wants to protect you from writing your result to a temporary object.

dellaert: Is the same true for fixed matrices, though?

varunagrawal: Looks like it. I am having compile time issues otherwise, specifically signature mismatch. This is the best explanation since we're trying to convert a block expression to a non-const matrix as a requirement.

eigen_error

PR #381, opened 2020-07-08
Second attempt at a MATLAB wrapper build fix.

mikesheffler:
1) Some serialization code was missing from `SOn.h` (not sure why this wouldn't have been a problem before building the MATLAB toolbox ...). I grabbed the added code from the `feature/new_wrapper` branch.

2) FrobeniusFactor stuff needed a couple `GTSAM_EXPORT` statements

Passes Windows building (don't have access to my Linux box at the moment)

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/381)

### DIFF
--- a/gtsam/geometry/SOn.h
+++ b/gtsam/geometry/SOn.h
@@ -30,6 +30,9 @@
 #include <vector>
 #include <random>
 
+ // For save/load
+#include <boost/serialization/split_member.hpp>
+
 namespace gtsam {
 
 namespace internal {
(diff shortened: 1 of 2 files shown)


gtsam/geometry/SOn.h
```diff
@@ -30,6 +30,9 @@
 #include <vector>
 #include <random>
 
+ // For save/load
+#include <boost/serialization/split_member.hpp>
```
dellaert, line 34: Why? Does vanilla serialization not work?

ProfFan: My bad, I did this to fix a boost bug revealed by the new test matrix, but forgot which bug it is...

dellaert: Not understanding this comment. The change was made by @mikesheffler, right? Also, if it is *not* needed, we should not do it - as it adds code and mystery.

mikesheffler: Both are true! It's Fan's code; I just replicated it exactly as it was on the `feature/new_wrapper` branch.

dellaert:
@mikesheffler thanks for this !!!! I'll let you merge and delete the branch, jsut in case you still had an outstanding change.

Issue #383, opened 2020-07-08
Is #include <boost/serialization/split_member.hpp> needed in SOn.h?


mikesheffler:
The fix to #377 (PR #381) involved the addition of

```
#include <boost/serialization/split_member.hpp>
```

to `SOn.h`. Quoth @dellaert :
> If this fixes the wrapper, I'm on board with merging ASAP - esp. before we release 4.0.3. But, please create an issue to examine whether the split_member is needed for anything, so we remember to remove it if not.

I'll go ahead and remove the include locally to see if I can still build.

mikesheffler:
Checked. It does not seem to be necessary

mikesheffler:
This should have closed with #384. Closing now.

PR #384, opened 2020-07-08
Unnecessary include statement

mikesheffler:
Answers #383 (the include is not, in fact necessary; @ProfFan , you might want to check `feature/new_wrapper` to see if it is needed there either)

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/384)


ProfFan:
Sure, I'll check later :)

mikesheffler:
Builds passed. I think this can just be merged in.

PR #385, opened 2020-07-08
Fix lambda check in logging optimizer

ProfFan:
We didn't check the maximum lambda in the logging optimizer :)

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/385)


dellaert:
Failing build?

cython/gtsam/utils/logging_optimizer.py
```diff
@@ -46,5 +46,6 @@ def gtsam_optimize(optimizer,
     def check_convergence(optimizer, current_error, new_error):
         return (optimizer.iterations() >= params.getMaxIterations()) or (
             gtsam.checkConvergence(params.getRelativeErrorTol(), params.getAbsoluteErrorTol(), params.getErrorTol(),
-                                   current_error, new_error))
+                                   current_error, new_error)) or (optimizer.lambda_() > params.getlambdaUpperBound())
```
varunagrawal, line 49: Should be `optimizer.lambda()` not `lambda_()` (no underscore).

ProfFan: should have underscore as lambda is a reserved word

ProfFan:
I see what is happening, we need to first check if it is a LM optimizer, otherwise it will not have lambda.

cython/gtsam/utils/logging_optimizer.py
```diff
@@ -46,5 +46,7 @@ def gtsam_optimize(optimizer,
     def check_convergence(optimizer, current_error, new_error):
         return (optimizer.iterations() >= params.getMaxIterations()) or (
             gtsam.checkConvergence(params.getRelativeErrorTol(), params.getAbsoluteErrorTol(), params.getErrorTol(),
-                                   current_error, new_error))
+                                   current_error, new_error)) or (
+            type(optimizer).__name__ == "LevenbergMarquardtOptimizer" and optimizer.lambda_() > params.getlambdaUpperBound())
```
dellaert, line 50: Oops, did you actually run the test? it should be lambda as that is wrapped. No problem with reserved word as it's a method I think.

ProfFan: I actually run the test, and indeed lambda is not usable, and is converted as `lambda_` in the wrapper.

dellaert: The *cython* wrapper?

ProfFan: Yes

ProfFan: https://colab.research.google.com/drive/1AJ-6mlrDhhV3Uqbn7ahHDEy6dUWsmsDe?usp=sharing

dellaert: But how did it ever work for me then?

ProfFan: The logic is in `wrap/FullyOverloadedFunction.cpp`:
```
/// Cython: Rename functions which names are python keywords
  static const std::array<std::string, 2> pythonKeywords;
  static std::string pyRename(const std::string& name) {
    if (std::find(pythonKeywords.begin(), pythonKeywords.end(), name) ==
        pythonKeywords.end())
      return name;
    else
      return name + "_";
  }
```

varunagrawal, line 50: You should use `isinstance` rather than typing up hard coded strings.

cython/gtsam/tests/test_logging_optimizer.py
```diff
@@ -43,6 +43,11 @@ def setUp(self):
         self.optimizer = gtsam.GaussNewtonOptimizer(
             graph, initial, self.params)
 
+        self.lmparams = gtsam.LevenbergMarquardtParams()
+        self.lmoptimizer = gtsam.LevenbergMarquardtOptimizer(
```
varunagrawal, line 47: I believe this can be done better.

I recommend a factory function that constructs the optimizer based on a flag or callable. 

Feel free to ignore and add a TODO if you're strapped for time.

cython/gtsam/utils/logging_optimizer.py
ProfFan: Sure :)

ProfFan: done

cython/gtsam/tests/test_logging_optimizer.py
ProfFan: I believe we can deal with this later :)

varunagrawal: Be sure to add a TODO :)

varunagrawal:
@dellaert we're blocked on your review of this PR. :slightly_smiling_face: 
I have already approved it if that helps.

PR #386, opened 2020-07-09
CombinedImuFactor Serialization Fix

### LINKED ISSUE #158: Serialization bug
## Description
There seems to be a bug in CombinedImuFactor.h in the PreintegrationCombinedParams struct. When the serialize function is called,  only the data in PreintegratedRotation::Params will be serialized, but not the data in PreintegrationParams.

## Steps to reproduce

I didn't experience any segfault. I just came across while resolving some other issue.

## Expected behavior
In the current implementation, it is likely that the data in the superclass PreintegrationParams won't be serialized and only the one in PreintegratedRotationParams.

The fix apprears to be trivial:
ar& BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegratedRotation::Params);
should probably be changed to:
ar& BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegrationParams);
in the serialize function.

It would seem logical since PreintegrationParams that's the actual superclass. I tried the change already and It compiles just fine. However, I don't have a way to validate that change.

One would need to write a proper unit test.

varunagrawal:
This PR does the following

1. Makes `CombinedImuFactor` adhere to the `Testable` trait so that traits based comparison can be performed.
2. Make `CombinedImuFactor` serializable.
3. Update the serialization code for all the IMU based factors to follow a common convention.
4. Replace `PreintegratedRotation::Params` with `PreintegratedRotationParams` (unnecessary scope).
5. Added unit tests to test serialization of all IMU based factors.

This PR should also fix #158.


gtsam/navigation/CombinedImuFactor.h
```diff
@@ -61,10 +63,19 @@ struct GTSAM_EXPORT PreintegrationCombinedParams : PreintegrationParams {
   Matrix3 biasOmegaCovariance;  ///< continuous-time "Covariance" describing gyroscope bias random walk
   Matrix6 biasAccOmegaInt;     ///< covariance of bias used for pre-integration
 
+  /// Default constructor makes unitialized params struct.
```
ProfFan, line 66: typo uninitialized

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -61,10 +63,19 @@ struct GTSAM_EXPORT PreintegrationCombinedParams : PreintegrationParams {
...
+  /// Used for serialization.
+  PreintegrationCombinedParams()
+      : PreintegrationParams(),
```
ProfFan, line 69: Not needed with empty args

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -86,24 +100,25 @@ PreintegrationCombinedParams(const Vector3& n_gravity) :
...
-    void serialize(ARCHIVE& ar, const unsigned int /*version*/) {
-    ar& BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegratedRotation::Params);
-    ar& BOOST_SERIALIZATION_NVP(biasAccCovariance);
-    ar& BOOST_SERIALIZATION_NVP(biasOmegaCovariance);
-    ar& BOOST_SERIALIZATION_NVP(biasAccOmegaInt);
+  void serialize(ARCHIVE& ar, const unsigned int /*version*/) {
+    namespace bs = ::boost::serialization;
+    ar & BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegrationParams);
+    ar & bs::make_nvp("biasAccCovariance",
+         bs::make_array(biasAccCovariance.data(), biasAccCovariance.size()));
```
ProfFan, line 111: Is this because Eigen is not correctly serialized?

varunagrawal: Yes. From my understanding and what I have seen, we need to extract the underlying data array for Boost.

gtsam/navigation/CombinedImuFactor.h
varunagrawal: Gotcha. Just felt better to be consistent in our coding standards.

gtsam/navigation/CombinedImuFactor.h
dellaert: Yeah, but why is the Matrix serialization in Matrix.h not triggered? Perhaps because these are fixed-size matrices? If so, maybe the solution for this *and all other future issues* is to change the signature of those functions to

```
    template<class Archive, typename Derived>
    void save(Archive & ar, const Eigen::MatrixBase<Derived>& m, unsigned int /*version*/) {
```

Does that work?

varunagrawal: Oh I didn't see that the serialization is defined in `Matrix.h`. From the looks of it, yes, the serialization definition is for dynamic matrices, and these (especially `preintMeasCov_`) are static, so Eigen is probably getting confused.

That should be a separate PR IMO, but I will test it out here first.

varunagrawal: So I had to go with `Eigen::PlainObjectBase` since `MatrixBase` doesn't support the `data()` method (strange?). Waiting on a local test(s) to pass before I create the PR.

varunagrawal: Actually that didn't work. I ended up doing something different which worked very nicely. Creating the PR now.

varunagrawal:
We need #391 merged in before I can update this PR.

gtsam/navigation/CombinedImuFactor.h
dellaert: Do fix

gtsam/navigation/CombinedImuFactor.h
dellaert: Do fix

PR #387, opened 2020-07-09
Minimum Feedback Arc Set (MFAS) for 1dsfm

akshay-krishnan:
This PR implements a solution to the [Minimum Feedback Arc Set](https://en.wikipedia.org/wiki/Feedback_arc_set#Minimum_feedback_arc_set) (MFAS) problem. We use the solution from Kyle Wilson and Noah Snavely's [ECCV14 paper](https://research.cs.cornell.edu/1dsfm/docs/1DSfM_ECCV14.pdf). 

---
This change is [Reviewable](https://reviewable.io/reviews/borglab/gtsam/387)


gtsam/sfm/mfas.h
```diff
@@ -0,0 +1,65 @@
+/*
+  This file defines functions used to solve a Minimum feedback arc set (MFAS)
```
varunagrawal, line 2: Please update the file brief to the GTRC copyright. Do keep the comment about how this is taken from another source.

gtsam/sfm/mfas.h
```diff
@@ -0,0 +1,65 @@
...
+
+/*
+ * Given a vector of KeyPairs that constitutes edges in a graph and the weights
+ * corresponding to these edges, this function changes all the weights to
+ * positive numbers by flipping the direction of the edges that have a negative
+ * weight. The changes are made in place.
+ * @param edges reference to vector of KeyPairs
+ * @param weights weights corresponding to edges
+ */
+void flipNegEdges(std::vector<KeyPair> &edges, std::vector<double> &weights);
```
varunagrawal, line 34: `flipNegativeEdges`?

gtsam/sfm/mfas.h
```diff
@@ -0,0 +1,65 @@
...
+ * Computes the MFAS ordering, ie an ordering of the nodes in the graph such
+ * that the source of any edge appears before its destination in the ordering.
+ * The weight of edges that are removed to obtain this ordering is minimized.
+ * @param edges: edges in the graph
+ * @param weights: weights corresponding to the edges (have to be positive)
+ * @param nodes: nodes in the graph
+ * @param ordered_positions: map from node to position in the ordering (0
+ * indexed)
+ */
+void mfasRatio(const std::vector<KeyPair> &edges,
```
varunagrawal, line 46: Is mfas in the function name necessary if we've defined a mfas namespace?

gtsam/sfm/mfas.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+void mfasRatio(const std::vector<KeyPair> &edges,
+               const std::vector<double> &weights, const KeyVector &nodes,
+               FastMap<Key, int> &ordered_positions) {
+  // initialize data structures
+  FastMap<Key, double> win_deg;
+  FastMap<Key, double> wout_deg;
+  FastMap<Key, vector<pair<int, double> > > inbrs;
+  FastMap<Key, vector<pair<int, double> > > onbrs;
```
varunagrawal, line 42: Use full names to make code more readable?

akshay-krishnan:
Thank you for reviewing @varunagrawal. I am currently working on changing this implementation, but I will make the changes you suggested too.

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * @brief Construct from the nodes in a graph (points in 3D), edges
+   * that are translation directions in 3D and the direction in
+   * which edges are to be projected.
+   * @param nodes Nodes in the graph
+   * @param relativeTranslations translation directions between nodes
+   * @param projectionDirection direction in which edges are to be projected
+   */
+  MFAS(const std::shared_ptr<std::vector<Key>> &nodes,
+       const std::shared_ptr<TranslationEdges> &relativeTranslations,
```
dellaert, line 50: Same: why a shared pointer? Just do const &

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+   */
+  std::map<KeyPair, double> computeOutlierWeights();
+
+  /**
+   * @brief Computes the 1D MFAS ordering of nodes in the graph
+   * @return orderedNodes: vector of nodes in the obtained order
+   */
+  std::vector<Key> computeOrdering();
+
+ private:
```
dellaert, line 80: In all other gtsam classes private instance variables are defined at the top of the class. Private methods can go here

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+
+  Given a weighted directed graph, the objective in a Minimum feedback arc set
+  problem is to obtain a graph that does not contain any cycles by removing
+  edges such that the total weight of removed edges is minimum.
+  @addtogroup SFM
+*/
+class MFAS {
+ public:
+  /**
+   * @brief Construct from the nodes in a graph (points in 3D), edges
```
dellaert, line 42: you say Point in 3D but you ask for Keys below

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * @brief Computes the "outlier weights" of the graph. We define the outlier weight
+   * of a edge to be zero if the edge in an inlier and the magnitude of its edgeWeight
+   * if it is an outlier. This function can only be used when constructing the 
+   * @return outlierWeights: map from an edge to its outlier weight.
+   */
+  std::map<KeyPair, double> computeOutlierWeights();
+
+  /**
+   * @brief Computes the 1D MFAS ordering of nodes in the graph
```
dellaert, line 75: We have a class Ordering in GTSAM. Is this an appropriate use of it?

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+   * @return orderedNodes: vector of nodes in the obtained order
+   */
+  std::vector<Key> computeOrdering();
+
+ private:
+  // pointer to nodes in the graph
+  const std::shared_ptr<std::vector<Key>> nodes_;
```
dellaert, line 82: Just store them, not shared pointers

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+  const std::shared_ptr<TranslationEdges> relativeTranslations_;
+
+  // relative translations when the object is initialized without using the actual
+  // relative translations, but with the weights from projecting in a certain 
+  // direction. This is used for unit testing, but not in practice. 
+  std::shared_ptr<TranslationEdges> relativeTranslationsForWeights_;
+
+  // edges with a direction such that all weights are positive
+  // i.e, edges that originally had negative weights are flipped
+  std::map<KeyPair, double> positiveEdgeWeights_;
```
dellaert, line 93: where are these computed? Note we don't use in GTSAM if we can help it. All methods should be const, so I expect these to be computed once in the constructor, and never modified :-)

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+  MFAS(const std::shared_ptr<std::vector<Key>> &nodes,
+       const std::map<KeyPair, double> &edgeWeights);
+
+  /**
+   * @brief Computes the "outlier weights" of the graph. We define the outlier weight
+   * of a edge to be zero if the edge in an inlier and the magnitude of its edgeWeight
+   * if it is an outlier. This function can only be used when constructing the 
+   * @return outlierWeights: map from an edge to its outlier weight.
+   */
+  std::map<KeyPair, double> computeOutlierWeights();
```
dellaert, line 72: should be const

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * @brief Computes the 1D MFAS ordering of nodes in the graph
+   * @return orderedNodes: vector of nodes in the obtained order
+   */
+  std::vector<Key> computeOrdering();
```
dellaert, line 78: should be const

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+
+#include <map>
+#include <memory>
+#include <vector>
+
+namespace gtsam {
+
+// used to represent edges between two nodes in the graph
+using KeyPair = std::pair<Key, Key>;
+using TranslationEdges = std::map<KeyPair, Unit3>;
```
ayushbaid, line 26: maybe name it UnitTranslationEdges?

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+  edges such that the total weight of removed edges is minimum.
+  @addtogroup SFM
+*/
+class MFAS {
+ public:
+  /**
+   * @brief Construct from the nodes in a graph (points in 3D), edges
+   * that are translation directions in 3D and the direction in
+   * which edges are to be projected.
+   * @param nodes Nodes in the graph
```
ayushbaid, line 45: Instead of nodes in a graph, we can add something more descriptive. Maybe something like pair of points in 3D?

Or are we making MFAS a generic class which works on any graph and not specific to SLAM?

gtsam/sfm/MFAS.h
```diff
@@ -0,0 +1,108 @@
...
+   * @brief Construct from the nodes in a graph and weighted directed edges
+   * between the graph. Not recommended for any purpose other than unit testing. 
+   * The computeOutlierWeights method will return an empty output if this constructor 
+   * is used. 
+   * When used in a translation averaging context, these weights are obtained
+   * by projecting edges in a particular direction.
+   * @param nodes: Nodes in the graph
+   * @param edgeWeights: weights of edges in the graph (map from edge to signed double)
+   */
+  MFAS(const std::shared_ptr<std::vector<Key>> &nodes,
```
ayushbaid, line 63: Is there anything else we can do? Mock the constructor? Or inherit this class, add this constructor, and test the child class?

gtsam/sfm/MFAS.cpp
```diff
@@ -0,0 +1,138 @@
...
+    in_weights[edge.second] += it->second;
+    out_weights[edge.first] += it->second;
+    in_neighbors[edge.second].push_back(edge.first);
+    out_neighbors[edge.first].push_back(edge.second);
+  }
+
+  // in each iteration, one node is appended to the ordered list
+  while (orderedNodes_.size() < nodes_->size()) {
+
+    // finding the node with the max heuristic score
```
ayushbaid, line 70: separate this in a function?

gtsam/sfm/MFAS.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+std::map<KeyPair, double> MFAS::computeOutlierWeights() {
+  // if ordering has not been computed yet
+  if (orderedNodes_.size() != nodes_->size()) {
+    computeOrdering();
+  }
+  // iterate over all edges
+  // start and end iterators depend on whether we are using relativeTranslations_ or
+  // relativeTranslationsForWeights_ to store the original edge directions
+  TranslationEdges::iterator start, end;
```
ayushbaid, line 112: rename them as startIterator, endIterator?

akshay-krishnan:
Thank you for the comments @dellaert and @ayushbaid. 

Prof. Frank, this class would be used on graphs with a large number of nodes (which represent cameras in our case). 1dsfm requires several instances of this class to compute the ordering for several sampled directions. These instances would run in parallel. I used the pointers for relativeTranslations_ and nodes_ to avoid storing/copying these large members for each instance. Yes, constant references would also serve the same purpose and I am happy to make that change, but for future reference please let me know why we prefer them.

akshay-krishnan:
I can also make most intermediate results internal to the methods, but its a little tricky for `positiveEdgeWeights_`. For some context, this is computed (but not returned) in `computeOrdering`, which is called by `computeOutlierWeights`. But then, `computeOutlierWeights` also needs `positiveEdgeWeights_`. Ultimately, 1dsfm only needs `computeOutlierWeights`, but I thought `computeOrdering` might be a nice method to have in the future. So the only way I could compute `positiveEdgeWeights_` in `computeOrdering` and use it in `computeOutlierWeights` without making in an instance variable is through an optional pointer argument to `computeOrdering`. Is that okay? Otherwise, I could merge both of them into one large method, which I think is not very clean.

dellaert:
> Thank you for the comments @dellaert and @ayushbaid.
> 
> Prof. Frank, this class would be used on graphs with a large number of nodes (which represent cameras in our case). 1dsfm requires several instances of this class to compute the ordering for several sampled directions. These instances would run in parallel. I used the pointers for relativeTranslations_ and nodes_ to avoid storing/copying these large members for each instance. Yes, constant reference members would also serve the same purpose and I am happy to make that change, but for future reference please let me know why we prefer them.

No, in that case shared pointer is fine, but then document it :-) although, I’m wondering about the design: why not have one class and run a method in parallel?

dellaert:
> I can also make most intermediate results internal to the methods, but its a little tricky for `positiveEdgeWeights_`. For some context, this is computed (but not returned) in `computeOrdering`, which is called by `computeOutlierWeights`. But then, `computeOutlierWeights` also needs `positiveEdgeWeights_`. Ultimately, 1dsfm only needs `computeOutlierWeights`, but I thought `computeOrdering` might be a nice method to have in the future. So the only way I could compute `positiveEdgeWeights_` in `computeOrdering` and use it in `computeOutlierWeights` without making in an instance variable is through an optional pointer argument to `computeOrdering`. Is that okay? Otherwise, I could merge both of them into one large method, which I think is not very clean.

Anything that is computed in the constructor is fine as an instance variable. The functional pattern we use is: only constructor is non-const.

akshay-krishnan:
> No, in that case shared pointer is fine, but then document it :-) although, I’m wondering about the design: why not have one class and run a method in parallel?

Yes, that can be done. But even then, do we need to copy these large data structures? 

In such a design, `computeOrdering` and `computeOutlierWeights` will have a const argument for the projection direction. But since `positiveEdgeWeights_` has to be shared from `computeOrdering`, it will either be non-const or have an optional pointer argument. I think the optional pointer argument is better, is that correct?

gtsam/sfm/MFAS.h
akshay-krishnan: While I do appreciate the name, I used the same that was used in TranslationRecovery.h for consistency. We might not use TranslationEdges now that we are changing the interface to vector of edges, so I'm not changing the name right now.

gtsam/sfm/MFAS.h
akshay-krishnan: I am updating the documentation to justify the use of pointers for nodes, as we discussed.

gtsam/sfm/MFAS.h
akshay-krishnan: Yeah it can be used, I am however considering changing the output of this function to a map from the original index to the ordered index. Please refer to comment in computeOutlierWeights() implementation as to why I think that is a good idea.

gtsam/sfm/MFAS.h
akshay-krishnan: I've updated the documentation to explain what these nodes represent when used in a translation averaging context.

gtsam/sfm/MFAS.cpp
akshay-krishnan: It is not used anywhere else, but yes, improves readability to factor that into a different function. I shall make that change in the next iteration if no other changes are suggested.

akshay-krishnan:
Prof. @dellaert  and @ayushbaid could you please re-review and let me know if more changes are necessary? I think it is better to make the change from the map to vector for representing edges in a separate PR.

gtsam/sfm/MFAS.cpp
```diff
@@ -0,0 +1,129 @@
...
+    : nodes_(nodes) {
+  // iterate over edges, obtain weights by projecting
+  // their relativeTranslations along the projection direction
+  for (auto it = relativeTranslations.begin();
+       it != relativeTranslations.end(); it++) {
+    edgeWeights_[it->first] = it->second.dot(projection_direction);
+  }
+}
+
+std::vector<Key> MFAS::computeOrdering() const {
```
dellaert, line 26: Maybe don’t go to four lines, but this function is indeed too long and complicated. Also ask yourself which STL algorithms you could be using.

dellaert:
@akshay-krishnan merge in develop before doing another push as we changed CI to GitHub actions...

akshay-krishnan:
@dellaert I have made some changes to improve the modularity and readability. Please re-review and let me know if there are more changes I can make.

dellaert (approved):
Awesome!

jingwuOUO (approved):

dellaert:
@varunagrawal please approve and merge or comment on what is missing

varunagrawal (approved):
LGTM

PR #391, opened 2020-07-10
Generic Eigen::Matrix serialization for boost

varunagrawal:
Updated the boost serialization code to handle both static and dynamic Eigen matrices.

Instead of doing this

```cpp
ar & bs::make_nvp("biasAccCovariance",
            bs::make_array(biasAccCovariance.data(), biasAccCovariance.size()));
```

we can now just do this

```cpp
ar& BOOST_SERIALIZATION_NVP(biasAccOmegaInt);
```

### DIFF
--- a/gtsam/base/Matrix.h
+++ b/gtsam/base/Matrix.h
@@ -549,16 +549,32 @@ namespace boost {
   namespace serialization {
 
     // split version - sends sizes ahead
-    template<class Archive>
-    void save(Archive & ar, const gtsam::Matrix & m, unsigned int /*version*/) {
+    template<class Archive,
+             typename Scalar_,
+             int Rows_,
+             int Cols_,
+             int Ops_,
+             int MaxRows_,
+             int MaxCols_>
+    void save(Archive & ar,
+              const Eigen::Matrix<Scalar_, Rows_, Cols_, Ops_, MaxRows_, MaxCols_> & m,
+              const unsigned int /*version*/) {
       const size_t rows = m.rows(), cols = m.cols();
       ar << BOOST_SERIALIZATION_NVP(rows);
       ar << BOOST_SERIALIZATION_NVP(cols);
       ar << make_nvp("data", make_array(m.data(), m.size()));
     }
 
-    template<class Archive>
-    void load(Archive & ar, gtsam::Matrix & m, unsigned int /*version*/) {
+    template<class Archive,
+             typename Scalar_,
+             int Rows_,
+             int Cols_,
+             int Ops_,
+             int MaxRows_,
+             int MaxCols_>
+    void load(Archive & ar,
+              Eigen::Matrix<Scalar_, Rows_, Cols_, Ops_, MaxRows_, MaxCols_> & m,
+              const unsigned int /*version*/) {
       size_t rows, cols;
       ar >> BOOST_SERIALIZATION_NVP(rows);
       ar >> BOOST_SERIALIZATION_NVP(cols);
@@ -566,8 +582,19 @@ namespace boost {
       ar >> make_nvp("data", make_array(m.data(), m.size()));
     }
 
+    // templated version of BOOST_SERIALIZATION_SPLIT_FREE(Eigen::Matrix);
+    template<class Archive,
+             typename Scalar_,
+             int Rows_,
+             int Cols_,
+             int Ops_,
+             int MaxRows_,
+             int MaxCols_>
+    void serialize(Archive & ar,
+              Eigen::Matrix<Scalar_, Rows_, Cols_, Ops_, MaxRows_, MaxCols_> & m,
+              const unsigned int version) {
+      split_free(ar, m, version);
+    }
+
   } // namespace serialization
 } // namespace boost
-
-BOOST_SERIALIZATION_SPLIT_FREE(gtsam::Matrix);
-
--- a/gtsam/base/Matrix.h
+++ b/gtsam/base/Matrix.h
@@ -548,6 +548,20 @@ GTSAM_EXPORT Vector columnNormSquare(const Matrix &A);
 namespace boost {
   namespace serialization {
 
+    /**
+     * Ref. https://stackoverflow.com/questions/18382457/eigen-and-boostserialize/22903063#22903063
+     * 
+     * Eigen supports calling resize() on both static and dynamic matrices.
+     * This allows for a uniform API, with resize having no effect if the static matrix
+     * is already the correct size.
+     * https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html#TutorialMatrixSizesResizing
+     * 
+     * We use all the Matrix template parameters to ensure wide compatibility.
+     * 
+     * eigen_typekit in ROS uses the same code
+     * http://docs.ros.org/lunar/api/eigen_typekit/html/eigen__mqueue_8cpp_source.html
+     */
+
     // split version - sends sizes ahead
     template<class Archive,
              typename Scalar_,
--- a/gtsam/navigation/ImuFactor.h
+++ b/gtsam/navigation/ImuFactor.h
@@ -164,7 +164,7 @@ private:
   void serialize(ARCHIVE & ar, const unsigned int /*version*/) {
     namespace bs = ::boost::serialization;
     ar & BOOST_SERIALIZATION_BASE_OBJECT_NVP(PreintegrationType);
-    ar & bs::make_nvp("preintMeasCov_", bs::make_array(preintMeasCov_.data(), preintMeasCov_.size()));
+    ar & BOOST_SERIALIZATION_NVP(preintMeasCov_);
   }
 };
 
--- a/gtsam/navigation/ManifoldPreintegration.h
+++ b/gtsam/navigation/ManifoldPreintegration.h
@@ -122,11 +122,11 @@ private:
     ar & BOOST_SERIALIZATION_NVP(deltaTij_);
     ar & BOOST_SERIALIZATION_NVP(deltaXij_);
     ar & BOOST_SERIALIZATION_NVP(biasHat_);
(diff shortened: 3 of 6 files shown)


varunagrawal:
The way I did it was to implement this change as part of #386 and update all the `make_array` calls, verify that tests pass and then pull out this functionality into its own PR.

Would creating this PR against the branch of #386 (instead of `develop`) give us more confidence in its functionality, while still being easy to review?

dellaert:
> Cool, but could you hunt down and replace the current make_array calls so the corresponding unit tests can attest to the correctness of this PR? If a template is never instantiated, I could be whatever...

I saw some comment in email that seems to be deleted. But, I did not agree with it anyway: make_array *is* used several places, mostly in navigation, so the above course of action is the right one for *this* PR.

varunagrawal:
Yes I agree with you. That was me shooting from the hip earlier. *sigh*

varunagrawal:
@dellaert updated all the `make_array` calls in the codebase. Tests pass locally, waiting on CI.

gtsam/base/Matrix.h
```diff
@@ -549,25 +549,52 @@ namespace boost {
...
-    template<class Archive>
-    void load(Archive & ar, gtsam::Matrix & m, unsigned int /*version*/) {
+    template<class Archive,
+             typename Scalar_,
+             int Rows_,
+             int Cols_,
+             int Ops_,
+             int MaxRows_,
+             int MaxCols_>
+    void load(Archive & ar,
```
dellaert, line 575: This cannot work for fixed arrays. They have no resize. So, you need to either specialize the template (preferred), or do some conditional logic below. I would also (unless you want to think about al possible combinations, remove all but the Scalar, Rows, Cols template parameters.

varunagrawal: I did some research and I found a couple of things:

1. Fixed sized matrices support `resize` for uniformity of the API though have assertions in place to ensure the resize is not incorrect. Quoting from the Eigen docs:

> All these methods are still available on fixed-size matrices, for the sake of API uniformity. Of course, you can't actually resize a fixed-size matrix. Trying to change a fixed size to an actually different value will trigger an assertion failure;

2. I spent a lot of time researching this particular problem yesterday, since I found this slightly odd too. This [StackOverflow answer](https://stackoverflow.com/a/22903063/1236990) was my main source of understanding, and I found confidence in this answer when I saw [eigen_typekit in ROS](http://docs.ros.org/lunar/api/eigen_typekit/html/eigen__mqueue_8cpp_source.html) use the same code. This [gist](https://gist.github.com/mtao/5798888) also provides the same code, though they seem independently developed since the gist doesn't use some in-built Eigen functionality. This [other gist](https://gist.github.com/artivis/288898ff877130e9a58ed04aeb9e59b5) does partial template specialization so putting it here as a refernce.

3. I agree with keeping only Scalar, Rows and Cols. It just seemed like a good idea to cover our bases by including all parameters, especially after seeing the examples in my above comment.

Thoughts?

dellaert:
I think that sounds like a plan. As long as this PR has the evidence (and maybe you can point out where it is in the PR comment) that static and dynamic matrices are correctly handled.
GitHub <notifications@github.com> wrote:
“@varunagrawal commented on this pull request.”


“In gtsam/base/Matrix.h:”

>        const size_t rows = m.rows(), cols = m.cols();
       ar << BOOST_SERIALIZATION_NVP(rows);
       ar << BOOST_SERIALIZATION_NVP(cols);
       ar << make_nvp("data", make_array(m.data(), m.size()));
     }
 
-    template<class Archive>
-    void load(Archive & ar, gtsam::Matrix & m, unsigned int /*version*/) {
+    template<class Archive,
+             typename Scalar_,
+             int Rows_,
+             int Cols_,
+             int Ops_,
+             int MaxRows_,
+             int MaxCols_>
+    void load(Archive & ar,
“I did some research and I found a couple of things:

1. Fixed sized matrices support resize for uniformity of the API though have assertions in place to ensure the resize is not incorrect. Quoting from the Eigen docs:

“All these methods are still available on fixed-size matrices, for the sake of API uniformity. Of course, you can't actually resize a fixed-size matrix. Trying to change a fixed size to an actually different value will trigger an assertion failure;”

1. I spent a lot of time researching this particular problem yesterday, since I found this slightly odd too. This StackOverflow answer was my main source of understanding, and I found confidence in this answer when I saw eigen_typekit in ROS use the same code. This gist also provides the same code, though they seem independently developed since the gist doesn't use some in-built Eigen functionality. This other gist does partial template specialization so putting it here as a refernce.
2. I agree with keeping only Scalar, Rows and Cols. It just seemed like a good idea to cover our bases by including all parameters, especially after seeing the examples in my above comment.

Thoughts?”

varunagrawal:
@dellaert I added a note not just in the PR but also the source code as a comment. This should help with future references.

varunagrawal:
Taken care of. :)

imu_tests

ProfFan:
It appears that after this PR I am seeing some warnings on the serialization tests:
```
In file included from /home/fan/Projects/CV/SLAM/gtsam/gtsam/3rdparty/Eigen/Eigen/Core:450,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/3rdparty/Eigen/Eigen/Dense:1,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/base/OptionalJacobian.h:22,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/base/Matrix.h:26,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/base/Manifold.h:22,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/geometry/BearingRange.h:21,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/geometry/Pose3.h:22,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/navigation/NavState.h:21,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/navigation/ManifoldPreintegration.h:24,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/navigation/CombinedImuFactor.h:26,
                 from /home/fan/Projects/CV/SLAM/gtsam/gtsam/navigation/CombinedImuFactor.cpp:23:
/home/fan/Projects/CV/SLAM/gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/DenseStorage.h: In member function ‘void boost::archive::detail::iserializer<Archive, T>::load_object_data(boost::archive::detail::basic_iarchive&, void*, unsigned int) const [with Archive = boost::archive::xml_iarchive; T = boost::optional<Eigen::Matrix<double, 3, 1> >]’:
/home/fan/Projects/CV/SLAM/gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/DenseStorage.h:194:66: warning: ‘<anonymous>.Eigen::DenseStorage<double, 3, 3, 1, 0>::m_data’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     DenseStorage(const DenseStorage& other) : m_data(other.m_data) {
                                                                  ^
/home/fan/Projects/CV/SLAM/gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/DenseStorage.h: In member function ‘void boost::archive::detail::iserializer<Archive, T>::load_object_data(boost::archive::detail::basic_iarchive&, void*, unsigned int) const [with Archive = boost::archive::text_iarchive; T = boost::optional<Eigen::Matrix<double, 3, 1> >]’:
/home/fan/Projects/CV/SLAM/gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/DenseStorage.h:194:66: warning: ‘<anonymous>.Eigen::DenseStorage<double, 3, 3, 1, 0>::m_data’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     DenseStorage(const DenseStorage& other) : m_data(other.m_data) {
                                                                  ^
/home/fan/Projects/CV/SLAM/gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/DenseStorage.h: In member function ‘void boost::archive::detail::iserializer<Archive, T>::load_object_data(boost::archive::detail::basic_iarchive&, void*, unsigned int) const [with Archive = boost::archive::binary_iarchive; T = boost::optional<Eigen::Matrix<double, 3, 1> >]’:
/home/fan/Projects/CV/SLAM/gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/DenseStorage.h:194:66: warning: ‘<anonymous>.Eigen::DenseStorage<double, 3, 3, 1, 0>::m_data’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     DenseStorage(const DenseStorage& other) : m_data(other.m_data) {
                                                                  ^
```

I changed some of these back to the old `data()`-based style and they disappear.

varunagrawal:
Yeah I'm seeing them too. I wonder if this PR uncovered some underlying issue with the serialization.

varunagrawal:
@ProfFan fixed the warnings in #404

jlblancoc:
For the records: if building with C++17 enabled, the following error shows up when deserializing gtsam factors:

```
/usr/include/eigen3/Eigen/src/Core/util/ForwardDeclarations.h:32: error: incomplete type ‘Eigen::internal::traits<boost::serialization::U>’ used in nested name specifier
   32 |   enum { has_direct_access = (traits<Derived>::Flags & DirectAccessBit) ? 1 : 0,
      |                              ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
```

After much digging, I found this to be a bug in the Eigen library when enabling C++17 with boost::serialization. 
The fix has not been released yet (v3.3.8 and/or v4.0.0?) so one has to manually pick [this patch](https://bitbucket.org/eigen/eigen/commits/28ca6976ccea/).

The only dirty solution seems to be a manual edit of /usr/ headers... :-( Alternatively, one could create a dedicated .cpp file just for serialization, and disable the c++17 flag for that file only until the next Eigen version is released.

Bright side is: it works.

varunagrawal:
I'll create an issue for this so we can track it in the future.

PR #392, opened 2020-07-10
Suppress 3rd Party Warnings in Release mode

varunagrawal:
Some of GTSAM's users have complained about GTSAM emitting warnings which are harmless but spam their logging system.

This PR suppresses the warnings on release build (only!) for 3rd party code to fix this issue.

Warnings from primarily GTSAM code will still show as is the best practice.

### DIFF
--- a/gtsam/CMakeLists.txt
+++ b/gtsam/CMakeLists.txt
@@ -106,7 +106,7 @@ add_library(gtsam ${gtsam_srcs})
 target_link_libraries(gtsam PUBLIC ${GTSAM_BOOST_LIBRARIES})
 target_link_libraries(gtsam PUBLIC ${GTSAM_ADDITIONAL_LIBRARIES})
 
-if(${CMAKE_BUILD_TYPE} STREQUAL "Release")
+if("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
   # Suppress warnings if Release build
   target_compile_options(gtsam PRIVATE -w)
 endif()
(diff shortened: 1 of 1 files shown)


jlblancoc:
Hmm... I don't think this is a good idea. 
Perhaps, only for the 3rd party projects, but not for the main gtsam/gtsam_unstable targets! Warnings are... warnings, and should be always considered :-) 

In fact, my position is closer to the opposite, adding `-Werror` but I recognize that's too drastic for large projects with many archs/compilers to be supported.

jlblancoc:
PS: If you were thinking of Windows users, it's true that there are tons of silly errors related to float/double/integer conversions that could be "safely" omitted. But then it would be better to make a finer control of what's disabled, via `/wdXXXX` flags. 

PS2: The commit adds "-w" without checking if we are in MSVC or GCC  ;-)

varunagrawal:
That's a great point @jlblancoc. I'll update this PR to only silence the 3rd party warnings.

varunagrawal:
@jlblancoc updated the PR.

gtsam/CMakeLists.txt
```diff
@@ -186,11 +186,17 @@ install(
 list(APPEND GTSAM_EXPORTED_TARGETS gtsam)
 set(GTSAM_EXPORTED_TARGETS "${GTSAM_EXPORTED_TARGETS}" PARENT_SCOPE)
 
-# make sure that ccolamd compiles even in face of warnings
+# Make sure that ccolamd compiles even in face of warnings
+# and suppress all warnings from 3rd party code if Release build
 if(WIN32)
-    set_source_files_properties(${3rdparty_srcs} PROPERTIES COMPILE_FLAGS "-w")
+  set_source_files_properties(${3rdparty_srcs} PROPERTIES COMPILE_FLAGS "-w")
```
jlblancoc, line 192: Maybe this should be "/w" ?

https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=vs-2019

varunagrawal: Looks like we found a bug then, since I didn't update that line of code except for the indentation. :tada:

Issue #242, opened 2020-03-06
Support for Landmarks as Vertices in g2o format


varunagrawal:
## Feature

Currently, GTSAM saves only camera poses as vertices in the g2o format using `writeG2o` and `readG2o`. We would like GTSAM to also save landmark points as vertices using the same machinery so we can save and retrieve 3D landmark points.

The format for landmarks as per the g2o repo is 

```
VERTEX_TRACKXYZ i x y z
```

This specifies a 3D point x<sub>i</sub> = (x, y, z)<sup>T</sup> which is located at (x, y, z)

## Motivation

This feature would allow us to read arbitrary g2o files and save/load both poses and landmarks.

## Pitch

Read and save both poses and points.

## Alternatives

N/A

## Additional context

https://github.com/RainerKuemmerle/g2o/wiki/File-format-slam-3d

PR #396, opened 2020-07-11
Support for landmarks in g2o files

varunagrawal:
Fix for #242.


varunagrawal:
So I only need to create unit tests for this via a sample dataset which also has landmarks. I believe the right approach is to just make my own simple dataset with a robot moving in a circle or a straight line.

gtsam/slam/dataset.h
```diff
@@ -76,14 +76,24 @@ enum KernelFunctionType {
...
+typedef std::pair<Key, Point2> IndexedLandmark;
 typedef std::pair<std::pair<Key, Key>, Pose2> IndexedEdge;
 
 /**
  * Parse TORO/G2O vertex "id x y yaw"
  * @param is input stream
  * @param tag string parsed from input stream, will only parse if vertex type
  */
-GTSAM_EXPORT boost::optional<IndexedPose> parseVertex(std::istream& is,
+GTSAM_EXPORT boost::optional<IndexedPose> parseVertexPose(std::istream& is,
```
dellaert, line 87: I see no reason for an API breaking name change, and if it were to be implemented we need a deprecated section with the old name.

varunagrawal: Sounds good. I'll use the compiler definition from #422 if that's okay?

dellaert: I prefer *not* to do the name change

varunagrawal:
It felt more consistent since I added a function named `parseVertexLandmark`. It also avoided ambiguity thanks to the name: you know you are parsing a g2o vertex which represents a pose.

If you still feel the consistency is not worth the extra deprecation effort, just give this message a *thumbs down* and I'll undo it. :slightly_smiling_face:

Issue #398, opened 2020-07-12
Build errors after #386 was merged.


Wallbraker:
## Description

Change 95b4a49f643efe0a8bdae4c04132cf6b110deab2 causes build errors in Kimera-VIO. See CI build: http://ci-sparklab.mit.edu:8080/blue/organizations/jenkins/MIT-SPARK-Kimera/detail/PR-100/2/pipeline

```
CMakeFiles/kimera_vio.dir/src/dataprovider/DataProviderModule.cpp.o:(.bss+0x0): multiple definition of `boost::archive::detail::extra_detail::init_guid<gtsam::PreintegrationCombinedParams>::g'
CMakeFiles/kimera_vio.dir/src/dataprovider/DataProviderInterface.cpp.o:(.bss+0x0): first defined here
CMakeFiles/kimera_vio.dir/src/dataprovider/EurocDataProvider.cpp.o:(.bss+0x10): multiple definition of `boost::archive::detail::extra_detail::init_guid<gtsam::PreintegrationCombinedParams>::g'
CMakeFiles/kimera_vio.dir/src/dataprovider/DataProviderInterface.cpp.o:(.bss+0x0): first defined here
CMakeFiles/kimera_vio.dir/src/dataprovider/KittiDataProvider.cpp.o:(.bss+0x0): multiple definition of `boost::archive::detail::extra_detail::init_guid<gtsam::PreintegrationCombinedParams>::g'
CMakeFiles/kimera_vio.dir/src/dataprovider/DataProviderInterface.cpp.o:(.bss+0x0): first defined here
CMakeFiles/kimera_vio.dir/src/frontend/StereoImuSyncPacket.cpp.o:(.bss+0x0): multiple definition of `boost::archive::detail::extra_detail::init_guid<gtsam::PreintegrationCombinedParams>::g'
CMakeFiles/kimera_vio.dir/src/dataprovider/DataProviderInterface.cpp.o:(.bss+0x0): first defined here
CMakeFiles/kimera_vio.dir/src/frontend/StereoVisionFrontEnd.cpp.o:(.bss+0x10): multiple definition of `boost::archive::detail::extra_detail::init_guid<gtsam::PreintegrationCombinedParams>::g'
CMakeFiles/kimera_vio.dir/src/dataprovider/DataProviderInterface.cpp.o:(.bss+0x0): first defined here
[...]

PR #399, opened 2020-07-12
Move the boost serialization declaration of CombinedImuFactor to cpp file

ProfFan:
#398


varunagrawal:
I am not completely happy with this solution, so I looked up [an alternative](https://stackoverflow.com/questions/32442778/boost-serialization-exporting-in-multiple-cpp-files/32447142#32447142).

@ProfFan I am going to hijack this PR and make those changes. I want to avoid declaring the Boost serialization export in multiple files.

PR #403, opened 2020-07-13
Basis functions

varunagrawal:
This PR adds support for Basis functions in GTSAM, such as Chebyshev, Chebyshev2 and Fourier.

Additionally it also has code for basis fitting which uses the newly introduced FunctorizedFactor.


varunagrawal:
Sounds good. :slightly_smiling_face: 

(PS build failed because Github had an outage around 2:30 am CT last night)

dellaert:
@gchenfc this is the PR I mentioned

dellaert:
Ping @JDFlorezCastillo @luoj1

varunagrawal:
FYI I may kill this PR because @dellaert in a previous meeting told me how to go about adding this in.

The idea is to basically create a branch in both GTSAM and Double-Expresso and move stuff incrementally so that things are perfectly in tune.

I'll also migrate double expresso to github.com so everyone who's tagged can look at the code. It's a lot of clever C++ (written by folks smarter than me) so don't hesitate to ask questions.

dellaert:
Don't kill before we chat

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,373 @@
...
+  template <int M>
+  class VectorEvaluationFunctor : protected EvaluationFunctor {
+   protected:
+    typedef Eigen::Matrix<double, M, 1> VectorM;
+    typedef Eigen::Matrix<double, M, -1> MatrixMN;
+    typedef Eigen::Matrix<double, /*MxMN*/ M, -1> Jacobian;
+    Jacobian H_;
+    void calculateJacobian() {
+      typedef Eigen::Matrix<double, M, M> MatrixM;
+      H_ = Eigen::kroneckerProduct(this->weights_, MatrixM::Identity());
```
JDFlorezCastillo, line 102: The [documentation](https://eigen.tuxfamily.org/dox/unsupported/group__KroneckerProduct__Module.html) warns against this syntax - it should be `kroneckerProduct(...).eval()`

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,373 @@
...
+    VectorM operator()(const MatrixMN& F,
+                       OptionalJacobian</*M*N*/ -1, -1> H = boost::none) const {
+      return apply(F, H);
+    }
+  };
+
+  /**
+   * Given M*N Matrix of M-vectors at N Chebyshev points, predict component for
+   * given row i, with 0<i<M.
+   */
```
JDFlorezCastillo, line 140: Component of what?

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,373 @@
...
+  template <int M>
+  class ComponentEvaluationFunctor : public EvaluationFunctor {
+   protected:
+    typedef Eigen::Matrix<double, M, -1> MatrixMN;
+    typedef Eigen::Matrix<double, /*1xMN*/ 1, -1> Jacobian;
+    size_t rowIndex_;
+    Jacobian H_;
+    void calculateJacobian(size_t N) {
```
JDFlorezCastillo, line 148: Is this a manual Kronecker product?

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,373 @@
...
+    using argument_type = typename DERIVED::Parameters;
+    using return_type = double;
+
+    DerivativeFunctor(size_t N, double x) : DerivativeFunctorBase(N, x) {}
+
+    DerivativeFunctor(size_t N, double x, double a, double b)
+        : DerivativeFunctorBase(N, x, a, b) {}
+
+    double apply(const typename DERIVED::Parameters& f,
+                 OptionalJacobian</*1xN*/ -1, -1> H = boost::none) const {
```
JDFlorezCastillo, line 266: Since 1 is known at compile time, perhaps do: `OptionalJacobian< 1, -1>` ?

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    /// Constructor, with optional interval [a,b]
+    VectorDerivativeFunctor(size_t N, double x, double a, double b)
+        : DerivativeFunctorBase(N, x, a, b) {
+      calculateJacobian();
+    }
+
+    VectorM apply(const MatrixMN& F,
+                  OptionalJacobian</*MxMN*/ -1, -1> H = boost::none) const {
```
JDFlorezCastillo, line 309: If M is known, could this instead be: `OptionalJacobian< M, -1>` ?
We interpreted that `M` is a template parameter, so it should already be defined.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,373 @@
...
+      if (H) *H = H_;
+      return F * this->weights_.transpose();
+    }
+    /// c++ sugar
+    VectorM operator()(const MatrixMN& F, OptionalJacobian</*MxMN*/ -1, -1> H =
+                                              boost::none) const {
+      return apply(F, H);
```
JDFlorezCastillo, line 316: Formatting?

JDFlorezCastillo (changes requested):
Partial review with comments, completed by @gchenfc, @silverwings-zero, and myself (@JDFlorezCastillo)

gtsam/basis/Basis.h
varunagrawal: That's only if we are replacing a matrix, which is not the case here.

> If you want to replace a matrix by its Kronecker product with some matrix, do NOT do this:

gtsam/basis/Basis.h
varunagrawal: Component of the vector.

gtsam/basis/Basis.h
varunagrawal: Kept it like this for consistency. This won't be a major improvement in efficiency since we barely use this particular functor.

gtsam/basis/Basis.h
varunagrawal: Yeah this is old code, definitely needs some more work.

gtsam/basis/tests/testFourier.cpp
```diff
@@ -0,0 +1,250 @@
...
+    graph.add(linearFactor);
+
+    // Create factor to predict value at x
+    PredictFactor<FourierBasis> predictFactor(
+        key, desiredValue, model, FourierBasis::EvaluationFunctor(N, x));
+
+    // Check expression Jacobians
+    EXPECT_CORRECT_FACTOR_JACOBIANS(predictFactor, values, 1e-5, 1e-9);
+
+    auto gaussianFactor = predictFactor.linearize(values);
```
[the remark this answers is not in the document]
varunagrawal: @dellaert can illuminate perhaps?

gtsam/basis/tests/testFourier.cpp
```diff
@@ -0,0 +1,250 @@
...
+  FourierBasis::DerivativeFunctor dfdx(7, x);
+  EXPECT_DOUBLES_EQUAL(numeric_dTdx(0, 0), dfdx(k7Coefficients), 1e-8);
+}
+
+//******************************************************************************
+TEST(Basis, VecDerivativeFunctor) {
+  using DotShape = typename FourierBasis::VectorDerivativeFunctor<2>;
+  const size_t N = 3;
+
+  // MATLAB example, Dec 27 2019, commit 014eded5
```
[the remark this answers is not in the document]
varunagrawal: Check the commit number?

varunagrawal:
So most of the comments by @JDFlorezCastillo seems to be in `Fourier.h` which was something @dellaert mainly worked on. I'll take the feedback into account but I may need help understanding some aspects myself.

gchenfc:
> So most of the comments by @JDFlorezCastillo seems to be in `Fourier.h` which was something @dellaert mainly worked on. I'll take the feedback into account but I may need help understanding some aspects myself.

Probably that was also because we didn't get through all the files - I think we didn't get to Chebyshev, Chebyshev2, or FitBasis

Also, I think most of the comments were superficial in any case

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,381 @@
...
+  /// Call weights for all x in vector, returns M*N matrix where M is the size
+  /// of the vector X.
+  static Matrix WeightMatrix(size_t N, const Vector& X) {
+    Matrix W(X.size(), N);
+    for (int i = 0; i < X.size(); i++)
+      W.row(i) = DERIVED::CalculateWeights(N, X(i));
+    return W;
+  }
+
+  /// Call weights for all x in vector, with interval [a,b]
```
dellaert, line 49: same

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,381 @@
...
+  /// Returns M*N matrix where M is the size of the vector X.
+  static Matrix WeightMatrix(size_t N, const Vector& X, double a, double b) {
+    Matrix W(X.size(), N);
+    for (int i = 0; i < X.size(); i++)
+      W.row(i) = DERIVED::CalculateWeights(N, X(i), a, b);
+    return W;
+  }
+
+  /**
+   * EvaluationFunctor at a given x, applied to Parameters.
```
dellaert, line 59: , calculates f(x;p).

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,381 @@
...
+   * This functor is used to evaluate a parameterized function at a given scalar
+   * value x. When given a specific N*1 vector of Parameters, returns the scalar
+   * value of the function at x, possibly with Jacobian wrpt the parameters.
+   */
+  class EvaluationFunctor {
+   protected:
+    typename DERIVED::Weights weights_;
```
dellaert, line 66: Again, looking at all derived classes, they just say

```
using Weights = Eigen::Matrix<double, /*1xN*/ 1, -1>;
```

i.e., Vector. So just define

```
using Weights = Vector;
```

front of Basis, and remove all mentions of DERIVED::Weights, as well as all typedefs in derived classes.

dellaert (changes requested):
Please do respond to all JD's in a meaningful way. I thought they were very germane and provided a set of fresh eyes. I helped below, and I think we have to resolve to mani issues before next review: (a) unsupported, (b) Weights.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,373 @@
...
+#include <gtsam/base/Matrix.h>
+#include <gtsam/base/OptionalJacobian.h>
+
+#include <gtsam/3rdparty/Eigen/unsupported/Eigen/KroneckerProduct>
+#include <iostream>
+
+/*
+ *  Concept of a Basis:
+ *    - type Weights, Parameters
+ *    - CalculateWeights(size_t N, double a=default, double b=default)
```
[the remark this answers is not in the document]
varunagrawal: That's an interesting point. In terms of the concept, I believe it is still valid since we do compute the weights for Chebyshev polynomials.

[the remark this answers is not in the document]
varunagrawal: Added the description. Thank you @dellaert!

gtsam/basis/tests/testFourier.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+//******************************************************************************
+TEST(Basis, BasisEvaluationFunctorDerivative) {
+  // If we give the H argument, we get the derivative of fx(0) wrpt coefficients
+  // Needs to be Matrix so it can be used by OptionalJacobian.
+  Matrix H(1, 3);
+  FourierBasis::EvaluationFunctor fx(3, 0);
+  EXPECT_DOUBLES_EQUAL(k3Coefficients[0] + k3Coefficients[1],
+                       fx(k3Coefficients, H), 1e-9);
+}
```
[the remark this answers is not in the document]
varunagrawal: Done!

gtsam/basis/tests/testFourier.cpp
```diff
@@ -0,0 +1,250 @@
...
+        key, desiredValue, model, FourierBasis::EvaluationFunctor(N, x));
+
+    // Check expression Jacobians
+    EXPECT_CORRECT_FACTOR_JACOBIANS(predictFactor, values, 1e-5, 1e-9);
+
+    auto gaussianFactor = predictFactor.linearize(values);
+    auto jacobianFactor =
+        boost::dynamic_pointer_cast<JacobianFactor>(gaussianFactor);
+    CHECK(jacobianFactor);  // makes sure it's indeed a JacobianFactor
+    EXPECT(assert_equal(linearFactor, *jacobianFactor, 1e-9));
```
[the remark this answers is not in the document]
varunagrawal: @dellaert this is for you :stuck_out_tongue_closed_eyes:

gtsam/basis/tests/testFourier.cpp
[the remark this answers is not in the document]
varunagrawal: Done

gtsam/basis/tests/testFourier.cpp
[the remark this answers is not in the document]
varunagrawal: I have updated the names but to be more descriptive of the behavior rather than the testing process.

gtsam/basis/tests/testFourier.cpp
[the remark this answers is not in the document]
varunagrawal: After looking at this in more detail, it is simply a test to check if the linearization of the factor is as expected. This is important since we deal with nonlinear functions all the time, but need to linearize for optimization.

gtsam/basis/tests/testFourier.cpp
```diff
@@ -0,0 +1,250 @@
...
+//******************************************************************************
+// Check derivative in two different ways: numerical and using D on f
+double proxy(double x) {
+  return FourierBasis::EvaluationFunctor(7, x)(k7Coefficients);
+}
+
+TEST(Basis, Derivative7) {
+  // Check Derivative evaluation at point x=0.2
+
+  // calculate expected values by numerical derivative of synthesis
```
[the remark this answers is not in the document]
varunagrawal: Good catch. Should be `proxy`.

gtsam/basis/tests/testFourier.cpp
```diff
@@ -0,0 +1,250 @@
...
+// generalization of the Fourier transform, and constitute a "pseudo-spectral"
+// parameterization. One way to do this is to establish hard constraints that
+// express the relationship between the new parameters and the coefficients.
+// For example, below I'd like the parameters to be the function values at
+// X = {0.1,0.2,0.3}, rather than a 3-vector of coefficients.
+// Because the values f(X) = at these points can be written as f(X) = W(X)*c,
+// we can simply express the coefficents c as c=inv(W(X))*f, which is a
+// generalized Fourier transform. That also means we can create factors with the
+// unknown f-values, as done manually below.
+// TEST(Basis, PseudoSpectral) {
```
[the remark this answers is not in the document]
varunagrawal: Uncommented.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,381 @@
...
+ * @date July 4, 2020
+ */
+
+#pragma once
+
+#include <gtsam/base/Matrix.h>
+#include <gtsam/base/OptionalJacobian.h>
+#include <gtsam/basis/ParameterMatrix.h>
+
+#include <gtsam/3rdparty/Eigen/unsupported/Eigen/KroneckerProduct>
```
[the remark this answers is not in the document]
varunagrawal: Should be `<unsupported/Eigen/KroneckerProduct>`, very strange structure.

gtsam/basis/Basis.h
varunagrawal: Done

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,381 @@
...
+ *    - CalculateWeights(size_t N, double a=default, double b=default)
+ */
+
+namespace gtsam {
+
+/// CRTP Base class for function bases
+template <typename DERIVED>
+class Basis {
+ public:
+  /// Call weights for all x in vector, returns M*N matrix where M is the size
```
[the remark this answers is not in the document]
varunagrawal: Done

gtsam/basis/Basis.h
varunagrawal: What if we use a parameterization that doesn't use a `Vector`? Code for the now and not the future?

gtsam/basis/Basis.h
varunagrawal: Updated to `OptionalJacobian<1, -1>`.

varunagrawal: I realized why I didn't do this before (and I am certain I gave Frank the same reasoning): `OptionalJacobian` doesn't support mixed static and dynamic sizes. We can either have it be fully static aka `OptionalJacobian<1, N>` or completely dynamic (which is what I have done).

Adding this support is not on my A->B and there is no real need for it per se since the ROI isn't valuable enough IMO.

gtsam/basis/Basis.h
varunagrawal: OptionalJacobian doesn't support mixed static and dynamic templating.

gtsam/basis/Basis.h
varunagrawal: Nope. This is the evaluation of the jacobian which I did by hand and derived this.

gtsam/basis/tests/testFourier.cpp
```diff
@@ -0,0 +1,250 @@
...
+  expected.row(1) << 1, cos(2), sin(2);
+  Vector2 X(1, 2);
+  Matrix actual = FourierBasis::WeightMatrix(3, X);
+  EXPECT(assert_equal(expected, actual, 1e-9));
+}
+
+//******************************************************************************
+TEST(Basis, Decomposition) {
+  // Create example sequence
+  Sequence sequence;
```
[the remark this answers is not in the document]
varunagrawal: I updated the docstring for `Sequence`.

gtsam/basis/tests/testFourier.cpp
```diff
@@ -0,0 +1,250 @@
...
+  const size_t N = 3;
+
+  // MATLAB example, Dec 27 2019, commit 014eded5
+  double h = 2 * M_PI / 16;
+  Vector2 dotShape(0.5556, -0.8315);  // at h/2
+  DotShape dotShapeFunction(N, h / 2);
+  Matrix23 theta = (Matrix32() << 0, 0, 0.7071, 0.7071, 0.7071, -0.7071)
+                       .finished()
+                       .transpose();
+  EXPECT(assert_equal(Vector(dotShape), dotShapeFunction(theta), 1e-4));
```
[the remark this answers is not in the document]
varunagrawal: @dellaert can you elaborate on this please? I can update the test once I understand what the intention was.

gtsam/basis/FitBasis.h
```diff
@@ -0,0 +1,84 @@
...
+
+  /// Constructor
+  FitBasis(size_t N, const Sequence& sequence, const SharedNoiseModel& model) {
+    GaussianFactorGraph::shared_ptr gfg = LinearGraph(sequence, model, N);
+    VectorValues solution = gfg->optimize();
+    parameters_ = solution.at(0);
+  }
+
+  /// Return Fourier coefficients
+  Parameters parameters() { return parameters_; }
```
gchenfc, line 81: function should be marked `const`?

gtsam/basis/FitBasis.h
```diff
@@ -0,0 +1,84 @@
...
+/// Our sequence representation is a map of {x: y} values where y = f(x)
+using Sequence = std::map<double, double>;
+
+/**
+ * Class that does Fourier Decomposition via least squares
+ */
+template <class Basis>
+class FitBasis {
+ public:
+  using Sample = std::pair<double, double>;
```
gchenfc, line 44: just curious, how come `Sequence` is defined in gtsam namespace and `Sample` is defined inside `FitBasis` ?

gtsam/basis/Fourier.h
```diff
@@ -0,0 +1,69 @@
...
+class GTSAM_EXPORT FourierBasis : public Basis<FourierBasis> {
+ public:
+  using Parameters = Eigen::Matrix<double, /*Nx1*/ -1, 1>;
+  using DiffMatrix = Eigen::Matrix<double, /*NxN*/ -1, -1>;
+
+  /**
+   *  Evaluate Real Fourier Weights of size N
+   *  e.g. N=5 yields bases: 1, cos(x), sin(x), cos(2*x), sin(2*x)
+   */
+  static Weights CalculateWeights(size_t N, double x) {
```
gchenfc, line 36: CRTP args `a`/`b`

gtsam/basis/Fourier.h
```diff
@@ -0,0 +1,69 @@
...
+      D(i, i + 1) = k;   // sin'(k*x) = k*cos(k*x)
+      D(i + 1, i) = -k;  // cos'(k*x) = -k*sin(k*x)
+      k += 1;
+    }
+
+    return D;
+  }
+
+  /// Get weights at a given x that calculate the derivative.
+  static Weights DerivativeWeights(size_t N, double x) {
```
gchenfc, line 63: CRTP args a & b

gtsam/basis/ParameterMatrix.h
```diff
@@ -0,0 +1,218 @@
...
+#include <gtsam/base/VectorSpace.h>
+
+#include <iostream>
+
+namespace gtsam {
+
+/**
+ * A matrix abstraction of MxN values at the Basis points.
+ * This class serves as a wrapper over an Eigen matrix.
+ * @tparam T: the type you would like to evaluate using polynomial
```
gchenfc, line 33: is `T` supposed to be `M` ???

Also can you give a quick reason this is necessary over just normal eigen matrices?  Is it for type-safety reasons?

gtsam/basis/basis.i
```diff
@@ -0,0 +1,126 @@
...
+  ParameterMatrix(const Matrix& matrix);
+
+  Matrix matrix() const;
+
+  void print(const string& s = "") const;
+};
+
+#include <gtsam/basis/BasisFactors.h>
+
+template <BASIS = {gtsam::Chebyshev2}>
```
gchenfc, line 68: Add other bases (`Chebychev` & `Fourier`) ?    (once missing CRTP requirements are filled)

gtsam/basis/basis.i
```diff
@@ -0,0 +1,126 @@
...
+                         const gtsam::noiseModel::Base* model, const size_t N,
+                         double x, double a, double b);
+};
+
+typedef gtsam::VectorEvaluationFactor<gtsam::Chebyshev2, 3>
+    VectorEvaluationFactorChebyshev2D3;
+typedef gtsam::VectorEvaluationFactor<gtsam::Chebyshev2, 4>
+    VectorEvaluationFactorChebyshev2D4;
+typedef gtsam::VectorEvaluationFactor<gtsam::Chebyshev2, 12>
+    VectorEvaluationFactorChebyshev2D12;
```
gchenfc, line 95: I wonder if there's any better way to do this?  Maybe that's a later problem.

gtsam/basis/tests/testChebyshev.cpp
```diff
@@ -0,0 +1,179 @@
...
+  // at times
+  const size_t m = 6;
+  Vector t(m);
+  t << -0.7, -0.4, 0.1, 0.3, 0.7, 0.9;
+  Vector x(m);
+  x << -0.7, -0.4, 0.1, 0.3, 0.7, 0.9;
+
+  for (size_t i = 0; i < m; i++) {
+    graph.emplace_shared<EvaluationFactor<Chebyshev1Basis>>(key, x(i), model, N,
+                                                            t(i));
```
gchenfc, line 86: Referencing above comment, if you add the optional args `a`, `b` then this will fail due to CRTP missing reqs

python/gtsam/preamble/basis.h
```diff
@@ -0,0 +1,12 @@
+/* Please refer to:
```
gchenfc, line 1: @varunagrawal I added this function for the new multiple-interface-files - can you check if I missed anything?

python/gtsam/specializations/basis.h
```diff
@@ -0,0 +1,12 @@
+/* Please refer to:
```
gchenfc, line 1: @varunagrawal I added this function - can you check if I missed anything?

gchenfc (approved):
I pushed a couple changes just to update the python wrapper / interface files.  Sorry for the commit noise - when pulling in latest changed from develop I accidentally did a squash merge and github got confused.

1. Mainly I'm not convinced kronecker product is necessary - I would propose to just do those manually to avoid the extra dependency / eigen compile flag needing to be set.
2. Make sure all CRTP requirements are satisfied by relevant classes.
3. Some documentation nits.

Personally, I'm satisfied otherwise

gtsam/basis/Chebyshev.h
```diff
@@ -0,0 +1,92 @@
...
+ * functions. In this sense, they are like the sines and cosines of the Fourier
+ * basis.
+ */
+struct Chebyshev2Basis : Basis<Chebyshev2Basis> {
+  using Parameters = Eigen::Matrix<double, -1, 1 /*Nx1*/>;
+
+  /**
+   *  Evaluate Chebyshev Weights on [-1,1] at any x up to order N-1 (N values)
+   */
+  static Weights CalculateWeights(size_t N, double x);
```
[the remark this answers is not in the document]
varunagrawal: IIRC this is because the derivatives for Chebyshev1 is not as straightforward? I'll look into adding the functions.

gtsam/basis/Chebyshev.h
```diff
@@ -0,0 +1,92 @@
...
+ * These are typically denoted with the symbol T_n, where n is the degree.
+ * The parameter N is the number of coefficients, i.e., N = n+1.
+ */
+struct Chebyshev1Basis : Basis<Chebyshev1Basis> {
+  using Parameters = Eigen::Matrix<double, -1, 1 /*Nx1*/>;
+
+  /**
+   *  Evaluate Chebyshev Weights on [-1,1] at x up to order N-1 (N values)
+   */
+  static Weights CalculateWeights(size_t N, double x);
```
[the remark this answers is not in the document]
varunagrawal: Do CRTP requirements need to be set? It's not the same as inheritance so adding them even when they don't make sense in the context isn't the right thing to do.

gtsam/basis/FitBasis.h
varunagrawal: This is a question for @dellaert

gtsam/basis/Fourier.h
varunagrawal: Yeah this is an example of not making sense. A Fourier series is not defined in a certain interval (unlike a Chebyshev series which is explicitly defined on [-1,1].

gtsam/basis/tests/testChebyshev.cpp
varunagrawal: CRTP doesn't require you to fulfill the same method signatures. It's essentially a C++ trick to get access to templating classes in the template.

gtsam/basis/basis.i
varunagrawal: What is the problem exactly?

varunagrawal:
Github is not allowing me to respond to your comment about the Kronecker Product. My consideration about that was that there is zero overhead in compilation since those are header only libraries, and they've been written by the maintainers of Eigen who are way smarter than us when it comes to writing optimized C++ code for numerical computing.

Yeah sure we could define our own KroneckerProduct function, but would it be as clean/efficient/bug-free as the default Eigen version? Not so sure about that.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,411 @@
...
+#include <unsupported/Eigen/KroneckerProduct>
+
+/**
+ * This file supports creating continuous functions `f(x;p)` as a linear
+ * combination of `basis functions` such as the Fourier basis on SO(2) or a set
+ * of Chebyshev polynomials on [-1,1].
+ * In the expression `f(x;p)` the variable `x` is
+ * the continuous argument at which teh function is evaluated, and `p` are
+ * parameters, e.g., coefficients used to weight the different basis functions.
+ * However, different parameterizations are also possible.
```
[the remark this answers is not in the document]
varunagrawal: Isn't this numerical analysis 101 stuff? I may be wrong since I've only ever sat in on one numerical analysis course.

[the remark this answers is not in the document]
gchenfc: Not sure - either way, JD and Michael use this code and it took us all ~an hour of reading through this code together to finally figure out what was going on :)   (this was prior to the last review cycle, it's much better now but still)

Reading through it a second time last night it still took me ~5 minutes to wrap my head around (no fault of your own, just I wasn't sure what "Weights" was referring to and I didn't see a literature reference.  I looked on [wikipedia](https://en.wikipedia.org/wiki/Chebyshev_polynomials) and [wolfram mathworld](https://mathworld.wolfram.com/ChebyshevPolynomialoftheSecondKind.html) but neither used the nomenclature "weights".

Maybe just giving a quick hint would help more users?  Just giving my impressions as someone new to the code :)

gtsam/basis/Chebyshev.h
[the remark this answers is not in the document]
gchenfc: Oh I guess I just assumed it would be best practice but maybe not?  I'm not that familiar.

I guess I was just thinking that, as a user, I was looking at `Basis` for all the functions I could use and wanted a Chebyshev1 EvaluatorFunctor with interval from 0 to 1 but then I got a compile error and was surprised (because that constructor existed in Basis but not for Chebyshev1).

Maybe a thought: that if it doesn't satisfy a CRTP requirement then maybe the CRTP base should not include that? (e.g. split into 2 bases to separate out functionality that isn't common)?  I think actually in this case there is a reasonable definition of the domain [a, b] for all inherited types (including Fourier, see comment below).  For derivatives, one option could be placing derivatives in a separate CRTP base?

Again, I don't know as much about this but just offering possible suggestions.

gtsam/basis/Fourier.h
gchenfc: I'd argue it could still make sense - I may want it periodic with period 1 or with period 10.  So although I guess you wouldn't need both `a` and `b` (unless you're concerned about numerical precision of pi and doing calculations centered at e.g. 1e9) it could still make sense to define the period.

gtsam/basis/Chebyshev.h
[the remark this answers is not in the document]
gchenfc: Just thought of another possible suggestion: variadic function templates?  This could maybe mitigate the confusion of a definition that exists in Basis but that cannot be used for certain basis functions, since the variadic template makes semantically clear that it depends on the derived implementation what type of calls are possible.

gtsam/basis/basis.i
gchenfc: e.g. If I'm a pure python user (`pip install gtsam`) and want a VectorEvaluationFactor with size 17.

Hardcoding a finite number of options seems not ideal, but not a huge deal for now.

gtsam/basis/tests/testChebyshev.cpp
gchenfc: Oh I see.  But in this case it would be logical to want to use such a Basis, right?  A Chebyshev1 that's scaled to a different domain?  Correct me if I'm wrong - I'm not super familiar with these stuffs.

gchenfc:
Regarding the kroneckerProduct, true, i noticed the re-compile after setting flag was instantaneous.  And to clarify, I meant that kroneckerProduct is only being used with a second argument of the identity matrix which is less of a "product" and more of just a storage/assignment operator (there's just a  bunch of multiplies by 0 and multiplies by 1).  Thus I would propose a function that does just these assignments instead of using kroneckerProduct.  I'm not sure what the eigen implementation is, but if naive, it could be o(n^4) multiplies which are not really necessary.

Regarding the force push on the wrapper, could you add back the `xxxFactor`s that you deleted? ;) JD and I were using those (and FitBasis would also be helpful, while you're at it...).  And also I think you forgot to add `basis.i` to the python CMakeLists?  I'm not clear on what was wrong with the version I had pushed.

gtsam/basis/basis.i
varunagrawal: Ah I see. Yeah that is a problem, let me add a TODO for it.

gtsam/basis/Basis.h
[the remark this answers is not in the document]
varunagrawal: Yeah I updated the docs. Weights should only refer to the barycentric interpolation weights. I've used Parameters in the docstring instead since they represent the coefficients.

gtsam/basis/Chebyshev.h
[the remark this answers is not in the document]
varunagrawal: I see what you mean. Giving users the ability to restrict the domain is easy so I'll add it.

gtsam/basis/ParameterMatrix.h
varunagrawal: The idea is to have a matrix where the rows define the dimensionality of the problem (e.g. `ParameterMatrix<traits<STATE>::dimension>`) but whose columns are variable in number and can be adjusted as per the value of `N`.

varunagrawal:
@gchenfc make a PR for those extra factors targeting this branch?

gtsam/basis/basis.i
varunagrawal: The key issue here is that we want type-safety by specifying the template, but then the wrapper becomes something annoying to deal with.

gtsam/basis/Chebyshev.h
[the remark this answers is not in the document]
varunagrawal: I don't see any math for how to restrict the limits of the polynomials for Cheb1 and Cheb2, but I'm putting in the limits for consistency.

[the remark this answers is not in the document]
varunagrawal: This actually goes back to the limit not making sense since for Chebyshev1 polynomials, a factor with `graph.emplace_shared<EvaluationFactor<Chebyshev1Basis>>(key, x(i), model, N, t(i), -1.0, 1.0);` essentially does nothing compared to `graph.emplace_shared<EvaluationFactor<Chebyshev1Basis>>(key, x(i), model, N, t(i));`.

This is esoteric knowledge however.

gtsam/basis/basis.i
ProfFan: There could be a `Dynamic` tag like Eigen, but you will need to deal with the dims in C++. Static dispatch should be optimized by the compiler, so it's mainly burden in coding not at runtime.

gchenfc:
> @gchenfc make a PR for those extra factors targeting this branch?

#832

gtsam/basis/basis.i
gchenfc: I'll resolve this because I think the TODO is sufficient.

But for the future, using just -1 for dynamic shouldn't be too difficult, just adding constructor args for `M` when template param `M`=-1, e.g. with partial template specialization?

gtsam/basis/ParameterMatrix.h
gchenfc: Hmm, still is there a reason to not just
```c++
template <int M>
using ParameterMatrix = Eigen::Matrix<double, M, -1>
```
?

varunagrawal: I honestly don't remember. There was a reason when I first implemented it. I'll have to go back and look at doubleExpresso to find the reason.

varunagrawal:
@gchenfc thanks a ton! Your fixes also fixed the Windows build.

cmake/HandleEigen.cmake
```diff
@@ -6,7 +6,7 @@ option(GTSAM_USE_SYSTEM_EIGEN "Find and use system-installed Eigen. If 'off', us
 if(NOT GTSAM_USE_SYSTEM_EIGEN)
   # This option only makes sense if using the embedded copy of Eigen, it is
   # used to decide whether to *install* the "unsupported" module:
-  option(GTSAM_WITH_EIGEN_UNSUPPORTED "Install Eigen's unsupported modules" OFF)
+  option(GTSAM_WITH_EIGEN_UNSUPPORTED "Install Eigen's unsupported modules" ON)
```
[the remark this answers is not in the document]
varunagrawal: Okay, removed the need for `EIGEN_UNSUPPORTED` by defining `kroneckerProductIdentity`. Please be sure to review that carefully. :slightly_smiling_face:

gtsam/basis/ParameterMatrix.h
varunagrawal: I remembered the reason.
Doing 
```cpp
template <int M>
using ParameterMatrix = Eigen::Matrix<double, M, -1>
```
causes the API to become far more verbose and non-intuitive.

E.g., we can no longer do `ParameterMatrix<12> X(N)` since it gives an error that we tried calling a vector method on a matrix object. We have to instead do
```cpp
ParameterMatrix<12> X;
X.resize(12, N); // why do we need to specify 12 again here? This is something eigen needs :(
```

In general, the fact that one of the dimensions is dynamic and the other has size greater than 1 makes things non-standard for Eigen since it assumes  everything is a Matrix and simple things like `X.col` don't compile.

I'll add this to the docstring for future reference.

varunagrawal:
@dellaert can this please get another review from you?

gtsam/basis/basis.i
```diff
@@ -131,13 +126,6 @@ virtual class ManifoldEvaluationFactor : gtsam::NoiseModelFactor {
 // `ComponentDerivativeFactor`
 
 #include <gtsam/basis/FitBasis.h>
-// We'll allow transparent binding of python dict to Sequence in this
-// compilation unit using pybind11/stl.h.
-// Another alternative would be making Sequence opaque in
-// python/gtsam/{preamble, specializations}, but std::map<double, double> is
-// common enough that it may cause collisions, and we don't need
-// reference-access anyway.
-#include <pybind11/stl.h>
```
gchenfc: (See also `specializations/basis.h`)

I should have made a python unit test in #832 , but can you give a MWE of how to use FitBasis in python?  I can't figure out how (before this commit, it was just `gtsam.FitBasisChebyshev2(2, {0:0, 1:1, 0.5:0.5}, noiseModel).parameters()`).

(I don't think it is possible with the version you have now; I get this error:
```python
>>> gtsam.Sequence({1:2})
  File "<stdin>", line 1, in <module>
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. gtsam.gtsam.Sequence()

Invoked with: {1: 2}
```

I think you also need the requisite line in the preamble, right?:
```c++
PYBIND11_MAKE_OPAQUE(std::map<double, double>);
```
)
But then the syntax for using e.g. FitBasis is a little bit cumbersome:
```python
data = gtsam.Sequence()
for k, v in {0:0, 0.5:0.5, 1:1}.items():
    data[k] = v
gtsam.FitBasisChebyshev2(2, data, noiseModel).parameters()
# syntax with <pybind11/stl.h>:
# gtsam.FitBasisChebyshev2(2, {0:0, 1:1, 0.5:0.5}, noiseModel).parameters()
```
(the constructor `gtsam.Sequence({...})` doesn't exist)

---

By the way, although not an immediate issue, this also has the side-effect of causing binding collisions.  For example, if we ever had another type:
```c++
py::bind_map<gtsam::Sequence>(m_, "Sequence");
py::bind_map<std::map<double, double>>(m_, "AnotherMapDouble");
```
this will throw the runtime python error:
```
>>> import gtsam
  File "<stdin>", line 1, in <module>
  File "/home/gerry/miniconda3/lib/python3.8/site-packages/gtsam-4.1.0-py3.8.egg/gtsam/__init__.py", line 2, in <module>
    from .gtsam import *
ImportError: generic_type: type "AnotherMapDouble" is already registered!
```

This collision occurs if we have any other `map<double, double>` in the entire gtsam namespace (even outside the basis module, e.g. in the geometry module).  In contrast, including `pybind11/stl.h` as I had done before stays within the compilation unit, i.e. it doesn't leave basis.i and we don't get any binding collisions.

Sorry for the long comment but I had already spent a pretty considerable time thinking about / testing these considerations when I made #832 and had pretty intentionally chosen to do it this way, so we could have maybe saved time if we had discussed this together :)

python/gtsam/specializations/basis.h
```diff
@@ -10,3 +10,5 @@
  * with `PYBIND11_MAKE_OPAQUE` this allows the types to be modified with Python,
  * and saves one copy operation.
  */
+
+py::bind_map<gtsam::Sequence>(m_, "Sequence");
```
gchenfc, line 14: see comment above

gchenfc (commented):
Sorry for the long comment regarding wrapping Sequence.
tl;dr, can you please provide a MWE of using FitBasis in python?  I'm having trouble on my end getting a version to work with your wrapper changes.
*Edit: figured it out - I "approve" this but do just have some comments and you can decide what you think

gtsam/basis/ParameterMatrix.h
```diff
@@ -79,29 +79,27 @@ class ParameterMatrix {
...
-   * @param index: The row index to set.
-   * @param row: The Eigen (row) vector used to set the values.
+   * Get the matrix column specified by `index`.
+   * @param index: The column index to retrieve.
    */
-  void setRow(size_t index, const Eigen::Matrix<double, 1, -1>& row) {
-    matrix_.row(index) = row;
+  Eigen::Matrix<double, M, 1> col(size_t index) const {
+    return matrix_.col(index);
   }
```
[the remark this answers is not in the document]
varunagrawal: Would this work? Isn't p2 now const?

gtsam/basis/basis.i
varunagrawal: The issue with the stl.h include is that it now pulls in that file for all languages, such as Matlab. Did you try testing that with the matlab wrapper? Or a setup where someone only needs Matlab but no python wrapping?

gtsam/basis/Basis.h
```diff
@@ -56,6 +55,26 @@ namespace gtsam {
...
+ */
+template<size_t M>
+Matrix kroneckerProductIdentity(const Weights& x) {
+  const auto I = Eigen::Matrix<double, M, M>::Identity();
+  Matrix result(M, x.cols() * M);
+  result.setZero();
+
+  for (int i = 0; i < x.cols(); i++) {
+    result.block(0, i*M, M, M) = x(i) * I;
+  }
```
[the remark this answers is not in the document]
varunagrawal: Sweet. I was aiming for correctness over efficiency, so I'll add this.

gtsam/basis/basis.i
varunagrawal: I'll also double check the binding collisions, in which case this change is no bueno.

gtsam/basis/ParameterMatrix.h
gchenfc: Ahh got it - makes sense!  Thanks!

gtsam/basis/ParameterMatrix.h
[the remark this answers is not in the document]
gchenfc: Ya, so what'll happen is `p2.row(r)` will return a temporary Matrix (copy) and that temporary matrix is modified then immediately discarded.

With stricter compile flags conceivably it could throw unused warnings or something, but this passed on my machine with default gtsam settings.

In any case, in retrospect after your comment about the purpose of ParameterMatrix, it's a lot more obvious these are just mimicking Eigen's behavior so I don't think this is needed anymore.

gtsam/basis/basis.i
gchenfc: Oh shoot sorry I forgot about the matlab wrapper.  My mistake!!!

Binding collisions still applies tho.  I'd be ok calling this a later problem, since currently there's nothing else trying to bind to `std::map<double, double>` (build for now, not for the future)

Not sure but I think maybe placing Sequence inside FitBasis's scope, or inside a different namespace might fix binding collisions???  Something about pybind11 `py::module_local()` but I didn't read in detail: https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#binding-stl-containers

varunagrawal: Nah I want to finish this now. A later problem is not a good enough excuse for me.

varunagrawal: Great find on `py::module_local`. That should do the trick.

> This module-local binding is designed to avoid potential conflicts between module bindings (for example, from two separate modules each attempting to bind std::vector<int> as a python type).
 
> It is possible to override this behavior to force a definition to be either module-local or global. To do so, you can pass the attributes `py::module_local()` (**to make the binding module-local**) or py::module_local(false) (to make the binding global) into the py::bind_vector or py::bind_map arguments

varunagrawal: Uhh guess not

ProfFan: I am lost, what is your issue here?

gchenfc: @ProfFan 
Trying to bind two different `std::map<double, double>` with
```c++
py::bind_map<std::map<double, double>>(m_, "TypeName1");
py::bind_map<std::map<double, double>>(m_, "TypeName2");
```
will cause a python runtime error on import.  This occurs even if the types are bound in different interface files.

I think changing `m_` to a "more local module" would fix this, if possible, but I haven't played with it.  It may also just not be possible to do what we want.

ProfFan: module_local is per-module, if we are building several modules (.so) then it should work.

varunagrawal: I figured it out!!!

Now you can do `coefficients = gtsam.FitBasisChebyshev2(2, {0:0, 1:1, 0.5:0.5}, noiseModel).parameters()` no problem.

varunagrawal: Unfortunately, the module level thing doesn't work since we are linking all the separate .so files into one. By module, I believe they mean individual files that are not linked together, e.g.

```
gtsam
	- base
		- __init__.py
		- base.so
	- basis
		- __init__.py
		- basis.so
	- linear
		- __init__.py
		- linear.so
...
```

varunagrawal:
Okay the python tests are breaking and I can't figure out why. I am super slammed with other stuff, so let's add the wrapper stuff in a separate PR?

varunagrawal:
Note to squash and merge.

gtsam/basis/basis.i
gchenfc: Opening a PR where I added the <pybind11/stl.h> include to ".../preamble/basis.h" and everything seems to work fine.

This is a separate PR but I think:
1) as Varun discovered, the bindings for `IndexPairXXX` should be updated/fixed at some point
2) binding collisions is actually not something we should be worried about because we shouldn't really have more than one type describing a `map<double, double>` anyway, i.e. just opaque-ify and bind std::map<double, double>

gtsam/basis/Basis.h
dellaert: Add a comment what Jacobian is then? e.g.
```
/*
Calculate the `1*(M*N)` Jacobian of `apply` with respect to the M*N parameter matrix `P`.
We flatten assuming column-major order, e.g., if N=3 and M=2, we have 
	H=[w(0) 0  w(1) 0  w(2) 0] for rowIndex==0
	H=[0 w(0)  0 w(1)  0 w(2)] for rowIndex==1
i.e., one row of the Kronecker product of weights_ with the MxM identity matrix.
See also VectorEvaluationFunctor.
*/
```

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+   */
+  template <int M>
+  class VectorEvaluationFunctor : protected EvaluationFunctor {
+   protected:
+    using VectorM = Eigen::Matrix<double, M, 1>;
+    using Jacobian = Eigen::Matrix<double, /*MxMN*/ M, -1>;
+    Jacobian H_;
+
+    void calculateJacobian() {
+      using MatrixM = Eigen::Matrix<double, M, M>;
```
dellaert, line 161: unused...

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+  - `CalculateWeights(size_t N, double x, double a=default, double b=default)`
+  - `DerivativeWeights(size_t N, double x, double a=default, double b=default)`
+ */
+
+namespace gtsam {
+
+using Weights = Eigen::Matrix<double, 1, -1>; /* 1xN vector */
+
+/**
+ * @brief Function for computing the kronecker product of Weight matrix with
```
dellaert, line 59: Weight matrix -> the 1*N Weight vector `w`

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+ * Identity matrix efficiently. The primary reason for this is so that we don't
```
dellaert, line 60: the M*M identity matrix `I`

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+      return apply(F, H);
+    }
+  };
+
+  /**
+   * Manifold EvaluationFunctor at a given x, applied to ParameterMatrix<M>.
+   * This functor is used to evaluate a parameterized function at a given scalar
+   * value x. When given a specific M*N parameters, returns an M-vector the M
+   * corresponding functions at x, possibly with Jacobian wrpt the parameters.
+   */
```
dellaert, line 247: Add:
```
The difference with the VectorEvaluationFunctor is that after computing the M*1 vector
xi=F(x;P), with x a scalar and P the M*N parameter vector, we also retract xi back to the
T manifold. For example, if T==Rot3, then we first compute a 3-vector xi using x and P,
and then map that 3-vector xi back to the Rot3 manifold, yielding a valid 3D rotation.
```

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+
+    DerivativeFunctorBase(size_t N, double x, double a, double b)
+        : weights_(DERIVED::DerivativeWeights(N, x, a, b)) {}
+
+    void print(const std::string& s = "") const {
+      std::cout << s << (s != "" ? " " : "") << weights_ << std::endl;
+    }
+  };
+
+  /// Given values f at the Chebyshev points, predict derivative at x
```
dellaert, line 312: As I did above, expand to a full doxygen comment and explain this better? Also, not Chebyshev in this file.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+    /// c++ sugar
+    VectorM operator()(
+        const ParameterMatrix<M>& F,
+        OptionalJacobian</*MxMN*/ -1, -1> H = boost::none) const {
+      return apply(F, H);
+    }
+  };
+
+  /**
+   * Given M*N Matrix of M-vectors at N Chebyshev points, predict derivative for
```
dellaert, line 379: Not Chebyshev, expand a bit.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+
+  /// Compute derivative vector of ParameterMatrix at specified point.
+  template <int M>
+  class VectorDerivativeFunctor : protected DerivativeFunctorBase {
+   protected:
+    using VectorM = Eigen::Matrix<double, M, 1>;
+    using Jacobian = Eigen::Matrix<double, /*MxMN*/ M, -1>;
+    Jacobian H_;
+
+    void calculateJacobian() {
```
dellaert, line 343: Copy suggested comment from VectorEvaluationFunctor.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+   * given row i, with 0<=i<M.
+   */
+  template <int M>
+  class ComponentDerivativeFunctor : protected DerivativeFunctorBase {
+   protected:
+    using Jacobian = Eigen::Matrix<double, /*1xMN*/ 1, -1>;
+    size_t rowIndex_;
+    Jacobian H_;
+
+    void calculateJacobian(size_t N) {
```
dellaert, line 389: Copy suggested comment from VectorComponentFunctor.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+   * corresponding functions at x, possibly with Jacobian wrpt the parameters.
+   */
+  template <int M>
+  class VectorEvaluationFunctor : protected EvaluationFunctor {
+   protected:
+    using VectorM = Eigen::Matrix<double, M, 1>;
+    using Jacobian = Eigen::Matrix<double, /*MxMN*/ M, -1>;
+    Jacobian H_;
+
+    void calculateJacobian() {
```
dellaert, line 160: Comment on what the Jacobian is, e.g.

```
/*
Calculate the `M*(M*N)` Jacobian of `apply` with respect to the M*N parameter matrix `P`.
We flatten assuming column-major order, e.g., if N=3 and M=2, we have
	H =[
		w(0) 0  w(1) 0  w(2) 0
		0 w(0)  0 w(1)  0 w(2)
	   ]
i.e., the Kronecker product of weights_ with the MxM identity matrix.
*/
```

Note *not* Doxygen as is private and not part of API

gtsam/basis/BasisFactors.h
```diff
@@ -0,0 +1,240 @@
...
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ *  @file  BasisFactors.h
+ *  @author Varun Agrawal
```
dellaert, line 14: At least add a @brief comment?

gtsam/basis/BasisFactors.h
```diff
@@ -0,0 +1,240 @@
...
+#pragma once
+
+#include <gtsam/basis/Basis.h>
+#include <gtsam/nonlinear/FunctorizedFactor.h>
+
+namespace gtsam {
+
+/**
+ * Factor for scalar BASIS evaluation.
+ * @param BASIS: The basis class to use e.g. Chebyshev2
```
dellaert, line 26: @tparam ?

dellaert (changes requested):
Lots of comments in Basis.h

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+#include <gtsam/base/OptionalJacobian.h>
+#include <gtsam/basis/ParameterMatrix.h>
+
+#include <iostream>
+
+/**
+ * This file supports creating continuous functions `f(x;p)` as a linear
+ * combination of `basis functions` such as the Fourier basis on SO(2) or a set
+ * of Chebyshev polynomials on [-1,1].
+ * In the expression `f(x;p)` the variable `x` is
```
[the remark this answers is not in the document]
varunagrawal: Still get the same thing. :confused:

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+   */
+  static Matrix WeightMatrix(size_t N, const Vector& X) {
+    Matrix W(X.size(), N);
+    for (int i = 0; i < X.size(); i++)
+      W.row(i) = DERIVED::CalculateWeights(N, X(i));
+    return W;
+  }
+
+  /**
+   * Call weights for all x in vector X, with interval [a,b].
```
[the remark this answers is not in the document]
varunagrawal: In that case, the method `CalculateWeights` already exists for computing a weight vector at a specific `x`. `WeightMatrix` computes the weight vectors for multiple `x` values.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+ * to optimize over the parameters `p`.
+
+ * This functionality is implemented using the `CRTP` or "Curiously recurring
+ * template pattern" C++ idiom, which is a meta-programming technique in which
+ * the derived class is passed as a template argument to `Basis<DERIVED>`.
+ * The DERIVED class is assumed to satisfy a C++ concept,
+ * i.e., we expect it to define the following types and methods:
+
+  - type `Parameters`: the parameters `p` in f(x;p)
+  - `CalculateWeights(size_t N, double x, double a=default, double b=default)`
```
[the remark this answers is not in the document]
varunagrawal: The weights are interpolation weights (aka Barycentric interpolation weights in the case of Chebyshev2).

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+ * @brief Function for computing the kronecker product of Weight matrix with
+ * Identity matrix efficiently. The primary reason for this is so that we don't
+ * need to use the Eigen Unsupported library.
+ *
+ * @tparam M Size of the identity matrix.
+ * @param x The weights of the polynomial.
+ * @return Matrix
+ */
+template <size_t M>
+Matrix kroneckerProductIdentity(const Weights& x) {
```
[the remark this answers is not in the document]
varunagrawal: Only place where it is used. We don't need a Kronecker Product anywhere else.

gtsam/basis/basis.i
ProfFan: I think the best way is to alias **IN Python**. Just create an alias in the Python files and nothing to worry, yay.

gtsam/basis/Basis.h
```diff
@@ -0,0 +1,430 @@
...
+ * This functionality is implemented using the `CRTP` or "Curiously recurring
+ * template pattern" C++ idiom, which is a meta-programming technique in which
+ * the derived class is passed as a template argument to `Basis<DERIVED>`.
+ * The DERIVED class is assumed to satisfy a C++ concept,
+ * i.e., we expect it to define the following types and methods:
+
+  - type `Parameters`: the parameters `p` in f(x;p)
+  - `CalculateWeights(size_t N, double x, double a=default, double b=default)`
+  - `DerivativeWeights(size_t N, double x, double a=default, double b=default)`
+ */
```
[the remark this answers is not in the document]
varunagrawal: Done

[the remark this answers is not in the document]
varunagrawal: Once we get the `FitBasis` working in the wrapper, I can add a note to look at the example script.

gtsam/basis/BasisFactors.h
```diff
@@ -0,0 +1,243 @@
...
+#pragma once
+
+#include <gtsam/basis/Basis.h>
+#include <gtsam/nonlinear/FunctorizedFactor.h>
+
+namespace gtsam {
+
+/**
+ * @brief Factor for scalar BASIS evaluation.
+ *
```
dellaert, line 28: This needs to be explained. When is this factor used? What is z? What error does it evaluate?

gtsam/basis/BasisFactors.h
```diff
@@ -0,0 +1,243 @@
...
+  EvaluationFactor(Key key, const double &z, const SharedNoiseModel &model,
+                   const size_t N, double x, double a, double b)
+      : Base(key, z, model, typename BASIS::EvaluationFunctor(N, x, a, b)) {}
+
+  virtual ~EvaluationFactor() {}
+};
+
+/**
+ * Prior factor for BASIS vector evaluation on ParameterMatrix of size (M,N).
+ * @param BASIS: The basis class to use e.g. Chebyshev2
```
dellaert, line 72: Same comment. Also, please don’t use the word *prior* in this file. They are *unary* factors, but not priors.

gtsam/basis/BasisFactors.h
```diff
@@ -0,0 +1,243 @@
...
+            key, z, model,
+            typename BASIS::template VectorComponentFunctor<P>(N, i, x, a, b)) {
+  }
+
+  virtual ~VectorComponentFactor() {}
+};
+
+/**
+ * Prior factor for BASIS Manifold evaluation on type T.
+ * @param BASIS: The basis class to use e.g. Chebyshev2
```
dellaert, line 136: Explain. Example. Not prior.

gtsam/basis/BasisFactors.h
```diff
@@ -0,0 +1,243 @@
...
+      : Base(
+            key, z, model,
+            typename BASIS::template ManifoldEvaluationFunctor<T>(N, x, a, b)) {
+  }
+
+  virtual ~ManifoldEvaluationFactor() {}
+};
+
+/**
+ * Factor for scalar BASIS derivative evaluation.
```
dellaert, line 164: Explain.

gtsam/basis/Chebyshev.cpp
```diff
@@ -0,0 +1,97 @@
...
+ * ((b'-a') * (x - a) / (b - a)) + a'
+ *
+ * @param x Value to scale to new range.
+ * @param a Original lower limit.
+ * @param b Original upper limit.
+ * @param t1 New lower limit.
+ * @param t2 New upper limit.
+ * @return double
+ */
+double scale(double x, double a, double b, double t1, double t2) {
```
dellaert, line 35: static ?

gtsam/basis/Chebyshev.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  Ux(0) = 1;
+  Ux(1) = 2 * x;
+  for (size_t i = 2; i < N; i++) {
+    // instead of cos(i*acos(x)), this recurrence is much faster
+    Ux(i) = 2 * x * Ux(i - 1) - Ux(i - 2);
+  }
+  return Ux;
+}
+Weights Chebyshev2Basis::DerivativeWeights(size_t N, double x, double a,
```
dellaert, line 53: Add blank line

gtsam/basis/Chebyshev.cpp
```diff
@@ -0,0 +1,97 @@
...
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file Chebyshev.cpp
+ * @brief Chebyshev basis decompositions
+ * @author Varun Agrawal, Frank Dellaert
```
dellaert, line 15: Add Jing Dong to all code copied from DoubleExpresso

gtsam/basis/Chebyshev.h
```diff
@@ -0,0 +1,111 @@
...
+   * @param N Degree of the polynomial.
+   * @param x Point to evaluate polynomial at.
+   * @param a Lower limit of polynomial (default=-1).
+   * @param b Upper limit of polynomial (default=1).
+   * @return Weights
+   */
+  static Weights DerivativeWeights(size_t N, double x, double a = -1,
+                                   double b = 1);
+};
+// Chebyshev2Basis
```
dellaert, line 64: Formatting?

gtsam/basis/Chebyshev.h
```diff
@@ -0,0 +1,111 @@
...
+
+/**
```
dellaert, line 66: Switch so  first kind comes first in this file and .cpp?

gtsam/basis/Chebyshev2.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    for (size_t k = 1; k <= last_k; ++k)
+      weights(i) -= 2 * cos(2 * k * theta) / (4 * k * k - 1);
+    weights(i) *= (b - a) / K;
+  }
+
+  return weights;
+}
+
+}  // namespace gtsam
```
dellaert, line 205: add return

gtsam/basis/Chebyshev2.h
```diff
@@ -0,0 +1,137 @@
...
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file Chebyshev2.h
+ * @brief Chebyshev parameterizations on Chebyshev points of second kind.
```
dellaert, line 14: Propose replacing by

@brief Pseudo-spectral parameterization for Chebyshev polynomials of the second kind.

In a pseudo-spectral case, rather than the parameters acting as weights for the bases polynomials (as in Chebyshev2Basis), here the parameters are the *values* at a specific set of points in the interval, the “Chebyshev points”. These values uniquely determine the polynomial that interpolates them at the Chebyshev points.

gtsam/basis/Chebyshev2.h
```diff
@@ -0,0 +1,137 @@
...
+
+#include <boost/function.hpp>
+
+namespace gtsam {
+
+/**
+ * Chebyshev Interpolation on Chebyshev points of the second kind
+ * Note that N here, the #points, is one less than N from Trefethen00book
+ * (pg.42)
+ */
```
dellaert, line 37: Maybe add a reference here to book

gtsam/basis/FitBasis.h
```diff
@@ -0,0 +1,88 @@
...
+  static GaussianFactorGraph::shared_ptr LinearGraph(
+      const Sequence& sequence, const SharedNoiseModel& model, size_t N) {
+    NonlinearFactorGraph graph = NonlinearGraph(sequence, model, N);
+    Values values;
+    values.insert<Parameters>(0, Parameters::Zero(N));
+    GaussianFactorGraph::shared_ptr gfg = graph.linearize(values);
+    return gfg;
+  }
+
+  /// Constructor
```
dellaert, line 77: Add proper doxygen with parameters

gtsam/basis/FitBasis.h
```diff
@@ -0,0 +1,88 @@
...
+
+/// Our sequence representation is a map of {x: y} values where y = f(x)
+using Sequence = std::map<double, double>;
+/// A sample is a key-value pair from a sequence.
+using Sample = std::pair<double, double>;
+
+/**
+ * Class that does regression via least squares
+ * Example usage:
+ *  auto fit = FitBasis<Chebyshev2>(3, data_points, noise_model);
```
dellaert, line 43: 3 should be last argument

gtsam/basis/Fourier.h
dellaert: You are correct. I propose throwing an exception below as not implemented yet. If someone needs it, they can implement. And, @gchenfc , if you feel called, just add a Pr after this is landed. It needs new unit tests and scaling the derivative weights, I think. It could be *very simple* as I think it is just multiplying the derivative weights with a scalar.

gtsam/basis/ParameterMatrix.h
```diff
@@ -0,0 +1,215 @@
...
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file ParamaterMatrix.h
```
dellaert, line 13: Rethinking this. Do we really need this? maybe we should just use `Eigen::Matrix<double, M, -1>` and kill this file.

gtsam/basis/basis.i
dellaert: Do it :-)

dellaert (changes requested):
Getting there, but before adding to GTSAM we really need better docs.

Please reply to every comment (with Done or thumbs up for most) which makes it really easy to re-review.

gtsam/basis/basis.i
varunagrawal: @ProfFan can you please explain more?

I came up with an alternative, where in the wrapper definition, we use `std::map` directly rather than `gtsam::Sequence`, so Pybind11 can wrap to a dict easily.

gtsam/basis/basis.i
```diff
@@ -0,0 +1,145 @@
...
+};
+
+// TODO(gerry): Add `DerivativeFactor`, `VectorDerivativeFactor`, and
+// `ComponentDerivativeFactor`
+
+#include <gtsam/basis/FitBasis.h>
+template <BASIS = {gtsam::FourierBasis, gtsam::Chebyshev1Basis,
+                   gtsam::Chebyshev2Basis, gtsam::Chebyshev2}>
+class FitBasis {
+  FitBasis(size_t N, const std::map<double, double>& sequence,
```
varunagrawal, line 133: @gchenfc try this out! I hit upon this idea last night.

gtsam/basis/Fourier.h
varunagrawal: Who is correct? :sweat_smile:

gtsam/basis/ParameterMatrix.h
varunagrawal: I've already explained this in a previous comment. Pasting here verbatim :slightly_smiling_face: 

Doing 
```cpp
template <int M>
using ParameterMatrix = Eigen::Matrix<double, M, -1>
```
causes the API to become far more verbose and non-intuitive.

E.g., we can no longer do `ParameterMatrix<12> X(N)` since it gives an error that we tried calling a vector method on a matrix object. We have to instead do
```cpp
ParameterMatrix<12> X;
X.resize(12, N); // why do we need to specify 12 again here? This is something eigen needs :(
```

In general, the fact that one of the dimensions is dynamic and the other has size greater than 1 makes things non-standard for Eigen since it assumes  everything is a Matrix and simple things like `X.col` don't compile.

I'll add this to the docstring for future reference.

varunagrawal: Our other option is is to just use `Matrix<double, -1, -1>` but then we lose compile time checks on the matrix size.

dellaert: It should be very easy to make a small function that creates the right matrix. Could we replac entire file contents with code below?

```
template <int M>
using ParameterMatrix = Eigen::Matrix<double, M, -1>;

template <int M>
ParameterMatrix<M> MakeParameters(int  N) {
  ParameterMatrix<M> P;
  P.resize(M, N);
  return P;
}

```

and say 

```
auto P = MakeParameters<12>(100);
```

varunagrawal: That solves the initialization problem but not the various vector-matrix confusion errors that Eigen throws on compilation.

I did the whole typedef and actually using the matrix, and it didn't compile.

varunagrawal:
Wrapper stuff has been fixed. I have to address Frank's comments but I am swamped with Skydio and ICRA so pushing this back a bit.

I'll get to it eventually :)

dellaert:
@varunagrawal - you seem to be working on many PR's except this one :-) Let's land it after you address comments, people are waiting for it...

varunagrawal:
Yeah though those other PRs are related to the state estimator.

gtsam/basis/Chebyshev.cpp
varunagrawal: Done

gtsam/basis/FitBasis.h
varunagrawal: Done

varunagrawal: Wait really? Shouldn't the polynomial degree be the first argument so that it is consistent?

gtsam/basis/FitBasis.h
varunagrawal: Done

gtsam/basis/basis.i
varunagrawal: Done

gtsam/basis/Chebyshev.cpp
varunagrawal: Doing for all code except `BasisFactors.h` since I explicitly wrote that (when we switched away from Expressions).

gtsam/basis/Chebyshev.h
varunagrawal: Huh clang-format doesn't fix this automatically. If I move it up to the closing bracket, it then works.

gtsam/basis/Chebyshev2.h
varunagrawal: Isn't `Trefenthen00Book` the reference in that comment?

gtsam/basis/basis.i
varunagrawal: @gchenfc gentle reminder.

gtsam/basis/Fourier.h
varunagrawal: I added something but I am going to leave the rest of this to @gchenfc or whomever needs it in the future. Not on my A->B but will assist if someone raises an issue.

gtsam/basis/FitBasis.h
```diff
@@ -0,0 +1,94 @@
...
+
+namespace gtsam {
+
+/// Our sequence representation is a map of {x: y} values where y = f(x)
+using Sequence = std::map<double, double>;
+/// A sample is a key-value pair from a sequence.
+using Sample = std::pair<double, double>;
+
+/**
+ * Class that does regression via least squares
```
dellaert, line 41: explain what the arguments are in the example.

dellaert (changes requested):
Getting really close. A few more requested comments.

gtsam/basis/BasisFactors.h
varunagrawal: For this and the above comments, I believe we're on the same page, I've just been writing the docs from a more numerical analysis perspective(?). What I'm trying to convey is that if some oracle evaluated the unknown `f(x;p)` and gave us `z`, we're using these factors to ensure that when we evaluate the polynomial `q(x)` we get the same `z`. For the manifold case, this means we enforce the same type and value.

varunagrawal: `same value of T` should probably be `same value of type T`

dellaert: And I’m saying the documentation as you wrote it it is highly confusing and will make sure the readers of this code are **not** on the same page :-)

varunagrawal: Haha this is why we have you as our esteemed reviewer :D

PR #419, opened 2020-07-21
Values Print Formatting

varunagrawal:
Added some formatting to the values printing to make viewing it easier.


dellaert:
@varunagrawal i made a milestone for all 4.1.1 changes, and added this to it. 4.1 is *just* pybind wrapper *and* removing all 4.0 deprecated methods. 4.1.1 can follow relatively soon after that, or we can just leave things in develop for a while. I don’t want yo hold up the 4.q version bump with a whole lot of PRs. (did you change some back to 4.1??)

gtsam/base/GenericValue.h
```diff
@@ -84,7 +84,7 @@ class GenericValue: public Value {
 
   /// Virtual print function, uses traits
   virtual void print(const std::string& str) const {
-    std::cout << "(" << demangle(typeid(T).name()) << ") ";
+    std::cout << "(" << demangle(typeid(T).name()) << ") " << std::endl;
```
jlblancoc, line 87: "\n" is preferred to std::endl since it might be faster. endl implies a stream sync, AFAIK.

This applies to everywhere in this PR...

varunagrawal:
@dellaert sounds good. Since this is a print update, I wasn't sure if it should be in a minor version bump rather than a patch version bump.

varunagrawal:
@jlblancoc great catch. The `\n` makes total sense here; I was waiting on getting #358 merged in before finishing up this PR :smile:

gtsam/base/GenericValue.h
```diff
@@ -83,8 +83,8 @@ class GenericValue: public Value {
   }
 
   /// Virtual print function, uses traits
-  void print(const std::string& str) const override {
-    std::cout << "(" << demangle(typeid(T).name()) << ") ";
+  virtual void print(const std::string& str) const override {
```
dellaert, line 86: no virtual if override

PR #424, opened 2020-07-23
replace boost random with std random

acxz:
follow up in spirit to #233 
this pr removes all boost random

### DIFF
--- a/timing/timeMatrixOps.cpp
+++ b/timing/timeMatrixOps.cpp
@@ -34,7 +34,7 @@ using boost::format;
 using namespace boost::lambda;
 
 static std::mt19937 rng;
-static std::uniform_real_distribution<> rg(-1.0, 0.0);
+static std::uniform_real_distribution<> uniform(-1.0, 0.0);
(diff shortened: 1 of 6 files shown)


timing/timeMatrixOps.cpp
```diff
@@ -33,7 +33,8 @@ using namespace std;
 using boost::format;
 using namespace boost::lambda;
 
-static boost::variate_generator<boost::mt19937, boost::uniform_real<> > rng(boost::mt19937(), boost::uniform_real<>(-1.0, 0.0));
+static std::mt19937 rng;
+static std::uniform_real_distribution<> rg(-1.0, 0.0);
```
dellaert, line 37: maybe call this uniform?

acxz: Yeah was debating on change the name to be more appropriate.

acxz: changed `rg` -> `uniform`
`rni` -> `ri` -> `uniform_i`
`rnj` -> `rj` -> `uniform_j`

PR #425, opened 2020-07-23
remove unused base variable

acxz:
just a warning fix in the functorized factor


acxz:
np just helping out where I can

PR #427, opened 2020-07-24
Fix test for FunctorizedFactor printing

varunagrawal:
Needed to remove a single space.


varunagrawal:
This is a change needed for #358 so I'll fix it there.

PR #429, opened 2020-07-24
Travis timeout fixes

varunagrawal:
Travis has been timing out a lot recently due to the increase in unit tests (especially for DEBUG builds).

This PR increases the number of make cores from 2 to 3.
I also removed the `sudo` requirement since travis has deprecated that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: cpp
 cache: ccache
-sudo: required
 dist: xenial
 
 addons:
@@ -33,7 +32,7 @@ stages:
 
 env:
   global:
-    - MAKEFLAGS="-j2"
+    - MAKEFLAGS="-j3"
     - CCACHE_SLOPPINESS=pch_defines,time_macros
 
 # Compile stage without building examples/tests to populate the caches.


jlblancoc:
I'm afraid there are only 2 cores in Travis: https://docs.travis-ci.com/user/reference/overview/#virtualization-environments

So the -j3 has no actual effect...

dellaert:
Actually that might not be quite true, right? A thread might be blocked on a file resource allowing two other threads to use the cores fully. But proof is in pudding, and @varunagrawal  might want to look at the timings in the build to verify...

dellaert:
See https://stackoverflow.com/questions/15289250/make-j4-or-j8/15295032

jlblancoc:
That's true...

jlblancoc:
If you are interested, we could prepare a configuration file for CircleCI instead... from my experience, it's much faster to launch and build... ?

dellaert:
> If you are interested, we could prepare a configuration file for CircleCI instead... from my experience, it's much faster to launch and build... ?

Let's see how the Github actions work?

varunagrawal:
If we can move to GitHub Actions this weekend, I am completely fine with closing this PR.

Whatever makes me not have to manually restart CI tasks constantly.

dellaert:
Just merge it. I think it might help a bit

PR #434, opened 2020-07-26
Fix virtual/override (C++11) keywords and enforce future usage via compile errors

jlblancoc:
Use `virtual` only in base classes, `override` only in derived classes. 
Add flags so the compiler emits an error if these rules are not observed in the future.

This makes the code easier to understand to newcomers, and avoids by design some errors and hard-to-debug bugs like unintentionally declaring a method that one believes is overriding a virtual method but it's not (e.g. a missing "const" or an additional/missing parameter).

PR #435, opened 2020-07-26
Fix clang warnings

jlblancoc:
Follow up of #434


jlblancoc:
> Could that Clique virtual constructor also addressed in this PR, or is that unrelated?

I would say it's unrelated, but if we confirm it was new after the merge, I would investigate it too...

jlblancoc:
Anyway: I could reproduce your Clique warning with clang-12, will address it here too.

jlblancoc:
@dellaert Done! Take a look at the commit logs for details on changes... 
PS: I'll not merge myself this time ;-) 
Good night!

PR #436, opened 2020-07-27
fix clang warnings on tests

acxz:
follow up in spirit to #435 but running `make check` with clang


gtsam_unstable/discrete/tests/testScheduler.cpp
```diff
@@ -75,7 +75,7 @@ DiscreteFactorGraph createExpected() {
   // Mutual exclusion for students
   expected.addAllDiff(A, J);
 
-  return expected;
+  return std::move(expected);
```
jlblancoc, line 78: Ahh! We would have all these for free in C++17 😉
https://en.cppreference.com/w/cpp/language/copy_elision

PR #437, opened 2020-07-27
[DoNotMerge] Add variadic-template N-ary ExpressionFactor

jlblancoc:
I found this useful to have...

Should `ExpressionFactor2` remain as a special case, or could it be replaced by a:
```cpp
template <xxx> 
using ExpressionFactor2=ExpressionFactorN<...>;
```
?

Note that such a change would change the API slightly so it's not backward compatible, hence I didn't do it.

### DIFF
--- a/gtsam/base/GenericValue.h
+++ b/gtsam/base/GenericValue.h
@@ -198,4 +198,12 @@ const ValueType& Value::cast() const {
   return dynamic_cast<const GenericValue<ValueType>&>(*this).value();
 }
 
+/** Functional constructor of GenericValue<T> so T can be automatically deduced
+  */
+template<class T>
+GenericValue<T> genericValue(const T& v) {
+  return GenericValue<T>(v);
+}
+
+
 } /* namespace gtsam */
--- a/gtsam/nonlinear/ExpressionFactor.h
+++ b/gtsam/nonlinear/ExpressionFactor.h
@@ -34,6 +34,9 @@ namespace gtsam {
  * such instances, the user should declare derived classes from this template,
  * implementing expresion(), serialize(), clone(), print(), and defining the
  * corresponding `struct traits<NewFactor> : public Testable<NewFactor> {}`.
+ *
+ * \tparam T Type for measurements.
+ *
  */
 template<typename T>
 class ExpressionFactor: public NoiseModelFactor {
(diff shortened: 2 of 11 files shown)


dellaert:
Cool. Will leave @varunagrawal to take a deeper look. Taking @ProfFan off as he's super busy
You could deprecate ExpressionFactor2, maybe put in a stub that just calls new one?

jlblancoc:
My only doubts are on ExpressionFactor2. Options: 
a) Deprecate current ExpressionFactor2. Add a "using X=Y;" for ExpressionFactor{2,3,4...} as instances of the new generic one... but with which names? It cannot be the same, of course...
b) Just replace ExpressionFactor2 with the "using X=Y;" of the new template. This is backward incompatible only due to missing `evaluateError()`... how much important is that?
c) Just leave ExpressionFactor2 as is... perhaps deprecate it, recommending the use of the new generic type so future releases can clean up ExpressionFactor2 and save space.

dellaert:
> My only doubts are on ExpressionFactor2. Options:
> a) Deprecate current ExpressionFactor2. Add a "using X=Y;" for ExpressionFactor{2,3,4...} as instances of the new generic one... but with which names? It cannot be the same, of course...
> b) Just replace ExpressionFactor2 with the "using X=Y;" of the new template. This is backward incompatible only due to missing `evaluateError()`... how much important is that?
> c) Just leave ExpressionFactor2 as is... perhaps deprecate it, recommending the use of the new generic type so future releases can clean up ExpressionFactor2 and save space.

I like option b, but you can have cake and eat it , too, by deprecating

```
class ExpressionFactor2 : ExpressionFactorN {
evaluateError()
}
```

no?

jlblancoc:
Sure 👍

jlblancoc:
Done. 

@varunagrawal , @dellaert : Please, take a look at the last bunch of commits... the deprecation of `ExpressionFactor2` led to one thing, then to another... ;-)
I like things like the new `{ {k,v},{k,v},... }` initializer for `Values`, it's handy. Let me know your thoughts.

jlblancoc:
> @jlblancoc hate to say it but I think you might have to split this PR and add a test in testValues for the Values change :-)
> When you do, make sure to start from develop and/or merge in so you get (slightly) faster travis.

Ok! It was already indirectly tested in the modified tests, but it's better to have dedicated unit tests for the new feature of Values, you're right. Split into #442 and a new PR to come after #442 is merged.

PR #440, opened 2020-07-28
Minor updates to the pybind wrapper

varunagrawal:
1. Fixed the `python-install` make command.
2. Updated the `__init__.py` file to have a common numpy import.
3. Cleaned up the `setup.py` a bit.
4.  ~~Updated the cpp template to include the stl.h header so KeyVector works~~

### DIFF
--- a/python/gtsam_py/python/__init__.py
+++ b/python/gtsam_py/python/__init__.py
@@ -1,9 +1,11 @@
 from .gtsam import *
 
+import numpy as np
+
+
 def Point2(x=0, y=0):
-    import numpy as np
     return np.array([x, y], dtype=float)
 
+
 def Point3(x=0, y=0, z=0):
-    import numpy as np
     return np.array([x, y, z], dtype=float)
--- a/python/gtsam_py/python/__init__.py
+++ b/python/gtsam_py/python/__init__.py
@@ -1,11 +1,23 @@
 from .gtsam import *
 
-import numpy as np
+
+def _init():
+    import numpy as np
+
+    global Point2  # export function
+
+    def Point2(x=0, y=0):
+        return np.array([x, y], dtype=float)
+
+    global Point3  # export function
+
+    def Point3(x=0, y=0, z=0):
+        return np.array([x, y, z], dtype=float)
+
+    # for interactive debugging
+    if __name__ == "__main__":
+        # we want all definitions accessible
+        globals().update(locals())
 
 
-def Point2(x=0, y=0):
-    return np.array([x, y], dtype=float)
-
-
-def Point3(x=0, y=0, z=0):
-    return np.array([x, y, z], dtype=float)
+_init()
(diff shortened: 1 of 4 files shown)


ProfFan:
wait, I think KeyVector will not need this because it's been dealt with in stl_bind,

please refer to `python/gtsam_py/preamble.h` and `specializations.h`

ProfFan:
Previously: `keyvector.at(i)`

Now: `keyvector[i]`

dellaert:
@varunagrawal text @ProfFan to coordinate :-)

varunagrawal:
The issue is wrt TBB allocators. @ProfFan said he'll add a permafix for it tomorrow.

python/gtsam_py/python/__init__.py
```diff
@@ -1,9 +1,11 @@
 from .gtsam import *
 
+import numpy as np
```
ProfFan, line 3: I think it's better to not import numpy in __init__.py so we don't get `gtsam.np`?

varunagrawal: Ah that's a good point. Though I am confused why the PointX API has changed and not just handled via GTSAM?

ProfFan: Point2 and Point3 are now just numpy arrays, so (

varunagrawal: Huh that's strange. They were numpy arrays in the cython wrapper too, that's what eigency was for :-)

varunagrawal: Let me see what I can do about it

ProfFan: No they are not
```
import gtsam
p=gtsam.Point3(0,0,0)
>>> type(p)
gtsam.gtsam.Point3
>>> p.vector()
array([0., 0., 0.])
```

dellaert:
BTW. A style comment. We should always import classes if used more than once in a unit, and nothing else.

```
import gtsam
from gtsam import Pose2

x = Pose3()
y = Pose3()
z = gtsam.Rot3()

foo = gtsam.bar(x, y, z)

```

Maybe this can go in a doc somewhere.

python/gtsam_py/python/__init__.py
varunagrawal: Yeah the typedefing to Vectors is the key here. I have a fix for the original numpy issue though.

ProfFan: So I think this change can be reversed? After that we can merge the setup.py changes :)

varunagrawal:
@ProfFan I added a nicer way that doesn't involve multiple import statements. This was important because potentially every time someone calls `Point2`, the numpy module is reimported, which is unnecessary overhead.

ProfFan:
@varunagrawal Have you tested this `_init()` method? If so then good2go

varunagrawal:
@ProfFan I did test it. I don't push till I test everything locally. :grin:

PR #442, opened 2020-07-28
Add C++11 brace initializer_list to Values

jlblancoc:
Add: 
- Constructor from `initializer_list`
- `genericValue()` helper macro for the new constructor.
- Unit tests.

### DIFF
--- a/gtsam/nonlinear/tests/testValues.cpp
+++ b/gtsam/nonlinear/tests/testValues.cpp
@@ -607,5 +607,35 @@ TEST(Values, Demangle) {
 }
 
 /* ************************************************************************* */
+TEST(Values, brace_initializer) {
+  const Pose2 poseA(1.0, 2.0, 0.3), poseC(.0, .0, .0);
+  const Pose3 poseB(Pose2(0.1, 0.2, 0.3));
+
+  {
+    Values values;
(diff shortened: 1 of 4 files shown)


gtsam/nonlinear/tests/testValues.cpp
```diff
@@ -606,6 +606,36 @@ TEST(Values, Demangle) {
...
+
+  {
+    Values values;
+    EXPECT_LONGS_EQUAL(0, values.size());
+    values = { {key1, genericValue(1.0)} };
+    EXPECT_LONGS_EQUAL(1, values.size());
+    CHECK(values.at<double>(key1) == 1.0);
+  }
+  {
+    Values values = { {key1, genericValue(poseA)}, {key2, genericValue(poseB)} };
```
dellaert, line 622: I’m assuming there is no magic that gets rid of the `genericValue` :-) ?

jlblancoc: I tried hard with all my magic... not powerful enough! :-)

jlblancoc:
> Very cool. Maybe someone needs to take a look at the whole Values thing, with modern c++ eyes

When time comes for requiring C++17 to users, a good option would be `std::variant<>` of common types (so they get allocated on the stack if possible) + std::any for any user-defined, less common types.

PR #444, opened 2020-07-28
Add N-ary template expression factors

### LINKED ISSUE #443: Deserialized expression factors (RangeFactor, may be others?) evaluate the wrong expression
## Description

Expression factors use the virtual method `expression()` to build the actual gtsam::Expression that evaluates the error and Jacobians. 

During de-serialization, the base class ExpressionFactor is called first, and it calls expression() in its "read()" step, invoking the derived class specific code to build the factor expression. However, all the required parameters are not yet loaded, since that will happen in the the de-serialization method one call stack up (!).

The bug was hidden, can be revealed via this new unit test: 

https://github.com/borglab/gtsam/blob/6223be5df67e19d83db46475d1b540e7c5036f8e/gtsam/sam/tests/testRangeFactor.cpp#L149-L168

which fails in the second CHECK(). 

## Solution

A PR fixing this will come (after #442 is merged), as part 2/2 of #437. 
Fix is this simple: 

https://github.com/borglab/gtsam/blob/053d293a86e2e7d07901d30b52fa75fd80c7c74e/gtsam/sam/RangeFactor.h#L136-L144

Please, check if there are other ExpressionFactor's out there with parameters in the need to get fixed too.

PS: this is the result of a few hours of frustrating, mystifying errors :-)

jlblancoc:
* Add variadic-template N-ary ExpressionFactor
* Deprecate ExpressionFactor2
* More unit tests for RangeFactor
* Closes #443

### DIFF
--- a/gtsam/nonlinear/ExpressionFactor.h
+++ b/gtsam/nonlinear/ExpressionFactor.h
@@ -34,6 +34,9 @@ namespace gtsam {
  * such instances, the user should declare derived classes from this template,
  * implementing expresion(), serialize(), clone(), print(), and defining the
  * corresponding `struct traits<NewFactor> : public Testable<NewFactor> {}`.
+ *
+ * \tparam T Type for measurements.
+ *
  */
 template<typename T>
 class ExpressionFactor: public NoiseModelFactor {
@@ -279,6 +282,9 @@ class ExpressionFactor2 : public ExpressionFactor<T> {
         "ExpressionFactor", boost::serialization::base_object<ExpressionFactor<T> >(*this));
   }
 };
+/// traits
+template <typename T, typename A1, typename A2>
+struct traits<ExpressionFactor2<T,A1,A2>> : public Testable<ExpressionFactor2<T,A1,A2>> {};
 // ExpressionFactor2
 
 }// \ namespace gtsam
--- a/gtsam/nonlinear/ExpressionFactor.h
+++ b/gtsam/nonlinear/ExpressionFactor.h
@@ -295,7 +295,7 @@ struct traits<ExpressionFactorN<T, Args...>>
 // ExpressionFactorN
 
 
-#if defined(GTSAM_ALLOW_DEPRECATED_SINCE_V4)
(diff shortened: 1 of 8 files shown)


gtsam/nonlinear/ExpressionFactor.h
```diff
@@ -222,26 +227,98 @@ class ExpressionFactor: public NoiseModelFactor {
...
+  }
+};
+/// traits
+template <typename T, typename... Args>
+struct traits<ExpressionFactorN<T, Args...>>
+    : public Testable<ExpressionFactorN<T, Args...>> {};
+// ExpressionFactorN
+
+
+#if defined(GTSAM_ALLOW_DEPRECATED_SINCE_V4)
```
dellaert, line 298: V41

gtsam/sam/tests/testRangeFactor.cpp
```diff
@@ -47,25 +47,29 @@ double measurement(10.0);
 /* ************************************************************************* */
 Vector factorError2D(const Pose2& pose, const Point2& point,
     const RangeFactor2D& factor) {
-  return factor.evaluateError(pose, point);
+  const auto &keys = factor.keys();
+  return factor.unwhitenedError({{keys[0], genericValue(pose)}, {keys[1], genericValue(point)}});
```
dellaert, line 51: Hmmm. Since in the factor we know all types, is there not a "variadic" way to make an overload that justs accept the pose and point?

jlblancoc:
Is deprecated V41 already defined or should I add it? (Couldn't find it
yesterday...)

dellaert:
> Is deprecated V41 already defined or should I add it? (Couldn't find it yesterday...)

It's in a different PR, waiting on @step13489mcc

gtsam/nonlinear/ExpressionFactor.h
jlblancoc: done.

gtsam/sam/tests/testRangeFactor.cpp
jlblancoc: :+1: Much easier to use. Done too in 0ee5fc58f

PR #445, opened 2020-07-28
remove empty smart folder

acxz:
I was wondering if we can remove the `gtsam/smart/` folder? It's been empty for 6 years.

### DIFF
--- a/gtsam/CMakeLists.txt
+++ b/gtsam/CMakeLists.txt
@@ -14,7 +14,6 @@ set (gtsam_subdirs
     sam
     sfm
     slam
-    smart
 	  navigation
 )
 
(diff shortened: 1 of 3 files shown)


acxz:
CI timed out. Reopening PR.

varunagrawal:
We need to add some way to have external users re-run the CI without having to reopen the PR :/

acxz:
The CI has timed out three times already on the TBB test build run, I don't think restarting the CI will work for the fourth time.

> We need to add some way to have external users re-run the CI without having to reopen the PR :/

While that would be nice, I really don't think there is a way to do that. Haven't seen a solution in other repos I have PR'd to either. Plus, the CI should be in control of the main devs.

varunagrawal:
@acxz we're planning on moving to Github Actions since they are faster and more parallelizable. I wouldn't really worry about the Travis CI at the moment.

dellaert:
@ProfFan whats stopping the GitHub actions switch?

ProfFan:
@dellaert Nothing, so should have no problem once we merge the pybind stuff.

dellaert:
Rather than closing and re-opening , mention me and I can restart the steps that fail. It’s always the same

acxz:
@dellaert looks like it has timed out once more.

PR #449, opened 2020-07-30
Allow factors to be disabled manually

jlblancoc:
`NonLinearFactor::active()` always returned true by default. This PR allows this to be overriden via a new `Factor::active(bool)`. 

Why the implementation has been moved down to Factor instead of NonLinearFactor since?
- Factor already had a serializable state, whose version has been bumped. Introducing serialize in NonLinearFactor would have implied way more code changes in derived classes.
- Conceptually, it's a generic status that applied to plain linear factor (doesn't need the "Values").

See new unit tests for usage example.

### DIFF
--- a/gtsam/inference/Factor.h
+++ b/gtsam/inference/Factor.h
@@ -71,10 +71,13 @@ typedef FastSet<FactorIndex> FactorIndexSet;
     /// The keys involved in this factor
     KeyVector keys_;
 
+    /// Whether this factor is active (default=true) and should be included in optimizations.
+    bool active_ = true;
+
     /// @name Standard Constructors
     /// @{
 
-    /** Default constructor for I/O */
+    /** Default constructor for serialization */
     Factor() {}
 
     /** Construct factor from container of keys.  This constructor is used internally from derived factor
@@ -128,6 +131,13 @@ typedef FastSet<FactorIndex> FactorIndexSet;
     */
     size_t size() const { return keys_.size(); }
 
+    /// See NonLinearFactor::active().
+    bool active() const { return active_; }
+
+    /// Changes the default "active" status of this factor.
+    void active(bool newActiveValue) { active_ = newActiveValue; }
+
+
     /// @}
 
 
@@ -163,8 +173,9 @@ typedef FastSet<FactorIndex> FactorIndexSet;
     /** Serialization function */
     friend class boost::serialization::access;
     template<class Archive>
-    void serialize(Archive & ar, const unsigned int /*version*/) {
+    void serialize(Archive & ar, const unsigned int version) {
       ar & BOOST_SERIALIZATION_NVP(keys_);
+      if (version>=1) ar & BOOST_SERIALIZATION_NVP(active_);
     }
 
     /// @}
@@ -172,3 +183,5 @@ typedef FastSet<FactorIndex> FactorIndexSet;
   };
 
 }
+
+BOOST_CLASS_VERSION(gtsam::Factor, 1)
--- a/gtsam/nonlinear/NonlinearFactor.h
+++ b/gtsam/nonlinear/NonlinearFactor.h
@@ -102,15 +102,19 @@ public:
 
   /**
    * Checks whether a factor should be used based on a set of values.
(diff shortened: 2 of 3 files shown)


gtsam/inference/Factor.h
```diff
@@ -163,12 +173,15 @@ typedef FastSet<FactorIndex> FactorIndexSet;
     /** Serialization function */
     friend class boost::serialization::access;
     template<class Archive>
-    void serialize(Archive & ar, const unsigned int /*version*/) {
+    void serialize(Archive & ar, const unsigned int version) {
       ar & BOOST_SERIALIZATION_NVP(keys_);
+      if (version>=1) ar & BOOST_SERIALIZATION_NVP(active_);
```
ProfFan, line 178: Probably need to check other places for latest version number?

jlblancoc: I think "version" is local to each class, so it's only required to update it for the Factor base class... 
https://theboostcpplibraries.com/boost.serialization-archive

gtsam/inference/Factor.h
```diff
@@ -128,6 +131,13 @@ typedef FastSet<FactorIndex> FactorIndexSet;
     */
     size_t size() const { return keys_.size(); }
 
+    /// See NonLinearFactor::active().
+    bool active() const { return active_; }
+
+    /// Changes the default "active" status of this factor.
+    void active(bool newActiveValue) { active_ = newActiveValue; }
```
dellaert, line 138: I know some code bases like this style, but we have not done this in GTSAM. Prefer `setActive` or have `activate` and `deactivate` methods. That being said, I don't like imperative change, at all.

jlblancoc:
Ok, got it...

Just if there is a short answer: what would be a more desirable alternative? Write new factors with a boolean ("switch") value attached, or alike?

dellaert:
> Ok, got it...
> 
> Just if there is a short answer: what would be a more desirable alternative? Write new factors with a boolean ("switch") value attached, or alike?

Depends: what's the use case? What is the size of the graphs associated with the use case? You can filter and create a new graph: it's just a vector of shared pointers. If it's about outlier rejection, perhaps robust error metrics is the better answer. It's a conversation :-)

mikesheffler:
>what would be a more desirable alternative? Write new factors with a boolean ("switch") value attached, or alike?

It's kind of old now, but are you familiar with Suenderhauf and Protzel's Switchable Constraints? https://nikosuenderhauf.github.io/projects/switchableConstraints/

There's another group https://github.com/wvu-navLab/ICE using GTSAM to do robust estimation (they got going right before robust estimation was introduced in GTSAM proper) and were looking at techniques to basically make a constraint meaningless (no information) instead of formally shutting it off (or, at least, that's the way I read it).
 
I'll admit, I've pined for a kill switch on factors in the past. Especially when not using smart factors, large vision problems can be _brutal_ to reestablish through filtering of the graph. Maybe not worth changing the whole programming paradigm over, though 😜

dellaert:
I am merely arguing above that it needs a very good justification. If there is a clear use case I'm not opposed to it, we just have to carefully weigh pros and cons.

jlblancoc:
> It's kind of old now, but are you familiar with Suenderhauf and Protzel's Switchable Constraints? https://nikosuenderhauf.github.io/projects/switchableConstraints/

@mikesheffler Thanks for the references! Yes, I saw *one* of those papers years ago, but never went down to their implementation, honestly, neither was aware there were so many papers in that line.

@dellaert Thanks for the pointers! I guess I'm more an old-school (or chaotic?) imperative+OOP programmer; but I found more and more talented people over the years arguing for functional programming, so I guess it must have clear advantages :-) 

In my case, I wanted to check the optimization result manually enabling/disabling a few (<100) factors in a graph of a few 1000s factors. Of course, there are always alternatives: e.g. to reset the smart pointers of factors to be ignored, keeping a copy somewhere to restore it back based on its index. Not hard. 
So, if the programming paradigm is an important design criteria, this PR doesn't have a "very good justification" in the sense that *there are* alternatives...

PR #452, opened 2020-08-01
Format gtsam/base - no code change

varunagrawal:
Run a full format on all the files in `gtsam/base` using Googe style formatting.

I plan to create multiple PRs to make it a transition.


jlblancoc:
Would you be willing to also add a `.clang-format` file at the root directory specifying the desired format, so all sort of IDEs automatically use it to enforce format complying?

```
clang-format --dump-config -style=Google  > .clang-format
```

It would be super useful for those of us using QtCreator, Atom, VSCode, etc. Clang-format plugins exist for most editors, actually.

Once the transition is finished, you could even add an additional step in CI ([example](https://github.com/MRPT/mrpt/blob/develop/travis/check_style.sh)) to test that future PRs fulfill the .clang-format style (Google's, or any custom style).

dellaert:
No mass formatting.

varunagrawal:
Understood

jlblancoc:
> No mass formatting.

Sure, it creates a reset point in git history. 
If you still want to add the .clang-format file, it could be used to "apply format to section", whenever a particular part of the code is modified. Just an idea...

dellaert:
> > No mass formatting.
> 
> Sure, it creates a reset point in git history.
> If you still want to add the .clang-format file, it could be used to "apply format to section", whenever a particular part of the code is modified. Just an idea...

If that is possible, I like that! Maybe @varunagrawal can test it out?

jlblancoc:
> If that is possible, I like that! Maybe @varunagrawal can test it out?

I think it would be doable with the command above (+ any manual tweak to the format per your preferences if different than Google's style), then reformatting just a section is possible from Visual Studio or QtCreator IDEs, at least (tested personally).

PR #453, opened 2020-08-01
Better frobenius factors

dellaert:
Previously, the vectorized Lie algebra generators were re-computed for *every* factor. This change allows to pass them in using a shared pointer, and hence compute them only once.

Also adds Testable for Frobenius factors, and cleaned up a little bit of rot while doing this.

### DIFF
--- a/timing/timeFrobeniusFactor.cpp
+++ b/timing/timeFrobeniusFactor.cpp
@@ -13,12 +13,11 @@
  * @file    timeFrobeniusFactor.cpp
  * @brief   time FrobeniusFactor with BAL file
  * @author  Frank Dellaert
- * @date    June 6, 2015
+ * @date    2019
  */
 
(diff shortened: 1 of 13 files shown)


timing/timeFrobeniusFactor.cpp
```diff
@@ -66,15 +62,16 @@ int main(int argc, char* argv[]) {
 
   // Build graph
   NonlinearFactorGraph graph;
-  // graph.add(NonlinearEquality<SO4>(0, SO4()));
+  // graph.add(NonlinearEquality<SOn>(0, SOn::identity(4)));
```
ProfFan, line 65: commented out code

dellaert: Yeah - not fixing that now. Was there before and orthogonal to PR

PR #459, opened 2020-08-06
Fix pose2 align bug.

Alexma3312:
In pose2 align function, the code does not align with the formula:
    cos ~ sum(dpx*dqx + dpy*dqy) and sin ~ sum(-dpy*dqx + dpx*dqy)

This is clearly due to an incorrect variable declaration as **dp** should generate from **cp** and **dq** should generate from **cq**. To correct the code, I just switch the **dp** and **dq** variable.


ProfFan:
@Alexma3312 If this is not covered by a unit test, please add one?

Alexma3312:
> @Alexma3312 If this is not covered by a unit test, please add one?

I did not add a new unit test because there is a unit test in testPose2.cpp to test on this align function.
After this modification, the unit test can still pass.

PR #463, opened 2020-08-08
Update translation averaging to use vector of binary measurements instead of map

akshay-krishnan:
This PR updates the input to translation averaging from a map of KeyPair to Unit3 to a vector of binary measurements of Unit3. In doing so, binary measurement is also introduced and wrapped in gtsam.h

Note: TranslationRecovery is not wrapped yet. This requires us to wrap vector<BinaryMeasurement<Unit3>> and will be addressed in a separate PR.

### DIFF
--- a/gtsam/sfm/BinaryMeasurement.h
+++ b/gtsam/sfm/BinaryMeasurement.h
@@ -18,6 +18,13 @@
  * @brief Binary measurement for representing edges in the epipolar graph
  */
 
+#include <ostream>
+
+
+#include <gtsam/base/Testable.h>
+#include <gtsam/base/Lie.h>
+#include <gtsam/nonlinear/NonlinearFactor.h>
+
 #include <gtsam/inference/Key.h>
 #include <gtsam/linear/NoiseModel.h>
 
@@ -71,22 +78,17 @@ public:
     }
 
     /** equals */
-    bool equals(const BetweenMeasurement& expected, double tol=1e-9) const {
-      const BetweenMeasurement<VALUE> *e =  dynamic_cast<const  BetweenMeasurement<VALUE>*> (&expected);
-      return e != nullptr && key1_ == expected.key1() && 
-             key2_ == expected.key2() 
+    bool equals(const BinaryMeasurement& expected, double tol=1e-9) const {
+      const BinaryMeasurement<VALUE> *e =  dynamic_cast<const  BinaryMeasurement<VALUE>*> (&expected);
+      return e != nullptr && key1_ == e->key1_ && 
+             key2_ == e->key2_
              && traits<VALUE>::Equals(this->measured_, e->measured_, tol) && 
-             noiseModel_.equals(expected.noiseModel());
+             noiseModel_->equals(expected.noiseModel());
     }
 
     /** return the measured */
-    const VALUE& measured() const {
+    VALUE measured() const {
       return measured_;
-    }
-
-    /** number of variables attached to this measurement */
-    std::size_t size() const {
-      return 2;
     }
   }; // \class BetweenMeasurement
 }
--- a/gtsam/sfm/BinaryMeasurement.h
+++ b/gtsam/sfm/BinaryMeasurement.h
@@ -1,0 +1,92 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010-2020, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+#pragma once
+
+/**
+ * @file BinaryMeasurement.h
+ * @author Akshay Krishnan
+ * @date July 2020
+ * @brief Binary measurement for representing edges in the epipolar graph
+ */
+
+#include <gtsam/inference/Key.h>
+#include <gtsam/linear/NoiseModel.h>
+
+namespace gtsam {
+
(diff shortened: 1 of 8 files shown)


gtsam/sfm/BinaryMeasurement.h
```diff
@@ -0,0 +1,92 @@
...
+
+ * -------------------------------------------------------------------------- */
+
+#pragma once
+
+/**
+ * @file BinaryMeasurement.h
+ * @author Akshay Krishnan
+ * @date July 2020
+ * @brief Binary measurement for representing edges in the epipolar graph.
```
dellaert, line 18: Not necessarily an epipolar graph...

akshay-krishnan:
Currently, the translation averaging code expects a map from KeyPair to Unit3 for input unit translations. The problem with this was that it was difficult to wrap a map, and it is also unnecessary and more computationally expensive. So we decided to switch to a vector of Unit3 measurements. We could not use a BetweenFactor or a NoiseModelFactor2 because these factors involve error functions that are different from the chordal distance used in translation averaging. The binary measurement class does not possess an error method and contains just the right amount of information to characterize rotations and unit translations between cameras in SfM.

gtsam/sfm/BinaryMeasurement.h
akshay-krishnan: Yeah, I updated the comment.

tests/testTranslationRecovery.cpp
```diff
@@ -69,10 +71,14 @@ TEST(TranslationRecovery, BAL) {
 
   // Check result for first two translations, determined by prior
   EXPECT(assert_equal(Point3(0, 0, 0), result.at<Point3>(0)));
-  EXPECT(assert_equal(Point3(2 * w_aZb.point3()), result.at<Point3>(1)));
+  EXPECT(assert_equal(Point3(2 * w_aZb_stored.point3()), result.at<Point3>(1)));
```
swarrier246, line 74: why this multiplication by 2?

akshay-krishnan: Because scale = 2.0. The result is expected to be the point at a distance 2 in the direction along w_aZb

tests/testTranslationRecovery.cpp
```diff
@@ -49,15 +49,17 @@ TEST(TranslationRecovery, BAL) {
...
-  const Pose3 wTa = poses.at<Pose3>(0), wTb = poses.at<Pose3>(1),
-              wTc = poses.at<Pose3>(2);
-  const Point3 Ta = wTa.translation(), Tb = wTb.translation(),
-               Tc = wTc.translation();
-  const Rot3 aRw = wTa.rotation().inverse();
-  const Unit3 w_aZb = relativeTranslations.at({0, 1});
-  EXPECT(assert_equal(Unit3(Tb - Ta), w_aZb));
-  const Unit3 w_aZc = relativeTranslations.at({0, 2});
-  EXPECT(assert_equal(Unit3(Tc - Ta), w_aZc));
+  Unit3 w_aZb_stored; // measurement between 0 and 1 stored for next unit test
```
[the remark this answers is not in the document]
akshay-krishnan: Because we need to use it in line 74. 
The old code stores both w_aZb and w_aZc. Ive renamed it to w_aZb_stored because I use w_aZb in the for loop for something else. I agree its not a great name. 
Storing from a map is easy because we can always look up using "at". But since it is a vector, we need to iterate over the edges to look up. I'm doing the iteration in lines 53-62 for another unit test, and at the same time storing this values.

akshay-krishnan:
@swarrier246 Yes, by definition the Binary Measurement class contains the measurement between two keys. The noise model is a property of the measurement and is therefore included. Error is excluded from this definition because if we include it becomes more than just a measurement - a factor. 
Another question you could ask is why we don't wrap the factor itself - why use measurement at all? We could, but the error function is a property of the optimizer and an external pipeline that is simply using the optimizer should not have to worry about it. If at a later point, we decide to use a different averaging algorithm, the interface will be the same. Another advantage of the measurement is that we can use the same for rotations (because its templated), although rotation averaging uses a different error function. 
I hope that makes sense.

ProfFan:
Just to chime in, I think the name `BinaryMeasurement` is a bit misleading...

dellaert:
To be clear: Factor.h, the absolute base class :-) Take a look at SymbolicFactor to see a derived class.

dellaert:
Wondering even whether it even should simply be a MeasurementFactor, with arbitrary arity...

akshay-krishnan:
Thank you for suggesting the Factor base class Prof. Frank, it looks cool! 
I am open to making this change. However, I have the following doubts: 

- We can create a class MeasurementFactor derived from Factor. This will eliminate the methods to access the keys, but the methods for measurement, noise model, printing, and equals (which is the major portion of the code) will still remain. 
- This is optional, but it is useful to have a specialized version of MeasurementFactor that only has two keys for use in SfM. Methods like key1() and key2() are more intuitive than having to use front() and back(). If we do that, we end up with an additional class and more code.
 
I also don't know how it can be used in a factor graph without an error function, but if that is not a major advantage, then could you please let me know what the benefit of the change would be?

dellaert:
More I think about, and given the time pressure, let's just land as is. I think we should do another PR in which it derives from Factor, at least.

akshay-krishnan:
Unable to merge because Travis is timing out for one case :(
@ProfFan would you know how to solve this?

dellaert:
I just restarted the one missing job manually. This is *very* annoying

jlblancoc:
> I just restarted the one missing job manually. This is _very_ annoying

In what state is Github Actions CI integration? Is there anything blocking it?

PR #465, opened 2020-08-11
Added Spectra 0.9.0 to 3rdparty

dellaert:
Adding Spectra Eigenvalue solver for Shonan Averaging certificate

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -23,3 +23,5 @@ ordering library
     - Included unmodified in gtsam/3rdparty/metis
     - Licenced under Apache License v 2.0, provided in
       gtsam/3rdparty/metis/LICENSE.txt
+- Spectra v0.9.0: Sparse Eigenvalue Computation Toolkit as a Redesigned ARPACK.
+    - Licenced under MPL2, provided at https://github.com/yixuan/spectra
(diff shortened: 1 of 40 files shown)


varunagrawal:
Just a suggestion (no need to accept it :) ), would git submodules be a better way to add 3rd party modules? You get finer control of the version included and it is very easy to update the version via git.

jlblancoc:
I will delay reviewing the PR until Frank gives his opinion on using submodules. Indeed, it's a much cleaner way to keep dependencies around. The main caveats are: (1) having to update the git clone instructions; and (2) dealing with users complaining if they didn't read the instructions... :-)

PS: @dellaert congrats of this awesome paper! ;-)

dellaert:
Thanks @jlblancoc 
We can start doing this in 4.1 but I'd like to postpone and release Shonan code in 4.0.4

dellaert:
main caveats are: (1) having to update the git clone instructions; and (2) dealing with users complaining if they didn't read the instructions... :-)

Indeed, 2 is a big issue

ProfFan:
@dellaert  I have started to feel that too, although I introduced the wrap submodule in the pybind branch. I now think probably a monorepo is a better choice for users...

PR #469, opened 2020-08-13
Replace Travis CI and Appveyor with Github Actions

varunagrawal:


ProfFan:
@varunagrawal I thought we are cherry-picking the commit from `feature/new_wrapper`? I think I can handle it today.

varunagrawal:
@ProfFan yeah there were a looooot of commits to go through and this seemed faster.

ProfFan:
@varunagrawal I propose that I take over, the Python build in this PR is for the pybind wrapper.

varunagrawal:
I updated the scripts and the yml files to work with the Cython wrapper. So far everything seems to be working.

It honestly doesn't matter to me which of us works on this as long as we land this quickly. :)

dellaert:
@ProfFan is right - better coordination here is needed to avoid double work :-(

varunagrawal:
Alternatively, @ProfFan can merge off the latest develop and update this (some minor modifications, specifically only PyBind11 related).

I have added a bunch of fixes to ensure the checks work correctly (which the Pybind PR doesn't have).

dellaert:
> Alternatively, @ProfFan can merge off the latest develop and update this (some minor modifications, specifically only PyBind11 related).
> 
> I have added a bunch of fixes to ensure the checks work correctly (which the Pybind PR doesn't have).

@varunagrawal  The issue seems to be that you guys did not coordinate (and your message does not indicate you are coordinating now) so the changes are somehow incompatible. Let's stop having the conversation here and connect with @ProfFan directly to decide how to do it.

varunagrawal:
Deffered to #472

PR #470, opened 2020-08-14
Fix FrobeniusBetweenFactor declaration

varunagrawal:
This is the reason the Windows build is breaking on Appveyor.


ProfFan:
Nice!

varunagrawal:
@dellaert can we force merge this? Travis is constantly timing out. :cry:

dellaert:
> @dellaert can we force merge this? Travis is constantly timing out. 😢

How about commenting out in this PR the (2?) builds in travis that time out? I think it's always gcc+Debug. Then this will benefit all future PR's until we *finally* switch over to GH Actions.

PR #473, opened 2020-08-16
Feature/shonan averaging

dellaert:
This is *the* PR for Shonan Averaging, which has been in development for over a year. It's pretty been pretty well tested and benchmarked, but give it a read, please ?

Note this PR targets the dataset refactor. Once that merges I will change target to develop.


dellaert:
Note, I will Squash-merge after it's approved!

gtsam.h
```diff
@@ -2895,6 +2897,133 @@ class BinaryMeasurement {
 typedef gtsam::BinaryMeasurement<gtsam::Unit3> BinaryMeasurementUnit3;
 typedef gtsam::BinaryMeasurement<gtsam::Rot3> BinaryMeasurementRot3;
 
+#include <gtsam/sfm/ShonanAveraging.h>
```
ProfFan, line 2900: A lot of stuff got exported, but no corresponding Python tests, so no way to know if they are correct?

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -0,0 +1,861 @@
...
+  auto lambdas = solver.eigenvalues();
+  double minEigenValue = lambdas(0).real();
+  for (size_t i = 1; i < lambdas.size(); i++) {
+    minEigenValue = min(lambdas(i).real(), minEigenValue);
+  }
+#else
+  double minEigenValue;
+  bool success = SparseMinimumEigenValue(A, S, &minEigenValue, minEigenVector);
+  if (!success) {
+    throw std::runtime_error(
```
ProfFan, line 647: Probably should fall back to the slow version instead of bailing out?

gtsam.h
dellaert: There is an entire python repo with tests. I will add that in a different PR

gtsam/sfm/ShonanAveraging.cpp
dellaert: That is *extremely* slow, and for large problems will sink your machine :-) I removed the slow code and moved it to the tests.

gtsam.h
dellaert: I did check wrapper is correctly built ;-)

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -0,0 +1,841 @@
...
+  Values lifted = LiftTo<SOn>(p, values);
+  for (auto it : lifted.filter<SOn>()) {
+    // Create a tangent direction xi with eigenvector segment v_i
+    // Assumes key is 0-based integer
+    const Vector xi = MakeATangentVector(p, minEigenVector, it.key);
+    // Move the old value in the descent direction
+    it.value = it.value.retract(xi);
+  }
+  return lifted;
+}
```
dellaert, line 701: @david-m-rosen the retract here is the Cayley map (exponential map is not implemented beyond SO(4))

gtsam/sfm/tests/testShonanAveraging.cpp
```diff
@@ -0,0 +1,367 @@
...
+  Values initialQ4 =
+      ShonanAveraging3::LiftwithDescent(4, Qstar3, minEigenVector);
+  EXPECT_LONGS_EQUAL(3, initialQ4.size());
+  Matrix expected(4, 4);
+  expected << 0.88891, 0.0257057, -0.0639449, -0.452868,  //
+      0.0257057, 0.994052, 0.0147965, 0.104791,           //
+      -0.0639455, 0.0147967, 0.963192, -0.260678,         //
+      0.452868, -0.104791, 0.260678, 0.846154;
+  EXPECT(assert_equal(SOn(expected), initialQ4.at<SOn>(0), 1e-5));
+}
```
dellaert, line 196: @david-m-rosen here is the descent check that fails

david-m-rosen: OK got it. Let me put some thought behind this then -- it should be possible to fix this up without requiring that we use any particular eigenvector ...

david-m-rosen: OK, if this case is meant to be a unit test **just** for the LiftwithDescent method, then it can be radically simplified.  As currently implemented, LiftwithDescent just performs the lift of the input indicated in eq. (27) of the paper, constructs the tangent vector from eq. (28), and then performs retraction -- that is, this function doesn't actually care whether the input v is a minimum eigenvector or not.  

I would suggest just replacing this with a simple hard-coded example involving an initial set of 2 or 3 rotations, a corresponding fixed vector v, and then just verifying that the returned result is what it should be.  This would entirely avoid the issue of having to compute **specific** eigenvectors across multiple platforms

dellaert: That sounds like a plan :-)

PR #476, opened 2020-08-17
Improve Github Actions

varunagrawal:
Currently, Github Actions is very verbose, causing quite a bit of noise and making debugging hard.
This is a result of experimentation in order to get it working correctly.

This PR addresses these issues by making Actions as terse as possible.


dellaert:
Let me know when it’s ready for review. I see some builds, especially Mac and windows, sometimes cancel. For now I only made one of the builds a requirement. Btw if some of the builds can be combined (why are python builds separate?) it would save some CO2...

.github/workflows/build-linux.yml
```diff
@@ -1,6 +1,6 @@
 name: Linux CI
 
-on: [push, pull_request]
+on: [pull_request]
```
ProfFan, line 3: This is intentional (to have a small check for pushes, and only do mac/etc on PRs)

dellaert: I like that, too. Also, what happens if I push to an existing PR?

ProfFan: Nothing? if develop is not merged?

If you are talking about an existing PR after this, yes these actions will be run twice (but in a good way because one will be the PR build and another the BRANCH build) just like before with Travis.

varunagrawal: Gotcha. It seemed redundant to have a check for pushed commits that aren't in a PR, but the reasoning is valid.

ProfFan:
@varunagrawal I think we should probably merge this PR as is, we can do the improvements later.

varunagrawal:
@ProfFan I agree. Updated the PR and this is ready to go.

PR #480, opened 2020-08-19
Fix/shonan unit tests

dellaert:
Tested pybind wrapper with Shonan unit tests. 

All worked but I had to re-add the method for random initialization at level p.


dellaert:
@ProfFan This has Shonan unit tests and all appear to work. I propose you approve, merge, and then I will approve and merge the Mega change.

PR #481, opened 2020-08-19
CI For Special Cases

varunagrawal:
Add CI for following cases:

- Allow features deprecated in version 4.1.
- Use Quaternions instead of Rotation matrices.
- Enable typedef of Points to Vector.

This is done for Ubuntu (with both GCC and clang) and macOS.

### DIFF
--- a/.github/workflows/build-special.yml
+++ b/.github/workflows/build-special.yml
@@ -19,16 +19,11 @@ jobs:
       matrix:
         # Github Actions requires a single row to be added to the build matrix.
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
-        name: [
+        name:
+          [
             ubuntu-gcc-deprecated,
             ubuntu-gcc-quaternions,
             ubuntu-gcc-points-vector,
-            macos-xcode-deprecated,
-            macos-xcode-quaternions,
-            macos-xcode-points-vector,
-            ubuntu-clang-deprecated,
-            ubuntu-clang-quaternions,
-            ubuntu-clang-points-vector
           ]
 
         build_type: [Debug, Release]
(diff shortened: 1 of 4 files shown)


ProfFan:
@varunagrawal Great, also, could you turn off the `ubuntu-18.04-gcc-9` CI for Python here too? It always timeouts. Just comment them out saying "Disabled for now because of timeouts", thanks!

.github/workflows/build-special.yml
```diff
@@ -0,0 +1,153 @@
...
+
+      matrix:
+        # Github Actions requires a single row to be added to the build matrix.
+        # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
+        name: [
+            ubuntu-gcc-deprecated,
+            ubuntu-gcc-quaternions,
+            ubuntu-gcc-points-vector,
+            macos-xcode-deprecated,
+            macos-xcode-quaternions,
```
ProfFan, line 27: I think we just need to test all these with gcc on ubuntu, to reduce CI estate.

varunagrawal: Sounds good. I wasn't sure what coverage we were aiming for hence I threw the kitchen sink at it.

dellaert:
29 runs now?? I guess @ProfFan will remove typedef and hence cut down. Maybe you should coordinate on order of merging and conflict resolution for those,

ProfFan:
@dellaert We could merge this first?

varunagrawal:
More than happy to make the updates once we enable that typedef to be the default.

varunagrawal:
Merging. We can re-evaluate the CI after we make Point2/3 default to Vector types.

PR #483, opened 2020-08-20
Shonan averaging cli

dellaert:
A C++ and python driver for Shonan averaging (followed by translation averaging)

### DIFF
--- a/examples/ShonanAveragingCLI.cpp
+++ b/examples/ShonanAveragingCLI.cpp
@@ -1,0 +1,125 @@
+/* ----------------------------------------------------------------------------
+
+* GTSAM Copyright 2010, Georgia Tech Research Corporation,
+* Atlanta, Georgia 30332-0415
+* All Rights Reserved
+* Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+* See LICENSE for the license information
+* -------------------------------------------------------------------------- */
+
+/**
+ * @file    ShonanAveragingCLI.cpp
+ * @brief   Run Shonan Rotation Averaging Algorithm on a file or example dataset
+ * @author  Frank Dellaert
+ * @date    August, 2020
+ *
+ * Example usage:
+ *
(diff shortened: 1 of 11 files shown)


dellaert:
@ProfFan this unfortunately triggers the first issue with pybind. In particular, keys() throws this error:

```
  File "/Users/dellaert/git/github/python/gtsam/examples/ShonanAveragingCLI.py", line 46, in estimate_poses_given_rot
    keys = factor.keys()
TypeError: Unable to convert function return value to a Python type! The signature was
        (self: gtsam.gtsam.NonlinearFactor) -> std::__1::vector<unsigned long long, tbb::tbb_allocator<unsigned long long> >

Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.
```

It's strange as I get the same with `Pose3SLAMExample_g2o.py`, and I thought all python examples ran?

ProfFan:
this is beacuse tbb changed the return type

Check specialization.h and change (there should be an ifdef added)

dellaert:
> this is beacuse tbb changed the return type
> 
> Check specialization.h and change (there should be an ifdef added)

@ProfFan I don't understand. Please ping me on Slack when you can chat about this.

dellaert:
> > this is beacuse tbb changed the return type
> > Check specialization.h and change (there should be an ifdef added)
> 
> @ProfFan I don't understand. Please ping me on Slack when you can chat about this.

Actually, I see now: TBB flag adds allocator to the vectors. I created issue  #484

ProfFan:
@dellaert Do you want me to create a PR to develop or here?

dellaert:
> @dellaert Do you want me to create a PR to develop or here?

Separate PR. Several current examples fail with TBB on.

dellaert:
@varunagrawal, @ProfFan now ready for review
@jingwuOUO FYI

PR #493, opened 2020-08-24
Make sure our readme is proper restructuredText

ProfFan:
Otherwise pypi will not allow the wheel to be uploaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ mapping (SAM) in robotics and vision, using Factor Graphs and Bayes
 Networks as the underlying computing paradigm rather than sparse
 matrices.
 
-The current support matrix is::
+The current support matrix is:
 
 | Platform     | Compiler  | Build Status  |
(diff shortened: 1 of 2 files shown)


varunagrawal:
We don't need restructured text for the Readme. We need to set a value in the setup.py to enable using markdown.

varunagrawal:
Huh that change makes no sense to me...

ProfFan:
@varunagrawal Oh I don't know if that is allowed :) Could you point me to a doc about this? I can change it really quick

ProfFan:
Explanation: :: marks the next section to be code in restructuredText, so this table will become a code block and render proper restructuredText.

varunagrawal:
Here we go: https://stackoverflow.com/a/26737258/1236990

I was most happy when this feature was released.

ProfFan:
@varunagrawal Done :) Thanks for the heads up!

PR #501, opened 2020-08-31
Example Datasets with Python Package

### LINKED ISSUE #500: FindExampleDataFile in python wrapper does not work
## Description

Seems like the path used in gtsam::findExampleDataFile is the one from the python wheel build.

```
ValueError: gtsam::findExampleDataFile could not find a matching file in
/Users/runner/work/gtsam-manylinux-build/gtsam-manylinux-build/gtsam/examples/Data
```

## Steps to reproduce

```python
import gtsam
_ = gtsam.findExampleDataFile("toyExample.g2o")
```

## Expected behavior

Should look in directory where python wheel was installed

## Environment

MacOS, python 3.8

## Additional information

Could be hard to solve

varunagrawal:
Updated the `setup.py` to include the example data files and datasets. Resolves #500.

Note that we now install the data to a common "gtsam_example_data" directory, since we need CMake to know the path of the directory when building the `.so` file. This is the only way I've found to break this circular dependency (setup.py depends on .so and .so depends on where setup.py installs the data), so any better suggestions or recommendations are welcome!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -428,14 +428,6 @@ add_subdirectory(CppUnitLite)
 if(GTSAM_BUILD_PYTHON)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/wrap/cmake")
     add_subdirectory(python)
-    if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-        # Common directory which always has write access
-        # This is because the exact location of the Python installation may vary
-        # depending on various factors such as python version and use of virtual envs.
-        set(GTSAM_PYTHON_DATASET_DIR "/tmp/gtsam_example_data")
-    elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
-        set(GTSAM_PYTHON_DATASET_DIR "$ENV{LOCALAPPDATA}/gtsam_example_data")
-    endif()
 endif()
 
 # Build GTSAM library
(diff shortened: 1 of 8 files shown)


ProfFan:
I think this would not work for wheels. I propose we have a Python version of this function, which uses `import site` to get the folder

varunagrawal:
I actually checked that and it works with the wheel format. I am double checking right now.

varunagrawal:
Yup, on running `python setup.py bdist_wheel`, I see output for adding the data files to the binary. :tada:

ProfFan:
I mean we should not touch anything outside of the site-packages directory, per the standard guidelines for python packages.

dellaert:
> site-packages directory

Indeed, @varunagrawal, does the code now refer to the site-packages? That would be powerful magic, and would be the only way it *could* work for people that use "pip install gtsam", right?

varunagrawal:
It doesn't quite unfortunately, because the C++ function `findExampleDataFile` has no idea what the site-packages directory can be at compile time.

I have seen packages like `nltk` be a bit more relaxed when it comes to packaging data (they download the corpus data into the home directory) but I agree with @ProfFan that it's good to stick with standards to avoid code rot (amongst other things).

varunagrawal:
I'll add the custom function in a bit.

dellaert:
@varunagrawal Yeah let's not be relaxed. If there is no c++ way, is there a python way as Fan suggested?

varunagrawal:
@dellaert there is no C++ way without some crazy magic due to the circular dependency (and even then it could break).

I am creating the python version. Building the wheel as of this comment.

varunagrawal:
@dellaert @ProfFan all done!

PR #502, opened 2020-08-31
Update NavState Printing

varunagrawal:
Update the printing of `NavState` so it is easier to read.

Example:
```cpp
NavState s;
cout << s << endl;
```

Before:

```sh
R: [
	1, 0, 0;
	0, 1, 0;
	0, 0, 1
]
p: 0
0
0
v: 0
0
0
```

After:

```sh
R: [
	1, 0, 0;
	0, 1, 0;
	0, 0, 1
]
p: 0 0 0
v: 0 0 0
```

### DIFF
--- a/gtsam/geometry/Point2.cpp
+++ b/gtsam/geometry/Point2.cpp
@@ -24,12 +24,6 @@ using namespace std;
 namespace gtsam {
 
 /* ************************************************************************* */
-ostream &operator<<(ostream &os, const Point2& p) {
-  os << p.transpose();
-  return os;
-}
-
-/* ************************************************************************* */
 double norm2(const Point2& p, OptionalJacobian<1,2> H) {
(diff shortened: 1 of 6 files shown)


varunagrawal:
So this change actually affects Vector2 and Vector3 directly. Why not just update the transpose at the source rather than have `.transpose()` calls littered everywhere?

dellaert:
> So this change actually affects Vector2 and Vector3 directly. Why not just update the transpose at the source rather than have `.transpose()` calls littered everywhere?

What about Vector4, Vector5? Spirit of my comment is: Eigen is Eigen, not some weird hybrid we create on top of it.

varunagrawal:
Fair enough. I'll update this PR to only change the printing of `NavState` then. :slightly_smiling_face:

PR #505, opened 2020-09-01
Fixed issue with Coriolis acceleration

tsetterf:
The corrections for dP and dV are in the navigation frame, whereas xi is in the local frame. The easiest fix is just to rotate it into the local frame within the coriolis function. I have updated the Jacobians as well.

### DIFF
--- a/gtsam/navigation/NavState.cpp
+++ b/gtsam/navigation/NavState.cpp
@@ -215,41 +215,40 @@ NavState NavState::update(const Vector3& b_acceleration, const Vector3& b_omega,
 Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
     OptionalJacobian<9, 9> H) const {
   TIE(nRb, n_t, n_v, *this);
-  Matrix3 bRn = nRb.matrix().transpose();
   const double dt2 = dt * dt;
   const Vector3 omega_cross_vel = omega.cross(n_v);
 
-  Vector9 xi;
-  Matrix3 D_dP_R;
-  dR(xi) << nRb.unrotate((-dt) * omega, H ? &D_dP_R : 0);
-  dP(xi) << ((-dt2) * omega_cross_vel); // NOTE(luca): we got rid of the 2 wrt INS paper
-  dV(xi) << ((-2.0 * dt) * omega_cross_vel);
+  // Get perturbations in nav frame
+  Vector9 n_xi, xi;
+  Matrix3 D_dR_R, D_dP_R, D_dV_R, D_body_nav;
+  dR(n_xi) << ((-dt) * omega);
+  dP(n_xi) << ((-dt2) * omega_cross_vel); // NOTE(luca): we got rid of the 2 wrt INS paper
+  dV(n_xi) << ((-2.0 * dt) * omega_cross_vel);
   if (secondOrder) {
     const Vector3 omega_cross2_t = omega.cross(omega.cross(n_t));
-    dP(xi) -= (0.5 * dt2) * omega_cross2_t;
-    dV(xi) -= dt * omega_cross2_t;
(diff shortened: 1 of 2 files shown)


gtsam/navigation/NavState.cpp
```diff
@@ -228,18 +228,27 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
     dP(xi) -= (0.5 * dt2) * omega_cross2_t;
     dV(xi) -= dt * omega_cross2_t;
   }
+
```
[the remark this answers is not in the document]
tsetterf: Yes, that's a good point.

tsetterf:
I am convinced that as written `dP` adn `dV` are in nav coordinates. `omega` is planet angular velocity in the planet-centered planet-fixed frame (i.e., the nav frame). `n_v` and `n_t` are in the nav frame (as annotated). Therefore their cross products are in the nav frame.
```
const Vector3 omega_cross_vel = omega.cross(n_v);
const Vector3 omega_cross2_t = omega.cross(omega.cross(n_t));
```

These are the only vector terms in `dP` and `dV`, so they are also in the nav frame (prior to my fix).
```
dP(xi) << ((-dt2) * omega_cross_vel); 
dV(xi) << ((-2.0 * dt) * omega_cross_vel);
dP(xi) -= (0.5 * dt2) * omega_cross2_t;
dV(xi) -= dt * omega_cross2_t;
```

I agree that it seems like this code is not unit tested in a rich enough scenario to reveal the bug. It is tested for self consistency, ensuring that the analytical and numerical Jacobians match.

gtsam/navigation/tests/testNavState.cpp
```diff
@@ -192,6 +192,49 @@ TEST(NavState, Coriolis2) {
   EXPECT(assert_equal(numericalDerivative21(coriolis, state2, true), aH));
 }
 
+TEST(NavState, Coriolis3) {
```
tsetterf, line 195: I added this basic unit test. It fails on the old NavState, and passes on my branch.

dellaert:
Aha, I think I might have had it backwards, indeed ! You are *converting* from nav to body, to *agree* with the  convention in ImuFactor - that tangent space is in body coordinates. That sounds great, and especially the addition of a fail-first-then-succeed unit test provides a warm feeling :-)

Will review again and merge on CI.

gtsam/navigation/NavState.cpp
```diff
@@ -228,18 +229,27 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
...
+  // NOTE(tim): position and velocity changes dP and dV are in nav frame; need
+  // to put them into local frame for subsequent retract
+  dP(xi) = bRn * dP(xi);
+  dV(xi) = bRn * dV(xi);
```
dellaert, line 236: Not loving this imperative code here now, so I propose you build n_xi above (you forgot one unrotate, BTW), and then here:
```
// Convert to body-coordinates:
Vector9 xi;
xi << bRn * dR(n_xi), bRn * dP(n_xi), bRn * dV(n_xi);
```

Makes it clearer overall as well.

gtsam/navigation/tests/testNavState.cpp
dellaert: Nice!

gtsam/navigation/NavState.cpp
tsetterf: I don't think I forgot one unrotate, since dR was already unrotated in your original code. My style was meant to minimize the changes to existing code. But since you asked, I have pushed a commit that I think makes it clearer what is going on, mostly as you suggest.

dellaert:
OK, this is awesome! Many thanks for this excellent contribution! I will merge now.

PR #506, opened 2020-09-03
New readBal function

varunagrawal:
This PR adds a new function readBal which returns the SfmData object, allowing for single line calling.

This function is particularly useful with the wrapper.

E.g.
```cpp
SfmData data = readBal("example.bal")
```
rather than
```
SfmData data;
readBAL("example.bal", data);
```

*Note* I followed Google style when naming the function.


varunagrawal:
@dellaert @ProfFan this is a quick one. :)

dellaert (approved):

PR #509, opened 2020-09-03
Update minimum Boost version required

### LINKED ISSUE #508: Bump minimum Boost version
## Description

Currently, the minimum Boost version required is 1.43. The version required by the tests and Actions CI is 1.69.0. 
Moreover, some tests won't pass with Boost 1.43 due to bugs that have since been fixed in the subsequent versions.

It would be nice to have consistency to avoid unnecessary pain for users. Recommended version is 1.58 (thanks @chrisbeall).

## Steps to reproduce

1. Install Boost 1.43
2. Compile GTSAM and run tests, sit back and watch things break.

## Expected behavior

All tests should pass and Boost should not be an issue.

## Environment

N/A

varunagrawal:
This PR fixes #508.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,7 +166,7 @@ if(MSVC AND BUILD_SHARED_LIBS)
 endif()
 
 # Store these in variables so they are automatically replicated in GTSAMConfig.cmake and such.
-set(BOOST_FIND_MINIMUM_VERSION 1.43)
+set(BOOST_FIND_MINIMUM_VERSION 1.58)
 set(BOOST_FIND_MINIMUM_COMPONENTS serialization system filesystem thread program_options date_time timer chrono regex)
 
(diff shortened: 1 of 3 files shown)


ProfFan:
I would prefer a warning instead of hard fail (if GTSAM still builds against 1.43?)

varunagrawal:
I haven't tried building against 1.43 but from my experience with the CI, I don't think it will.

varunagrawal:
@ProfFan my main concern is that if we just give a warning (rather than failing), some users may not see the warning as important and come and raise issues here. By providing a hard fail, we emphasize their Boost version is not supported.

Thoughts @dellaert?

dellaert:
> @ProfFan my main concern is that if we just give a warning (rather than failing), some users may not see the warning as important and come and raise issues here. By providing a hard fail, we emphasize their Boost version is not supported.
> 
> Thoughts @dellaert?

Agreed. We should put minimum version in cmake which is the *true* minimal version for a bug-free GTSAM. I think cmake should hard-error if not found.

ProfFan (approved):
Sure, @dellaert agreed :)

PR #511, opened 2020-09-06
Wrap the DSFMap class so SFM can use them

ProfFan:
Note: this contains a workaround on the template specialization limit of the wrapper.

### DIFF
--- a/gtsam/base/DSFMap.h
+++ b/gtsam/base/DSFMap.h
@@ -120,4 +120,12 @@ class IndexPair : public std::pair<size_t,size_t> {
   inline size_t i() const { return first; };
   inline size_t j() const { return second; };
 };
+
+typedef std::vector<IndexPair> IndexPairVector;
+typedef std::set<IndexPair> IndexPairSet;
+
+inline IndexPairVector IndexPairSetAsArray(IndexPairSet& set) { return IndexPairVector(set.begin(), set.end()); }
+
+typedef std::map<IndexPair, IndexPairSet> IndexPairSetMap;
+typedef DSFMap<IndexPair> DSFMapIndexPair;
 }  // namespace gtsam
--- a/gtsam/base/DSFMap.h
+++ b/gtsam/base/DSFMap.h
@@ -21,6 +21,7 @@
 #include <cstdlib>  // Provides size_t
 #include <map>
 #include <set>
+#include <vector>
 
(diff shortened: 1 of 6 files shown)


ProfFan:
@dellaert @varunagrawal

gtsam/base/DSFMap.h
```diff
@@ -120,4 +121,12 @@ class IndexPair : public std::pair<size_t,size_t> {
...
   inline size_t j() const { return second; };
 };
+
+typedef std::vector<IndexPair> IndexPairVector;
+typedef std::set<IndexPair> IndexPairSet;
+
+inline IndexPairVector IndexPairSetAsArray(IndexPairSet& set) { return IndexPairVector(set.begin(), set.end()); }
+
+typedef std::map<IndexPair, IndexPairSet> IndexPairSetMap;
+typedef DSFMap<IndexPair> DSFMapIndexPair;
```
varunagrawal, line 131: `using` instead of `typedef`?
My understanding is that `using` plays nicer with templates. I am not sure if the wrapper supports this though.

gtsam/gtsam.i
```diff
@@ -259,11 +259,59 @@ class IndexPair {
...
+  // common STL methods
+  size_t size() const;
+  bool empty() const;
+  void clear();
+
+  // structure specific methods
+  gtsam::IndexPairSet at(gtsam::IndexPair& key);
+};
+
+class DSFMapIndexPair {
```
varunagrawal, line 310: Keep this as templated for future proofing?

varunagrawal (approved):
Functionally looks great!
Some design/style recommendations though.

gtsam/gtsam.i
ProfFan: That's wont work because of the nested template in IndexPairSetMap

gtsam/base/DSFMap.h
ProfFan: I think other such definitions use typedef, so let's keep it consistent :)

gtsam/gtsam.i
ProfFan: Sadly *

ProfFan:
@swarrier246 Now you can switch back to develop :)

swarrier246:
Thanks, @ProfFan ! Appreciate the help

PR #513, opened 2020-09-07
Wrap translation averaging

akshay-krishnan:
This PR wraps translation averaging to provide a Python interface. The included python test verifies that the interface works well.

### DIFF
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2914,13 +2914,6 @@ class BinaryMeasurement {
 typedef gtsam::BinaryMeasurement<gtsam::Unit3> BinaryMeasurementUnit3;
 typedef gtsam::BinaryMeasurement<gtsam::Rot3> BinaryMeasurementRot3;
 
-class BinaryMeasurementsUnit3 {
-  BinaryMeasurementsUnit3();
-  size_t size() const;
-  gtsam::BinaryMeasurement<gtsam::Unit3> at(size_t idx) const;
-  void push_back(const gtsam::BinaryMeasurement<gtsam::Unit3>& measurement);
-};
-
 #include <gtsam/sfm/ShonanAveraging.h>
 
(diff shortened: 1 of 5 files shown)


gtsam/gtsam.i
```diff
@@ -2914,6 +2914,13 @@ class BinaryMeasurement {
 typedef gtsam::BinaryMeasurement<gtsam::Unit3> BinaryMeasurementUnit3;
 typedef gtsam::BinaryMeasurement<gtsam::Rot3> BinaryMeasurementRot3;
 
+class BinaryMeasurementsUnit3 {
+  BinaryMeasurementsUnit3();
```
dellaert, line 2918: All these are generic and should be just added above in `BinaryMeasurement`

dellaert (approved):
Awesome, just one small comment...

gtsam/gtsam.i
akshay-krishnan: Great, I did not know that. Removed it.

dellaert: *Actually*, I apologize but I was wrong: 
1) I did not see this was about the cevtor of measurements, so my comment was wrong
2) by the specialization stuff, these methods are indeed available in a *pythonesque* way in the python wrapper. However, I think it still needs to be here for the MATLAB wrapper. @ProfFan can confirm.
PS @ProfFan this state of affairs is really not great, even thought not necessarily in your plate: I think we should adapt the wrap language to deal with these things in a uniform and transparent way. After all, the wrap code is a code *generator*. Surely this should not be on the user.

ProfFan: Yes, you still need to create the`XxxVector`, `XxxMap` type aliases for the MATLAB wrapper. I demonstrated this in #511 for DSFMap, you can refer to that PR @akshay-krishnan .

I agree that we should extend the wrap language to deal with this, but currently I have no bandwidth on it, and I cannot think of an idiomatic way to do this...

PR #518, opened 2020-09-11
ISAM2 Evaluate Nonlinear Error

varunagrawal:
This PR adds two helper methods to `ISAM2Result` which allows wrapper access to the `errorBefore` and `errorAfter` nonlinear errors of ISAM2.

The helper methods are setup to leverage the boost optional capabilities, such that if `evaluateNonlinearError` is `False`, they will return `nan`. This works great with the python wrapper i.e. the result is printed out as `nan`.

Example use case:

```python
isam2_parameters = gtsam.ISAM2Params()
isam2_parameters.setEvaluateNonlinearError(True)
#
#
#
result = self.isam.update(graph, initial_estimate)
print("Before Error {0}".format(result.getErrorBefore()))
print("After Error  {0}".format(result.getErrorAfter()))
```

This method has proven super useful in helping to debug various setup issues in ISAM2 by evaluating the error.

Any feedback involving location of tests for these methods?


varunagrawal:
Reminder :slightly_smiling_face:

dellaert (approved):

PR #525, opened 2020-09-16
Wrap upgrade

varunagrawal:
Perform a subtree update of wrap.

### Motivation

Pybind11 had a relative submodule which even though we didn't use, would cause errors in Github Actions due to the way the checkout action is set up.

After raising this issue with the folks at Pybind11, they agreed it was problematic and updated the codebase to not have the submodule anymore. This PR adds in those changes.

**tl;dr** Pybind11 made some updates, so we updated it too so Actions doesn't complain.


ProfFan (approved):
LGTM:)

varunagrawal:
I'm going to merge this and then investigate the build issues. It seems to be on Github's end from what I can glean.

Issue #522, opened 2020-09-13
Default Python version is bad


varunagrawal:
## Description

The default Python version in CMake is `Default`. When turning on the `GTSAM_BUILD_PYTHON` flag, the `Default` values causes an error with Pybind11.

```sh
Found PythonInterp: /home/varun/.pyenv/shims/python (found version "3.8.5")
 Found PythonLibs: /usr/lib/libpython3.8.so (found version "3.8.5")
 CMake Error at wrap/pybind11/tools/pybind11Tools.cmake:16 (find_package):
   find_package called with invalid argument "Default"
 Call Stack (most recent call first):
   wrap/pybind11/CMakeLists.txt:33 (include)

 pybind11 v2.5.dev1
```

Setting the Python version to a specific version, such as `3.8`, fixes the issue, so the proposed solution would be to set the default to empty ("") and use the CMake variables from Python to set the default version value. @dellaert @ProfFan 

## Steps to reproduce

1. Use fresh build directory.
2. Set the `GTSAM_BUILD_PYTHON` flag to `ON`.
3. Run `cmake ..`

## Expected behavior

CMake should not throw an error and should configure the build without issues.

## Environment

- OS: Arch Linux
- Clang compiler

PR #526, opened 2020-09-18
Automatically find Python3 version

### LINKED ISSUE #519: Matlab wrapper uses python 3
Problem:
[ 56%] Generating wrap/gtsam/gtsam_wrapper.cpp
Ignoring classes: ['gtsam::Point2', 'gtsam::Point3']
  File "/path/gtsam/wrap/matlab_wrapper.py", line 1770, in <module>
    cc_content = wrapper.wrap()
  File "/path/gtsam/wrap/matlab_wrapper.py", line 1657, in wrap
    self.wrap_namespace(self.module)
  File "/path/gtsam/wrap/matlab_wrapper.py", line 1096, in wrap_namespace
    self.wrap_namespace(element, namespaces)
  File "/path/gtsam/wrap/matlab_wrapper.py", line 1101, in wrap_namespace
    class_text = self.wrap_instantiated_class(element, "".join(namespace.full_namespaces()))
  File "/path/gtsam/wrap/matlab_wrapper.py", line 1037, in wrap_instantiated_class
    ).splitlines()) + '\n'
  File "/path/gtsam/wrap/matlab_wrapper.py", line 698, in wrap_class_constructors
    methods_wrap = textwrap.indent(textwrap.dedent("""\

To generate the Matlab wrapper it requires python 3.  I fixed by changing "Default" to my python 3 version.

https://github.com/borglab/gtsam/blob/de75367e5e8f49f468474a75e9fd4bdb15fc856e/CMakeLists.txt#L104

varunagrawal:
Use CMake to identify best Python3 version if `Default` requested.
Note that the CMake code looks specifically for Python3 as we shouldn't be supporting the EOL'd Python2.

This also helps get rid of a bunch of checks for the `Default` value.

Fixes #519 and #522.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,11 @@ if(GTSAM_BUILD_PYTHON)
     endif()
 
     if(${GTSAM_PYTHON_VERSION} STREQUAL "Default")
-        set(GTSAM_PYTHON_VERSION "${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}" CACHE STRING "The version of Python to build the wrappers against." FORCE)
+        set(GTSAM_PYTHON_VERSION "${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}"
+                CACHE
+                STRING
+                "The version of Python to build the wrappers against."
+                FORCE)
     endif()
 
     if(GTSAM_UNSTABLE_BUILD_PYTHON)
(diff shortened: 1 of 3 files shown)


ProfFan:
https://cmake.org/cmake/help/latest/module/FindPython.html : 
"Python3 is searched and will fall back to Python2"

varunagrawal:
I wanted to fail if Python3 is not found, rather than fall back to Python2. I strongly believe we should not support Python2 on our end.

ProfFan (approved):

fangxu622:
I have the same questions # 
can't build matlab wrapper,  #552 
but build python wrapper is ok~

PR #532, opened 2020-09-19
Adding logmap API which takes the object to the function argument

ayushbaid:
The `LieGroup` supports `Logmap` (log-map of the argument at identity) as well as `logmap` (logmap of the element between the current object and the argument).

The `logmap` variant is present in gtsam.i, and hence not available in the python wrapper.

This PR adds the `logmap` function where-ever `Logmap` was defined.

Testing details: I build the gtsam library and python wrapper, and checked Rot3. As the change is similar in every place, it should work for all cases.

### DIFF
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -361,7 +361,6 @@ class LieScalar {
   // Lie group
   static gtsam::LieScalar Expmap(Vector v);
   static Vector Logmap(const gtsam::LieScalar& p);
-  Vector logmap(const gtsam::LieScalar& p);
 };
 
 #include <gtsam/base/deprecated/LieVector.h>
@@ -509,7 +508,6 @@ class StereoPoint2 {
   // Lie Group
   static gtsam::StereoPoint2 Expmap(Vector v);
   static Vector Logmap(const gtsam::StereoPoint2& p);
-  Vector logmap(const gtsam::StereoPoint2& p);
 
   // Standard Interface
   Vector vector() const;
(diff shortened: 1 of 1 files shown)


ProfFan (approved):
Awesome :)

dellaert:
Will merge as we discussed LieXXX stuff in meeting

varunagrawal:
Ummm this broke the Matlab wrapper.

I don't understand why `logmap` was added to `LieMatrix` and `LieVector` if they have been deprecated?
@ayushbaid

dellaert:
How did it break the matlab wrapper?
And: if we deprecate LieXXX, we should do it forcefully. I talked to @ProfFan about it

varunagrawal:
Apparently the `logmap` functions are not defined in the `LieMatrix` and `LieVector` classes.

```
/home/varun/borglab/gtsam/build/gtsam/wrap/gtsam/gtsam_wrapper.cpp: In function ‘void gtsamLieVector_logmap_193(int, mxArray**, int, const mxArray**)’:
/home/varun/borglab/gtsam/build/gtsam/wrap/gtsam/gtsam_wrapper.cpp:4328:32: error: ‘struct gtsam::LieVector’ has no member named ‘logmap’; did you mean ‘Logmap’?
 4328 |   out[0] = wrap< Vector >(obj->logmap(p));
      |                                ^~~~~~
      |                                Logmap
/home/varun/borglab/gtsam/build/gtsam/wrap/gtsam/gtsam_wrapper.cpp: In function ‘void gtsamLieMatrix_logmap_212(int, mxArray**, int, const mxArray**)’:
/home/varun/borglab/gtsam/build/gtsam/wrap/gtsam/gtsam_wrapper.cpp:4492:32: error: ‘struct gtsam::LieMatrix’ has no member named ‘logmap’; did you mean ‘Logmap’?
 4492 |   out[0] = wrap< Vector >(obj->logmap(p));
      |                                ^~~~~~
      |                                Logmap
```

ayushbaid:
So should I remove this from LieMatrix and LieVector for the time being? Before we can figure out the deprecation possibility for these 2 classes?

varunagrawal:
I made a PR since it was quick and I need this to work for another PR.

@ayushbaid can you please take a look at it? It should be quick :)

PR #533, opened 2020-09-20
Added accelerated power method to compute eigenpair fast

jingwuOUO:


gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+#include <random>
+
+namespace gtsam {
+
+using Sparse = Eigen::SparseMatrix<double>;
+
+/* ************************************************************************* */
+/// MINIMUM EIGENVALUE COMPUTATIONS
+
+struct PowerFunctor {
```
david-m-rosen, line 44: This strikes me as unnecessarily complicated ... is a struct _really_ necessary here?  Wouldn't it be much simpler to have a single (standalone) function  with a signature like:

``` bool powerMethod(const SparseMatrix& A, const Vector& v0, int max_iters, double eps, Vector& v, int& num_iters, bool accelerated = false, double beta = 0) ```

File: gtsam/sfm/PowerMethod.h [hunk 2]
```diff
@@ -0,0 +1,367 @@
...
+   *
+   */
+
+  // Const reference to an externally-held matrix whose minimum-eigenvalue we
+  // want to compute
+  const Sparse &A_;
+
+  const Matrix &S_;
+
+  const int m_n_; // dimension of Matrix A
```
david-m-rosen, line 68: Why do we need to store this value?

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+  const int m_nev_; // number of eigenvalues required
```
david-m-rosen, line 69: Similarly, why is this needed?  (The power method only ever returns a **single** Ritz pair ... )

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+private:
+  Vector ritz_val_; // all ritz eigenvalues
+  Matrix ritz_vectors_; // all ritz eigenvectors
+  Vector ritz_conv_; // store whether the ritz eigenpair converged
+  Vector sorted_ritz_val_; // sorted converged eigenvalue
+  Matrix sorted_ritz_vectors_; // sorted converged eigenvectors
+
+public:
+ // Constructor
+ explicit PowerFunctor(const Sparse& A, const Matrix& S, int nev, int ncv,
```
david-m-rosen, line 89: The fact that there's an A AND an S here tells me that this is NOT just doing a basic power method computation -- it looks like this is doing the full minimum eigenpair computation.

You should separate the (optionally accelerated) power method from the full minimum-eigenvalue computation by writing two separate functions.  This is logically cleaner/clearer, and it's better for code re-use and unit testing.

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+       m_n_(A.rows()),
+       m_nev_(nev),
+       // m_ncv_(ncv > m_n_ ? m_n_ : ncv),
+       accelerated_(accelerated),
+       beta_(beta),
+       m_niter_(0) {
+         // Do nothing
+       }
+
+ int rows() const { return A_.rows(); }
```
david-m-rosen, line 102: Why are these necessary?

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+ int cols() const { return A_.cols(); }
+
+ // Matrix-vector multiplication operation
+ Vector perform_op(const Vector& x1, const Vector& x0) const {
```
david-m-rosen, line 106: Why is this necessary?  Why not just write x1 = A*x0 as needed?

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+   // Do the multiplication
+   Vector x2 = A_ * x1 - beta_ * x0;
+   x2.normalize();
+   return x2;
+  }
+
+  Vector perform_op(const Vector& x1, const Vector& x0, const double beta) const {
```
david-m-rosen, line 113: Similarly, why is this needed?

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+    for (int i = 0; i < len; i++) {
+      long m_rand = next_long_rand(m_rand);
+      res[i] = double(m_rand) / double(m_max) - double(0.5);
+    }
+    res.normalize();
+    return res;
+  }
+
+  /// Tuning the momentum beta using the Best Heavy Ball algorithm in Ref(3)
+  void setBeta() {
```
david-m-rosen, line 166: Again, I would suggest that you break this out as a standalone function (implemented in the same file as the power method) -- possibly as a helper function that is called from inside the main powerMethod() function, if beta actually _needs_ to be estimated.   

[Note that it's possible to run the accelerated power method using a _pre-specified_ value of beta (i.e., without having to estimate a good value empirically).  This is actually the case for our intended application (computing the minimum eigenvalue of the certificate matrix) ... ]

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+            break;
+          }
+        }
+      }
+    }
+
+    beta_ = betas[maxIndex];
+  }
+
+  void perturb(int i) {
```
david-m-rosen, line 207: You can replace this entire function with a single call to Eigen Vector's Random() static construtor

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+      disturb(i) = uniform01(generator);
+    }
+    disturb.normalize();
+
+    Vector x0 = ritz_vectors_.col(i);
+    double magnitude = x0.norm();
+    ritz_vectors_.col(i) = x0 + 0.03 * magnitude * disturb;
+  }
+
+  void init(const Vector x0, const Vector x00) {
```
david-m-rosen, line 227: Initialization should be handled outside the powerMethod() function

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+    }
+    init(x0, x00);
+  }
+
+  bool converged(double tol, int i) {
+    Vector x = ritz_vectors_.col(i);
+    double theta = x.transpose() * A_ * x;
+
+    // store the ritz eigen value
+    ritz_val_(i) = theta;
```
david-m-rosen, line 277: You should NOT put side-effects like this inside a function that looks like it's just checking for convergence; this is going to do weird stuff if unwary programmers try to use it in the obvious way.  Functions like this should be ```const```

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+
+    // update beta
+    if (accelerated_) beta_ = std::max(beta_, theta * theta / 4);
```
david-m-rosen, line 280: beta should never be changed within a single execution of this method ...

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+
+    Vector diff = A_ * x - theta * x;
+    double error = diff.lpNorm<1>();
```
david-m-rosen, line 283: The Ritz residual should be measured using the L2 norm

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,367 @@
...
+      }
+    }
+    return num_converge;
+  }
+
+  size_t num_iterations() {
+    return m_niter_;
+  }
+
+  void sort_eigenpair() {
```
david-m-rosen, line 307: I'm not sure why this function is necessary, since the power method only ever computes a single eigenpair ...

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -470,8 +471,54 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
   return Lambda - Q_;
 }
 
-/* ************************************************************************* */
-/// MINIMUM EIGENVALUE COMPUTATIONS
+// Alg.6 from paper Distributed Certifiably Correct Pose-Graph Optimization
+static bool PowerMinimumEigenValue(
```
david-m-rosen, line 475: You only actually need S here (check Alg 6 in the paper)

david-m-rosen (changes requested):
Hey @jingwuOUO,

Looks like you have all of the necessary pieces here, but this implementation looks to me like it's _waaaaay_ more complicated than it needs to be.

I think it would be much better to refactor this into two simple **standalone** functions (i.e. not structs, which in my opinion are way overkill for such simple functions):

- ```powerMethod()```, which implements **only** the basic (optionally-accelerated) power method
- ```computeMinimumEigenPair()```, which implements Alg 6 in the paper (using ```powerMethod()```)

You should be able to implement a basic version of ```PowerMethod()``` in ~30 lines of code, and similarly for ```computeMinimumEigenPair()``` 

Importantly: everything involving initialization, etc. should be done _**outside**_ of the ```powerMethod()``` function -- this function should take as input the starting vector v0 (perhaps with an optional default random initialization of appropriate dimension, in case the user doesn't have, or doesn't want to supply, an explicit initial estimate).

dellaert:
@david-m-rosen i will take over from @jingwuOUO for a bit but will take your comments into account.

gtsam/sfm/PowerMethod.h
dellaert: It's necessary as we will get rid of big sparse matrices alltogether

gtsam/sfm/PowerMethod.h
dellaert: we don't

gtsam/sfm/PowerMethod.h
dellaert: Not required and removed

gtsam/sfm/PowerMethod.h
dellaert: @jingwuOUO pls reply to David about this. This is something I need clarity on before changing anything

gtsam/sfm/PowerMethod.h
dellaert: I agree. I thing see why Jinf did it, but will try a different way. Do we ever get a random pattern of "converged" pairs?

gtsam/sfm/ShonanAveraging.cpp
dellaert: @jingwuOUO please reply. I am also not clear on why we need S.

gtsam/sfm/PowerMethod.h
dellaert: Hmmm. I actually would like 3 eigenpairs, will gtell you why later. @jingwuOUO  can it actually do this?

gtsam/sfm/PowerMethod.h
david-m-rosen: Ahhh right.  In that case (and just as a suggestion): might it still make more sense to separate whatever functor you want to use for matrix-vector multiplies from the logic of the power method?  (I say this because I could see both the graphical matrix-vector multiply functor *and* the power method being independently useful for many things -- might be easier from the perspective of modularity and code reuse not to tie their implementations together in a *single* struct ... )

gtsam/sfm/PowerMethod.h
david-m-rosen: @dellaert the (optionally accelerated) power method only computes a **single** eigenpair estimate.  

It *is* possible to estimate k > 1 eigenpairs using the power method as a *subroutine* through a process called *deflation*, but this involves performing k sequential runs of the power method to compute the eigenpairs one by one.  If you really do need multiple eigenpairs, a proper Krylov subspace method like the Lanczos method really is the way to go.

gtsam/sfm/PowerMethod.h
david-m-rosen: There's only every one Ritz pair that's being estimated, so it seems like there's a lot more work being done here than is necessary.  A simple implementation of this function would be something like:

```
theta = x.dot(A*x);   // Compute Ritz value (eigenvalue estimate), assuming x is normalized
double res = (A*x - theta*x).norm();  // Compute Ritz residual: L2 norm of eigenvalue equation *A*x = lambda*x
return res <= eps;  // Check whether Ritz residual satisfies user-specified tolerance
```

gtsam/sfm/PowerMethod.h
dellaert: Yeah, that was my understanding too. But @jingwuOUO why do we then have *multiple* pairs in this code?

gtsam/sfm/PowerMethod.h
jingwuOUO: Yes, the power method only returns a single Ritz pair. m_nev_ is used when we want to sort all ritz vectors.

gtsam/sfm/PowerMethod.h
jingwuOUO: The S is passed here because I want to use the first row of S as the initial x0 as mention in the III.C part of Computational Enhancements for Certifiably Correct SLAM paper. You're right, it's less confusing if I can pass only the first row.

Now the PowerFunctor is only doing the power method.

gtsam/sfm/PowerMethod.h
david-m-rosen: But the power method only computes a **single** Ritz vector (the one that gets returned when the function terminates).  So what are these other not-Ritz-vectors that we're keeping around, and why would be interesting in "sorting" them?

gtsam/sfm/PowerMethod.h
jingwuOUO: David is correct, this line should be deleted.

gtsam/sfm/PowerMethod.h
jingwuOUO: Just to make sure, previous in the email the equation is * || Ax - theta*x || < eps *, so *||* here is l2 norm? If yes, then this should be changed to l2 norm rather than l1.

gtsam/sfm/PowerMethod.h
jingwuOUO: That was my misunderstanding.  We actually do not need this sort_eigenpair(). Power method is returning only one eigenpair.

If we want to implement lanczos method, this can be used in the future.

gtsam/sfm/ShonanAveraging.cpp
jingwuOUO: Sorry for the misunderstanding here. The S has different meanings in two paper. So in the code, S is used here to provide the initial vector x0 in power method.

(1) In the Yulun20 paper, S = S(X), here S is A in the code. X is actually S in the code.

(2) In our ECCV paper, A in the code is the certificate matrix (which is S in (1)). S in the code is S \in St(d, p)^n in our paper.

gtsam/sfm/PowerMethod.h
david-m-rosen: Unless otherwise specified, it's generally understood that ||x|| means L2 norm (standard Euclidean norm).  This should be the L2 norm.

gtsam/sfm/ShonanAveraging.cpp
david-m-rosen: OK, so just to clarify then:

- A is the matrix whose eigenpair we want to compute
- S is meant to be the first-order-critical Stiefel manifold element found by local search in the Riemannian Staircase in Shonan Averaging.  We only actually use this for constructing a good starting vector x0 for the power method, so we can get a cleaner (and more general) interface by replacing S with an (optional) argument of the form ```const Vector& x0```, which provides an (optional) initialization for the power method; if the user doesn't supply this argument, we can just take a randomly-sampled vector of appropriate dimension.

gtsam/sfm/PowerMethod.h
dellaert: Why was it ever inserted, though? What paper recommends this? What code did this?

dellaert:
@jingwuOUO back to you. However, I'd like to speak to you first before you start editing, so let's schedule a meeting.

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,315 @@
...
+                                 1.01 * maxBeta, 1.5 * maxBeta};
+
+    Matrix ritzVectors;
+    ritzVectors.resize(dim_, 10);
+    ritzVectors.setZero();
+    for (size_t i = 0; i < 10; i++) {
+      for (size_t k = 0; k < betas.size(); ++k) {
+        for (size_t j = 1; j < 10; j++) {
+          if (j < 2) {
+            Vector x0 = Vector::Random(dim_);
```
dellaert, line 170: This might have to have -0.5 added

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,315 @@
...
+                       bool accelerated = false, double beta = 0)
+      : A_(A), dim_(A.rows()), nrRequired_(nrRequired),
+        accelerated_(accelerated), beta_(beta), nrIterations_(0) {
+    Vector x0;
+    Vector x00;
+    if (!S.isZero(0)) {
+      x0 = S.row(1);
+      x00 = S.row(0);
+    } else {
+      x0 = Vector::Random(dim_);
```
dellaert, line 101: this might have to have -0.5 as well

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,315 @@
...
+public:
+  // Constructor
+  explicit PowerMethod(const Operator &A, const Matrix &S, int nrRequired = 1,
+                       bool accelerated = false, double beta = 0)
+      : A_(A), dim_(A.rows()), nrRequired_(nrRequired),
+        accelerated_(accelerated), beta_(beta), nrIterations_(0) {
+    Vector x0;
+    Vector x00;
+    if (!S.isZero(0)) {
+      x0 = S.row(1);
```
dellaert, line 98: +1 on optional initialization vector.

dellaert:
Given our conversation, please:
1) remove the sorting
2) split up the PowerMethod into two *classes* (not structs): 
- a base class PowerMethod
- a derived class AcceleratedPowerMethod
3) for PowerMethod, just have the one vector that is iterated
4) for AcceleratedPowerMethod, also add beta_ and the "previousVector_". If actually needed, that is: I think you can probably do without it.
5) replace S with an optional initial vector (for both classes)

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,247 @@
...
+
+  Vector update() const {
+    Vector y = update(this->ritzVectors_, previousVector_, beta_);
+    previousVector_ = this->ritzVectors_;
+    return y;
+  }
+
+  /// Tuning the momentum beta using the Best Heavy Ball algorithm in Ref(3)
+  void setBeta() {
+    if (PowerMethod<Operator>::dim_ < 10) return;
```
dellaert, line 209: what's up with this? Document in docstring above.

gtsam/sfm/PowerMethod.h
```diff
@@ -0,0 +1,247 @@
...
+    double maxBeta = beta_;
+    size_t maxIndex;
+    std::vector<double> betas = {2 / 3 * maxBeta, 0.99 * maxBeta, maxBeta,
+                                 1.01 * maxBeta, 1.5 * maxBeta};
+
+    Matrix tmpRitzVectors;
```
dellaert, line 215: How about renaming and saying
```Matrix R = Matrix::Zero(dim_,10);```

File: gtsam/sfm/ShonanAveraging.cpp [hunk 21]
```diff
@@ -470,8 +471,52 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
   return Lambda - Q_;
 }
 
-/* ************************************************************************* */
-/// MINIMUM EIGENVALUE COMPUTATIONS
+// Alg.6 from paper Distributed Certifiably Correct Pose-Graph Optimization
```
dellaert, line 474: Document all arguments carefully.

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -470,8 +471,52 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
...
+    // minimum eigenvalue, so just return this solution
+    *minEigenValue = lmEigenValue;
+    if (minEigenVector) {
+      *minEigenVector = lmOperator.eigenvectors();
+      minEigenVector->normalize();  // Ensure that this is a unit vector
+    }
+    return true;
+  }
+
+  Sparse C = lmEigenValue * Matrix::Identity(A.rows(), A.cols()) - A;
```
dellaert, line 504: const ?

dellaert (changes requested):
OK, This is starting to look good. Still, a large number of cleanup/documentation needed before this can go into GTSAM. I'd also like you to split the header into two files (and the tests) and move them to the linear sub-directory, as they are actually general and have nothing to do with sfm per se.

dellaert:
Just a comment: please do not force-push a public branch, ever. It breaks the review process, and a number of other things. If this is about having the latest develop, just*merge* develop into your branch.

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+    return y;
+  }
+
+  // Update the vector by dot product with A_
+  Vector update() const { return update(ritzVector_); }
+
+  // Perturb the initial ritzvector
+  void perturb() {
+    // generate a 0.03*||x_0||_2 as stated in David's paper
+    std::mt19937 rng(42);
```
dellaert, line 84: Add seed as an optional argument in constructor, with default value 42,  and pass it to perturb.

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+  double ritzValue_;   // all Ritz eigenvalues
+  Vector ritzVector_;  // all Ritz eigenvectors
+
+ public:
+  // Constructor
+  explicit PowerMethod(const Operator &A,
+                       const boost::optional<Vector> initial = boost::none)
+      : A_(A), dim_(A.rows()), nrIterations_(0) {
+    Vector x0;
+    x0 = initial ? Vector::Random(dim_) : initial.get();
```
dellaert, line 58: combine with above line

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+  Vector update() const { return update(ritzVector_); }
+
+  // Perturb the initial ritzvector
+  void perturb() {
+    // generate a 0.03*||x_0||_2 as stated in David's paper
+    std::mt19937 rng(42);
+    std::uniform_real_distribution<double> uniform01(0.0, 1.0);
+
+    int n = dim_;
+    // Vector disturb(n);
```
dellaert, line 88: kill all commented out code

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+  Vector update(const Vector &x) const {
+    Vector y = A_ * x;
+    y.normalize();
+    return y;
+  }
+
+  // Update the vector by dot product with A_
+  Vector update() const { return update(ritzVector_); }
+
+  // Perturb the initial ritzvector
```
dellaert, line 81: This is not well documented. Where does this come from?

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+    Vector disturb = Vector::Random(n);
+    disturb.normalize();
+
+    Vector x0 = ritzVector_;
+    double magnitude = x0.norm();
+    ritzVector_ = x0 + 0.03 * magnitude * disturb;
+  }
+
+  // Perform power iteration on a single Ritz value
+  // Updates ritzValue_
```
dellaert, line 101: You resolved my comment but still not document return value

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+    const Vector diff = A_ * x - theta * x;
+    double error = diff.norm();
+    return error < tol;
+  }
+
+  // Return the number of iterations
+  size_t nrIterations() const { return nrIterations_; }
+
+  // Start the iteration until the ritz error converge
+  int compute(int maxIterations, double tol) {
```
dellaert, line 118: maxIterations can't be negative so change type to size_t (check other variables as well)

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    return std::min(1, nrConverged);
+  }
+
+  // Return the eigenvalue
+  double eigenvalues() const { return ritzValue_; }
+
+  // Return the eigenvector
+  const Vector eigenvectors() const { return ritzVector_; }
```
dellaert, line 136: Please respond to my question. Is this true?? I thought we had to multiply still with some matrix in some cases?

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -472,6 +473,57 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
 
 /* ************************************************************************* */
 /// MINIMUM EIGENVALUE COMPUTATIONS
+// Alg.6 from paper Distributed Certifiably Correct Pose-Graph Optimization, 
+// it takes in the certificate matrix A as input, the maxIterations and the 
+// minEigenvalueNonnegativityTolerance is set to 1000 and 10e-4 ad default, 
+// there are two parts
+// in this algorithm:
+// (1)
```
dellaert, line 481: complete

dellaert (changes requested):
Still a number of things. Don't force-push

david-m-rosen:
@dellaert I'd also be happy to give this a close look whenever you have a version you're happy with, if you'd like a second set of eyes on it [I've not been tracking this closely for the past few weeks, as I got the sense that you and Jing wanted to take the lead on this for a bit :-) ].

dellaert:
> @dellaert I'd also be happy to give this a close look whenever you have a version you're happy with, if you'd like a second set of eyes on it [I've not been tracking this closely for the past few weeks, as I got the sense that you and Jing wanted to take the lead on this for a bit :-) ].

Yeah, my latest comments are just nits, so maybe you can take a look at the algorithmic aspects and comments? Would like to land this soon. Note, it's in linear subdirectory now and the intent is it is generic, not specific to Shonan. All Shonan-specific stuff should be in ShonanAveraging...

david-m-rosen (changes requested):
@dellaert @jingwuOUO  OK I finished giving this another close look.

Overall the Shonan Averaging code looks **much** cleaner -- there are a few small changes to the minimum eigenvalue computation that should be done to speed up the method, but after having restructured the code, these amount to changing 1 or 2 lines.  This part looks good to me.

The power method and accelerated power method code still needs a bit of work -- there's still weird stuff being done in a few places.  The unit tests also need to be fixed -- right now the logic in these tests is actually **completely broken** (see comments).

gtsam/linear/PowerMethod.h
jingwuOUO: This is now not used. Changed to Vector::Random().

gtsam/linear/PowerMethod.h
jingwuOUO: Didn't notice this wrong order! Thanks!

gtsam/linear/PowerMethod.h
jingwuOUO: It is now moved to ShonanAveraging.cpp, it comes from Part III.C David's paper Rosen, D. and Carlone, L., 2017, September. Computational enhancements for certifiably correct SLAM. Already added this in the documentation.

gtsam/linear/PowerMethod.h
jingwuOUO: Added more detailed documentation now.

gtsam/linear/PowerMethod.h
jingwuOUO: My understanding is we don't need to multiply with some matrix. @david-m-rosen what do you think?

jingwuOUO:
> @dellaert @jingwuOUO OK I finished giving this another close look.
> 
> Overall the Shonan Averaging code looks **much** cleaner -- there are a few small changes to the minimum eigenvalue computation that should be done to speed up the method, but after having restructured the code, these amount to changing 1 or 2 lines. This part looks good to me.
> 
> The power method and accelerated power method code still needs a bit of work -- there's still weird stuff being done in a few places. The unit tests also need to be fixed -- right now the logic in these tests is actually **completely broken** (see comments).

Yes. The test previous is not well-written. There are mistakes that I didn't fix. Now I have corrected the input with right value. For powermethod, the initial vector is a random one. And for accelerated, the initial is the row0 of aim matrix. 

Also revised the code according to your comments.

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+    ritzValue_ = 0.0;
+
+    // initialize Ritz eigen vectors
+    ritzVector_ = Vector::Zero(dim_);
+
+    ritzVector_.col(0) = update(x0);
+    perturb();
+  }
+
+  // Update the vector by dot product with A_
```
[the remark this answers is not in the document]
jingwuOUO: Renaming the function, thanks for the advice. Added what this function actually returns.

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+  // Return the number of iterations
+  size_t nrIterations() const { return nrIterations_; }
+
+  // Start the iteration until the ritz error converge
+  int compute(int maxIterations, double tol) {
+    // Starting
+    int nrConverged = 0;
+
+    for (int i = 0; i < maxIterations; i++) {
+      nrIterations_ += 1;
```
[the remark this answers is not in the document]
jingwuOUO: Thanks! It is!

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+    Vector x0 = ritzVector_;
+    double magnitude = x0.norm();
+    ritzVector_ = x0 + 0.03 * magnitude * disturb;
+  }
+
+  // Perform power iteration on a single Ritz value
+  // Updates ritzValue_
+  bool iterateOne(double tol) {
+    const Vector x = ritzVector_;
+    double theta = x.transpose() * A_ * x;
```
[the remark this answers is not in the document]
jingwuOUO: Got it

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,133 @@
...
+   * enhancements for certifiably correct SLAM. In Proceedings of the
+   * International Conference on Intelligent Robots and Systems.
+   * 2) Yulun Tian and Kasra Khosoussi and David M. Rosen and Jonathan P. How,
+   * 2020, Aug, Distributed Certifiably Correct Pose-Graph Optimization, Arxiv
+   * 3) C. de Sa, B. He, I. Mitliagkas, C. Ré, and P. Xu, “Accelerated
+   * stochastic power iteration,” in Proc. Mach. Learn. Res., no. 84, 2018, pp.
+   * 58–67
+   *
+   * It performs the following iteration: \f$ x_{k+1} = A * x_k + \beta *
+   * x_{k-1} \f$ where A is the certificate matrix, x is the Ritz vector
```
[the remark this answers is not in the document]
jingwuOUO: Renaming it to "aim matrix", or any other more suitable name?

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,133 @@
...
+
+ public:
+  // Constructor from aim matrix A (given as Matrix or Sparse), optional intial
+  // vector as ritzVector
+  explicit AcceleratedPowerMethod(
+      const Operator &A, const boost::optional<Vector> initial = boost::none)
+      : PowerMethod<Operator>(A, initial) {
+    Vector x0;
+    // initialize ritz vector
+    x0 = initial ? Vector::Random(this->dim_) : initial.get();
```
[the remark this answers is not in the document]
jingwuOUO: Yes. The previous order is wrong. Corrected it.

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,139 @@
...
+    Vector y = A_ * x;
+    y.normalize();
+    return y;
+  }
+
+  // Update the vector by dot product with A_
+  Vector update() const { return update(ritzVector_); }
+
+  // Perturb the initial ritzvector
+  void perturb() {
```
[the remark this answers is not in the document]
jingwuOUO: Moved it to ShonanAveraging.cpp

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,133 @@
...
+    this->ritzVector_ = update(previousVector_, x0, beta_);
+    this->perturb();
+
+    // set beta
+    Vector init_resid = this->ritzVector_;
+    const double up = init_resid.transpose() * this->A_ * init_resid;
+    const double down = init_resid.transpose().dot(init_resid);
+    const double mu = up / down;
+    beta_ = mu * mu / 4;
+    setBeta();
```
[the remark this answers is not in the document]
jingwuOUO: 1. Added beta argument in the constructor so that user can directly set the beta value. If there is no default value provide by the user, then it will call the `estimateBeta()` 
2. I don't understand why Shonan will already have a good idea about what beta should be. My understanding is the optimal beta will  be the {\lamda_2}^2/4 of the aim matrix, did we get the approximate value in previous power method? If yes, is it the  ritz value associated with the maximum eigenvector, because we assume here \lamda_1 is close to \lamda_2?

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,133 @@
...
+ public:
+  // Constructor from aim matrix A (given as Matrix or Sparse), optional intial
+  // vector as ritzVector
+  explicit AcceleratedPowerMethod(
+      const Operator &A, const boost::optional<Vector> initial = boost::none)
+      : PowerMethod<Operator>(A, initial) {
+    Vector x0;
+    // initialize ritz vector
+    x0 = initial ? Vector::Random(this->dim_) : initial.get();
+    Vector x00 = Vector::Random(this->dim_);
```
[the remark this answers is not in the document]
jingwuOUO: OK. Changed to this initialization. But actually I found if using this initialization strategy, the converged eigenVector becomes a little bit far way from the correct eigenVector in the unittest. Even when I increased the iteration upper limit a lot, the converged eigenVector does not change much.

Anyway, I just increased the error threshold in the `assert_equal(ev2, apf.eigenvector(), 3e-5)`  from 3e-5 to 4e-5.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,133 @@
...
+
+  // Update the ritzVector with beta and previous two ritzVector
+  Vector update() const {
+    Vector y = update(this->ritzVector_, previousVector_, beta_);
+    previousVector_ = this->ritzVector_;
+    return y;
+  }
+
+  // Tuning the momentum beta using the Best Heavy Ball algorithm in Ref(3)
+  void setBeta() {
```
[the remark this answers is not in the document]
jingwuOUO: Changed the name, added beta argument in the constructor so that the user can set it directly.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,133 @@
...
+    double maxBeta = beta_;
+    size_t maxIndex;
+    std::vector<double> betas = {2 / 3 * maxBeta, 0.99 * maxBeta, maxBeta,
```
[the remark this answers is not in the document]
jingwuOUO: Yes, your consideration is right. Have changed the code here to avoid the above mentioned situation.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,133 @@
...
+                                 1.01 * maxBeta, 1.5 * maxBeta};
+
+    Matrix R = Matrix::Zero(this->dim_, 10);
+    for (size_t i = 0; i < 10; i++) {
```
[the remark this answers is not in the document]
jingwuOUO: This is the *number of Iterations T* in the Best Heavy Ball algorithm. It means we need to do T times iteration to find the initial ritz vector that returns the largest Rayleigh quotient (But here I didn't randomly initialize this vector, have corrected this).

The paper did not mention how to set the T, so I use 10 here. But to be honest, I didn't find information regarding how to set this iteration time. Do you have idea? Or I think we can just skip the outer iteration since we've already have a `w` vector in the algorithm, which is also the initial ritz vector.

The algorithm is the B.3 in this [link](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6557638/)

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,133 @@
...
+            R.col(0) = update(x0, x00, betas[k]);
+            R.col(1) = update(R.col(0), x0, betas[k]);
+          } else {
+            R.col(j) = update(R.col(j - 1), R.col(j - 2), betas[k]);
+          }
+        }
+        const Vector x = R.col(9);
+        const double up = x.transpose() * this->A_ * x;
+        const double down = x.transpose().dot(x);
+        const double mu = up / down;
```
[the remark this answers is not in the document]
jingwuOUO: I refer to [this ](https://en.wikipedia.org/wiki/Rayleigh_quotient) to compute the Rayleigh quotient, so this is not correct? 

I changed it to 
`const double up = init_resid.dot( this->A_ * init_resid );
    const double down = init_resid.dot(init_resid);
    const double mu = up / down;
    double maxBeta = mu * mu / 4;`

gtsam/linear/tests/testAcceleratedPowerMethod.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <random>
+
+using namespace std;
+using namespace gtsam;
+using symbol_shorthand::X;
+
+/* ************************************************************************* */
+TEST(AcceleratedPowerMethod, acceleratedPowerIteration) {
+  // test power iteration, beta is set to 0
+  Sparse A(6, 6);
```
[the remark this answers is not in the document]
jingwuOUO: This is wrong......Have corrected it.

gtsam/linear/tests/testAcceleratedPowerMethod.cpp
```diff
@@ -0,0 +1,110 @@
...
+  A.coeffRef(0, 0) = 6;
+  A.coeffRef(0, 0) = 5;
+  A.coeffRef(0, 0) = 4;
+  A.coeffRef(0, 0) = 3;
+  A.coeffRef(0, 0) = 2;
+  A.coeffRef(0, 0) = 1;
+  Vector initial = Vector6::Zero();
+  const Vector6 x1 = (Vector(6) << 1.0, 0.0, 0.0, 0.0, 0.0, 0.0).finished();
```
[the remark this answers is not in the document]
jingwuOUO: Fixed

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -472,6 +473,57 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
...
+// in this algorithm:
+// (1) 
+static bool PowerMinimumEigenValue(
+    const Sparse &A, const Matrix &S, double *minEigenValue,
+    Vector *minEigenVector = 0, size_t *numIterations = 0,
+    size_t maxIterations = 1000,
+    double minEigenvalueNonnegativityTolerance = 10e-4) {
+
+  // a. Compute dominant eigenpair of S using power method
+  const boost::optional<Vector> initial(S.row(0));
```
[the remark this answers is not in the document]
jingwuOUO: Moved the `perturb()` in Shonan. And also only do the perturb for accelerated power method.

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -472,6 +473,57 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
...
+  PowerMethod<Sparse> lmOperator(A, initial);
+
```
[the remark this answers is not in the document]
jingwuOUO: Got it!

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -472,6 +473,57 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
...
+    *minEigenValue = lmEigenValue;
+    if (minEigenVector) {
+      *minEigenVector = lmOperator.eigenvectors();
+      minEigenVector->normalize();  // Ensure that this is a unit vector
+    }
+    return true;
+  }
+
+  const Sparse C = lmEigenValue * Matrix::Identity(A.rows(), A.cols()) - A;
+  AcceleratedPowerMethod<Sparse> minShiftedOperator(C, initial);
```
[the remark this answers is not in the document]
jingwuOUO: Yes.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -59,66 +60,68 @@ class AcceleratedPowerMethod : public PowerMethod<Operator> {
   // Constructor from aim matrix A (given as Matrix or Sparse), optional intial
   // vector as ritzVector
   explicit AcceleratedPowerMethod(
-      const Operator &A, const boost::optional<Vector> initial = boost::none)
+      const Operator &A, const boost::optional<Vector> initial = boost::none,
+      const double initialBeta = 0.0)
```
david-m-rosen, line 64: ```const``` not necessary here, since this is passing-by-copy

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -59,66 +60,68 @@ class AcceleratedPowerMethod : public PowerMethod<Operator> {
...
-    const double down = init_resid.transpose().dot(init_resid);
-    const double mu = up / down;
-    beta_ = mu * mu / 4;
-    setBeta();
+    this->ritzVector_ = initial ? initial.get() : Vector::Random(this->dim_);
+    this->ritzVector_.normalize();
+    previousVector_ = Vector::Zero(this->dim_);
+
+    // initialize beta_
+    if (!initialBeta) {
```
david-m-rosen, line 72: I still don't agree with having ```estimateBeta()``` here -- I **really** don't think it's a good idea to have a heavy-duty operation like ```estimateBeta()``` be a part of the constructor.

Notice, for example, that it would be impossible to construct an instance of this class for later use without that code immediately running

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -59,66 +60,68 @@ class AcceleratedPowerMethod : public PowerMethod<Operator> {
...
+      estimateBeta();
+    } else {
+      beta_ = initialBeta;
+    }
   }
 
-  // Update the ritzVector with beta and previous two ritzVector
-  Vector update(const Vector &x1, const Vector &x0, const double beta) const {
+  // Update the ritzVector with beta and previous two ritzVector, and return
```
david-m-rosen, line 79: This docstring **still** does not accurately describe what this function does!  [This function doesn't "**update**" ] anything!

gtsam/linear/tests/testAcceleratedPowerMethod.cpp
```diff
@@ -42,26 +40,24 @@ TEST(AcceleratedPowerMethod, acceleratedPowerIteration) {
...
-  EXPECT_LONGS_EQUAL(1, apf.eigenvectors().cols());
-  EXPECT_LONGS_EQUAL(6, apf.eigenvectors().rows());
+  apf.compute(50, 1e-4);
+  EXPECT_LONGS_EQUAL(6, apf.eigenvector().rows());
 
-  Vector6 actual1 = apf.eigenvectors();
-  // actual1(0) = abs (actual1(0));
-  EXPECT(assert_equal(x1, actual1));
+  Vector6 actual1 = apf.eigenvector();
+  EXPECT(assert_equal(x1, actual1, 1e-4));
```
david-m-rosen, line 58: You cannot check eigenvectors this way, because eigenvectors are **not** unique.  

Instead, you should check that the returned Ritz pair has a Ritz residual that satisfies the required tolerance, and that the Ritz value is within the Ritz tolerance of the maximum-magnitude eigenvalue

gtsam/linear/tests/testPowerMethod.cpp
```diff
@@ -42,23 +40,22 @@ TEST(PowerMethod, powerIteration) {
...
-  EXPECT_LONGS_EQUAL(1, pf.eigenvectors().cols());
-  EXPECT_LONGS_EQUAL(6, pf.eigenvectors().rows());
+  pf.compute(50, 1e-4);
+  EXPECT_LONGS_EQUAL(6, pf.eigenvector().rows());
 
   const Vector6 x1 = (Vector(6) << 1.0, 0.0, 0.0, 0.0, 0.0, 0.0).finished();
-  Vector6 actual0 = pf.eigenvectors();
-  EXPECT(assert_equal(x1, actual0));
+  Vector6 actual0 = pf.eigenvector();
+  EXPECT(assert_equal(x1, actual0, 1e-4));
```
david-m-rosen, line 55: Same as above: you cannot check eigenvectors this way

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -471,57 +471,86 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
...
+}
+
+/* ************************************************************************* */
+// Perturb the initial initialVector by adding a spherically-uniformly
+// distributed random vector with 0.03*||initialVector||_2 magnitude to
+// initialVector
+// ref : Part III. C, Rosen, D. and Carlone, L., 2017, September. Computational
+// enhancements for certifiably correct SLAM. In Proceedings of the
+// International Conference on Intelligent Robots and Systems.
+Vector perturb(const Vector &initialVector) {
```
david-m-rosen, line 488: This looks good, but I think it should also be ```const```?

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -471,57 +471,86 @@ Sparse ShonanAveraging<d>::computeA(const Values &values) const {
...
+// minEigenvalueNonnegativityTolerance is set to 1000 and 10e-4 ad default,
 // there are two parts
 // in this algorithm:
-// (1) 
+// (1) compute the maximum eigenpair (\lamda_dom, \vect{v}_dom) of A by power
+// method. if \lamda_dom is less than zero, then return the eigenpair. (2)
+// compute the maximum eigenpair (\theta, \vect{v}) of C = \lamda_dom * I - A by
+// accelerated power method. Then return (\lamda_dom - \theta, \vect{v}).
 static bool PowerMinimumEigenValue(
     const Sparse &A, const Matrix &S, double *minEigenValue,
```
david-m-rosen, line 512: OK this looks pretty good now, but I think it might be a *bit* cleaner to use ```double &``` instead of ```double*``` for ```minEigenValue``` and ```size_t& ``` instead of ```size_t*``` for ```numIterations```

(I like to avoid pointers whenever I can -- lots of creative ways to cause horrible segfaults with these :-P)

gtsam/linear/PowerMethod.h
```diff
@@ -46,94 +46,72 @@ class PowerMethod {
...
+  bool converged(double tol) {
     const Vector x = ritzVector_;
-    double theta = x.transpose() * A_ * x;
-
     // store the Ritz eigen value
-    ritzValue_ = theta;
-
-    const Vector diff = A_ * x - theta * x;
-    double error = diff.norm();
+    ritzValue_ = x.dot(A_ * x);
```
david-m-rosen, line 84: This line is **changing** ```ritzValue_```  -- simple test functions like this should **not** cause side-effects!

You should change this to be a temporary value, and make this function ```const```.

gtsam/linear/tests/testAcceleratedPowerMethod.cpp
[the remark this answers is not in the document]
david-m-rosen: It might be better here to make ```initial``` a hard-coded vector instead of a random sample -- this would make it easier to reproduce failing tests at different times, different machines, etc.

gtsam/linear/tests/testPowerMethod.cpp
```diff
@@ -42,23 +40,22 @@ TEST(PowerMethod, powerIteration) {
...
-  A.coeffRef(0, 0) = 3;
-  A.coeffRef(0, 0) = 2;
-  A.coeffRef(0, 0) = 1;
-  Vector initial = Vector6::Zero();
+  A.coeffRef(1, 1) = 5;
+  A.coeffRef(2, 2) = 4;
+  A.coeffRef(3, 3) = 3;
+  A.coeffRef(4, 4) = 2;
+  A.coeffRef(5, 5) = 1;
+  Vector initial = Vector6::Random();
```
david-m-rosen, line 48: Same as above -- I think it would be better to make this a hard-coded vector

gtsam/linear/PowerMethod.h
david-m-rosen: @dellaert   What's the question being asked here?

david-m-rosen (changes requested):
OK, still a few things that need to be fixed, but I think this is getting pretty close now :-)

gtsam/linear/AcceleratedPowerMethod.h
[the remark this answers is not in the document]
jingwuOUO: Yes. I should follow this naming style.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -59,66 +60,68 @@ class AcceleratedPowerMethod : public PowerMethod<Operator> {
...
+                        const double beta) const {
     Vector y = this->A_ * x1 - beta * x0;
     y.normalize();
     return y;
   }
 
-  // Update the ritzVector with beta and previous two ritzVector
-  Vector update() const {
-    Vector y = update(this->ritzVector_, previousVector_, beta_);
+  // Update the ritzVector with beta and previous two ritzVector, and return
```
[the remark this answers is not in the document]
jingwuOUO: Renamed this func and also added more detailed documentation.

gtsam/linear/AcceleratedPowerMethod.h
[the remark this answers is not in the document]
jingwuOUO: I see. Changed it.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,136 @@
...
+    Vector y = this->A_ * x1 - beta * x0;
+    y.normalize();
+    return y;
+  }
+
+  // Update the ritzVector with beta and previous two ritzVector, and return
+  // x_{k+1} = A * x_k - \beta * x_{k-1} / || A * x_k - \beta * x_{k-1} ||
+  Vector powerIteration() const {
+    Vector y = powerIteration(this->ritzVector_, previousVector_, beta_);
+    previousVector_ = this->ritzVector_;
```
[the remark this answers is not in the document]
jingwuOUO: I've changed the update part outside this function, and have changed the documentation. This func only performs accelerated power iteration, updating is outside this func.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,136 @@
...
+    return y;
+  }
+
+  // Tuning the momentum beta using the Best Heavy Ball algorithm in Ref(3)
```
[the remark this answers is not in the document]
jingwuOUO: That's really good suggestion! Have added more illustration in the code so that others can understand this algorithm faster and easier.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,136 @@
...
+            R.col(j) = powerIteration(R.col(j - 1), R.col(j - 2), betas[k]);
+          }
+        }
+        const Vector x = R.col(9);
+        const double up = x.dot(this->A_ * x);
+        const double down = x.dot(x);
+        const double mu = up / down;
+        if (mu * mu / 4 > maxBeta) {
+          maxIndex = k;
+          maxBeta = mu * mu / 4;
```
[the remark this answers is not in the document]
jingwuOUO: Yes, I should reset the beta according to the maxBeta after each t iteration

gtsam/linear/PowerMethod.h
```diff
@@ -46,94 +46,72 @@ class PowerMethod {
...
     ritzVector_ = Vector::Zero(dim_);
-
-    ritzVector_.col(0) = update(x0);
-    perturb();
+    ritzVector_ = powerIteration(x0);
   }
 
-  // Update the vector by dot product with A_
-  Vector update(const Vector &x) const {
+  // Update the vector by dot product with A_, and return A * x / || A * x ||
```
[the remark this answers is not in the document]
jingwuOUO: Changed the update description. Thanks for pointing that out. I searched and found this should be called matrix-vector product?

gtsam/linear/PowerMethod.h
```diff
@@ -46,94 +46,72 @@ class PowerMethod {
...
-    ritzVector_ = x0 + 0.03 * magnitude * disturb;
-  }
+  //  Update the vector by dot product with A_, and return A * x / || A * x ||
+  Vector powerIteration() const { return powerIteration(ritzVector_); }
 
-  // Perform power iteration on a single Ritz value
-  // Updates ritzValue_
-  bool iterateOne(double tol) {
+  // After Perform power iteration on a single Ritz value, if the error is less
+  // than the tol then return true else false
```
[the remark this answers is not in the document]
jingwuOUO: Ok, this is more clear

gtsam/linear/PowerMethod.h
```diff
@@ -46,94 +46,72 @@ class PowerMethod {
...
     // Starting
-    int nrConverged = 0;
+    bool isConverged = false;
 
-    for (int i = 0; i < maxIterations; i++) {
-      nrIterations_ += 1;
-      ritzVector_ = update();
-      nrConverged = iterateOne(tol);
-      if (nrConverged) break;
+    for (size_t i = 0; i < maxIterations; i++) {
```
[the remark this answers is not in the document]
jingwuOUO: Thanks! this is more compact.

gtsam/linear/tests/testAcceleratedPowerMethod.cpp
jingwuOUO: OK. That also solve my previous question about the increasing error in comparing the eigenvector after changed the initialization ritzVector.

gtsam/sfm/ShonanAveraging.cpp
jingwuOUO: Non-member function can't add const

gtsam/linear/tests/testPowerMethod.cpp
jingwuOUO: Good suggestion!

gtsam/linear/tests/testAcceleratedPowerMethod.cpp
jingwuOUO: About this, I have question maybe I don't fully understand. 
Because we're computing the maximum eigenvalue and its according eigenvector. Suppose there is no double or more root of this maximum eigenvalue(the input matrix in unittest does not have), and the ritzVector is always normalized, then isn't the eigenvector here only have two form, v and -v? So that's why I set the unittest to compare the eigenvector.

gtsam/sfm/ShonanAveraging.cpp
jingwuOUO: I changed this part. I'm not 100% sure if there is further consideration to use `double*` than `double &` and `size_t` because this form follows previous code by @dellaert . Could you take a look?

david-m-rosen (changes requested):
A few final edits (all minor at this point), and I think this will be ready to go.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -59,66 +60,68 @@ class AcceleratedPowerMethod : public PowerMethod<Operator> {
...
-    const double mu = up / down;
-    beta_ = mu * mu / 4;
-    setBeta();
+    this->ritzVector_ = initial ? initial.get() : Vector::Random(this->dim_);
+    this->ritzVector_.normalize();
+    previousVector_ = Vector::Zero(this->dim_);
+
+    // initialize beta_
+    if (!initialBeta) {
+      estimateBeta();
```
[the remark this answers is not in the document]
david-m-rosen: Update: I still think ```estimateBeta()``` should not be called here ...

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -76,61 +76,97 @@ class AcceleratedPowerMethod : public PowerMethod<Operator> {
     }
   }
 
-  // Update the ritzVector with beta and previous two ritzVector, and return
-  // x_{k+1} = A * x_k - \beta * x_{k-1} / || A * x_k - \beta * x_{k-1} ||
-  Vector powerIteration(const Vector &x1, const Vector &x0,
+  // Run accelerated power iteration on the ritzVector with beta and previous
```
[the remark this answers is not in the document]
jingwuOUO: Also added parenthesis in numerator.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -76,61 +76,97 @@ class AcceleratedPowerMethod : public PowerMethod<Operator> {
...
+  // two ritzVector, and return x_{k+1} = A * x_k - \beta * x_{k-1} / || A * x_k
+  // - \beta * x_{k-1} ||
+  Vector acceleratedPowerIteration () const {
+    Vector y = acceleratedPowerIteration(this->ritzVector_, previousVector_, beta_);
     return y;
   }
 
   // Tuning the momentum beta using the Best Heavy Ball algorithm in Ref(3)
   void estimateBeta() {
     // set beta
```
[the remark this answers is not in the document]
jingwuOUO: Changed the comment.

gtsam/linear/tests/testPowerMethod.cpp
```diff
@@ -0,0 +1,105 @@
...
+  // Store the max eigenvalue and its according eigenvector
+  const auto ev1 = solver.eigenvalues()(maxIdx).real();
+  auto ev2 = solver.eigenvectors().col(maxIdx).real();
+
+  Vector initial = Vector4::Random();
+  PowerMethod<Matrix> pf(L.first, initial);
+  pf.compute(50, 1e-4);
+  EXPECT_DOUBLES_EQUAL(ev1, pf.eigenvalue(), 1e-8);
+  auto actual2 = pf.eigenvector();
+  EXPECT(assert_equal(ev2, actual2, 3e-5));
```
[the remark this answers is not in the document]
jingwuOUO: About why using this unittest, the idea comes from that we want to use the graph.hessian() later to extract the A matrix. There are some tests using the EigenSolver from Eigen to see whether the eigenpair result is always "constant" and also to compare the results of power/accelerated method at initial (but now more clear about the correctness of power method). 
But I'm still a little confused here, for the specific matrix A in this unittest, there is no two eigenvectors for one eigenvalue, and the eigenvector is normalized, the only uncertainty here is the sign of the  eigenvector, whether +/-. So I guess it's fine to keep this. I'm not sure. I'm ok with deleting the checking specific eigenvector part.

jingwuOUO:
Updated according to last review by David.

Something I'm not sure about. I see previous `SparseMinimumEigenValue` is not included in Shonan, so I did the same with `PowerMinimumEigenValue`, not sure why. Do we need to add `PowerMinimumEigenValue` func into Shonan so that I can also add `purturb` func in Shonan. @dellaert

gtsam/linear/tests/testPowerMethod.cpp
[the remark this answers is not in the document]
david-m-rosen: The problem with:

```   EXPECT(assert_equal(ev2, actual2, 3e-5)); ```

is that you are trying to enforce the condition that the (unit) eigenvector ```actual2``` returned by the power method is *exactly* equal to the eigenvector ```ev2``` returned by Eigen.  But there are **always** _at least 2_ unit eigenvectors associated with any given eigenvalue, and in general there will be _infinitely many_ (if that eigenvalue happens to have multiplicity > 1).

So the question here is: why would we _ever_ expect the power method to return _that specific_ eigenvector?

The answer is: we _wouldn't_, because eigenvectors are **never unique**.  And if they're not unique, we shouldn't be trying to enforce uniqueness in a unit test.  This doesn't make mathematical sense.

The correct test of "eigenvector-ness" is the eigenvector equation ```A*x = lambda*x```.  Specifically, we should be checking something like a normalized Ritz residual:

```eps := ||A*x  - lambda*x|| / ||x||```

Note that this error measure does **not** depend upon *which specific* eigenvector we get, since *any* eigenvector ```x``` for eigenvalue ```lambda``` would have ```eps = 0```.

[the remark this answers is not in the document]
Comment (jingwuOUO) [reply gtsam/linear/tests/testPowerMethod.cpp]: Got it!

File: gtsam/linear/AcceleratedPowerMethod.h [hunk 68]
```diff
@@ -0,0 +1,171 @@
...
+
+using Sparse = Eigen::SparseMatrix<double>;
+
+/* ************************************************************************* */
+/// MINIMUM EIGENVALUE COMPUTATIONS
+
+// Template argument Operator just needs multiplication operator
+template <class Operator>
+class AcceleratedPowerMethod : public PowerMethod<Operator> {
+  /**
```
dellaert, line 36: doxygen comments should go right above class. Incorporate template argument doc there.

gtsam/linear/AcceleratedPowerMethod.h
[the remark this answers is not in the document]
dellaert: @jingwuOUO is there a reason `estimateBeta` can't be moved out as a static function??

gtsam/linear/AcceleratedPowerMethod.h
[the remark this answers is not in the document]
dellaert: I still see a mismatch between doc and argument names. Also, should be a doxygen comment, /** */

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,171 @@
...
+  // two ritzVector, and return y = (A * x0 - \beta * x00) / || A * x0
+  // - \beta * x00 ||
+  Vector acceleratedPowerIteration (const Vector &x1, const Vector &x0,
+                        const double beta) const {
+    Vector y = this->A_ * x1 - beta * x0;
+    y.normalize();
+    return y;
+  }
+
+  // Run accelerated power iteration on the ritzVector with beta and previous
```
dellaert, line 89: make doxygen comment. And what do x0 and x00 refer to here. Clarify by using a name and symbol together, e.g. the first ritzvector r0 and second ritzvector r1

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,171 @@
...
+  // two ritzVector, and return y = (A * x0 - \beta * x00) / || A * x0
+  // - \beta * x00 ||
+  Vector acceleratedPowerIteration () const {
+    Vector y = acceleratedPowerIteration(this->ritzVector_, previousVector_, beta_);
+    return y;
+  }
+
+  // Tuning the momentum beta using the Best Heavy Ball algorithm in Ref(3)
+  void estimateBeta() {
```
dellaert, line 98: I propose to make this a const method and return beta.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -76,61 +76,97 @@ class AcceleratedPowerMethod : public PowerMethod<Operator> {
...
-    std::vector<double> betas = {2 / 3 * maxBeta, 0.99 * maxBeta, maxBeta,
-                                 1.01 * maxBeta, 1.5 * maxBeta};
+    std::vector<double> betas;
 
     Matrix R = Matrix::Zero(this->dim_, 10);
-    for (size_t i = 0; i < 10; i++) {
-      Vector x0 = Vector::Random(this->dim_);
-      x0.normalize();
-      Vector x00 = Vector::Zero(this->dim_);
+    size_t T = 10;
```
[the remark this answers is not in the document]
dellaert: Please do?

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,120 @@
...
+#include <vector>
+
+namespace gtsam {
+
+using Sparse = Eigen::SparseMatrix<double>;
+
+/* ************************************************************************* */
+/// MINIMUM EIGENVALUE COMPUTATIONS
+
+// Template argument Operator just needs multiplication operator
```
dellaert, line 37: Document this class with doxygen comment, include this comment as well of course

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,120 @@
...
+  size_t nrIterations_;  // number of iterations
+
+  double ritzValue_;   // Ritz eigenvalue
+  Vector ritzVector_;  // Ritz eigenvector
+
+ public:
+  // Constructor
+  explicit PowerMethod(const Operator &A,
+                       const boost::optional<Vector> initial = boost::none)
+      : A_(A), dim_(A.rows()), nrIterations_(0) {
```
dellaert, line 56: Does .rows() work with the factor graph unit tests?

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,120 @@
...
+    Vector x0;
+    x0 = initial ? initial.get() : Vector::Random(dim_);
+    x0.normalize();
+
+    // initialize Ritz eigen value
+    ritzValue_ = 0.0;
+
+    // initialize Ritz eigen vector
+    ritzVector_ = Vector::Zero(dim_);
```
dellaert, line 65: is this not redundant?

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,120 @@
...
+    ritzVector_ = powerIteration(x0);
+  }
+
+  // Run power iteration on the vector, and return A * x / || A * x ||
```
dellaert, line 69: Single-line doxygen is OK but starts with ///

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,120 @@
...
+  Vector powerIteration(const Vector &x) const {
+    Vector y = A_ * x;
+    y.normalize();
+    return y;
+  }
+
+  // Run power iteration on the vector, and return A * x / || A * x ||
```
dellaert, line 76: same. Also, what is x?

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,120 @@
...
+  Vector powerIteration() const { return powerIteration(ritzVector_); }
+
+  // After Perform power iteration on a single Ritz value, check if the Ritz
```
dellaert, line 79: Make doxygen.

gtsam/linear/tests/testAcceleratedPowerMethod.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  Vector6 actual1 = apf.eigenvectors();
+  // actual1(0) = abs (actual1(0));
+  EXPECT(assert_equal(x1, actual1));
+
+  EXPECT_DOUBLES_EQUAL(ev1, apf.eigenvalues(), 1e-5);
+}
+
+/* ************************************************************************* */
+TEST(AcceleratedPowerMethod, useFactorGraph) {
```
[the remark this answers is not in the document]
dellaert: It's a test for factor graph operator, at least that part should be tested somewhere. We should have tests for a sparse and dense matrix, as well. Simplicity would be good.

gtsam/linear/tests/testPowerMethod.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <random>
+
+using namespace std;
+using namespace gtsam;
+using symbol_shorthand::X;
+
+/* ************************************************************************* */
+TEST(PowerMethod, powerIteration) {
+  // test power iteration, beta is set to 0
+  Sparse A(6, 6);
```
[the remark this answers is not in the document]
dellaert: What should be the logic? Maybe we should meet about this

dellaert (changes requested):
Still some comments by David that have not been addressed. Also, please read for 5 minutes about doxygen and then please apply that to entire PR.

gtsam/linear/PowerMethod.h
jingwuOUO: Now using fg.hessian().first (which is also Matrix type), this works.

jingwuOUO:
So we need to add unittest for sparse and dense matrix, can I create some sparse and dense matrix myself? And then add it in these two test file.

dellaert:
> So we need to add unittest for sparse and dense matrix, can I create some sparse and dense matrix myself? And then add it in these two test file.

Addressed in meeting. Please re-request review.

gtsam/linear/tests/testPowerMethod.cpp
```diff
@@ -0,0 +1,142 @@
...
+  const double ritzResidual = (A * actual1 - ritzValue * actual1).norm();
+  EXPECT_DOUBLES_EQUAL(0, ritzResidual, 1e-5);
+
+  const double ev1 = 6.0;
+  EXPECT_DOUBLES_EQUAL(ev1, pf.eigenvalue(), 1e-5);
+}
+
+/* ************************************************************************* */
+TEST(PowerMethod, useFactorGraphSparse) {
+  // Let's make a scalar synchronization graph with 4 nodes
```
dellaert, line 65: I like your two new examples ! It's clear you read my blog post :-) But, the issue is it's still copy/paste. Could you add them to a header file in this directory? See smallExample.h as an example.

gtsam/linear/PowerMethod.h
```diff
@@ -0,0 +1,150 @@
...
+
+namespace gtsam {
+
+using Sparse = Eigen::SparseMatrix<double>;
+
+/**
+ * \brief Compute maximum Eigenpair with power method
+ *
+ * References :
+ * 1) Rosen, D. and Carlone, L., 2017, September. Computational
```
dellaert, line 38: Note sure this reference is appropriate for power method. More for accelerated. Power method itself is as old as the street.

gtsam/linear/AcceleratedPowerMethod.h
```diff
@@ -0,0 +1,174 @@
...
+   * two ritzVector x0 and x00, and return y = (A * x0 - \beta * x00) / || A * x0
+   * - \beta * x00 ||
+   */
+  Vector acceleratedPowerIteration () const {
+    Vector y = acceleratedPowerIteration(this->ritzVector_, previousVector_, beta_);
+    return y;
+  }
+
+  /// Tuning the momentum beta using the Best Heavy Ball algorithm in Ref(3)
+  double estimateBeta() const {
```
dellaert, line 99: Explain again why this is still a method? I thought David had asked to make this a free function or at least a static method.

gtsam/sfm/ShonanAveraging.cpp
dellaert: Actually, if it's a local function this can just be made static in this file.

dellaert (approved):
I will approve but trust you to address the final few comments I have.

gtsam/linear/PowerMethod.h
jingwuOUO: Added _Matrix Computation_ as reference and noted pages, which is also cited by Yunlun_arxiv paper

gtsam/linear/AcceleratedPowerMethod.h
david-m-rosen: @dellaert @jingwuOUO I think it's fine to have this as a method (in fact, probably necessary, since we need access to the matrix-vector multiplication operator that's passed in during construction).  What I **don't** want is this function being called inside the constructor (as currently happens in line 71), since this is a heavy-duty operation that is not actually required in order to construct an instance of this object -- permitting a default value of beta = 0.0 is perfectly fine, and I think arguably the "right" default behavior, since that will exactly correspond to the usual (simple) power iteration

gtsam/sfm/ShonanAveraging.cpp
david-m-rosen: OK, I don't really have strong feelings about this one way or the other -- happy to resolve this as-is if you are @dellaert?

david-m-rosen:
Hi @jingwuOUO, the only changes I see that remain to be done here are:

-  Adding the number of iterations T used in estimateBeta() as an (optional) argument (with default value e.g. 10)
- Not calling estimateBeta() from within the constructor for the accelerated power method, since this is a heavy-duty method.  It would be better to allow the user to specify this directly, as an optional argument with a default value of 0 (corresponding to a standard power iteration).

dellaert:
@jingwuOUO pls fix last two issues so we can finally merge?

PR #535, opened 2020-09-22
Wrapping MFAS and 1dsfm python example

akshay-krishnan:
This PR provides a python wrapper to the MFAS class. 
It adds an example that shows how 1dsfm uses MFAS and translation recovery together to estimate global translations from unit translation measurements given global rotations. 
There is also a small change to the MFAS API. The unused shared_ptr member has been removed.

### DIFF
--- a/gtsam/sfm/MFAS.cpp
+++ b/gtsam/sfm/MFAS.cpp
@@ -7,8 +7,6 @@
 
 #include <gtsam/sfm/MFAS.h>
 
-#include <boost/shared_ptr.hpp>
-
 #include <algorithm>
 #include <map>
 #include <unordered_map>
--- a/gtsam/sfm/MFAS.cpp
+++ b/gtsam/sfm/MFAS.cpp
@@ -113,10 +113,8 @@ void removeNodeFromGraph(const Key node,
   graph.erase(node);
 }
 
-MFAS::MFAS(const boost::shared_ptr<vector<Key>> nodes,
-           const TranslationEdges& relativeTranslations,
-           const Unit3& projectionDirection)
-    : nodes_(nodes) {
+MFAS::MFAS(const TranslationEdges& relativeTranslations,
+           const Unit3& projectionDirection) {
   // Iterate over edges, obtain weights by projecting
   // their relativeTranslations along the projection direction
   for (const auto& measurement : relativeTranslations) {
--- a/gtsam/sfm/MFAS.cpp
+++ b/gtsam/sfm/MFAS.cpp
@@ -7,6 +7,8 @@
 
 #include <gtsam/sfm/MFAS.h>
 
+#include <boost/shared_ptr.hpp>
+
 #include <algorithm>
 #include <map>
 #include <unordered_map>
@@ -111,7 +113,7 @@ void removeNodeFromGraph(const Key node,
   graph.erase(node);
 }
 
-MFAS::MFAS(const std::shared_ptr<vector<Key>>& nodes,
+MFAS::MFAS(const boost::shared_ptr<vector<Key>> nodes,
            const TranslationEdges& relativeTranslations,
            const Unit3& projectionDirection)
     : nodes_(nodes) {
--- a/gtsam/sfm/tests/testMFAS.cpp
+++ b/gtsam/sfm/tests/testMFAS.cpp
@@ -9,8 +9,6 @@
 
 #include <CppUnitLite/TestHarness.h>
 
-#include <iostream>
-
 using namespace std;
 using namespace gtsam;
 
--- a/gtsam/sfm/tests/testMFAS.cpp
+++ b/gtsam/sfm/tests/testMFAS.cpp
@@ -6,8 +6,12 @@
  */
 
 #include <gtsam/sfm/MFAS.h>
-#include <iostream>
+
+#include <boost/shared_ptr.hpp>
+#include <boost/make_shared.hpp>
 #include <CppUnitLite/TestHarness.h>
+
+#include <iostream>
 
 using namespace std;
 using namespace gtsam;
@@ -46,7 +50,7 @@ map<MFAS::KeyPair, double> getEdgeWeights(const vector<MFAS::KeyPair> &edges,
 // test the ordering and the outlierWeights function using weights2 - outlier
 // edge is rejected when projected in a direction that gives weights2
 TEST(MFAS, OrderingWeights2) {
-  MFAS mfas_obj(make_shared<vector<Key>>(nodes), getEdgeWeights(edges, weights2));
+  MFAS mfas_obj(boost::make_shared<vector<Key>>(nodes), getEdgeWeights(edges, weights2));
 
   vector<Key> ordered_nodes = mfas_obj.computeOrdering();
 
@@ -76,7 +80,7 @@ TEST(MFAS, OrderingWeights2) {
 // weights1 (outlier edge is accepted when projected in a direction that
 // produces weights1)
 TEST(MFAS, OrderingWeights1) {
-  MFAS mfas_obj(make_shared<vector<Key>>(nodes), getEdgeWeights(edges, weights1));
+  MFAS mfas_obj(boost::make_shared<vector<Key>>(nodes), getEdgeWeights(edges, weights1));
 
   vector<Key> ordered_nodes = mfas_obj.computeOrdering();
 
--- a/gtsam/sfm/tests/testMFAS.cpp
+++ b/gtsam/sfm/tests/testMFAS.cpp
@@ -7,8 +7,6 @@
 
 #include <gtsam/sfm/MFAS.h>
 
-#include <boost/shared_ptr.hpp>
-#include <boost/make_shared.hpp>
 #include <CppUnitLite/TestHarness.h>
 
 #include <iostream>
@@ -43,14 +41,13 @@ map<MFAS::KeyPair, double> getEdgeWeights(const vector<MFAS::KeyPair> &edges,
   for (size_t i = 0; i < edges.size(); i++) {
     edgeWeights[edges[i]] = weights[i];
   }
-  cout << "returning edge weights " << edgeWeights.size() << endl;
   return edgeWeights;
 }
 
 // test the ordering and the outlierWeights function using weights2 - outlier
 // edge is rejected when projected in a direction that gives weights2
 TEST(MFAS, OrderingWeights2) {
-  MFAS mfas_obj(boost::make_shared<vector<Key>>(nodes), getEdgeWeights(edges, weights2));
+  MFAS mfas_obj(getEdgeWeights(edges, weights2));
 
   vector<Key> ordered_nodes = mfas_obj.computeOrdering();
 
@@ -80,7 +77,7 @@ TEST(MFAS, OrderingWeights2) {
 // weights1 (outlier edge is accepted when projected in a direction that
 // produces weights1)
 TEST(MFAS, OrderingWeights1) {
-  MFAS mfas_obj(boost::make_shared<vector<Key>>(nodes), getEdgeWeights(edges, weights1));
+  MFAS mfas_obj(getEdgeWeights(edges, weights1));
 
   vector<Key> ordered_nodes = mfas_obj.computeOrdering();
 
--- a/python/gtsam/examples/TranslationAveragingExample.py
+++ b/python/gtsam/examples/TranslationAveragingExample.py
@@ -32,8 +32,8 @@ def get_data() -> Tuple[gtsam.Values, List[gtsam.BinaryMeasurementUnit3]]:
     that lie on a circle and face the center. The poses of 8 cameras are obtained from SFMdata
     and the unit translations directions between some camera pairs are computed from their
     global translations. """
-    # Using toy dataset in SfMdata for example.
-    wTc_list = SFMdata.createPoses(gtsam.Cal3_S2(50.0, 50.0, 0.0, 50.0, 50.0))
+    fx, fy, s, u0, v0 = 50.0, 50.0, 0.0, 50.0, 50.0
+    wTc_list = SFMdata.createPoses(gtsam.Cal3_S2(fx, fy, s, u0, v0))
     # Rotations of the cameras in the world frame.
     wRc_values = gtsam.Values()
     # Normalized translation directions from camera i to camera j
@@ -88,10 +88,10 @@ def prune_to_inliers(w_iZj_list: gtsam.BinaryMeasurementsUnit3) -> gtsam.BinaryM
         for keypair, weight in outlier_weight_dict.items():
             avg_outlier_weights[keypair] += weight / len(outlier_weights)
 
-    # Remove w_relative_tranlsations that have weight greater than threshold, these are outliers.
+    # Remove w_iZj that have weight greater than threshold, these are outliers.
     w_iZj_inliers = gtsam.BinaryMeasurementsUnit3()
-    [w_iZj_inliers.append(Z) for Z in w_iZj_list if avg_outlier_weights[(
-        Z.key1(), Z.key2())] < OUTLIER_WEIGHT_THRESHOLD]
+    [w_iZj_inliers.append(Z) for w_iZj in w_iZj_list if avg_outlier_weights[(
+        w_iZj.key1(), w_iZj.key2())] < OUTLIER_WEIGHT_THRESHOLD]
 
     return w_iZj_inliers
 
@@ -108,7 +108,7 @@ def estimate_poses(i_iZj_list: gtsam.BinaryMeasurementsUnit3,
         wRc_values: Rotations of the cameras in the world frame.
 
     Returns:
-        Values: Estimated poses.
+        gtsam.Values: Estimated poses.
     """
 
     # Convert the translation direction measurements to world frame using the rotations.
--- a/python/gtsam/examples/TranslationAveragingExample.py
+++ b/python/gtsam/examples/TranslationAveragingExample.py
@@ -89,7 +89,7 @@ def estimate_poses(relative_translations: gtsam.BinaryMeasurementsUnit3,
 
     # Compute average of outlier weights. Each outlier weight is a map from a pair of Keys (camera IDs) to a weight,
     # where weights are proportional to the probability of the edge being an outlier.
-    avg_outlier_weights = defaultdict(lambda: 0.0)
+    avg_outlier_weights = defaultdict(float)
     for outlier_weight_dict in outlier_weights:
         for keypair, weight in outlier_weight_dict.items():
             avg_outlier_weights[keypair] += weight / len(outlier_weights)
--- a/python/gtsam/examples/TranslationAveragingExample.py
+++ b/python/gtsam/examples/TranslationAveragingExample.py
@@ -1,0 +1,93 @@
+from collections import Counter
+import functools
+import operator
+
+import numpy as np
+
+import gtsam
+from gtsam.examples import SFMdata
+
+max_1dsfm_projection_directions = 50
+outlier_weight_threshold = 0.1
+
+def get_data():
+    """"Returns data from SfMData.createPoses(). This contains the global rotations and the unit translations directions."""
+    # Using toy dataset in SfMdata for example.
+    poses = SFMdata.createPoses(gtsam.Cal3_S2(50.0, 50.0, 0.0, 50.0, 50.0))
+    rotations = gtsam.Values()
+    translation_directions = []
+    for i in range(0, len(poses) - 2):
+        # Add the rotation
+        rotations.insert(i, poses[i].rotation())
+        # Create unit translation measurements with next two poses
+        for j in range(i+1, i+3):
+            i_Z_j = gtsam.Unit3(poses[i].rotation().unrotate(poses[j].translation() - poses[i].translation()))
+            translation_directions.append(gtsam.BinaryMeasurementUnit3(
+                i, j, i_Z_j, gtsam.noiseModel.Isotropic.Sigma(3, 0.01)))
+    # Add the last two rotations.
+    rotations.insert(len(poses) - 1, poses[-1].rotation())
+    rotations.insert(len(poses) - 2, poses[-2].rotation())
+    return (rotations, translation_directions)
+
+
+def estimate_poses_given_rot(measurements: gtsam.BinaryMeasurementsUnit3,
+                             rotations: gtsam.Values):
+    """Estimate poses given normalized translation directions and rotations between nodes.
+
+    Arguments:
+        measurements - List of translation direction from the first node to the second node in the coordinate frame of the first node.
+        rotations {Values} -- Estimated rotations
+
+    Returns:
+        Values -- Estimated poses.
+    """
+
+    # Convert the translation directions to global frame using the rotations.
+    w_measurements = gtsam.BinaryMeasurementsUnit3()
+    for measurement in measurements:
+        w_measurements.append(gtsam.BinaryMeasurementUnit3(measurement.key1(), measurement.key2(
+        ), gtsam.Unit3(rotations.atRot3(measurement.key1()).rotate(measurement.measured().point3())), measurement.noiseModel()))
+
+    # Indices of measurements that are to be used as projection directions. These are randomly chosen.
+    indices = np.random.choice(len(w_measurements), min(
+        max_1dsfm_projection_directions, len(w_measurements)), replace=False)
+    # Sample projection directions from the measurements.
+    projection_directions = [w_measurements[idx].measured() for idx in indices]
+    
+    outlier_weights = []
+    # Find the outlier weights for each direction using MFAS.
+    for direction in projection_directions:
+        algorithm = gtsam.MFAS(w_measurements, direction)
+        outlier_weights.append(algorithm.computeOutlierWeights())
+
+    # Compute average of outlier weights.
+    avg_outlier_weights = {}
+    for outlier_weight_dict in outlier_weights:
+        for k, v in outlier_weight_dict.items():
+            if k in avg_outlier_weights:
+                avg_outlier_weights[k] += v/len(outlier_weights)
+            else:
+                avg_outlier_weights[k] = v/len(outlier_weights)
+
+    # Remove measurements that have weight greater than threshold.
+    inlier_measurements = gtsam.BinaryMeasurementsUnit3()
+    [inlier_measurements.append(m) for m in w_measurements if avg_outlier_weights[(m.key1(), m.key2())] < outlier_weight_threshold]
+
+    # Run the optimizer to obtain translations for normalized directions.
+    translations = gtsam.TranslationRecovery(inlier_measurements).run()
+
+    poses = gtsam.Values()
+    for key in rotations.keys():
+        poses.insert(key, gtsam.Pose3(
+            rotations.atRot3(key), translations.atPoint3(key)))
+    return poses
+
+def main():
+    rotations, translation_directions = get_data()
+    poses = estimate_poses_given_rot(translation_directions, rotations)
+    print("**** Translation averaging output ****")
+    print(poses)
+    print("**************************************")
+
+if __name__ == '__main__':
+    main()
--- a/python/gtsam/examples/TranslationAveragingExample.py
+++ b/python/gtsam/examples/TranslationAveragingExample.py
@@ -90,7 +90,7 @@ def prune_to_inliers(w_iZj_list: gtsam.BinaryMeasurementsUnit3) -> gtsam.BinaryM
 
     # Remove w_iZj that have weight greater than threshold, these are outliers.
     w_iZj_inliers = gtsam.BinaryMeasurementsUnit3()
-    [w_iZj_inliers.append(Z) for w_iZj in w_iZj_list if avg_outlier_weights[(
+    [w_iZj_inliers.append(w_iZj) for w_iZj in w_iZj_list if avg_outlier_weights[(
         w_iZj.key1(), w_iZj.key2())] < OUTLIER_WEIGHT_THRESHOLD]
 
     return w_iZj_inliers
--- a/python/gtsam/examples/TranslationAveragingExample.py
+++ b/python/gtsam/examples/TranslationAveragingExample.py
@@ -59,7 +59,7 @@ def get_data() -> Tuple[gtsam.Values, List[gtsam.BinaryMeasurementUnit3]]:
     return wRc_values, i_iZj_list
 
 
-def prune_to_inliers(w_iZj_list: gtsam.BinaryMeasurementsUnit3) -> gtsam.BinaryMeasurementsUnit3:
+def filter_outliers(w_iZj_list: gtsam.BinaryMeasurementsUnit3) -> gtsam.BinaryMeasurementsUnit3:
     """Removes outliers from a list of Unit3 measurements that are the 
     translation directions from camera i to camera j in the world frame."""
 
@@ -120,7 +120,7 @@ def estimate_poses(i_iZj_list: gtsam.BinaryMeasurementsUnit3,
             i_iZj.key1(), i_iZj.key2(), w_iZj, i_iZj.noiseModel()))
 
     # Remove the outliers in the unit translation directions.
-    w_iZj_inliers = prune_to_inliers(w_iZj_list)
+    w_iZj_inliers = filter_outliers(w_iZj_list)
 
     # Run the optimizer to obtain translations for normalized directions.
     wtc_values = gtsam.TranslationRecovery(w_iZj_inliers).run()
--- a/python/gtsam/examples/TranslationAveragingExample.py
+++ b/python/gtsam/examples/TranslationAveragingExample.py
@@ -1,17 +1,27 @@
-from collections import Counter
-import functools
-import operator
+"""
+GTSAM Copyright 2010-2018, Georgia Tech Research Corporation,
+Atlanta, Georgia 30332-0415
+All Rights Reserved
+Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+See LICENSE for the license information
+
+This example shows how 1dsfm uses outlier rejection (MFAS) and optimization (translation recovery)
+together for estimating global translations from relative translation directions and global rotations.
+The purpose of this example is to illustrate the connection between these two classes using a small SfM dataset.
+
+Author: Akshay Krishnan
+Date: September 2020
+"""
 
 import numpy as np
 
 import gtsam
 from gtsam.examples import SFMdata
 
-max_1dsfm_projection_directions = 50
-outlier_weight_threshold = 0.1
 
 def get_data():
-    """"Returns data from SfMData.createPoses(). This contains the global rotations and the unit translations directions."""
+    """"Returns data from SfMData.createPoses(). This contains global rotations and unit translations directions."""
     # Using toy dataset in SfMdata for example.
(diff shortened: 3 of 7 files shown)


python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+    outlier_weights = []
+    # Find the outlier weights for each direction using MFAS.
+    for direction in projection_directions:
+        algorithm = gtsam.MFAS(w_measurements, direction)
+        outlier_weights.append(algorithm.computeOutlierWeights())
+
+    # Compute average of outlier weights.
+    avg_outlier_weights = {}
+    for outlier_weight_dict in outlier_weights:
+        for k, v in outlier_weight_dict.items():
```
swarrier246, line 82: maybe rename k,v to something more descriptive?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def estimate_poses_given_rot(measurements: gtsam.BinaryMeasurementsUnit3,
+                             rotations: gtsam.Values):
+    """Estimate poses given normalized translation directions and rotations between nodes.
+
+    Arguments:
+        measurements {BinaryMeasurementsUnit3}- List of translation direction from the first node to
+        the second node in the coordinate frame of the first node.
+        rotations {Values} -- Estimated rotations
```
johnwlambert, line 51: Thanks for the PR, Akshay. Which way do the estimated rotations point? `wRc`, or `cRw`? Could we clarify that in the docstring?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+            translation_directions.append(gtsam.BinaryMeasurementUnit3(
+                i, j, i_Z_j, gtsam.noiseModel.Isotropic.Sigma(3, 0.01)))
+    # Add the last two rotations.
+    rotations.insert(len(poses) - 1, poses[-1].rotation())
+    rotations.insert(len(poses) - 2, poses[-2].rotation())
+    return (rotations, translation_directions)
+
+
+def estimate_poses_given_rot(measurements: gtsam.BinaryMeasurementsUnit3,
+                             rotations: gtsam.Values):
```
johnwlambert, line 45: missing return type on the function

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+    """Estimate poses given normalized translation directions and rotations between nodes.
+
+    Arguments:
+        measurements {BinaryMeasurementsUnit3}- List of translation direction from the first node to
+        the second node in the coordinate frame of the first node.
```
johnwlambert, line 50: nitpick: would be nice to indent the second line

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+    # Remove measurements that have weight greater than threshold.
+    inlier_measurements = gtsam.BinaryMeasurementsUnit3()
+    [inlier_measurements.append(m) for m in w_measurements if avg_outlier_weights[(
+        m.key1(), m.key2())] < outlier_weight_threshold]
+
+    # Run the optimizer to obtain translations for normalized directions.
+    translations = gtsam.TranslationRecovery(inlier_measurements).run()
+
+    poses = gtsam.Values()
+    for key in rotations.keys():
```
johnwlambert, line 97: can we rename `key` to something more descriptive?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+        the second node in the coordinate frame of the first node.
+        rotations {Values} -- Estimated rotations
+
+    Returns:
+        Values -- Estimated poses.
+    """
+
+    # Some hyperparameters.
+    max_1dsfm_projection_directions = 50
+    outlier_weight_threshold = 0.1
```
johnwlambert, line 59: is this specified in the paper, or a hyperparameter for the user to tune based on their dataset? might be nice to add a small clarification. perhaps these should be module level constants, like `OUTLIER_WEIGHT_THRESHOLD`?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+Date: September 2020
+"""
+
+import numpy as np
+
+import gtsam
+from gtsam.examples import SFMdata
+
+
+def get_data():
```
johnwlambert, line 23: missing type hint on return argument

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+    """"Returns data from SfMData.createPoses(). This contains global rotations and unit translations directions."""
+    # Using toy dataset in SfMdata for example.
+    poses = SFMdata.createPoses(gtsam.Cal3_S2(50.0, 50.0, 0.0, 50.0, 50.0))
+    rotations = gtsam.Values()
+    translation_directions = []
+    for i in range(0, len(poses) - 2):
+        # Add the rotation
+        rotations.insert(i, poses[i].rotation())
```
johnwlambert, line 31: would be nice to clarify what rotations represents: is it `wRc` or `cRw` here?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+    rotations.insert(len(poses) - 2, poses[-2].rotation())
+    return (rotations, translation_directions)
+
+
+def estimate_poses_given_rot(measurements: gtsam.BinaryMeasurementsUnit3,
+                             rotations: gtsam.Values):
+    """Estimate poses given normalized translation directions and rotations between nodes.
+
+    Arguments:
+        measurements {BinaryMeasurementsUnit3}- List of translation direction from the first node to
```
johnwlambert, line 49: if the type hint is already present in the function signature, we should exclude it from the docstring

johnwlambert: (otherwise the docstring becomes bloated, IMHO)

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+    rotations.insert(len(poses) - 1, poses[-1].rotation())
+    rotations.insert(len(poses) - 2, poses[-2].rotation())
+    return (rotations, translation_directions)
+
+
+def estimate_poses_given_rot(measurements: gtsam.BinaryMeasurementsUnit3,
+                             rotations: gtsam.Values):
+    """Estimate poses given normalized translation directions and rotations between nodes.
+
+    Arguments:
```
johnwlambert, line 48: Are we using the "Google" style guide? If so, should be named `Args` (https://google.github.io/styleguide/pyguide.html#383-functions-and-methods)

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+        outlier_weights.append(algorithm.computeOutlierWeights())
+
+    # Compute average of outlier weights.
+    avg_outlier_weights = {}
+    for outlier_weight_dict in outlier_weights:
+        for k, v in outlier_weight_dict.items():
+            if k in avg_outlier_weights:
+                avg_outlier_weights[k] += v / len(outlier_weights)
+            else:
+                avg_outlier_weights[k] = v / len(outlier_weights)
```
johnwlambert, line 86: branching logic for dict insertion can usually be avoided with `setdefault` or `defaultdict` or `get(blah,[])` etc. Can this be written as a dict comprehension?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,112 @@
...
+                poses[j].translation() - poses[i].translation()))
+            translation_directions.append(gtsam.BinaryMeasurementUnit3(
+                i, j, i_Z_j, gtsam.noiseModel.Isotropic.Sigma(3, 0.01)))
+    # Add the last two rotations.
+    rotations.insert(len(poses) - 1, poses[-1].rotation())
+    rotations.insert(len(poses) - 2, poses[-2].rotation())
+    return (rotations, translation_directions)
+
+
+def estimate_poses_given_rot(measurements: gtsam.BinaryMeasurementsUnit3,
```
johnwlambert, line 44: i think the function name could be improved a bit more, since we estimate poses given rotations and directions, not just given rot

johnwlambert (changes requested):

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: I understand the confusion, but don't you think estimate_poses_from_rotations_and_unit_translations() is a bit too long? Even then it would not clarify what the unit translations are. Maybe I can change it to estimate_poses() and let the input arguments speak for the rest?

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: Do you mean a translation averaging python module? It is not part of the MFAS module, since its used outside. Should we capitalize the names here?
Yes, the values are from Kyle Wilson's code. He actually uses 48 for max_1dsfm_projection_directions. I will update that.

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: Not sure what to rename it, any suggestions?

python/gtsam/examples/TranslationAveragingExample.py
ProfFan: `estimate_poses_from_rot_unitt`?

johnwlambert: or maybe just `estimate_poses()`?

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: Have clarified in the comment above.

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: I updated it to `estimate_poses()`.

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: Got it. Would we expect a user to choose different hyperparameters than these? if so, these would go into argparse.

From what you mentioned, it seems like these are relatively fixed constants, in which case I would just define these at the top of this file as `OUTLIER_WEIGHT_THRESHOLD` etc, otherwise they come across as magic numbers

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: didn't quite follow this part, what does each key represent?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,122 @@
...
+        for keypair, weight in outlier_weight_dict.items():
+            avg_outlier_weights[keypair] += weight / len(outlier_weights)
+
+    # Remove w_relative_tranlsations that have weight greater than threshold, these are outliers.
+    inlier_w_relative_translations = gtsam.BinaryMeasurementsUnit3()
+    [inlier_w_relative_translations.append(Z) for Z in w_relative_translations
+        if avg_outlier_weights[(Z.key1(), Z.key2())] < outlier_weight_threshold]
+
+    # Run the optimizer to obtain translations for normalized directions.
+    w_translations = gtsam.TranslationRecovery(
```
johnwlambert, line 103: I'm still a bit hesitant about our notations here. My preference is too be overly explicit, like this: https://github.com/argoai/argoverse-api#a-note-regarding-coordinate-transforms

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,122 @@
...
+        outlier_weights.append(algorithm.computeOutlierWeights())
+
+    # Compute average of outlier weights. Each outlier weight is a map from a pair of Keys (camera IDs) to a weight,
+    # where weights are proportional to the probability of the edge being an outlier.
+    avg_outlier_weights = defaultdict(lambda: 0.0)
+    for outlier_weight_dict in outlier_weights:
+        for keypair, weight in outlier_weight_dict.items():
+            avg_outlier_weights[keypair] += weight / len(outlier_weights)
+
+    # Remove w_relative_tranlsations that have weight greater than threshold, these are outliers.
```
johnwlambert, line 97: typo on `tranlsations`

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,122 @@
...
+
+    outlier_weights = []
+    # Find the outlier weights for each direction using MFAS.
+    for direction in projection_directions:
+        algorithm = gtsam.MFAS(w_relative_translations, direction)
+        outlier_weights.append(algorithm.computeOutlierWeights())
+
+    # Compute average of outlier weights. Each outlier weight is a map from a pair of Keys (camera IDs) to a weight,
+    # where weights are proportional to the probability of the edge being an outlier.
+    avg_outlier_weights = defaultdict(lambda: 0.0)
```
johnwlambert, line 92: i think `avg_outlier_weights = defaultdict(float)` is the more idiomatic way

johnwlambert (changes requested):

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: what does keypair mean?

akshay-krishnan: Keypair is actually a tuple (i, j), a pair of camera indices. I called it keypair because that is what the underlying C++ structure is called.

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: I understand the concern, at the same time, I think the notation that you point to `p_dst = dst_SE3_src * p_src` might contradict with the GTSAM convention (if I am not wrong). Ideally I think we should prefer `dst_p = dst_SE3_src * src_p`. Is there a variable name you recommend here? I called them translations to be somewhat consistent in the file (we have relative_translations), but `w_T_c` would work as well.

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,122 @@
...
+from typing import Tuple, List
+
+import numpy as np
+
+import gtsam
+from gtsam.examples import SFMdata
+
+
+def get_data() -> Tuple[gtsam.Values, List[gtsam.BinaryMeasurementUnit3]]:
+    """"Returns data from SfMData.createPoses(). This contains global rotations and unit translations directions."""
```
johnwlambert, line 27: i might add a little more here, or in the module docstring, explaining where the data comes from. Are we perturbing known poses with noise, and then trying to recover the ground truth from the noisy rotations?

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: Thanks Akshay. I think it would be even more readable if we did it in the code itself. What do you think about the following:

```python
    # Using toy dataset in SfMdata for example.
    wTc_list = SFMdata.createPoses(gtsam.Cal3_S2(50.0, 50.0, 0.0, 50.0, 50.0))
    # Rotations of the cameras in the world frame
    wRc_list = gtsam.Values()
    # Normalized translation directions for pairs of cameras - from camera i to camera j,
    # in the coordinate frame of the first camera.
    jti_directions = []
    for i in range(0, len(poses) - 2):
        # Add the rotation.
        wRc_list.insert(i, wTc_list[i].rotation())
```

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: i'm good with either of those. I just get very confused when I read only `rotation` or `pose`. I would much prefer if we use something like the following throughout this whole file:
`wTc` for SE(3) transformations from camera to world
`wtc` for translations from camera to world
`wRc` for rotations from camera to world

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: maybe we could rename it as `wRc`?

johnwlambert (changes requested):

akshay-krishnan:
Thanks @johnwlambert. Yeah that sounds good, what do you think I should call the relative translations and measurements? Is that naming okay?

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: Can I use `wRc_values` instead of `wRc_list`? Because it is not really a list. 
Also, I think direction from i to j should be `itj_direction` since i is the reference frame? 
And if this direction is expressed in the world frame, can I call it `w_itj_direction`?

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: [Key](https://github.com/borglab/gtsam/blob/ec04369c8870754c058714221f5b36737b3db770/gtsam/inference/Key.h) is a type in GTSAM. If I am not wrong, it is equivalent to a size_t (like an unsigned integer). Translation Averaging, like much of other gtsam code, uses Keys to represent Camera IDs. gtsam.Values is used to store the rotations, and it supports lookup by the key of the camera. 
I don't think calling it wRc would be correct, since it is not a rotation. Maybe we could call it camera_id in external code, but doing so here would hide the fact that it is actually a gtsam Key.

akshay-krishnan: The term keypair used at another place in the code is a pair of keys. The pybind wrapper allows us to treat it like a tuple of ints in python.

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: I think we can still improve this docstring -- i don't think it explains all that is happening here in this function

johnwlambert:
Sure, `wRc_values` is fine with me

gtsam/sfm/MFAS.cpp
```diff
@@ -111,10 +113,8 @@ void removeNodeFromGraph(const Key node,
   graph.erase(node);
 }
 
-MFAS::MFAS(const std::shared_ptr<vector<Key>>& nodes,
-           const TranslationEdges& relativeTranslations,
-           const Unit3& projectionDirection)
-    : nodes_(nodes) {
+MFAS::MFAS(const TranslationEdges& relativeTranslations,
+           const Unit3& projectionDirection) {
```
johnwlambert, line 117: are the `nodes` no longer required here? are they initialized elsewhere?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,122 @@
...
+This example shows how 1dsfm uses outlier rejection (MFAS) and optimization (translation recovery)
+together for estimating global translations from relative translation directions and global rotations.
+The purpose of this example is to illustrate the connection between these two classes using a small SfM dataset.
+
+Author: Akshay Krishnan
+Date: September 2020
+"""
+
+from collections import defaultdict
+from typing import Tuple, List
```
johnwlambert, line 18: prefer to have these alphabetically, `List, Tuple`

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: Hmm i'm a bit confused here. The relative direction in the world frame is just temporary (for the purpose of generating actual relative translation directions, in a camera's own frame), right?

Because the E matrix will only give us in one's own frame.

johnwlambert: if so, i’m good with any of your suggestions about the names for it

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,122 @@
...
+    rotations = gtsam.Values()
+    # Normalized translation directions for pairs of cameras - from first camera to second,
+    # in the coordinate frame of the first camera.
+    translation_directions = []
+    for i in range(0, len(poses) - 2):
+        # Add the rotation.
+        rotations.insert(i, poses[i].rotation())
+        # Create unit translation measurements with next two poses.
+        for j in range(i + 1, i + 3):
+            i_Z_j = gtsam.Unit3(poses[i].rotation().unrotate(
```
johnwlambert, line 40: i think this line is not super readable. can we break it up by semantics, or add inline comments to explain it?

I think it will probably be easier to parse once I can figure out which way `poses[j]` represents.

gtsam/sfm/MFAS.cpp
akshay-krishnan: Nodes are no longer required. They were not being used since I refactored this class a while ago. I had missed removing this unused member then.

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,126 @@
...
+        # Create unit translation measurements with next two poses.
+        for j in range(i + 1, i + 3):
+            i_iZj = gtsam.Unit3(wTc[i].rotation().unrotate(
+                wTc[j].translation() - wTc[i].translation()))
+            i_iZj_list.append(gtsam.BinaryMeasurementUnit3(
+                i, j, i_iZj, gtsam.noiseModel.Isotropic.Sigma(3, 0.01)))
+    # Add the last two rotations.
+    wRc_values.insert(len(wTc) - 1, wTc[-1].rotation())
+    wRc_values.insert(len(wTc) - 2, wTc[-2].rotation())
+    return (wRc_values, i_iZj_list)
```
johnwlambert, line 54: Parentheses are not needed here on return args

johnwlambert:
Much improved with the new variable names, thanks Akshay.

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,126 @@
...
+
+    wTc_values = gtsam.Values()
+    for key in wRc_values.keys():
+        wTc_values.insert(key, gtsam.Pose3(
+            wRc_values.atRot3(key), wtc_values.atPoint3(key)))
+    return wTc_values
+
+
+def main():
+    wRc_values, w_iZj_list = get_data()
```
johnwlambert, line 118: i think the 2nd return argument should be `i_iZj_list`, not `w_iZj_list`

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,126 @@
...
+    wRc_values.insert(len(wTc) - 2, wTc[-2].rotation())
+    return (wRc_values, i_iZj_list)
+
+
+def estimate_poses(i_iZj_list: gtsam.BinaryMeasurementsUnit3,
+                   wRc_values: gtsam.Values) -> gtsam.Values:
+    """Estimate poses given rotations and normalized translation directions between cameras.
+
+    Args:
+        iZj_list -- List of normalized translation direction measurements between camera pairs,
```
johnwlambert, line 62: i think this should say
```python
Args:
     i_iZj_list - ...
````

Comment (johnwlambert) [reply python/gtsam/examples/TranslationAveragingExample.py:62]: Also, these should be colons here, not hyphens (`argname: ...`, not `argname -- ...` (https://google.github.io/styleguide/pyguide.html#383-functions-and-methods):

```python
        wRc_values: Rotations of the cameras in the world frame.
    Returns:
        Values: Estimated poses.
```

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,126 @@
...
+    for i_iZj in i_iZj_list:
+        w_iZj = gtsam.Unit3(wRc_values.atRot3(i_iZj.key1())
+                                      .rotate(i_iZj.measured().point3()))
+        w_iZj_list.append(gtsam.BinaryMeasurementUnit3(
+            i_iZj.key1(), i_iZj.key2(), w_iZj, i_iZj.noiseModel()))
+
+    # Indices of measurements that are to be used as projection directions. 
+    # These are randomly chosen.
+    sampled_indices = np.random.choice(len(w_iZj_list), min(
+        MAX_1DSFM_PROJECTION_DIRECTIONS, len(w_iZj_list)), replace=False)
```
johnwlambert, line 83: i think this would be more readable as follows:
```python
# all sampled directions must be unique
num_directions_to_sample = min(MAX_1DSFM_PROJECTION_DIRECTIONS, len(w_iZj_list))
sampled_indices = np.random.choice(len(w_iZj_list), num_directions_to_sample, replace=False)
```

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: Got it. thanks for the clarification. We still have a typo `w_relative_tranlsations`, and that variable should be renamed in the comment too

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,126 @@
...
+    # Sample projection directions from the measurements.
+    projection_directions = [w_iZj_list[idx].measured()
+                             for idx in sampled_indices]
+
```
johnwlambert, line 87: i think this function is too long. can we split out the outlier rejection part, so it becomes `prune_to_inliers()` or `keep_inliers_only()` or `compute_inliers()`?
```python
def prune_to_inliers(w_iZj_list: gtsam.BinaryMeasurementsUnit3) -> gtsam.BinaryMeasurementsUnit3:
   """ """
    # Indices of measurements that are to be used as projection directions. 
    # These are randomly chosen.
    sampled_indices = np.random.choice(len(w_iZj_list), min(
        MAX_1DSFM_PROJECTION_DIRECTIONS, len(w_iZj_list)), replace=False)

# Sample projection directions from the measurements.
    projection_directions = [w_iZj_list[idx].measured()
                             for idx in sampled_indices]

    outlier_weights = []
    # Find the outlier weights for each direction using MFAS.
    for direction in projection_directions:
        algorithm = gtsam.MFAS(w_iZj_list, direction)
        outlier_weights.append(algorithm.computeOutlierWeights())

    # Compute average of outlier weights. Each outlier weight is a map from a pair of Keys 
    # (camera IDs) to a weight, where weights are proportional to the probability of the edge 
    # being an outlier.
    avg_outlier_weights = defaultdict(float)
    for outlier_weight_dict in outlier_weights:
        for keypair, weight in outlier_weight_dict.items():
            avg_outlier_weights[keypair] += weight / len(outlier_weights)

    # Remove w_relative_tranlsations that have weight greater than threshold, these are outliers.

    w_iZj_inliers = gtsam.BinaryMeasurementsUnit3()
    [w_iZj_inliers.append(Z) for Z in w_iZj_list
        if avg_outlier_weights[(Z.key1(), Z.key2())] < OUTLIER_WEIGHT_THRESHOLD]
    return w_iZj_inliers
```

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,126 @@
...
+OUTLIER_WEIGHT_THRESHOLD = 0.1
+
+
+def get_data() -> Tuple[gtsam.Values, List[gtsam.BinaryMeasurementUnit3]]:
+    """"Returns global rotations and unit translation directions between 8 cameras 
+    that lie on a circle and face the center. The poses of 8 cameras are obtained from SFMdata 
+    and the unit translations directions between some camera pairs are computed from their 
+    global translations. """
+    # Using toy dataset in SfMdata for example.
+    wTc = SFMdata.createPoses(gtsam.Cal3_S2(50.0, 50.0, 0.0, 50.0, 50.0))
```
johnwlambert, line 36: maybe this should be named `wTc_list` instead of `wTc` since it represents multiple poses?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,126 @@
...
+    # Normalized translation directions from camera i to camera j
+    # in the coordinate frame of camera i.
+    i_iZj_list = []
+    for i in range(0, len(wTc) - 2):
+        # Add the rotation.
+        wRc_values.insert(i, wTc[i].rotation())
+        # Create unit translation measurements with next two poses.
+        for j in range(i + 1, i + 3):
+            i_iZj = gtsam.Unit3(wTc[i].rotation().unrotate(
+                wTc[j].translation() - wTc[i].translation()))
```
johnwlambert, line 48:
```python
# compute the ray from pose i to pose j
w_iZj = wTc[j].translation() - wTc[i].translation()
wRi = wTc[i].rotation()
iRw = wRi.inverse()
i_iZj = gtsam.Unit3(iRw.rotate(w_iZj))
```

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: Yeah, wonder why I did that.

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: Okay, I am not very familiar with Python conventions, was just following Shonan Averaging. https://github.com/borglab/gtsam/blob/develop/python/gtsam/examples/ShonanAveragingCLI.py

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: But we could use the `unrotate` method right? Instead of taking the inverse and then rotating.

akshay-krishnan: Also, w_iZj in your code is not a ray, if you are defining ray as the unit translation direction.

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Args:
+        i_iZj_list: List of normalized translation direction measurements between camera pairs, 
+                    Z here refers to measurements. The measurements are of camera j with reference 
+                    to camera i (iZj), in camera i's coordinate frame (i_). iZj represents a unit 
+                    vector to j in i's frame and is not a transformation. 
+        wRc_values: Rotations of the cameras in the world frame.
+
+    Returns:
+        Values: Estimated poses.
```
johnwlambert, line 111: looks good. i think this return arg should be named `wTc_values`, not `Values`

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,144 @@
...
+        outlier_weights.append(algorithm.computeOutlierWeights())
+
+    # Compute average of outlier weights. Each outlier weight is a map from a pair of Keys
+    # (camera IDs) to a weight, where weights are proportional to the probability of the edge
+    # being an outlier.
+    avg_outlier_weights = defaultdict(float)
+    for outlier_weight_dict in outlier_weights:
+        for keypair, weight in outlier_weight_dict.items():
+            avg_outlier_weights[keypair] += weight / len(outlier_weights)
+
```
johnwlambert, line 90: I think we still need to rename `w_relative_tranlsation` in the line below

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,144 @@
...
+OUTLIER_WEIGHT_THRESHOLD = 0.1
+
+
+def get_data() -> Tuple[gtsam.Values, List[gtsam.BinaryMeasurementUnit3]]:
+    """"Returns global rotations and unit translation directions between 8 cameras
+    that lie on a circle and face the center. The poses of 8 cameras are obtained from SFMdata
+    and the unit translations directions between some camera pairs are computed from their
+    global translations. """
+    # Using toy dataset in SfMdata for example.
+    wTc_list = SFMdata.createPoses(gtsam.Cal3_S2(50.0, 50.0, 0.0, 50.0, 50.0))
```
johnwlambert, line 36: i would probably write the following to make this more readable:
```python
fx, fy, s, u0, v0 = 50.0, 50.0, 0.0, 50.0, 50.0
wTc_list = SFMdata.createPoses(gtsam.Cal3_S2(fx, fy, s, u0, v0))
```

python/gtsam/examples/TranslationAveragingExample.py
johnwlambert: unrotate works too, i find it a bit less intuitive,  but your call

johnwlambert:
Hi Akshay -- looks good. I just left last 4 things, approving now, I trust you'll address them before merge.

johnwlambert (approved):

python/gtsam/examples/TranslationAveragingExample.py
akshay-krishnan: Are variable names mentioned in return args? This is the type not the name.

johnwlambert: from what i understand, the variable name can be listed in return args. but the type does not need to be repeated there, since it's redundant with the return arg type hint in the function signature

johnwlambert (approved):

dellaert (approved):
Looks OK to me, but with pybind we should be able to use native python collections, right @ProfFan ? I’m ok with doing that in another PR

gtsam/sfm/MFAS.cpp
```diff
@@ -7,6 +7,8 @@
 
 #include <gtsam/sfm/MFAS.h>
 
+#include <boost/shared_ptr.hpp>
```
ayushbaid, line 10: Is shared_ptr being used right now? I see it has been removed from one instance.

gtsam/sfm/tests/testMFAS.cpp
```diff
@@ -6,9 +6,11 @@
  */
 
 #include <gtsam/sfm/MFAS.h>
-#include <iostream>
+
 #include <CppUnitLite/TestHarness.h>
 
+#include <iostream>
```
ayushbaid, line 12: do we need iostream now?

python/gtsam/examples/TranslationAveragingExample.py
```diff
@@ -0,0 +1,144 @@
...
+            i_iZj = gtsam.Unit3(i_itj)
+            i_iZj_list.append(gtsam.BinaryMeasurementUnit3(
+                i, j, i_iZj, gtsam.noiseModel.Isotropic.Sigma(3, 0.01)))
+    # Add the last two rotations.
+    wRc_values.insert(len(wTc_list) - 1, wTc_list[-1].rotation())
+    wRc_values.insert(len(wTc_list) - 2, wTc_list[-2].rotation())
+    return wRc_values, i_iZj_list
+
+
+def prune_to_inliers(w_iZj_list: gtsam.BinaryMeasurementsUnit3) -> gtsam.BinaryMeasurementsUnit3:
```
ayushbaid, line 62: maybe prune_ouliers? or filter_outliers?

ayushbaid (commented):
Minor comments.

Otherwise, LGTM

swarrier246 (approved):

PR #536, opened 2020-09-22
Add numericalDerivative Dimension Template Parameter

varunagrawal:
- Update `numericalDerivative` functions to take in optional dimension template parameter.
- Added corresponding tests.

I needed this for the `ParameterMatrix` since `numericalDerivativeXY` complains about dynamic sized matrices if `N` is not provided (aka [line 128](https://github.com/borglab/gtsam/blob/a4c3623701a4b0077ced76af45c3927e9727ca05/gtsam/base/numericalDerivative.h#L128)).


varunagrawal:
@ProfFan @dellaert  this should be a quick review :slightly_smiling_face:

dellaert (approved):

PR #537, opened 2020-09-22
Expose 3d point attribute p of SfmTrack in wrapper

johnwlambert:
- Grant `SfmTrack` 3d point access to wrapper
- Update the list of updated python examples, to reflect available examples implemented already in C++


gtsam/gtsam.i
```diff
@@ -2847,6 +2847,7 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 
 #include <gtsam/slam/dataset.h>
 class SfmTrack {
```
dellaert, line 2849: I’d prefer the name `point3`

dellaert (approved):

gtsam/gtsam.i
johnwlambert: sure, so rename `p` to be `point3`? I think we will need to rename it in the following examples in this PR then since it's a breaking change:
- `examples/CreateSFMExampleData.cpp`
- `timing/timeSFMBALnavTcam.cpp`
- `timing/timeSFMBAL.cpp`
- `examples/SFMExample_bal_COLAMD_METIS.cpp`
- `examples/SFMExampleExpressions_bal.cpp`
- `examples/SFMExample_bal.cpp`

varunagrawal: Nope, rename `get3dPoint` to `point3`.

johnwlambert: Ah I see, @dellaert do you mind confirming that and I can quickly make the change? I guess “get*()” is not preferred in GTSAM?

dellaert: Indeed, no get...

johnwlambert: pushed the update, should be good to go now

PR #538, opened 2020-09-22
Remove logmap function from LieVector and LieMatrix

varunagrawal:
Fixes regression introduced in #532.


ayushbaid (approved):
LGTM.

Simple change reverting the function addition

dellaert (approved):

PR #539, opened 2020-09-23
Add calibrate with jacobians for Cal3Bundler

### LINKED ISSUE #237: Mismatch between calibrate functions for Cal3_S2 and Cal3Bundler

varunagrawal:
Added an additional `calibrate` method for `Cal3Bundler` which compute jacobians.

Since the normal calibration method is iterative, I have approximated the jacobian as if a single iteration step has occurred, which should suffice assuming the projection is close to the ideal measurements.

I would definitely like a thorough review on this though.

Fixes #237

### DIFF
--- a/gtsam/geometry/Cal3Bundler.cpp
+++ b/gtsam/geometry/Cal3Bundler.cpp
@@ -121,23 +121,23 @@ Point2 Cal3Bundler::calibrate(const Point2& pi,
     throw std::runtime_error(
         "Cal3Bundler::calibrate fails to converge. need a better initialization");
 
-  x = invKPi.x(), y = invKPi.y();
-  const double xx = x * x, yy = y * y;
-  const double rr = xx + yy;
-  const double g = (1 + k1_ * rr + k2_ * rr * rr);
-  const double inv_f = 1 / f_, inv_g = 1 / g;
+  // We make use of the Implicit Function Theorem to compute the Jacobians from uncalibrate
+  // Given f(pi, pn) = uncalibrate(pn) - pi, and g(pi) = calibrate, we can easily compute the Jacobians
+  // df/pi = -I (pn and pi are independent args)
+  // Dcal = -inv(H_uncal_pn) * df/pi = -inv(H_uncal_pn) * (-I) = inv(H_uncal_pn)
+  // Dp = -inv(H_uncal_pn) * df/K = -inv(H_uncal_pn) * H_uncal_K
+  Matrix H_uncal_K, H_uncal_pn;
 
-  // Approximate the jacobians via a single iteration of g.
-  if (Dcal) {
-    *Dcal << -inv_f * x, -x * inv_g * rr, -x * inv_g * rr * rr, -inv_f * y,
-        -y * inv_g * rr, -y * inv_g * rr * rr;
+  if (Dcal || Dp) {
+    // Compute uncalibrate Jacobians
+    uncalibrate(pn, H_uncal_K, H_uncal_pn);
+  }
+
+  if (Dcal) { 
+    *Dcal = -H_uncal_pn.inverse() * H_uncal_K;
   }
   if (Dp) {
-    const double dg_du = (2 * k1_ * x * inv_f) + (4 * k2_ * rr * x * inv_f);
-    const double dg_dv = (2 * k1_ * y * inv_f) + (4 * k2_ * rr * y * inv_f);
-
-    *Dp << (inv_f * inv_g) - (pn.x() * inv_g * dg_du), -pn.x() * inv_g * dg_dv,
-        -pn.y() * inv_g * dg_du, (inv_f * inv_g) - (pn.x() * inv_g * dg_dv);
+    *Dp = H_uncal_pn.inverse();
   }
 
   return pn;
--- a/gtsam/geometry/Cal3Bundler.cpp
+++ b/gtsam/geometry/Cal3Bundler.cpp
@@ -94,7 +94,9 @@ Point2 Cal3Bundler::uncalibrate(const Point2& p, //
 }
 
 /* ************************************************************************* */
-Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
+Point2 Cal3Bundler::calibrate(const Point2& pi, OptionalJacobian<2, 3> Dcal,
+                              OptionalJacobian<2, 2> Dp,
+                              const double tol) const {
   // Copied from Cal3DS2 :-(
   // but specialized with k1,k2 non-zero only and fx=fy and s=0
   const Point2 invKPi((pi.x() - u0_)/f_, (pi.y() - v0_)/f_);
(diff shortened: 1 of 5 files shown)


varunagrawal:
Okay so this is breaking because calling code is getting confused which version of `calibrate` to call.

@dellaert should I merge the two `calibrate` functions into one?

dellaert:
> Okay so this is breaking because calling code is getting confused which version of `calibrate` to call.
> 
> @dellaert should I merge the two `calibrate` functions into one?

Yes :-)

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -118,6 +120,25 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
     throw std::runtime_error(
         "Cal3Bundler::calibrate fails to converge. need a better initialization");
 
+  const double x = invKPi.x(), y = invKPi.y();
```
dellaert, line 123: remove invKPi variable, just init x and y on line 102, and use to init `pn` on line 105.

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -118,6 +120,25 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
...
+  const double xx = x * x, yy = y * y;
+  const double rr = xx + yy;
+  const double g = (1 + k1_ * rr + k2_ * rr * rr);
+  const double inv_f = 1 / f_, inv_g = 1 / g;
+
+  // Approximate the jacobians via a single iteration of g.
```
dellaert, line 129: Does this comment still hold? You are doing multiple iterations above

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -118,6 +120,25 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
...
+  if (Dcal) {
+    *Dcal << -inv_f * x, -x * inv_g * rr, -x * inv_g * rr * rr, -inv_f * y,
```
dellaert, line 131: Where does this come from? Document

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -118,6 +120,25 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
...
+        -y * inv_g * rr, -y * inv_g * rr * rr;
+  }
+  if (Dp) {
+    const double dg_du = (2 * k1_ * x * inv_f) + (4 * k2_ * rr * x * inv_f);
```
dellaert, line 135: same

gtsam/geometry/Cal3Bundler.cpp
varunagrawal: We're doing multiple iterations but when calculating the jacobians, I assume only one iteration is done. So it neither the initial estimate or the final iterate (which could be after n<=10 iterations), but after a single iteration.

gtsam/geometry/Cal3Bundler.cpp
varunagrawal: We still need invKPi in the iteration...

gtsam/geometry/Cal3Bundler.cpp
varunagrawal: What would you recommend adopting? This was one of the reasons this PR took as long as it did.

dellaert: `calibrate` takes undistorted coordinates and makes them into pixels. This is iterative, but at end we have the correct relationship. Then we ask how the answer changes *differentially*, after we found this answer. So, it clearly needs to be the final iterate, as I understand it. So, I'd recommend taking rr, g etc out of the loop, initialize them from x and y, and then update at end of the loop so they are valid for the last estimate.

You still have to document how you got the derivatives. These things are more clear when there is a closed form solution, but it's not obvious when we are iterating on a nonlinear relationship.

gtsam/gtsam.i
```diff
@@ -1053,7 +1053,7 @@ class Cal3_S2Stereo {
 class Cal3Bundler {
   // Standard Constructors
   Cal3Bundler();
-  Cal3Bundler(double fx, double k1, double k2, double u0, double v0);
+  Cal3Bundler(double fx, double k1, double k2, double u0, double v0, double tol);
```
dellaert, line 1056: Wrapper can’t do default arguments  so either undo or add another version.

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -25,13 +25,13 @@ namespace gtsam {
...
-    f_(1), k1_(0), k2_(0), u0_(0), v0_(0) {
+    f_(1), k1_(0), k2_(0), u0_(0), v0_(0), tol_(1e-5) {
 }
 
 /* ************************************************************************* */
-Cal3Bundler::Cal3Bundler(double f, double k1, double k2, double u0, double v0) :
-    f_(f), k1_(k1), k2_(k2), u0_(u0), v0_(v0) {
-}
+Cal3Bundler::Cal3Bundler(double f, double k1, double k2, double u0, double v0,
+                         double tol)
```
dellaert, line 33: I like it

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -94,21 +94,24 @@ Point2 Cal3Bundler::uncalibrate(const Point2& p, //
...
 /* ************************************************************************* */
-Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
+Point2 Cal3Bundler::calibrate(const Point2& pi,
+                              OptionalJacobian<2, 3> Dcal,
+                              OptionalJacobian<2, 2> Dp) const {
   // Copied from Cal3DS2 :-(
   // but specialized with k1,k2 non-zero only and fx=fy and s=0
-  const Point2 invKPi((pi.x() - u0_)/f_, (pi.y() - v0_)/f_);
+  double x = (pi.x() - u0_)/f_, y = (pi.y() - v0_)/f_;
+  const Point2 invKPi(x, y);
```
dellaert, line 103: Not needed, x  ,y are never changed. Ensure this by making const. In fact add const to all things that never change

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -118,6 +121,25 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
     throw std::runtime_error(
         "Cal3Bundler::calibrate fails to converge. need a better initialization");
 
+  x = invKPi.x(), y = invKPi.y();
+  const double xx = x * x, yy = y * y;
```
dellaert, line 125: My suggestion was to take rr and g out of the loop above and make them non-const

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -118,6 +121,25 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
...
+  const double rr = xx + yy;
+  const double g = (1 + k1_ * rr + k2_ * rr * rr);
+  const double inv_f = 1 / f_, inv_g = 1 / g;
+
+  // Approximate the jacobians via a single iteration of g.
```
dellaert, line 130: Obsolete comment if you buy my argument above, but you are still using the initial estimate for jacobians?

varunagrawal:
@dellaert great call on the Implicit Function theorem. Thinking in math was the right approach :smile: 
This is now ready for review.

+ @johnwlambert

gtsam/geometry/tests/testCal3Bundler.cpp
```diff
@@ -69,11 +71,20 @@ TEST( Cal3Bundler, Duncalibrate)
   Matrix numerical2 = numericalDerivative22(uncalibrate_, K, p);
   CHECK(assert_equal(numerical1,Dcal,1e-7));
   CHECK(assert_equal(numerical2,Dp,1e-7));
-  CHECK(assert_equal(numerical1,K.D2d_calibration(p),1e-7));
-  CHECK(assert_equal(numerical2,K.D2d_intrinsic(p),1e-7));
-  Matrix Dcombined(2,5);
-  Dcombined << Dp, Dcal;
-  CHECK(assert_equal(Dcombined,K.D2d_intrinsic_calibration(p),1e-7));
```
varunagrawal, line 76: These methods are deprecated, which is why I removed these tests.

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -118,6 +121,25 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, const double tol) const {
     throw std::runtime_error(
         "Cal3Bundler::calibrate fails to converge. need a better initialization");
 
+  // We make use of the Implicit Function Theorem to compute the Jacobians from uncalibrate
+  // Given f(pi, pn) = uncalibrate(pn) - pi, and g(pi) = calibrate, we can easily compute the Jacobians
+  // df/pi = -I (pn and pi are independent args)
+  // Dcal = -inv(H_uncal_pn) * df/pi = -inv(H_uncal_pn) * (-I) = inv(H_uncal_pn)
+  // Dp = -inv(H_uncal_pn) * df/K = -inv(H_uncal_pn) * H_uncal_K
+  Matrix H_uncal_K, H_uncal_pn;
```
dellaert, line 129: These matrices should be fixed-size. Also, cache H_uncal_pn here? (In a fixed size Matrix) finally, you can avoid calculating H_uncal_K using `Dcal ? &H_uncal_K : nullptr in first argument below.

dellaert (commented):
Awesome. You *need* to use fixed-size matrices if you know sizes, esp in functions like these that could be called for all points.

Maybe make an issue to hunt down and fix other implicit theorem instances?

dellaert (approved):

dellaert:
Approved as I’ll trust you’ll make those final changes

gtsam/geometry/Cal3Bundler.cpp
varunagrawal: Quick question about caching H_uncal_pn: should I be calling uncalibrate with H_uncal_pn always?

PR #464, opened 2020-08-10
Update Simitary3 with Align feature

Alexma3312:
**This PR adds four functions into the Similarity3 class.**
The code format is based on Google format and GTSAM::Pose3::Align().

1. transformFrom - Similarity transform of a Pose3 object
2. Align(point pairs) - Calculate Similarity transform from matched point pairs
3. rotationAveraging - Calculate average rotation from a list of rotations
4. Align(pose pairs) - Calculate Similarity transform from matched pose pairs

**Visualization of unit tests for the Align functions:**
1. Calculate Similarity transform from matched point pairs
Input:
r_point1 = Point3(0, 0, 0) g_point1 = Point3(6, 8, 10)
r_point2 = Point3(3, 0, 0) g_point2 = Point3(6, 2, 10)
r_point3 = Point3(3, 0, 4) g_point3 = Point3(6, 2, 18)
pr1

Output:
gRr = Rot3.Rz(math.radians(-90))
s = 2
gtr = Point3(6, 8, 10)

2. Calculate Similarity transform from matched pose pairs
r_pose1 = Pose3(Rot3(1, 0, 0, 0, 1, 0, 0, 0, 1), Point3(0, 0, 0))
r_pose2 = Pose3(Rot3(-1, 0, 0, 0, 1, 0, 0, 0, 1), Point3(4, 0, 0))
g_pose1 = Pose3(Rot3(-1, 0, 0, 0, 1, 0, 0, 0, -1), Point3(4, 6, 10))
g_pose2 = Pose3(Rot3(1, 0, 0, 0, 1, 0, 0, 0, -1), Point3(-4, 6, 10))
pr2

Output:
gRr = Rot3.Ry(math.radians(180))
s = 2
gtr = Point3(4, 6, 10)

PR #542, opened 2020-09-29
Point3 mean to throw an exception when input size=0 .

Alexma3312:
https://github.com/borglab/gtsam/pull/464 follow up.
A simple fix to throw an exception when the input size equals to zero in the Point3 mean functions. @jlblancoc

### DIFF
--- a/gtsam/geometry/Point3.cpp
+++ b/gtsam/geometry/Point3.cpp
@@ -77,6 +77,7 @@ double dot(const Point3 &p, const Point3 &q, OptionalJacobian<1, 3> H1,
 
 Point3Pair mean(const std::vector<Point3Pair> &abPointPairs) {
   const size_t n = abPointPairs.size();
+  if (n == 0) throw std::invalid_argument("input should have at least 1 pair of points");
   Point3 aCentroid(0, 0, 0), bCentroid(0, 0, 0);
   for (const Point3Pair &abPair : abPointPairs) {
     aCentroid += abPair.first;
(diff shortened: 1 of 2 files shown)


gtsam/geometry/Point3.cpp
```diff
@@ -77,6 +77,7 @@ double dot(const Point3 &p, const Point3 &q, OptionalJacobian<1, 3> H1,
 
 Point3Pair mean(const std::vector<Point3Pair> &abPointPairs) {
   const size_t n = abPointPairs.size();
+  if (n == 0) throw std::invalid_argument("input should have at least 1 pair of points");
```
varunagrawal, line 80: Slightly more descriptive error message?

`Point3Pair::mean input vector is empty`

gtsam/geometry/Point3.h
```diff
@@ -62,6 +62,7 @@ GTSAM_EXPORT double dot(const Point3& p, const Point3& q,
 /// mean
 template <class CONTAINER>
 GTSAM_EXPORT Point3 mean(const CONTAINER& points) {
+  if (points.size() == 0) throw std::invalid_argument("input should have at least 1 point");
```
varunagrawal, line 65: Same as above.

varunagrawal (approved):
LGTM except for a minor comment.

gtsam/geometry/Point3.cpp
Alexma3312: sure

dellaert (approved):

PR #547, opened 2020-09-30
Feature/robust shonan

lucacarlone:
Added option to use a Huber norm in Shonan Rotation Averaging.
The main changes were:

- added a useHuber option in Shonan

- added machinery to convert binaryMeasurements to use a robust noise model

- adjusted the FrobeniusFactor-related code to cope with possibly robust costs

- updated the example ShonanAveragingCLI.cpp


gtsam/sfm/BinaryMeasurement.h
```diff
@@ -71,6 +71,14 @@ template <class T> class BinaryMeasurement : public Factor {
     this->noiseModel_->print("  noise model: ");
   }
 
+  // TODO: make this more general?
```
dellaert, line 74: In GTSAM we prefer to have const methods only; please refactor by passing the robust model in the constructor.

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -53,7 +53,8 @@ ShonanAveragingParameters<d>::ShonanAveragingParameters(
       optimalityThreshold(optimalityThreshold),
       alpha(alpha),
       beta(beta),
-      gamma(gamma) {
+      gamma(gamma),
+	  useHuber(false){
```
dellaert, line 57: Cool, but please format all changed sections with google style, indentation seems off

gtsam/sfm/ShonanAveraging.h
```diff
@@ -151,6 +164,15 @@ class GTSAM_EXPORT ShonanAveraging {
     return measurements_[k];
   }
 
+  /// wrap factors with robust Huber loss
+  static Measurements makeNoiseModelRobust(Measurements measurements){
+	  Measurements robustMeasurements = measurements;
```
dellaert, line 169: I’m not in favor of this imperative scheme, but FWIW you are passing by value so the copy was already made :-)

gtsam/slam/FrobeniusFactor.h
```diff
@@ -34,7 +34,7 @@ namespace gtsam {
  * isotropic. If it is, we extend to 'n' dimensions, otherwise we throw an
  * error. If defaultToUnit == false throws an exception on unexepcted input.
  */
-GTSAM_EXPORT boost::shared_ptr<noiseModel::Isotropic>
+GTSAM_EXPORT SharedNoiseModel
```
dellaert, line 37: Comment above behavior for robust noise model

.cproject
```diff
@@ -1,3563 +1,82 @@
...
-				<buildArguments>-j5</buildArguments>
-				<buildTarget>wrap</buildTarget>
-				<stopOnError>true</stopOnError>
-				<useDefaultCommand>true</useDefaultCommand>
-				<runAllBuilders>true</runAllBuilders>
-			</target>
-		</buildTargets>
-	</storageModule>
-</cproject>
+<?xml version="1.0" encoding="UTF-8"?>
```
dellaert, line 1: Please just delete this file :-)

dellaert (changes requested):
Awesome! However, please refactor to pass noise model in constructor of binary measurements. Imperative code goes against GTSAM design guidelines.

varunagrawal:
@dellaert @lucacarlone I went ahead and addressed the review comments while also removing the `goto` in the `FrobeniusFactor`.

dellaert:
I’ll review, but FWIW a goto can be the best choice within a function. It’s bad reputation stems from very long programs with lots of goto statements.

lucacarlone:
@varunagrawal  this is great!! it was still on my calendar, but had to push it back for a while. Is there anything else for which you need my help on this branch?

varunagrawal:
@lucacarlone I think I've got this covered. The PR was already in great shape to begin with. 😊

I only need folks to review all my PRs now.

varunagrawal:
@dellaert gentle reminder :slightly_smiling_face:

examples/ShonanAveragingCLI.cpp
```diff
@@ -85,11 +92,14 @@ int main(int argc, char* argv[]) {
...
+  auto lmParams = LevenbergMarquardtParams::CeresDefaults();
   if (d == 2) {
     cout << "Running Shonan averaging for SO(2) on " << inputFile << endl;
-    ShonanAveraging2 shonan(inputFile);
+    ShonanAveraging2::Parameters parameters(lmParams);
+    parameters.setUseHuber(useHuberLoss);
+    ShonanAveraging2 shonan(inputFile,parameters);
     auto initial = shonan.initializeRandomly(rng);
-    auto result = shonan.run(initial);
+    auto result = shonan.run(initial,pMin);
```
dellaert, line 102: spacing. Did you lint?

examples/ShonanAveragingCLI.cpp
```diff
@@ -58,6 +61,10 @@ int main(int argc, char* argv[]) {
       "Write solution to the specified file")(
       "dimension,d", po::value<int>(&d)->default_value(3),
       "Optimize over 2D or 3D rotations")(
+      "useHuberLoss,h", po::value<bool>(&useHuberLoss)->default_value(false),
+	  "set True to use Huber loss")(
+	  "pMin,p", po::value<int>(&pMin)->default_value(3),
```
dellaert, line 66: spacing. Please format (only parts that have been changed to avoid large diff) and also lint

examples/ShonanAveragingCLI.cpp
```diff
@@ -101,9 +111,11 @@ int main(int argc, char* argv[]) {
     poses = initialize::computePoses<Pose2>(result.first, &poseGraph);
   } else if (d == 3) {
     cout << "Running Shonan averaging for SO(3) on " << inputFile << endl;
-    ShonanAveraging3 shonan(inputFile);
+    ShonanAveraging3::Parameters parameters(lmParams);
+    parameters.setUseHuber(useHuberLoss);
+    ShonanAveraging3 shonan(inputFile,parameters);
     auto initial = shonan.initializeRandomly(rng);
-    auto result = shonan.run(initial);
+    auto result = shonan.run(initial,pMin);
```
dellaert, line 118: spacing

gtsam/sfm/BinaryMeasurement.h
```diff
@@ -45,11 +45,43 @@ template <class T> class BinaryMeasurement : public Factor {
...
   SharedNoiseModel noiseModel_; ///< Noise model
 
-public:
+ public:
   BinaryMeasurement(Key key1, Key key2, const T &measured,
-                    const SharedNoiseModel &model = nullptr)
-      : Factor(std::vector<Key>({key1, key2})), measured_(measured),
-        noiseModel_(model) {}
+                    const SharedNoiseModel &model = nullptr,
+                    bool useHuber = false)
```
dellaert, line 51: I'm a bit puzzled about this flag. Why not just give BinaryMeasurement a robust error model as the input argument. I think the code should be re-factored to do just that, construct the noise model out of this class.

gtsam/sfm/ShonanAveraging.h
```diff
@@ -53,6 +53,8 @@ struct GTSAM_EXPORT ShonanAveragingParameters {
   double alpha;                 // weight of anchor-based prior (default 0)
   double beta;                  // weight of Karcher-based prior (default 1)
   double gamma;                 // weight of gauge-fixing factors (default 0)
+  bool useHuber;  // if enabled, the Huber loss is used in the optimization
```
dellaert, line 56: Spacing for comments is inconsistent. (Format changed part only)

gtsam/sfm/ShonanAveraging.h
```diff
@@ -77,6 +79,18 @@ struct GTSAM_EXPORT ShonanAveragingParameters {
 
   void setGaugesWeight(double value) { gamma = value; }
   double getGaugesWeight() { return gamma; }
+
+  void setUseHuber(bool value) { useHuber = value; }
+  bool getUseHuber() { return useHuber; }
+
+  void print() const {
```
dellaert, line 86: document with singe line `///`

gtsam/sfm/ShonanAveraging.h
```diff
@@ -77,6 +79,18 @@ struct GTSAM_EXPORT ShonanAveragingParameters {
...
+    std::cout << " ShonanAveragingParameters: " << std::endl;
+    std::cout << " alpha: " << alpha  << std::endl;
+    std::cout << " beta: " << beta  << std::endl;
+    std::cout << " gamma: " << gamma  << std::endl;
+    std::cout << " useHuber: " << useHuber  << std::endl;
+    std::cout << " --------------------------" << std::endl;
```
dellaert, line 92: kill this. Should be responsibility of caller

gtsam/sfm/ShonanAveraging.h
```diff
@@ -151,6 +164,15 @@ class GTSAM_EXPORT ShonanAveraging {
     return measurements_[k];
   }
 
+  /// wrap factors with robust Huber loss
+  Measurements makeNoiseModelRobust(const Measurements& measurements) const {
+    Measurements robustMeasurements = measurements;
+    for (auto &measurement : robustMeasurements) {
+      measurement = BinaryMeasurement<Rot>(measurement, true);
```
dellaert, line 171: imperative, without any good reason. Please create a new Measurements variables, reserve space, and emplace_back :-)

gtsam/slam/FrobeniusFactor.cpp
```diff
@@ -23,31 +23,47 @@ using namespace std;
...
+      sigmas = robust->noise()->sigmas();
+    } else {
+      sigmas = model->sigmas();
+    }
+
     size_t n = sigmas.size();
     if (n == 1) {
       sigma = sigmas(0); // Rot2
-      goto exit;
+      exit = true;
```
dellaert, line 43: restore goto flow. It was good.

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -338,11 +336,23 @@ template <typename T>
...
-    throw std::invalid_argument(
-        "Shonan averaging noise models must be isotropic.");
+  double sigma;
+  if (isotropic) {
+    sigma = isotropic->sigma();
+  } else {
+    const auto &robust = boost::dynamic_pointer_cast<noiseModel::Robust>(
+        measurement.noiseModel());
+    if (robust) {
+      std::cout << "Verification of optimality does not work with robust cost "
```
dellaert, line 346: It should throw.

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -793,6 +803,12 @@ std::pair<Values, double> ShonanAveraging<d>::run(const Values &initialEstimate,
   for (size_t p = pMin; p <= pMax; p++) {
     // Optimize until convergence at this level
     Qstar = tryOptimizingAt(p, initialSOp);
+    if(parameters_.useHuber){ // in this case, there is no optimality verification
+      if(pMin!=pMax)
+         std::cout << "When using robust norm, Shonan only tests a single rank" << std::endl;
```
dellaert, line 808: throw

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -819,20 +835,33 @@ template class ShonanAveraging<2>;
...
                                    const Parameters &parameters)
-    : ShonanAveraging<3>(measurements, parameters) {}
+    : ShonanAveraging<3>(parameters.useHuber?
+        makeNoiseModelRobust(measurements) : measurements, parameters) {}
 
 ShonanAveraging3::ShonanAveraging3(string g2oFile, const Parameters &parameters)
-    : ShonanAveraging<3>(parseMeasurements<Rot3>(g2oFile), parameters) {}
+    : ShonanAveraging<3>(
+          parameters.useHuber
+              ? makeNoiseModelRobust(parseMeasurements<Rot3>(g2oFile))
```
dellaert, line 862: Create a parseMeasurements version that takes the flag rather than complicate the logic of these constructors

dellaert (changes requested):
You asked for my review :-) Sorry about it being late.

gtsam/sfm/ShonanAveraging.h
varunagrawal: This is the result of the formatter.

gtsam/sfm/ShonanAveraging.cpp
varunagrawal: `parseMeasurements` is a part of `dataset.cpp` whereas `makeNoiseModelRobust` is of `ShonanAveraging.h`. This kind of specialization of `parseMeasurements` seems finicky to me.

gtsam/sfm/ShonanAveraging.cpp
varunagrawal: This causes the constructor to error out when using the huber noise model. Thoughts?

varunagrawal: Also, I just noticed that the else condition doesn't assign a value to sigma. Updating it.

varunagrawal: Actually this whole if-else doesn't make any sense to me. If the noise model is required to be isotropic, shouldn't we also be checking if the robust model's noise model is isotropic?

varunagrawal:
@lucacarlone should the `Kappa` method be something like this instead?

```cpp
template <typename T>
static double Kappa(const BinaryMeasurement<T> &measurement) {
  const auto &robust =
      boost::dynamic_pointer_cast<noiseModel::Robust>(measurement.noiseModel());
  if (robust) {
    std::cout << "Verification of optimality does not work with robust cost "
                 "function"
              << std::endl;
    const auto &isotropic = boost::dynamic_pointer_cast<noiseModel::Isotropic>(
        measurement.noiseModel()->noise());
  } else {
    const auto &isotropic = boost::dynamic_pointer_cast<noiseModel::Isotropic>(
        measurement.noiseModel());
  }

  if (isotropic) {
    sigma = isotropic->sigma();
    return 1.0 / (sigma * sigma);

  } else {
    throw std::invalid_argument(
        "Shonan averaging noise models must be isotropic (but robust losses "
        "are allowed).");
  }
}
```

gtsam/sfm/ShonanAveraging.cpp
ProfFan: Try remove the autos?

varunagrawal: I don't think that will help, because there are test cases where we specifically use the huber model and this if-case will succeed. Throwing an exception will cause the test to fail.

gtsam/sfm/ShonanAveraging.cpp
dellaert: Just add it here as a local helper method

gtsam/sfm/ShonanAveraging.cpp
dellaert: You can test an exception, no? You can't check whether something prints out. Anyway, printing a warning is not the way to handle a wrong use of the module.

varunagrawal: I am not sure I understand. Shonan should still be able to work with a robust model, only losing the power to provide the certificate of optimality, right?

gtsam/sfm/ShonanAveraging.cpp
varunagrawal: I came up with something neater: `maybeRobust(measurements, useHuber)`. If `useHuber` is false, just return the measurements.

varunagrawal: This also make the constructor more functional, since we can use it across all constructors as simple as `maybeRobust(measurements)` or `maybeRobust(parseMeasurements<Rot3>(g2oFile))`.

gtsam/sfm/ShonanAveraging.cpp
dellaert: There is a ‘robust’ flag somewhere, right? If the flag is set, then robust should be allowed without a message. If it is not set it should throw. Printing out a message and then muddling on is not an option in my view.

varunagrawal: Gotcha. I did misunderstand what you meant, sorry about that.

lucacarlone: my 2 cents: it might be fishy to just disable the optimality verification when using robust loss, since the user might still think s/he is getting a certifiably-optimal solution. I suggest adding a flag "certifyOptimality" and throw an exception when certifyOptimality = true and we use a robust loss. The flag certifyOptimality might be useful also in the non-robust case since the user might want to disable the certification to save time/computation.

lucacarlone:
@varunagrawal : let me know if there are pending tasks on this PR that you want me to help with. It seems that except the "throw issue" above, everything else has been solved (thanks!).

varunagrawal:
@lucacarlone I am going to need your review since I've made quite a few changes, and wanted to cross-check with you if the overall code is indeed what you expected.

gtsam/sfm/ShonanAveraging.h
```diff
@@ -47,12 +47,16 @@ struct GTSAM_EXPORT ShonanAveragingParameters {
   using Anchor = std::pair<size_t, Rot>;
 
   // Paremeters themselves:
```
lucacarlone, line 49: Paremeters -> Parameters

also: following GaussNewtonParams, LevenbergMarquardtParams, etc, shouldn't we call this class "ShonanAveragingParams"?

lucacarlone (commented):
I mostly looked at the "robust" aspects and everything seems to be in place.

varunagrawal:
@dellaert gentle reminder.

gtsam/sfm/ShonanAveraging.h
```diff
@@ -46,13 +46,17 @@ struct GTSAM_EXPORT ShonanAveragingParameters {
...
-  double beta;                  // weight of Karcher-based prior (default 1)
-  double gamma;                 // weight of gauge-fixing factors (default 0)
+  // Parameters themselves:
+  LevenbergMarquardtParams lm;  ///< LM parameters
+  double optimalityThreshold;   ///< threshold used in checkOptimality
+  Anchor anchor;                ///< pose to use as anchor if not Karcher
+  double alpha;                 ///< weight of anchor-based prior (default 0)
+  double beta;                  ///< weight of Karcher-based prior (default 1)
+  double gamma;                 ///< weight of gauge-fixing factors (default 0)
+  ///< if enabled, the Huber loss is used (default false)
```
dellaert, line 56: `///<` is for a one-line comment *after* the declaration.

dellaert (approved):

Issue #548, opened 2020-10-01
VC 2016 compile bug?


dellaert:
## Description

It seems VC 2016 fails to complete check after @Alexma3312 ’s PR was merged:
```

Event.cpp
1114
D:\a\gtsam\gtsam\gtsam/geometry/Point3.h(69): error C2491: 'gtsam::mean': definition of dllimport function not allowed (compiling source file D:\a\gtsam\gtsam\gtsam_unstable\geometry\BearingS2.cpp) [D:\a\gtsam\gtsam\build\gtsam_unstable\gtsam_unstable.vcxproj]
1115
D:\a\gtsam\gtsam\gtsam/geometry/Point3.h(69): error C2491: 'gtsam::mean': definition of dllimport function not allowed (compiling source file D:\a\gtsam\gtsam\gtsam_unstable\geometry\Event.cpp) [D:\a\gtsam\gtsam\build\gtsam_unstable\gtsam_unstable.vcxproj]
1116
  Pose3Upright.cpp
```

Alexma3312:
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2491?view=vs-2019

dellaert:
I think the issue is the name conflict. I am fixing in a branch now...

ProfFan:
Should this be closed?

dellaert:
Fixed in #549

PR #549, opened 2020-10-01
Rename mean to means + some refactoring

dellaert:
(Hopefully) fix issue #548, as well as some re-factoring of code using Point3 pairs. 
Also two small unrelated re-factors in Pose3/Similarity3.

### DIFF
--- a/gtsam/geometry/Point3.h
+++ b/gtsam/geometry/Point3.h
@@ -67,8 +67,8 @@ GTSAM_EXPORT Point3 mean(const CONTAINER& points) {
   return sum / points.size();
 }
 
-/// mean of Point3 pair 
-GTSAM_EXPORT Point3Pair mean(const std::vector<Point3Pair>& abPointPairs);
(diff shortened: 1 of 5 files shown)


gtsam/geometry/Point3.h
```diff
@@ -68,8 +68,8 @@ GTSAM_EXPORT Point3 mean(const CONTAINER& points) {
   return sum / points.size();
 }
 
-/// mean of Point3 pair 
-GTSAM_EXPORT Point3Pair mean(const std::vector<Point3Pair>& abPointPairs);
+/// mean of Point3 pair
```
varunagrawal, line 71: `means` in the docstring, and better docstring maybe?

varunagrawal (approved):
(Very) Minor fix needed.

gtsam/geometry/Point3.h
dellaert: Done

PR #553, opened 2020-10-02
Find python if using Default

varunagrawal:
This PR updates the `FindPython` cmake call to only happen when the `Default` argument is used.


ProfFan (approved):

varunagrawal:
@johnwlambert hopefully this should fix your issue.

PR #557, opened 2020-10-06
refactor cmake scripts into smaller files

jlblancoc:
I thought this is a good step before addressing #478 , since adding support for more system libraries will add a lot of extra cmake code, so it might be better to keep it hierarchically organized, one script for each external dependency.

It passes all tests in a local build.


cmake/handle_eigen.cmake
```diff
@@ -0,0 +1,75 @@
+###############################################################################
+# Option for using system Eigen or GTSAM-bundled Eigen
+### These patches only affect usage of MKL. If you want to enable MKL, you *must*
+### use our patched version of Eigen
+### See:  http://eigen.tuxfamily.org/bz/show_bug.cgi?id=704 (Householder QR MKL selection)
+###       http://eigen.tuxfamily.org/bz/show_bug.cgi?id=705 (Fix MKL LLT return code)
```
ProfFan, line 6: These two are fixed in 2015, I am curious that are these still relevant for us?

jlblancoc: You're right. I checked it and apparently the current embedded copy of Eigen doesn't have any patch applied. 
Will remove the comments. 

There were a number of comments that looked doubtful to me. Those 100% immediately-clear wrong or outdated were removed, but I left most of them untouched in the first commit of this PR to first check if all tests passed.

jlblancoc:
An additional change: the system version of Eigen3 is used by default **if** it is successfully first found in the system. In other words: 
- Is Eigen missing? -> Quietly select embedded version.
- Is Eigen available as system package? -> Quietly select it.

In any case, one can override the selection with the same good-old cmake variable `GTSAM_USE_SYSTEM_EIGEN`.

I hope you agree with this policy.

dellaert:
> An additional change: the system version of Eigen3 is used by default **if** it is successfully first found in the system. In other words:
> 
> * Is Eigen missing? -> Quietly select embedded version.
> * Is Eigen available as system package? -> Quietly select it.
> 
> In any case, one can override the selection with the same good-old cmake variable `GTSAM_USE_SYSTEM_EIGEN`.
> 
> I hope you agree with this policy.

I think this is not a bad policy, but it is a change that could break - silently and in a way that is hard to debug - several installs of GTSAM out in the world. Regardless, I think that change should not be part of this PR. If we do it, it should be it’s own PR and widely disseminated/documented.

CMakeLists.txt
```diff
@@ -45,399 +38,21 @@ if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
...
-if(GTSAM_BUILD_PYTHON)
-    if(${GTSAM_PYTHON_VERSION} STREQUAL "Default")
-        # Get info about the Python3 interpreter
-        # https://cmake.org/cmake/help/latest/module/FindPython3.html#module:FindPython3
-        find_package(Python3 COMPONENTS Interpreter Development)
-
-        if(NOT ${Python3_FOUND})
-            message(FATAL_ERROR "Cannot find Python3 interpreter. Please install Python >= 3.6.")
-        endif()
+include(cmake/handle_boost.cmake)            # Boost
```
dellaert, line 41: I seem to remember that we used a different naming convention for cmake files, before, camelCase? Regardless, it should be uniform.

dellaert (changes requested):
Love it, but blocking for now as I think Eigen change is not kosher in this PR. After that is reversed, i will gladly approve.

CMakeLists.txt
jlblancoc: Hmmm... correct, most use camelCase. Done.

jlblancoc:
> Love it, but blocking for now as I think Eigen change is not kosher in this PR. After that is reversed, i will gladly approve.

Reverted too.

dellaert (approved):

jlblancoc:
On Wed, Oct 7, 2020 at 5:06 PM Frank Dellaert <notifications@github.com> wrote:
>
> @dellaert approved this pull request.


Ok, thanks.
This one can be merged now...

dellaert:
Thanks again !!!!!

PR #558, opened 2020-10-07
Fix Python serialization

ProfFan:
To unblock GTSFM.


dellaert (approved):

dellaert:
Python CI fails?

ProfFan:
@dellaert I think it's timing out, I think I can deal with that later

PR #556, opened 2020-10-05
Add python equivalent for  SFMExample_bal.cpp

johnwlambert:
Adds Bundle Adjustment functionality to the Python wrapper.

Should provide the equivalent of the C++ example (https://github.com/borglab/gtsam/blob/develop/examples/SFMExample_bal.cpp)

Present ambiguity @dellaert  -- is
```c++
typedef gtsam::PinholeCamera<gtsam::Cal3Bundler> PinholeCameraCal3Bundler;
```
or
```c++
typedef gtsam::PinholeCamera<gtsam::Cal3Bundler> SfmCamera;
```
preferred? We use both throughout codebase currently.

Introduces:

- PriorFactor for SfmCamera
- GeneralSFMFactor<SfmCamera,Point3>
See Line 34 and Line 68 of https://github.com/borglab/gtsam/blob/develop/examples/SFMExample_bal.cpp

A later PR should include
- SfmData -> set up of collection is considered part of “initialization”. Can add .push_back() method to this struct, Unit tests of newly wrapped functionality

PR #560, opened 2020-10-10
Fix LLVM repo keys

ProfFan:
This should fix the problem in https://github.com/borglab/gtsam/pull/556/checks?check_run_id=1234393111

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -52,6 +52,7 @@ jobs:
           if [ "${{ matrix.compiler }}" = "clang" ] && [ "${{ matrix.version }}" = "9" ]; then
             # pool.sks-keyservers.net is the SKS GPG global keyserver pool
             # 15CF4D18AF4F7421 is the GPG key for the LLVM apt repository
+            # This key is not in the keystore by default for Ubuntu so we need to add it.
             gpg --keyserver pool.sks-keyservers.net --recv-key 15CF4D18AF4F7421
             gpg -a --export 15CF4D18AF4F7421 | sudo apt-key add -
             sudo add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main"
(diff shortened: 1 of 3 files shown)


ProfFan:
@johnwlambert Please review this :)

johnwlambert:
Looks good, thanks @ProfFan for adding this fix.

Would you mind adding a few more inline comments before we merge, just to increase the readability? Perhaps something like
```python
# We need to use a keyserver because.... we use this keyserver because...
# matrix.version 9 represents ...
# 15CF4D18AF4F7421 is the hash for ...
if [ "${{ matrix.compiler }}" = "clang" ] && [ "${{ matrix.version }}" = "9" ]; then
            gpg --keyserver pool.sks-keyservers.net --recv-key 15CF4D18AF4F7421
            gpg -a --export 15CF4D18AF4F7421 | sudo apt-key add -
```

ProfFan:
I think as this is under clang if it's pretty self-explanatory...

dellaert:
> I think as this is under clang if it's pretty self-explanatory...

Please be responsive to the reviewer's comment, rather than dismissive.

dellaert (changes requested):
Please implement John's comment

ProfFan:
@johnwlambert @dellaert done

johnwlambert (approved):

ProfFan:
@dellaert Please accept :)

dellaert:
> @dellaert Please accept :)

Did you actually read John’s comment? Maybe I’m just missing where you explain why we need a key server etc as he asked…

johnwlambert:
Hi Fan, we need this to merge the SFM Example BAL updates, I think a brief comment about why we needed the keyserver would help. You could also consider making that hash key a bash variable, so that is a bit easier to read

ProfFan:
@dellaert I think I have addressed @johnwlambert 's comments, could you approve this PR so we can merge?

dellaert (approved):

PR #563, opened 2020-10-12
Fix warning on clang

dellaert:
Clang started spewing out warnings after cmake changes, because of this new flag. Makes me think there is some conditional logic missing :-) I commented out for now but suggest @jlblancoc takes over this PR and branch.

### DIFF
--- a/cmake/GtsamBuildTypes.cmake
+++ b/cmake/GtsamBuildTypes.cmake
@@ -1,3 +1,5 @@
+include(CheckCXXCompilerFlag) # for check_cxx_compiler_flag()
+
 # Set cmake policy to recognize the AppleClang compiler
 # independently from the Clang compiler.
 if(POLICY CMP0025)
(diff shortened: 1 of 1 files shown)


jlblancoc:
Copy that

jlblancoc:
I can't reproduce with my default clang version. 
After `cmake .`, what are your values for:  CMAKE_CXX_COMPILER_ID and CMAKE_CXX_COMPILER_VERSION? (first two lines of configuration list at the end)

dellaert:
> I can't reproduce with my default clang version.
> After `cmake .`, what are your values for: CMAKE_CXX_COMPILER_ID and CMAKE_CXX_COMPILER_VERSION? (first two lines of configuration list at the end)

Yeah, this is latest Xcode

```
Apple clang version 12.0.0 (clang-1200.0.32.2)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
```

Which file can I find those flags in? grep yields no results

jlblancoc:
> Clang started spewing out warnings after cmake changes, because of this new flag. Makes me think there is some conditional logic missing :-) I commented out for now but suggest @jlblancoc takes over this PR and branch.

Perhaps I misunderstood your problem from the beginning: is the problem that clang doesn't recognize the flag `-Wsuggest-override` (hence you commented it out), or that the flag is doing its job and detecting warnings in many places? 

If it's the latter, we should fix it by finding out what are the offending headers and either fixing them by adding the `override` if they belong to gtsam, or including them with `-isystem` instead of `-I` if they are 3rd party code.

dellaert:
The flag is not recognized...

jlblancoc:
@dellaert This should work now...

dellaert:
Yay, thanks! BTW I might be less responsive (to anything) until CVPR deadline.

dellaert:
Note I cannot merge I because I was the original author of this PR :-) could you approve?

jlblancoc (approved):

jlblancoc:
Done. Feel free to merge

dellaert:
Done, and thanks!!!

PR #564, opened 2020-10-12
Fix/plane factor

dellaert:
Adding unit test by Marco (@mcamurri) to debug issue #561 

Note it does not seem to be the adding of a second factor that causes the problem. For example, commenting out both plane priors but adding two plane factors works. In fact *and this is crazy* just uncommenting the prior on P2 makes the solver fail. This is crazy as the prior is not even supposed to play a role when we eliminate P1.

I'm stumped. Could be a subtle memory bug in OrientedPlane3 or any of the derivatives, but I checked at least some of the prime suspects and did not see anything obviously wrong.

More investigation needed


mcamurri:
Thank you very much @dellaert for considering the issue. I will try this PR very soon. All the credit for the discovery goes to @dwisth who found the issue and wrote the initial version of this test. I just played with it for a while.

gtsam/slam/tests/testOrientedPlane3Factor.cpp
```diff
@@ -170,6 +174,75 @@ TEST( OrientedPlane3DirectionPrior, Constructor ) {
...
+    initialEstimate.insert(P(2), p2);
+    initialEstimate.insert(X(0), x0);
+
+    GaussNewtonParams params;
+    GTSAM_PRINT(graph);
+    Ordering ordering = list_of(P(1))(P(2))(X(0));  // make sure P1 eliminated first
+    params.setOrdering(ordering);
+    params.setLinearSolverType("SEQUENTIAL_QR");  // abundance of caution
+    params.setVerbosity("TERMINATION");  // show info about stopping conditions
+    GaussNewtonOptimizer optimizer(graph, initialEstimate);
```
ProfFan, line 238: @dellaert `param` is not passed in there so it defaults to MULTIFRONTAL_CHOLESKY regardless of the above. I fixed that and the optimization converged.

ProfFan: BTW, `MULTIFRONTAL_QR` also works.

dellaert: Doh!!! That means it’s not memory, probably, just extremely unconditioned? You could get the Hessian and do SVD to check.

dellaert:
@mcamurri ?

mcamurri:
Hi @dellaert, we had to put this issue aside for a while but we are back on it. We'll share some updates soon. @dwisth

mcamurri:
See https://github.com/borglab/gtsam/pull/680 for further developments

dellaert:
I merged #680 and #362 in here

dwisth:
Just an update, I've been looking into this and have had some luck with the "relative plane factor" as described in [Kaess2015](https://www.cs.cmu.edu/~kaess/pub/Kaess15icra.pdf) (Equation 25). This is basically a ternary factor, with a local linearisation point for the plane. 

It seems to work well for unit tests and some real datasets. I'm just polishing it off, doing some more testing, and will try to push soon.

dellaert:
That makes a lot of sense, given our discussion. I'll assume you're adding a factor. And that we can also remove the one "error" function in favor of one with analytical derivatives :-)

dwisth:
I just:

- Added the new plane factor, called `LocalOrientedPlane3Factor`.
- Removed the `Unit3::error()` function as you suggested, and switched all functions that were using this to `Unit3::errorVector()`. 

The only thing I'm stuck on now is the `OrientedPlane3 lm_rotation_error` unit test. I can't quite figure out why it doesn't optimize correctly (this is the same reason that I suggested switching to `Unit3::localCoordinates` instead of `Unit3::errorVector`). It might be something particular to the unit test but I'm not quite sure what it is. Any ideas?

gtsam/geometry/OrientedPlane3.cpp
```diff
@@ -17,8 +17,10 @@
  * @brief  A plane, represented by a normal direction and perpendicular distance
  */
 
+#include <gtsam/base/numericalDerivative.h>
```
dellaert, line 20: Do we still need this?

gtsam/geometry/OrientedPlane3.h
```diff
@@ -24,6 +24,9 @@
 #include <gtsam/geometry/Unit3.h>
 #include <gtsam/geometry/Pose3.h>
 
+#include <algorithm>
```
dellaert, line 27: Move to cpp?

gtsam/geometry/Unit3.cpp
```diff
@@ -236,7 +226,7 @@ Vector2 Unit3::errorVector(const Unit3& q, OptionalJacobian<2, 2> H_p,
 /* ************************************************************************* */
 double Unit3::distance(const Unit3& q, OptionalJacobian<1, 2> H) const {
   Matrix2 H_xi_q;
-  const Vector2 xi = error(q, H ? &H_xi_q : nullptr);
+  const Vector2 xi = errorVector(q, boost::none, H ? &H_xi_q : nullptr);
```
dellaert, line 229: I think this PR should not modify Unit3, right?

gtsam/geometry/tests/testUnit3.cpp
```diff
@@ -487,10 +463,10 @@ TEST(Unit3, ErrorBetweenFactor) {
 TEST(Unit3, CopyAssign) {
   Unit3 p{1, 0.2, 0.3};
 
-  EXPECT(p.error(p).isZero());
+  EXPECT(p.errorVector(p).isZero());
```
dellaert, line 466: Same, Unit3 should not be touched in this PR I think

gtsam/navigation/AttitudeFactor.cpp
```diff
@@ -29,13 +29,13 @@ Vector AttitudeFactor::attitudeError(const Rot3& nRb,
     Matrix23 D_nRef_R;
     Matrix22 D_e_nRef;
     Unit3 nRef = nRb.rotate(bRef_, D_nRef_R);
-    Vector e = nZ_.error(nRef, D_e_nRef);
+    Vector e = nZ_.errorVector(nRef, boost::none, D_e_nRef);
```
dellaert, line 32: Not this PR -> restrict to plane only

gtsam/slam/RotateFactor.h
```diff
@@ -104,7 +104,7 @@ class RotateDirectionsFactor: public NoiseModelFactor1<Rot3> {
   /// vector of errors returns 2D vector
   Vector evaluateError(const Rot3& iRc, boost::optional<Matrix&> H = boost::none) const override {
     Unit3 i_q = iRc * c_z_;
-    Vector error = i_p_.error(i_q, H);
+    Vector error = i_p_.errorVector(i_q, boost::none, H);
```
dellaert, line 107: Not this PR

dellaert (commented):
I think we should restrict this PR to the plane factors only, no changes are needed to Unit3 and other factors that derive from it. Regarding the unit test that doesn’t pass, if this is not important to your current use case, I would just comment it out. I looked at it as well and I could not figure out what was wrong either.

gtsam/geometry/OrientedPlane3.cpp
dwisth: no, I'm removing this

gtsam/geometry/OrientedPlane3.h
dwisth: not required any more, removing `#include`

gtsam/geometry/Unit3.cpp
dwisth: I've reverted the changes to `Unit3::error()` that I made

dellaert:
Cool, thanks !
@ProfFan can you review and approve? (I created PR and can't approve)
One thing that still is not working is CI: still the failing test?

gtsam/geometry/OrientedPlane3.cpp
```diff
@@ -42,53 +43,47 @@ OrientedPlane3 OrientedPlane3::transform(const Pose3& xr, OptionalJacobian<3, 3>
...
-}
-
-/* ************************************************************************* */
-Vector3 OrientedPlane3::errorVector(const OrientedPlane3& other, OptionalJacobian<3, 3> H1,
+Vector3 OrientedPlane3::errorVector(const OrientedPlane3& other,
+                                    OptionalJacobian<3, 3> H1,
                                     OptionalJacobian<3, 3> H2) const {
   Matrix22 H_n_error_this, H_n_error_other;
-  Vector2 n_error = n_.errorVector(other.normal(), H1 ? &H_n_error_this : 0,
+  Vector2 n_error = n_.errorVector(other.n_, H1 ? &H_n_error_this : 0,
```
ProfFan, line 65: Wonder why this change, but feel free to ignore.

gtsam/slam/tests/testOrientedPlane3Factor.cpp
```diff
@@ -32,95 +36,125 @@ using namespace std;
...
 
   // Given two noisy measurements of equal weight, expect result between the two
   OrientedPlane3 expected_plane_landmark(-1.0, 0.0, 0.0, 2.0);
   EXPECT(assert_equal(optimized_plane_landmark, expected_plane_landmark));
 }
 
-// *************************************************************************
+// // *************************************************************************
+/*
 TEST (OrientedPlane3Factor, lm_rotation_error) {
```
ProfFan, line 87: Maybe add an explanation why this is commented out.

gtsam/slam/tests/testOrientedPlane3Factor.cpp
ProfFan: I didn't follow up on this, is this still relevant?

gtsam_unstable/slam/tests/testLocalOrientedPlane3Factor.cpp
```diff
@@ -0,0 +1,240 @@
...
+  auto optimized_plane_landmark = result_values.at<OrientedPlane3>(P(0));
+
+  // Given two noisy measurements of equal weight, expect result between the two
+  OrientedPlane3 expected_plane_landmark(-1.0, 0.0, 0.0, 2.0);
+  EXPECT(assert_equal(optimized_plane_landmark, expected_plane_landmark));
+}
+
+// *************************************************************************
+/*
+TEST (LocalOrientedPlane3Factor, lm_rotation_error) {
```
ProfFan, line 82: same, why this is commented out?

ProfFan (approved):
In general LGTM, but I think it's better if there are inline comments on why tests are disabled.

gtsam/slam/tests/testOrientedPlane3Factor.cpp
dwisth: No, we've now changed and removed this part of the code. The fix is implemented in the new `LocalOrientedPlane3Factor`.

gtsam_unstable/slam/tests/testLocalOrientedPlane3Factor.cpp
dwisth: Both of these unit tests are commented out because they fail, but I can't figure out why - see my comments [here](https://github.com/borglab/gtsam/pull/564#issuecomment-779290861). @ProfFan - any intuition as to why these tests might be failing?

dellaert: @ProfFan is super-busy with paper deadline and has probably no Plane3 experience. I think he is just suggesting you add a comment as to why we commented this out, and add a TODO to figure out the reason later. Then we can land this PR after CI runs.

dwisth: No problem, I just pushed some comments for these commented out unit tests

dellaert:
Merged, thanks a lot all!!!!

hunkyu:
Has this problem been solved? Which version is the final version? I found that gtsam 4.0.3 still has problems @dellaert

PR #567, opened 2020-10-15
Fix inteminent sks keyserver failure

ProfFan:
Hope CI works

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -53,8 +53,9 @@ jobs:
             # pool.sks-keyservers.net is the SKS GPG global keyserver pool
             # 15CF4D18AF4F7421 is the GPG key for the LLVM apt repository
             # This key is not in the keystore by default for Ubuntu so we need to add it.
-            gpg --keyserver pool.sks-keyservers.net --recv-key 15CF4D18AF4F7421
-            gpg -a --export 15CF4D18AF4F7421 | sudo apt-key add -
+            LLVM_KEY=15CF4D18AF4F7421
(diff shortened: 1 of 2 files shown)


johnwlambert:
Thanks @ProfFan . Any idea why 2 of the checks were cancelled?

.github/workflows/build-linux.yml
```diff
@@ -53,8 +53,9 @@ jobs:
             # pool.sks-keyservers.net is the SKS GPG global keyserver pool
             # 15CF4D18AF4F7421 is the GPG key for the LLVM apt repository
             # This key is not in the keystore by default for Ubuntu so we need to add it.
-            gpg --keyserver pool.sks-keyservers.net --recv-key 15CF4D18AF4F7421
-            gpg -a --export 15CF4D18AF4F7421 | sudo apt-key add -
+            LLVM_KEY=15CF4D18AF4F7421
+            gpg --keyserver ipv4.pool.sks-keyservers.net --recv-key $LLVM_KEY || gpg --keyserver ha.pool.sks-keyservers.net --recv-key $LLVM_KEY
```
johnwlambert, line 57: Thanks for this update Fan. Do you mind adding a brief note about what ipv4 represents, vs. the pool we used beforehand?

johnwlambert (changes requested):

johnwlambert:
Will they time out for all future branches, as well? Can we increase the timeout length, or should we remove those 2 from the CI?

ProfFan:
This is not related to the macOS timeouts, GitHub uses MacStadium for the macOS CI, and they will terminate your long-running jobs at random.

.github/workflows/build-linux.yml
ProfFan: done

johnwlambert (approved):

PR #569, opened 2020-10-18
Feature/wrap sfm data

swarrier246:
I've written a python unittest for the functions I've used from the SfmData and SfmTrack wrapper. Will add tests for other methods after I've had some doubts clarified.
Changed "measurements" in SfmTrack to Measurements, to maintain uniformity in how vectors are named (specializations.h). That change was also reflected in all other files using track.measurements


gtsam/slam/dataset.h
```diff
@@ -219,15 +220,21 @@ struct SfmTrack {
   SfmTrack(): p(0,0,0) {}
   Point3 p; ///< 3D position of the point
   float r, g, b; ///< RGB color of the 3D point
-  std::vector<SfmMeasurement> measurements; ///< The 2D image projections (id,(u,v))
+  std::vector<SfmMeasurement> Measurements; ///< The 2D image projections (id,(u,v))
```
akshay-krishnan, line 223: I think this should be renamed to `measurements`. The convention is to use small letters to start member variable names.

gtsam/gtsam.i
```diff
@@ -2759,21 +2759,35 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 };
 
 #include <gtsam/slam/dataset.h>
+// Dummy classes, for MATLAB wrappers
+class SfmMeasurement{};
+class SiftIndex{ };
+class SfmMeasurements{};
+class SfmCamera{};
```
akshay-krishnan, line 2766: I think the MATLAB wrapper also needs them to have basic methods to do anything with them. Like at() or push_back() for vectors. That is the case for other vectors that have been wrapped in this file.

gtsam/gtsam.i
```diff
@@ -2759,21 +2759,35 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
...
+
 class SfmTrack {
+  SfmTrack();
   Point3 point3() const;
   size_t number_measurements() const;
-  pair<size_t, gtsam::Point2> measurement(size_t idx) const;
-  pair<size_t, size_t> siftIndex(size_t idx) const;
+  void setP(gtsam::Point3& p_);
+  gtsam::SfmMeasurement measurement(size_t idx) const;
+  gtsam::SiftIndex siftIndex(size_t idx) const;
```
akshay-krishnan, line 2774: Do the existing versions of these methods not work? Is there a reason for the change?

gtsam/slam/dataset.h
```diff
@@ -236,13 +243,23 @@ struct SfmTrack {
...
+
 };
 
+
 /// Define the structure for the camera poses
 typedef PinholeCamera<Cal3Bundler> SfmCamera;
 
 /// Define the structure for SfM data
 struct SfmData {
+  std::vector<SfmMeasurement> Measurements;
```
akshay-krishnan, line 262: Could we use lower case `measurements` for the member name here as well?

akshay-krishnan: Could you also please document what these measurements are? I understand the one in SfmTrack, but what does the one in SfmData represent?

gtsam/gtsam.i
akshay-krishnan, line 2766: I dont think we need to wrap SfmCamera, this is being wrapped as PinholeCameraCal3Bundler in another PR, and there are a few methods in other classes that use it. Please refer to https://github.com/borglab/gtsam/pull/556. Perhaps we can merge that PR if it is needed for this code.

python/gtsam/specializations.h
```diff
@@ -13,3 +13,6 @@ py::bind_vector<std::vector<gtsam::BinaryMeasurement<gtsam::Unit3> > >(m_, "Bina
 py::bind_map<gtsam::IndexPairSetMap>(m_, "IndexPairSetMap");
 py::bind_vector<gtsam::IndexPairVector>(m_, "IndexPairVector");
 py::bind_map<gtsam::KeyPairDoubleMap>(m_, "KeyPairDoubleMap");
+py::bind_vector<std::vector<gtsam::SfmMeasurement> >(m_, "Measurement");
+py::bind_vector<std::vector<gtsam::SiftIndex> >(m_, "SiftIndexVector");
+py::bind_vector<std::vector<gtsam::SfmCamera> >(m_, "cameras");
```
akshay-krishnan, line 18: Do we need vector cameras? Maybe we could call them VectorPinholeCameraCal3Bundler after that PR has been merged. Or maybe in that PR itself, if we need them.

akshay-krishnan (changes requested):

gtsam/slam/dataset.h
```diff
@@ -260,6 +277,14 @@ struct SfmData {
   SfmTrack track(size_t idx) const {
     return tracks[idx];
   }
+
+  void add_track(SfmTrack t) {
+    tracks.push_back(t);
+  }
+
+  void add_camera(SfmCamera cam){
```
akshay-krishnan, line 285: I think this should be a const reference, like `void add_camera(const SfmCamera& cam)`.

gtsam/slam/dataset.h
```diff
@@ -260,6 +277,14 @@ struct SfmData {
   SfmTrack track(size_t idx) const {
     return tracks[idx];
   }
+
+  void add_track(SfmTrack t) {
```
akshay-krishnan, line 281: Please change the argument to a const reference, unless there's a good reason not to.

gtsam/slam/dataset.h
```diff
@@ -236,13 +243,23 @@ struct SfmTrack {
   Point3 point3() const {
     return p;
   }
+  void add_measurement(pair<size_t, gtsam::Point2> m) {
```
akshay-krishnan, line 246: Please use const reference arguments.

akshay-krishnan (changes requested):

gtsam/gtsam.i
swarrier246: i haven't implemented the matlab wrapper yet. These are dummy classes

gtsam/slam/dataset.h
swarrier246: it's the same: measurements within the tracks. But I'll update that in documentation

python/gtsam/preamble.h
```diff
@@ -10,3 +10,6 @@ PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Pose3>);
 PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose3> > >);
 PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose2> > >);
 PYBIND11_MAKE_OPAQUE(std::vector<gtsam::IndexPair>);
+PYBIND11_MAKE_OPAQUE(std::vector<gtsam::SfmMeasurement>);
```
johnwlambert, line 13: why do we need to add these here? Didn't catch what `PYBIND11_MAKE_OPAQUE` does for us

gtsam/gtsam.i
johnwlambert: i propose we come up with a better name than `siftIndex` while we are at it here, since we won't to be using non-SIFT detections, descriptions, and correspondences very soon

swarrier246: do we need them at all? Because to my understanding, SiftIndex seems to be the feature coordinates at a particular measurement - then why is SfmMeasurement a pair of <size_t, Point2>? Why not <size_t, SiftIndex>?

python/gtsam/preamble.h
swarrier246: This explains it well: https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html

gtsam/slam/dataset.h
```diff
@@ -236,13 +243,23 @@ struct SfmTrack {
...
+    Measurements.push_back(m);
+  }
+
+  SfmMeasurements& measurements() {
```
akshay-krishnan, line 250: Could you point me to where this method is being used? I'm not sure why it is returned by reference.

swarrier246: Nowhere, actually. It was originally intended to return all measurements in a track(I wanted to avoid using a for loop to get measurements, but it returns an SfmMeasurement object, which we'd need to wrap to get the elements of, so I decided not to use it)

python/CMakeLists.txt
```diff
@@ -36,6 +36,9 @@ set(ignore
     gtsam::BetweenFactorPose3s
     gtsam::Point2Vector
     gtsam::Pose3Vector
+    gtsam::SfmMeasurement
+    gtsam::SfmCamera
+    gtsam::SiftIndex
```
akshay-krishnan, line 41: All three of these are not being wrapped. I think they can be removed.

python/gtsam/preamble.h
```diff
@@ -10,3 +10,5 @@ PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Pose3>);
 PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose3> > >);
 PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose2> > >);
 PYBIND11_MAKE_OPAQUE(std::vector<gtsam::IndexPair>);
+//PYBIND11_MAKE_OPAQUE(std::vector<gtsam::SfmMeasurement>);
```
akshay-krishnan, line 13: Remove commented lines?

python/gtsam/specializations.h
```diff
@@ -13,3 +13,5 @@ py::bind_vector<std::vector<gtsam::BinaryMeasurement<gtsam::Unit3> > >(m_, "Bina
 py::bind_map<gtsam::IndexPairSetMap>(m_, "IndexPairSetMap");
 py::bind_vector<gtsam::IndexPairVector>(m_, "IndexPairVector");
 py::bind_map<gtsam::KeyPairDoubleMap>(m_, "KeyPairDoubleMap");
+//py::bind_vector<std::vector<gtsam::SfmMeasurement> >(m_, "Measurement");
```
akshay-krishnan, line 16: Remove commented lines?

python/gtsam/tests/test_SfmData.py
```diff
@@ -0,0 +1,79 @@
...
+    def setUp(self):
+        """Initialize SfmData and SfmTrack"""
+        self.data = gtsam.SfmData()
+        self.tracks = gtsam.SfmTrack()
+
+    def test_tracks(self):
+        """Test functions in SfmTrack"""
+        # measurement is of format (camera_idx, imgPoint)
+        # create camera indices for two cameras
+        i1, i2 = np.random.randint(5), np.random.randint(5)
```
akshay-krishnan, line 36: Is the randomness of any use here? You might as well pick random numbers yourself. It helps to have determinism in tests.

python/gtsam/tests/test_SfmData.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+
+    def test_data(self):
+        """Test functions in SfmData"""
+        #cam1 = gtsam.SfmCamera(1500, 1200, 0, 640, 480)
+        # Create new track with 3 measurements
+        track2 = gtsam.SfmTrack()
+        i1, i2, i3 = np.random.randint(5), np.random.randint(5), np.random.randint(5)
+        uv_i1 = gtsam.Point2(np.random.randint(5), np.random.randint(5))
```
akshay-krishnan, line 63: I don't think the randomness here is particularly useful. It doesn't hurt much either so its up to you. Also, point2 can be double, so why use randint?

python/gtsam/tests/test_SfmData.py
```diff
@@ -0,0 +1,79 @@
+"""
+GTSAM Copyright 2010-2019, Georgia Tech Research Corporation,
+Atlanta, Georgia 30332-0415
+All Rights Reserved
+
+See LICENSE for the license information
+
+Unit tests for testing dataset access.
+Author: Frank Dellaert (Python: Sushmita Warrier)
```
akshay-krishnan, line 9: Ive not seen this kind of separation in author names before. Totally fine if it has been done elsewhere. Otherwise, I think you can list names in the order of the contribution to this file?

akshay-krishnan (changes requested):
Looking much cleaner now! Leaving some final comments.

python/gtsam/tests/test_SfmData.py
swarrier246: I think I based this off of John's PR #556 . Let me know if you still want me to change it

python/gtsam/tests/test_SfmData.py
swarrier246: No particular reason, just seemed simpler. I'll add arbitrary numbers instead of the np.random at all places

python/gtsam/tests/test_SfmData.py
akshay-krishnan: No, its up to you.

python/gtsam/tests/test_SfmData.py
```diff
@@ -0,0 +1,81 @@
...
+        self.tracks.set_point3(gtsam.Point3(2.5, 3.3, 1.2))
+        np.testing.assert_array_almost_equal(
+            gtsam.Point3(2.5,3.3,1.2), 
+            self.tracks.point3()
+        )
+
+
+    def test_data(self):
+        """Test functions in SfmData"""
+        #cam1 = gtsam.SfmCamera(1500, 1200, 0, 640, 480)
```
akshay-krishnan, line 60: Commented code, can be removed.

python/gtsam/tests/test_SfmData.py
```diff
@@ -0,0 +1,81 @@
...
+# pylint: disable=invalid-name, no-name-in-module, no-member
+
+from __future__ import print_function
+
+import unittest
+
+import numpy as np
+
+import gtsam
+#from gtsam import SfmCamera
```
akshay-krishnan, line 20: Commented code, can be removed.

python/gtsam/tests/test_SfmData.py
```diff
@@ -0,0 +1,81 @@
...
+from gtsam.utils.test_case import GtsamTestCase
```
akshay-krishnan, line 21: I am not sure if this is standard practice for importing in Python. Hasn't gtsam already been imported above?

python/gtsam/preamble.h
```diff
@@ -9,4 +9,4 @@ PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Point2, Eigen::aligned_allocator<gtsam::
 PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Pose3>);
 PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose3> > >);
 PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose2> > >);
-PYBIND11_MAKE_OPAQUE(std::vector<gtsam::IndexPair>);
+PYBIND11_MAKE_OPAQUE(std::vector<gtsam::IndexPair>);
```
akshay-krishnan, line 12: Did you change any spacing or remove a new line here? Undoing that would remove this file from the PR.

gtsam/gtsam.i
```diff
@@ -2759,21 +2759,29 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
...
   size_t number_cameras() const;
   size_t number_tracks() const;
   gtsam::PinholeCamera<gtsam::Cal3Bundler> camera(size_t idx) const;
   gtsam::SfmTrack track(size_t idx) const;
+  void add_track(const gtsam::SfmTrack& t) ;
+  void add_camera(const gtsam::SfmCamera& cam);
 };
 
 gtsam::SfmData readBal(string filename);
+bool writeBAL(string filename, gtsam::SfmData& data);
```
akshay-krishnan, line 2784: It looks like the filename argument is a `const std::string& ` in dataset.h

akshay-krishnan (changes requested):

python/gtsam/tests/test_SfmData.py
swarrier246: Yeah, other tests have done it too (refer https://github.com/borglab/gtsam/blob/develop/python/gtsam/tests/test_Point2.py)

gtsam/gtsam.i
swarrier246: that is repeated across a lot of functions(eg readBal, load2D, etc). I'm not sure why it is being used the way it is. Maybe @ProfFan could shed some light on it?

ProfFan: I think this is just history :)

dellaert: const and & are mostly ignored in .i files. Just using string is fine.

gtsam/gtsam.i
```diff
@@ -2759,21 +2759,29 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 };
 
 #include <gtsam/slam/dataset.h>
+
 class SfmTrack {
+  SfmTrack();
   Point3 point3() const;
   size_t number_measurements() const;
+  void set_point3(gtsam::Point3& p_);
```
dellaert, line 2767: I'd rather see a new constructor than an imperative "set". "add" is fine as it is "part of the construction" of the object.

gtsam/slam/dataset.h
```diff
@@ -221,10 +221,15 @@ struct SfmTrack {
   float r, g, b; ///< RGB color of the 3D point
   std::vector<SfmMeasurement> measurements; ///< The 2D image projections (id,(u,v))
   std::vector<SiftIndex> siftIndices;
+  
   /// Total number of measurements in this track
   size_t number_measurements() const {
     return measurements.size();
   }
+  /// Set 3D point
+  void set_point3(const Point3& p_){
```
dellaert, line 230: see my comment from .i file.

gtsam/slam/dataset.h
```diff
@@ -233,11 +238,17 @@ struct SfmTrack {
   SiftIndex siftIndex(size_t idx) const {
     return siftIndices[idx];
   }
+  /// Get 3D point
```
dellaert, line 241: how about returning a const& ?

gtsam/slam/dataset.h
```diff
@@ -233,11 +238,17 @@ struct SfmTrack {
...
   Point3 point3() const {
     return p;
   }
+  /// Add measurement to track
+  void add_measurement(const pair<size_t, gtsam::Point2>& m) {
+    measurements.push_back(m);
```
dellaert, line 247: can you use emplace?

dellaert (commented):
I'll just comment, as Akshay requested changes, he can approve after you make changes.

python/gtsam/tests/test_SfmData.py
```diff
@@ -0,0 +1,79 @@
...
+        # add measurements to the track
+        track2.add_measurement(m_i1)
+        track2.add_measurement(m_i2)
+        track2.add_measurement(m_i3)
+        self.data.add_track(self.tracks)
+        self.data.add_track(track2)
+        # Number of tracks in SfmData is 2
+        self.assertEqual(self.data.number_tracks(), 2)
+        # camera idx of first measurement of second track corresponds to i1
+        self.assertEqual(self.data.track(1).measurement(0)[0], i1)
```
johnwlambert, line 76: is the measurement returned as a Numpy array, rather than as a typed SfmMeasurement object? is that preferred?

johnwlambert: Discussed Pair vs. NamedTuple, but ship has sailed, resolved.

akshay-krishnan: Sorry I missed the discussion @johnwlambert . Do you think it would more readable if the tuple was unpacked locally like `cam_idx, point = tracks.measurements(0)`? Might save some effort on creating a class for the measurements and changing existing code in dataset.h.

gtsam/gtsam.i
```diff
@@ -2759,21 +2759,29 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 };
 
 #include <gtsam/slam/dataset.h>
+
 class SfmTrack {
-  Point3 point3() const;
+  SfmTrack();
+  SfmTrack(const gtsam::Point3& pt);
```
akshay-krishnan, line 2765: About this new constructor, what happens to the measurements if you use them to triangulate a point? Should the constructor not accept measurements as an argument as well?

python/gtsam/preamble.h
akshay-krishnan: This is not resolved yet.

akshay-krishnan (changes requested):

gtsam/gtsam.i
swarrier246: you mean a vector of measurements as a constructor argument?

dellaert: No, it's acceptable to add them one at a time.

akshay-krishnan (approved):
Nice work! Thanks for repeatedly making the requested changes.

ProfFan:
Mac builds are broken by upstream boost issue https://github.com/boostorg/serialization/issues/219

Issue #292, opened 2020-05-06
[Feature Request] Allow use of system libraries for 3rdparty codebases


acxz:
## Feature

While having 3rdparty codebases inside of GTSAM's codebase is very useful, such as keeping track of dependencies and building off of one version, it is also useful to allow for the use of system libraries, so that users can have consistent versions across their own projects and computer and that upstream can be followed readily for bug fixes/new features.

## Motivation

All 3rdparty libraries in GTSAM are included in most distributions package managers, (def in Debian/Ubuntu and Arch, most likely in Fedora as well). In some cases having the GTSAM 3rdparty libraries causes conflicts. For example the ArchLinux package has been patched so that GTSAM's metis does not conflict with the actual metis library. At least having the option to use system libraries will help alleviate this and allow more freedom on the user's end. This also decreases the GTSAM installed size as well.

## Pitch

Add flags and relevant linking code in the CMake procedure to allow for the use of system level libraries over 3rdparty libraries. Similar to the how users can use system installed eigen.

## Additional context

Most distros like to see codebases 'devendored' so as to decrease the amount of code reuse, linking errors, etc. It is also the reason why distro like to dynamically link against libraries instead of statically linking everything.

ruffsl:
This may be a good step forward for packaging and distribution. I was about to open the ticket requesting about distributing binaries via Debian packages. There's a number of slam frameworks within the ROS community that never quite get released into binary distributions of release distros for easy end user installment given their dependencies never get released/shipped.

For example, packages like SLAM-Toolbox can use multiple different backends, like Ceres, G2O, SPA, and GTSAM, but end up deprecating backends that aren't released as they can't be used and thus not as useful to maintain.

https://github.com/SteveMacenski/slam_toolbox

There are also some impressive VIO SLAM packages I'd like to see disrupt the robotic software stack, but the usually get stunted to user-exercises in source builds before they ever get community traction. I'd like to see Kimera-VIO make it into a ROS release, but that will require a number of upstream dependencies to be released as well, including GTSAM.

https://github.com/MIT-SPARK/Kimera-VIO

I see there's a debian folder here that was last touched a while ago. In the interim of getting a full-blown Debian package released upstream, merely providing Debian packages within the ROS repository via bloom and the ROS build farm maybe a simpler/quicker method of distributing debians for downstream libraries.

https://index.ros.org/doc/ros2/Tutorials/Releasing-a-ROS-2-package-with-bloom

acxz:
[...]

PR #570, opened 2020-10-19
Use METIS system library if so selected

jlblancoc:
This addresses the libmetis part of #292.

Defines a new cmake flag `GTSAM_USE_SYSTEM_METIS` (default=off) and uses system libraries in GNU/Linux systems if set to ON. This has been tested with `METISOrderingExample` for both configurations. 

Note: the current CI won't test the GTSAM_USE_SYSTEM_METIS=ON case... 
Perhaps one of the cases in the test matrix should address using all system libraries (Eigen, Metis,...)?

(cc: @berndpfrommer )


acxz (commented):
@jlblancoc 

glad to see you working on this! Had some questions about this though.

Shouldn't https://github.com/borglab/gtsam/blob/develop/gtsam/inference/Ordering.cpp#L28
be changed to just `<metis.h>`
as well as https://github.com/borglab/gtsam/blob/develop/gtsam_unstable/partition/FindSeparator-inl.h#L25 being changed to `<metislib.h>` in this PR?

jlblancoc:
> @jlblancoc
> 
> glad to see you working on this! Had some questions about this though.
> 
> Shouldn't https://github.com/borglab/gtsam/blob/develop/gtsam/inference/Ordering.cpp#L28
> be changed to just `<metis.h>`
> as well as https://github.com/borglab/gtsam/blob/develop/gtsam_unstable/partition/FindSeparator-inl.h#L25 being changed to `<metislib.h>` in this PR?

Good catch! :+1:

acxz:
based on my past effort on this task, I fear you might not be able to use `<metislib.h>`. Re: https://github.com/borglab/gtsam/issues/292#issuecomment-656276457

varunagrawal (approved):
LGTM. I am blown away by your CMake skills @jlblancoc!

varunagrawal:
We should create an issue for @acxz's comments and tackle them in the next PR. This PR is atomically very nice.

acxz:
I don't want to be a downer but I don't think this PR should be merged before addressing the issues I have raised.
For one we are still using the older include path which contains the 3rdparty path. This means that even if I am setting the `SYSTEM_METIS` flag to be on, I still won't be using the metis system include file.
Even if that is changed, system metis does not provide `metislib.h` which means we won't be able to use the functionality of it. Thus rendering the `SYSTEM_METIS` flag broken. If its going to be broken when merged in, we should just not merge it in.

varunagrawal:
Fair point. I'm going to leave this as approved and trust @jlblancoc to add in the updates before we merge.

varunagrawal (changes requested):
We need to make the changes as recommended by @acxz.

varunagrawal:
@jlblancoc I'm merging this. 🙂

jlblancoc:
> Let's merge? @jlblancoc please do the honors.

@varunagrawal  Sorry, I've been busy and missed a *lot* of what's happening here... so glad you guys made it work :-)

PR #572, opened 2020-10-23
Add alternativeName() for symbols

jlblancoc:
What do you think of this feature? 

It allows writing code with symbol named according to their physical meaning. Check the example in the new unit test. 

Of course, the default KeyFormatter would still return the underlying one-character symbol name, but if someone want to have custom names printed, a custom KeyFormatter would be required.

### DIFF
--- a/gtsam/inference/Symbol.h
+++ b/gtsam/inference/Symbol.h
@@ -165,13 +165,15 @@ inline Key Z(std::uint64_t j) { return Symbol('z', j); }
 
 /** Generates symbol shorthands with alternative names different than the
  * one-letter predefined ones. */
-inline std::function<Key(std::uint64_t)> alternativeName(const char c) {
-	return [c](std::uint64_t j) { return gtsam::Symbol(c, j); };
-}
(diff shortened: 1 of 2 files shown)


dellaert (commented):
That's cool. But it took a while before I figured out what was going on :-) I think a functor might be easier to grok than a function producing a function. E.g.

```
gtsam:SymbolGenerator Z('x'), DZ('v'), DDZ('a');
```

implemented like
```
class SymbolGenerator {
const char c_;
public:
SymbolGenerator(const char c) c_(c) {} 
Symbol operator()(const std::uint64_t j) const { return Symbol(c_, j); }
};
```

What do you think?

jlblancoc:
Updated!
One remaining doubt: what would be better, `gtsam::symbol_shorthand::SymbolGenerator` or `gtsam::SymbolGenerator` ?

jlblancoc:
PS: If it's the shorter `gtsam:SymbolGenerator`, you could merge straight away.

dellaert (approved):
Cool!

PR #576, opened 2020-10-31
adding additional Unit3 support in the wrapper: PriorFactorUnit3, Values::insert/update/at(Unit3)

tmcg0:
This very simple PR adds some additional functionality for the GTSAM wrapper for Unit3 types. Specifically:
- adding PriorFactorUnit3 as a wrapper type
- Adding Values methods update(), insert(), at() that support Unit3 types.

(apologies if there is a good reason this functionality is missing, or if it's added somewhere else that I've missed)

### DIFF
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2157,6 +2157,7 @@ class Values {
   void insert(size_t j, const gtsam::SOn& P);
   void insert(size_t j, const gtsam::Rot3& rot3);
   void insert(size_t j, const gtsam::Pose3& pose3);
+  void insert(size_t j, const gtsam::Unit3& unit3);
   void insert(size_t j, const gtsam::Cal3_S2& cal3_s2);
   void insert(size_t j, const gtsam::Cal3DS2& cal3ds2);
   void insert(size_t j, const gtsam::Cal3Bundler& cal3bundler);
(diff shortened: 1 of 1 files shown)


dellaert:
This is awesome, but check why some of the CI runs fail?

tmcg0:
Hmm. The code I added was really small and shouldn't have broken anything. I'll try to dig in to why the MacOS CI fails. A quick look at the CI logs makes it look like it's something else going on.

ProfFan:
It appears to be some compatibility problem on latest boost (which is what brew tracks). I don't have bandwidth now unfortunately.

tmcg0:
> It appears to be some compatibility problem on latest boost (which is what brew tracks). I don't have bandwidth now unfortunately.

Yep. I don't have a Mac, so I likely won't be able to track this down. I'm a bit confused as how the latest commit in develop was passing. Did brew track a new boost in the last four days?

tmcg0:
The failing CI in this PR should be fixed by [the latest PR which fixed some boost compatibility in Brew](https://github.com/borglab/gtsam/pull/577). Thanks @ProfFan for fixing that!

ProfFan:
@tmcg0 Glad to help :)

Please merge in or rebase to latest develop to fix CI.

tmcg0:
Looks like one of the Windows CI timed out after 6 hours. Also looks like that's happening to other recent PRs.

ProfFan:
Since it's only the Debug version I think it's timing out. If that is the only blocker then I think it's fine.

dellaert (approved):
Thanks!!! Very useful :-) And you helped us debug the Mac issue.

PR #577, opened 2020-11-02
Homebrew latest boost has buggy boost.serialization

ProfFan:


dellaert (approved):
Let's merge if CI on Mac actually succeeds, otherwise we might have to check a bit more

dellaert:
@ProfFan : CI says: 
```
Error: Calling Non-checksummed download of boost formula file from an arbitrary URL is disabled! 
Use 'brew extract' or 'brew create' and 'brew tap-new' to create a formula file in a tap on GitHub instead.
```

PR #579, opened 2020-11-03
Consistent interface for pixel center

varunagrawal:
Defined u0 and v0 in all camera models for consistent interface.

### DIFF
--- a/gtsam/geometry/Cal3Bundler.h
+++ b/gtsam/geometry/Cal3Bundler.h
@@ -98,11 +98,23 @@ public:
     return k2_;
   }
 
+  /// image center in x
+  inline double px() const {
+    return u0_;
+  }
+
+  /// image center in y
+  inline double py() const {
+    return v0_;
+  }
+
+  /// @deprecated
   /// get parameter u0
   inline double u0() const {
     return u0_;
   }
 
+  /// @deprecated
   /// get parameter v0
   inline double v0() const {
     return v0_;
--- a/gtsam/geometry/Cal3Bundler.h
+++ b/gtsam/geometry/Cal3Bundler.h
@@ -108,17 +108,17 @@ public:
     return v0_;
   }
 
-  /// @deprecated
+#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V41
   /// get parameter u0
   inline double u0() const {
     return u0_;
   }
 
-  /// @deprecated
   /// get parameter v0
   inline double v0() const {
     return v0_;
   }
+#endif
 
 
   /**
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2766,7 +2766,9 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 class SfmTrack {
   SfmTrack();
 
-  double r, g, b;
+  double r;
+  double g;
+  double b;
   // TODO Need to close wrap#10 to allow this to work.
   // std::vector<pair<size_t, gtsam::Point2>> measurements;
 
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -849,8 +849,6 @@ class Cal3_S2 {
   double skew() const;
   double px() const;
   double py() const;
-  double u0() const;
-  double v0() const;
   gtsam::Point2 principalPoint() const;
   Vector vector() const;
   Matrix K() const;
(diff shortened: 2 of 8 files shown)


dellaert (changes requested):
There is already px and py ?? I do not support to duplicate versions of all these things

varunagrawal:
> There is already px and py ?? I do not support to duplicate versions of all these things

`Cal3Bundler` doesn't have `px` or `py` so this causes undefined function errors when working with different camera models (e.g. using `Cal3_S2` in the main code and then loading from a `Cal3Bundler` from a BAL file. I figured having `u0` and `v0` made more sense, though I can make Cal3Bundler have `px` & `py` instead?

swarrier246:
> > There is already px and py ?? I do not support to duplicate versions of all these things
> 
> `Cal3Bundler` doesn't have `px` or `py` so this causes undefined function errors when working with different camera models (e.g. using `Cal3_S2` in the main code and then loading from a `Cal3Bundler` from a BAL file. I figured having `u0` and `v0` made more sense, though I can make Cal3Bundler have `px` & `py` instead?

Wouldn't it be advisable to remove the px and py in the other models and reflect those changes across all places that use it, then? Might cause some avoidable confusion otherwise.

gtsam/gtsam.i
```diff
@@ -597,6 +597,7 @@ class Rot3 {
   Rot3(double R11, double R12, double R13,
       double R21, double R22, double R23,
       double R31, double R32, double R33);
+  Rot3(double w, double x, double y, double z);
```
swarrier246, line 600: Is this related to this PR? (I'm assuming it's a quaternion rot constructor?)

varunagrawal:
> > > There is already px and py ?? I do not support to duplicate versions of all these things
> > 
> > 
> > `Cal3Bundler` doesn't have `px` or `py` so this causes undefined function errors when working with different camera models (e.g. using `Cal3_S2` in the main code and then loading from a `Cal3Bundler` from a BAL file. I figured having `u0` and `v0` made more sense, though I can make Cal3Bundler have `px` & `py` instead?
> 
> Wouldn't it be advisable to remove the px and py in the other models and reflect those changes across all places that use it, then? Might cause some avoidable confusion otherwise.

That would break backwards compatibility. Also a lot of internal code uses the `px` and `py` functions.

gtsam/gtsam.i
varunagrawal: It isn't, but it was minor and I didn't want to create a whole new PR for it.

dellaert:
Please request a re-review after adding px and py to missing Cal. Remove all u0 and v0 that were not there yet. Pls deprecate u0 and v0 if there was any.

gtsam/geometry/Cal3Bundler.h
```diff
@@ -98,11 +98,23 @@ class GTSAM_EXPORT Cal3Bundler {
...
+  inline double px() const {
+    return u0_;
+  }
+
+  /// image center in y
+  inline double py() const {
+    return v0_;
+  }
+
+  /// @deprecated
```
dellaert, line 111: No, we have this #ifdef thingy, remember ? :-)

gtsam/gtsam.i
```diff
@@ -2761,7 +2762,16 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
 };
 
 #include <gtsam/slam/dataset.h>
+
 class SfmTrack {
+  SfmTrack();
+
+  double r;
```
dellaert, line 2769: double r,g,b;

gtsam/gtsam.i
```diff
@@ -2761,7 +2762,16 @@ virtual class EssentialMatrixFactor : gtsam::NoiseModelFactor {
...
+  double g;
+  double b;
+  //TODO Need to close wrap#10 to allow this to work.
```
dellaert, line 2772: Please use Google-style linting, which should flag an issue here.

dellaert (approved):
Thanks, pls address comments though.

swarrier246 (commented):
LGTM!

gtsam/geometry/Cal3Bundler.h
varunagrawal: Uhhh I'm stupid

varunagrawal:
Will squash and merge once CI passes.

gtsam/gtsam.i
varunagrawal: The wrap parser doesn't support this syntax. :(

PR #585, opened 2020-11-08
Avoid redundant calls to error() in NonLinearOptimizer

jlblancoc:
A simple fix. `error()` was called two times in each iteration (or 3 times if verbose-level was high enough). 
Surely not a huge efficient loss, haven't measured it, but for large graphs it might be noticeable.


varunagrawal (approved):
LGTM

jlblancoc:
> LGTM

:+1:

PR #593, opened 2020-11-10
Ubuntu+GCC9 CI stage

varunagrawal:
Github Actions has improved its performance and this particular stage no longer times out, so reintroducing it.


ProfFan (approved):
Great!

swarrier246 (approved):

PR #594, opened 2020-11-10
Remove goto

varunagrawal:
Replace the `goto` commands with some simple boolean flag use.


dellaert (changes requested):
Again, I think goto is cleaner and more understandable here. It has a bad reputation from the 80s, but works in these settings.

PR #601, opened 2020-11-15
New print assertions

varunagrawal:
This PR adds new assertion functions to be able to test print outputs in a clean, functional way, reducing copy-pasta.

With these assertions, we can test for consistent printing across platforms, to improve the overall GTSAM experience.

- `assert_stdout_equal` checks for the output when using `std::cout <<`.
- `assert_print_equal` checks the output when using a class' `print` function, e.g. `pose.print()`.

### DIFF
--- a/gtsam/base/TestableAssertions.h
+++ b/gtsam/base/TestableAssertions.h
@@ -23,6 +23,7 @@
 #include <boost/optional.hpp>
 #include <map>
 #include <iostream>
+#include <sstream>
 #include <vector>
 
 namespace gtsam {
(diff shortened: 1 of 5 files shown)


varunagrawal:
I came up with this PR because I found myself reinventing the stringstream-cout-redirect wheel too many times. :slightly_smiling_face:

dellaert (approved):

jlblancoc:
@varunagrawal What about adding all `print()` methods an extra optional parameter of type `std::ostream`, defaulting to `std::cout`, and which could be set to a local `std::sstringstream` in these functions? That idea has been around in my head for a long time, but it would imply touching many (all!) print() methods... 
Thoughts?

varunagrawal:
I have been thinking about the same thing, but I believe the better approach would be to overload the `<<` operator for all classes that support the `print` function. Then the use of a `stringstream` would be trivial.

I did contemplate capturing the output in a stringstream, but there were 2 problems I wasn't sure about:
1. For different systems and compilers, it made sense to perform the redirection so that we were reproducing real printing conditions, rather than a pseudo-print. The last thing I wanted was the headache of users complaining about printing being bad.
2. The `print()` function is what we use in the wrapper to print details of an object, so I wasn't sure how adding an optional parameter would affect the wrapper's operation. @ProfFan is the wrapper expert and I just have working knowledge of it. Given that Fan is pretty busy these days with SwiftFusion, I opted out of that extra pain.

PR #602, opened 2020-11-15
Publish GTSAM artifacts

varunagrawal:
Updated the Actions workflow to upload the corresponding build directories, so we can use the latest version of GTSAM in CI pipelines for other repos (e.g. GTDynamics).

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -83,3 +83,8 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           bash .github/scripts/unix.sh -t
+      - name: Upload build directory
+        uses: actions/upload-artifact@v2
+          with:
+            name: gtsam-${{ matrix.name }}
+            path: $GITHUB_WORKSPACE/build/
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -85,6 +85,6 @@ jobs:
           bash .github/scripts/unix.sh -t
       - name: Upload build directory
         uses: actions/upload-artifact@v2
-          with:
-            name: gtsam-${{ matrix.name }}
-            path: $GITHUB_WORKSPACE/build/
+        with:
+          name: gtsam-${{ matrix.name }}
+          path: $GITHUB_WORKSPACE/build/
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -85,7 +85,7 @@ jobs:
           bash .github/scripts/unix.sh -t
       - name: Upload build directory
         uses: actions/upload-artifact@v2
-        if: ${{ matrix.build_type }} == "Release"
+        if: ${{ matrix.build_type  == "Release" }}
         with:
           name: gtsam-${{ matrix.name }}-${{ matrix.build_type }}
           path: ${{ github.workspace }}/build/
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -86,5 +86,5 @@ jobs:
       - name: Upload build directory
         uses: actions/upload-artifact@v2
         with:
-          name: gtsam-${{ matrix.name }}
+          name: gtsam-${{ matrix.name }}-${{ matrix.build_type }}
           path: ${{ github.workspace }}/build/
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -87,4 +87,4 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: gtsam-${{ matrix.name }}
-          path: $GITHUB_WORKSPACE/build/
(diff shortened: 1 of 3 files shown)


.github/workflows/build-linux.yml
```diff
@@ -82,4 +82,9 @@ jobs:
...
         if: runner.os == 'Linux'
         run: |
-          bash .github/scripts/unix.sh -t
\ No newline at end of file
+          bash .github/scripts/unix.sh -t
+      - name: Upload build directory
+        uses: actions/upload-artifact@v2
+        with:
+          name: gtsam-${{ matrix.name }}-${{ matrix.build_type }}
+          path: ${{ github.workspace }}/build/
```
gchenfc, line 90: newline at end of file ?

.github/workflows/build-macos.yml
```diff
@@ -51,3 +51,8 @@ jobs:
         if: runner.os == 'macOS'
         run: |
           bash .github/scripts/unix.sh -t
+      - name: Upload build directory
+        uses: actions/upload-artifact@v2
+        with:
+          name: gtsam-${{ matrix.name }}-${{ matrix.build_type }}
+          path: ${{ github.workspace }}/build/
```
gchenfc, line 58: newline

.github/workflows/build-windows.yml
```diff
@@ -75,4 +75,9 @@ jobs:
...
           cmake --build build --config ${{ matrix.build_type }} --target check.base
           cmake --build build --config ${{ matrix.build_type }} --target check.base_unstable
-          cmake --build build --config ${{ matrix.build_type }} --target check.linear
\ No newline at end of file
+          cmake --build build --config ${{ matrix.build_type }} --target check.linear
+      - name: Upload build directory
+        uses: actions/upload-artifact@v2
+        with:
+          name: gtsam-${{ matrix.name }}-${{ matrix.build_type }}
+          path: ${{ github.workspace }}/build/
```
gchenfc, line 83: newline

gchenfc (commented):
Cool!

2 comments:
1. is there a reason not to do this for python and special cases?
2. The artifacts are pretty big since they are uploading the entire build directory, especially the debug zip which is ~10GB.  This has the dual byproducts of extra CI time (~15min for debug to upload the file) and a pretty time to download the artifacts down from github.  Neither of these are deal breakers for me, but just thought I'd point out how large the files are.  Perhaps an argument could be made for a more selective approach?

ProfFan:
@gchenfc Python wheels need very special build environments (check gtsam-manylinux-build) to make them run on every possible Linux distro you have. That is also very expensive (currently needs a custom GitHub runner hosted on GCE).

Also I second Gerry's 2nd point here - I think GitHub has a 10G-ish quota on storage space?

.github/workflows/build-linux.yml
varunagrawal: I checked the file again and it already has a newline.

.github/workflows/build-macos.yml
varunagrawal: Already has a newline :slightly_smiling_face:

.github/workflows/build-windows.yml
varunagrawal: Already has a newline :slightly_smiling_face:

varunagrawal:
I am not sure about the file size limit because Actions is not complaining about it, but you are right, if the upload is adding to the overall time, it doesn't make sense to add all the files. I wonder if zipping up the build directory would make a difference?

.github/workflows/build-linux.yml
gchenfc: sorry my bad!

gchenfc:
[github billing](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions)
Not sure what account borglab is on but this would seem to imply that indeed there are storage limits which will incur additional cost.

Hmm.  Maybe not using debug builds and also only uploading builds on merge with develop?
I think the zipped files are ~1/4 the size of the original directory and take virtually no time to zip

varunagrawal:
> GitHub Actions usage is free for public repositories and self-hosted runners.

No worries here.

gchenfc (approved):
Ahh, didn't read carefully that the limits/billing only applied to private repos.
Personally I would prefer only enabling this on release builds just to speed up builds, but up to you!

varunagrawal:
I completely agree with you on the Release part. For downstream use, Debug builds for the CI just seem like an unnecessary time sink.

varunagrawal:
Also, "free for public repos" is one of the reasons why I convinced Frank to move from Bitbucket to Github. We have so much tooling available to make life easy now.

varunagrawal:
Okay wow, release builds are less than 500 MB while debug builds are > 4 GB!!! Great recommendation @gchenfc.

PR #608, opened 2020-11-19
Mark Values::at return type as const

### LINKED ISSUE #606: Error-prone assignable lvalue returned by Values.at<T>()
## Description

Possible bug in `operator =` of `gtsam::Value` 
I'll write a small test code to reproduce later, but just wanted to leave it reported here as a reminder, since this drove me crazy for a long time.

`Values::at<T>(key)` returns a const ref, but at least in the case of "T=gtsam::Vector", it's **assignable** without any warning or error (!).

## Steps to reproduce

Better example will be provided, this is just a snippet:

```cpp
Values v;
v.insert(...)
v.at<gtsam::Vector>(key) = gtsam::Vector();
// Correct usage would be: v.update(...);
```

## Expected behavior

IMO, the "=" above should not compile, since at() returns `const Value&`. At least, it was confusing to me and didn't give any clue about the value not being actually updated, which should be done via `update()`.

## Environment

gtsam develop branch. Ubuntu 20.04, default g++.

varunagrawal:
Fixes #606

### DIFF
--- a/gtsam/nonlinear/Values-inl.h
+++ b/gtsam/nonlinear/Values-inl.h
@@ -338,19 +338,18 @@ namespace gtsam {
    }  // internal
 
    /* ************************************************************************* */
-   template<typename ValueType>
-   ValueType Values::at(Key j) const {
+   template <typename ValueType>
+   const ValueType Values::at(Key j) const {
      // Find the item
      KeyValueMap::const_iterator item = values_.find(j);
 
      // Throw exception if it does not exist
-     if(item == values_.end())
-       throw ValuesKeyDoesNotExist("at", j);
+     if (item == values_.end()) throw ValuesKeyDoesNotExist("at", j);
 
-    // Check the type and throw exception if incorrect
-    // h() split in two lines to avoid internal compiler error (MSVC2017)
-    auto h = internal::handle<ValueType>();
-    return h(j,item->second);
+     // Check the type and throw exception if incorrect
+     // h() split in two lines to avoid internal compiler error (MSVC2017)
+     auto h = internal::handle<ValueType>();
+     return h(j, item->second);
   }
 
   /* ************************************************************************* */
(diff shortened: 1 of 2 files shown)


gtsam/nonlinear/Values-inl.h
```diff
@@ -338,19 +338,18 @@ namespace gtsam {
...
+     if (item == values_.end()) throw ValuesKeyDoesNotExist("at", j);
 
-    // Check the type and throw exception if incorrect
-    // h() split in two lines to avoid internal compiler error (MSVC2017)
-    auto h = internal::handle<ValueType>();
-    return h(j,item->second);
+     // Check the type and throw exception if incorrect
+     // h() split in two lines to avoid internal compiler error (MSVC2017)
+     auto h = internal::handle<ValueType>();
+     return h(j, item->second);
```
jlblancoc, line 352: Indent?

varunagrawal: I didn't notice this before, but the entire file has 3 space indentation. I guess I'll have to update it.

varunagrawal: That's why the formatter decided to add an extra indent.

jlblancoc:
In principle... it LGTM. However, if you don't mind, please wait until I add a minimal unit test failing before, passing after, ok? I would push directly here to your branch, ok?

varunagrawal:
Sounds great!

jlblancoc:
This fixes my original problem, thanks!
However, I can't add tiny unit test to verify that the older code no longer compiles without c++17's `if constexpr()`... it could be added with c++11 and SFINAE, but it's probably not worth the messy unit test.

But before merging, please check out the additional point I'm adding to #606.

jlblancoc:
(just a reminder) this pr is waiting for a design decision by @dellaert in https://github.com/borglab/gtsam/issues/606#issuecomment-730918399

dellaert (changes requested):
Large diff because of formatting changes. Could you undo all formatting changes and just leave the changes for this PR? Also messes up "blame".

varunagrawal:
@dellaert all done.

dellaert (approved):
Nice!

PR #615, opened 2020-11-27
vector of cameras and triangulation function wrapped

swarrier246:
wrapped vector of cameras and triangulation function for individual calibrations. Unit tests added


gtsam/geometry/triangulation.h
```diff
@@ -494,5 +496,8 @@ TriangulationResult triangulateSafe(const CameraSet<CAMERA>& cameras,
     }
 }
 
+typedef CameraSet<PinholeCamera<Cal3Bundler>> CameraSetCal3Bundler;
```
akshay-krishnan, line 499: Could you add a comment to explain these typedefs are for the wrapper?

gtsam/gtsam.i
```diff
@@ -1149,7 +1175,13 @@ gtsam::Point3 triangulatePoint3(const gtsam::Pose3Vector& poses,
...
     gtsam::Cal3Bundler* sharedCal, const gtsam::Point2Vector& measurements,
     double rank_tol, bool optimize);
-
+gtsam::Point3 triangulatePoint3(const gtsam::CameraSetCal3_S2& cameras,
+    const gtsam::Point2Vector& measurements, double rank_tol,
+    bool optimize);
+gtsam::Point3 triangulatePoint3(const gtsam::CameraSetCal3Bundler& cameras,
+    const gtsam::Point2Vector& measurements, double rank_tol,
+    bool optimize);
+
```
akshay-krishnan, line 1184: nit: It looks like you have an empty space on this new line.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,21 +15,24 @@
 import gtsam as g
 from gtsam.utils.test_case import GtsamTestCase
 from gtsam import Cal3_S2, Cal3Bundler, Rot3, Pose3, \
-    PinholeCameraCal3_S2, Point3, Point2Vector, Pose3Vector, triangulatePoint3
+    PinholeCameraCal3_S2, Point3, Point2Vector, Pose3Vector, triangulatePoint3, CameraSetCal3_S2, CameraSetCal3Bundler, PinholeCameraCal3Bundler
```
akshay-krishnan, line 18: Line too lengthy, split into two lines?

python/gtsam/tests/test_Triangulation.py
```diff
@@ -76,5 +79,63 @@ def test_TriangulationExample(self):
...
+        measurements.append(z1)
+        measurements.append(z2)
+        cameras.append(camera1)
+        cameras.append(camera2)
+
+        optimize = True
+        rank_tol = 1e-9
+
+        triangulated_landmark = triangulatePoint3(cameras, measurements, rank_tol, optimize)
+        self.gtsamAssertEquals(landmark, triangulated_landmark, 1e-2)
```
akshay-krishnan, line 108: 1e-2? Does it not work with a smaller error margin?

python/gtsam/tests/test_Triangulation.py
```diff
@@ -76,5 +79,63 @@ def test_TriangulationExample(self):
...
+        z1 = camera1.project(landmark)
+        z2 = camera2.project(landmark)
+        # two cameras
+        measurements = Point2Vector()
+        cameras = CameraSetCal3Bundler()
+
+        measurements.append(z1)
+        measurements.append(z2)
+        cameras.append(camera1)
+        cameras.append(camera2)
```
akshay-krishnan, line 130: It is good practice (it might also be my personal choice) to create a variable where it is used. That would mean 
```
        measurements = Point2Vector()
        measurements.append(z1)
        measurements.append(z2)

        cameras = CameraSetCal3Bundler()
        cameras.append(camera1)
        cameras.append(camera2)
```
Would you say this is more readable? It is up to you ultimately.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -76,5 +79,63 @@ def test_TriangulationExample(self):
...
+
+        optimize = True
+        rank_tol = 1e-9
+
+        triangulated_landmark = triangulatePoint3(cameras, measurements, rank_tol, optimize)
+        self.gtsamAssertEquals(landmark, triangulated_landmark, 1e-2) 
+
```
akshay-krishnan, line 137: Out of curiosity, what does rank_tol mean and why do we have to use a higher error tolerance when the rank_tol is so low?

gtsam/geometry/triangulation.h
```diff
@@ -24,6 +24,8 @@
 #include <gtsam/slam/TriangulationFactor.h>
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
 #include <gtsam/inference/Symbol.h>
+#include <gtsam/geometry/Cal3_S2.h>
```
akshay-krishnan, line 27: Includes have to be in alphabetical order.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,21 +15,24 @@
...
         # Looking along X-axis, 1 meter above ground plane (x-y)
         upright = Rot3.Ypr(-np.pi / 2, 0., -np.pi / 2)
-        pose1 = Pose3(upright, Point3(0, 0, 1))
-        camera1 = PinholeCameraCal3_S2(pose1, sharedCal)
+        self.pose1 = Pose3(upright, Point3(0, 0, 1))
 
         # create second camera 1 meter to the right of first camera
-        pose2 = pose1.compose(Pose3(Rot3(), Point3(1, 0, 0)))
-        camera2 = PinholeCameraCal3_S2(pose2, sharedCal)
+        self.pose2 = self.pose1.compose(Pose3(Rot3(), Point3(1, 0, 0)))
```
akshay-krishnan, line 28: Looks like you dont explicitly need to make them members. i.e maybe you declare pose1 and pose2 in setUp as they were done before and use them in the other tests?

akshay-krishnan (changes requested):

python/gtsam/tests/test_Triangulation.py
akshay-krishnan: My bad, did not read the diff correctly. You may ignore this, marking as resolved :/

python/gtsam/tests/test_Triangulation.py
akshay-krishnan: nit: You could also create the cameras vector immediately after camera1 and camera2 were defined.

varunagrawal:
@swarrier246 can you please run `git merge develop` and push again? That will ensure CI does not fail due to being an older version.

python/gtsam/tests/test_Triangulation.py
swarrier246: it does. changed it.

python/gtsam/tests/test_Triangulation.py
swarrier246: rank_tol is the SVD rank tolerance, used to check if the triangulation is underconstrained. The error tolerance, I believe is the floating point digits comparison when checking for equality between the two Point3s.
I've changed the error tolerance to be lower, and the rank_tol value is the same as used throughout gtsam for triangulateDLT. Haven't changed it

swarrier246:
> @swarrier246 can you please run `git merge develop` and push again? That will ensure CI does not fail due to being an older version.

It says already up to date. Could there be another reason the CI is failing?

varunagrawal:
That means you haven't done a `git pull origin develop` for a while :)

gtsam/gtsam.i
```diff
@@ -1109,6 +1109,32 @@ typedef gtsam::PinholeCamera<gtsam::Cal3_S2> PinholeCameraCal3_S2;
...
+  CameraSetCal3Bundler();
+
+  // common STL methods
+  size_t size() const;
+  bool empty() const;
+  void clear();
+
+  // structure specific methods
+  gtsam::PinholeCameraCal3Bundler at(size_t i) const;
+  void push_back(gtsam::PinholeCameraCal3Bundler& cam) const;
```
akshay-krishnan, line 1122: `const gtsam::PinholeCameraCal3Bundler&`

gtsam/gtsam.i
```diff
@@ -1109,6 +1109,32 @@ typedef gtsam::PinholeCamera<gtsam::Cal3_S2> PinholeCameraCal3_S2;
...
+  CameraSetCal3_S2();
+
+  // common STL methods
+  size_t size() const;
+  bool empty() const;
+  void clear();
+
+  // structure specific methods
+  gtsam::PinholeCameraCal3_S2 at(size_t i) const;
+  void push_back(gtsam::PinholeCameraCal3_S2& cam) const;
```
akshay-krishnan, line 1135: const argument?

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,123 @@
...
+        # z1 = camera1.project(self.landmark)
+        # z2 = camera2.project(self.landmark)
         #
         # measurements = Point2Vector()
         # measurements.append(z1)
         # measurements.append(z2)
         #
         # triangulated_landmark = triangulatePoint3(poses,bundlerCal, measurements, rank_tol, optimize)
-        # self.gtsamAssertEquals(landmark, triangulated_landmark,1e-9)
+        # self.gtsamAssertEquals(self.landmark, triangulated_landmark,1e-9)
```
akshay-krishnan, line 82: I wonder why there is so much commented code here, can this be removed?

akshay-krishnan (changes requested):

python/gtsam/tests/test_Triangulation.py
swarrier246: that is an example of triangulation with bundler calibration. Have left it as it was. @ProfFan , should I put this into a different unit test?

swarrier246: removed the commented code for now, since it is similar to Cal3_S2, and there is a unittest that uses distinct Bundler calibrations. Triangulation with a shared Bundler calibration can be rewitten as a separate unit test, if required

akshay-krishnan (approved):

python/CMakeLists.txt
```diff
@@ -38,6 +38,8 @@ set(ignore
     gtsam::Pose3Vector
     gtsam::KeyVector
     gtsam::BinaryMeasurementsUnit3
+    gtsam::CameraSetCal3Bundler
+    gtsam::CameraSetCal3_S2
```
varunagrawal, line 42: Why are these here? `CameraSet` is a valid type rather than a typedef of an STL container. You should be able to access the class methods without specializing this class.

python/gtsam/preamble.h
```diff
@@ -10,3 +10,5 @@ PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Pose3>);
 PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose3> > >);
 PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose2> > >);
 PYBIND11_MAKE_OPAQUE(std::vector<gtsam::IndexPair>);
+PYBIND11_MAKE_OPAQUE(gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3Bundler> >);
+PYBIND11_MAKE_OPAQUE(gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3_S2> >);
```
varunagrawal, line 14: Maybe not needed because `CameraSet` is a defined class.

python/gtsam/specializations.h
```diff
@@ -13,3 +13,5 @@ py::bind_vector<std::vector<gtsam::BinaryMeasurement<gtsam::Unit3> > >(m_, "Bina
 py::bind_map<gtsam::IndexPairSetMap>(m_, "IndexPairSetMap");
 py::bind_vector<gtsam::IndexPairVector>(m_, "IndexPairVector");
 py::bind_map<gtsam::KeyPairDoubleMap>(m_, "KeyPairDoubleMap");
+py::bind_vector<gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3_S2> > >(m_, "CameraSetCal3_S2");
+py::bind_vector<gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3Bundler> > >(m_, "CameraSetCal3Bundler");
```
varunagrawal, line 17: Again, perhaps not needed since the subclassing mechanism should give you access to the `vector` methods.

varunagrawal (changes requested):
Too much copy-paste going on which can be easily eliminated.

python/CMakeLists.txt
dellaert: @ProfFan should be able to explain. Is it not explained in the README? I asked @ProfFan to ask a README to explain what goes here.

python/gtsam/specializations.h
dellaert: same: ask @ProfFan , this might be above Sushmita's (and my) paygrade. @ProfFan needs to be documented !!!

python/CMakeLists.txt
ProfFan: It is a STL container (and actually templated on the inner type as well), so this is required.

python/gtsam/specializations.h
ProfFan: This is a case that has never been encountered before, I'll document this case shortly after this week (too busy)

varunagrawal:
@swarrier246 @dellaert @ProfFan I did some playing around with the wrapper last night for this particular CameraSet case, and the good news is that we don't need to add those classes to the CMake or specializations files. Pybind11 is able to glean the methods from the defined subclass without issues.

The bad news is that the wrapper binding was failing for me due to the remnants of #237 which is my bad. I didn't open a tracking issue as @dellaert asked me to, but as consolation I have opened issue #621 as well as a PR #623 to remedy this.

python/CMakeLists.txt
ProfFan: Note this is not merely for accessibility:
- You do not ignore the class, then you have STL binding but needs one extra copy (`<stl.h>`)
- You ignore and `MAKE_OPAQUE`, then you have a fast binding without copy (`<stl_bind.h>`)

The automatic STL binding is nice, but I think it's always better to avoid overheads :)

gtsam/gtsam.i
```diff
@@ -1150,7 +1176,13 @@ gtsam::Point3 triangulatePoint3(const gtsam::Pose3Vector& poses,
...
     gtsam::Cal3Bundler* sharedCal, const gtsam::Point2Vector& measurements,
     double rank_tol, bool optimize);
-
+gtsam::Point3 triangulatePoint3(const gtsam::CameraSetCal3_S2& cameras,
+    const gtsam::Point2Vector& measurements, double rank_tol,
+    bool optimize);
+gtsam::Point3 triangulatePoint3(const gtsam::CameraSetCal3Bundler& cameras,
+    const gtsam::Point2Vector& measurements, double rank_tol,
+    bool optimize);
+
```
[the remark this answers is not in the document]
swarrier246: Templates aren't supported for free functions currently. I've opened an issue at borglab/wrap#14 for this. Can we proceed for the moment with repeated functions?

python/CMakeLists.txt
varunagrawal: I see. This raises the question of whether code that accepts a CameraSet would accept a List or would it be expecting an object of type CameraSet. This is the weird thing about this scenario since CameraSet is not a vector alias, rather a vector subclass.

gtsam/gtsam.i
```diff
@@ -1068,6 +1068,23 @@ typedef gtsam::PinholeCamera<gtsam::Cal3DS2> PinholeCameraCal3DS2;
...
+  bool empty() const;
+  void clear();
+
+  // structure specific methods
+  T at(size_t i) const;
+  void push_back(const T& cam);
+};
+
+typedef gtsam::CameraSet<gtsam::PinholeCameraCal3_S2> CameraSetCal3_S2;
+typedef gtsam::CameraSet<gtsam::PinholeCameraCal3Bundler> CameraSetCal3Bundler;
```
varunagrawal, line 1086: Since `CameraSet` requires only one template, you can create a template list on line 1071 rather than defining multiple typedefs.

gtsam/gtsam.i
```diff
@@ -1099,7 +1116,6 @@ class StereoCamera {
 
 #include <gtsam/geometry/triangulation.h>
 
-// Templates appear not yet supported for free functions
```
varunagrawal, line 1102: Please keep this comment and add info about the issue in the wrap repo.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
 import gtsam as g
 from gtsam.utils.test_case import GtsamTestCase
 from gtsam import Cal3_S2, Cal3Bundler, Rot3, Pose3, \
-    PinholeCameraCal3_S2, Point3, Point2Vector, Pose3Vector, triangulatePoint3
+    PinholeCameraCal3_S2, PinholeCameraCal3Bundler, Point3, \
+    Point2Vector, Pose3Vector, triangulatePoint3, \
+    CameraSetCal3_S2, CameraSetCal3Bundler
```
varunagrawal, line 20: sort imports

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
...
+        self.pose2 = self.pose1.compose(Pose3(Rot3(), Point3(1, 0, 0)))
 
         # landmark ~5 meters infront of camera
-        landmark = Point3(5, 0.5, 1.2)
+        self.landmark = Point3(5, 0.5, 1.2)
+
+    
+    def test_TriangulationExample(self):
+        # Some common constants
+        sharedCal = Cal3_S2(1500, 1200, 0, 640, 480)
```
varunagrawal, line 38: Can be made common.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
...
+
+        triangulated_landmark = triangulatePoint3(cameras, measurements, rank_tol, optimize)
+        self.gtsamAssertEquals(self.landmark, triangulated_landmark, 1e-9)
+
+    def test_distinct_Ks_Bundler(self):
+        # two cameras
+        K1 = Cal3Bundler(1500, 0, 0, 640, 480)
+        camera1 = PinholeCameraCal3Bundler(self.pose1, K1)
+
+        K2 = Cal3Bundler(1600, 0, 0, 650, 440)
```
varunagrawal, line 100: K1, K2, camera1, camera2 can be instantiated in the `setUp` method since they are all common.
Similarly you can generate the measurements in the `setUp` since they all use the same cameras and the same landmark, this will require you to have less copy-paste everywhere.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
...
+        # Project two landmarks into two cameras and triangulate
+        z1 = camera1.project(self.landmark)
+        z2 = camera2.project(self.landmark)
+
+        measurements = Point2Vector()
+        measurements.append(z1)
+        measurements.append(z2)
+
+        optimize = True
+        rank_tol = 1e-9
```
varunagrawal, line 116: Pass these as keyword-arguments.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
...
+        # Project two landmarks into two cameras and triangulate
+        z1 = camera1.project(self.landmark)
+        z2 = camera2.project(self.landmark)
+        
+        measurements = Point2Vector()
+        measurements.append(z1)
+        measurements.append(z2)
+
+        optimize = True
+        rank_tol = 1e-9
```
varunagrawal, line 90: Pass these as keyword arguments.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
...
+    Point2Vector, Pose3Vector, triangulatePoint3, \
+    CameraSetCal3_S2, CameraSetCal3Bundler
 
 class TestVisualISAMExample(GtsamTestCase):
-    def test_TriangulationExample(self):
-        # Some common constants
-        sharedCal = Cal3_S2(1500, 1200, 0, 640, 480)
-
+    def setUp(self):
+        # Set up two camera poses
```
varunagrawal, line 24: Please use proper docstring conventions aka triple-quotes.

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
...
-        pose2 = pose1.compose(Pose3(Rot3(), Point3(1, 0, 0)))
-        camera2 = PinholeCameraCal3_S2(pose2, sharedCal)
+        self.pose2 = self.pose1.compose(Pose3(Rot3(), Point3(1, 0, 0)))
 
         # landmark ~5 meters infront of camera
-        landmark = Point3(5, 0.5, 1.2)
+        self.landmark = Point3(5, 0.5, 1.2)
+
+    
+    def test_TriangulationExample(self):
```
varunagrawal, line 36: docstring please

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
...
-        #
-        # measurements = Point2Vector()
-        # measurements.append(z1)
-        # measurements.append(z2)
-        #
-        # triangulated_landmark = triangulatePoint3(poses,bundlerCal, measurements, rank_tol, optimize)
-        # self.gtsamAssertEquals(landmark, triangulated_landmark,1e-9)
+        self.gtsamAssertEquals(self.landmark, triangulated_landmark,1e-2)
+
+    def test_distinct_Ks(self):
```
varunagrawal, line 69: docstring please

python/gtsam/tests/test_Triangulation.py
```diff
@@ -15,66 +15,108 @@
...
+        measurements.append(z1)
+        measurements.append(z2)
+
+        optimize = True
+        rank_tol = 1e-9
+
+        triangulated_landmark = triangulatePoint3(cameras, measurements, rank_tol, optimize)
+        self.gtsamAssertEquals(self.landmark, triangulated_landmark, 1e-9)
+
+    def test_distinct_Ks_Bundler(self):
```
varunagrawal, line 95: docstring please

gtsam/geometry/triangulation.h
```diff
@@ -494,5 +496,9 @@ TriangulationResult triangulateSafe(const CameraSet<CAMERA>& cameras,
     }
 }
 
+// Vector of Cameras - used by the Python/MATLAB wrapper
+using CameraSetCal3Bundler = CameraSet<PinholeCamera<Cal3Bundler>>;
+using CameraSetCal3_S2 = CameraSet<PinholeCamera<Cal3_S2>>;
```
varunagrawal, line 501: I understand why you need this.

@ProfFan  seems like the wrapper does not maintain ordering of declarations, so even though the typedef exists in `gtsam.i`, these aliases are required to avoid `undefined` errors.

varunagrawal (changes requested):
This PR can be made quite a bit leaner. Be sure to be cognizant of repetitive code, especially when you can abstract it away in functions/methods.

varunagrawal:
@swarrier246 please also format all the python code you've added. If you use VS Code, you can install autopep8 really easily and then just hit Ctrl+Shift+I.

gtsam/geometry/triangulation.h
varunagrawal: I spoke to @swarrier246 offline and we figured out how to not use the `using` alias just for the sake of the wrapper.

gtsam/gtsam.i
swarrier246: Leaving the typedef there for shorter type name and to enforce uniformity in naming conventions

gtsam/geometry/triangulation.h
ProfFan: I think that is very strange as typedefs should be generated on the *top*. BTW, I thought we are ignoring the CameraSet for better performance, no?

swarrier246: We're keeping the `using `. How would ignoring CameraSet improve performance? Won't it break triangulatepoint3? That function expects a CameraSet

ProfFan: I think it will not as ignore will only skip wrapping the CameraSet class, the function signature is still there. I suggest you look at the generated C++ code and do a diff to understand what is going on there.

varunagrawal: I looked into the generated cpp and the typedefs are definitely not generated at the top or in sequence. There are cases where a type is used even before the typedef declaration has been made known, resulting in the wrapper throwing `undefined reference` errors.

I think @ProfFan and @swarrier246 are not on the same page. Fan is referring to updating the preamble.h so that CameraSet is treated as a Python list in the wrapper. Honestly, I don't like that design since we need to touch multiple files to incorporate it, plus a python list accepts all kinds of objects while the CameraSet should only accept specific camera models. Converting to a list kills the type information and can result in potential bugs and errors.

swarrier246: I kinda agree with @varunagrawal 's reasoning on this one. @ProfFan is it a significant performance issue?

ProfFan: I think Varun you have a misunderstanding here. *Binding as list* does not make it a Python `list`. You still need to manually create a `CameraSet` object, just that object has a `list` like interface and can be used as if it is a list. You can try putting in something that is not a `Camera`, and it will fail.

varunagrawal: Ah I see. Then yeah if we get the performance boost, it makes sense to do that. Then we don't need to add all the STL vector methods in the wrapper like it is being currently done, just the class definition should be enough.

@swarrier246 really sorry about this, but I agree with Fan on reverting back to the preamble.h and CMake updates.

gtsam/gtsam.i
swarrier246: rendered moot by adding preamble and CMake updates

gtsam/geometry/triangulation.h
```diff
@@ -18,56 +18,61 @@
...
 #include <gtsam/geometry/CameraSet.h>
+#include <gtsam/geometry/PinholeCamera.h>
 #include <gtsam/geometry/Pose2.h>
 #include <gtsam/slam/TriangulationFactor.h>
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
 #include <gtsam/inference/Symbol.h>
 
-namespace gtsam {
+namespace gtsam
+{
```
varunagrawal, line 31: The formatting of this file is off. Why is that?

gtsam/geometry/triangulation.h
```diff
@@ -18,56 +18,61 @@
...
-/// Exception thrown by triangulateDLT when landmark is behind one or more of the cameras
-class GTSAM_EXPORT TriangulationCheiralityException: public std::runtime_error {
-public:
-  TriangulationCheiralityException() :
-      std::runtime_error(
-          "Triangulation Cheirality Exception: The resulting landmark is behind one or more cameras.") {
-  }
-};
+  /// Exception thrown by triangulateDLT when SVD returns rank < 3
+  class GTSAM_EXPORT TriangulationUnderconstrainedException : public std::runtime_error
```
varunagrawal, line 34: Even the class should start at the 1st column, not have indentations.

gtsam/gtsam.i
```diff
@@ -118,708 +118,735 @@
  *  - TODO: Add generalized serialization support via boost.serialization with hooks to matlab save/load
  */
 
-namespace gtsam {
+namespace gtsam
+{
```
varunagrawal, line 122: Formatting of this file is off as well. Did you change your formatter settings?

python/gtsam/preamble.h
```diff
@@ -5,8 +5,10 @@ PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Key, tbb::tbb_allocator<gtsam::Key>>);
...
-PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Point2, Eigen::aligned_allocator<gtsam::Point2> >);
+PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Point2, Eigen::aligned_allocator<gtsam::Point2>>);
 PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Pose3>);
-PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose3> > >);
-PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose2> > >);
+PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose3>>>);
+PYBIND11_MAKE_OPAQUE(std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose2>>>);
 PYBIND11_MAKE_OPAQUE(std::vector<gtsam::IndexPair>);
+PYBIND11_MAKE_OPAQUE(gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3Bundler>>);
+PYBIND11_MAKE_OPAQUE(gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3_S2>>);
```
varunagrawal, line 14: Newline at the end.

varunagrawal (changes requested):
Almost done. Only issue is formatting.

johnwlambert:
Hi Sushmita, thanks for the PR. I see +3403 lines, -3152 lines in the diff -- I think a lot of things got autoformatted perhaps? can we revert those?

varunagrawal (approved):
LGTM

ProfFan (approved):
Great! Thanks Sushmita and Varun!

PR #616, opened 2020-11-28
Miscellaneous fixes

varunagrawal:
1. Suppress warnings for clang as well as gcc.
2. Fix warning for potentially unassigned pointer.
3. Plotting code fixes.

### DIFF
--- a/gtsam/CMakeLists.txt
+++ b/gtsam/CMakeLists.txt
@@ -199,7 +199,7 @@ if(WIN32)
 else()
   if("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
     # Suppress all warnings from 3rd party sources.
-    set_source_files_properties(${3rdparty_srcs} PROPERTIES COMPILE_FLAGS "-w")
+    set_source_files_properties(${3rdparty_srcs} PROPERTIES COMPILE_FLAGS "-w -Wno-everything")
   else()
     set_source_files_properties(${3rdparty_srcs} PROPERTIES COMPILE_FLAGS "-Wno-error")
   endif()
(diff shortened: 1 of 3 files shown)


gtsam/CMakeLists.txt
```diff
@@ -199,7 +199,7 @@ if(WIN32)
 else()
   if("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
     # Suppress all warnings from 3rd party sources.
-    set_source_files_properties(${3rdparty_srcs} PROPERTIES COMPILE_FLAGS "-w")
+    set_source_files_properties(${3rdparty_srcs} PROPERTIES COMPILE_FLAGS "-w -Wno-everything")
```
dellaert, line 202: what's this?

python/gtsam/utils/plot.py
```diff
@@ -36,18 +36,15 @@ def set_axes_equal(fignum):
     ax.set_zlim3d([origin[2] - radius, origin[2] + radius])
 
 
-def ellipsoid(xc, yc, zc, rx, ry, rz, n):
+def ellipsoid(rx, ry, rz, n):
```
dellaert, line 39: API change? Why?

gtsam/CMakeLists.txt
varunagrawal: That's the flag to suppress warnings in Clang. `-w` already does so for GCC, but Clang still throws warnings in Release mode.

python/gtsam/utils/plot.py
varunagrawal: The center point values are not used in the function anywhere. They are simply wasteful args that serve no purpose.

dellaert (approved):

PR #617, opened 2020-11-28
Gnc

lucacarlone:
Added Graduated Non-Convexity (GNC) as a nonlinear optimizer in GTSAM. This is an implementation of the GNC paper (https://arxiv.org/pdf/1909.08605.pdf), but it uses iterative solvers in the inner loop instead of non-minimal global solvers.

Outstanding tasks:
- [x] reintroduce “using BaseOptimizer = BaseOptimizerParameters::OptimizerType;”  to let the user choose the iterative solver for the inner loop.
- [x] move GncOptimizer and its params to a proper class/file when ready.
- [x] populate the TLS option (now the code only implements GM = Geman McClure).
- [x] double-check that the 0.5 in front of the factors in gtsam does not require a modification to the weight update rule.

### DIFF
--- a/gtsam/nonlinear/GaussNewtonOptimizer.h
+++ b/gtsam/nonlinear/GaussNewtonOptimizer.h
@@ -29,7 +29,7 @@ class GaussNewtonOptimizer;
  */
 class GTSAM_EXPORT GaussNewtonParams : public NonlinearOptimizerParams {
 public:
-  typedef GaussNewtonOptimizer OptimizerType;
+  using OptimizerType = GaussNewtonOptimizer;
 };
 
 /**
--- a/gtsam/nonlinear/GaussNewtonOptimizer.h
+++ b/gtsam/nonlinear/GaussNewtonOptimizer.h
@@ -28,6 +28,8 @@ class GaussNewtonOptimizer;
  * NonlinearOptimizationParams.
  */
 class GTSAM_EXPORT GaussNewtonParams : public NonlinearOptimizerParams {
+public:
+  typedef GaussNewtonOptimizer OptimizerType;
 };
 
 /**
--- a/gtsam/nonlinear/GncOptimizer.h
+++ b/gtsam/nonlinear/GncOptimizer.h
@@ -203,9 +203,11 @@ public:
     double mu = initializeMu();
     double mu_prev = mu;
 
-    // handle the degenerate case for TLS cost that corresponds to small
-    // maximum residual error at initialization
-    if (mu <= 0 && params_.lossType == GncParameters::TLS) {
+    // handle the degenerate case that corresponds to small
+    // maximum residual errors at initialization
+    // For GM: if residual error is small, mu -> 0
+    // For TLS: if residual error is small, mu -> -1
+    if (mu <= 0) {
       if (params_.verbosityGNC >= GncParameters::VerbosityGNC::SUMMARY) {
         std::cout << "GNC Optimizer stopped because maximum residual at "
                      "initialization is small." << std::endl;
@@ -230,6 +232,10 @@ public:
       GaussNewtonOptimizer baseOptimizer_iter(graph_iter, state_);
       result = baseOptimizer_iter.optimize();
 
+      // update mu
+      mu_prev = mu;
+      mu = updateMu(mu);
+
       // stopping condition
       if (checkMuConvergence(mu, mu_prev)) {
         // display info
@@ -240,9 +246,6 @@ public:
         }
         break;
       }
-      // otherwise update mu
-      mu_prev = mu;
-      mu = updateMu(mu);
     }
     return result;
   }
--- a/gtsam/nonlinear/GncOptimizer.h
+++ b/gtsam/nonlinear/GncOptimizer.h
@@ -36,6 +36,9 @@ namespace gtsam {
 template<class BaseOptimizerParameters>
 class GncParams {
 public:
+  //typedef BaseOptimizerParameters::OptimizerType GncOptimizerType;
+  typedef BaseOptimizerParameters::OptimizerType OptimizerType;
+
   /** Verbosity levels */
   enum VerbosityGNC {
     SILENT = 0, SUMMARY, VALUES
@@ -45,8 +48,6 @@ public:
   enum RobustLossType {
     GM /*Geman McClure*/, TLS /*Truncated least squares*/
   };
-
-  using BaseOptimizer = GaussNewtonOptimizer; // BaseOptimizerParameters::OptimizerType;
 
   /// Constructor
   GncParams(const BaseOptimizerParameters& baseOptimizerParams) :
@@ -145,6 +146,11 @@ template<class GncParameters>
 class GncOptimizer {
 public:
   // types etc
+//  typedef BaseOptimizerParameters::OptimizerType GncOptimizerType;
+ // typedef GncParameters::BaseOptimizerParameters::OptimizerType BaseOptimizer; //
+  //typedef BaseOptimizerParameters::OptimizerType BaseOptimizer;
+  //typedef GaussNewtonOptimizer BaseOptimizer;
+  typedef GncParameters::OptimizerType BaseOptimizer;
 
 private:
   NonlinearFactorGraph nfg_;
@@ -198,7 +204,7 @@ public:
   Values optimize() {
     // start by assuming all measurements are inliers
     weights_ = Vector::Ones(nfg_.size());
-    GaussNewtonOptimizer baseOptimizer(nfg_, state_);
+    BaseOptimizer baseOptimizer(nfg_, state_);
     Values result = baseOptimizer.optimize();
     double mu = initializeMu();
     double mu_prev = mu;
@@ -229,7 +235,7 @@ public:
 
       // variable/values update
       NonlinearFactorGraph graph_iter = this->makeWeightedGraph(weights_);
-      GaussNewtonOptimizer baseOptimizer_iter(graph_iter, state_);
+      BaseOptimizer baseOptimizer_iter(graph_iter, state_);
       result = baseOptimizer_iter.optimize();
 
       // update mu
--- a/gtsam/nonlinear/GncOptimizer.h
+++ b/gtsam/nonlinear/GncOptimizer.h
@@ -176,10 +176,10 @@ class GncOptimizer {
     }
     // set initial mu
     switch (params_.lossType) {
-      case GncParameters::GM:
+      case GncLossType::GM:
         // surrogate cost is convex for large mu
         return 2 * rmax_sq / params_.barcSq;  // initial mu
-      case GncParameters::TLS:
+      case GncLossType::TLS:
         /* initialize mu to the value specified in Remark 5 in GNC paper.
          surrogate cost is convex for mu close to zero
          degenerate case: 2 * rmax_sq - params_.barcSq < 0 (handled in the main loop)
@@ -198,10 +198,10 @@ class GncOptimizer {
   /// Update the gnc parameter mu to gradually increase nonconvexity.
   double updateMu(const double mu) const {
     switch (params_.lossType) {
-      case GncParameters::GM:
+      case GncLossType::GM:
         // reduce mu, but saturate at 1 (original cost is recovered for mu -> 1)
         return std::max(1.0, mu / params_.muStep);
-      case GncParameters::TLS:
+      case GncLossType::TLS:
         // increases mu at each iteration (original cost is recovered for mu -> inf)
         return mu * params_.muStep;
       default:
@@ -214,10 +214,10 @@ class GncOptimizer {
   bool checkMuConvergence(const double mu) const {
     bool muConverged = false;
     switch (params_.lossType) {
-      case GncParameters::GM:
+      case GncLossType::GM:
         muConverged = std::fabs(mu - 1.0) < 1e-9;  // mu=1 recovers the original GM function
         break;
-      case GncParameters::TLS:
+      case GncLossType::TLS:
         muConverged = false;  // for TLS there is no stopping condition on mu (it must tend to infinity)
         break;
       default:
@@ -242,10 +242,10 @@ class GncOptimizer {
   bool checkWeightsConvergence(const Vector& weights) const {
     bool weightsConverged = false;
     switch (params_.lossType) {
-      case GncParameters::GM:
+      case GncLossType::GM:
         weightsConverged = false;  // for GM, there is no clear binary convergence for the weights
         break;
-      case GncParameters::TLS:
+      case GncLossType::TLS:
         weightsConverged = true;
         for (size_t i = 0; i < weights.size(); i++) {
           if (std::fabs(weights[i] - std::round(weights[i]))
@@ -315,7 +315,7 @@ class GncOptimizer {
 
     // update weights of known inlier/outlier measurements
     switch (params_.lossType) {
-      case GncParameters::GM: {  // use eq (12) in GNC paper
+      case GncLossType::GM: {  // use eq (12) in GNC paper
         for (size_t k : unknownWeights) {
           if (nfg_[k]) {
             double u2_k = nfg_[k]->error(currentEstimate);  // squared (and whitened) residual
@@ -325,7 +325,7 @@ class GncOptimizer {
         }
         return weights;
       }
-      case GncParameters::TLS: {  // use eq (14) in GNC paper
+      case GncLossType::TLS: {  // use eq (14) in GNC paper
         double upperbound = (mu + 1) / mu * params_.barcSq;
         double lowerbound = mu / (mu + 1) * params_.barcSq;
         for (size_t k : unknownWeights) {
--- a/gtsam/nonlinear/GncOptimizer.h
+++ b/gtsam/nonlinear/GncOptimizer.h
@@ -10,8 +10,8 @@
  * -------------------------------------------------------------------------- */
 
 /**
- * @file    testGncOptimizer.cpp
- * @brief   Unit tests for GncOptimizer class
+ * @file    GncOptimizer.cpp
+ * @brief   The GncOptimizer class
  * @author  Jingnan Shi
  * @author  Luca Carlone
  * @author  Frank Dellaert
--- a/gtsam/nonlinear/GncOptimizer.h
+++ b/gtsam/nonlinear/GncOptimizer.h
@@ -340,7 +340,7 @@ public:
 
     // update weights of known inlier/outlier measurements
     switch (params_.lossType) {
-    case GncParameters::GM: // use eq (12) in GNC paper
+    case GncParameters::GM: { // use eq (12) in GNC paper
       for (size_t k : unknownWeights) {
         if (nfg_[k]) {
           double u2_k = nfg_[k]->error(currentEstimate); // squared (and whitened) residual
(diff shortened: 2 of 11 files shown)


tests/testGncOptimizer.cpp
```diff
@@ -0,0 +1,625 @@
...
+  /// get expected values by optimizing outlier-free graph
+  Values expectedWithOutliers = LevenbergMarquardtOptimizer(*graph, *initial).optimize();
+  // as expected, the following test fails due to the presence of an outlier!
+  // CHECK(assert_equal(expected, expectedWithOutliers, 1e-3));
+
+  // GNC
+  // Note: in difficult instances, we set the odometry measurements to be inliers,
+  // but this problem is simple enought to succeed even without that assumption
+  // std::vector<size_t> knownInliers;
+  GncParams<GaussNewtonParams> gncParams = GncParams<GaussNewtonParams>();
```
dellaert, line 612: Just say `GncParams<GaussNewtonParams> gncParams;` ?

tests/testGncOptimizer.cpp
```diff
@@ -0,0 +1,625 @@
...
+#include <CppUnitLite/TestHarness.h>
+
+using namespace std;
+using namespace gtsam;
+
+using symbol_shorthand::X;
+using symbol_shorthand::L;
+static double tol = 1e-7;
+
+/* ************************************************************************* */
```
dellaert, line 39: Probably need to move to a .h file now

tests/testGncOptimizer.cpp
```diff
@@ -0,0 +1,625 @@
...
+  BaseOptimizerParameters baseOptimizerParams;
+  /// any other specific GNC parameters:
+  RobustLossType lossType = GM; /* default loss*/
+  size_t maxIterations = 100; /* maximum number of iterations*/
+  double barcSq = 1.0; /* a factor is considered an inlier if factor.error() < barcSq. Note that factor.error() whitens by the covariance*/
+  double muStep = 1.4; /* multiplicative factor to reduce/increase the mu in gnc */
+  VerbosityGNC verbosityGNC = SILENT; /* verbosity level */
+  std::vector<size_t> knownInliers = std::vector<size_t>(); /* slots in the factor graph corresponding to measurements that we know are inliers */
+
+  void setLossType(const RobustLossType type) {
```
dellaert, line 73: add comments to all methods, at least `/// one-liner doxygen`

tests/testGncOptimizer.cpp
```diff
@@ -0,0 +1,625 @@
...
+  enum VerbosityGNC {
+    SILENT = 0, SUMMARY, VALUES
+  };
+
+  /** Choice of robust loss function for GNC */
+  enum RobustLossType {
+    GM /*Geman McClure*/, TLS /*Truncated least squares*/
+  };
+
+  using BaseOptimizer = GaussNewtonOptimizer; // BaseOptimizerParameters::OptimizerType;
```
dellaert, line 53: Oh ?

dellaert (commented):
I left some comments. What is the main templating issue you wanted me to look at ?

tests/testGncOptimizer.cpp
lucacarlone: solved

tests/testGncOptimizer.cpp
lucacarlone: yes, this exactly the problem we need to solve. I can do:
using BaseOptimizer = BaseOptimizerParameters::OptimizerType;
Then I can go to LevenbergMarquardtParams and define OptimizerType as LevenbergMarquardtOptimizer.
However, LevenbergMarquardtParams does not know about LevenbergMarquardtOptimizer, so I'm not sure how to make that define work. did you have something different in mind?

dellaert: I think you need to modify LevenbergMarquardtParams to add the typedef there

lucacarlone: ok I'll try that, but I suspect an easier way would be to template on the Optimizer instead of the Params.

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,381 @@
...
+    }
+    // set initial mu
+    switch (params_.lossType) {
+    case GncParameters::GM:
+      return 2 * rmax_sq / params_.barcSq; // initial mu
+    case GncParameters::TLS:
+      // initialize mu to the value specified in Remark 5 in GNC paper
+      // degenerate case: residual is close to zero so mu approximately equals
+      // to -1
+      return params_.barcSq / (2 * rmax_sq - params_.barcSq);
```
lucacarlone, line 270: @jingnanshi : I don't think mu can be negative. does it make sense to have a lower bound for mu in this and the GM case? it the max residual is below barcsq, it means there is no outlier and maybe we can just "return"

jingnanshi: @lucacarlone I completely agree. I put the check outside the initializeMu function (see line 210 to 217) so that we can easily just return the current result.

lucacarlone: great, @jingnanshi !

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,381 @@
...
+    }
+  }
+
+  /// check if we have reached the value of mu for which the surrogate loss matches the original loss
+  bool checkMuConvergence(const double mu, const double mu_prev) const {
+    switch (params_.lossType) {
+    case GncParameters::GM:
+      return std::fabs(mu - 1.0) < 1e-9; // mu=1 recovers the original GM function
+    case GncParameters::TLS:
+      return std::fabs(mu - mu_prev) < params_.relativeMuTol;
```
lucacarlone, line 297: @jingnanshi : this does not make much sense, since the decrease in mu is constant. maybe @pantonante has a better check?

lucacarlone:
@jingnanshi @dellaert @pantonante : I fixed the outstanding issue with optimizerType. I also added a couple of unit tests (and comments) to make sure the inlier threshold is set correctly. @jingnanshi : the only thing I'm not convinced about is the stopping condition for TLS - please check my comment. Besides that, this PR is ready for review!

gtsam/nonlinear/GncOptimizer.h
jingnanshi: @lucacarlone : good catch! this should be cost - cost_prev

lucacarlone: I think Pasquale checks if the weights are (close to) binary, which is the case after TLS converges.

jingnanshi: @lucacarlone : I just checked Pasquale's email, and he suggested using the relative change of the weighted sum of the squared residuals, which is what I used in TEASER++. I think I'll implement a checkConvergence function that checks mu or relative cost change depending on loss type.

pantonante: Yes, I suggest to use the relative change of the weighted sum of the squared residuals because it works well in practice. Here seems you are checking that they are not changing much, this doesn't work for TLS.

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,388 @@
...
+
+  /// Default constructor
+  GncParams() :
+      baseOptimizerParams() {
+  }
+
+  /// GNC parameters
+  BaseOptimizerParameters baseOptimizerParams; /*optimization parameters used to solve the weighted least squares problem at each GNC iteration*/
+  /// any other specific GNC parameters:
+  RobustLossType lossType = GM; /* default loss*/
```
pantonante, line 65: Probably we want TLS as default loss type.

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,425 @@
...
+    switch (params_.lossType) {
+    case GncParameters::GM:
+      return checkMuConvergence(mu);
+    case GncParameters::TLS:
+      return checkCostConvergence(cost, prev_cost);
+    default:
+      throw std::runtime_error(
+          "GncOptimizer::checkMuConvergence: called with unknown loss type.");
+    }
+  }
```
pantonante, line 345: This will work well, but it looks weird that we use two very different stopping criteria for the two robust costs. I would use the cost based convergence test in both cases as default and add a parameter to support three convergence tests:
1. cost based only (default)
2. weighs based only
3. [optional] cost based OR weights base (whatever converges first)
I think this would add clarity overall.

lucacarlone: for simplicity, I would add both stopping conditions (for both costs) and give the user control of the tolerances for stopping, such that s/he can enable/disable the conditions s/he prefers.

lucacarlone: @jingnanshi : let me know if you can take a look at this, otherwise I can do it later in December.

jingnanshi: @lucacarlone @pantonante For the TLS case, since mu is increasing, it is difficult to check for mu (unless I have misunderstood it). What do you guys recommend?

lucacarlone: @jingnanshi : Pasquale was suggesting two checks I think:
1) relative change in cost: e.g., | cost_t - cost_t-1 | /  |cost_t | < params_.relativeCostTol (then stop)
2) weights: e.g., if | weight - round(weights) | < params_.weightConvergenceTol  (then stop)
In the latter round(weights) rounds the entries of weights in {0;1} and this is mostly effective for TLS (it checks whether you have a clear decision inlier/outliers).

Note: you can also keep the stopping condition based on mu for GM, while there is no such condition for TLS

gtsam/nonlinear/GncOptimizer.h
lucacarlone: @jingnanshi : I would implement this in the main loop (before "if (checkMuConvergence(mu, mu_prev))"): I would simply add "if (checkCostConvergence(..)  OR checkWeightsConvergence(..))" and implement these in a separate function. Let me know if you want me to take a stab - I can try to do that tonight

jingnanshi: @lucacarlone : That makes sense. I won't be able to do it tonight unfortunately so feel free to take a look if you have time!

lucacarlone: @jingnanshi @pantonante : I fixed the stopping conditions and a few other things (including setting TLS to default) - please review carefully and let me know if you have more comments.

jingnanshi: @lucacarlone : Had two very minor fixes. Everything looks good!

lucacarlone: perfect - thanks @jingnanshi !

gtsam/nonlinear/GaussNewtonOptimizer.h
```diff
@@ -28,6 +28,8 @@ class GaussNewtonOptimizer;
  * NonlinearOptimizationParams.
  */
 class GTSAM_EXPORT GaussNewtonParams : public NonlinearOptimizerParams {
+public:
+  typedef GaussNewtonOptimizer OptimizerType;
```
dellaert, line 32: Change to `using OptimizerType = GaussNewtonOptimizer`

gtsam/nonlinear/LevenbergMarquardtParams.h
```diff
@@ -40,6 +42,7 @@ class GTSAM_EXPORT LevenbergMarquardtParams: public NonlinearOptimizerParams {
 
   static VerbosityLM verbosityLMTranslator(const std::string &s);
   static std::string verbosityLMTranslator(VerbosityLM value);
+  typedef LevenbergMarquardtOptimizer OptimizerType;
```
dellaert, line 45: Change to `using OptimizerType = LevenbergMarquardtOptimizer `

tests/testGncOptimizer.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+using symbol_shorthand::L;
+using symbol_shorthand::X;
+static double tol = 1e-7;
+
+/* ************************************************************************* */
+TEST(GncOptimizer, gncParamsConstructor) {
+  // check params are correctly parsed
+  LevenbergMarquardtParams lmParams;
+  GncParams<LevenbergMarquardtParams> gncParams1(lmParams);
```
dellaert, line 47: Do we need the <> here (and below) or can the type be derived from the argument? I think it can and should be omitted.

tests/testGncOptimizer.cpp
```diff
@@ -0,0 +1,670 @@
...
+  // has to have Gaussian noise models !
+  auto fg = example::createReallyNonlinearFactorGraph(); // just a unary factor
+                                                         // on a 2D point
+
+  Point2 p0(3, 3);
+  Values initial;
+  initial.insert(X(1), p0);
+
+  LevenbergMarquardtParams lmParams;
+  GncParams<LevenbergMarquardtParams> gncParams(lmParams);
```
dellaert, line 84: If default. I think we should test here the use of 
```
GncParams<LevenbergMarquardtParams> gncParams;
```
i.e. without arguments

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+#pragma once
+
+#include <gtsam/nonlinear/LevenbergMarquardtOptimizer.h>
+#include <gtsam/nonlinear/GaussNewtonOptimizer.h>
+#include <gtsam/nonlinear/NonlinearFactorGraph.h>
+
+namespace gtsam {
+
+/* ************************************************************************* */
+template<class BaseOptimizerParameters>
```
dellaert, line 36: There is enough code here to warrant a separate file for GncParams

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+class GncParams {
+public:
+  /** For each parameter, specify the corresponding optimizer: e.g., GaussNewtonParams -> GaussNewtonOptimizer */
+  typedef typename BaseOptimizerParameters::OptimizerType OptimizerType;
+
+  /** Verbosity levels */
+  enum VerbosityGNC {
```
dellaert, line 43: GNC here is redundant

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+    SILENT = 0, SUMMARY, VALUES
+  };
+
+  /** Choice of robust loss function for GNC */
+  enum RobustLossType {
```
dellaert, line 48: I propose to make this its own class called `GncLossType`, not an inner class. Will make things less verbose.

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+      baseOptimizerParams(baseOptimizerParams) {
+  }
+
+  /// Default constructor
+  GncParams() :
+      baseOptimizerParams() {
+  }
+
+  /// GNC parameters
+  BaseOptimizerParameters baseOptimizerParams; /*optimization parameters used to solve the weighted least squares problem at each GNC iteration*/
```
dellaert, line 63: Single line comments after a line should use doxygen `///< Optimization parameters...
Also, it seems this line is too long so probably you did not format with Google style? Please do so, and if comment is not single line, add it above with doxygen /** ... */

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+public:
+  /** For each parameter, specify the corresponding optimizer: e.g., GaussNewtonParams -> GaussNewtonOptimizer */
+  typedef typename BaseOptimizerParameters::OptimizerType OptimizerType;
+
+  /** Verbosity levels */
+  enum VerbosityGNC {
+    SILENT = 0, SUMMARY, VALUES
+  };
+
+  /** Choice of robust loss function for GNC */
```
dellaert, line 47: Single line doxygen should use `///`

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+
+  /// Default constructor
+  GncParams() :
+      baseOptimizerParams() {
+  }
+
+  /// GNC parameters
+  BaseOptimizerParameters baseOptimizerParams; /*optimization parameters used to solve the weighted least squares problem at each GNC iteration*/
+  /// any other specific GNC parameters:
+  RobustLossType lossType = TLS; /* default loss*/
```
dellaert, line 65: `///< Default loss.`

Note capitalization and punctuation :-)

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+
+/* ************************************************************************* */
+template<class GncParameters>
+class GncOptimizer {
+public:
+  /** For each parameter, specify the corresponding optimizer: e.g., GaussNewtonParams -> GaussNewtonOptimizer */
+  typedef typename GncParameters::OptimizerType BaseOptimizer;
+
+private:
+  NonlinearFactorGraph nfg_;
```
dellaert, line 160: Document

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+        }
+      }
+    }
+  }
+
+  /// Access a copy of the internal factor graph
+  NonlinearFactorGraph getFactors() const {
+    return NonlinearFactorGraph(nfg_);
+  }
+  /// Access a copy of the internal values
```
dellaert, line 195: Here and below: add return. Make sure to format with clang-format or cpp-format w Google style.

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -0,0 +1,459 @@
...
+      // according to remark mu = params_.barcSq / (2 * rmax_sq - params_.barcSq) = params_.barcSq/ excessResidual
+      // however, if the denominator is 0 or negative, we return mu = -1 which leads to termination of the main GNC loop
+      return (2 * rmax_sq - params_.barcSq) > 0 ? params_.barcSq / (2 * rmax_sq - params_.barcSq)  :  -1;
+    default:
+      throw std::runtime_error(
+          "GncOptimizer::initializeMu: called with unknown loss type.");
+    }
+  }
+
+  /// update the gnc parameter mu to gradually increase nonconvexity
```
dellaert, line 305: Capitalize and add period to all comments.

dellaert (commented):
Awesome! First round, will look again after comments resolved :-)

tests/testGncOptimizer.cpp
lucacarlone: @dellaert : it does not work if I omit - should I do something else to make it derive the type?

gtsam/nonlinear/GncOptimizer.h
lucacarlone: @dellaert : what do you envision we should put in the new class? this is currently just an enum

tests/testGncOptimizer.cpp
dellaert: Hmmm, I had expected that to work, but maybe it's only since C++17: https://en.cppreference.com/w/cpp/language/class_template_argument_deduction

gtsam/nonlinear/GncOptimizer.h
dellaert: Just before the params class: otherwise you have to append the entire (templated) parameter class name which is unwieldy.

gtsam/nonlinear/GncParams.h
```diff
@@ -0,0 +1,155 @@
...
+  double relativeCostTol = 1e-5;  ///< If relative cost change is below this threshold, stop iterating
+  double weightsTol = 1e-4;  ///< If the weights are within weightsTol from being binary, stop iterating (only for TLS)
+  Verbosity verbosity = SILENT;  ///< Verbosity level
+  std::vector<size_t> knownInliers = std::vector<size_t>();  ///< Slots in the factor graph corresponding to measurements that we know are inliers
+
+  /// Set the robust loss function to be used in GNC (chosen among the ones in GncLossType).
+  void setLossType(const GncLossType type) {
+    lossType = type;
+  }
+  /// Set the maximum number of iterations in GNC (changing the max nr of iters might lead to less accurate solutions and is not recommended).
```
dellaert, line 80: Add newline between methods...

dellaert (approved):
Very cool! 
Just a few more code nits, feel free to merge after addressed ;-)

Now we need to apply it to Shonan and compare with robuts rotation averaging, right @jingwuOUO  :-) ?

lucacarlone:
all set! I'll merge after @pantonante and @jingnanshi  approve!

jingnanshi (approved):
:+1:

pantonante (approved):

PR #623, opened 2020-11-30
Jacobians for Camera models

### LINKED ISSUE #621: Mismatch between calibrate functions for Cal3_S2 and Cal3DS2
## Description

This is the same issue as #237 but for `Cal3DS2`. We need to use the Implicit Function theorem to fix this in `Cal3DS2_Base.cpp`.

varunagrawal:
- Add jacobians for calibrate function of Cal3DS2 based models using implicit function theorem.
- Consistent naming of jacobian parameters.
- Added tests for jacobians.
- Some simple formatting.
- Fixed docs for implicit function theorem.
- Added parentheses to conform with Google style.

Fixes #621

### DIFF
--- a/gtsam/geometry/Cal3Bundler.h
+++ b/gtsam/geometry/Cal3Bundler.h
@@ -18,6 +18,7 @@
 
 #pragma once
 
+#include <gtsam/geometry/Cal3DS2_Base.h>
 #include <gtsam/geometry/Point2.h>
 
 namespace gtsam {
--- a/gtsam/geometry/Cal3Bundler.h
+++ b/gtsam/geometry/Cal3Bundler.h
@@ -18,7 +18,7 @@
 
 #pragma once
 
-#include <gtsam/geometry/Cal3DS2_Base.h>
+#include <gtsam/geometry/Cal3.h>
 #include <gtsam/geometry/Point2.h>
 
 namespace gtsam {
--- a/gtsam/geometry/Cal3DS2_Base.h
+++ b/gtsam/geometry/Cal3DS2_Base.h
@@ -19,42 +19,10 @@
 
 #pragma once
 
+#include <gtsam/geometry/Cal3.h>
 #include <gtsam/geometry/Point2.h>
 
 namespace gtsam {
-
-/**
- * Function which makes use of the Implicit Function Theorem to compute the
- * Jacobians of `calibrate` using `uncalibrate`.
- * Given f(pi, pn) = uncalibrate(pn) - pi, and g(pi) = calibrate, we can
- * easily compute the Jacobians:
- * df/pi = -I (pn and pi are independent args)
- * Dp = -inv(H_uncal_pn) * df/pi = -inv(H_uncal_pn) * (-I) = inv(H_uncal_pn)
- * Dcal = -inv(H_uncal_pn) * df/K = -inv(H_uncal_pn) * H_uncal_K
- *
- * @tparam T Calibration model.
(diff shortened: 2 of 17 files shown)


dellaert (commented):
The implicit theorem code seems very similar in all cases, on the verge of copy/pasta. I don't think we have a base class for all calibrations, but perhaps we can isolate these (3 now?) implicit theorem calls into one function accessible to all?

varunagrawal:
I have 2 questions about that:

1. I need to have the signature of the common function have `OptionalJacobian<2, -1>` since the number of calibration parameters vary for different calibration models. However, when I try to do this, I get the error: 
    ```
    error: cannot convert ‘OptionalJacobian<2,9>’ to ‘OptionalJacobian<-1,-1>’
      192 |   Dcalibrate(pn, Dcal, Dp);
    ```
    This seems to be because the dynamic version of OptionalJacobian doesn't have a constructor from a fixed sized matrix.

2. I am not sure how to call the `uncalibrate` method for the subclasses. Calling `uncalibrate` always calls the base class `uncalibrate`, causing `Cal3Unified` results to be wrong. Any suggestions about this (other than using function passing)?

ProfFan:
I don't really see the problem here, I think just make the function virtual is enough?

varunagrawal:
Wouldn't virtual mean that the subclasses need to define their implementations of the function? That would make defeat the purpose.

varunagrawal:
@ProfFan I think I see what you mean. `uncalibrate` should be virtual, I agree. Thanks for the tip! :smile:

varunagrawal:
`virtual`/`override` has a similar problem because the OptionalJacobian sizes are different :disappointed:

dellaert:
I was not talking about a common method. I was talking a bout a *free* function that can be called even by calibration classes that have zero inheritance structure in common. So, virtual/override is moot. And, yes, it will have to be templated on the Jacobian dimensions.

varunagrawal:
Yeah I was considering a follow up PR to have a common base class for all calibration models since there is a lot of copy-pasta here. I'm creating a new file called `Cal3.h`.

gtsam/geometry/Cal3DS2_Base.h
```diff
@@ -39,22 +41,43 @@ namespace gtsam {
 class GTSAM_EXPORT Cal3DS2_Base {
 
 protected:
-
-  double fx_, fy_, s_, u0_, v0_ ; // focal length, skew and principal point
-  double k1_, k2_ ; // radial 2nd-order and 4th-order
-  double p1_, p2_ ; // tangential distortion
+ double fx_, fy_, s_, u0_, v0_;  // focal length, skew and principal point
```
dellaert, line 44: `///<` is doxygen one-liner convention

dellaert (approved):
Very nice !!!!

gtsam/geometry/Cal3Bundler.h
```diff
@@ -18,6 +18,7 @@
 
 #pragma once
 
+#include <gtsam/geometry/Cal3.h>
```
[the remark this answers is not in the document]
varunagrawal: `CalBundler.cpp` uses the `calibrateJacobians` method. I figured it was good style to keep includes in the header.

gtsam/geometry/Cal3DS2_Base.h
varunagrawal: I'm done refactoring all the calibration models. This has been taken care of in that PR.

PR #624, opened 2020-12-01
IMU Parameters Documentation

### LINKED ISSUE #213: IMUFactor example noise/bias units
Hi, I am trying to use GTSAM to create a system using IMU and GPS measurements together in a factor graph.

I was following the example IMUFactor. It has the following values which were being computed from the sensor specifications
```

// We use the sensor specs to build the noise model for the IMU factor.
  double accel_noise_sigma = 0.0003924;
  double gyro_noise_sigma = 0.000205689024915;
  double accel_bias_rw_sigma = 0.004905;
  double gyro_bias_rw_sigma = 0.000001454441043;
```

However there are no units attached to these. I am trying to better model the IMU I am using which is Bosch BMI160. I cannot find any values which resemble these. 

https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BMI160-DS000.pdf

Any help in adapting the sample to use this IMU would be of great help!

varunagrawal:
Fixes #213


dellaert:
Attach a PDF ?

varunagrawal:
The PDF is attached. It is the last of the 3 files, or am I misreading your comment?

cmake/GtsamTesting.cmake
```diff
@@ -88,7 +88,7 @@ enable_testing()
 
 option(GTSAM_BUILD_TESTS                 "Enable/Disable building of tests"          ON)
 option(GTSAM_BUILD_EXAMPLES_ALWAYS       "Build examples with 'make all' (build with 'make examples' if not)"       ON)
-	option(GTSAM_BUILD_TIMING_ALWAYS         "Build timing scripts with 'make all' (build with 'make timing' if not"    OFF)
+option(GTSAM_BUILD_TIMING_ALWAYS         "Build timing scripts with 'make all' (build with 'make timing' if not"    OFF)
```
dellaert, line 91: remove this from diff

dellaert (changes requested):
Sorry, but the table about units - which I gather is the only thing you added - does not make sense in this document. I encourage you to read the document (as I did, again, for the review), and then you'll see this table come as a complete surprise. Perhaps this table, which refers to variables in the IMU factor, should go as documentation in the code?

cmake/GtsamTesting.cmake
varunagrawal: Aye aye

dellaert (approved):
yay!

dellaert:
Feel free to merge and delete branch after checks passed :-)

varunagrawal:
Don't you love how unicode lets us put all the math symbols directly in an easy-to-read manner? :D

cmake/GtsamTesting.cmake
```diff
@@ -88,7 +88,7 @@ enable_testing()
 
 option(GTSAM_BUILD_TESTS                 "Enable/Disable building of tests"          ON)
 option(GTSAM_BUILD_EXAMPLES_ALWAYS       "Build examples with 'make all' (build with 'make examples' if not)"       ON)
-option(GTSAM_BUILD_TIMING_ALWAYS         "Build timing scripts with 'make all' (build with 'make timing' if not"    OFF)
+	option(GTSAM_BUILD_TIMING_ALWAYS         "Build timing scripts with 'make all' (build with 'make timing' if not"    OFF)
```
jlblancoc: oops... :-)

mintar:
I'm pretty sure that this is incorrect:

```c++
Vector3 biasAcc_; ///< The units for stddev are σ = m/s² or m √Hz/s²
Vector3 biasGyro_; ///< The units for stddev are σ = rad/s or rad √Hz/s
```
It should be:

```c++
Vector3 biasAcc_; ///< The units for stddev are σ = m √Hz/s²
Vector3 biasGyro_; ///< The units for stddev are σ = rad √Hz/s
```

See [my comment here](https://github.com/borglab/gtsam/issues/213#issuecomment-979843896) for an in-depth explanation.

dellaert:
@varunagrawal if you agree change it?

PR #626, opened 2020-12-01
Cal3 Base Class

varunagrawal:
I went ahead and created the common base class for all calibration models, which is `Cal3`.
This base class helps remove A LOT of redundancy and copy-pasta. I also took advantage of this opportunity to modernize the code as much as possible, using better conventions (such as base-class equality checking and `std::fabs`).

Also did a lot of formatting for parts that I've touched. As a result, this is a big PR and may need to go through multiple iterations.

### DIFF
--- a/gtsam/geometry/Cal3.cpp
+++ b/gtsam/geometry/Cal3.cpp
@@ -62,11 +62,11 @@ void Cal3::print(const std::string& s) const { gtsam::print((Matrix)K(), s); }
 
 /* ************************************************************************* */
 bool Cal3::equals(const Cal3& K, double tol) const {
-  return (std::abs(fx_ - K.fx_) < tol) && (std::abs(fy_ - K.fy_) < tol) &&
-         (std::abs(s_ - K.s_) < tol) && (std::abs(u0_ - K.u0_) < tol) &&
-         (std::abs(v0_ - K.v0_) < tol);
+  return (std::fabs(fx_ - K.fx_) < tol) && (std::fabs(fy_ - K.fy_) < tol) &&
+         (std::fabs(s_ - K.s_) < tol) && (std::fabs(u0_ - K.u0_) < tol) &&
+         (std::fabs(v0_ - K.v0_) < tol);
 }
 
 /* ************************************************************************* */
 
-}  // namespace gtsam
+}  // \ namespace gtsam
--- a/gtsam/geometry/Cal3.cpp
+++ b/gtsam/geometry/Cal3.cpp
@@ -22,28 +22,25 @@
 #include <iostream>
 
 namespace gtsam {
-using namespace std;
 
 /* ************************************************************************* */
 Cal3::Cal3(double fov, int w, int h)
     : s_(0), u0_((double)w / 2.0), v0_((double)h / 2.0) {
   double a = fov * M_PI / 360.0;  // fov/2 in radians
-  fx_ =
-      (double)w / (2.0 * tan(a));  //    old formula: fx_ = (double) w * tan(a);
+  // old formula: fx_ = (double) w * tan(a);
+  fx_ = double(w) / (2.0 * tan(a));
   fy_ = fx_;
 }
 
 /* ************************************************************************* */
 Cal3::Cal3(const std::string& path)
     : fx_(320), fy_(320), s_(0), u0_(320), v0_(140) {
-  char buffer[200];
-  buffer[0] = 0;
-  sprintf(buffer, "%s/calibration_info.txt", path.c_str());
+  const auto buffer = path + std::string("/calibration_info.txt");
   std::ifstream infile(buffer, std::ios::in);
 
-  if (infile)
+  if (infile) {
     infile >> fx_ >> fy_ >> s_ >> u0_ >> v0_;
-  else {
+  } else {
     throw std::runtime_error("Cal3: Unable to load the calibration");
   }
 
@@ -51,9 +48,9 @@ Cal3::Cal3(const std::string& path)
 }
 
 /* ************************************************************************* */
-ostream& operator<<(ostream& os, const Cal3& cal) {
-  os << "{fx: " << cal.fx() << ", fy: " << cal.fy() << ", s:" << cal.skew()
-     << ", px:" << cal.px() << ", py:" << cal.py() << "}";
+std::ostream& operator<<(std::ostream& os, const Cal3& cal) {
+  os << "{ fx: " << cal.fx() << ", fy: " << cal.fy() << ", s: " << cal.skew()
+     << ", px: " << cal.px() << ", py: " << cal.py() << " }";
   return os;
 }
 
@@ -62,9 +59,9 @@ void Cal3::print(const std::string& s) const { gtsam::print((Matrix)K(), s); }
 
 /* ************************************************************************* */
 bool Cal3::equals(const Cal3& K, double tol) const {
-  return (std::fabs(fx_ - K.fx_) < tol) && (std::fabs(fy_ - K.fy_) < tol) &&
-         (std::fabs(s_ - K.s_) < tol) && (std::fabs(u0_ - K.u0_) < tol) &&
-         (std::fabs(v0_ - K.v0_) < tol);
+  return (std::fabs(fx_ - K.fx_) < tol && std::fabs(fy_ - K.fy_) < tol &&
+          std::fabs(s_ - K.s_) < tol && std::fabs(u0_ - K.u0_) < tol &&
+          std::fabs(v0_ - K.v0_) < tol);
 }
 
 /* ************************************************************************* */
--- a/gtsam/geometry/Cal3.cpp
+++ b/gtsam/geometry/Cal3.cpp
@@ -33,12 +33,11 @@ Cal3::Cal3(double fov, int w, int h)
 }
 
 /* ************************************************************************* */
-Cal3::Cal3(const std::string& path)
-    : fx_(320), fy_(320), s_(0), u0_(320), v0_(140) {
+Cal3::Cal3(const std::string& path) {
   const auto buffer = path + std::string("/calibration_info.txt");
   std::ifstream infile(buffer, std::ios::in);
 
-  if (infile) {
+  if (infile && !infile.eof()) {
     infile >> fx_ >> fy_ >> s_ >> u0_ >> v0_;
   } else {
     throw std::runtime_error("Cal3: Unable to load the calibration");
--- a/gtsam/geometry/Cal3.cpp
+++ b/gtsam/geometry/Cal3.cpp
@@ -1,0 +1,72 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file   Cal3.cpp
+ * @brief  Common code for all calibration models.
+ * @author Frank Dellaert
+ */
+
+#include <gtsam/geometry/Cal3.h>
+
+#include <cmath>
+#include <fstream>
+#include <iostream>
+
+namespace gtsam {
+using namespace std;
+
+/* ************************************************************************* */
+Cal3::Cal3(double fov, int w, int h)
+    : s_(0), u0_((double)w / 2.0), v0_((double)h / 2.0) {
+  double a = fov * M_PI / 360.0;  // fov/2 in radians
+  fx_ =
+      (double)w / (2.0 * tan(a));  //    old formula: fx_ = (double) w * tan(a);
+  fy_ = fx_;
+}
+
+/* ************************************************************************* */
+Cal3::Cal3(const std::string& path)
+    : fx_(320), fy_(320), s_(0), u0_(320), v0_(140) {
+  char buffer[200];
+  buffer[0] = 0;
+  sprintf(buffer, "%s/calibration_info.txt", path.c_str());
+  std::ifstream infile(buffer, std::ios::in);
+
+  if (infile)
+    infile >> fx_ >> fy_ >> s_ >> u0_ >> v0_;
+  else {
+    throw std::runtime_error("Cal3: Unable to load the calibration");
+  }
+
+  infile.close();
+}
+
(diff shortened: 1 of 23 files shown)


gtsam/geometry/Cal3.cpp
```diff
@@ -0,0 +1,72 @@
...
+    : s_(0), u0_((double)w / 2.0), v0_((double)h / 2.0) {
+  double a = fov * M_PI / 360.0;  // fov/2 in radians
+  fx_ =
+      (double)w / (2.0 * tan(a));  //    old formula: fx_ = (double) w * tan(a);
+  fy_ = fx_;
+}
+
+/* ************************************************************************* */
+Cal3::Cal3(const std::string& path)
+    : fx_(320), fy_(320), s_(0), u0_(320), v0_(140) {
```
jlblancoc, line 38: I would prefer `double fx_ = 320.0f;` etc. in the .h (C++11 feature) instead of here at the .cpp. It's easier for the reader to see what are the defaults, and makes harder to be inconsistent between different ctors.

gtsam/geometry/Cal3.cpp
```diff
@@ -0,0 +1,72 @@
...
+  char buffer[200];
+  buffer[0] = 0;
+  sprintf(buffer, "%s/calibration_info.txt", path.c_str());
```
jlblancoc, line 41: Security issue: potential out-of-buffer write. 
Why not?
```
const auto buffer = path  + std::string("/calibration_info.txt");
``` 
(would be shorter with c++14's [s-operator](https://en.cppreference.com/w/cpp/string/basic_string/operator%22%22s), but gtsam uses c++11 only).

gtsam/geometry/Cal3.cpp
```diff
@@ -0,0 +1,72 @@
...
+  std::ifstream infile(buffer, std::ios::in);
+
+  if (infile)
+    infile >> fx_ >> fy_ >> s_ >> u0_ >> v0_;
```
jlblancoc, line 45: error check:  `if (infile)` at the end?

gtsam/geometry/Cal3.h
```diff
@@ -61,6 +61,133 @@ void calibrateJacobians(const Cal& calibration, const Point2& pn,
...
+ public:
+  enum { dimension = 5 };
+  ///< shared pointer to calibration object
+  using shared_ptr = boost::shared_ptr<Cal3>;
+
+  /// @name Standard Constructors
+  /// @{
+
+  /// Create a default calibration that leaves coordinates unchanged
+  Cal3() : fx_(1), fy_(1), s_(0), u0_(0), v0_(0) {}
```
jlblancoc, line 82: refer to comment above. If defaults are like `double fx_ = 1;` then : `Cal3() = default;`.

gtsam/geometry/Cal3.h
```diff
@@ -61,6 +61,133 @@ void calibrateJacobians(const Cal& calibration, const Point2& pn,
...
+
+  /// constructor from doubles
+  Cal3(double fx, double fy, double s, double u0, double v0)
+      : fx_(fx), fy_(fy), s_(s), u0_(u0), v0_(v0) {}
+
+  /// constructor from vector
+  Cal3(const Vector& d)
+      : fx_(d(0)), fy_(d(1)), s_(d(2)), u0_(d(3)), v0_(d(4)) {}
```
jlblancoc, line 90: throw is unexpected vector length?

gtsam/geometry/Cal3.h
```diff
@@ -61,6 +61,133 @@ void calibrateJacobians(const Cal& calibration, const Point2& pn,
...
+   * @param w image width
+   * @param h image height
+   */
+  Cal3(double fov, int w, int h);
+
+  /// @}
+  /// @name Advanced Constructors
+  /// @{
+
+  /// load calibration from location (default name is calibration_info.txt)
```
jlblancoc, line 104: Document expected file content format...

gtsam/geometry/Cal3.h
```diff
@@ -61,6 +61,133 @@ void calibrateJacobians(const Cal& calibration, const Point2& pn,
...
+    return K;
+  }
+
+#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V41
+  /** @deprecated The following function has been deprecated, use K above */
+  Matrix3 matrix() const { return K(); }
+#endif
+
+  /// return inverted calibration matrix inv(K)
+  Matrix3 matrix_inverse() const {
```
jlblancoc, line 166: This method name is all lowercase, principalPoint is camel case... (??).

jlblancoc (commented):
Good! I just left some random comments...

gtsam/geometry/Cal3.cpp
varunagrawal: I like your suggestion, but the default constructor uses `fx_ = 1`, `fy_ = 1` etc. Should I replace that with the defaults from this constuctor (from string path)?

gtsam/geometry/Cal3.h
varunagrawal: Ah great catch. It would be an API change, but renaming it to `inverse()` makes more sense?

gtsam/geometry/Cal3.cpp
jlblancoc: Hmm... good point. I think you should leave them like `double fx_ = 1.0;` etc. for consistency with the default ctor, then here in this ctor from a file, the defaults actually shouldn't matter since at the ctor should either: a) return with the new values from the file, or b) throw an exception if error loading... right?

gtsam/geometry/Cal3.h
varunagrawal: Nice catch! I updated all the vector constructors to have fixed size vectors so this is caught at compile-time (aka no nasty surprises).

gtsam/geometry/Cal3.cpp
varunagrawal: That's true. Let me update that really quick.

gtsam/geometry/Cal3.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <fstream>
+#include <iostream>
+
+namespace gtsam {
+
+/* ************************************************************************* */
+Cal3::Cal3(double fov, int w, int h)
+    : s_(0), u0_((double)w / 2.0), v0_((double)h / 2.0) {
+  double a = fov * M_PI / 360.0;  // fov/2 in radians
+  // old formula: fx_ = (double) w * tan(a);
```
dellaert, line 30: ?

gtsam/geometry/Cal3.h
```diff
@@ -61,6 +61,143 @@ void calibrateJacobians(const Cal& calibration, const Point2& pn,
...
+    return K;
+  }
+
+#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V41
+  /** @deprecated The following function has been deprecated, use K above */
+  Matrix3 matrix() const { return K(); }
+#endif
+
+  /// Return inverted calibration matrix inv(K)
+  Matrix3 inverse() const {
```
dellaert, line 176: move to cpp?

gtsam/geometry/Cal3Bundler.h
```diff
@@ -28,23 +28,25 @@ namespace gtsam {
...
-class GTSAM_EXPORT Cal3Bundler {
+class GTSAM_EXPORT Cal3Bundler : public Cal3 {
 
-private:
-  double f_; ///< focal length
-  double k1_, k2_; ///< radial distortion
-  double u0_, v0_; ///< image center, not a parameter to be optimized but a constant
-  double tol_; ///< tolerance value when calibrating
+ private:
+  double f_ = 1.0f;               ///< focal length
```
dellaert, line 34: why do we still have this?

gtsam/geometry/Cal3Bundler.h
```diff
@@ -109,6 +111,11 @@ class GTSAM_EXPORT Cal3Bundler {
     return v0_;
   }
 
+  Matrix3 K() const; ///< Standard 3*3 calibration matrix
```
dellaert, line 114: this is in base class, should be omitted or made override?

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -52,18 +42,23 @@ Vector3 Cal3Bundler::vector() const {
...
 }
 
 /* ************************************************************************* */
 bool Cal3Bundler::equals(const Cal3Bundler& K, double tol) const {
-  if (std::abs(f_ - K.f_) > tol || std::abs(k1_ - K.k1_) > tol
-      || std::abs(k2_ - K.k2_) > tol || std::abs(u0_ - K.u0_) > tol
-      || std::abs(v0_ - K.v0_) > tol)
-    return false;
-  return true;
+  return (std::fabs(f_ - K.f_) < tol && std::fabs(k1_ - K.k1_) < tol &&
```
dellaert, line 59: what about base? And I think f should go :-)

gtsam/geometry/Cal3DS2_Base.h
```diff
@@ -34,65 +34,58 @@ namespace gtsam {
  * but using only k1,k2,p1, and p2 coefficients.
  * K = [ fx s u0 ; 0 fy v0 ; 0 0 1 ]
  * rr = Pn.x^2 + Pn.y^2
- * \hat{Pn} = (1 + k1*rr + k2*rr^2 ) Pn + [ 2*p1 Pn.x Pn.y + p2 (rr + 2 Pn.x^2) ;
+ * \hat{Pn} = (1 + k1*rr + k2*rr^2 ) Pn + [ 2*p1 Pn.x Pn.y + p2 (rr + 2 Pn.x^2)
```
dellaert, line 37: What, no unicode??? ;-)

gtsam/geometry/Cal3Fisheye.h
```diff
@@ -160,17 +151,21 @@ class GTSAM_EXPORT Cal3Fisheye {
...
 
-  /// Given delta vector, update calibration
-  Cal3Fisheye retract(const Vector& d) const;
-
-  /// Given a different calibration, calculate update to obtain it
-  Vector localCoordinates(const Cal3Fisheye& T2) const;
-
   /// Return dimensions of calibration manifold object
-  virtual size_t dim() const { return 9; }
+  virtual size_t dim() const { return dimension; }
```
dellaert, line 155: add to base and override if not just `return Dim(); `

gtsam/geometry/Cal3Fisheye.h
```diff
@@ -160,17 +151,21 @@ class GTSAM_EXPORT Cal3Fisheye {
...
 
   /// Return dimensions of calibration manifold object
-  static size_t Dim() { return 9; }
+  static size_t Dim() { return dimension; }
```
dellaert, line 158: THis might be harder to put in base class, with static involved?

gtsam/geometry/Cal3_S2.cpp
```diff
@@ -22,95 +22,53 @@
...
-    if(Dp)
-        *Dp << inv_fx, -inv_fx_s_inv_fy, 0, inv_fy;
-    return point;
+Point2 Cal3_S2::calibrate(const Point2& p, OptionalJacobian<2, 5> Dcal,
+                          OptionalJacobian<2, 2> Dp) const {
+  const double u = p.x(), v = p.y();
+  double delta_u = u - u0_, delta_v = v - v0_;
+  double inv_fx = 1 / fx_, inv_fy = 1 / fy_;
+  double inv_fy_delta_v = inv_fy * delta_v,
+         inv_fx_s_inv_fy = inv_fx * s_ * inv_fy;
```
dellaert, line 59: spacing?

gtsam/geometry/tests/testCal3_S2Stereo.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/* ************************************************************************* */
+TEST(Cal3_S2Stereo, calibrate_homogeneous) {
+  Vector3 intrinsic(2, 3, 1);
+  Vector3 image(1320.3, 1740, 1);
+  CHECK(assert_equal((Vector)intrinsic, (Vector)K.calibrate(image)));
+}
+
+//TODO(Varun) Add calibrate and uncalibrate methods
```
dellaert, line 60: Stick these in a different branch, PR, rather than landing commented out code.

dellaert (approved):
A bunch of comments, but awe-inspiring nice PR !

gtsam/geometry/Cal3.cpp
varunagrawal: Yeah I should just kill that.

gtsam/geometry/Cal3Bundler.h
varunagrawal: Cal3Bundler doesn't use `fx_` or `fy_` like the base class, just a single, common `f` value. Though now that I think about it, I can do something about it.

gtsam/geometry/Cal3Bundler.h
varunagrawal: Should override, since we should force skew to be 0, even if it is read from a file.

gtsam/geometry/Cal3Bundler.cpp
varunagrawal: Nice catch. Don't know how I missed that, then again this is a ginormous PR.

gtsam/geometry/Cal3DS2_Base.h
varunagrawal: Aye aye, Cap'n.

gtsam/geometry/Cal3Fisheye.h
varunagrawal: Each model has a different dimension (though some have common values for `dimension`), so I'll have to pick and choose.

varunagrawal: Actually, I can just get rid of the subclass `dim()` methods and use your suggestion of `return Dim()`.

gtsam/geometry/Cal3Fisheye.h
varunagrawal: Yes especially since we are using this to differentiate the various dimensionalities.

gtsam/geometry/Cal3Fisheye.h
varunagrawal: Just checked with a unit test: we will need to override `dim()`. :disappointed:

varunagrawal:
All done! I'll merge once CI passes. :tada:

PR #629, opened 2020-12-03
SymbolGenerator: add chr() and made constexpr-capable

jlblancoc:
That it is: 
- add a `chr()` method to recover the underlying char symbol, and
- Make possible to write things like:

```cpp
constexpr auto sQpp = gtsam::SymbolGenerator('a');
//...

    for (const auto &kv : values) {
      gtsam::Symbol s(kv.key);
      switch (s.chr()) {
      case sQpp.chr():    // this requires SymbolGenerator to be constexpr
        // ...
      }
   }


```


varunagrawal (approved):
LGTM

PR #632, opened 2020-12-05
NonlinearFactorGraph saveGraph

varunagrawal:
Function to save graph as graphviz file via wrapper.

### DIFF
--- a/gtsam/nonlinear/NonlinearFactorGraph.h
+++ b/gtsam/nonlinear/NonlinearFactorGraph.h
@@ -11,7 +11,7 @@
 
 /**
  * @file    NonlinearFactorGraph.h
- * @brief   Factor Graph Constsiting of non-linear factors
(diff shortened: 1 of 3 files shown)


gtsam/nonlinear/NonlinearFactorGraph.h
```diff
@@ -115,6 +115,11 @@ namespace gtsam {
     void saveGraph(std::ostream& stm, const Values& values = Values(),
       const GraphvizFormatting& graphvizFormatting = GraphvizFormatting(),
       const KeyFormatter& keyFormatter = DefaultKeyFormatter) const;
+  
+    /** Write the graph in GraphViz format to file for visualization */
```
dellaert, line 119: Mention that this is only here to make wrapping possible, as wrapped languages do not have access to C++ streams.

dellaert (approved):

Issue #627, opened 2020-12-02
Problem with IMU preintegration during impacts


raabuchanan:
There may be an issue with IMU preintegration for impacts and impulses. Bellow is a link to simulated data from an IMU in a robot's foot as it takes a step. The figure shows the raw acceleration and rotation velocities in the top two plots. The bottom plot is the result of continuous IMU preintegration and the Z velocity is shown in cyan. The velocity should return to zero as the foot makes contact but instead stays negative.

![simulation-rf](https://user-images.githubusercontent.com/5701905/100895847-e7f12e80-34b5-11eb-8f79-55eb774610e4.png)

[Step Dataset](https://drive.google.com/drive/folders/1sFaGuOueWxV4IAz65uJ9IrYTLNa67zja?usp=sharing) 
rosbag and csvs are provided. In the rosbag, the relevant topics are:

- /anymal/RH_FLATFOOT_IMU (IMU signal)
- /kins_ros/propagated_state (IMU preintegration)
  - custom message definition is provided in link

tmcg0:
Forgive me if this is a naïve response: but it doesn't look like you have any position reference like a GPS. If you're merely doing pure IMU gyro/accelerometer dead reckoning, velocity drift due to sensor noise is expected, right?

dellaert:
Thanks @raabuchanan ! Tim, you are spot on, but it would not drift this far in this short amount of time. We actually asked Russell to create this issue to debug the IMU factor in the face of impacts. 

@raabuchanan I assume you will also create a PR that has the unit test in it?

raabuchanan:
@dellaert Yes the unit test is coming

@tmcg0 In addition to Frank's point, this is simulated data without noise. Other kinds of motions without impacts have virtually no drift over several minutes.

tmcg0:
@raabuchanan Great point, I totally missed that this was simulated data.

raabuchanan:
See PR #633

PR #633, opened 2020-12-08
Adds unit test for imu preintegration of a single step

raabuchanan:
Related to issue #627 

This adds a unit test which integrates simulated data from an IMU in a foot while to foot takes a single step. See issue for related dataset.

Currently the unit test actually passes which is in contradiction to what I am seeing in the issue. I need to investigate the problem further but wanted to put up this unit test to keep everyone in the loop. Also someone may spot something wrong that I'm doing. I will write more unit tests to try and understand the problem.


varunagrawal:
I believe we can merge all 3 files into 1 so we have a comprehensive test file that is easy to modify. While I like the structure a lot, it seems like a lot of added complexity.

varunagrawal:
@raabuchanan I've made some updates to the code to make it agnostic to users. I'd love to work with you on this PR and figure this issue out.

Currently, I am a bit confused on what are we testing for specifically? Is there a specific assertion you want to test for?

dellaert:
@varunagrawal , please help land? Also, don’t force-push :-)

varunagrawal:
Since @raabuchanan got the Imu impact working, what should we be testing for? The same thing as before?

@raabuchanan please could you refresh my memory on the goal of this test? I believe you had a slide on it.

raabuchanan:
Hey @varunagrawal yes I fixed the problem on my side. I've just pushed a small cleanup now.

raabuchanan:
Hi @varunagrawal yes I fixed the issue on my side and this passes now. This test checks a sequence of simulated IMU data that the relative Z position is zero. It also checks that there are no rotations. I will push a small cleanup soon but I can add more documentation if you want.

varunagrawal:
Awesome. I'll clean it up a little bit and then we can merge.

tests/testImuPreintegration.cpp
```diff
@@ -0,0 +1,131 @@
...
+      results.push_back(atof(str.c_str()));
+    }
+    ImuMeasurement measurement;
+    measurement.dt = static_cast<size_t>(1e9 * (1 / rate));
+    measurement.time = results[2];
+    measurement.I_a_WI = {results[29], results[30], results[31]};
+    measurement.I_w_WI = {results[17], results[18], results[19]};
+    imuMeasurements.push_back(measurement);
+
+    // cout << "IMU measurement " << measurement << endl;
```
varunagrawal, line 68: We should get rid of the print statements. Tests should be silent and fast.

tests/testImuPreintegration.cpp
```diff
@@ -0,0 +1,131 @@
...
+  for (size_t n = 1; n < imuMeasurements.size(); n++) {
+    // integrate
+    imuPreintegrated.integrateMeasurement(imuMeasurements[n].I_a_WI,
+                                          imuMeasurements[n].I_w_WI, 1 / rate);
+    // predict
+    propState = imuPreintegrated.predict(initialNavState, priorImuBias);
+    position = propState.pose().translation();
+    velocity = propState.velocity();
+    // cout << "IMU Position " << position.transpose() << endl;
+    // cout << "IMU Velocity " << velocity.transpose() << endl;
```
varunagrawal, line 111: Need to remove these prints.

varunagrawal (approved):
Very minor comments, but LGTM.

PR #644, opened 2020-12-28
Wrapping prior factor and general sfm factor or Cal3Bundler

ayushbaid:
The python API has support for `PriorFactor` and `GeneralSFMFactor2` for calibration class `Cal3_S2`, but not for `Cal3Bundler`.

In this PR, we add the same wraps for `Cal3Bundler` calibration class.

### DIFF
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2520,7 +2520,7 @@ class NonlinearISAM {
 #include <gtsam/geometry/StereoPoint2.h>
 
 #include <gtsam/nonlinear/PriorFactor.h>
-template<T = {Vector, gtsam::Point2, gtsam::StereoPoint2, gtsam::Point3, gtsam::Rot2, gtsam::SO3, gtsam::SO4, gtsam::SOn, gtsam::Rot3, gtsam::Pose2, gtsam::Pose3, gtsam::Unit3, gtsam::Cal3_S2, gtsam::Cal3Bundler, gtsam::CalibratedCamera, gtsam::PinholeCameraCal3_S2, gtsam::imuBias::ConstantBias, gtsam::PinholeCamera<gtsam::Cal3Bundler>}>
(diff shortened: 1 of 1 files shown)


gtsam/gtsam.i
```diff
@@ -2666,14 +2666,17 @@ typedef gtsam::GeneralSFMFactor<gtsam::PinholeCameraCal3_S2, gtsam::Point3> Gene
 typedef gtsam::GeneralSFMFactor<gtsam::PinholeCameraCal3DS2, gtsam::Point3> GeneralSFMFactorCal3DS2;
 typedef gtsam::GeneralSFMFactor<gtsam::PinholeCamera<gtsam::Cal3Bundler>, gtsam::Point3> GeneralSFMFactorCal3Bundler;
 
-template<CALIBRATION = {gtsam::Cal3_S2}>
+template<CALIBRATION>
```
varunagrawal, line 2669: Since this is only 1 template, you can just list it out like so:
```
template<CALIBRATION = {gtsam::Cal3_S2, gtsam::Cal3DS2, gtsam::Cal3Bundler}>
```

and remove the typedefs.

varunagrawal, line 2669: Since this is a single template parameter, you can just list out the template arguments and get rid of the typedefs, like this:

```cpp
template<CALIBRATION = {gtsam::Cal3_S2, gtsam::Cal3DS2, gtsam::Cal3Bundler}>
```

varunagrawal (changes requested):
Great work, just 1 small nitpick.

dellaert (approved):
Please try Varun's suggestion, then feel free to merge.

gtsam/gtsam.i
ayushbaid: This works! Thanks!

varunagrawal (approved):
LGTM

PR #645, opened 2020-12-29
Python support for CMake version < 3.12

varunagrawal:
I got a bit too aggressive with my `FindPython` requirements and this is causing cmake issues for people, so adding support for CMake <3.12.


varunagrawal:
@dlee640 this might help you out.

PR #647, opened 2020-12-31
Pose3 interpolateRt method

varunagrawal:
New interpolateRt method that might be more intuitive than minimizing geodesic distance.

### DIFF
--- a/gtsam/geometry/tests/testPose3.cpp
+++ b/gtsam/geometry/tests/testPose3.cpp
@@ -1041,8 +1041,8 @@ TEST(Pose3, interpolate) {
 
   // Non-trivial interpolation, translation value taken from output.
   Pose3 expected2(interpolate(T2.rotation(), T3.rotation(), t),
-                  Point3(3.39016299515, -7.62798017104, 5.03165553689));
-  EXPECT(assert_equal(expected2, interpolate(T2, T3, t)));
+                  interpolate(T2.translation(), T3.translation(), t));
+  EXPECT(assert_equal(expected2, T2.interpolateRt(T3, t)));
 }
 
 /* ************************************************************************* */
--- a/gtsam/geometry/tests/testPose3.cpp
+++ b/gtsam/geometry/tests/testPose3.cpp
@@ -1016,6 +1016,20 @@ TEST(Pose3, TransformCovariance6) {
 TEST(Pose3, interpolate) {
   EXPECT(assert_equal(T2, interpolate(T2,T3, 0.0)));
   EXPECT(assert_equal(T3, interpolate(T2,T3, 1.0)));
+
+  // Example from Peter Corke
+  // https://robotacademy.net.au/lesson/interpolating-pose-in-3d/
+  double s = 0.0759; // corresponds to the 10th element when calling `ctraj` in the video
+  Pose3 O;
+  Pose3 F(Rot3::Roll(0.6).compose(Rot3::Pitch(0.8)).compose(Rot3::Yaw(1.4)), Point3(1, 2, 3));
+
+  // The expected answer matches the result presented in the video.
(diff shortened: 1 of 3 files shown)


dellaert (changes requested):
Actually, the old code is correct. The interpolate function is *defined* as doing interpolation in tangent space. If the rotations of both endpoints are different, there is no expectation that the translation linearly interpolates.

If the rotations are the *same*, I think it does linearly interpolate translation (but I'm not even sure of that).

varunagrawal:
> Actually, the old code is correct. The interpolate function is _defined_ as doing interpolation in tangent space. If the rotations of both endpoints are different, there is no expectation that the translation linearly interpolates.
> 
> If the rotations are the _same_, I think it does linearly interpolate translation (but I'm not even sure of that).

I though about that at first, that perhaps the interpolation of the translations needn't be linear betweem them if the rotations are different, but after looking it up and [checking the implementation of other libraries](https://github.com/ros/geometry2/blob/7ea64d25a0631fe91bd277e0d9195f18cc437151/tf2/src/cache.cpp#L176-L196), it seems that the translations are indeed linear even if the rotations are different.

This is one of the reasons I put that example from Peter Corke's video lecture as a unit test. It is based off his Robotics Toolbox for Matlab.

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -1016,6 +1016,33 @@ TEST(Pose3, TransformCovariance6) {
...
   EXPECT(assert_equal(T3, interpolate(T2,T3, 1.0)));
+
+  // Trivial example: start at origin and move to (1, 0, 0) while rotating pi/2
+  // about z-axis.
+  Pose3 start;
+  Pose3 end(Rot3::Rz(M_PI_2), Point3(1, 0, 0));
+  // This interpolation is easy to calculate by hand.
+  double t = 0.5;
+  Pose3 expected0(Rot3::Rz(M_PI_4), Point3(0.5, 0, 0));
+  EXPECT(assert_equal(expected0, interpolate(start, end, t)));
```
varunagrawal, line 1027: @dellaert I added this simple test to verify since it is easy to visualize and imaging: a body turning about Z as it moves from the origin to `(1, 0, 0)`.

The previous code fails with this result:
```
Not equal:
expected:
 R: [
	0.707107, -0.707107, 0;
	0.707107, 0.707107, 0;
	0, 0, 1
]
t: 0.5   0   0
actual:
 R: [
	0.707107, -0.707107, 0;
	0.707107, 0.707107, 0;
	0, 0, 1
]
t:       0.5 -0.207107         0
```

It doesn't make any sense why the body would off the X axis.

dellaert: This is because the interpolation is in *tangent* space. Straight lines in tangent space correspond to screw motions.

varunagrawal: Can you please provide a reference I can read up on?

dellaert: https://en.wikipedia.org/wiki/Slerp

The reference is the documentation of "interpolate". I think you expected something else, which is fine. I emailed with a different approach (interpolateRt). You can document the difference with interpolate<Pose3> there.

dellaert (changes requested):
Cool. Please document the difference with interpolate<Pose3> in the doxygen? And I'd also change the test to use interpolateRt now, right? interpolate<Rot3> should be tested in testRot3, and interpolate<Point3> should be tested in testPoint3.

varunagrawal:
I see. By shortest distance I meant straight line. Will update.

dellaert (approved):
Now it's perfect :-)

PR #649, opened 2021-01-01
Add dataset directly to package

### LINKED ISSUE #648: findExampleDataFiles broken on macOS
## Description

`findExampleDataFile` should work on macOS but it does not.

## Steps to reproduce

1. `pip3 install --user --pre gtsam`
2. try find any data file.

## Expected behavior

Should return the path of data file.

## Environment

macOS Big Sur, Python 3.9.1

## Additional information

Upon inspection the data files were installed to 
```
/Users/proffan/Library/Python/3.9/gtsam/Data/sphere2500.txt
```
not 
```
/Users/proffan/Library/Python/3.9/lib/python/site-packages/gtsam/Data/sphere2500.txt
```

I think this behavior is wrong as this is the "root" of the directory hierarchy.

Furthermore, a non-user install trys to write to `/usr/local/gtsam`, again not correct.

I may be wrong here as I have touched the wheels myself before uploading, though. Need someone to test a from-source install and the wheel to be sure.

varunagrawal:
This PR fixes #648.

### DIFF
--- a/.github/scripts/python.sh
+++ b/.github/scripts/python.sh
@@ -43,7 +43,7 @@ if [ -z ${PYTHON_VERSION+x} ]; then
     exit 127
 fi
 
-PYTHON="python${PYTHON_VERSION}"
+PYTHON="python${PYTHON_MAJOR_VERSION}"
 
 if [[ $(uname) == "Darwin" ]]; then
     brew install wget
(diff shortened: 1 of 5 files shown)


ProfFan (approved):
Great, thanks Varun!

So now data goes to `site-packages/gtsam/example/Data`, right?

varunagrawal:
Yes it should!

varunagrawal:
I made a small mistake in the inclusion patterns which is why the build failed. Should work now.

varunagrawal:
@ProfFan this now stores the data in `site-packages/gtsam/Data` (no `example`) but that is fine since the definition in the python utils looks for the data in that location.

Main point is that the data is now stored under the module's directory tree. :tada:

ProfFan:
Great! :)

PR #653, opened 2021-01-04
Updates to use the new version of wrap

varunagrawal:
This PR adds a bunch of changes so that the updates made to `wrap` to be used as a package still work seamlessly in the current set up of GTSAM.


varunagrawal:
Shoot, the issue is still here. Closing.

PR #656, opened 2021-01-05
Fix/warnings

dellaert:
Fixed warnings I found when compiling with latest clang.
This includes forgotten virtual destructor in Cal3.


varunagrawal (commented):
Cool! Any reason we're switching from const references in the for-ranges to const values?

dellaert:
> Cool! Any reason we're switching from const references in the for-ranges to const values?

The boost-created range functionality does not yield references. That's what the warning was.

dellaert:
@varunagrawal when you approve, feel free to immediately merge.

varunagrawal (approved):
LGTM

PR #657, opened 2021-01-05
Fix warnings

varunagrawal:
This PR fixes a bunch of cmake and compiler warnings.

### DIFF
--- a/gtsam/3rdparty/metis/CMakeLists.txt
+++ b/gtsam/3rdparty/metis/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.0)
 project(METIS)
 
 # Add flags for currect directory and below
--- a/gtsam/nonlinear/tests/testValues.cpp
+++ b/gtsam/nonlinear/tests/testValues.cpp
@@ -175,9 +175,7 @@ TEST(Values, basic_functions)
 {
   Values values;
   const Values& values_c = values;
-  Matrix23 M1, M2;
-  M1 << 0, 0, 0, 0, 0, 0;
-  M2 << 0, 0, 0, 0, 0, 0;
+  Matrix23 M1 = Matrix23::Zero(), M2 = Matrix23::Zero();
   values.insert(2, Vector3(0, 0, 0));
   values.insert(4, Vector3(0, 0, 0));
   values.insert(6, M1);
--- a/gtsam/nonlinear/tests/testValues.cpp
+++ b/gtsam/nonlinear/tests/testValues.cpp
@@ -175,10 +175,13 @@ TEST(Values, basic_functions)
 {
   Values values;
   const Values& values_c = values;
-  values.insert(2, Vector3());
-  values.insert(4, Vector3());
-  values.insert(6, Matrix23());
-  values.insert(8, Matrix23());
+  Matrix23 M1, M2;
+  M1 << 0, 0, 0, 0, 0, 0;
+  M2 << 0, 0, 0, 0, 0, 0;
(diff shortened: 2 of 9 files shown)


gtsam/nonlinear/tests/testValues.cpp
```diff
@@ -175,10 +175,13 @@ TEST(Values, basic_functions)
...
-  values.insert(2, Vector3());
-  values.insert(4, Vector3());
-  values.insert(6, Matrix23());
-  values.insert(8, Matrix23());
+  Matrix23 M1, M2;
+  M1 << 0, 0, 0, 0, 0, 0;
+  M2 << 0, 0, 0, 0, 0, 0;
+  values.insert(2, Vector3(0, 0, 0));
+  values.insert(4, Vector3(0, 0, 0));
+  values.insert(6, M1);
```
dellaert, line 183: How about Z_2x3 ?

gtsam/3rdparty/metis/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
```
dellaert, line 1: We can't make changes inside metis. Licensing and upgrade difficulties.

gtsam/slam/GeneralSFMFactor.h
```diff
@@ -70,6 +70,7 @@ class GeneralSFMFactor: public NoiseModelFactor2<CAMERA, LANDMARK> {
 protected:
 
   Point2 measured_; ///< the 2D measurement
+  bool verbose_; ///< Flag for print verbosity
```
dellaert, line 73: Adding a boolean flag for 100k factors in an SFM problem. Problematic. Propose to leave this out of this PR and discuss a better strategy

dellaert (changes requested):
Nice. But. Some issues, in comments.

gtsam/3rdparty/metis/CMakeLists.txt
varunagrawal: Interesting. I saw the git-blame for this CMake file and this file has been updated before, hence I went ahead. Does this new information change anything?

gtsam/nonlinear/tests/testValues.cpp
varunagrawal: `Z_2x3` isn't defined but I see what you mean.

gtsam/3rdparty/metis/CMakeLists.txt
dellaert: Actually, maybe the Makefile is what we added, and we should not make changes in the library itself.

gtsam/slam/GeneralSFMFactor.h
varunagrawal: Alright. Should I leave the exception printing in? Ideally that should be part of a logging framework. I'll already looked up some options.

dellaert: No, no printing. Let's leave for when we have logging, which would be great. I do like glog :-)

gtsam/3rdparty/metis/CMakeLists.txt
varunagrawal: Yup, the library is unchanged. :slightly_smiling_face: Guess we'll keep this update.

gtsam/slam/GeneralSFMFactor.h
varunagrawal: Yeah glog and CodeCov (for unit test code coverage) are two things on my radar. I'll create an issue for it that someone else can tackle.

dellaert (approved):

PR #659, opened 2021-01-06
Adding default point color to `SfmTrack` to fix equality check

ayushbaid:
The `equals` method fails right now when we run bundle adjustment on the same input twice. I checked that all the 2D measurements and 3d point match, but the color comparison fails. 

Adding a default value fixes the comparison.


varunagrawal (approved):
LGTM though it would be nice to have methods to set/get those values. Perhaps put in the default values in the constructor?

johnwlambert (approved):

gtsam/slam/dataset.h
```diff
@@ -221,7 +221,7 @@ struct SfmTrack {
   SfmTrack(): p(0,0,0) {}
   SfmTrack(const gtsam::Point3& pt) : p(pt) {}
```
johnwlambert, line 222: @varunagrawal @ayushbaid perhaps we should do
```
SfmTrack(r = 0, g = 0, b = 0): p(0,0,0), r(r), g(g), b(b) {}
SfmTrack(const gtsam::Point3& pt, r = 0, g = 0, b = 0) : p(pt), r(r), g(g), b(b) {}
```
so we can use the RGB values later in rendering?

johnwlambert: perhaps `unsigned char` is more appropriate?

varunagrawal: `size_t`

varunagrawal: Actually keeping it float makes sense in case the values aren't in the 0-255 range.

ProfFan: I would say it's probably better to rethink this a bit: what if the image is 12bit per channel? What if the input is grayscale? What are the implications of assuming 8bit RGB?

johnwlambert: totally fair, good points

gtsam/slam/dataset.h
```diff
@@ -218,8 +218,9 @@ typedef std::pair<size_t, size_t> SiftIndex;
 
 /// Define the structure for the 3D points
 struct SfmTrack {
-  SfmTrack(): p(0,0,0) {}
-  SfmTrack(const gtsam::Point3& pt) : p(pt) {}
+  SfmTrack(float r = 0, float g = 0, float b = 0): p(0,0,0), r(r), g(g), b(b) {}
+  SfmTrack(const gtsam::Point3& pt, float r = 0, float g = 0, float b = 0) : p(pt), r(r), g(g), b(b) {}
```
varunagrawal, line 222: Formatting please. Also, a good convention is to have struct/class variables be something like `r_` and then have the initializer list be `r_(r)`. This is consistent with Google C++ style which we follow.

gtsam/slam/dataset.h
```diff
@@ -218,8 +218,9 @@ typedef std::pair<size_t, size_t> SiftIndex;
...
+ 
   Point3 p; ///< 3D position of the point
   float r, g, b; ///< RGB color of the 3D point
```
varunagrawal, line 225: Please add a getter for this?

Something as simple as 
```cpp
Point3 rgb() const { return Point3(r_, g_, b_); }
```

varunagrawal (changes requested):
Some minor things to complete this PR.

gtsam/slam/dataset.h
johnwlambert: Added, thanks

gtsam/slam/dataset.h
johnwlambert: I like the style suggestion, but this would involve renaming all of the variables in the class, since none adhere to that guideline. @dellaert would you like to see a big change like that?

gtsam/slam/dataset.h
johnwlambert: @ProfFan maybe we can discuss this a bit more -- I do think that 8bit is much more common for our use cases in GTSAM and GTSFM, and I also think that if the input is grayscale, they wouldn't use the RGB attributes in the first place. And I think this attribute is purely for 3d point cloud visualization purposes, and I don't think grayscale values would contribute meaningfully to a 3d visualization.

ayushbaid: Agree with John. As this is just used for visualization, we can just stick with float R, G, B values, normalized to [0, 1] range.

gtsam/slam/dataset.h
varunagrawal: I would vote for it since due to the previous lack of getters, it won't be an API change, so we should be able to maintain backwards compatibility.

dellaert: Actually, for *classes* the convention is underscore, for structs it is not (and should not).
It is an API change as these fields are public.

dellaert (approved):
Approving *without* the name change for the fields.

gtsam/slam/dataset.h
johnwlambert: Got it, thanks

varunagrawal: Ah you're right. My mistake.

PR #662, opened 2021-01-11
Serialization unit tests in python for SfmData

johnwlambert:


python/gtsam/tests/test_SfmData.py
```diff
@@ -75,5 +75,12 @@ def test_data(self):
         cam_idx, img_measurement = self.data.track(1).measurement(0)
         self.assertEqual(cam_idx, i1)
 
+        
+    def test_serialize(self):
+        """Test that serialization works for SfmData objects"""
+        self.assertTrue(isinstance(self.data, gtsam.SfmData))
+        serialized_data = self.data.serialize()
+        self.assertTrue(isinstance(serialized_data, str))
```
varunagrawal, line 83: This test should be a whole-round test, where you serialize, then deserialize and verify whether the object is still the same.

johnwlambert: Good point, thanks, I'll update that

dellaert:
Let’s land?

PR #670, opened 2021-01-14
Fix GTSAM_EXPORT

RamadanAhmed:
Fix GTSAM_EXPORT for some classes and function

- Remove GTSAM_EXPORT from template function because its must be implmented in header and so there is nothing to export
- Add GTSAM_EXPORT to Shanon Averaging2/3 to export these classes that are used in examples


varunagrawal (approved):
Awesome! This was on my TODO list, so thanks a ton for the PR.

PR #671, opened 2021-01-15
FunctorizedFactor2

varunagrawal:
This PR adds a binary version of the `FunctorizedFactor`.

Since a lot of operations we perform involve at least 2 keys (e.g. projection of unknown landmarks with unknown cameras), this factor provides a very easy way to add custom functionals involving 2 keys, such as custom sensors for specific projects.

Please refer to the tests for an example of a simple Matrix-Vector product.


dellaert (approved):
Cool

PR #673, opened 2021-01-17
Cleaning SmartStereoProjectionPoseFactor

ToniRV:
- ToDo:
- [x] Split .h/.cpp
- [x] Use const&, avoid needless and costly copies
- [x] Lint everything to 80 char otw it's annoying to read.
- [x] Add asserts where the user could mess up badly (i.e. different sizes for measurements and keys, etc)
- [x] had to add gtsam_unstable to examples... Can I avoid that? Yes, moved the example to gtsam_unstable
- [x] calibration() returns a weird const vector, if const of the K_ elements was the intended purpose, the const should be inside the shared_ptr... otherwise the original const is useless.
- [ ] Add a unique Id (Key) to the smart factor, otw there is a ton of bookkeeping to be done by the user (perhaps in another PR)


ProfFan:
@ToniRV There is a dedicated examples for unstable: 
https://github.com/borglab/gtsam/tree/develop/gtsam_unstable/examples

ToniRV:
@ProfFan I guess the problem then is that [examples/ISAM2_SmartFactorStereo_IMU.cpp](https://github.com/borglab/gtsam/blob/develop/examples/ISAM2_SmartFactorStereo_IMU.cpp)
should be in the `gtsam_unstable/examples` folder instead of `examples`?

dellaert:
I think that’s right. Feel free to move it!
thanks for this PR!

gtsam_unstable/slam/SmartStereoProjectionPoseFactor.h
```diff
@@ -1,6 +1,7 @@
 /* ----------------------------------------------------------------------------
 
- * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * GTSAM Copyright 2010, Georgia Tech Research Corpo
+ * ation,
```
dellaert, line 4: Fix

gtsam_unstable/slam/SmartStereoProjectionPoseFactor.h
```diff
@@ -28,8 +30,10 @@ namespace gtsam {
  * @addtogroup SLAM
  *
  * If you are using the factor, please cite:
- * L. Carlone, Z. Kira, C. Beall, V. Indelman, F. Dellaert, Eliminating conditionally
- * independent sets in factor graphs: a unifying perspective based on smart factors,
+ * L. Carlone, Z. Kira, C. Beall, V. Indelman, F. Dellaert, Eliminating
```
dellaert, line 33: Nicely format after auto format?

dellaert (approved):
I’d move the example to unstable, but other than that, looks great!

gtsam_unstable/slam/SmartStereoProjectionPoseFactor.h
ToniRV: Yes, `clang` 80char limit. Fixed now.

dellaert (approved):
Cool. I'll merge.

PR #674, opened 2021-01-17
CMake scripts compatible with gtsam as git submodule

jlblancoc:
Minor changes to make cmake scripts not to assume `CMAKE_SOURCE_DIR` is identical to `GTSAM_SOURCE_DIR` (and the same for `*_BINARY_DIR`). 

PS: The `GTSAM_` prefix comes from the `project(GTSAM)` in the root `CMakeLists.txt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ include(GtsamTesting)
 include(GtsamPrinting)
 
 # guard against in-source builds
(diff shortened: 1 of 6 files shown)


dellaert (approved):
Trusting you on this...

jlblancoc:
Thanks! 
It was already tested locally, and tests passed so... let's merge it :+1:

PR #675, opened 2021-01-18
changing robust noise model to Gaussian

ayushbaid:
The previous code was converting Gaussian noise model on `Pose3` to robust noise model on `Rot3`. This simple PR scraps off the robust wrapping over the Gaussian noise model.

### DIFF
--- a/gtsam/sfm/ShonanAveraging.cpp
+++ b/gtsam/sfm/ShonanAveraging.cpp
@@ -880,9 +880,7 @@ static BinaryMeasurement<Rot3> convert(
         "parseMeasurements<Rot3> can only convert Pose3 measurements "
         "with Gaussian noise models.");
   const Matrix6 M = gaussian->covariance();
-  auto model = noiseModel::Robust::Create(
-      noiseModel::mEstimator::Huber::Create(1.345),
-      noiseModel::Gaussian::Covariance(M.block<3, 3>(3, 3)));
+  auto model = noiseModel::Gaussian::Covariance(M.block<3, 3>(3, 3));
   return BinaryMeasurement<Rot3>(f->key1(), f->key2(), f->measured().rotation(),
                                  model);
 }


varunagrawal (changes requested):
Sorry, we can't undo that since that was part of the PR to add robust noise models to Shonan.

varunagrawal (approved):
Nvm the last review. The flow of the code indicates this is correct. Nice job @ayushbaid!

johnwlambert:
Thanks for the PR @ayushbaid.

@dellaert and @lucacarlone -- do you mind taking a look since you all edited Shonan most recently? It seems like ShonanAveraging3 now automatically forces the robust option. Should there be an if/else statement to switch off between 
```
auto model = noiseModel::Robust::Create(
noiseModel::mEstimator::Huber::Create(1.345),
noiseModel::Gaussian::Covariance(M.block<3, 3>(3, 3)));
```
vs.
```
auto model = noiseModel::Gaussian::Covariance(M.block<3, 3>(3, 3), true)
```
Most recent change is here:
https://github.com/borglab/gtsam/pull/547/files#diff-2681e2ffa3c440f21039b01556f84cf32000a6c0c251989f657835ee69622367L850

varunagrawal:
That flag is not necessary since `convert` is called by `extractRot3Measurements` which in turn is called by `maybeRobust`. `maybeRobust` takes the flag for using Huber loss and does the conversion accordingly.

Of course, a 2nd look is always greatly appreciated.

ayushbaid:
> That flag is not necessary since `convert` is called by `extractRot3Measurements` which in turn is called by `maybeRobust`. `maybeRobust` takes the flag for using Huber loss and does the conversion accordingly.
> 
> Of course, a 2nd look is always greatly appreciated.

Yup. I think the conversion will happen in the `maybeRobust` function depending on the parameters.

dellaert (approved):
LGTM, seems we were double robustifying before?

ayushbaid:
> LGTM, seems we were double robustifying before?

Yep! And the first one was without checking the parameters

PR #676, opened 2021-01-19
additional `sparseJacobian` functions (new)

gchenfc:
This replaces #610 to reduce commit- and diff- noise.

The purpose of this PR is to provide a convenience function for obtaining the sparse (augmented) Jacobian of a `GaussianFactorGraph`.

Note that, as it stands, 99% of this PR is copy-pasta from `GaussianFactorGraph::sparseJacobian()`.  Although I would prefer to make a version that is more elegant, this is the original way Mandy/Fan's code was written and I guess we can do that in a future PR.  I added a TODO.


gchenfc:
Sorry, I used a "merge" instead of "cherry-pick" because I thought it would better preserve history, but now it is way too messy.  Closing and trying again.

PR #677, opened 2021-01-19
additional `sparseJacobian` functions (new)

gchenfc:
This replaces #610 to reduce commit- and diff- noise.

The purpose of this PR is to provide a convenience function for obtaining the sparse (augmented) Jacobian of a GaussianFactorGraph.

Note that, as it stands, 99% of this PR is copy-pasta from GaussianFactorGraph::sparseJacobian(). Although I would prefer to make a version that is more elegant, this is the original way Mandy/Fan's code was written and I guess we can do that in a future PR. I added a TODO.


gtsam/linear/SparseEigen.h
```diff
@@ -0,0 +1,119 @@
...
+#include <Eigen/Sparse>
+
+namespace gtsam {
+
+typedef Eigen::SparseMatrix<double> SpMat;
+
+/// Constructs an Eigen-format SparseMatrix of a GaussianFactorGraph
+SpMat sparseJacobianEigen(
+    const GaussianFactorGraph &gfg, const Ordering &ordering) {
+  // TODO(gerry): eliminate copy/pasta by making GaussianFactorGraph version
```
dellaert, line 38: I thought you had done that already? Or is that "the next PR" ?

gtsam/linear/SparseEigen.h
```diff
@@ -0,0 +1,119 @@
+/* ----------------------------------------------------------------------------
```
dellaert, line 1: naming convention for non-class files starts with lowercase, i.e. sparseEigen.h. But, an even better suggetion is to rename SpMat below (not a good name) to SparseEigen. Then we can keep the filename and this file is about how to create one.

gtsam/linear/tests/testGaussianFactorGraph.cpp
```diff
@@ -73,37 +82,48 @@ TEST(GaussianFactorGraph, sparseJacobian) {
...
-
-  EXPECT(assert_equal(expected, actual));
+  const Key x123 = 0, x45 = 1;
+  gfg.add(x123, (Matrix(2, 3) << 1, 2, 3, 5, 6, 7).finished(),
+          Vector2(4, 8), model);
+  gfg.add(x123, (Matrix(2, 3) << 9, 10, 0, 0, 0, 0).finished(),
+          x45,  (Matrix(2, 2) << 11, 12, 14, 15.).finished(),
+          Vector2(13, 16), model);
+
+  // Check version for MATLAB - NOTE that we transpose this!
```
dellaert, line 94: Diff is weird. Can you move back up to minimize diff?

dellaert (approved):
Bit confused that your more general code is not here but I guess that will be next PR..

gtsam/linear/SparseEigen.h
gchenfc: Ya, I was under the impression that you wanted just the bare minimum thing needed for A->B, which is efficient code for returning an Eigen-sparse-matrix format jacobian?

gchenfc: But I'm more than happy to put that in this PR

gtsam/linear/SparseEigen.h
gchenfc: done

gtsam/linear/tests/testGaussianFactorGraph.cpp
gchenfc: done

gtsam/linear/SparseEigen.h
dellaert: No, my comment on the other PR was not that these more generic, copy/pasta avoiding functions were not useful, just that they should be inside gfg.h, and then *used* in a small header-only function for the Eigen version.

dellaert: So, I propose you merge this, then create a next PR with those utilities to get rid of copy/pasta. No sense in throwing away that work.

gchenfc: ok!

PR #678, opened 2021-01-19
Better printing

varunagrawal:
This PR adds some printing updates and fixes which makes debugging easier.


gtsam/base/Matrix.cpp
```diff
@@ -153,7 +153,7 @@ const Eigen::IOFormat& matlabFormat() {
 /* ************************************************************************* */
 //3 argument call
 void print(const Matrix& A, const string &s, ostream& stream) {
-  cout << s << A.format(matlabFormat()) << endl;
+  stream << s << A.format(matlabFormat()) << endl;
```
ProfFan, line 156: oh this is a typo! nice catch :)

ProfFan (approved):

PR #681, opened 2021-01-20
Known Issues Section

### LINKED ISSUE #590: testSmartProjectionFactor unit test fails with march=native
## Description

`testSmartProjectionFactor` fails on Ubuntu 16.04, gcc 5.4.0 with default configuration. The default configuration uses `march=native`, which is not exercised by CI. Test passes with `GTSAM_BUILD_WITH_MARCH_NATIVE=OFF`.

```
126/243 Test #126: testSmartProjectionFactor ..............***Failed    0.02 sec
/home/cbeall/git/gtsam/gtsam/slam/tests/testSmartProjectionFactor.cpp:492: Failure: "Exception: 
Indeterminant linear system detected while working near variable
1 (Symbol: 1).

Thrown when a linear system is ill-posed.  The most common cause for this
error is having underconstrained variables.  Mathematically, the system is
underdetermined.  See the GTSAM Doxygen documentation at
http://borg.cc.gatech.edu/ on gtsam::IndeterminantLinearSystemException for
more information." 
There were 1 failures
```

I'll try to look into this myself later, but wanted to make a report.

## Steps to reproduce

Get latest develop, configure, `make check`.

## Expected behavior

Test passes.

## Environment

Ubuntu 16.04, gcc 5.4.0

varunagrawal:
Added a new section in `INSTALL.md` for known issues which discusses conditions for the issue in #590.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -13,7 +13,7 @@ $ make install
 ## Important Installation Notes
 
 1. GTSAM requires the following libraries to be installed on your system:
-    - BOOST version 1.58 or greater (install through Linux repositories or MacPorts)
+    - BOOST version 1.67 or greater (install through Linux repositories or MacPorts)
     - Cmake version 3.0 or higher
     - Support for XCode 4.3 command line tools on Mac requires CMake 2.8.8 or higher
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -41,11 +41,6 @@ $ make install
     - MacOS 10.6 - 10.14
     - Windows 7, 8, 8.1, 10
 
-    Known issues:
-
-    - MSVC 2013 is not yet supported because it cannot build the serialization module
-      of Boost 1.55 (or earlier).
-
 2. GTSAM makes extensive use of debug assertions, and we highly recommend you work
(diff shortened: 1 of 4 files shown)


chrisbeall:
I tried on Ubuntu 16.04 with Boost 1.68, compiled locally with gcc 5.4 (not from PPA), and the problem persists. It appears that the compiler matters, too.

I think it's better to document this issue in the readme or install notes under "Known Issues" than to force users to upgrade Boost because Ubuntu 16.04 with Boost 1.58 is still widely used. A CMake warning might be ok, too.

varunagrawal:
@chrisbeall interesting. What would be the minimum supported compiler for GCC? We use GCC 5.5.0 in the CI and but turn off march_native.

jlblancoc (commented):
Not ideal, but this is probably a good deal: reporting the known issues which we have not been able to fix or would require users to upgrade to fix a bug in a part (serialization) that they may not need at all for their projects. 

I defer the PR approval to Frank or other core developers, since this is more a "policy" than a technical discussion.

varunagrawal:
@jlblancoc I updated the PR to only have the known issues section. Does it look good to you?

dellaert (approved):

PR #682, opened 2021-01-20
Improve `sparseJacobianEigen()`

gchenfc:
#677 contained copy/pasted code from `GaussianFactorGraph::sparseJacobian` in `sparseJacobianEigen`.  This PR makes the more generic, `GaussianFactorGraph::sparseJacobianInPlace<T>(T& entries, ...)` and uses it instead in `sparseJacobianEigen`.

It also adds an optional Ordering argument to better match the other jacobian functions, and optional nrows/ncols parameters since these are necessary to keep track of for sparse matrices.

Finally, efficiency is *slightly* improved by inserting directly as described [here](https://eigen.tuxfamily.org/dox/group__TutorialSparse.html#title3), benchmarked by running `sparseJacobianEigen` on a linearized `BALfinal-394-100368-pre`:
| Description | Time |
| --- | --- |
| `T` as `vector<Eigen::Triple>` and using `setFromTriplets` (original) | 0.67s |
| `T` as `vector<boost::tuple>` and using `SparseEigen::insert` | 0.72s |
| `T` as `vector<Eigen::Triple>` and using `SparseEigen::insert` (new) | 0.63s |

Notes about the diff:
* GaussianFactorGraph.cpp <-> GaussianFactorGraph-impl.h, the function implementation moved from `sparseJacobian` to `sparseJacobianInPlace` with the following modifications:
  * (in-place) `entries` is no longer constructed
  * `col`/`row` variables were renamed to `nrows` and `ncols` to match parameter names
  * entries in the `b` vector are only added if they are non-zero
* in `testGaussianFactorGraph` one entry A(5, 7) = 0 was removed since it has value 0 (see previous bullet point)


gtsam/base/Matrix.h
```diff
@@ -73,6 +75,12 @@ GTSAM_MAKE_MATRIX_DEFS(9);
 typedef Eigen::Block<Matrix> SubMatrix;
 typedef Eigen::Block<const Matrix> ConstSubMatrix;
 
+/// Sparse matrix representation as vector of tuples.
+/// See SparseMatrix.h for additional representations SparseMatrixEigenTriplets
+/// and SparseMatrixEigen
+typedef std::vector<boost::tuple<size_t, size_t, double>>
```
dellaert, line 81: Please move to gfg. Ideally: no change to this file

gtsam/linear/GaussianFactorGraph-impl.h
```diff
@@ -0,0 +1,100 @@
+/* ----------------------------------------------------------------------------
```
dellaert, line 1: Why templated?? If currently we only use double, undo the template and just stick it into gfg.cpp, please. We can always template *when* we need it.

dellaert (commented):
Two comments before doing a full review.

gtsam/linear/GaussianFactorGraph-impl.h
gchenfc: Yes, I will fix this.

The trouble was that SparseEigen needed to call this function with template parameter `vector<Eigen::Triplet>` which is defined in `<Eigen/Sparse>`, so defining it in `gfg.cpp` would require either including `<Eigen/Sparse>` or forward declaring something that the caller must include.  I thought for a long time about how avoid templating and previously, I couldn't think of a way to avoid it.

*However*, once I made the "efficiency improvements" then I found 2 ways around it:
1) Just define our own version of Triplet that's more efficient than boost::tuple<size_t, size_t, double>
2) Use boost::tuple and be slightly less efficient (as per the table, about 12% slower).

I will do the former.

gtsam/linear/GaussianFactorGraph.h
```diff
@@ -180,20 +180,84 @@ namespace gtsam {
     ///@name Linear Algebra
     ///@{
 
+    /// Sparse matrix representation as vector of tuples.
+    typedef std::vector<boost::tuple<size_t, size_t, double>>
+        SparseMatrixBoostTriplets;
+    /// Sparse matrix representation as vector of slightly more efficient
+    /// tuples.
+    typedef std::vector<std::tuple<int, int, double>> SparseMatrixGtsamTriplets;
```
dellaert, line 188: What's the reasoning here? I'd prefer to keep only one of these.

gtsam/linear/GaussianFactorGraph.h
```diff
@@ -180,20 +180,84 @@ namespace gtsam {
...
+    SparseMatrixBoostTriplets sparseJacobian(
+        const Ordering& ordering) const;
+
+    /** Returns a sparse augmented Jacobian with default Ordering */
+    SparseMatrixBoostTriplets sparseJacobian(size_t& nrows,
+                                             size_t& ncols) const;
+
+    /** Returns a sparse augmented Jacobian without with default ordering and
+     * outputting its dimensions */
+    SparseMatrixBoostTriplets sparseJacobian() const;
```
dellaert, line 214: Why do we all of a sudden have that many versions? Which ones do we need *now* ? Which are built "for the future"?

gtsam/linear/GaussianFactorGraph.h
```diff
@@ -180,20 +180,84 @@ namespace gtsam {
...
+    Matrix sparseJacobian_(const Ordering& ordering) const;
+
+    /** Returns a matrix-form sparse augmented Jacobian with default Ordering
+     * @param[out] nrows The number of rows in the Jacobian
+     * @param[out] ncols The number of columns in the Jacobian
+     */
+    Matrix sparseJacobian_(size_t& nrows, size_t& ncols) const;
+
+    /** Returns a matrix-form sparse augmented Jacobian with default ordering
+     * and without outputting its dimensions */
```
dellaert, line 237: Same comment.

dellaert (changes requested):
Gerry, I appreciated the enthusiasm, but it seems you created a (confusing) zoo of options for some hypothetical user in the future. Let's not: we need an opinionated API that provides just what is needed. Try to remember why this PR exists, and satisfy *that* use case. Leave everything else untouched. I think we can change to your "fast type" without changing the API (substantially).

gtsam/linear/GaussianFactorGraph.h
gchenfc: Right now, only need to add the one with all the arguments:
`SparseMatrixBoostTriplets sparseJacobian(const Ordering& ordering,
                                             size_t& nrows,
                                             size_t& ncols) const;`

I'll remove the others.

I should note though that I'm not sure how people are "supposed" to use the version without the `nrows` and `ncols` arguments since otherwise you cannot tell how large the real matrix is (there may be zero columns/zero rows at the end that you would never know about).  But I guess we'll leave that up to users to figure out.

gtsam/linear/GaussianFactorGraph.h
gchenfc: Originally the code used `SparseMatrixBoostTriplets`.  I added `SparseMatrixGtsamTriplets` because it's about 14% faster and 33% less memory intensive (16 bytes instead of 24 bytes per nnz entry) in my tests.  Reason: 1) `size_t` is 8 bytes, and even if we use int instead, `boost::tuple` aligns to 8 bytes.  We don't have to typedef these?

Old version kept for backwards compatibility.

gtsam/linear/GaussianFactorGraph.h
gchenfc: done

gtsam/linear/GaussianFactorGraph.h
gchenfc: done

gtsam/linear/GaussianFactorGraph.h
gchenfc: I removed the typedefs but both are still used (the std::tuple version is used in the "fast" version).

gtsam/linear/GaussianFactorGraph.cpp
```diff
@@ -138,56 +147,78 @@ namespace gtsam {
...
+      nrows += jacobianFactor->rows();
     }
-    return vector<triplet>(entries.begin(), entries.end());
+
+    ncols++;  // +1 for b-column
   }
 
   /* ************************************************************************* */
-  Matrix GaussianFactorGraph::sparseJacobian_() const {
+  using BoostTriplets = std::vector<boost::tuple<size_t, size_t, double> >;
```
dellaert, line 185: Why ??? Just make everything fast unless there is a compelling reason *not* to

gtsam/linear/SparseEigen.h
```diff
@@ -30,89 +30,33 @@
...
-    }
-
-    // Increment row index
-    row += jacobianFactor->rows();
-  }
-
-  // ...and make a sparse matrix with it.
-  SparseEigen Ab(row, currentColIndex + 1);
-  Ab.setFromTriplets(entries.begin(), entries.end());
+  gttic_(SparseEigen_sparseJacobianEigen);
```
dellaert, line 40: Great copy/pasta removal!

gtsam/linear/GaussianFactorGraph.h
```diff
@@ -181,19 +181,31 @@ namespace gtsam {
...
      */
     Matrix sparseJacobian_() const;
 
+    /** Returns a sparse matrix with `int` indices instead of `size_t` for
+     * slightly faster performance
+     * @param ordering the column ordering
+     * @param[out] nrows The number of rows in the Jacobian
+     * @param[out] ncols The number of columns in the Jacobian
+     */
+    std::vector<std::tuple<int, int, double> > sparseJacobianFast(
```
dellaert, line 206: Why not just make sparseJacobian fast ?

dellaert (commented):
Still confused, see comments. Just switch to int already *unless* there is a great reason NOT to.

gtsam/linear/GaussianFactorGraph.h
gchenfc: I was concerned about backwards compatibility, but if you think this isn't a big issue, then I would gladly make everything "fast".

API Changes:
1) If people don't use `auto` and use the actual type in declarations, e.g. `std::vector<boost::tuple<size_t, size_t, double>> jacobian = sparseJacobian()`, then obviously this will need to change
2) `boost::tuple` and `std::tuple` actually have different access methods, where `boost::tuple` uses member function `get<N>()` while `std::tuple` uses a free function `std::get<N>(tuple)` so even code that uses `auto` will break.

`boost::tuple<int, int, double>` is no faster than `boost::tuple<size_t, size_t, double>` on GCC 7-10 that I tested on Ubuntu 18.04 since it's 8 byte aligned, whereas `std::tuple` can align in 4 bytes.

gtsam/linear/SparseEigen.h
gchenfc: :)

gtsam/linear/GaussianFactorGraph.cpp
gchenfc: (see other comment thread)

gtsam/linear/GaussianFactorGraph.h
dellaert: OK, now I'm confused. What *is* faster then?

ProfFan: A custom tuple type will definitely be much faster (and deterministic)

gchenfc: `std::tuple` is faster (`boost::tuple` is the slow one)

A custom struct was the same speed as `std::tuple`, but perhaps it might vary by compiler/architecture

dellaert: I see. Well, then, let's use std::tuple and size_t and be done with it :-) No fast or slow versions.

gchenfc: Sorry, last clarification: `std::tuple<int, int, double>` is faster than `std::tuple<size_t, size_t, double>`.

| fast | slow |
|--|--|
|`std::tuple<int...>` | `std::tuple<size_t,...>`|
| custom struct | `boost::tuple<size_t, ...>`|
|  | `boost::tuple<int, ...>` |

where both "fast" options have the same speed and all 3 "slow" options have the same speed.

I will assume `std::tuple` and `int` is ok with you.  (Eigen sparse doesn't support unsigned indexing for some reason, and I don't think >2 billion entries aka 6GB worth will be needed)

dellaert: OK

gchenfc: done

gtsam/linear/GaussianFactorGraph.cpp
```diff
@@ -100,33 +100,35 @@ namespace gtsam {
...
+      ncols += dims[key];
     }
 
     // Iterate over all factors, adding sparse scalar entries
-    typedef boost::tuple<size_t, size_t, double> triplet;
-    vector<triplet> entries;
-    size_t row = 0;
-    for (const sharedFactor& factor : *this) {
+    SparseTriplets entries;
+    entries.reserve(60 * size());
```
dellaert, line 129: Why 60? poseSLAM=6+6=12, vanilla SFM = 6+3=9, IMU = 6+6+3+3+6=24

gtsam/linear/GaussianFactorGraph.cpp
```diff
@@ -138,52 +140,61 @@ namespace gtsam {
...
+              "GaussianFactorGraph contains a factor that is neither a "
+              "JacobianFactor nor a HessianFactor.");
       }
 
       // Whiten the factor and add entries for it
       // iterate over all variables in the factor
       const JacobianFactor whitened(jacobianFactor->whiten());
       for (JacobianFactor::const_iterator key = whitened.begin();
-          key < whitened.end(); ++key) {
+           key < whitened.end(); ++key) {
```
dellaert, line 152: auto?

dellaert (approved):
Yay!!!
Just two nits... I'll leave you to merge after that and after CI succeeds. Email me if you can't for some reason.

gtsam/linear/GaussianFactorGraph.cpp
gchenfc: done

gtsam/linear/GaussianFactorGraph.cpp
gchenfc: Good question - that was from Fan/Mandy's original code, but good point.  I think 30 is more appropriate than 60.

I just tried with a couple examples in the BAL dataset and they're 26, StereoVOExample_large is 27, (very small) g2o example was 33, and Pose2SLAMStressTest was 15.

gchenfc: done

dellaert:
Nice!

PR #684, opened 2021-01-23
changed barcsq to a vector to allow each factor to have a different inlier threshold

lucacarlone:
I realized that having a default inlier threshold equal to 1 does not make sense for the gtsam user since:
1) "1" is not a reasonable error threshold (for Gaussian noise, we should use the chi2inv to establish a reasonable maximum error) and 
2) the factors can have different dimensions, hence it is desirable to be able to set a different threshold for each factor.

The PR involves:

- [x] making barcsq into a vector
- [x] adding/adjusting the unit tests accordingly
- [x] initializing barcsq using chi2inv

### DIFF
--- a/gtsam/nonlinear/GncOptimizer.h
+++ b/gtsam/nonlinear/GncOptimizer.h
@@ -29,7 +29,18 @@
 #include <gtsam/nonlinear/GncParams.h>
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
 
+//#include <boost/math/distributions/inverse_chi_squared.hpp>
+#include <boost/math/distributions/chi_squared.hpp>
+
 namespace gtsam {
+/*
+ * Quantile of chi-squared distribution with given degrees of freedom at probability alpha.
+ * Equivalent to chi2inv in Matlab.
+ */
+static double Chi2inv(const double alpha, const size_t dofs) {
+  boost::math::chi_squared_distribution<double> chi2(dofs);
+  return boost::math::quantile(chi2, alpha);
+}
 
 /* ************************************************************************* */
 template<class GncParameters>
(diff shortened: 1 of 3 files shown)


lucacarlone:
@dellaert @jingnanshi @pantonante : I completed a small enhancement for GNC. please take a look and approve if you like it.

@dellaert : please take a look at the 2 lines commented at line 658 of testGncOptimizer. It seems that "dim()" segfaults when we use the default noise model in the factor. This might be a bug in gtsam.

dellaert (approved):
Nice! Sorry about delay, teaching large class ;-)

dellaert (changes requested):
PR is fine, but changing to "request changes" as I just noticed CI fails on Linux. Probably something small but needs to be fixed before we can merge.

dellaert:
> @dellaert : please take a look at the 2 lines commented at line 658 of testGncOptimizer. It seems that "dim()" segfaults when we use the default noise model in the factor. This might be a bug in gtsam.

Is that why CI fails? Let me check.

lucacarlone:
@dellaert: let me know if you get some insight into what's failing in CI. I took another pass and cleaned up the comments. @jingnanshi  @pantonante : feel free to approve if this looks good.

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -29,7 +29,18 @@
 #include <gtsam/nonlinear/GncParams.h>
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
 
+//#include <boost/math/distributions/inverse_chi_squared.hpp>
```
jingnanshi, line 32: We might want to remove this line.

jingnanshi (approved):
Everything looks good! I left one comment regarding a commented out include file.

gtsam/nonlinear/GncOptimizer.h
lucacarlone: good catch - done!

dellaert:
Still some failures on Ubuntu, though?

lucacarlone:
@dellaert : I only changed comments so far, I hoped you could look into the dim() issue. Let me know if that's not feasible.

dellaert:
> @dellaert : I only changed comments so far, I hoped you could look into the dim() issue. Let me know if that's not feasible.

Dim issue is separate, I think. CI failure is unrelated

lucacarlone:
@dellaert : @jingnanshi fixed the issue! at this point, this branch should be good to go

dellaert (approved):

PR #690, opened 2021-02-03
Fix/shonan averaging dense to sparse

ToniRV:


ToniRV:
Nice, it worked! @dellaert

dellaert (approved):
Cool, thanks !!

PR #691, opened 2021-02-05
Avoid target collision if gtsam used as submodule

jlblancoc:
CMake ends with an error if gtsam is used as a git submodule in another project also having an `uninstall()` target. 

This PR simply defines another target and makes the top-level one to depend on it, or fallback to the usual way if no other top-level `uninstall` exists.


varunagrawal (approved):
LGTM

Could you also add in a comment specifying the submodule use-case?

PR #693, opened 2021-02-07
Update LieGroups.lyx

melhashash:
Fix typo in the equation in item 1 section 5.3


mikesheffler (commented):
Pretty sure this is correct. I don't think the corresponding .pdf is automatically generated (note that it is present in the source tree), so I don't want to approve it yet.

Should @melhashash generate a .pdf, or does someone in particular create it?

dellaert:
I think this is right. I'll approve but if you can render  the PDF that would be better. I'll wait to merge until I heard from you.

dellaert (approved):

melhashash:
I have generated the pdf. 

Note: the font has changed, and references are added at the end (it was missing in the original document).

@dellaert @mikesheffler

dellaert:
Awesome, thanks !

PR #702, opened 2021-02-24
add Sim(3)-based alignment to the wrapper

johnwlambert:
- Fix pose notation in Sim(3) Align() code and unit tests
- Add more docstrings
- Move Sim(3) out of unstable to stable GTSAM
- Add Sim(3) to python wrapper
- Add python unit tests

3rd python unit test checks alignment here between 4 pose-pairs:
<p align="center">
  
</p>


ayushbaid:
Hi John. Is this algorithm the same as what we are trying to implement in GTSFM?

johnwlambert:
> Hi John. Is this algorithm the same as what we are trying to implement in GTSFM?

Yes, although PosePairs (wTi, wTi_) from wTi_list, wTi_list_ is the one that was suggested: https://github.com/borglab/gtsam/blob/develop/gtsam_unstable/geometry/Similarity3.cpp#L170

gtsam/geometry/Similarity3.cpp
```diff
@@ -24,7 +24,7 @@
 namespace gtsam {
 
 using std::vector;
-using PointPairs = vector<Point3Pair>;
+typedef vector<Point3Pair> PointPairs;
```
akshay-krishnan, line 27: Why not Point3Pairs? I understand its not related to your change, but I dont know why we dropped the 3 there.

akshay-krishnan (changes requested):
I think we should also add some unit tests for Similarity3 if there are not any already, especially since its being taken out of unstable.

gtsam/geometry/Similarity3.cpp
johnwlambert: I agree that would be better. Yeah just legacy code, we could improve it likely

johnwlambert:
> I think we should also add some unit tests for Similarity3 if there are not any already, especially since its being taken out of unstable.

Agreed, yeah we should do that

akshay-krishnan:
It looks like there is a testSimilarity3 in unstable which broke because we moved the file. Maybe we can try moving that as well, and that might fix the CI?

johnwlambert:
> It looks like there is a [testSimilarity3](https://github.com/borglab/gtsam/blob/develop/gtsam_unstable/geometry/tests/testSimilarity3.cpp) in unstable which broke because we moved the file. Maybe we can try moving that as well, and that might fix the CI?

Good point. But also seems like `PointPairs` isn't making its way into the gtsam namespace, also:
```python
/home/runner/work/gtsam/gtsam/build/python/gtsam.cpp: In function ‘void pybind11_init_gtsam(pybind11::module&)’:
/home/runner/work/gtsam/gtsam/build/python/gtsam.cpp:1163:45: error: ‘PointPairs’ in namespace ‘gtsam’ does not name a type
         .def_static("Align",[](const gtsam::PointPairs& abPointPairs){return gtsam::Similarity3::Align(abPointPairs);}, py::arg("abPointPairs"));
                                             ^
/home/runner/work/gtsam/gtsam/build/python/gtsam.cpp:1163:129: error: expected ‘)’ before string constant
         .def_static("Align",[](const gtsam::PointPairs& abPointPairs){return gtsam::Similarity3::Align(abPointPairs);}, py::arg("abPointPairs"));
```

gtsam/geometry/Similarity3.cpp
ProfFan: Move this typedef to the header

johnwlambert: thanks @ProfFan , just moved it to the `Similarity3.h` file

gtsam/geometry/Similarity3.h
```diff
@@ -30,6 +30,8 @@ namespace gtsam {
 // Forward declarations
 class Pose3;
 
+typedef std::vector<Point3Pair> PointPairs;
```
varunagrawal, line 33: Keep this as `using Point3Pairs= vector<Point3Pair>`. C++11 compliant and preferred (as it was previously). :slightly_smiling_face:

varunagrawal (commented):
Looks good, but please rename `PointPairs` to `Point3Pairs` and use the preferred `using` idiom for typedefs. Also @johnwlambert be sure to link your commits to your account.

gtsam/geometry/Similarity3.h
johnwlambert: thanks for the review @varunagrawal. I'll make the change -- is there a specific reason `using` is preferred over `typedef`?

varunagrawal: https://stackoverflow.com/a/23196675/1236990

asa:
Using is more “modern”.

gtsam/geometry/Similarity3.h
dellaert: If Pose3Pairs is defined in Pose3.h, Point3Pairs (indeed, rename it?) should be defined in Point3.h

gtsam/geometry/Similarity3.cpp
johnwlambert: should be updated now to `Point3Pairs`

akshay-krishnan (changes requested):

gtsam/geometry/Similarity3.h
```diff
@@ -125,7 +125,10 @@ class Similarity3: public LieGroup<Similarity3, 7> {
   GTSAM_UNSTABLE_EXPORT static Similarity3 Align(const std::vector<Point3Pair>& abPointPairs);
   
   /**
-   *  Create Similarity3 by aligning at least two pose pairs
+   * Create Similarity3 by aligning at least two pose pairs
+   * Given a list of pairs in world frame w1, and a list of pairs in another world
+   * frame w2, will compute the best-fit Similarity3 transformation to align them.
+   * `w2Sw1` will returned for pairs [ (w2Ti1,w1Ti1), (w2Ti2,w1Ti2), (w2Ti3,w1Ti3) ]
```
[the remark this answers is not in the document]
johnwlambert: If they're in the same world frame, then I think there is no need for alignment? So I think the prior documentation is confusing @akshay-krishnan

gtsam/geometry/Similarity3.cpp
```diff
@@ -174,18 +174,19 @@ Similarity3 Similarity3::Align(const vector<Pose3Pair> &abPosePairs) {
...
   vector<Rot3> rotations;
-  PointPairs abPointPairs;
+  Point3Pairs abPointPairs;
   rotations.reserve(n);
   abPointPairs.reserve(n);
   Pose3 wTa, wTb;
   for (const Pose3Pair &abPair : abPosePairs) {
     std::tie(wTa, wTb) = abPair;
-    rotations.emplace_back(wTa.rotation().compose(wTb.rotation().inverse()));
+    Rot3 aRb = wTa.rotation().between(wTb.rotation());
```
dellaert, line 183: const

gtsam/geometry/Similarity3.h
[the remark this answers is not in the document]
dellaert: I agree that this comment is misleading. there is a set A and a set B, both specified wrpt to the *same* world frame, and we want to get aSb

gtsam/geometry/tests/testSimilarity3.cpp
```diff
@@ -346,7 +346,7 @@ TEST(Similarity3, AlignPoint3_3) {
 //******************************************************************************
 // Align with Pose3 Pairs
 TEST(Similarity3, AlignPose3) {
-  Similarity3 expected_aSb(Rot3::Ry(180 * degree), Point3(2, 3, 5), 2.0);
+  Similarity3 expected_bSa(Rot3::Ry(180 * degree), Point3(2, 3, 5), 2.0);
```
dellaert, line 349: we talk about aSb all the time and that's what we should test.

gtsam/geometry/tests/testSimilarity3.cpp
```diff
@@ -364,8 +364,8 @@ TEST(Similarity3, AlignPose3) {
   // Cayley transform cannot accommodate 180 degree rotations,
   // hence we only test for Expmap
 #ifdef GTSAM_ROT3_EXPMAP
-  Similarity3 actual_aSb = Similarity3::Align(correspondences);
-  EXPECT(assert_equal(expected_aSb, actual_aSb));
+  Similarity3 actual_bSa = Similarity3::Align(correspondences);
```
dellaert, line 367: ab_correspondences should give aSb

dellaert (changes requested):
Awesome! But several red flags on coordinate conventions

gtsam/geometry/Similarity3.h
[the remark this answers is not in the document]
johnwlambert: Thanks for the review @dellaert . Question: in the Similarity3 `transformFrom` unit test, the code states https://github.com/borglab/gtsam/blob/develop/gtsam_unstable/geometry/tests/testSimilarity3.cpp#L261:

```c++
// Group action on Pose3
TEST(Similarity3, GroupActionPose3) {
  Similarity3 bSa(Rot3::Ry(180 * degree), Point3(2, 3, 5), 2.0);
  // Create source poses
  Pose3 Ta1(Rot3(), Point3(0, 0, 0));
  Pose3 Ta2(Rot3(-1, 0, 0, 0, -1, 0, 0, 0, 1), Point3(4, 0, 0));
  // Create destination poses
  Pose3 expected_Tb1(Rot3(-1, 0, 0, 0, 1, 0, 0, 0, -1), Point3(4, 6, 10));
  Pose3 expected_Tb2(Rot3(1, 0, 0, 0, -1, 0, 0, 0, -1), Point3(-4, 6, 10));
  EXPECT(assert_equal(expected_Tb1, bSa.transformFrom(Ta1)));
  EXPECT(assert_equal(expected_Tb2, bSa.transformFrom(Ta2)));
}
```
By `Ta1` I assume that we mean `wTa1`. But that means the semantics state:
```
expected_wTb1 = bSa.transformFrom(wTa1)
```
which seems contrary to our GTSAM convention.

gtsam/geometry/Similarity3.h
johnwlambert: Updated

gtsam/geometry/Similarity3.cpp
```diff
@@ -174,18 +174,19 @@ Similarity3 Similarity3::Align(const vector<Pose3Pair> &abPosePairs) {
 
   // calculate rotation
   vector<Rot3> rotations;
-  PointPairs abPointPairs;
+  Point3Pairs abPointPairs;
   rotations.reserve(n);
   abPointPairs.reserve(n);
-  Pose3 wTa, wTb;
+  Pose3 aTi, bTi;
```
varunagrawal, line 180: Please add a comment on what frame `i` is?

gtsam/geometry/tests/testSimilarity3.cpp
```diff
@@ -260,18 +260,22 @@ TEST(Similarity3, GroupAction) {
...
+  Pose3 eTo2(Rot3(-1, 0, 0, 0, -1, 0, 0, 0, 1), Point3(4, 0, 0));
+
+  // Create destination poses (two objects now living in the world/city "w" frame)
+  // Desired to place the objects into the world frame for tracking
+  Pose3 expected_wTo1(Rot3(-1, 0, 0, 0, 1, 0, 0, 0, -1), Point3(4, 6, 10));
+  Pose3 expected_wTo2(Rot3(1, 0, 0, 0, -1, 0, 0, 0, -1), Point3(-4, 6, 10));
+
+  // objects now live in the world frame, instead of in the egovehicle frame
+  EXPECT(assert_equal(expected_wTo1, wSe.transformFrom(eTo1)));
+  EXPECT(assert_equal(expected_wTo2, wSe.transformFrom(eTo2)));
```
varunagrawal, line 278: Love the comments!

gtsam/gtsam.i
```diff
@@ -448,6 +448,16 @@ class Point3 {
   void serialize() const;
 };
 
+#include <gtsam/geometry/Point3.h>
+class Point3Pairs
+{
```
varunagrawal, line 453: Formatting? We use the Google/1TBS style rather than the K&R/Linux kernel style.

gtsam/gtsam.i
```diff
@@ -766,6 +776,16 @@ class Pose3 {
   void serialize() const;
 };
 
+#include <gtsam/geometry/Pose3.h>
+class Pose3Pairs
+{
```
varunagrawal, line 781: Formatting here as well.

varunagrawal (approved):
LGTM with some minor changes. :slightly_smiling_face:

gtsam/geometry/Similarity3.h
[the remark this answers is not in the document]
johnwlambert: This was resolved, notation was backwards/non-standard previously in the unit tests

gtsam/geometry/tests/testSimilarity3.cpp
johnwlambert: thanks!

gtsam/geometry/Similarity3.cpp
johnwlambert: done, thanks

gtsam/gtsam.i
johnwlambert: done, thanks. I moved the `{` brace on the same line as `class Point3Pairs` now

gtsam/gtsam.i
johnwlambert: done

gtsam/geometry/tests/testSimilarity3.cpp
johnwlambert: updated

gtsam/geometry/Similarity3.cpp
johnwlambert: updated

dellaert:
Hmmmm, CI fails, but it seems with a compiler error, and only on Linux :-(

dellaert:
(Other than that LGTM, but that needs to be investigated before we can merge)

johnwlambert:
> (Other than that LGTM, but that needs to be investigated before we can merge)

Yeah, @ProfFan thought it was likely OOM issue from the large gtsam.i file?

dellaert:
Definitely possible. @jingwuOUO reported similar issues on her Linux box. @ProfFan and @varunagrawal what can we do to split up gtsam.i? One file per subdir?

varunagrawal:
I don't think we even need that. We need to figure out a way to break down the `gtsam.cpp` (which is fed to Pybind) into smaller files that can be parallelly built. At the end of the day, it generates one shared object file so that would be the easiest way to do it.

jingwuOUO:
When `make -j1 python-install`, it often causes my computer freezing and raise error `g++-7: internal compiler error: Killed (program cc1plus)`, my processor is i7-8550U @ 1.80GHz x 8.

varunagrawal:
What happens when you run `make -j8`?

I agree that the Pybind11 wrapper is a major memory hog. I had to update my swap from 8 GB to 24 GB to prevent my laptop from freezing as well.

jingwuOUO:
Tried, it will crash sooner. The same error message.

jingwuOUO:
Oh my swap is 8GB. I might need to update.

johnwlambert:
CI passes now that TBB + Python is disabled in the CI

gtsam/geometry/Point3.h
```diff
@@ -38,6 +38,8 @@ typedef Vector3 Point3;
 typedef std::pair<Point3, Point3> Point3Pair;
 GTSAM_EXPORT std::ostream &operator<<(std::ostream &os, const gtsam::Point3Pair &p);
 
+using Point3Pairs = std::vector<Point3Pair>;
```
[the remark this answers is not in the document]
johnwlambert: done

varunagrawal:
Interesting. I wonder if TBB is doing something to cause the OOM (instead of the wrapper). We need to update the TBB dependent code to use the new API.

PR #706, opened 2021-03-05
Fixing translation averaging for the panorama case

akshay-krishnan:
When the translation recovery class was used with unit translations that are all zero (all cameras at the same point), it would cause a segmentation fault. 
This PR adds a unit test for this case and fixes the segmentation fault.

### DIFF
--- a/gtsam/sfm/TranslationRecovery.cpp
+++ b/gtsam/sfm/TranslationRecovery.cpp
@@ -81,6 +81,7 @@ void TranslationRecovery::addPrior(
     const double scale, NonlinearFactorGraph *graph,
     const SharedNoiseModel &priorNoiseModel) const {
   auto edge = relativeTranslations_.begin();
+  if(edge == relativeTranslations_.end()) return;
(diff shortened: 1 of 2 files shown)


dellaert (approved):
Approved, but I'd like to know what the underlying GTSAM bug is that threw a segfault :-) We should fix that with an exception.

dellaert:
Also, CI does not pass?

johnwlambert:
> Also, CI does not pass?

Seems to be a homebrew issue, Fan suggested a fix and I put it in a different PR

akshay-krishnan:
> Approved, but I'd like to know what the underlying GTSAM bug is that threw a segfault :-) We should fix that with an exception.

I believe the segfault was in TranslationRecovery::addPrior, [this line](https://github.com/borglab/gtsam/blob/759dcaa1dd9bd799c6a531561563acfa9ae5032b/gtsam/sfm/TranslationRecovery.cpp#L85) specifically. Using the first edge in the graph when it is empty.

dellaert:
Ah, so it was your code, not GTSAM :-)

johnwlambert:
@akshay-krishnan can you pull in develop?

PR #708, opened 2021-03-07
Enabling pickle support to python types generated by the pybind wrapper

ayushbaid:
This PR will enable some python types for GTSAM to be [pickled](https://docs.python.org/3/library/pickle.html). Popular frameworks like numpy and pytorch already support pickling. This addition will enable better compatibility with multi-processing in Python and direct use in frameworks like Dask.

### DIFF
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -797,6 +797,12 @@ class Unit3 {
   size_t dim() const;
   gtsam::Unit3 retract(Vector v) const;
   Vector localCoordinates(const gtsam::Unit3& s) const;
+
+  // enabling serialization functionality
+  void serialize() const;
+
+  // enabling function to compare objects
+  bool equals(const gtsam::Unit3& expected, double tol) const;
 };
 
 #include <gtsam/geometry/EssentialMatrix.h>
--- a/python/gtsam/tests/test_pickle.py
+++ b/python/gtsam/tests/test_pickle.py
@@ -44,5 +44,3 @@ class TestPickle(GtsamTestCase):
     def test_unit3_roundtrip(self):
         obj = Unit3(Point3(1, 1, 0))
         self.assertEqualityOnPickleRoundtrip(obj)
-
-    
--- a/python/gtsam/tests/test_pickle.py
+++ b/python/gtsam/tests/test_pickle.py
@@ -1,0 +1,48 @@
+"""
+GTSAM Copyright 2010-2020, Georgia Tech Research Corporation,
+Atlanta, Georgia 30332-0415
+All Rights Reserved
+
+See LICENSE for the license information
+
+Unit tests to check pickling.
+
+Author: Ayush Baid
+"""
+from gtsam import Cal3Bundler, PinholeCameraCal3Bundler, Point2, Point3, Pose3, Rot3, SfmTrack, Unit3 
+
+from gtsam.utils.test_case import GtsamTestCase
+
+class TestPickle(GtsamTestCase):
+    """Tests pickling of types where it has been configured."""
+
+    def test_cal3Bundler_roundtrip(self):
+        obj = Cal3Bundler(fx=100, k1=0.1, k2=0.2, u0=100, v0=70)
+        self.assertEqualityOnPickleRoundtrip(obj)
+    
+    def test_pinholeCameraCal3Bundler_roundtrip(self):
+        obj = PinholeCameraCal3Bundler(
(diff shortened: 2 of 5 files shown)


python/gtsam/tests/test_pickle.py
```diff
@@ -0,0 +1,48 @@
...
+
+Unit tests to check pickling.
+
+Author: Ayush Baid
+"""
+from gtsam import Cal3Bundler, PinholeCameraCal3Bundler, Point2, Point3, Pose3, Rot3, SfmTrack, Unit3 
+
+from gtsam.utils.test_case import GtsamTestCase
+
+class TestPickle(GtsamTestCase):
```
ayushbaid, line 16: I created this single file for all the pickle tests, but I prefer having them in their individual test files.

The problem is that we currently have the python test file for `Pose3` and not for other types. So I might need to create a file for each class I am supporting right now, and more classes in the future.

wrap/gtwrap/pybind_wrapper.py
```diff
@@ -292,6 +296,37 @@ def wrap_namespace(self, namespace):
...
+        """
+        if cpp_class not in [
+            "gtsam::Cal3Bundler",
+            "gtsam::PinholeCameraCal3Bundler", 
+            "gtsam::Pose3", 
+            "gtsam::Rot3", 
+            "gtsam::SfmTrack"
+            "gtsam::Unit3", 
+        ]:
+            # TODO: fix this check by just checking if serialization is supported.
```
ayushbaid, line 314: I am relying on the serialization function to make my life easier. Things like `BetweenFactor` do not support the serialization functionality. I know this check against a list of 5 objects (what is needed for GT-SFM right now) is not optimal, and I request alternate suggestions.

gtsam/gtsam.i
```diff
@@ -797,6 +797,12 @@ class Unit3 {
   size_t dim() const;
   gtsam::Unit3 retract(Vector v) const;
   Vector localCoordinates(const gtsam::Unit3& s) const;
+
```
johnwlambert, line 800: Thanks for the PR, @ayushbaid . Is Unit3 the only class that needs serialization here, and if so, why note Rot3, Pose3, etc?

python/gtsam/tests/test_pickle.py
johnwlambert: Maybe the code is easier to understand if all of the serialization code is together (less copy-paste elsewhere)? But maybe less in keeping of the GTSAM philosophy of a single test file per class. So I'm not sure, perhaps it's ok right now

wrap/gtwrap/pybind_wrapper.py
johnwlambert: Is it correct to just return an empty string for these 6 classes?

wrap/gtwrap/pybind_wrapper.py
```diff
@@ -292,6 +296,37 @@ def wrap_namespace(self, namespace):
...
+            "gtsam::Cal3Bundler",
+            "gtsam::PinholeCamera<gtsam::Cal3Bundler>", 
+            "gtsam::Pose3", 
+            "gtsam::Rot3", 
+            "gtsam::SfmTrack",
+            "gtsam::Unit3", 
+        ]:
+            # TODO: fix this check by just checking if serialization is supported.
+            return ""
+        return textwrap.dedent('''
```
johnwlambert, line 316: could you maybe add some inline comments explaining why the textwrap dedent module is needed here, and why the C++ code needs to be provided as a string in Python?

gtsam/gtsam.i
ayushbaid: All other classes already have the `serialization` and `equals` API exposed in gtsam.i. `Unit3` was missing it.

wrap/gtwrap/pybind_wrapper.py
ayushbaid: This won't apply now. With help from @ProfFan, we now add pickle support bundled with serialization. So all classes which can be serialized are supported now.

wrap/gtwrap/pybind_wrapper.py
ayushbaid: This same structure has been used in this file. Do I still add the comments?

ProfFan:
You need to start from a fresh copy of `develop`, `update_wrap` and then add the changes.

ayushbaid:
Closing this PR as the pybind code has been added to the wrap repo.

PR #709, opened 2021-03-08
feature/pickle-support

ayushbaid:
Adding pickle support by using serialization


ayushbaid:
This PR just has the wrap code pulled in. The build fails because there are GTSAM classes which do not have a public constructor with no arguments. Example: `gtsam::LinearContainerFactor`

ProfFan:
Oh. Maybe then you need to update wrap to include a mechanism like `serialize()` marker, for example a `pickle()` marker.

ProfFan (approved):

ProfFan:
This is great! Thanks @ayushbaid !

Issue #407, opened 2020-07-14
[WIP] New MATLAB wrapper


ProfFan:
We need to finish this before 4.1.

- [x] code generation for some method signatures is broken (wrong type)
- [x] matlab compiles (only for gtsam, `gtsam_unstable` is broken even in develop)
- [x] we have not tested on MATLAB at all (I don't know how to load the module)
- [x] resolve overloads or add numbering?
- [x] resolve the issue where the numbering of C++ and matlab functions mismatch
- [x] resolve the crashes on factor construction
- [x] `This` declarations are broken because the concrete type is dropped during parsing

ProfFan:
@dellaert

ProfFan:
I have updated the `feature/new_wrapper` branch to reflect the changes.

ProfFan:
I have finished the wrapping for matlab and briefly tested on MATLAB (loading and simple function works), could anyone help me test it by:

- Clone the `feature/new_wrapper` branch
- `git submodule update --init --recursive`
- mkdir build && cd build
- `cmake -DCMAKE_BUILD_TYPE=Release -DGTSAM_TYPEDEF_POINTS_TO_VECTORS=ON -DGTSAM_INSTALL_MATLAB_TOOLBOX=ON -DCMAKE_INSTALL_PREFIX=../gtsam_install -DGTSAM_CUSTOM_MATLAB_PATH=<your matlab>/MATLAB/R2020a/bin ..`
- `make -j8 install`
- Go to `../gtsam_install/gtsam_toolbox` and run the matlab wrapper

And report the issues you have? Many thanks!

@varunagrawal @mikesheffler @dellaert

ProfFan:
I am getting errors like this when running the examples:
```
PlanarSLAMExample
Error using gtsam.NonlinearFactorGraph
Error: File: NonlinearFactorGraph.m Line: 341 Column: 5
'push_back' is already defined as a method in the 'NonlinearFactorGraph' class.

Error in PlanarSLAMExample (line 29)
graph = NonlinearFactorGraph;
 ```

So we probably need to do some name mangling here.

ProfFan:
Same goes for the file name resolution. Really need some help from someone with better MATLAB experience...

varunagrawal:
I'll take a look at this later today.

ProfFan:
@varunagrawal Thank you in advance :) I need a list of problems so I can tweak the generation code...

ProfFan:
[...]

PR #711, opened 2021-03-10
Use wrap to perform Matlab wrapping

varunagrawal:
This PR updates the Matlab wrapper so that it is completely reliant on `wrap`.

I've also made a git template on the org called `gtsam-project-matlab` which follows the exact same structure as the `example_project`, which is why this PR also deletes the `example_project`.


varunagrawal:
I also want to check if this fixes #407.

varunagrawal:
A couple of tests are failing due to segfaults. I would wait to merge this PR until we get those fixed.

- testSerialization.m
- testPlanarSLAMExample.m

I'll look into these tomorrow since I have some other stuff to finish up today.

varunagrawal:
I got all the MATLAB tests to pass!!!!!!!

Issue #634, opened 2020-12-09
error: no type named 'library_version_type' in namespace 'boost::serialization'


DYFeng:
## Description

Compile on macOS failed.

## Steps to reproduce

1.  git checkout develop
2. cmake .. && make

```
In file included from /Users/dyfeng/Projects/gtsam/gtsam/inference/BayesTree.cpp:21:
In file included from /Users/dyfeng/Projects/gtsam/gtsam/inference/BayesTree.h:22:
In file included from /Users/dyfeng/Projects/gtsam/gtsam/inference/Key.h:20:
In file included from /Users/dyfeng/Projects/gtsam/gtsam/base/FastList.h:25:
/usr/local/include/boost/serialization/list.hpp:53:33: error: no type named 'library_version_type' in namespace 'boost::serialization'; did you mean
      'item_version_type'?
    const boost::serialization::library_version_type library_version(
          ~~~~~~~~~~~~~~~~~~~~~~^
```

## Environment

OS: macOS
Apple clang version 12.0.0 (clang-1200.0.32.2)
Target: x86_64-apple-darwin20.1.0
boost: 1.74.0

varunagrawal:
So this is actually an issue with Boost 1.74. I encountered this issue today morning and found this: https://github.com/boostorg/serialization/issues/217#issuecomment-737284894 

@DYFeng can you please downgrade to Boost 1.73?

wolfv:
Adding 

```
#include <boost/serialization/version.hpp>
```


to LossFunctions.h and SubGraphBuilder.h fixes this issue.

Yuanzhuangz:
> Adding
> 
> ```
> #include <boost/serialization/version.hpp>
> ```
> 
> to LossFunctions.h and SubGraphBuilder.h fixes this issue.

why can't i find LossFunction.h,where is it?

wwtinwhu:
> Adding
> 
> ```
> #include <boost/serialization/version.hpp>
> ```
> 
> to LossFunctions.h and SubGraphBuilder.h fixes this issue.

Hello, after I adding it, this issue still not solved.  Any advice? Thanks! @DYFeng @wolfv @varunagrawal 
### Environment
OS: ubuntu22.04
Compiler: gcc 11.4
boost: 1.74
gtsam version: 4.0.3

wwtinwhu:
> > Adding
> > ```
> > #include <boost/serialization/version.hpp>
> > ```
> > 
> > 
> >     
> >       
> >     
[...]

PR #713, opened 2021-03-11
Fix/boost issue

dellaert:
Fixes issue #634 
Added #include <boost/serialization/version.hpp> headers...

### DIFF
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -1809,7 +1809,7 @@ class GaussianFactorGraph {
 };
 
 #include <gtsam/linear/GaussianConditional.h>
-virtual class GaussianConditional : gtsam::GaussianFactor {
+virtual class GaussianConditional : gtsam::JacobianFactor {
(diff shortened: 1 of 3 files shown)


gtsam/gtsam.i
```diff
@@ -1809,7 +1809,7 @@ class GaussianFactorGraph {
 };
 
 #include <gtsam/linear/GaussianConditional.h>
-virtual class GaussianConditional : gtsam::GaussianFactor {
+virtual class GaussianConditional : gtsam::JacobianFactor {
```
varunagrawal, line 1812: Why change the super class for this? Doesn't GaussianFactor make sense for a GaussianConditional?

varunagrawal (approved):
LGTM with a question.

gtsam/gtsam.i
dellaert: JacobianFactor is in turn a GaussianFactor

PR #720, opened 2021-03-18
Fix CI with separate boost install

varunagrawal:
Github Actions no longer supports boost, so we install it separately.
Also turned off some checks since the wrapper causes an OOM exception.


varunagrawal:
Yussss!!!!

PR #721, opened 2021-03-20
Unit tests for Jacobian of PartialPriorFactor<Pose3>

miloknowles:
Adds unit tests for `PartialPriorFactor<Pose3>` to show that the Jacobians for this factor are incorrect. Specifically, the factor precomputes a Jacobian that is correct for an identity pose, but doesn't take into account the rotation/translation at the current linearization point. See `/gtsam_unstable/slam/tests/testPartialPriorFactor.cpp` for the relevant tests (the identity case succeeds while the others fail). This is addressing the post [here](https://groups.google.com/g/gtsam-users/c/jiiYCer8nUY) on the Google group.

Haven't contributed to GTSAM before so apologies if I'm missing a part of the workflow. I thought I'd open the issue now, and start working on a fix in the meantime.

### DIFF
--- a/gtsam/base/ProductLieGroup.h
+++ b/gtsam/base/ProductLieGroup.h
@@ -161,6 +161,9 @@ public:
     }
     return v;
   }
+  static TangentVector LocalCoordinates(const ProductLieGroup& p, ChartJacobian Hp = boost::none) {
+    return Logmap(p, Hp);
+  }
   ProductLieGroup expmap(const TangentVector& v) const {
     return compose(ProductLieGroup::Expmap(v));
   }
--- a/gtsam_unstable/slam/PartialPriorFactor.h
+++ b/gtsam_unstable/slam/PartialPriorFactor.h
@@ -111,14 +111,14 @@ namespace gtsam {
     Vector evaluateError(const T& p, boost::optional<Matrix&> H = boost::none) const override {
       if (H) {
         Matrix H_logmap;
-        p.localCoordinates(T::identity(), H_logmap);
+        T::Logmap(p, H_logmap);
         (*H) = Matrix::Zero(indices_.size(), T::dimension);
         for (size_t i = 0; i < indices_.size(); ++i) {
           (*H).row(i) = H_logmap.row(indices_.at(i));
         }
       }
       // Compute the tangent vector representation of T and select relevant parameters.
-      const Vector& full_logmap = p.localCoordinates(T::identity());
+      const Vector& full_logmap = T::Logmap(p);
       Vector partial_logmap = Vector::Zero(T::dimension);
       for (size_t i = 0; i < indices_.size(); ++i) {
         partial_logmap(i) = full_logmap(indices_.at(i));
--- a/gtsam_unstable/slam/PartialPriorFactor.h
+++ b/gtsam_unstable/slam/PartialPriorFactor.h
@@ -107,16 +107,23 @@ namespace gtsam {
 
     /** Returns a vector of errors for the measured tangent parameters.  */
     Vector evaluateError(const T& p, boost::optional<Matrix&> H = boost::none) const override {
+      Eigen::Matrix<double, T::dimension, T::dimension> H_local;
+
+      // If the Rot3 Cayley map is used, Rot3::LocalCoordinates will throw a runtime error
+      // when asked to compute the Jacobian matrix (see Rot3M.cpp).
+      #ifdef GTSAM_ROT3_EXPMAP
+      const Vector full_tangent = T::LocalCoordinates(p, H ? &H_local : nullptr);
+      #else
+      const Vector full_tangent = T::Logmap(p, H ? &H_local : nullptr);
+      #endif
+
       if (H) {
-        Matrix H_local;
-        T::LocalCoordinates(p, H_local);
         (*H) = Matrix::Zero(indices_.size(), T::dimension);
         for (size_t i = 0; i < indices_.size(); ++i) {
           (*H).row(i) = H_local.row(indices_.at(i));
         }
       }
-      // Compute the tangent vector representation of T and select relevant parameters.
-      const Vector& full_tangent = T::LocalCoordinates(p);
+      // Select relevant parameters from the tangent vector.
       Vector partial_tangent = Vector::Zero(indices_.size());
       for (size_t i = 0; i < indices_.size(); ++i) {
         partial_tangent(i) = full_tangent(indices_.at(i));
--- a/gtsam_unstable/slam/PartialPriorFactor.h
+++ b/gtsam_unstable/slam/PartialPriorFactor.h
@@ -119,7 +119,7 @@ namespace gtsam {
       }
       // Compute the tangent vector representation of T and select relevant parameters.
       const Vector& full_logmap = T::Logmap(p);
-      Vector partial_logmap = Vector::Zero(T::dimension);
+      Vector partial_logmap = Vector::Zero(indices_.size());
       for (size_t i = 0; i < indices_.size(); ++i) {
         partial_logmap(i) = full_logmap(indices_.at(i));
       }
--- a/gtsam_unstable/slam/PartialPriorFactor.h
+++ b/gtsam_unstable/slam/PartialPriorFactor.h
@@ -17,8 +17,6 @@
 
 #pragma once
 
-#include <Eigen/Core>
-
 #include <gtsam/nonlinear/NonlinearFactor.h>
 #include <gtsam/base/Lie.h>
 
@@ -110,21 +108,21 @@ namespace gtsam {
     /** Returns a vector of errors for the measured tangent parameters.  */
     Vector evaluateError(const T& p, boost::optional<Matrix&> H = boost::none) const override {
       if (H) {
-        Matrix H_logmap;
-        T::Logmap(p, H_logmap);
+        Matrix H_local;
+        T::LocalCoordinates(p, H_local);
         (*H) = Matrix::Zero(indices_.size(), T::dimension);
         for (size_t i = 0; i < indices_.size(); ++i) {
-          (*H).row(i) = H_logmap.row(indices_.at(i));
+          (*H).row(i) = H_local.row(indices_.at(i));
         }
       }
       // Compute the tangent vector representation of T and select relevant parameters.
-      const Vector& full_logmap = T::Logmap(p);
-      Vector partial_logmap = Vector::Zero(indices_.size());
+      const Vector& full_tangent = T::LocalCoordinates(p);
+      Vector partial_tangent = Vector::Zero(indices_.size());
       for (size_t i = 0; i < indices_.size(); ++i) {
-        partial_logmap(i) = full_logmap(indices_.at(i));
+        partial_tangent(i) = full_tangent(indices_.at(i));
       }
 
-      return partial_logmap - prior_;
+      return partial_tangent - prior_;
     }
 
     // access
--- a/gtsam_unstable/slam/PartialPriorFactor.h
+++ b/gtsam_unstable/slam/PartialPriorFactor.h
@@ -31,11 +31,9 @@ namespace gtsam {
    *
    * The prior vector used in this factor is stored in compressed form, such that
    * it only contains values for measurements that are to be compared, and they are in
-   * the same order as VALUE::Logmap().  The mask will determine which components to extract
-   * in the error function.
+   * the same order as VALUE::Logmap(). The provided indices will determine which components to
+   * extract in the error function.
    *
-   * For practical use, it would be good to subclass this factor and have the class type
-   * construct the mask.
    * @tparam VALUE is the type of variable the prior effects
    */
   template<class VALUE>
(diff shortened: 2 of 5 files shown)


dellaert:
Nice. I would put the fix in this same PR, as we can’t merge failing tests :-) do you know how to fix it?

miloknowles:
I think so - the `evaluateError` function already calls `LieGroup::logmap`to get the tangent vector, so getting the Jacobian from `logmap` and then selecting the relevant row(s) seems like an easy solution.

Unrelated, but I'd also like to rename the "mask" argument in the factor to something else, since it's a list of tangent vector indices rather than a binary mask (I found this slightly confusing at first).

miloknowles:
Whoops, going to avoid the "mask" renaming since `DGroundConstraint` and others depend on this member. One question: should I prefer `localCoordinates` or `logmap`?

dellaert:
> I think so - the `evaluateError` function already calls `LieGroup::logmap`to get the tangent vector, so getting the Jacobian from `logmap` and then selecting the relevant row(s) seems like an easy solution.
> 
> Unrelated, but I'd also like to rename the "mask" argument in the factor to something else, since it's a list of tangent vector indices rather than a binary mask (I found this slightly confusing at first).

Ah, unrelated PRs are unrelated :-) Let's keep this one easy to review? If you've already don it no issue.

localcoordinates is indeed preferred as that will allow any manifold type as well.

miloknowles:
Already made the change, apologies - I'll keep my PRs minimal in the future though! :)

Switched to `localCoordinates`, although I only get the correct Jacobians when I use `p.localCoordinates(T::identity())` instead of the other way around, for reasons I don't quite understand...

Based on `PoseRotationPrior` and `PoseTranslationPrior`, it seems like the correct way to compute error is `prior_.localCoordinates(p)`. However, since we only store *part* of the prior tangent vector in this factor, we would have to assume values for the other parameters to follow this pattern. I'm not sure how to do that except for by assuming the identity, in which case we may as well use `Logmap` or `LocalCoordinates`. Let me know what you think would be best.

dellaert:
Interesting. I propose you try something that you think is right, but do it in a very test-driven way. Definitely test away from zero, and for non-trivial manifolds, e.g. Pose3.

CI seems to be failing now, BTW

gtsam_unstable/slam/tests/testPartialPriorFactor.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+
+/* ************************************************************************* */
+TEST(PartialPriorFactor, JacobianAtIdentity)
+{
+  Key poseKey(1);
+  Pose3 measurement = Pose3::identity();
+  SharedNoiseModel model = noiseModel::Isotropic::Sigma(1, 0.25);
+
+  TestPartialPriorFactor factor(poseKey, kIndexTy, measurement.translation().x(), model);
```
bricerebsamen, line 49:
```suggestion
  TestPartialPriorFactor factor(poseKey, kIndexTy, measurement.translation().y(), model);
```

gtsam_unstable/slam/tests/testPartialPriorFactor.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+
+/* ************************************************************************* */
+TEST(PartialPriorFactor, JacobianPartialTranslation) {
+  Key poseKey(1);
+  Pose3 measurement(Rot3::RzRyRx(0.15, -0.30, 0.45), Point3(-5.0, 8.0, -11.0));
+  SharedNoiseModel model = noiseModel::Isotropic::Sigma(1, 0.25);
+
+  TestPartialPriorFactor factor(poseKey, kIndexTy, measurement.translation().x(), model);
```
bricerebsamen, line 73:
```suggestion
  TestPartialPriorFactor factor(poseKey, kIndexTy, measurement.translation().y(), model);
```

miloknowles: Thanks for catching that, @bricerebsamen!

miloknowles:
@dellaert I decided to use `T::Logmap` since I think most GTSAM users will be using this factor with Lie groups anyways. I've added unit tests for `PartialPriorFactor<Pose2/Pose3>` at a variety of different linearization points and am pretty confident the Jacobians are correct now. Interestingly, it looks like PR #724 is fixing the exact same Jacobian problem in the UnaryFactor tutorial. Will deal with the CI failures now.

dellaert:
> @dellaert I decided to use `T::Logmap` since I think most GTSAM users will be using this factor with Lie groups anyways. I've added unit tests for `PartialPriorFactor<Pose2/Pose3>` at a variety of different linearization points and am pretty confident the Jacobians are correct now. Interestingly, it looks like PR #724 is fixing the exact same Jacobian problem in the UnaryFactor tutorial. Will deal with the CI failures now.

Hmmm. What's the problem using `T::LocalCoordinates`? It seems Logmap is artificially constraining the usefulness. Maybe I don't see the problem.

miloknowles:
> > @dellaert I decided to use `T::Logmap` since I think most GTSAM users will be using this factor with Lie groups anyways. I've added unit tests for `PartialPriorFactor<Pose2/Pose3>` at a variety of different linearization points and am pretty confident the Jacobians are correct now. Interestingly, it looks like PR #724 is fixing the exact same Jacobian problem in the UnaryFactor tutorial. Will deal with the CI failures now.
> 
> Hmmm. What's the problem using `T::LocalCoordinates`? It seems Logmap is artificially constraining the usefulness. Maybe I don't see the problem.

I forgot to mention - `PoseRTV` does not implement `LocalCoordinates`, only `Logmap`, which is why I had ended up using `Logmap`.

dellaert:
> > > @dellaert I decided to use `T::Logmap` since I think most GTSAM users will be using this factor with Lie groups anyways. I've added unit tests for `PartialPriorFactor<Pose2/Pose3>` at a variety of different linearization points and am pretty confident the Jacobians are correct now. Interestingly, it looks like PR #724 is fixing the exact same Jacobian problem in the UnaryFactor tutorial. Will deal with the CI failures now.
> > 
> > 
> > Hmmm. What's the problem using `T::LocalCoordinates`? It seems Logmap is artificially constraining the usefulness. Maybe I don't see the problem.
> 
> I forgot to mention - `PoseRTV` does not implement `LocalCoordinates`, only `Logmap`, which is why I had ended up using `Logmap`. There are some some partial prior factors in `DynamicsPriors.h` that are templated on the `PoseRTV` type.

So, PoseRTV is superseded by NavState, but that's a different PR :-)

But the problem with LocalCoordinates might be easily solved. It might simply be a missing entry at line 78 in PoseRTV.h ?


```
  /// @name Manifold
  /// @{
  using Base::dimension;
  using Base::dim;
  using Base::Dim;
  using Base::retract;
  using Base::localCoordinates;
  /// @}
```

miloknowles:
> > > > @dellaert I decided to use `T::Logmap` since I think most GTSAM users will be using this factor with Lie groups anyways. I've added unit tests for `PartialPriorFactor<Pose2/Pose3>` at a variety of different linearization points and am pretty confident the Jacobians are correct now. Interestingly, it looks like PR #724 is fixing the exact same Jacobian problem in the UnaryFactor tutorial. Will deal with the CI failures now.
> > > 
> > > 
> > > Hmmm. What's the problem using `T::LocalCoordinates`? It seems Logmap is artificially constraining the usefulness. Maybe I don't see the problem.
> > 
> > 
> > I forgot to mention - `PoseRTV` does not implement `LocalCoordinates`, only `Logmap`, which is why I had ended up using `Logmap`. There are some some partial prior factors in `DynamicsPriors.h` that are templated on the `PoseRTV` type.
> 
> So, PoseRTV is superseded by NavState, but that's a different PR :-)
> 
> But the problem with LocalCoordinates might be easily solved. It might simply be a missing entry at line 78 in PoseRTV.h ?
> 
> ```
>   /// @name Manifold
>   /// @{
>   using Base::dimension;
>   using Base::dim;
>   using Base::Dim;
>   using Base::retract;
>   using Base::localCoordinates;
>   /// @}
> ```

I think because `ProductLieGroup` doesn't implement `LocalCoordinates` (just `Logmap`), we can't use the simple fix :(
Maybe that's a different PR too?

dellaert:
> I think because `ProductLieGroup` doesn't implement `LocalCoordinates` (just `Logmap`), we can't use the simple fix :(
> Maybe that's a different PR too?

No, I think that's an oversight and we could put that in this PR. *OR* switch to a different type for testing this PR. Adding PoseRTV code is probably a mistake, so maybe just switch to NavState which should have all the manifold goodness.

miloknowles:
> > I think because `ProductLieGroup` doesn't implement `LocalCoordinates` (just `Logmap`), we can't use the simple fix :(
> > Maybe that's a different PR too?
> 
> No, I think that's an oversight and we could put that in this PR. _OR_ switch to a different type for testing this PR. Adding PoseRTV code is probably a mistake, so maybe just switch to NavState which should have all the manifold goodness.

Sorry for all the back and forth - I opted to just add `LocalCoordinates` to `ProductLieGroup` so that we get it in `PoseRTV`. Unfortunately `NavState` doesn't have `LocalCoordinates` either, unless I'm missing some kind of manifold inheritance. Also, replacing `PoseRTV` with `NavState` would require a lot of changes in `DynamicsPriors` and `dynamics/tests/testIMUSystem.cpp`, and I didn't want to make the review any more involved.

gtsam/base/ProductLieGroup.h
```diff
@@ -161,6 +161,9 @@ class ProductLieGroup: public std::pair<G, H> {
     }
     return v;
   }
+  static TangentVector LocalCoordinates(const ProductLieGroup& p, ChartJacobian Hp = boost::none) {
```
dellaert, line 164: Awesome!

gtsam_unstable/slam/PartialPriorFactor.h
```diff
@@ -102,35 +100,34 @@ namespace gtsam {
...
+      if (H) {
+        Matrix H_local;
+        T::LocalCoordinates(p, H_local);
+        (*H) = Matrix::Zero(indices_.size(), T::dimension);
+        for (size_t i = 0; i < indices_.size(); ++i) {
+          (*H).row(i) = H_local.row(indices_.at(i));
+        }
+      }
+      // Compute the tangent vector representation of T and select relevant parameters.
+      const Vector& full_tangent = T::LocalCoordinates(p);
```
dellaert, line 119: should not be reference I think

gtsam_unstable/slam/PartialPriorFactor.h
```diff
@@ -102,35 +100,34 @@ namespace gtsam {
...
-      // FIXME: this was originally the generic retraction - may not produce same results
-      Vector full_logmap = T::Logmap(p);
-//      Vector full_logmap = T::identity().localCoordinates(p); // Alternate implementation
-      Vector masked_logmap = Vector::Zero(this->dim());
-      for (size_t i=0; i<mask_.size(); ++i)
-        masked_logmap(i) = full_logmap(mask_[i]);
-      return masked_logmap - prior_;
+      if (H) {
+        Matrix H_local;
+        T::LocalCoordinates(p, H_local);
```
dellaert, line 112: Move outside of if statement and write
```
Matrix H_local;
const Vector full_tangent = T::LocalCoordinates(p, H? & H_local : nullptr);
```

dellaert (commented):
Looking good! A few more comments re modern Jacobian style. Re-request review if CI succeeds and I'll approve (or Brice can).

gtsam_unstable/slam/PartialPriorFactor.h
```diff
@@ -102,35 +100,41 @@ namespace gtsam {
...
-//      Vector full_logmap = T::identity().localCoordinates(p); // Alternate implementation
-      Vector masked_logmap = Vector::Zero(this->dim());
-      for (size_t i=0; i<mask_.size(); ++i)
-        masked_logmap(i) = full_logmap(mask_[i]);
-      return masked_logmap - prior_;
+      Eigen::Matrix<double, T::dimension, T::dimension> H_local;
+
+      // If the Rot3 Cayley map is used, Rot3::LocalCoordinates will throw a runtime error
+      // when asked to compute the Jacobian matrix (see Rot3M.cpp).
+      #ifdef GTSAM_ROT3_EXPMAP
```
dellaert, line 114: Sure, but why would we do this for every possible T?
But I'll let this slip, as (a) the fix is not straightforward, (b) this is not an often used factor, and (c) GTSAM_ROT3_EXPMAP is the default...

dellaert (approved):

dellaert:
Thanks @miloknowles !!!!

gtsam_unstable/slam/PartialPriorFactor.h
miloknowles: Sorry, I'm not sure I follow. Do you mean we just shouldn't use this factor if `GTSAM_ROT3_EXPMAP` is turned off?

PR #730, opened 2021-04-04
addPriorDouble

dellaert:
Added double as template argument

### DIFF
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2091,7 +2091,7 @@ class NonlinearFactorGraph {
   gtsam::KeySet keys() const;
   gtsam::KeyVector keyVector() const;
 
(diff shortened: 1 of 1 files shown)


gchenfc (commented):
Could we also add the fixed-vector versions:
`Vector1, Vector2, Vector3, Vector4, Vector5, Vector6` ?

Might be a little bit excessive but otherwise causes annoying errors:
```
Attempting to retrieve value with key "24209042725404672", type stored in Values is
gtsam::GenericValue<Eigen::Matrix<double, 6, 1, 0, 6, 1> > but requested type was
Eigen::Matrix<double, -1, 1, 0, -1, 1>
```
that are quite annoying to work around from the Python side.
*edit, I guess strictly speaking Vector2 and Vector3 aren't necessary since one could use Point2 and Point3, and Vector1 should be quite rare...  But Vector6 is quite common in GTDynamics

dellaert:
@gchenfc feel free to do a different PR. But the possible confusion between static and dynamic vector types makes that a less straightforward PR. In any case, orthogonal to this one, which solves a simple problem in GTDynamics.

dellaert:
@gchenfc please approve this so I can merge :-) CI seems OK, only the one time-out...

gchenfc (approved):

PR #733, opened 2021-04-06
Matlab Wrapper function to extract Vectors from a Values object

gchenfc:
@mattking-smith and I added a function in `utilities.h`, which is primarily used for the matlab wrapper, to extract all the vectors from a Values object.

Similar to extractPose3 and extractPose2, this just makes a lot of matlab code significantly faster than for-looping.


gtsam/nonlinear/utilities.h
```diff
@@ -142,6 +143,22 @@ Matrix extractPose3(const Values& values) {
   return result;
 }
 
+/// Extract all Vector values into a single tall vector
```
dellaert, line 146: A single tall vector? I don’t like that at all :-) at least Pose3 gets extracted into a matrix.
What is the use case here, really?
And what happens if you want fixed size vectors?

gchenfc: This cannot handle fixed size vectors (they don't get returned), but from matlab it's actually quite difficult (impossible other than Point2/Point3?) to place fixed-size vectors into a Values object anyway in the current develop branch.

The output is a single tall vector since (a) it's very rare to have fixed-size vectors anyway and (b) since the vectors are potentially different sizes, then you can't place in a matrix without zero-padding.  It's actually quite easy in matlab to `reshape(vectors, N, [])` if you know all your vectors are the same size.

As for the use case, lots of linear dynamics problems will have vectors (as we have seen) and extracting the vectors from a Values result using `for i =1:N; result.atVector(i); end` is often glacially slow.  I remember actually having this experience over the summer as well and as a dirty hack I even manually parsed the serialized strings to make extracting values faster in matlab (Adam was on an old version of gtsam so I couldn't make PR's to the develop branch).  Therefore, I believe these types of convenience extraction functions have utility.

The choice to use dynamic vectors instead of fixed-size is less obvious, but my naive and subjective opinion is that for the flexibility it is reasonable.

ProfFan: Maybe it's better to figure out why the extraction is so slow. Recently Varun has revamped a lot of the old code, so maybe performance has improved? (and the code is now easier to read)

mattking-smith: Hi All,

I cannot tell if this issue has been resolved or not. @gchenfc I am still using the custom extract with the reshape command, but I know that @dellaert wanted another solution. What is the end user (like a corporate sponsor) suppose to do to use the extract for vectors from a value object in MATLAB?

mattking-smith (approved):

gtsam/nonlinear/utilities.h
varunagrawal: You do a `values.atVector(key)`. @mattking-smith please do not approve PRs without providing _sufficient reasoning_ on the approval especially if your issue is not quite as related. :slightly_smiling_face:

dellaert: > You do a `values.atVector(key)`. @mattking-smith please do not approve PRs without providing _sufficient reasoning_ on the approval especially if your issue is not quite as related. 🙂

Also, Matt, can't approve your *own* PR :-)

gtsam/nonlinear/utilities.h
```diff
@@ -142,6 +143,22 @@ Matrix extractPose3(const Values& values) {
...
+Vector extractVector(const Values& values) {
```
dellaert, line 147: I think an API that makes more sense is to give a character, say 'x`' that will be used to select all values with a particular Symbol, e.g. x4, x5, x6, and that will assume that they are all of the same dimension. Use case is then:
```
result = extractVectors(values, 'x')
```
and if there are 200 such vectors, and they are 5D, the result is a 200x5 Matrix (not 5x200, we want to be consistent with other methods in this file). You can thrown an exception if those values have different dimensions.

dellaert (changes requested):

dellaert:
Also, merge in develop and resolve conflicts (carefully!)

gtsam/nonlinear/utilities.h
gchenfc: @varunagrawal I think this was an efficiency question.  I probably should have made the PR description more clear, but to clarify:

Doing `values.atVector(key)` for, say, a hundred keys is glacially slow and will actually consume the vast majority of the runtime for many problems.  I was telling Matt that one time I actually used something like `fscanf('%f ', values.serialize())` or something because that was a thousand times faster than `values.atVector(key)` for every key.

I'm pretty sure the other functions in `utilities.h` were written for the same reason - someone else realized how slow it is to loop through keys in matlab.

I can work together with matt to fix this up after ICRA.

dellaert: Yes, it's a valid PR. I only have issues with the API and the "one big vector". See my review comment.

varunagrawal: My response was more directed towards the corporate sponsor part. I imagine for most cases, getting all the vectors as one big matrix would not be a major necessity.

dellaert (approved):
I love that we are reviving this PR. Just change the docs etc to read `mxn` not ‘NxM’, because:
- matrices are typically mxn 
- N is a better constant for dimensionality IMHO

mattking-smith:
> I love that we are reviving this PR. Just change the docs etc to read `mxn` not ‘NxM’, because:
> 
> * matrices are typically mxn
> * N is a better constant for dimensionality IMHO

I am going to verify this update on my computer now.

matlab/gtsam_tests/testUtilities.m
```diff
@@ -52,5 +52,5 @@
 values.insert(2, (13:18)');
 values.insert(7, Pose3());
 actual = utilities.extractVector(values);
-expected = reshape(1:18, 3, 6);
+expected = reshape(1:18, 6, 3)';
```
dellaert, line 55: Hmmm. It should be 3*6, right?

mattking-smith:
@gchenfc I just switched to this branched, compiled and tried running the `testUtilities.m` and I immediately ran into:
```
>> testUtilities
Error using gtsam_wrapper
Exception from gtsam:
eliminateMultifrontal expects 1 arguments, not 0


Error in gtsam.utilities.extractVector (line 3)
        varargout{1} = gtsam_wrapper(2379, varargin{:});

Error in testUtilities (line 54)
actual = utilities.extractVector(values);
```

I am guessing that this may be related to https://github.com/borglab/gtsam/issues/954#issue-1073641616 and to your current [wrapper issue](https://github.com/borglab/wrap/issues/134).

mattking-smith (changes requested):
@gchenfc I believe this this needs work. I am failing the basic unit test:
```
>> testUtilities
Error using gtsam_wrapper
Exception from gtsam:
eliminateMultifrontal expects 1 arguments, not 0


Error in gtsam.utilities.extractVector (line 3)
        varargout{1} = gtsam_wrapper(2379, varargin{:});

Error in testUtilities (line 54)
actual = utilities.extractVector(values);
```

matlab/gtsam_tests/testUtilities.m
gchenfc: matlab reshapes columns-first so we have to do 6, 3 then transpose

Just to be clear, `expected` will be:
```
1 2 3 4 5 6
7 8 9 10 11 12
13 14 15 16 17 18
```
This is what you're expecting, right?

gchenfc:
Oops I had a typo in unit test, can you try again with updated .m file?

No clue why EliminateMultifrontal is involved - but if it's still having issues after updated .m file we can chat to resolve.

mattking-smith:
> Oops I had a typo in unit test, can you try again with updated .m file?

I looked at the file changes you made and implemented them on my system so I could quickly test your fix without recompiling and I found the following error:

```
>> actual = utilities.extractVectors(values);

CHECK('extractVector', all(actual == expected, 'all'));
Error using gtsam.utilities.extractVectors (line 5)
Arguments do not match any overload of function extractVectors
```

Looking at the output of the wrapper definition I see the following:
```
function varargout = extractVectors(varargin)
      if length(varargin) == 2 && isa(varargin{1},'gtsam.Values') && isa(varargin{2},'char')
        varargout{1} = gtsam_wrapper(89, varargin{:});
      else
        error('Arguments do not match any overload of function extractVectors');
      end
```

So I tried the following tests:
```
>> actual = utilities.extractVectors(values,'7')

actual =

     []

>> actual = utilities.extractVectors(values,'x1')
Error using gtsam_wrapper
Exception from gtsam:
The error message identifier is invalid.


Error in gtsam.utilities.extractVectors (line 3)
        varargout{1} = gtsam_wrapper(89, varargin{:});
 
>> actual = utilities.extractVectors(values,'x')

actual =

     []
```
Am I inputting the symbol correctly?

gchenfc:
Omg I'm so careless - it should be the last one:

```
utilities.extractVectors(values,'x')
```

Actually the test should be like this:

```
% test extractVectors
values = Values();
values.insert(symbol('x', 0), (1:6)');
values.insert(symbol('x', 1), (7:12)');
values.insert(symbol('x', 2), (13:18)');
values.insert(symbol('x', 7), Pose3());
actual = utilities.extractVectors(values, 'x');
expected = reshape(1:18, 6, 3)';
CHECK('extractVectors', all(actual == expected, 'all'));
```

Sorry for the confusion
Since pushing causes CI to take a long time and slows down everyone else, could you try this before I push these changes?

mattking-smith:
> % test extractVectors
> values = Values();
> values.insert(symbol('x', 0), (1:6)');
> values.insert(symbol('x', 1), (7:12)');
> values.insert(symbol('x', 2), (13:18)');
> values.insert(symbol('x', 7), Pose3());
> actual = utilities.extractVectors(values, 'x');
> expected = reshape(1:18, 6, 3)';
> CHECK('extractVectors', all(actual == expected, 'all'));

Running this snippet of code resulted in:
```
>> % test extractVectors
values = Values();
values.insert(symbol('x', 0), (1:6)');
values.insert(symbol('x', 1), (7:12)');
values.insert(symbol('x', 2), (13:18)');
values.insert(symbol('x', 7), Pose3());
actual = utilities.extractVectors(values, 'x')
expected = reshape(1:18, 6, 3)';
CHECK('extractVectors', all(actual == expected, 'all'));

actual =

     1     2     3     4     5     6
     7     8     9    10    11    12
    13    14    15    16    17    18
```

This is the correct format and executed flawlessly.

mattking-smith:
> Sorry for the confusion Since pushing causes CI to take a long time and slows down everyone else, could you try this before I push these changes?

Can you push these changes so I can pull and then just test the .m files real quick?

gchenfc:
pushed - thanks so much for testing for me!!!

mattking-smith (approved):
> pushed - thanks so much for testing for me!!!

I have pulled and verified this fix. 
`>> testUtilities`

Thank you for doing this @gchenfc

mattking-smith:
>Since pushing causes CI to take a long time and slows down everyone else, could you try this before I push these changes?

Once the CI tests are successfully completed, we should be able to wrap up this PR @dellaert. Nice work @gchenfc.

matlab/gtsam_tests/testUtilities.m
dellaert: Yep, did not see the transpose... Sorry!

PR #735, opened 2021-04-07
KeyFormatter - boost::function to std::function

varunagrawal:
This PR updates the KeyFormatter typedef from `boost::function` to `std::function`.

@gchenfc and I discovered that the pybind11 wrapper plays well with std::function but not with boost::function. So this change will enable the direct use of `gtsam::KeyFormatter` in the wrapped code without any major modifications or hacks.

The best part is that std::function is a drop-in replacement to boost::function for this particular use case, so there should be no breaking change.

### DIFF
--- a/gtsam/inference/Key.h
+++ b/gtsam/inference/Key.h
@@ -25,7 +25,6 @@
 #include <gtsam/base/types.h>
 #include <gtsam/dllexport.h>
 
-#include <boost/function.hpp>
 #include <functional>
 
 #include <iosfwd>
--- a/gtsam/inference/Key.h
+++ b/gtsam/inference/Key.h
@@ -26,13 +26,14 @@
 #include <gtsam/dllexport.h>
 
 #include <boost/function.hpp>
+#include <functional>
 
 #include <iosfwd>
 
 namespace gtsam {
 
 /// Typedef for a function to format a key, i.e. to convert it to a string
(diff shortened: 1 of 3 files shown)


gtsam/inference/Key.h
```diff
@@ -26,13 +26,14 @@
 #include <gtsam/dllexport.h>
 
 #include <boost/function.hpp>
+#include <functional>
```
dellaert, line 29: So do we still need boost function header?

dellaert (approved):
Modulo comment

gtsam/inference/Key.h
gchenfc: done.  Side effect: Symbol and LabeledSymbol now need to include `boost/function.hpp` but they arguably should have included this to begin with for exactly this reason because they both explicitly use boost::function.

We should probably go through and replace all boost::function with std::function but probably best for a different PR since this is focused on KeyFormatter

varunagrawal: We did because some files that included Key.h had a dependency on boost::function.

+1 for @gchenfc's comment.

dellaert: > We did because some files that included Key.h had a dependency on boost::function.
> 
> +1 for @gchenfc's comment.

That's kitchen-sink header thinking. Headers should be included where used, not anywhere else.

PR #738, opened 2021-04-08
Update "print" function arguments

gchenfc:
After https://github.com/borglab/wrap/pull/86 gets merged, it will allow optional arguments for the gtsam wrapper interface file.  This updates the `gtsam.i` definitions to take advantage of these for the print statements.

This actually basically accomplishes the same thing as overloading... except that it also supports kwargs which can be handy.  e.g.
```python
factor.print_('factor: ', lambda key : return "the key is {}".format(key))
factor.print_(keyFormatter = lambda key : return "the key is {}".format(key))
```

Admittedly, the value add is not super large but imho it can still simplify the wrapping a little bit.


gchenfc:
Note: this does the same as https://github.com/borglab/gtsam/pull/750/commits/d0d8f480395e1373203dc5c24ef0bf9b8d6e6741 from #750 but keeps the same default arguments as the c++ defaults, and also adds the keyFormatter optional arguments to the classes which accept it.

dellaert:
Ping me for a re-review when you and @varunagrawal agree on a solution

gtsam/gtsam.i
```diff
@@ -1236,7 +1239,8 @@ class SymbolicBayesNet {
   SymbolicBayesNet();
   SymbolicBayesNet(const gtsam::SymbolicBayesNet& other);
   // Testable
-  void print(string s="") const;
+  void print(string s = "BayesNet",
```
varunagrawal, line 1242: Should this be SymbolicBayesNet?

gtsam/gtsam.i
```diff
@@ -1582,7 +1590,8 @@ class VectorValues {
 #include <gtsam/linear/GaussianFactor.h>
 virtual class GaussianFactor {
   gtsam::KeyVector keys() const;
-  void print(string s="") const;
+  void print(string s = "Factor",
```
varunagrawal, line 1593: Keep it empty like in the header file?

gtsam/gtsam.i
```diff
@@ -1610,7 +1619,8 @@ virtual class JacobianFactor : gtsam::GaussianFactor {
   JacobianFactor(const gtsam::GaussianFactorGraph& graph);
 
   //Testable
-  void print(string s="") const;
+  void print(string s = "Factor",
```
varunagrawal, line 1622: Keep it empty like in the header file?

gtsam/gtsam.i
```diff
@@ -1659,7 +1669,8 @@ virtual class HessianFactor : gtsam::GaussianFactor {
 
   //Testable
   size_t size() const;
-  void print(string s="") const;
+  void print(string s = "Factor",
```
varunagrawal, line 1672: Keep it empty like in the header file?

gtsam/gtsam.i
```diff
@@ -1684,7 +1695,8 @@ class GaussianFactorGraph {
   GaussianFactorGraph(const gtsam::GaussianBayesTree& bayesTree);
 
   // From FactorGraph
-  void print(string s="") const;
+  void print(string s = "FactorGraph",
```
varunagrawal, line 1698: Leave this empty or change it to `GaussianFactorGraph`?

varunagrawal (changes requested):
Looks fine to me, but we we should make the default `s` values consistent with the C++ defaults or update them to the subclass value.

varunagrawal:
I went ahead and updated this PR since I want to land it at the earliest.

dellaert (approved):

gtsam/gtsam.i
dellaert: Should be same default as c++ which should not change @varunagrawal so change before merge if you *did* change.

varunagrawal: This is the default for the base class, so it will also print in the subclass. Should I override the print method?

varunagrawal: To clarify, the `print` in `FactorGraph` has "FactorGraph" as the default and since `GaussianFactorGraph` inherits from `FactorGraph`, it also prints "FactorGraph" as the default (which to me isn't quite right).

dellaert: Not this PR to fix it

varunagrawal:
Okay I am merging this and will create a new PR for overriding the `print`s.

varunagrawal (approved):

PR #739, opened 2021-04-12
Wrapper update

gchenfc:
This merges in the updates from [wrap](https://github.com/borglab/wrap)

Includes:
- multiple interface files cmake command
- KeyFormatter updates for python
- support for templated base class
- redirects c++ stdout to python's sys.stdout for python


dellaert (approved):
Thanks for the docs update, too!

PR #744, opened 2021-04-17
Remove unused body_P_sensor param

asa:
Parm is mentioned but no longer used in CombinedImuFactor.

### DIFF
--- a/gtsam/navigation/CombinedImuFactor.h
+++ b/gtsam/navigation/CombinedImuFactor.h
@@ -211,7 +211,7 @@ public:
    * @param measuredAcc Measured acceleration (in body frame, as given by the
    * sensor)
    * @param measuredOmega Measured angular velocity (as given by the sensor)
-   * @param deltaT Time interval between two consecutive IMU measurements
+   * @param dt Time interval between two consecutive IMU measurements
    */
   void integrateMeasurement(const Vector3& measuredAcc,
       const Vector3& measuredOmega, const double dt) override;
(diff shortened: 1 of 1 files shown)


dellaert:
Why not? Was there a PR removing its use. Should the right PR to add that functionality?

asa:
I am not sure of the history on this? Just noticed we had a param mentioned, but not interface for it and thought we might want to remove it till it's implemented or notate it as a todo.

varunagrawal:
I thought `body_P_sensor` was defined on the `PreintegrationParams` and not on the factor itself. As such removing that comment should be the right thing to do.

varunagrawal:
@asa can you also update the docstring for `dt`. It is right above your edit and is labelled as `deltaT`.

@dellaert I can confirm this is the case. There was a refactor 6 years ago to define `PreintegrationParams` and it seems the docstrings were not updated.

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -212,8 +212,6 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurements : public PreintegrationType
    * sensor)
    * @param measuredOmega Measured angular velocity (as given by the sensor)
    * @param deltaT Time interval between two consecutive IMU measurements
```
varunagrawal, line 214: Please update the documentation for this parameter as well?

varunagrawal (approved):
LGTM with a minor request.

gtsam/navigation/CombinedImuFactor.h
asa: done

PR #750, opened 2021-04-21
Wrapper Update

varunagrawal:
1. Fix global variable assignment.
2. Wrap enums defined inside classes.
3. Added default values for `print()` so that `__repr__` works.


dellaert:
Description is meaningless ;-) What is the change to wrap that is merged here?

dellaert:
PR comments are docs

dellaert (approved):
Great catch, as print was broken in python. Merge after CI suceeds.

PR #753, opened 2021-04-23
Adding factor which considers the essential matrix and camera calibration as variable

ayushbaid:


ayushbaid:
This is a work-in-progress PR. I have some tests failing and have to write down the test for Expression factors.

johnwlambert:
Which cost function are you using? Squared Sampson, Squared SED, or something else?

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,119 @@
...
+   * @param K calibration (common for both images) for key calibrationKey
+   * @param H1 optional jacobian in E
+   * @param H2 optional jacobian in K
+   * @return * Vector
+   */
+  Vector evaluateError(const EssentialMatrix& E, const CALIBRATION& K,
+    boost::optional<Matrix&> H1 = boost::none, boost::optional<Matrix&> H2 = boost::none) const override {
+    Vector error(1);
+    // converting from pixel coordinates to normalized coordinates cA and cB
+    Matrix cA_H_K; // dcA/dK
```
dellaert, line 86: Use fixed size

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,119 @@
...
+    Matrix cB_H_K; // dcB/dK
+    Point2 cA = K.calibrate(pA_, cA_H_K);
```
dellaert, line 88: Use ` H ? & cB_H_K : nullptr` To only evaluate derivatives if needed

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,119 @@
...
+    Point2 cB = K.calibrate(pB_, cB_H_K);
+
+    // Homogeneous the coordinates
+    Vector3 vA = EssentialMatrix::Homogeneous(cA);
+    Vector3 vB = EssentialMatrix::Homogeneous(cB);
+
+    if (H2){
+      // compute the jacobian of error w.r.t K
+
+      // dvX / dcX [3x2] = [1, 0], [0, 1], [0, 0]
```
dellaert, line 98: Compute this *fixed size* matrix in the constructor

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,119 @@
...
+      Matrix v_H_c = (Matrix(3, 2) << 1.0, 0.0, 0.0, 1.0, 0.0, 0.0).finished(); // [3x2]
+
+      // computing dvA/dK = dvA/dcA * dcA/dK and dVB/dK = dvB/dcB * dcB/dK
+      Matrix vA_H_K = v_H_c * cA_H_K;
```
dellaert, line 102: Fixed size

dellaert: Also, take a look at v_h_C and what this means for this product :-)

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,119 @@
...
+      Matrix vB_H_K = v_H_c * cB_H_K;
+
+      // error function f = vB.T * E * vA
+      // H2 = df/dK = vB.T * E.T * dvA/dK + vA.T * E * dvB/dK
+      *H2 = vB.transpose() * E.matrix().transpose() * vA_H_K + vA.transpose() * E.matrix() * vB_H_K;
+    }
+
+    error << E.error(vA, vB, H1);
```
dellaert, line 110: Same `?` trick

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,119 @@
...
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file EssentialMatrixWithCalibrationFactor.h
```
dellaert, line 13: Does it have to be in it’s on file, or can you just be added to the other factor file. You could call it essential matrix factory 3 :-)

gtsam/slam/tests/testEssentialMatrixWithCalibrationFactor.cpp
```diff
@@ -0,0 +1,446 @@
+/**
+ * @file testEssentialMatrixWithCalibrationFactor.cpp
+ * @brief Test EssentialMatrixWithCalibrationFactor class
+ * @author Ayush Baid
+ * @author Akshay Krishnan
+ * @date April 22, 2021
+ */
+
+#include <gtsam/slam/EssentialMatrixWithCalibrationFactor.h>
```
dellaert, line 9: Same comment, I would probably prefer to keep it in the same file and reuse the examples. It’s all two- view geometry so all this copy/paste does not really needed.

gtsam/slam/tests/testEssentialMatrixWithCalibrationFactor.cpp
```diff
@@ -0,0 +1,455 @@
...
+// TODO: maybe default value not good; assert with 0th
+Cal3_S2 trueK = Cal3_S2();
+// PinholeCamera<Cal3_S2> camera2(data.cameras[1].pose(), trueK);
+Rot3 trueRotation(c1Rc2);
+Unit3 trueDirection(c1Tc2);
+EssentialMatrix trueE(trueRotation, trueDirection);
+
+double baseline = 0.1;  // actual baseline of the camera
+
+Point2 pA(size_t i) { return data.tracks[i].measurements[0].second; }
```
akshay-krishnan, line 51: Maybe use a good name for this? pA looks like a variable more than a function. gePixelFromA or something? Also consider passing in data instead of using global variables. Same for other variables below.

dellaert:
Should work for any calibration :-)

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,127 @@
...
+          (Matrix(3, 2) << 1.0, 0.0, 0.0, 1.0, 0.0, 0.0).finished();  // [3x2]
+
+      // computing dvA/dK = dvA/dcA * dcA/dK and dVB/dK = dvB/dcB * dcB/dK
+      Matrix vA_H_K = v_H_c * cA_H_K;
+      Matrix vB_H_K = v_H_c * cB_H_K;
+
+      // error function f = vB.T * E * vA
+      // H2 = df/dK = vB.T * E.T * dvA/dK + vA.T * E * dvB/dK
+      *H2 = vB.transpose() * E.matrix().transpose() * vA_H_K +
+            vA.transpose() * E.matrix() * vB_H_K;
```
akshay-krishnan, line 115: We might have made a mistake here. 
I think it should be 
f = vB.T * E * vA
df/dvB = (E * vA).T = vA.T * E.T
df/dvA = vB.T * E
(transpose on E in a different place). Following up offline.

akshay-krishnan (commented):
Leaving some initial comments. Following up on the jacobian offline.

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,127 @@
...
+namespace gtsam {
+
+/**
+ * Factor that evaluates algebraic epipolar error (K^-1 p)'E (K^-1 p) for given
+ * essential matrix and calibration shared between two images.
+ */
+template <class CALIBRATION>
+class EssentialMatrixWithCalibrationFactor
+    : public NoiseModelFactor2<EssentialMatrix, CALIBRATION> {
+  Point2 pA_, pB_;  ///< points in pixel coordinates
```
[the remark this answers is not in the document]
akshay-krishnan: Sorry my bad, they go at the top, but maybe add "private" to make it explicit. :)

gtsam/slam/tests/testEssentialMatrixWithCalibrationFactor.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+// The rotation between body and camera is:
+gtsam::Point3 bX(1, 0, 0), bY(0, 1, 0), bZ(0, 0, 1);
+gtsam::Rot3 cRb = gtsam::Rot3(bX, bZ, -bY).inverse();
+
+namespace example1 {
+
+const string filename = findExampleDataFile("5pointExample1.txt");
+SfmData data;
+bool readOK = readBAL(filename, data);
```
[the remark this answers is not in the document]
akshay-krishnan: Looks like you do it below in a unit test. Will the other unit tests run if that fails?

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   *  Constructor
+   *  @param essentialMatrixKey Essential Matrix variable key
+   *  @param calibrationKey Calibration variable key
+   *  @param pA point in first camera, in pixel coordinates
+   *  @param pB point in second camera, in pixel coordinates
+   *  @param model noise model is about dot product in ideal, homogeneous coordinates
+   */
+  EssentialMatrixWithCalibrationFactor(Key essentialMatrixKey, Key calibrationKey, const Point2& pA, const Point2& pB,
```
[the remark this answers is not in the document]
ayushbaid: I will reformat now. 

As there is no format file, I just added `{ BasedOnStyle: Google }` to my VSCode settings. Anything else I should use?

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
ayushbaid: It will depend on the number of params for K, Hence I wont be able to use a fixed size.

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
ayushbaid: Thanks. Will do.

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
ayushbaid: if H1 is `nullptr`, it will be passed in as `nullptr`. So why do I need an explicit check.

gtsam/slam/tests/testEssentialMatrixWithCalibrationFactor.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+// Noise model for first type of factor is evaluating algebraic error
+noiseModel::Isotropic::shared_ptr model1 =
+    noiseModel::Isotropic::Sigma(1, 0.01);
+// Noise model for second type of factor is evaluating pixel coordinates
+noiseModel::Unit::shared_ptr model2 = noiseModel::Unit::Create(2);
+
+// The rotation between body and camera is:
+gtsam::Point3 bX(1, 0, 0), bY(0, 1, 0), bZ(0, 0, 1);
+gtsam::Rot3 cRb = gtsam::Rot3(bX, bZ, -bY).inverse();
```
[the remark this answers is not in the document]
ayushbaid: But this is not exactly identity?

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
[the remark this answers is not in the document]
dellaert: Right, use cppformat w Google style in VS code

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
dellaert: Not true at all: CALIBRATION is known at compile time

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
dellaert: Make sure to address this! It will save lots compute

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
dellaert: You are correct!

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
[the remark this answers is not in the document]
dellaert: indeed :-)

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,127 @@
...
+    // Homogeneous the coordinates
+    Vector3 vA = EssentialMatrix::Homogeneous(cA);
+    Vector3 vB = EssentialMatrix::Homogeneous(cB);
+
+    if (H2) {
+      // compute the jacobian of error w.r.t K
+
+      // dvX / dcX [3x2] = [1, 0], [0, 1], [0, 0]
+      Matrix v_H_c =
+          (Matrix(3, 2) << 1.0, 0.0, 0.0, 1.0, 0.0, 0.0).finished();  // [3x2]
```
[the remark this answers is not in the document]
dellaert: I think this matrix should disappear altogether, see my comment above

gtsam/slam/tests/testEssentialMatrixWithCalibrationFactor.cpp
dellaert: Please keep naming and mve back to old file, where this terminology is being used.

gtsam/slam/tests/testEssentialMatrixWithCalibrationFactor.cpp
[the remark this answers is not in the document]
akshay-krishnan: I see, yeah never mind then.

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,127 @@
...
+#include <iostream>
+
+namespace gtsam {
+
+/**
+ * Factor that evaluates algebraic epipolar error (K^-1 p)'E (K^-1 p) for given
+ * essential matrix and calibration shared between two images.
+ */
+template <class CALIBRATION>
+class EssentialMatrixWithCalibrationFactor
```
johnwlambert, line 37: is there a simpler name here? Not sure that it conveys the overall gist of the factor.

`EssentialMatrixProjectionFactor`?

ayushbaid: This name is a bit tricky as we are not doing any reprojection to compute error, which is what this name suggests to me. 

The word projection does not hint that calibration is a parameter.

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
ayushbaid: Changed it. Thanks!

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
ayushbaid: As this is identity matrix, I just initialize the Jacobian in the `homogeneous` function to identity.

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
ayushbaid: I added fixed size variable. The product looks right to me. Is it wrong?

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
ayushbaid: I am calling it `EssentialMatrixFactor4` as we have 3 factors already.

gtsam/slam/tests/testEssentialMatrixWithCalibrationFactor.cpp
ayushbaid: merged the files

ayushbaid:
> Should work for any calibration :-)

I cannot get the `calibrate` function of `Cal3Bundler` class to converge.

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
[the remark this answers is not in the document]
ayushbaid: added the private tag

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
```diff
@@ -0,0 +1,127 @@
...
+              << std::endl;
+  }
+
+  /// vector of errors returns 1D vector
+  /**
+   * @brief Calculate the algebraic epipolar error p' (K^-1)' E K p.
+   *
+   * @param E essential matrix for key essentialMatrixKey
+   * @param K calibration (common for both images) for key calibrationKey
+   * @param H1 optional jacobian in E
```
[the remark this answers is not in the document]
ayushbaid: thanks. fixed

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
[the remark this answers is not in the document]
ayushbaid: removed it as I moved the jacobian computation in the function which homogenizes the coordinates.

gtsam/slam/EssentialMatrixWithCalibrationFactor.h
ayushbaid: as discussed, looks right because the error computed is f = vA.T @ E @ vB.

dellaert:
> > Should work for any calibration :-)
> 
> I cannot get the `calibrate` function of `Cal3Bundler` class to converge.

Bizarre. But at least it will *compile* with it, right?

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -267,28 +269,121 @@ class EssentialMatrixFactor3: public EssentialMatrixFactor2 {
...
+   * @param K calibration (common for both images) for key keyK
+   * @param H1 optional jacobian of error w.r.t E
+   * @param H2 optional jacobian of error w.r.t K
+   * @return * Vector 1D vector of algebraic error
+   */
+  Vector evaluateError(
+      const EssentialMatrix& E, const CALIBRATION& K,
+      boost::optional<Matrix&> H1 = boost::none,
+      boost::optional<Matrix&> H2 = boost::none) const override {
+    Vector error(1);
```
dellaert, line 357: Declare where used, not before. And make const.

ayushbaid:
> > > Should work for any calibration :-)
> > 
> > 
> > I cannot get the `calibrate` function of `Cal3Bundler` class to converge.
> 
> Bizarre. But at least it will _compile_ with it, right?

Yes. No issues with compilation. The test for minimization doesnt work.

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -267,28 +269,121 @@ class EssentialMatrixFactor3: public EssentialMatrixFactor2 {
...
+class EssentialMatrixFactor4
+    : public NoiseModelFactor2<EssentialMatrix, CALIBRATION> {
+ private:
+  Point2 pA_, pB_;  ///< points in pixel coordinates
+
+  typedef NoiseModelFactor2<EssentialMatrix, CALIBRATION> Base;
+  typedef EssentialMatrixFactor4 This;
+
+  static const int DimK = FixedDimension<CALIBRATION>::value;
+  typedef Eigen::Matrix<double, 2, DimK> JacobianCalibration;
```
dellaert, line 311: nice!

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -267,28 +269,121 @@ class EssentialMatrixFactor3: public EssentialMatrixFactor2 {
...
+      boost::optional<Matrix&> H2 = boost::none) const override {
+    Vector error(1);
+    // converting from pixel coordinates to normalized coordinates cA and cB
+    JacobianCalibration cA_H_K;  // dcA/dK
+    JacobianCalibration cB_H_K;  // dcB/dK
+    Point2 cA = K.calibrate(pA_, H2 ? &cA_H_K : 0);
+    Point2 cB = K.calibrate(pB_, H2 ? &cB_H_K : 0);
+
+    // Homogeneous the coordinates
+    Matrix32 vA_H_cA, vB_H_cB;
```
dellaert, line 365: These are *not* needed as we know they are constant and super, simple. In fact, `E^T vA_H_cA` is
```
e11 e21 e31   1 0   e11 e21
e12 e22 e32 * 0 1 = e12 e22
e13 e23 e33   0 0   e13 e23   
```
i.e., `E.matrix().transpose().leftCols<2>()`
Similarly, E * vB_H_cB = E.matrix().leftCols<2>()`.

Unit test of course :-)

ayushbaid: Thanks. updated

gtsam/slam/EssentialMatrixFactor.h
ayushbaid:
```const Vector error(1);
error << E.error(vA, vB, H1) 
```
does not compile. 

```
 candidate: ‘Eigen::CommaInitializer<Derived> Eigen::DenseBase<Derived>::operator<<(const Scalar&) [with Derived = Eigen::Matrix<double, -1, 1>; Eigen::DenseBase<Derived>::Scalar = double]’ <near match>
  144 | inline CommaInitializer<Derived> DenseBase<Derived>::operator<< (const Scalar& s)
      |                                  ^~~~~~~~~~~~~~~~~~
passing ‘const Vector*’ {aka ‘const Eigen::Matrix<double, -1, 1>*’} as ‘this’ argument discards qualifiers

```

gtsam/slam/tests/testEssentialMatrixWithCalibrationFactor.cpp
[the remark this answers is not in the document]
ayushbaid: This check is not very helpful tbh. If the file does not exist, the test does not end up running:

```
Child aborted***Exception:   0.21 sec
terminate called after throwing an instance of 'std::invalid_argument'
  what():  gtsam::findExampleDataFile could not find a matching file
```

I dont think there will be a case when the file exists but is not a valid BAL file (like the review should not let this happen).

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -267,28 +269,121 @@ class EssentialMatrixFactor3: public EssentialMatrixFactor2 {
...
+  typedef NoiseModelFactor2<EssentialMatrix, CALIBRATION> Base;
+  typedef EssentialMatrixFactor4 This;
+
+  static const int DimK = FixedDimension<CALIBRATION>::value;
+  typedef Eigen::Matrix<double, 2, DimK> JacobianCalibration;
+
+ public:
+  /**
+   *  Constructor
+   *  @param keyE Essential Matrix variable key
```
akshay-krishnan, line 316: Can we add the direction of the essential matrix? A to B or B to A?

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -267,28 +269,121 @@ class EssentialMatrixFactor3: public EssentialMatrixFactor2 {
...
+      const std::string& s = "",
+      const KeyFormatter& keyFormatter = DefaultKeyFormatter) const override {
+    Base::print(s);
+    std::cout << "  EssentialMatrixFactor4 with measurements\n  ("
+              << pA_.transpose() << ")' and (" << pB_.transpose() << ")'"
+              << std::endl;
+  }
+
+  /**
+   * @brief Calculate the algebraic epipolar error p' (K^-1)' E K p.
```
akshay-krishnan, line 345: Maybe write this error in terms of pA and pB?

gtsam/slam/tests/testEssentialMatrixFactor.cpp
```diff
@@ -351,7 +353,112 @@ TEST (EssentialMatrixFactor3, minimization) {
...
+        boost::bind(&EssentialMatrixFactor4<Cal3_S2>::evaluateError, factor, _1,
+                    _2, boost::none, boost::none);
+    HEexpected = numericalDerivative21<Vector1, EssentialMatrix, Cal3_S2>(
+        f, trueE, trueK);
+    HKexpected = numericalDerivative22<Vector1, EssentialMatrix, Cal3_S2>(
+        f, trueE, trueK);
+
+    // Verify the Jacobian is correct
+    EXPECT(assert_equal(HEexpected, HEactual, 1e-8));
+    EXPECT(assert_equal(HKexpected, HKactual, 1e-8));
```
akshay-krishnan, line 385: Is this different from using EXPECT_CORRECT_FACTOR_JACOBIANS?

gtsam/slam/tests/testEssentialMatrixFactor.cpp
```diff
@@ -526,7 +633,59 @@ TEST (EssentialMatrixFactor3, extraTest) {
...
+  truth.insert(1, trueE);
+  truth.insert(2, trueK);
+  EXPECT_DOUBLES_EQUAL(0, graph.error(truth), 1e-8);
+
+  // Check error at initial estimate
+  Values initial;
+  EssentialMatrix initialE =
+      trueE.retract((Vector(5) << 0.1, -0.1, 0.1, 0.1, -0.1).finished());
+  Cal3Bundler initialK =
+      trueK.retract((Vector(3) << 0.1, -0.02, 0.03).finished());
```
akshay-krishnan, line 655: Did you get these values from elsewhere in gtsam? I wonder if the convergence problem is because it is too far from trueK. Even 0.02 and 0.03 might be too much for distortion parameters. The tolerance used in calibrate() is 1e-5.  I'm not sure about this though.

akshay-krishnan: And 0.1 is too small for focal length, which has a true value of 500?

ayushbaid:
I have added prior on K for the minimization test. However, there doesn't seem to be any any significant change in the value of K during optimization.

I checked the Jacobians, and they are non-zero.

@akshay-krishnan

dellaert:
> I have added prior on K for the minimization test. However, there doesn't seem to be any any significant change in the value of K during optimization.
> 
> I checked the Jacobians, and they are non-zero.
> 
> @akshay-krishnan

Jacobians being non-zero is not a good diagnostic. You need to check the rank of the Hessian, using linearizeToHessianFactor. Then do DVD on it and investigate the spectrum of eigen values.

ayushbaid:
Thanks Frank for the comment, I checked the singular values of the information matrix from the Hessian factor.
```
2.10e+05 4.61e+04 6.69e+03 4.11e+02 3.84e-02 3.09e-02 2.44e-02 1.79e-02 1.25e-03 4.86e-05
```

This is a very wide range of values, and something is going wrong. I have no hypothesis right now on what might be an issue.

ayushbaid:
> > Thanks Frank for the comment, I checked the singular values of the information matrix from the Hessian factor.
> > ```
> > 2.10e+05 4.61e+04 6.69e+03 4.11e+02 3.84e-02 3.09e-02 2.44e-02 1.79e-02 1.25e-03 4.86e-05
> > ```
> > 
> > 
> > This is a very wide range of values, and something is going wrong. I have no hypothesis right now on what might be an issue.
> 
> Is this _without a prior_ ? If so I would expect 7 non-zero values (sorted in decreasing magnitude) and zeros after that. Can you verify that?

This is without the prior, yes. But adding a prior also wont contribute to the Hessian evaluated on the initial values as the prior is on the initial values as mean too. I have checked this.

dellaert:
> > > Thanks Frank for the comment, I checked the singular values of the information matrix from the Hessian factor.
> > > ```
> > > 2.10e+05 4.61e+04 6.69e+03 4.11e+02 3.84e-02 3.09e-02 2.44e-02 1.79e-02 1.25e-03 4.86e-05
> > > ```
> > > 
> > > 
> > > This is a very wide range of values, and something is going wrong. I have no hypothesis right now on what might be an issue.
> > 
> > 
> > Is this _without a prior_ ? If so I would expect 7 non-zero values (sorted in decreasing magnitude) and zeros after that. Can you verify that?
> 
> This is without the prior, yes. But adding a prior also wont contribute to the Hessian evaluated on the initial values as the prior is on the initial values as mean too. I have checked this.

Sorry, but that does not make sense :-) Information is separate from error. In a Kalman filter, for example, you can precompute all hessians/covariances regardless of the measurements. We can discuss this offline.

I am just a bit surprised that we get more than 7 DOF information in the Hessian at all. Maybe it is "fake" information because of linearization. But *regardless* I would expect the SVD spectrum to fill up when you add a prior on K. I would experiment with different values for the noise models (for both image measurements and K) and understand their effect on the singular values and rank. Maybe by plotting a 2D array of bar graphs with log-scale Y.

ayushbaid:
I did some experiments with the number of input points and different types of priors.

The singular values are obtained by the information matrix obtained at the linearization of the graph to Hessian at the initial values.


## Effect of number of input correspondences
With a prior on `K` centered at the initial value of K, the following graph shows the ratios of singular values
singularVals_numPoints

## Effect of noise model
Trying out different values of sigma on the noise model on algebraic error. The prior is still on initial K.
singularVals_noiseLevel_initialKprior

Now, the prior is on the true K.
singularVals_noiseLevel_trueKprior

akshay-krishnan:
Sorry for the delayed response Ayush, great graphs! Questions: 
- how many points did you use for the 2nd and 3rd graphs (noise model tuning)?
- what is the value of sigma for the two noise models you used for the first graph (influence of number of points)?

johnwlambert:
Discussion from meeting:

1. try putting very small covariance on K? as if fixed, so just 5 points?
2. Maybe migrate the Sampson error into gtsam, out of gtsfm, into FundamentalMatrix? Maybe static method inside EssentialMatrix? And perhaps use Sampson here?

gtsam/slam/tests/testEssentialMatrixFactor.cpp
```diff
@@ -351,7 +353,117 @@ TEST (EssentialMatrixFactor3, minimization) {
   EXPECT_DOUBLES_EQUAL(0, graph.error(result), 1e-4);
 }
 
-} // namespace example1
+//*************************************************************************
+TEST(EssentialMatrixFactor4, factor) {
+  Key keyE(1);
+  Key keyK(1);
```
akshay-krishnan, line 359: Not sure how this test passed, shouldnt the keyK be initialized to 2?

gtsam/slam/tests/testEssentialMatrixFactor.cpp
```diff
@@ -351,7 +353,117 @@ TEST (EssentialMatrixFactor3, minimization) {
...
+  truth.insert(1, trueE);
+  truth.insert(2, trueK);
+  EXPECT_DOUBLES_EQUAL(0, graph.error(truth), 1e-8);
+
+  // Check error at initial estimate
+  Values initial;
+  EssentialMatrix initialE =
+      trueE.retract((Vector(5) << 0.1, -0.1, 0.1, 0.1, -0.1).finished());
+  Cal3_S2 initialK =
+      trueK.retract((Vector(5) << 0.1, -0.08, 0.01, -0.05, 0.06).finished());
```
akshay-krishnan, line 428: I tried different things here. 
If you add the prior on trueK instead of initialK and start from trueK, the test passes (note initialE != trueE).  
But if the prior is on initialK, K just does not change (even if the covariance is sigma on prior is super high). And E does not converge to the right solution either (which makes sense, as K is incorrect). 

With 5 points, we can only hope to optimize E right? Maybe we should add another test, where we have 8 points and the prior on initialK (not trueK).

dellaert: Yeah, actually, this makes sense: the prior disambiguates in favor of the K that's given. If we are only given 5 points, there is no redundancy and all 5 DOF are devoted to (R, d), also 5DOF.

I agree that with a new test we can check that in the case of > 5 correspondences the calibration is nudged toward the correct K. But, we will never get more than 7DOF from two-view correspondences, so the best we can hope for is that 2DOF of the calibration (any 2D linear combination) is corrected. Typically, we want the focal length to be corrected most, and hence constrain that the least, I'd say.

gtsam/slam/tests/testEssentialMatrixFactor.cpp
ayushbaid: you are right.

ayushbaid:
This PR should depend on #792 if it leads to better performance.

dellaert:
I'll be traveling next week but let me know when ready to re-review. In the meantime, check the CI: 22 checks are failing :-)

gtsam/slam/tests/testEssentialMatrixFactor.cpp
```diff
@@ -514,24 +729,81 @@ TEST (EssentialMatrixFactor3, extraTest) {
...
+    Hexpected2 =
+        numericalDerivative22<Vector2, EssentialMatrix, double>(f, bodyE, d);
 
     // Verify the Jacobian is correct
     EXPECT(assert_equal(Hexpected1, Hactual1, 1e-6));
     EXPECT(assert_equal(Hexpected2, Hactual2, 1e-8));
   }
 }
+/*
+TEST(EssentialMatrixFactor4, minimizationWithStrongCal3BundlerPrior2) {
```
akshay-krishnan, line 746: This test does not pass because calibrate() does not converge for our data. Should we remove it? @ayushbaid

akshay-krishnan:
> I'll be traveling next week but let me know when ready to re-review. In the meantime, check the CI: 22 checks are failing :-)

Ayush is also travelling, so I have pushed some changes. We are still using the algebraic error here. I think the switch to Sampson error can be done in a different PR, if we need to. Please take a look when you have time.

gtsam/slam/tests/testEssentialMatrixFactor.cpp
dellaert: Yes, if you're creating a factor, please use the macro. Change for all uses in this file.

gtsam/slam/tests/testEssentialMatrixFactor.cpp
```diff
@@ -351,7 +349,227 @@ TEST (EssentialMatrixFactor3, minimization) {
...
+  EXPECT(assert_equal(trueK, actualK, 1e-2));
+
+  // Check error at result
+  EXPECT_DOUBLES_EQUAL(0, graph.error(result), 1e-4);
+
+  // Check errors individually
+  for (size_t i = 0; i < 7; i++)
+    EXPECT_DOUBLES_EQUAL(
+        0,
+        actualE.error(EssentialMatrix::Homogeneous(actualK.calibrate(pA(i))),
```
dellaert, line 519: maybe add `using EssentialMatrix::Homogeneous` at top of file

dellaert (approved):
Approved subject to CI passing

dellaert:
Ci passes, so feel free to merge after latest comments have been addressed and CI passes again.

examples/CreateSFMExampleData.cpp
```diff
@@ -63,49 +57,68 @@ void createExampleBALFile(const string& filename, const vector<Point3>& P,
...
   vector<Point3> P;
-  P += Point3(0, 0, 100), Point3(-10, 0, 100), Point3(10, 0, 100), //
-  Point3(0, 50, 50), Point3(0, -50, 50), Point3(-20, 0, 80), Point3(20, -50, 80);
+  P += Point3(0, 0, 100), Point3(-10, 0, 100), Point3(10, 0, 100),  //
+      Point3(0, 50, 50), Point3(0, -50, 50), Point3(-20, 0, 80),
+      Point3(20, -50, 80);
 
   // Assumes example is run in ${GTSAM_TOP}/build/examples
-  const string filename = "../../examples/data/5pointExample2.txt";
+  const string filename = "../../examples/Data/5pointExample2.txt";
```
johnwlambert, line 90: why uppercase Data now, instead of lowercase data?

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -267,28 +269,128 @@ class EssentialMatrixFactor3: public EssentialMatrixFactor2 {
...
-}// gtsam
+/**
+ * Binary factor that optimizes for E and calibration K using the algebraic
+ * epipolar error (K^-1 pA)'E (K^-1 pB). The calibration is shared between two
+ * images.
+ *
+ * Note: As correspondences between 2d coordinates can only recover 7 DoF,
+ * this factor should always be used with a prior factor on calibration.
+ * Even with a prior, we can only optimize 2 DoF in the calibration. So the
+ * prior should have a noise model with very low sigma in the remaining
```
johnwlambert, line 304: Thanks for the PR, guys. should we give a definition for what range constitutes a "very low" sigma?

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -267,28 +269,128 @@ class EssentialMatrixFactor3: public EssentialMatrixFactor2 {
...
+    Point2 cB = K.calibrate(pB_, H2 ? &cB_H_K : 0, boost::none);
+
+    // convert to homogeneous coordinates
+    Vector3 vA = EssentialMatrix::Homogeneous(cA);
+    Vector3 vB = EssentialMatrix::Homogeneous(cB);
+
+    if (H2) {
+      // compute the jacobian of error w.r.t K
+
+      // error function f = vA.T * E * vB
```
johnwlambert, line 377: do we define `vA` anywhere?

examples/CreateSFMExampleData.cpp
akshay-krishnan: The folder in gtsam is now called "Data". I believe this file is a bit old.

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -7,35 +7,35 @@
...
 
 namespace gtsam {
 
 /**
  * Factor that evaluates epipolar error p'Ep for given essential matrix
  */
-class EssentialMatrixFactor: public NoiseModelFactor1<EssentialMatrix> {
-
-  Vector3 vA_, vB_; ///< Homogeneous versions, in ideal coordinates
+class EssentialMatrixFactor : public NoiseModelFactor1<EssentialMatrix> {
```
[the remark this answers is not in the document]
akshay-krishnan: For sure, sorry about that. Will keep in mind next time :)

gtsam/slam/EssentialMatrixFactor.h
akshay-krishnan: I think this is relative, it depends on the sigma that is not low. There are examples in the unit tests, which are still empirical. Not sure if there is an accurate answer that can be briefly documented.

gtsam/slam/EssentialMatrixFactor.h
akshay-krishnan: In lines above: 
```
    // convert to homogeneous coordinates   
     Vector3 vA = EssentialMatrix::Homogeneous(cA);
```
Is it not as obvious as I thought? We can add documentation then.

gtsam/slam/tests/testEssentialMatrixFactor.cpp
akshay-krishnan: Does that work for class names? I get an `error: ‘gtsam::EssentialMatrix’ is not a namespace or unscoped enum`.

PR #758, opened 2021-04-27
Avoid potential wrong memory access

jlblancoc:
(Spotted while randomly browsing the code)

If the user uses an invalid index, the [] operator won't check it and the program will access invalid memory. Using at() would throw instead.


dellaert (approved):
Coolio

PR #763, opened 2021-05-04
Update boost download link

varunagrawal:
As per [this notice](https://www.boost.org/users/news/boost_has_moved_downloads_to_jfr.html), the download link for boost has changed since Bintray has been retired. This is causing the CI to fail, which this PR fixes.


ProfFan (approved):
LGTM

yetongumich (commented):
LGTM

Issue #764, opened 2021-05-07
Custom factors & Python packaging


Skydes:
Hi there,

I have been playing with the Python API recently and I am really in awe of how neat it is! Coming from Ceres, everything seems so easy with GTSAM. Huge thanks to the team for pulling this :)

I want to optimize camera poses/calibration via reprojections while keeping the 3D points fixed (similar to https://github.com/borglab/gtsam/issues/697). There is no factor for this and I want to avoid writing a custom C++ factor, mostly because dependencies could then not anymore be handled by Python.

1. I hacked around using a prior on the points with very low`1e-8` noise, but this is hacky and inefficient, so I am looking for better :) Is there a way to freeze some optimization variables? Ceres has `SetParameterBlockConstant` but I couldn't find anything similar in GTSAM.
2. If not, any ETA on supporting custom factors in Python? This was mentioned in https://github.com/borglab/gtsam/issues/728.
3. We also use conda to handle dependencies, so I would be ok with writing a C++ factor if there was a GTSAM conda package.

ProfFan:
On (1), I think you can use zero covariance which will make the constraint an equality constraint :)
On (2), #748 is on my table now, I will start implementing it maybe tomorrow.
On (3), we currently do not use conda for distribution, but it should be fairly straightforward (contributions welcomed!)

Skydes:
Thanks for the quick reply.
1. For large graphs I observe that a zero covariance gives `inf` cost, but a `1e-8` sigma works fine.
2. That would be awesome! On our side I can think of many cases for which this would be useful.
3. I might look into it then.

ProfFan:
Python factor is now in develop, please try and test it hard so we can know what to change :)

halajun:
> 

Hi @ProfFan. I saw an example file called CustomFactorExample.py, and from here seems python custom factor has been supported? If so, how can I get to know about the details of defining custom factor, except this example? Thanks!

halajun:
> > 
> 
> Hi @ProfFan. I saw an example file called CustomFactorExample.py, and from here seems python custom factor has been supported? If so, how can I get to know about the details of defining custom factor, except this example? Thanks!

No worries. I searched the library globally and found the source files (CustomFactor.h, CustomFactor.cpp) in "nonlinear" folder. Cheers!

ProfFan:
[...]

Issue #748, opened 2021-04-20
Values-based FunctorizedFactor (for Python supercharging)


gchenfc:
## Feature

now that wrap supports functionals, could we make a factor in GTSAM similar to FunctorizedFactor but that accepts a functional on a values object rather than with a fixed list of actual types (i.e. in the style of NoiseModelFactor instead of NoiseModelFactorN)?

This would supercharge the python API so that you could write basically any custom factor you want in pure python without any re-compiles (although at the expense of slow runtime of the "linearize" step, if you choose to use this factor)

Example constructor:
```c++
  using OptionalJacobians = boost::optional<std::Vector<Matrix>&>;
  FunctorizedFactor(
      KeyVector keys, const SharedNoiseModel &model,
      const std::function<Vector(Values, OptionalJacobians)> unwhitenedError)
      : NoiseModelFactor(model, keys),
        func_(func) {}
```
Then in python:
```python
def myFactorError(x, H):
  # do calculations
  # populate H
  return z - h(x)
graph.add(FunctorizedFactor([symbol('x', 0), ...], noiseModel, myFactorError)
```

## Motivation

Fully custom factors in pure python - faster prototyping and easier adoption of GTSAM (based on my chats, it seems a lot of people intimidated by GTSAM are mostly intimidated by the C++ custom factors)

## Additional context

Currently there are only unary and binary versions: `FunctorizedFactor1` and `FunctorizedFactor2`.  They're templated which makes it difficult to anticipate in the python wrapper every combination of template arguments a user might want.  Much easier IMO to just make a Values-based version to expose to the python wrapper:
`FunctorizedFactor(keys, noiseModel, lambda values : return z - h(values))`

Should be relatively trivial - almost just a wrapping of `NoiseModelFactor`

ProfFan:
OK, I have a working prototype here: https://github.com/ProfFan/python-factor-test

If anyone wants to test it, just build like any other python project.

I'll start upstreaming if there is no objection to the API.

@gchenfc @varunagrawal @dellaert @Skydes

ProfFan:
Example code is in https://github.com/ProfFan/python-factor-test/blob/master/python/tests/test_custom_factor.py

Excerpt:
```
    def test_jacobian(self):
        """Tests if the factor result matches the GTSAM Pose2 unit test"""

        gT1 = Pose2(1, 2, np.pi/2)
        gT2 = Pose2(-1, 4, np.pi)

[...]

PR #767, opened 2021-05-14
Custom Factors in Python

ProfFan:
Per #748 and #764, this is a factor that allow one to create the evaluateError function in Python.

The unit tests document the usage.

It also contains a fix for the proper binding of `Point3Pairs`. @johnwlambert

### DIFF
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2170,6 +2170,21 @@ virtual class NoiseModelFactor: gtsam::NonlinearFactor {
 virtual class CustomFactor: gtsam::NoiseModelFactor {
   // Note CustomFactor will not be wrapped for MATLAB, as there is no supporting machinery there.
   CustomFactor();
+  /*
+   * Example:
+   * ```
+   * def error_func(this: CustomFactor, v: gtsam.Values, H: List[np.ndarray]):
+   *    <calculated error>
+   *    if not H is None:
+   *        <calculate the Jacobian>
+   *        H[0] = J1 # 2-d numpy array for a Jacobian block
+   *        H[1] = J2
+   *        ...
+   *    return error # 1-d numpy array
+   *
+   * cf = CustomFactor(noise_model, keys, error_func)
+   * ```
+   */
   CustomFactor(const gtsam::SharedNoiseModel& noiseModel, const gtsam::KeyVector& keys, const gtsam::CustomErrorFunction& errorFunction);
 };
 
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2166,6 +2166,12 @@ virtual class NoiseModelFactor: gtsam::NonlinearFactor {
   Vector whitenedError(const gtsam::Values& x) const;
 };
 
+#include <gtsam/nonlinear/CustomFactor.h>
+virtual class CustomFactor: gtsam::NoiseModelFactor {
+  CustomFactor();
+  CustomFactor(const gtsam::SharedNoiseModel& noiseModel, const gtsam::KeyVector& keys, const gtsam::CustomErrorFunction& errorFunction);
+};
+
 #include <gtsam/nonlinear/Values.h>
 class Values {
   Values();
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2168,6 +2168,7 @@ virtual class NoiseModelFactor: gtsam::NonlinearFactor {
 
 #include <gtsam/nonlinear/CustomFactor.h>
 virtual class CustomFactor: gtsam::NoiseModelFactor {
+  // Note CustomFactor will not be wrapped for MATLAB, as there is no supporting machinery there.
   CustomFactor();
   CustomFactor(const gtsam::SharedNoiseModel& noiseModel, const gtsam::KeyVector& keys, const gtsam::CustomErrorFunction& errorFunction);
 };
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2185,7 +2185,10 @@ virtual class CustomFactor: gtsam::NoiseModelFactor {
    * cf = CustomFactor(noise_model, keys, error_func)
    * ```
    */
-  CustomFactor(const gtsam::SharedNoiseModel& noiseModel, const gtsam::KeyVector& keys, const gtsam::CustomErrorFunction& errorFunction);
+  CustomFactor(const gtsam::SharedNoiseModel& noiseModel, const gtsam::KeyVector& keys,
+               const gtsam::CustomErrorFunction& errorFunction);
+
+  void print(string s = "", gtsam::KeyFormatter keyFormatter = gtsam::DefaultKeyFormatter);
 };
 
 #include <gtsam/nonlinear/Values.h>
--- a/gtsam/gtsam.i
+++ b/gtsam/gtsam.i
@@ -2168,7 +2168,10 @@ virtual class NoiseModelFactor: gtsam::NonlinearFactor {
 
 #include <gtsam/nonlinear/CustomFactor.h>
 virtual class CustomFactor: gtsam::NoiseModelFactor {
-  // Note CustomFactor will not be wrapped for MATLAB, as there is no supporting machinery there.
+  /*
+   * Note CustomFactor will not be wrapped for MATLAB, as there is no supporting machinery there.
+   * This is achieved by adding `gtsam::CustomFactor` to the ignore list in `matlab/CMakeLists.txt`.
+   */
   CustomFactor();
   /*
    * Example:
--- a/gtsam/nonlinear/CustomFactor.cpp
+++ b/gtsam/nonlinear/CustomFactor.cpp
@@ -19,13 +19,36 @@
 
 namespace gtsam {
 
+/*
+ * Calculates the unwhitened error by invoking the callback functor (i.e. from Python).
+ */
 Vector CustomFactor::unwhitenedError(const Values& x, boost::optional<std::vector<Matrix>&> H) const {
   if(this->active(x)) {
     if(H) {
+      /*
+       * In this case, we pass the raw pointer to the `std::vector<Matrix>` object directly to pybind.
+       * As the type `std::vector<Matrix>` has been marked as opaque in `preamble.h`, any changes in
+       * Python will be immediately reflected on the C++ side.
+       *
+       * Example:
+       * ```
+       * def error_func(this: CustomFactor, v: gtsam.Values, H: List[np.ndarray]):
+       *    <calculated error>
+       *    if not H is None:
+       *        <calculate the Jacobian>
+       *        H[0] = J1
+       *        H[1] = J2
+       *        ...
+       *    return error
+       * ```
+       */
       return this->errorFunction(*this, x, H.get_ptr());
     } else {
-      JacobianVector dummy;
-      return this->errorFunction(*this, x, &dummy);
+      /*
+       * In this case, we pass the a `nullptr` to pybind, and it will translated to `None` in Python.
+       * Users can check for `None` in their callback to determine if the Jacobian is requested.
+       */
+      return this->errorFunction(*this, x, nullptr);
     }
   } else {
     return Vector::Zero(this->dim());
--- a/gtsam/nonlinear/CustomFactor.cpp
+++ b/gtsam/nonlinear/CustomFactor.cpp
@@ -24,6 +24,7 @@ namespace gtsam {
  */
 Vector CustomFactor::unwhitenedError(const Values& x, boost::optional<std::vector<Matrix>&> H) const {
   if(this->active(x)) {
+
     if(H) {
       /*
        * In this case, we pass the raw pointer to the `std::vector<Matrix>` object directly to pybind.
@@ -45,7 +46,7 @@ Vector CustomFactor::unwhitenedError(const Values& x, boost::optional<std::vecto
       return this->error_function_(*this, x, H.get_ptr());
     } else {
       /*
-       * In this case, we pass the a `nullptr` to pybind, and it will translated to `None` in Python.
+       * In this case, we pass the a `nullptr` to pybind, and it will translate to `None` in Python.
        * Users can check for `None` in their callback to determine if the Jacobian is requested.
        */
       return this->error_function_(*this, x, nullptr);
--- a/gtsam/nonlinear/CustomFactor.cpp
+++ b/gtsam/nonlinear/CustomFactor.cpp
@@ -1,0 +1,35 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file    CustomFactor.cpp
+ * @brief   Class to enable arbitrary factors with runtime swappable error function.
+ * @author  Fan Jiang
+ */
+
+#include <gtsam/nonlinear/CustomFactor.h>
+
+namespace gtsam {
+
+Vector CustomFactor::unwhitenedError(const Values& x, boost::optional<std::vector<Matrix>&> H) const {
+  if(this->active(x)) {
+    if(H) {
+      return this->errorFunction(*this, x, H.get_ptr());
+    } else {
+      JacobianVector dummy;
+      return this->errorFunction(*this, x, &dummy);
+    }
+  } else {
+    return Vector::Zero(this->dim());
+  }
+}
+
+}
--- a/gtsam/nonlinear/CustomFactor.cpp
+++ b/gtsam/nonlinear/CustomFactor.cpp
@@ -42,17 +42,34 @@ Vector CustomFactor::unwhitenedError(const Values& x, boost::optional<std::vecto
        *    return error
        * ```
        */
-      return this->errorFunction(*this, x, H.get_ptr());
+      return this->error_function_(*this, x, H.get_ptr());
     } else {
       /*
        * In this case, we pass the a `nullptr` to pybind, and it will translated to `None` in Python.
        * Users can check for `None` in their callback to determine if the Jacobian is requested.
        */
-      return this->errorFunction(*this, x, nullptr);
+      return this->error_function_(*this, x, nullptr);
     }
   } else {
     return Vector::Zero(this->dim());
   }
+}
+
+void CustomFactor::print(const std::string &s, const KeyFormatter &keyFormatter) const {
+  std::cout << s << "CustomFactor on ";
+  auto keys_ = this->keys();
+  bool f = false;
+  for (const Key &k: keys_) {
+    if (f)
+      std::cout << ", ";
+    std::cout << keyFormatter(k);
+    f = true;
+  }
+  std::cout << "\n";
+  if (this->noiseModel_)
+    this->noiseModel_->print("  noise model: ");
+  else
+    std::cout << "no noise model" << std::endl;
 }
 
 }
--- a/gtsam/nonlinear/CustomFactor.h
+++ b/gtsam/nonlinear/CustomFactor.h
@@ -35,7 +35,7 @@ class CustomFactor;
  * This is safe because this is passing a const pointer, and pybind11 will maintain the `std::vector` memory layout.
  * Thus the pointer will never be invalidated.
  */
-using CustomErrorFunction = std::function<Vector(const CustomFactor&, const Values&, const JacobianVector*)>;
+using CustomErrorFunction = std::function<Vector(const CustomFactor &, const Values &, const JacobianVector *)>;
 
 /**
  * @brief Custom factor that takes a std::function as the error
@@ -46,7 +46,7 @@ using CustomErrorFunction = std::function<Vector(const CustomFactor&, const Valu
  */
 class CustomFactor: public NoiseModelFactor {
 protected:
-   CustomErrorFunction error_function_;
+  CustomErrorFunction error_function_;
 
 protected:
 
@@ -66,7 +66,7 @@ public:
    * @param keys keys of the variables
    * @param errorFunction the error functional
    */
-  CustomFactor(const SharedNoiseModel& noiseModel, const KeyVector& keys, const CustomErrorFunction& errorFunction) :
+  CustomFactor(const SharedNoiseModel &noiseModel, const KeyVector &keys, const CustomErrorFunction &errorFunction) :
       Base(noiseModel, keys) {
     this->error_function_ = errorFunction;
   }
@@ -80,16 +80,22 @@ public:
   Vector unwhitenedError(const Values &x, boost::optional<std::vector<Matrix> &> H = boost::none) const override;
 
   /** print */
-  void print(const std::string& s,
-             const KeyFormatter& keyFormatter = DefaultKeyFormatter) const override;
+  void print(const std::string &s,
+             const KeyFormatter &keyFormatter = DefaultKeyFormatter) const override;
 
+  /**
+   * Mark not sendable
+   */
+  bool sendable() const override {
+    return false;
+  }
 
 private:
 
   /** Serialization function */
   friend class boost::serialization::access;
   template<class ARCHIVE>
-  void serialize(ARCHIVE & ar, const unsigned int /*version*/) {
+  void serialize(ARCHIVE &ar, const unsigned int /*version*/) {
     ar & boost::serialization::make_nvp("CustomFactor",
                                         boost::serialization::base_object<Base>(*this));
   }
--- a/gtsam/nonlinear/CustomFactor.h
+++ b/gtsam/nonlinear/CustomFactor.h
@@ -45,13 +45,13 @@ using CustomErrorFunction = std::function<Vector(const CustomFactor&, const Valu
  * This factor is mainly for creating a custom factor in Python.
  */
 class CustomFactor: public NoiseModelFactor {
-public:
-   CustomErrorFunction errorFunction;
+protected:
+   CustomErrorFunction error_function_;
 
 protected:
 
-  typedef NoiseModelFactor Base;
-  typedef CustomFactor This;
+  using Base = NoiseModelFactor;
+  using This = CustomFactor;
 
 public:
 
@@ -68,7 +68,7 @@ public:
    */
   CustomFactor(const SharedNoiseModel& noiseModel, const KeyVector& keys, const CustomErrorFunction& errorFunction) :
       Base(noiseModel, keys) {
-    this->errorFunction = errorFunction;
+    this->error_function_ = errorFunction;
   }
 
   ~CustomFactor() override = default;
@@ -81,22 +81,7 @@ public:
 
   /** print */
   void print(const std::string& s,
-             const KeyFormatter& keyFormatter = DefaultKeyFormatter) const override {
-    std::cout << s << "CustomFactor on ";
-    auto keys_ = this->keys();
-    bool f = false;
-    for (const Key& k: keys_) {
-      if (f)
-        std::cout << ", ";
-      std::cout << keyFormatter(k);
-      f = true;
-    }
-    std::cout << "\n";
-    if (this->noiseModel_)
-      this->noiseModel_->print("  noise model: ");
-    else
-      std::cout << "no noise model" << std::endl;
-  }
+             const KeyFormatter& keyFormatter = DefaultKeyFormatter) const override;
 
 
 private:
--- a/gtsam/nonlinear/CustomFactor.h
+++ b/gtsam/nonlinear/CustomFactor.h
@@ -1,0 +1,84 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file    CustomFactor.h
+ * @brief   Class to enable arbitrary factors with runtime swappable error function.
+ * @author  Fan Jiang
+ */
+
+#pragma once
+
+#include <gtsam/nonlinear/NonlinearFactor.h>
+
+using namespace gtsam;
+
+namespace gtsam {
+
+typedef std::vector<Matrix> JacobianVector;
+
+class CustomFactor;
+
+typedef std::function<Vector(const CustomFactor&, const Values&, const JacobianVector*)> CustomErrorFunction;
+
+/**
+ * @brief Custom factor that takes a std::function as the error
+ * @addtogroup nonlinear
+ * \nosubgrouping
+ *
+ * This factor is mainly for creating a custom factor in Python.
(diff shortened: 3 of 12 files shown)


gtsam/nonlinear/CustomFactor.cpp
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+#include <gtsam/nonlinear/CustomFactor.h>
+
+namespace gtsam {
+
+Vector CustomFactor::unwhitenedError(const Values& x, boost::optional<std::vector<Matrix>&> H) const {
+  if(this->active(x)) {
+    if(H) {
+      return this->errorFunction(*this, x, H.get_ptr());
```
dellaert, line 25: Document what the mechanism is by which python can access. I expect you'll talk about pybind, types, how it appears in python, what python function should do.

gtsam/nonlinear/CustomFactor.cpp
```diff
@@ -0,0 +1,35 @@
...
+    } else {
+      JacobianVector dummy;
+      return this->errorFunction(*this, x, &dummy);
```
dellaert, line 28: Document what happens in  this case.

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,86 @@
...
+import unittest
+from gtsam import Values, Pose2, CustomFactor
+
+import numpy as np
+
+import gtsam
+from gtsam.utils.test_case import GtsamTestCase
+
+
+class TestCustomFactor(GtsamTestCase):
```
dellaert, line 21: doc

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def test_new(self):
```
dellaert, line 23: doc! Don't you use a linter?

dellaert (changes requested):
You **have to** document this. You are the *only* one* who will do this. It has to also show up in some markdown file, dson't know which one. This is probably worth thinking about re the "readthedocs" change.

matlab/CMakeLists.txt
```diff
@@ -61,7 +61,8 @@ endif()
 # ignoring the non-concrete types (type aliases)
 set(ignore
     gtsam::Point2
-    gtsam::Point3)
+    gtsam::Point3
+    gtsam::CustomFactor)
```
gchenfc, line 65: Can you add a comment in gtsam.i that CustomFactor won't be wrapped for matlab?

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,86 @@
...
+        def error_func(this: CustomFactor, v: gtsam.Values, H: List[np.ndarray]):
+            return np.array([1, 0, 0])
+        
+        noise_model = gtsam.noiseModel.Unit.Create(3)
+        cf = CustomFactor(noise_model, gtsam.KeyVector([0]), error_func)
```
gchenfc, line 28: is `gtsam.KeyVector(...)` necessary?  I thought for KeyVector then

```python
cf = CustomFactor(noise_model, [0], error_func)
```

should work?  If not, is there a different type for which this would work?

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,86 @@
...
+
+        def error_func(this: CustomFactor, v: gtsam.Values, H: List[np.ndarray]):
+            # print(f"{this = },\n{v = },\n{len(H) = }")
+
+            key0 = this.keys()[0]
+            key1 = this.keys()[1]
+            gT1, gT2 = v.atPose2(key0), v.atPose2(key1)
+            error = Pose2(0, 0, 0).localCoordinates(gT1.between(gT2))
+            
+            if len(H) > 0:
```
gchenfc, line 63: I feel the pythonic syntax would be `if H is not None: `, is there any way we can achieve this?  I know one way would be to include pybind in `CustomFactor.h` then use `pybind::none`, but I guess this isn't ideal.  Hoping for a null_ptr way or something.

gtsam/nonlinear/CustomFactor.cpp
gchenfc, line 28: could this pass in null_ptr instead or something?  Possibly by something along the lines of

```c++
typedef std::function<Vector(const CustomFactor&, const Values&, std::optional<JacobianVector*>)> CustomErrorFunction;
```
or something ???

gtsam/nonlinear/CustomFactor.h
```diff
@@ -0,0 +1,84 @@
...
+
+using namespace gtsam;
+
+namespace gtsam {
+
+typedef std::vector<Matrix> JacobianVector;
+
+class CustomFactor;
+
+typedef std::function<Vector(const CustomFactor&, const Values&, const JacobianVector*)> CustomErrorFunction;
```
gchenfc, line 30: Reason for `JacobianVector *` to be const?

gtsam/gtsam.i
```diff
@@ -2166,6 +2166,12 @@ virtual class NoiseModelFactor: gtsam::NonlinearFactor {
   Vector whitenedError(const gtsam::Values& x) const;
 };
 
+#include <gtsam/nonlinear/CustomFactor.h>
+virtual class CustomFactor: gtsam::NoiseModelFactor {
+  CustomFactor();
+  CustomFactor(const gtsam::SharedNoiseModel& noiseModel, const gtsam::KeyVector& keys, const gtsam::CustomErrorFunction& errorFunction);
```
gchenfc, line 2172: can you add a documentation/comment here about the args of `CustomErrorFunction`, just so that anyone using gtsam.i as an API reference doesn't have to go digging around to figure out how to use it?

gchenfc (approved):
Cool!

See (minor) comments

gtsam/nonlinear/CustomFactor.h
ProfFan: It's a const pointer to an array (whose size is preallocated by GTSAM). 

Design note: pybind11 will invoke a copy if this is `JacobianVector &`, and modifications in Python will not be reflected.

I'll add this to the comments

python/gtsam/preamble.h
```diff
@@ -6,9 +6,17 @@ PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Key, tbb::tbb_allocator<gtsam::Key>>);
 PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Key>);
 #endif
 PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Point2, Eigen::aligned_allocator<gtsam::Point2> >);
+PYBIND11_MAKE_OPAQUE(gtsam::Point3Pairs);
```
johnwlambert, line 9: thanks for the PR, Fan. Was this opaque option something we forgot to add earlier? What's the advantage of it -- preventing a deep copy?

ProfFan: Yes, it prevents a deep copy.

varunagrawal: This tells Pybind11 not to perform a cpp-to-python conversion of types and instead treat this as its own type. This help prevents issues with pass-by-reference propagation between the two languages.

ProfFan: ^ much better explanation

gtsam/gtsam.i
varunagrawal: This should be add-able via the docstring mechanism. That way, the editor shows it. Adding usage docs should go in the cpp file and not in the .i file imo

python/gtsam/tests/test_custom_factor.py
varunagrawal: We should consider making KeyVector default to a regular list on the python side while keeping the allocator semantics on the cpp side

gtsam/nonlinear/CustomFactor.cpp
varunagrawal: That's apparently doable as per some issues on Pybind11, but we'd have to use boost::optional because cpp11

python/gtsam/tests/test_custom_factor.py
varunagrawal: Yeah I want this too. I'll take a look at potential fixes later tonight. Lots of stuff to do for internship.

python/gtsam/tests/test_custom_factor.py
ProfFan: done

python/gtsam/tests/test_custom_factor.py
ProfFan: done

gtsam/nonlinear/CustomFactor.cpp
ProfFan: done

gtsam/nonlinear/CustomFactor.cpp
ProfFan: done

matlab/CMakeLists.txt
ProfFan: done

python/gtsam/tests/test_custom_factor.py
ProfFan: Now it should work, I added the implicit conversion for `list`.

python/gtsam/tests/test_custom_factor.py
ProfFan: This is a great idea. I just switched to your suggestion :)

gtsam/nonlinear/CustomFactor.cpp
ProfFan: Yes, done :)

gtsam/nonlinear/CustomFactor.h
ProfFan: done

gtsam/gtsam.i
ProfFan: done

gtsam/gtsam.i
```diff
@@ -2166,6 +2166,28 @@ virtual class NoiseModelFactor: gtsam::NonlinearFactor {
...
+   *        <calculate the Jacobian>
+   *        H[0] = J1 # 2-d numpy array for a Jacobian block
+   *        H[1] = J2
+   *        ...
+   *    return error # 1-d numpy array
+   *
+   * cf = CustomFactor(noise_model, keys, error_func)
+   * ```
+   */
+  CustomFactor(const gtsam::SharedNoiseModel& noiseModel, const gtsam::KeyVector& keys, const gtsam::CustomErrorFunction& errorFunction);
```
varunagrawal, line 2188: Format just this block?

gtsam/nonlinear/CustomFactor.h
```diff
@@ -0,0 +1,92 @@
...
+
+/*
+ * NOTE
+ * ==========
+ * pybind11 will invoke a copy if this is `JacobianVector &`, and modifications in Python will not be reflected.
+ *
+ * This is safe because this is passing a const pointer, and pybind11 will maintain the `std::vector` memory layout.
+ * Thus the pointer will never be invalidated.
+ */
+typedef std::function<Vector(const CustomFactor&, const Values&, const JacobianVector*)> CustomErrorFunction;
```
varunagrawal, line 38: use `using` instead of `typedef`?

gtsam/nonlinear/CustomFactor.h
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * This factor is mainly for creating a custom factor in Python.
+ */
+class CustomFactor: public NoiseModelFactor {
+public:
+   CustomErrorFunction errorFunction;
+
+protected:
+
+  typedef NoiseModelFactor Base;
```
varunagrawal, line 53: `using` instead of `typedef` for `Base` and `This`.

python/gtsam/specializations.h
```diff
@@ -2,9 +2,12 @@
 // These are required to save one copy operation on Python calls
 #ifdef GTSAM_ALLOCATOR_TBB
 py::bind_vector<std::vector<gtsam::Key, tbb::tbb_allocator<gtsam::Key> > >(m_, "KeyVector");
+py::implicitly_convertible<py::list, std::vector<gtsam::Key, tbb::tbb_allocator<gtsam::Key> > >();
 #else
 py::bind_vector<std::vector<gtsam::Key> >(m_, "KeyVector");
+py::implicitly_convertible<py::list, std::vector<gtsam::Key> >();
```
varunagrawal, line 8: TIL something new! Awesome.

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,125 @@
...
+from gtsam.utils.test_case import GtsamTestCase
+
+class TestCustomFactor(GtsamTestCase):
+    def test_new(self):
+        """Test the creation of a new CustomFactor"""
+        def error_func(this: CustomFactor, v: gtsam.Values, H: List[np.ndarray]):
+            return np.array([1, 0, 0])
+        
+        noise_model = gtsam.noiseModel.Unit.Create(3)
+        cf = CustomFactor(noise_model, gtsam.KeyVector([0]), error_func)
```
varunagrawal, line 27: Update the test to use the `list` or add another one to show both options?

gtsam/nonlinear/CustomFactor.h
```diff
@@ -0,0 +1,92 @@
...
+    this->errorFunction = errorFunction;
+  }
+
+  ~CustomFactor() override = default;
+
+  /** Calls the errorFunction closure, which is a std::function object
+    * One can check if a derivative is needed in the errorFunction by checking the length of Jacobian array
+  */
+  Vector unwhitenedError(const Values& x, boost::optional<std::vector<Matrix>&> H = boost::none) const override;
+
```
varunagrawal, line 80: Formatting, and also maybe override the `print` and `operator<<` methods?

varunagrawal (approved):
Looks good overall, some comments and needs formatting. Awesome job @ProfFan!

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,125 @@
...
+            from the outside scope. Or the variables can be acquired by indexing `v`.
+            Jacobian is passed by modifying the H array of numpy matrices.
+            """
+
+            key0 = this.keys()[0]
+            key1 = this.keys()[1]
+            gT1, gT2 = v.atPose2(key0), v.atPose2(key1)
+            error = Pose2(0, 0, 0).localCoordinates(gT1.between(gT2))
+            
+            if not H is None:
```
varunagrawal, line 66: `if H is not None` is the pythonic way. :slightly_smiling_face:

gtsam/nonlinear/CustomFactor.h
```diff
@@ -0,0 +1,92 @@
...
+#pragma once
+
+#include <gtsam/nonlinear/NonlinearFactor.h>
+
+using namespace gtsam;
+
+namespace gtsam {
+
+typedef std::vector<Matrix> JacobianVector;
```
varunagrawal, line 26: `using` instead of `typedef`

gtsam/nonlinear/CustomFactor.h
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @brief Custom factor that takes a std::function as the error
+ * @addtogroup nonlinear
+ * \nosubgrouping
+ *
+ * This factor is mainly for creating a custom factor in Python.
+ */
+class CustomFactor: public NoiseModelFactor {
+public:
+   CustomErrorFunction errorFunction;
```
varunagrawal, line 49: Should we make this `private` so that the error function passed in from python doesn't need `this` to be the first argument?

varunagrawal (approved):
Some more stuff I noticed and should have added in my last review.

varunagrawal:
@ProfFan can you please re-request a review after you make the changes? Overall it looks great but I want to do a thorough review again so I can start using this ASAP.

gtsam/nonlinear/CustomFactor.h
ProfFan: I think all our old factors use typedef, so let's keep using typedef here?

python/gtsam/tests/test_custom_factor.py
ProfFan: Sure

gtsam/nonlinear/CustomFactor.h
ProfFan: I think this is not relevant, since the callback will need the `this` pointer because one can reuse the same callback for multiple factors.

ProfFan:
I am thinking about adding an optional constructor param that allows one to specify the name of the custom factor (stored as a string), so people can distinguish different `CustomFactor` s.

dellaert (changes requested):
MArkdown?

gtsam/nonlinear/CustomFactor.h
varunagrawal: Yes but that's legacy. @dellaert mentioned to me that when we update/write new code, we should use the modern conventions. I can push changes to this PR if you're okay with that. :slightly_smiling_face:

gtsam/nonlinear/CustomFactor.h
varunagrawal: Wouldn't we then just pass the same function to different factors. I don't think I understand but would one factor refer to another factor's errorFunction explicitly? That kinda seems more error prone to me (e.g. we are indexing a factor list and we accidentally index the wrong factor).

gtsam/gtsam.i
```diff
@@ -2166,6 +2166,31 @@ virtual class NoiseModelFactor: gtsam::NonlinearFactor {
   Vector whitenedError(const gtsam::Values& x) const;
 };
 
+#include <gtsam/nonlinear/CustomFactor.h>
+virtual class CustomFactor: gtsam::NoiseModelFactor {
+  // Note CustomFactor will not be wrapped for MATLAB, as there is no supporting machinery there.
```
dellaert, line 2171: But, it's still in `gtsam.i`. Explain (in comment) how the wrapper code knows to not wrap this and make sure MATLAB wrapper still works.

gtsam/nonlinear/CustomFactor.cpp
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#include <gtsam/nonlinear/CustomFactor.h>
+
+namespace gtsam {
+
+/*
+ * Calculates the unwhitened error by invoking the callback functor (i.e. from Python).
+ */
+Vector CustomFactor::unwhitenedError(const Values& x, boost::optional<std::vector<Matrix>&> H) const {
```
dellaert, line 25: very long lines. Are you sure you're linting/formatting with Google style?

gtsam/nonlinear/CustomFactor.h
dellaert: I don't understand either why it's public, and why it does not obey naming convention (trailing underscore).

gtsam/nonlinear/CustomFactor.h
dellaert: Please use "using". @varunagrawal let's keep this PR Fan's task :-)

gtsam/nonlinear/CustomFactor.h
```diff
@@ -0,0 +1,113 @@
...
+  ~CustomFactor() override = default;
+
+  /**
+    * Calls the errorFunction closure, which is a std::function object
+    * One can check if a derivative is needed in the errorFunction by checking the length of Jacobian array
+    */
+  Vector unwhitenedError(const Values &x, boost::optional<std::vector<Matrix> &> H = boost::none) const override;
+
+  /** print */
+  void print(const std::string& s,
```
dellaert, line 83: Can we add a cpp file to have the implementation of this?
We can then also avoid including `<iostream>` here? A bit baffled it is not included and still works, probably because NonLinearFactor is templated and it has to include it.

python/FACTORS.md
```diff
@@ -0,0 +1,78 @@
+# GTSAM Python-based factors
```
dellaert, line 1: Rename to "CustomFactors.md"

python/FACTORS.md
```diff
@@ -0,0 +1,78 @@
...
+
+One now can build factors purely in Python using the `CustomFactor` factor.
+
+## Theory
```
dellaert, line 5: I'd move this to the end and call it "implementation", and add some more meat copied from your excellent comments in the source files

python/gtsam/preamble.h
dellaert: Make sure this much better explanation is copied to wrapper docs.

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,156 @@
...
+import gtsam
+from gtsam.utils.test_case import GtsamTestCase
+
+
+class TestCustomFactor(GtsamTestCase):
+    def test_new(self):
+        """Test the creation of a new CustomFactor"""
+
+        def error_func(this: CustomFactor, v: gtsam.Values, H: List[np.ndarray]):
+            return np.array([1, 0, 0])
```
dellaert, line 26: doc! Don't you use a linter?

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,156 @@
...
+
+        noise_model = gtsam.noiseModel.Unit.Create(3)
+        cf = CustomFactor(noise_model, [0], error_func)
+
+    def test_call(self):
+        """Test if calling the factor works (only error)"""
+        expected_pose = Pose2(1, 1, 0)
+
+        def error_func(this: CustomFactor, v: gtsam.Values, H: List[np.ndarray]) -> np.ndarray:
+            key0 = this.keys()[0]
```
dellaert, line 45: doc! Don't you use a linter?

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,156 @@
...
+        np.testing.assert_allclose(J_cf, J_bf)
+        np.testing.assert_allclose(b_cf, b_bf)
+
+    def test_printing(self):
+        """Tests if the factor result matches the GTSAM Pose2 unit test"""
+        gT1 = Pose2(1, 2, np.pi / 2)
+        gT2 = Pose2(-1, 4, np.pi)
+
+        def error_func(this: CustomFactor, v: gtsam.Values, _: List[np.ndarray]):
+            key0 = this.keys()[0]
```
dellaert, line 105: doc! Don't you use a linter?

python/gtsam/tests/test_custom_factor.py
```diff
@@ -0,0 +1,156 @@
...
+    def test_no_jacobian(self):
+        """Tests that we will not calculate the Jacobian if not requested"""
+
+        gT1 = Pose2(1, 2, np.pi / 2)
+        gT2 = Pose2(-1, 4, np.pi)
+
+        expected = Pose2(2, 2, np.pi / 2)
+
+        def error_func(this: CustomFactor, v: gtsam.Values, H: List[np.ndarray]):
+            key0 = this.keys()[0]
```
dellaert, line 129: doc! Don't you use a linter?

dellaert (changes requested):
Cool ! Starting to get there.

python/gtsam/tests/test_custom_factor.py
ProfFan: This is a callback, not a class method, so no linter warnings.

gtsam/nonlinear/CustomFactor.h
ProfFan: Yup, I moved it anyways.

gtsam/nonlinear/CustomFactor.cpp
ProfFan: Yes, it's still shorter than 80 :)

gtsam/gtsam.i
ProfFan: done

python/FACTORS.md
ProfFan: done

python/FACTORS.md
ProfFan: done

python/gtsam/preamble.h
ProfFan: done

gtsam/nonlinear/CustomFactor.h
ProfFan: done

gtsam/nonlinear/CustomFactor.h
ProfFan: done

gtsam/nonlinear/CustomFactor.h
ProfFan: fixed

ProfFan:
I have addressed all comments @dellaert  @varunagrawal

dellaert (changes requested):
After design review, need to expand example to trajopt, and think about autodiff (after that).

ProfFan:
Changes:
- Per @dellaert 's request, I added the example trajectory optimization with CustomFactor
- Unit test of optimizing an actual graph
- Fixed linearization of `CustomFactor` with TBB. Now it will not deadlock trying to acquire the GIL

python/gtsam/examples/CustomFactorExample.py
```diff
@@ -0,0 +1,154 @@
+import gtsam
+import numpy as np
+
+from typing import List, Optional
+from functools import partial
+
+# Simulate a car for one second
```
dellaert, line 7: Make a function for this that you call below.

python/gtsam/examples/CustomFactorExample.py
```diff
@@ -0,0 +1,154 @@
...
+I = np.eye(1)
+gps_model = gtsam.noiseModel.Isotropic.Sigma(1, sigma_gps)
+
+
+def error_gps(this: gtsam.CustomFactor, values, jacobians: Optional[List[np.ndarray]], measurement: np.ndarray):
+    """GPS Factor error function
+    :param this: gtsam.CustomFactor handle
+    :param values: gtsam.Values
+    :param jacobians: Optional list of Jacobians
+    :param measurement: GPS measurement, to be filled with `partial`
```
dellaert, line 55: Can you put measurement first? That makes it like “currying” in functionali languages.

python/gtsam/examples/CustomFactorExample.py
```diff
@@ -0,0 +1,154 @@
...
+    key = this.keys()[0]
+    estimate = values.atVector(key)
+    error = estimate - measurement
+    if jacobians is not None:
+        jacobians[0] = I
+
+    return error
+
+
+for k in range(5):
```
dellaert, line 67: Document what’s happening here

dellaert: Also, formatting?

dellaert: Finally, maybe assign factor to a variable

python/gtsam/examples/CustomFactorExample.py
```diff
@@ -0,0 +1,154 @@
...
+    factor_graph.add(gtsam.CustomFactor(gps_model, [unknown[k]], partial(error_gps, measurement=np.array([g[k]]))))
+
+v = gtsam.Values()
```
dellaert, line 70: Remove new line at end and Document block

python/gtsam/examples/CustomFactorExample.py
```diff
@@ -0,0 +1,154 @@
...
+
+for i in range(5):
+    v.insert(unknown[i], np.array([0.0]))
+
+params = gtsam.GaussNewtonParams()
+optimizer = gtsam.GaussNewtonOptimizer(factor_graph, v, params)
```
dellaert, line 76: Same

python/gtsam/examples/CustomFactorExample.py
```diff
@@ -0,0 +1,154 @@
...
+
+result = optimizer.optimize()
+
+error = np.array([(result.atVector(unknown[k]) - x[k])[0] for k in range(5)])
```
dellaert, line 80: What’s happening??? Document

python/gtsam/examples/CustomFactorExample.py
```diff
@@ -0,0 +1,154 @@
...
+# Adding odometry will improve things a lot
+odo_model = gtsam.noiseModel.Isotropic.Sigma(1, sigma_odo)
+
+
+def error_odom(this: gtsam.CustomFactor, values, jacobians: Optional[List[np.ndarray]], measurement: np.ndarray):
+    """Odometry Factor error function
+    :param this: gtsam.CustomFactor handle
+    :param values: gtsam.Values
+    :param jacobians: Optional list of Jacobians
+    :param measurement: Odometry measurement, to be filled with `partial`
```
dellaert, line 94: Measurement first?

dellaert (changes requested):
Great improvement. Not approving yet because we need to talk about sendable

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -134,6 +134,14 @@ class GTSAM_EXPORT NonlinearFactor: public Factor {
    */
   shared_ptr rekey(const KeyVector& new_keys) const;
 
+  /**
+   * Should the factor be evaluated in the same thread as the caller
+   * This is to enable factors that has shared states (like the Python GIL lock)
+   */
+   virtual bool sendable() const {
```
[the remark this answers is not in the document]
ProfFan: I shamelessly borrowed this concept from Rust ([here](https://doc.rust-lang.org/nomicon/send-and-sync.html)). Basically it indicates whether an object can be shared with another thread. Maybe `shareable` is a better name?

python/CustomFactors.md
```diff
@@ -0,0 +1,110 @@
...
+
+```python
+noise_model = gtsam.noiseModel.Unit.Create(3)
+# constructor(<noise model>, <list of keys>, <error callback>)
+cf = gtsam.CustomFactor(noise_model, [X(0), X(1)], error_func)
+```
+
+## Example
+
+The following is a simple `BetweenFactor` implemented in Python.
```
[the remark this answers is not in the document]
ProfFan: It's not zero now, if you look at the last unit test.

python/gtsam/examples/CustomFactorExample.py
ProfFan: done

python/gtsam/examples/CustomFactorExample.py
ProfFan: done

python/gtsam/examples/CustomFactorExample.py
ProfFan: done

python/gtsam/examples/CustomFactorExample.py
ProfFan: done

python/gtsam/examples/CustomFactorExample.py
ProfFan: done

python/gtsam/examples/CustomFactorExample.py
```diff
@@ -0,0 +1,154 @@
+import gtsam
```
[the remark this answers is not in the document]
ProfFan: fixed

python/gtsam/examples/CustomFactorExample.py
ProfFan: done

dellaert (approved):

ProfFan:
Merged! 🤯

PR #771, opened 2021-05-28
Correct ImuFactorExamples frame description

### LINKED ISSUE #768: integrateMeasurement() data frame unclear
## Description

In [CombinedImuFactorsExample.cpp](https://github.com/borglab/gtsam/blob/ab92779b25b04b376fbbd1846bbbd21904c50e7a/examples/CombinedImuFactorsExample.cpp#L26) it is stated that the IMU data is provided aligned to NED (the navigation frame).

However, the documentation for [gtsam::PreintegratedCombinedMeasurements::integrateMeasurement()](https://gtsam.org/doxygen/4.0.0/a03435.html#a9c9d6f736444eab5d49fa9a9b305c738) states that the IMU data should be provided in body frame.

Only one of these can be correct and I am uncertain which is.

HMellor:
Fixes #768

### DIFF
--- a/examples/CombinedImuFactorsExample.cpp
+++ b/examples/CombinedImuFactorsExample.cpp
@@ -23,7 +23,8 @@
  *  A row starting with "i" is the first initial position formatted with
  *  N, E, D, qx, qY, qZ, qW, velN, velE, velD
  *  A row starting with "0" is an imu measurement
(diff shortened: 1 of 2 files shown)


varunagrawal (approved):
LGTM. Thanks @HMellor!

PR #774, opened 2021-05-30
Fix docstrings for NonlinearEquality

varunagrawal:
Fixed a typo for `NonlinearEquality2`, and took this opportunity to modernize the docstrings as well.

@dellaert I did have a question about `NonlinearEquality` though. Its docstring seems incorrect in that it doesn't constrain two variables to be equal (this is done by `NonlinearEquality2`). Recommendation?

### DIFF
--- a/gtsam/nonlinear/NonlinearEquality.h
+++ b/gtsam/nonlinear/NonlinearEquality.h
@@ -59,10 +59,10 @@ private:
   double error_gain_;
 
   // typedef to this class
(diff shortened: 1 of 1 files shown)


dellaert (approved):

varunagrawal:
@dellaert thoughts on the docstrings and the naming of the factors?

dellaert:
Don't rename, but feel free to make docs better

PR #775, opened 2021-05-30
Fix initial guess for Cal3Bundler `calibrate`

### LINKED ISSUE #770: Cal3Bundler calibrate() jacobian unit test failure
## Description

The Jacobians of the calibrate() method of Cal3Bundler fail the numerical derivative unit test for certain calibration and input values. Specifically, they fail for the default value of Cal3Bundler i.e f = 1 with 0 distortion. 

## Steps to reproduce

I've added a couple of unit tests in a branch [here](https://github.com/borglab/gtsam/blob/f64ef4ed4239371ad7e6fea008a10e95587c8584/gtsam/geometry/tests/testCal3Bundler.cpp#L104). The unit test called `DcalibrateDefault` is the one that fails.  

```cpp
TEST(Cal3Bundler, DcalibrateDefault) {
  Cal3Bundler trueK;
  Matrix Dcal, Dp;
  Point2 pn(0.5, 0.5);
  Point2 pi = trueK.uncalibrate(pn);
  Point2 actual = trueK.calibrate(pi, Dcal, Dp);
  CHECK(assert_equal(pn, actual, 1e-7));
  Matrix numerical1 = numericalDerivative21(calibrate_, trueK, pi);
  Matrix numerical2 = numericalDerivative22(calibrate_, trueK, pi);
  CHECK(assert_equal(numerical1, Dcal, 1e-5));
  CHECK(assert_equal(numerical2, Dp, 1e-5));
}
```
```sh
Failure: 

not equal:
expected = [
        -0.5, 0, 0;
        -0.5, 0, 0
]
actual = [
        -0.5, -0.25, -0.125;
        -0.5, -0.25, -0.125
]
actual - expected = [
        5.046e-11, -0.25, -0.125;
        5.046e-11, -0.25, -0.125
]
```

## Environment

Linux

## Additional information

The calibrate method uses the implicit function theorem to compute the derivatives, which internally uses the derivative of uncalibrate(). Since the unit test for uncalibrate() passes, I think it should be the implementation of the implicit function theorem that is causing the issue. 

The calibrate() method for Cal3Bundler uses an optimization to compute the point in calibrated coordinates. The distortion for this case is zero, the optimization converges in the first iteration itself, as expected.

varunagrawal:
This PR fixes #770.

I simply incorporated the radial distortion from the iteration loop to the initial guess and the test passed. This makes sense since when we tweak the radial distortion parameters a bit for finite difference, we want to be able to see that effect in the output.

### DIFF
--- a/gtsam/geometry/Cal3Bundler.cpp
+++ b/gtsam/geometry/Cal3Bundler.cpp
@@ -99,18 +99,19 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, OptionalJacobian<2, 3> Dcal,
   double x = (pi.x() - u0_) / fx_, y = (pi.y() - v0_) / fx_;
   const Point2 invKPi(x, y);
 
-  // initialize by ignoring the distortion at all, might be problematic for
-  // pixels around boundary
-  Point2 pn(x, y);
-
+  // initialize pn with distortion included
+  double px = pi.x(), py = pi.y(), rr = px * px + py * py;
+  double g = (1 + k1_ * rr + k2_ * rr * rr);
+  Point2 pn = invKPi / g;
+  
   // iterate until the uncalibrate is close to the actual pixel coordinate
   const int maxIterations = 10;
   int iteration;
   for (iteration = 0; iteration < maxIterations; ++iteration) {
     if (distance2(uncalibrate(pn), pi) <= tol_) break;
-    const double px = pn.x(), py = pn.y(), xx = px * px, yy = py * py;
-    const double rr = xx + yy;
-    const double g = (1 + k1_ * rr + k2_ * rr * rr);
+    px = pn.x(), py = pn.y();
+    rr = (px * px) + (py * py);
+    double g = (1 + k1_ * rr + k2_ * rr * rr);
     pn = invKPi / g;
   }
 
--- a/gtsam/geometry/Cal3Bundler.cpp
+++ b/gtsam/geometry/Cal3Bundler.cpp
@@ -96,11 +96,9 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, OptionalJacobian<2, 3> Dcal,
                               OptionalJacobian<2, 2> Dp) const {
   // Copied from Cal3DS2
   // but specialized with k1, k2 non-zero only and fx=fy and s=0
-  double x = (pi.x() - u0_) / fx_, y = (pi.y() - v0_) / fx_;
-  const Point2 invKPi(x, y);
-
+  double px = (pi.x() - u0_) / fx_, py = (pi.y() - v0_) / fx_;
+  const Point2 invKPi(px, py);
   Point2 pn;
-  double px = pi.x(), py = pi.y();
 
   // iterate until the uncalibrate is close to the actual pixel coordinate
   const int maxIterations = 10;
--- a/gtsam/geometry/Cal3Bundler.cpp
+++ b/gtsam/geometry/Cal3Bundler.cpp
@@ -105,8 +105,8 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, OptionalJacobian<2, 3> Dcal,
   int iteration = 0;
   do {
     // initialize pn with distortion included
-    double rr = (px * px) + (py * py);
-    double g = (1 + k1_ * rr + k2_ * rr * rr);
+    const double rr = (px * px) + (py * py);
+    const double g = (1 + k1_ * rr + k2_ * rr * rr);
     pn = invKPi / g;
 
     if (distance2(uncalibrate(pn), pi) <= tol_) break;
--- a/gtsam/geometry/Cal3Bundler.cpp
+++ b/gtsam/geometry/Cal3Bundler.cpp
@@ -99,21 +99,26 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, OptionalJacobian<2, 3> Dcal,
   double x = (pi.x() - u0_) / fx_, y = (pi.y() - v0_) / fx_;
   const Point2 invKPi(x, y);
 
-  // initialize pn with distortion included
-  double px = pi.x(), py = pi.y(), rr = px * px + py * py;
-  double g = (1 + k1_ * rr + k2_ * rr * rr);
-  Point2 pn = invKPi / g;
-  
+  Point2 pn;
+  double px = pi.x(), py = pi.y();
+
   // iterate until the uncalibrate is close to the actual pixel coordinate
   const int maxIterations = 10;
-  int iteration;
-  for (iteration = 0; iteration < maxIterations; ++iteration) {
-    if (distance2(uncalibrate(pn), pi) <= tol_) break;
-    px = pn.x(), py = pn.y();
-    rr = (px * px) + (py * py);
+  int iteration = 0;
+  do {
+    // initialize pn with distortion included
+    double rr = (px * px) + (py * py);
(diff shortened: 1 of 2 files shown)


johnwlambert:
Nice work, thanks Varun

ProfFan (approved):

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -99,18 +99,19 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, OptionalJacobian<2, 3> Dcal,
...
   const Point2 invKPi(x, y);
 
-  // initialize by ignoring the distortion at all, might be problematic for
-  // pixels around boundary
-  Point2 pn(x, y);
-
+  // initialize pn with distortion included
+  double px = pi.x(), py = pi.y(), rr = px * px + py * py;
+  double g = (1 + k1_ * rr + k2_ * rr * rr);
+  Point2 pn = invKPi / g;
```
akshay-krishnan, line 105: invKPi is in normalized coordinates but g is in pixel (unnormalized image) coordinates here. Shouldnt px and py be initialized to x and y respectively?

akshay-krishnan: If you change that, it becomes as though writing the first iteration of the loop outside the loop. 
So can the loop can be changed to a do-while? (I am not a fan of do-while or anything, but maybe this is a valid use case? )

akshay-krishnan (changes requested):
Nice work catching that!

gtsam/geometry/Cal3Bundler.cpp
varunagrawal: I did think about using a do-while yesterday. Let me update.

varunagrawal: I believe `g` should be initialized using the pixel coordinates since we capture the radial distortions that way. If `g` was initialized to `x` and `y` (aka the intrinsic coordinates), we wouldn't see any effect of the radial distortion in the inverse process.

dellaert:
both, please test the hell out of this :-) Also, please don't merge until @akshay-krishnan and you are on the same wavelength.

gtsam/geometry/Cal3Bundler.cpp
varunagrawal: Great references. I agree that g should be defined using the intrinsic coordinates and not the image coordinates. I like your suggestion to use invKPi since that makes sense to me as per your references. I'll update the PR later tonight.

varunagrawal:
@akshay-krishnan I added all the tests from your branch `cal3bundler-calibrate-test`. Please delete the branch once we merge this? Or if there is other stuff missing, let's land it!

gtsam/geometry/Cal3Bundler.cpp
johnwlambert: @akshay-krishnan could you explain this point a bit more? I didn't quite follow

akshay-krishnan: John, could you check the references? They explain the pinhole camera model with distortion parameters. 
In this model, 

Pi = K * Pn_distorted where Pn_distorted = g * Pn_undistorted.
And g is a nonlinear function of Pn_undistorted. 

In calibrate() we are trying to find Pn_undistorted by optimization. The initial guess for Pn_undistorted = (invK * Pi) / g (using the above). But since g is a function of Pn_undistorted itself, we need an initial guess for g as well. The best we can do is initialize g using Pn_distorted ( = invK * Pi). 
I hope that makes sense.

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -99,20 +99,26 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, OptionalJacobian<2, 3> Dcal,
   double x = (pi.x() - u0_) / fx_, y = (pi.y() - v0_) / fx_;
   const Point2 invKPi(x, y);
 
-  // initialize by ignoring the distortion at all, might be problematic for
-  // pixels around boundary
-  Point2 pn(x, y);
+  Point2 pn;
+  double px = pi.x(), py = pi.y();
```
akshay-krishnan, line 103: Per our discussion in the other thread, this should be x, y right? I was basically suggesting the following: 

```
  double px = (pi.x() - u0_) / fx_, py = (pi.y() - v0_) / fx_;  
  const Point2 invKPi(px, py);  
  Point2 pn;
```
Then the do-while as you have below. 
It might be worth adding another unit test for a `Cal3Bundler K(5, 0, 0, 2, 2);` i.e non-zero `u0_`, `v0_`.

gtsam/geometry/Cal3Bundler.cpp
```diff
@@ -99,20 +99,26 @@ Point2 Cal3Bundler::calibrate(const Point2& pi, OptionalJacobian<2, 3> Dcal,
...
-  for (iteration = 0; iteration < maxIterations; ++iteration) {
-    if (distance2(uncalibrate(pn), pi) <= tol_) break;
-    const double px = pn.x(), py = pn.y(), xx = px * px, yy = py * py;
-    const double rr = xx + yy;
-    const double g = (1 + k1_ * rr + k2_ * rr * rr);
+  int iteration = 0;
+  do {
+    // initialize pn with distortion included
+    double rr = (px * px) + (py * py);
+    double g = (1 + k1_ * rr + k2_ * rr * rr);
```
akshay-krishnan, line 111: can we retain the `const`?

varunagrawal: `g` gets updated every iteration so it is not semantically `const`.

gtsam/geometry/Cal3Bundler.cpp
varunagrawal: Ah good catch. Sorry I missed that.

gtsam/geometry/Cal3Bundler.cpp
dellaert: That's BS ;-)
Make const rr and g const, they are declared inside scope where they are const.

varunagrawal: Is that how it works? I didn't know `const` was scope-dependent (and haven't seen an example of it anywhere).

gtsam/geometry/tests/testCal3Bundler.cpp
```diff
@@ -89,6 +89,33 @@ TEST(Cal3Bundler, DcalibrateDefault) {
...
+  CHECK(assert_equal(expected, actual, 1e-7));
+  Matrix numerical1 = numericalDerivative21(uncalibrate_, K, p);
+  Matrix numerical2 = numericalDerivative22(uncalibrate_, K, p);
+  CHECK(assert_equal(numerical1, Dcal, 1e-7));
+  CHECK(assert_equal(numerical2, Dp, 1e-7));
+}
+
+/* ************************************************************************* */
+TEST(Cal3Bundler, DcalibratePrincipalPoint) {
+  Cal3Bundler K(2, 0, 0, 2, 2);
```
akshay-krishnan, line 107: There’s also a global static variable by name K, just FYI in case you missed that. I don’t think it should cause any issues though.

akshay-krishnan (approved):
LGTM, thanks for fixing and for being patient with my delayed review. :)

gtsam/geometry/tests/testCal3Bundler.cpp
varunagrawal: Yeah I know about that. The local scope will take precedence so it's all good.

PR #777, opened 2021-05-30
Fix rekey of LinearContainerFactor

### LINKED ISSUE #568: LinearFactorContainer rekey()
Hi all, 

It appears that the LinearFactorContainer has a broken rekey member function (inherited from NonlinearFactor).  

## Description

Upon construction, the LinearFactorContainer sets its keys attribute (inherited from NonlinearFactor) and stores a shared pointer to the linear factor:
https://github.com/borglab/gtsam/blob/develop/gtsam/nonlinear/LinearContainerFactor.cpp#L31

Then when NonlinearFactor's rekey function is called, the NonlinearFactor keys attribute is modified without also rekeying the stored factor.

Note the L2 vs L4 key difference in the image below:

failed_rekey

## Steps to reproduce

1. Create a LinearFactorContainer
2. Call its rekey() member function

## Expected behavior

The LinearFactorContainer should rekey its stored factor.

## Environment

C++, Ubuntu 20.04, stock gtsam 4.1 commit: 76a30a208a967634bcd33b530e7e74a5cd6aa7f3

varunagrawal:
As per #568, `LinearContainerFactor`'s rekey method fails because the `rekey` method doesn't update the keys for the Gaussian Factor and Linearization point properties.

This PR fixes #568 by overriding `rekey` and doing the necessary updates.

### DIFF
--- a/gtsam/nonlinear/LinearContainerFactor.cpp
+++ b/gtsam/nonlinear/LinearContainerFactor.cpp
@@ -165,6 +165,47 @@ NonlinearFactor::shared_ptr LinearContainerFactor::negateToNonlinear() const {
 }
 
 /* ************************************************************************* */
+NonlinearFactor::shared_ptr LinearContainerFactor::rekey(
+    const std::map<Key, Key>& rekey_mapping) const {
+  auto rekeyed_base_factor = Base::rekey(rekey_mapping);
+  // Update the keys to the properties as well
+  // Downncast so we have access to members
+  auto new_factor = boost::static_pointer_cast<LinearContainerFactor>(rekeyed_base_factor);
+  // Create a new Values to assign later
+  Values newLinearizationPoint;
+  for (size_t i = 0; i < factor_->size(); ++i) {
+    auto mapping = rekey_mapping.find(factor_->keys()[i]);
+    if (mapping != rekey_mapping.end())
+      new_factor->factor_->keys()[i] = mapping->second;
+      newLinearizationPoint.insert(mapping->second, linearizationPoint_->at(mapping->first));
+  }
+  new_factor->linearizationPoint_ = newLinearizationPoint;
+
+  // upcast back and return
+  return boost::static_pointer_cast<NonlinearFactor>(new_factor);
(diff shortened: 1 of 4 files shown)


gtsam/nonlinear/LinearContainerFactor.cpp
```diff
@@ -164,6 +164,47 @@ NonlinearFactor::shared_ptr LinearContainerFactor::negateToNonlinear() const {
...
+      newLinearizationPoint.insert(mapping->second, linearizationPoint_->at(mapping->first));
+  }
+  new_factor->linearizationPoint_ = newLinearizationPoint;
+
+  // upcast back and return
+  return boost::static_pointer_cast<NonlinearFactor>(new_factor);
+}
+
+/* ************************************************************************* */
+NonlinearFactor::shared_ptr LinearContainerFactor::rekey(
```
ProfFan, line 189: Maybe we can just reuse the method for `std::map` here?

ProfFan (approved):
LGTM

dellaert:
I’m ok if this fixes a bug , but I do want to note, for the record, that “are-keying” goes against the functional nature of GTSAM. I question the need for it and would recommend people using this to re-examine why they’re using it :-) Try to be functional unless it is *really* affecting performance.

gtsam/nonlinear/LinearContainerFactor.cpp
varunagrawal: Can you please explain more?

varunagrawal:
@pshea76 check out @dellaert's comment, that's great advice right there.

gtsam/nonlinear/LinearContainerFactor.cpp
varunagrawal: Oh I understand now. I don't think that is doable, because the `std::map` version is doing a replacement, while the other one is doing an update. The assignment in the `KeyVector` version makes it cleaner and I am not sure how we would create the map without making some assumptions about what the user intended.

PR #779, opened 2021-05-31
Added destructor for `CameraSet` to remove warning

varunagrawal:
Just the title.


ProfFan (approved):
LGTM, though I think maybe we need to rethink whether to keep inheriting `std` containers...

varunagrawal:
Yeah true. This is a legacy of some previous fixes I made where the destructors needed to be explicitly added to remove warnings. Hopefully this is the last of them.

Issue #746, opened 2021-04-19
SO3 matrix log when theta is close to zero


yiping:
https://github.com/borglab/gtsam/blob/43e8f1e5aeaf11890262722c1e5e04a11dbf9d75/gtsam/geometry/SO3.cpp#L279-L281

Should line 281 be

`magnitude = 0.5 - tr_3 / 12.0;`

instead?

tmcg0:
This is probably worth someone from the GTSAM team getting their eyes on. In practice this wouldn't affect the computation much because abs(tr_3) is always <1e-7. But there does seem to be an inconsistency between the comment at line 280 and the code at 281. I'll try to track down what it should be and post here.

tmcg0:
original function for magnitude is:
`magnitude = f(theta) = theta/(2.0*sin(theta))` # See [Blanco](https://jinyongjeong.github.io/Download/SE3/jlblanco2010geometry3d_techrep.pdf).

taylor series expansion of f(theta) at theta=0 should be:
1/2 + theta^2/12 + 7*theta^4/720 + H.O.T.

I'm a little bit rusty and unsure what's happening in the code. Is it that for small theta we're assuming that tr(R) ~ theta? And that's why we do this variable swap between `tr_3` and `theta`?

dellaert:
OK, so, the angle `theta` is a function of the trace *t*, which ranges from -1 to 3. Here is [Wolfram alpha's plot](https://www.wolframalpha.com/input/?i=acos%28%28t+-+1.0%29+%2F+2.0%29+):

![image](https://user-images.githubusercontent.com/10515273/120030639-b91bc700-bfc5-11eb-959c-f4f3faf12bce.png)

We are interested in getting the Taylor series of `theta/(2.0*sin(theta))` as you said, but in *t* (or *t-3*). For the life of me I can't remember how I got the Taylor of magnitude expansion in t. Maybe that's a fruitful direction to go? Both comment and code might be wrong.

ProfFan:
I remember doing this math when dealing with Cayley and stuff long long time ago. What I remember is that it's because the original form cannot be evaluated at tr_3 = 0, so one needs to not use the formula in first branch, but the original form to devise this taylor expansion.

tmcg0:
Glad I'm not the only rusty one here :)

So here's a snippet from Blanco:
![image](https://user-images.githubusercontent.com/31974575/120036742-36473c00-bfc6-11eb-8b96-388f88914a79.png)

[...]

PR #780, opened 2021-06-01
Bugfix: correct SO(3) logmap when theta small

tmcg0:
This PR fixes https://github.com/borglab/gtsam/issues/746. 

When mod(theta, 2pi) is small, the SO(3) logmap has to be handled specially (to avoid numerical problems) using the Taylor series expansion. Previously the code which implemented this first-order Taylor series expansion was incorrect (although the comment above it was correct). This PR corrects the line of code, as well as adds a comment above directing the user to the related issue (https://github.com/borglab/gtsam/issues/746), where the user can find the derivation of the equation.


varunagrawal (approved):
LGTM!

PR #783, opened 2021-06-04
Added Range to Point2

dellaert:
Added range function object to Point2 and added unit test in testRangeFactor.h to show usage.


ProfFan (approved):
LGTM

ProfFan:
I'll merge and delete branch since this is a minor patch.

PR #785, opened 2021-06-06
Fix bug in LinearContainerFactor and warnings about Point3

varunagrawal:
Just the title.


ProfFan (approved):
LGTM, but need a test case that captures this bug.

varunagrawal:
Thanks for the reminder @ProfFan. I added a simple test case which exposes the bug. Will merge once CI passes.

PR #752, opened 2021-04-23
Add a MagPoseFactor for using magnetometer measurements with Pose2/Pose3

miloknowles:
This PR adds a new `MagPoseFactor<POSE>` which extends the existing `MagFactor` to work with `Pose2` and `Pose3`. The factor also includes an optional `body_P_sensor` in case the magnetometer does not coincide with the body frame. I implemented this factor so that I could use magnetometer measurements within a VIO system, and thought it might be useful to others. It doesn't seem like the `MagFactor` gets much use, so no worries if you don't think this factor should be added.

PR #788, opened 2021-06-08
Move MagPoseFactor into gtsam

miloknowles:
As discussed in #752 with @varunagrawal, this PR moves `MagPoseFactor` out of unstable and into `gtsam`.


tmcg0:
Thanks for doing this @miloknowles! Excited to use it in my own work.

... I don't know why I thought it was going to be tougher to migrate this factor from unstable -> stable. (I assumed there were some python bindings that had to be edited too, but I suppose not)

Also looks like the xcode debug CI didn't want to play nice :(

miloknowles:
Yeah it looks like something went wrong with the CI ... does anyone know how to re-run tests?

dellaert (approved):

dellaert:
Thanks @miloknowles !!!

PR #791, opened 2021-06-10
Add transformFrom() to wrapper, for Point3 in Similarity3

johnwlambert:
Expose via the python wrapper `transformFrom()` for the `Point3` type. Previously only `transformFrom()` for the `Pose3` type was exported.


johnwlambert:
Thanks for the review @ProfFan

PR #792, opened 2021-06-10
Switching to squared sampson point-line error

ayushbaid:
The squared sampson error is the first order approximation of the gold-standard reprojection error. 

This is a work-in-progress PR, and I'll see if the switch actually improves performance. I will clean up the PR and add more documentation if we want to merge this.

### DIFF
--- a/gtsam/geometry/EssentialMatrix.cpp
+++ b/gtsam/geometry/EssentialMatrix.cpp
@@ -101,17 +101,15 @@ EssentialMatrix EssentialMatrix::rotate(const Rot3& cRb,
 }
 
 /* ************************************************************************* */
-double EssentialMatrix::error(const Vector3& vA, const Vector3& vB, //
-    OptionalJacobian<1, 5> H) const {
-
+double EssentialMatrix::error(const Vector3& vA, const Vector3& vB,  //
+                              OptionalJacobian<1, 5> H) const {
   // compute the epipolar lines
   Point3 lB = E_ * vB;
   Point3 lA = E_.transpose() * vA;
 
-
   // compute the algebraic error as a simple dot product.
-  double algebraic_error = dot(vA, lB); 
-  
+  double algebraic_error = dot(vA, lB);
+
   // compute the line-norms for the two lines
   Matrix23 I;
   I.setIdentity();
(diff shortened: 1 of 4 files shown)


ProfFan:
Wait, is this a WIP PR? @akshay-krishnan @ayushbaid ?

ProfFan:
Also this is a breaking PR.

akshay-krishnan:
Fan, yes this a WIP PR. We merged a portion of it in another PR. I am not sure how Ayush wants to proceed with this, I think we will open a new PR if we want to merge the remaining changes.

ProfFan:
Yeah what I mean is that this will change how old code behaves or not?

dellaert:
It seems this was merged in develop, not in another branch, so Fan questions are pertinent :-) Phan, why do you say this is a breaking PR? CI seems to work.

ProfFan:
@dellaert  Because
```c++
double EssentialMatrix::error(const Vector3& vA, const Vector3& vB, 
                              OptionalJacobian<1, 5> HE,
                              OptionalJacobian<1, 3> HvA,
                              OptionalJacobian<1, 3> HvB) const {
```
signature has changed, and
```
*HE = 2 * sampson_error * (numerator_H / denominator -
         algebraic_error * denominator_H / (denominator * denominator));
```
The first Jacobian has changed as well.

akshay-krishnan:
Fan, those changes have not been merged into develop. [Here](https://github.com/borglab/gtsam/blob/91674c80083da612fc724a65a917416bb5cda53d/gtsam/geometry/EssentialMatrix.cpp#L104) is the error() method in EssentialMatrix.cpp at develop. 
Ayush, merged some commits from this branch feature/sampson-epipolar-error into feature/essential-mat-with-approx-k which is what #753 used. But I removed those changes in [this commit](https://github.com/borglab/gtsam/pull/753/commits/14f8b8aa6279e5d407e46282f7ee82d044129adc) before merging #753. 

So git will say that this PR has been merged since its commits went into #753 but I removed most of the changes in those commits later on.

ProfFan:
@akshay-krishnan I see, thanks :)

PR #793, opened 2021-06-11
add expressions for cross() and dot()

jlblancoc:
Add `cross()` and `dot()` as expressions. Added the corresponding unit tests.


dellaert (approved):
Awesome!

PR #796, opened 2021-06-14
update boost::bind usage

acxz:
use <boost/bind/bind.hpp> instead of deprecated <boost/bind.hpp>

use boost::placeholders:: scope instead of global scope

remove and add <boost/bind/bind.hpp> in appropriate files

Essentially with recent boost versions, the usage of `<boost/bind.hpp>` is deprecated and `<boost/bind/bind.hpp>` along with using the above mentioned symbols in the `boost::placeholders` scope is recommended. This PR replaces the deprecated functionality throughout the repo.

Related warning messages look like the following:
```
/usr/include/boost/bind.hpp:36:1: note: ‘#pragma message: The practice of declaring the Bind placeholders (_1, _2, ...) in th     e global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIN     D_GLOBAL_PLACEHOLDERS to retain the current behavior.’                                                                       
   36 | BOOST_PRAGMA_MESSAGE(                                                     
      | ^~~~~~~~~~~~~~~~~~~~                  
```

Let me know if you want me to break it up into several smaller commits.


ProfFan (approved):
LGTM, thanks!

gtsam/base/numericalDerivative.h
```diff
@@ -19,20 +19,15 @@
 #pragma once
 
 #include <boost/function.hpp>
-#ifdef __GNUC__
```
varunagrawal, line 22: Any reason this was removed?

gtsam/linear/HessianFactor.cpp
```diff
@@ -32,14 +32,6 @@
 #include <boost/format.hpp>
 #include <boost/make_shared.hpp>
 #include <boost/tuple/tuple.hpp>
-#ifdef __GNUC__
```
varunagrawal, line 35: Same as above

gtsam/linear/JacobianFactor.cpp
```diff
@@ -35,14 +35,6 @@
 #include <boost/format.hpp>
 #include <boost/make_shared.hpp>
 #include <boost/array.hpp>
-#ifdef __GNUC__
```
varunagrawal, line 38: Same as above

gtsam/navigation/tests/testAHRSFactor.cpp
```diff
@@ -174,17 +174,22 @@ TEST(AHRSFactor, Error) {
 
   // Expected Jacobians
   Matrix H1e = numericalDerivative11<Vector3, Rot3>(
-      boost::bind(&callEvaluateError, factor, _1, x2, bias), x1);
+      boost::bind(&callEvaluateError, factor, boost::placeholders::_1, x2,
```
varunagrawal, line 177: Since you're using `boost::placeholders` in multiple places, you should use `using namespace boost::placeholders`. This should also reduce the diff size.

varunagrawal (changes requested):
What version of boost is this for? I don't seem to get this issue, and we don't want to break anything by adding features for a more recent version of boost.

gtsam/base/numericalDerivative.h
acxz: This was removed as I believe it was first introduced to surpress warnings from the previous `<boost/bind.hpp>` header. As we are not using that header, we can remove this `ifdef`.

gtsam/linear/HessianFactor.cpp
acxz: This was removed as I believe it was first introduced to surpress warnings from the previous `<boost/bind.hpp>` header. As we are not using that header, we can remove this `ifdef`.

gtsam/base/numericalDerivative.h
acxz: Even if we do use the older include, I believe the later versions of boost have fixed the associated warnings so there is no need to suppress them.

gtsam/linear/HessianFactor.cpp
acxz: Even if we do use the older include, I believe the later versions of boost have fixed the associated warnings so there is no need to suppress them.

gtsam/linear/JacobianFactor.cpp
acxz: This was removed as I believe it was first introduced to surpress warnings from the previous <boost/bind.hpp> header. As we are not using that header, we can remove this ifdef.

acxz: Even if we do use the older include, I believe the later versions of boost have fixed the associated warnings so there is no need to suppress them.

gtsam/navigation/tests/testAHRSFactor.cpp
acxz: `using namespace ...` should be used as little as possible as it clutters the scope. Even then it's recommend to never use the `using` directive (See: https://google.github.io/styleguide/cppguide.html#Namespaces).

I deemed that `numericalDerivative.h` had a large number of uses to justify it, whereas for the other files I did not believe it was necessary.

In fact, I would rather spend the time to change `numericalDerivative.h` to not have `using namespace` than the other way around. Let me know what you still think, I'll add `using namespace` if you want.

gtsam/linear/JacobianFactor.cpp
varunagrawal: You don't need to respond to each of these comments regarding the same issue :) We use this review style to mark areas so we don't miss it if it needs updating.

gtsam/navigation/tests/testAHRSFactor.cpp
varunagrawal: What is your definition of large number of uses? For us, it is generally 4 or more.

In that case, you should use `using boost:: placeholder::_1` etc since it avoids the issues of namespace pollution and other negative aspects. This is an acceptable pattern.

acxz:
> What version of boost is this for? I don't seem to get this issue, and we don't want to break anything by adding features for a more recent version of boost.

I've been seeing these warnings since boost 1.73. (https://github.com/boostorg/bind/commit/5c21b6a5251523453ef3d02cc7e44dbfb31275eb) I'm on 1.76 rn.

The CI passes so it doesn't break anything.

gtsam/navigation/tests/testAHRSFactor.cpp
acxz: I still think with `using boost::placeholder::_1` at the global scope you are polluting with `_1`. With `using` I believe it is recommended to use it locally in methods and not at the file level. At that point it should reduce the diff size a good bit.

I'll go ahead and do that and we can see if the diff size in appropriate.
(Sorry I'm being so adamant on not using `using` ;))

varunagrawal: It's not polluting the namespace since we're using that specific variable and only accessing that one variable.
It would be namespace pollution if other, unused, variables were being included into the namespace.

acxz:
did some digging and `<boost/bind/bind.hpp>` has been around since [`1.39`](https://github.com/boostorg/bind/blob/boost-1.39.0/include/boost/bind/bind.hpp), which is lower than what we required (`1.43`).

gtsam/navigation/tests/testAHRSFactor.cpp
acxz: Ah maybe pollution was the wrong word to use there. I agree with what you are describing, but there is the case, where if we `using boost::placeholder::_1` say in a header file, than `_1` becomes associated with `boost::placeholder::_1` and you can't use a local `_1`.

acxz: I've updated the PR, where for tests I am adding `using namespace boost::placeholders;` as those are the end of the compilation pipeline. But for the other files, I've kept them as is. In fact, I doubled down on my stance for `numericalDerivative.h` and added the `using namespace boost::placeholders;` in the local scopes of the methods.

ProfFan:
Looks good now I think

ProfFan:
@acxz The macOS CI is failing.

acxz:
Yep on it, forgot to compile the tests, going through some more files and fixing them up.

varunagrawal:
Please don't force push on public branches. :)

gtsam/navigation/tests/testAHRSFactor.cpp
varunagrawal: Will we ever use a local `_1` in any context? We prefer to use descriptive names so I don't see that being a problem. I still don't agree with your use of the full namespace everywhere, even in headers, the reasons being that the placeholders are not part of the public API (especially if you do `using boost::placeholders::_1;` inside the `gtsam` namespace) and the use of `using boost::placeholders::_1;` is a common pattern we have used in other code bases in the lab. There is little reason to make the code more verbose for the sake of being pedantic.

Also, you have included the header in test files which don't use the placeholder, and not included in test files that do. What's going on there?

acxz: > especially if you do `using boost::placeholders::_1` inside the gtsam namespace

I think this is fine, if done in cpp files, but should still not be done in header files. Here is a source on why: Page 109 [here](http://library.bagrintsev.me/CPP/Sutter.C%2B%2B%20Coding%20Standards.2005.pdf)

> Also, you have included the header in test files which don't use the placeholder, and not included in test files that do. What's going on there?

Can you point to these specifically? I thought I had added and removed the header appropriately, I guess I messed up for some of them. (Surprised the CI passes)

varunagrawal: Again, no reason to be pendantic about the standards if it makes the code unnecessarily verbose. We prefer readability over everything else. And as I mentioned above, the placeholders are not used as part of any public API.

gtsam/geometry/tests/testPinholeCamera.cpp
```diff
@@ -22,11 +22,13 @@
 #include <gtsam/base/Testable.h>
 #include <gtsam/base/numericalDerivative.h>
 
+#include <boost/bind/bind.hpp>
 #include <CppUnitLite/TestHarness.h>
 
 #include <cmath>
 #include <iostream>
 
+using namespace boost::placeholders;
```
varunagrawal, line 31: This is an example @acxz

acxz: It is used right here: https://github.com/borglab/gtsam/pull/796/files/650e432f52f58505eba8c8f6a918f03e733a7acc#diff-55c680128d3251f02bf918b24cbe80c0b30fe8ff56ecda7f187d6bb3f5fdbde2R70

acxz: Link doesn't really work that well, line 70 of the new file.

gtsam/navigation/tests/testAHRSFactor.cpp
acxz: > the placeholders are not used as part of any public API

I'm a bit confused about this. What are you referring to as the public API?

acxz: Are you referring to the header files users have access to? If so `gtsam/base/numericalDerivative.h` is a public header file.

Here is a minimal example, where the user would be frustrated.

exe.cpp
```
#include <gtsam/base/numericalDerivative.h>

#include <functional>
#include <iostream>
#include <typeinfo>

using namespace std::placeholders;
using namespace gtsam;

int main() { std::cout << typeid(_1).name() << std::endl; }
```

Having `using boost::placeholders::_1;` inside of the gtsam namespace in the header file of `numericalDerivative.h`.
Output:
```
/home/acxz/test-gtsam/exe.cpp:10:34: error: reference to ‘_1’ is ambiguous
   10 | int main() { std::cout << typeid(_1).name() << std::endl; }
      |                                  ^~
In file included from /usr/include/boost/bind/bind.hpp:2336,
                 from /usr/local/include/gtsam/base/numericalDerivative.h:22,
                 from /home/acxz/test-gtsam/exe.cpp:1:
/usr/include/boost/bind/placeholders.hpp:46:38: note: candidates are: ‘constexpr const boost::arg<1> boost::placeholders::_1’
   46 | BOOST_INLINE_CONSTEXPR boost::arg<1> _1;
      |                                      ^~
In file included from /usr/include/boost/function.hpp:23,
                 from /usr/local/include/gtsam/base/numericalDerivative.h:21,
                 from /home/acxz/test-gtsam/exe.cpp:1:
/usr/include/c++/11.1.0/functional:225:34: note:                 ‘const std::_Placeholder<1> std::placeholders::_1’
  225 |     extern const _Placeholder<1> _1;
      |                                  ^~
```

Having the `using` directive inside the scope of functions in the header file of `numericalDerivative.h`.
Output:
```
St12_PlaceholderILi1EE
```

As you can see the current change works as expected and returns the proper type, `std::placeholders::_1`. The first output gives us an ambiguous usage of `_1` and fails to compile.

gtsam/base/numericalDerivative.h
```diff
@@ -176,13 +169,14 @@ typename internal::FixedSizeMatrix<Y,X1>::type numericalDerivative21(const boost
...
       "Template argument X1 must be a manifold type.");
-  return numericalDerivative11<Y, X1, N>(boost::bind(h, _1, boost::cref(x2)), x1, delta);
+  return numericalDerivative11<Y, X1, N>(boost::bind(h, boost::placeholders::_1, boost::cref(x2)), x1, delta);
 }
 
 /** use a raw C++ function pointer */
 template<class Y, class X1, class X2>
 typename internal::FixedSizeMatrix<Y,X1>::type numericalDerivative21(Y (*h)(const X1&, const X2&), const X1& x1,
     const X2& x2, double delta = 1e-5) {
+  using namespace boost::placeholders;
```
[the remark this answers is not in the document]
acxz: The only variables in `boost::placeholders` are the `_X` where `X` goes from 1 to 9. [Link](https://github.com/boostorg/bind/blob/develop/include/boost/bind/placeholders.hpp)

As we use multiple `_X`, I thought it would more readable to use `boost::placeholders` instead of manually specifying each `boost::placeholders::_X`.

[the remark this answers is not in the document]
ProfFan: I think both are fine, and I personally prefer `using xxx::_1` in rarely changed code, and `using ns` in scripts.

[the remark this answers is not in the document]
acxz: @ProfFan can you clarify your preference concerning `using` in header files, specifically in an existing namespace scope vs a local function scope?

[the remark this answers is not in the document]
varunagrawal: @acxz yes I am aware. My recommendation is to use `using boost::placeholders::_1;` and `using boost::placeholders::_2;` etc at the top level of the namespace rather than including the full namespace via a `using namespace` statement.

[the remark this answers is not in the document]
acxz: So then is this the same concern as the previous comment chain? If so can we close this and just keep the conversation to one comment chain?

If you want me to use `using` instead of `using namespace` I can go ahead and do that, but do note that I'll be doing that inside of every method. I would prefer we resolve the first concern first before.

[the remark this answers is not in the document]
varunagrawal: Sorry no, having a `using` inside every method is not a good idea for reasons I've mentioned before.

gtsam/navigation/tests/testAHRSFactor.cpp
varunagrawal: The user can then simply remove the first `using namespace boost::placeholders;` and things should work no problem. I don't really see what the problem is here.

Moreover, the `develop` branch currently uses `_1` and this has not been a problem all this time.

gtsam/geometry/tests/testPinholeCamera.cpp
varunagrawal: Which new file? The diff for this file doesn't show anything else.

gtsam/navigation/tests/testAHRSFactor.cpp
acxz: > Moreover, the develop branch currently uses _1 and this has not been a problem all this time.

We actually get the same error with the current code.

> The user can then simply remove the first `using namespace boost::placeholders`;

The user's code does not have the line `using namespace boost::placeholders`; They are using `std::placeholders`.

gtsam/geometry/tests/testPinholeCamera.cpp
acxz: Sorry about that wasn't clear. Line 68 of the current `gtsam/geometry/tests/testPinholeCamera.cpp` or Line 70 of the same file with my commit applied.

https://github.com/acxz/gtsam/blob/650e432f52f58505eba8c8f6a918f03e733a7acc/gtsam/geometry/tests/testPinholeCamera.cpp#L70

gtsam/navigation/tests/testAHRSFactor.cpp
varunagrawal: The error has not been reported, because the placeholder is not a part of the public API.

Well, you use std::placeholders and then mention boost::placeholders in the following comment. Either way I don't see how that makes a difference.

varunagrawal: The issue with your example is that you have a `using namespace gtsam` which directly contradicts your previous arguments. The user can simply use gtsam::X for whatever they need and issue resolved.

acxz: > the placeholder is not a part of the public API.

I'm still a bit confused by this. By public API do you mean the header files that users have access to?

> Boost or std doesn't matter right? They can either use the boost one,

I'm actually not sure about that. There might be differences in std::bind and boost::bind, where it does matter, there might not be. (I personally feel like it doesn't.) The user will most likely err on the safer side if they are using `std::bind` and would want to use `std::placeholders::`.

> fully qualify the std one
> The user can simply use gtsam::X for whatever they need and issue resolved.

The user should not be at fault for having to use `using namespace` in `.cpp` files, especially code that is at the end of the compilation units, such as this example or tests. It is up to the libraries included to ensure included namespaces do not cause pollution or collisions. It is one thing if the user is also using `using boost::placeholders` along with `using std::placeholders` that's on them to fix the collision. But it is another thing if the collision error is coming from a library that they do not have control over. Sure they can fix the problem on their end, but they should not have experienced the problem in the first place if the library is created properly.

varunagrawal: > By public API do you mean the header files that users have access to?

By public API I mean that they will never need to use `_1` when calling a function in GTSAM. So they never have to deal with it wrt GTSAM code. The public API is any function that can be called by the user.

> The user should not be at fault for having to use using namespace in .cpp files, especially code that is at the end of the compilation units, such as this example or tests. It is up to the libraries included to ensure included namespaces do not cause pollution or collisions. It is one thing if the user is also using using boost::placeholders along with using std::placeholders that's on them to fix the collision. But it is another thing if the collision error is coming from a library that they do not have control over. Sure they can fix the problem on their end, but they should not have experienced the problem in the first place if the library is created properly.

Then the user can just be more specific and declare `using std::placeholders::_1;`. Again, I don't see what the problem is. We are not restricting them in any way.

acxz: > they will never need to use _1 when calling a function in GTSAM

Can they not use the functions in numericalDerivatives which you can pass placeholder arguments?

> Then the user can just be more specific and declare `using std::placeholders::_1;` This is the same as `using namespace std::placeholders;`. The `_1` is imported in both cases, leading to the same collision problem.

It seems that we are at a standstill. I suggest we wait for another opinion on the matter, before taking this conversation further.

gtsam/base/numericalDerivative.h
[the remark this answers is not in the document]
dellaert: OK, seems a whole argument developed here :-) I am in favor of explicitly using `boost::placeholders::_1` and 1`boost::placeholders::_2` in this file. The google style guide explicitly forbids importing namespaces, and it's a bad idea in header files, *even* inside a scope like above. The reason is (apparently) that some (microsoft!) compilers don't respect the scope.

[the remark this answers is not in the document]
ProfFan: Second Frank here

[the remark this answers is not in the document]
varunagrawal: Okay, then we should be consistent since GTDynamics doesn't explicitly follow this. :)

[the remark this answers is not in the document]
acxz: > some (microsoft!) compilers don't respect the scope

yikes that is truly unfortunate. I'll add a commit that explicitly uses `boost::placeholders::_X` in the header files.

gtsam/geometry/tests/testPinholeSet.cpp
```diff
@@ -20,7 +20,6 @@
 #include <gtsam/geometry/Pose3.h>
 #include <gtsam/base/numericalDerivative.h>
 #include <CppUnitLite/TestHarness.h>
-#include <boost/bind.hpp>
```
varunagrawal, line 23: I admire your thoroughness @acxz!

gtsam/linear/tests/testGaussianBayesNet.cpp
```diff
@@ -26,6 +26,8 @@
 #include <boost/assign/list_of.hpp>
 #include <boost/assign/std/list.hpp> // for operator +=
 using namespace boost::assign;
+#include <boost/bind/bind.hpp>
+using namespace boost::placeholders;
```
varunagrawal, line 30: formatting?

gtsam/linear/tests/testGaussianBayesTree.cpp
```diff
@@ -22,6 +22,8 @@
 #include <boost/assign/std/list.hpp> // for operator +=
 #include <boost/assign/std/set.hpp> // for operator +=
 using namespace boost::assign;
+#include <boost/bind/bind.hpp>
+using namespace boost::placeholders;
```
varunagrawal, line 26: formatting?

gtsam/navigation/tests/testManifoldPreintegration.cpp
```diff
@@ -98,7 +98,9 @@ TEST(ManifoldPreintegration, computeError) {
   pim.computeError(x1, x2, bias, aH1, aH2, aH3);
   boost::function<Vector9(const NavState&, const NavState&,
                           const imuBias::ConstantBias&)> f =
-      boost::bind(&ManifoldPreintegration::computeError, pim, _1, _2, _3,
+      boost::bind(&ManifoldPreintegration::computeError, pim,
```
varunagrawal, line 101: It would be nice to have consistency amongst tests in the use of `using namespace boost::placeholders;` but whatever.

gtsam/navigation/tests/testTangentPreintegration.cpp
```diff
@@ -105,7 +105,9 @@ TEST(TangentPreintegration, computeError) {
   pim.computeError(x1, x2, bias, aH1, aH2, aH3);
   boost::function<Vector9(const NavState&, const NavState&,
                           const imuBias::ConstantBias&)> f =
-      boost::bind(&TangentPreintegration::computeError, pim, _1, _2, _3,
+      boost::bind(&TangentPreintegration::computeError, pim,
```
varunagrawal, line 108: Similar comment on consistency.

gtsam/nonlinear/NonlinearEquality.h
```diff
@@ -87,7 +87,8 @@ class NonlinearEquality: public NoiseModelFactor1<VALUE> {
    * Constructor - forces exact evaluation
    */
   NonlinearEquality(Key j, const T& feasible,
-      const CompareFunction &_compare = boost::bind(traits<T>::Equals,_1,_2,1e-9)) :
+      const CompareFunction &_compare = boost::bind(traits<T>::Equals,
```
varunagrawal, line 90: Format just the edited sections? This is pretty easy with VS code.

gtsam/nonlinear/tests/testValues.cpp
```diff
@@ -29,6 +29,8 @@
 #include <boost/assign/std/vector.hpp>
 #include <boost/assign/list_of.hpp>
 using namespace boost::assign;
+#include <boost/bind/bind.hpp>
```
varunagrawal, line 32: formatting?

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
```diff
@@ -420,6 +423,8 @@ class EquivInertialNavFactor_GlobalVel : public NoiseModelFactor5<POSE, VELOCITY
     // Note: all delta terms refer to an IMU\sensor system at t0
     // Note: Earth-related terms are not accounted here but are incorporated in predict functions.
 
+    using namespace boost::placeholders;
```
varunagrawal, line 426: remove this?

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
```diff
@@ -437,18 +442,18 @@ class EquivInertialNavFactor_GlobalVel : public NoiseModelFactor5<POSE, VELOCITY
...
+    Matrix H_vel_bias = numericalDerivative11<Vector3, IMUBIAS>(boost::bind(&PreIntegrateIMUObservations_delta_vel, msr_gyro_t, msr_acc_t, msr_dt, delta_angles, delta_vel_in_t0, flag_use_body_P_sensor, body_P_sensor, boost::placeholders::_1), Bias_t0);
     Matrix H_vel_pos = Z_3x3;
 
-    Matrix H_angles_angles = numericalDerivative11<Vector3, Vector3>(boost::bind(&PreIntegrateIMUObservations_delta_angles, msr_gyro_t, msr_dt, _1, flag_use_body_P_sensor, body_P_sensor, Bias_t0), delta_angles);
-    Matrix H_angles_bias = numericalDerivative11<Vector3, IMUBIAS>(boost::bind(&PreIntegrateIMUObservations_delta_angles, msr_gyro_t, msr_dt, delta_angles, flag_use_body_P_sensor, body_P_sensor, _1), Bias_t0);
+    Matrix H_angles_angles = numericalDerivative11<Vector3, Vector3>(boost::bind(&PreIntegrateIMUObservations_delta_angles, msr_gyro_t, msr_dt, boost::placeholders::_1, flag_use_body_P_sensor, body_P_sensor, Bias_t0), delta_angles);
```
varunagrawal, line 455: Run formatting only on the edited sections? Remember that we use Google style for formatting. :)

gtsam_unstable/slam/InertialNavFactor_GlobalVelocity.h
```diff
@@ -234,38 +235,38 @@ class InertialNavFactor_GlobalVelocity : public NoiseModelFactor5<POSE, VELOCITY
...
+      Matrix H4_Vel = gtsam::numericalDerivative11<VELOCITY, POSE>(boost::bind(&InertialNavFactor_GlobalVelocity::evaluateVelocityError, this, Pose1, Vel1, Bias1, boost::placeholders::_1, Vel2), Pose2);
       *H4 = stack(2, &H4_Pose, &H4_Vel);
     }
 
     // Jacobian w.r.t. Vel2
     if (H5){
       if (Vel2.size()!=3) throw std::runtime_error("Frank's hack to make this compile will not work if size != 3");
-      Matrix H5_Pose = gtsam::numericalDerivative11<POSE, Vector3>(boost::bind(&InertialNavFactor_GlobalVelocity::evaluatePoseError, this, Pose1, Vel1, Bias1, Pose2, _1), Vel2);
-      Matrix H5_Vel = gtsam::numericalDerivative11<Vector3, Vector3>(boost::bind(&InertialNavFactor_GlobalVelocity::evaluateVelocityError, this, Pose1, Vel1, Bias1, Pose2, _1), Vel2);
+      Matrix H5_Pose = gtsam::numericalDerivative11<POSE, Vector3>(boost::bind(&InertialNavFactor_GlobalVelocity::evaluatePoseError, this, Pose1, Vel1, Bias1, Pose2, boost::placeholders::_1), Vel2);
```
varunagrawal, line 268: Similar comment about formatting

gtsam_unstable/slam/InvDepthFactorVariant1.h
```diff
@@ -106,13 +108,13 @@ class InvDepthFactorVariant1: public NoiseModelFactor2<Pose3, Vector6> {
 
     if (H1) {
       (*H1) = numericalDerivative11<Vector, Pose3>(
-          boost::bind(&InvDepthFactorVariant1::inverseDepthError, this, _1,
+          boost::bind(&InvDepthFactorVariant1::inverseDepthError, this, boost::placeholders::_1,
```
varunagrawal, line 111: similar comment about formatting

gtsam_unstable/slam/InvDepthFactorVariant3.h
```diff
@@ -229,13 +231,13 @@ class InvDepthFactorVariant3b: public NoiseModelFactor3<Pose3, Pose3, Vector3> {
...
-      (*H1) = numericalDerivative11<Vector,Pose3>(boost::bind(&InvDepthFactorVariant3b::inverseDepthError, this, _1, pose2, landmark), pose1);
+      (*H1) = numericalDerivative11<Vector,Pose3>(boost::bind(&InvDepthFactorVariant3b::inverseDepthError, this, boost::placeholders::_1, pose2, landmark), pose1);
 
     if(H2)
-      (*H2) = numericalDerivative11<Vector,Pose3>(boost::bind(&InvDepthFactorVariant3b::inverseDepthError, this, pose1, _1, landmark), pose2);
+      (*H2) = numericalDerivative11<Vector,Pose3>(boost::bind(&InvDepthFactorVariant3b::inverseDepthError, this, pose1, boost::placeholders::_1, landmark), pose2);
 
     if(H3)
-      (*H3) = numericalDerivative11<Vector,Vector3>(boost::bind(&InvDepthFactorVariant3b::inverseDepthError, this, pose1, pose2, _1), landmark);
+      (*H3) = numericalDerivative11<Vector,Vector3>(boost::bind(&InvDepthFactorVariant3b::inverseDepthError, this, pose1, pose2, boost::placeholders::_1), landmark);
```
varunagrawal, line 240: format only the edited parts?

gtsam/base/numericalDerivative.h
```diff
@@ -942,7 +935,7 @@ inline typename internal::FixedSizeMatrix<X2,X3>::type numericalHessian323(
     boost::function<double(const X1&, const X2&, const X3&)> f, const X1& x1,
     const X2& x2, const X3& x3, double delta = 1e-5) {
   return numericalHessian212<X2, X3>(
-      boost::function<double(const X2&, const X3&)>(boost::bind(f, boost::cref(x1), _1, _2)),
+      boost::function<double(const X2&, const X3&)>(boost::bind(f, boost::cref(x1), boost::placeholders::_1, boost::placeholders::_2)),
```
varunagrawal, line 938: Run formatting on the edited parts?

gtsam/linear/tests/testGaussianBayesNet.cpp
acxz: I believe this is formatted in line with the existing code.

acxz: Maybe add a suggestion?

gtsam/navigation/tests/testManifoldPreintegration.cpp
acxz: I didn't because we don't use namespaces at all in this file.

gtsam/nonlinear/NonlinearEquality.h
acxz: Sure

gtsam/nonlinear/NonlinearEquality.h
```diff
@@ -87,7 +87,8 @@ class NonlinearEquality: public NoiseModelFactor1<VALUE> {
...
+          boost::placeholders::_1,boost::placeholders::_2,1e-9)) :
```
acxz, line 91:
```suggestion
          boost::placeholders::_1, boost::placeholders::_2, 1e-9)) :
```

gtsam/nonlinear/NonlinearEquality.h
```diff
@@ -97,7 +98,8 @@ class NonlinearEquality: public NoiseModelFactor1<VALUE> {
    * Constructor - allows inexact evaluation
    */
   NonlinearEquality(Key j, const T& feasible, double error_gain,
-      const CompareFunction &_compare = boost::bind(traits<T>::Equals,_1,_2,1e-9)) :
+      const CompareFunction &_compare = boost::bind(traits<T>::Equals,
+          boost::placeholders::_1,boost::placeholders::_2,1e-9)) :
```
acxz, line 102:
```suggestion
          boost::placeholders::_1, boost::placeholders::_2, 1e-9)) :
```

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
```diff
@@ -304,41 +305,43 @@ class EquivInertialNavFactor_GlobalVel : public NoiseModelFactor5<POSE, VELOCITY
       boost::optional<Matrix&> H4 = boost::none,
       boost::optional<Matrix&> H5 = boost::none) const override {
 
+    using namespace boost::placeholders;
+
```
acxz, line 309:
```suggestion
```

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
```diff
@@ -420,6 +423,8 @@ class EquivInertialNavFactor_GlobalVel : public NoiseModelFactor5<POSE, VELOCITY
     // Note: all delta terms refer to an IMU\sensor system at t0
     // Note: Earth-related terms are not accounted here but are incorporated in predict functions.
 
+    using namespace boost::placeholders;
+
```
acxz, line 427:
```suggestion
```

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
acxz:
```suggestion
```

gtsam/linear/tests/testGaussianBayesNet.cpp
varunagrawal: In other test files, the includes and the usings are grouped together.

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
acxz, line 309:
```suggestion
```

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
```diff
@@ -304,41 +305,43 @@ class EquivInertialNavFactor_GlobalVel : public NoiseModelFactor5<POSE, VELOCITY
       boost::optional<Matrix&> H4 = boost::none,
       boost::optional<Matrix&> H5 = boost::none) const override {
 
+    using namespace boost::placeholders;
```
acxz, line 308:
```suggestion
```

gtsam/navigation/tests/testManifoldPreintegration.cpp
varunagrawal: No reason not to add them for this case.

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
acxz: Can you clarify Google style?
Do you want me to keep it 80 line width?

gtsam/linear/tests/testGaussianBayesNet.cpp
acxz: I think it is appropriate to follow local formatting more than global formatting.
That can be left for another commit/PR.

gtsam_unstable/slam/EquivInertialNavFactor_GlobalVel.h
varunagrawal: If you use `clang-format`, one of its options is to use `Google` as the formatting style.

acxz:
Let me know if there is anything else @varunagrawal 

tbh feel free to make changes yourself as well

varunagrawal (approved):
Apologies for the delay in reviewing, but excellent work! LGTM

PR #806, opened 2021-06-29
Add user-defined copy constructor for Rot2

jaelrod:
In aarch64 platform architecture (running on a Jetson Xavier NX), it appears that copy elision performed as a return value optimization when returning a Rot2 from a function by-value results in its values `Rot2::c_` and `Rot2::s_` being initialized improperly or not at all.

Adding a user-defined copy constructor for Rot2 to properly return the value to the caller resolves this problem.

(also consolidated/simplified the Rot2 code normalizing its cosine and sine)

### DIFF
--- a/gtsam/geometry/Rot2.h
+++ b/gtsam/geometry/Rot2.h
@@ -50,6 +50,9 @@ namespace gtsam {
 
     /** default constructor, zero rotation */
     Rot2() : c_(1.0), s_(0.0) {}
+    
+    /** copy constructor */
+    Rot2(const Rot2& r) : Rot2(r.c_, r.s_) {}
 
     /// Constructor from angle in radians == exponential map at identity
     Rot2(double theta) : c_(cos(theta)), s_(sin(theta)) {}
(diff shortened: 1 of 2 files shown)


gtsam/geometry/Rot2.h
```diff
@@ -50,6 +50,9 @@ namespace gtsam {
 
     /** default constructor, zero rotation */
     Rot2() : c_(1.0), s_(0.0) {}
+    
+    /** copy constructor */
+    Rot2(const Rot2& r) : Rot2(r.c_, r.s_) {}
```
dellaert, line 55: Is this not default?

dellaert (approved):

dellaert:
I'll run CI and merge

gtsam/geometry/Rot2.h
jaelrod: Do you mean is this not the copy constructor that is generated by default by the compiler? I imagine that it probably is, but it is hard to say whether it is the compiler-generated version of the copy constructor itself or the copy constructor elision for return value optimization that was causing the error without fully dissecting that assembly I posted in the issue reply.

PR #811, opened 2021-07-08
Python fisheye interface

roderick-koehle:
This patch completes the python wrapper for the Unified and Omnidirectional camera model.
It also fixes the member function exposure to the calibrate / uncalibrate of the omnidirectional subclass.

Sample snipped to test the Omnidirectional / Unified Fisheye Model:
```python
import numpy as np
import gtsam
fx, fy, s, u0, v0 = 2, 2, 0, 0, 0
k1, k2, p1, p2, xi = 0, 0 , 0, 0, 1
stereographic = gtsam.Cal3Unified(fx, fy, s, u0, v0, k1, k2, p1, p2, xi)
x, y, z = 1, 0, 1
u, v = stereographic.uncalibrate([x, y])
correct = (u == 2/(1+np.linalg.norm(np.array[x, y, z])))
```

Sample snipped to test the Equidistant Fisheye Model:
```python
import numpy as np
import gtsam
equidistant = gtsam.Cal3Fisheye()
x, y, z = 1, 0, 1
u, v = equidistant .uncalibrate([x, y])
correct = (u == np.arctan2(x, z))
```


ProfFan:
@roderick-koehle Thanks for the contribution! Could you also add a unit test in `python/tests` (and preferably an example in `python/examples`)?

gtsam/geometry/SimpleCamera.h
```diff
@@ -33,7 +34,8 @@ namespace gtsam {
   using PinholeCameraCal3Bundler = gtsam::PinholeCamera<gtsam::Cal3Bundler>;
   using PinholeCameraCal3DS2 = gtsam::PinholeCamera<gtsam::Cal3DS2>;
   using PinholeCameraCal3Unified = gtsam::PinholeCamera<gtsam::Cal3Unified>;
-
+  using PinholeCameraCal3Fisheye = gtsam::PinholeCamera<gtsam::Cal3Fisheye>;
```
dellaert, line 37: Hmmm. why do we even have these typedefs and includes here? Seems like it exacerbates header bloat/compile times.

gtsam/geometry/triangulation.h
```diff
@@ -499,6 +501,8 @@ TriangulationResult triangulateSafe(const CameraSet<CAMERA>& cameras,
 // Vector of Cameras - used by the Python/MATLAB wrapper
 using CameraSetCal3Bundler = CameraSet<PinholeCamera<Cal3Bundler>>;
 using CameraSetCal3_S2 = CameraSet<PinholeCamera<Cal3_S2>>;
+using CameraSetCal3Fisheye = CameraSet<PinholeCamera<Cal3Fisheye>>;
+using CameraSetCal3Unified = CameraSet<PinholeCamera<Cal3Unified>>;
```
dellaert, line 505: Similar question. Maybe we should pare down and only list here what's absolutely required in triangulation. Caller can include.

gtsam/gtsam.i
```diff
@@ -977,6 +977,52 @@ class Cal3Bundler {
   void pickle() const;
 };
 
+#include <gtsam/geometry/Cal3Fisheye.h>
+class Cal3Fisheye {
```
dellaert, line 981: should this be a virtual class inheriting from a base class?

python/gtsam/tests/test_Cal3Fisheye.py
```diff
@@ -0,0 +1,105 @@
...
+
+class TestCal3Fisheye(GtsamTestCase):
+
+    def test_Cal3Fisheye(self):
+        K = gtsam.Cal3Fisheye()
+        self.assertEqual(K.fx(), 1.)
+        self.assertEqual(K.fy(), 1.)
+
+    def test_distortion(self):
+        "Equidistant fisheye model of focal length f, defined as r/f = tan(theta)"
```
dellaert, line 27: Make this into triple-quote doctsring?

dellaert (changes requested):
Awesome! Some of my comments - I realize - are about the state of things and not your PR specifically, but nevertheless a good time to think about whether to make them worse (include bloat).

Before approving let's have some discussion about that and one iteration on comments.

gtsam/geometry/SimpleCamera.h
varunagrawal: My guess is that it is an artifact of before we understood the wrapper's behavior really well. The wrapper has seen some nice improvements in the past year so we may as well kill this, while moving the typedef for `Cal3_S2` inside the deprecation block.

gtsam/gtsam.i
varunagrawal: Yeah this should technically inherit from `Cal3` (similar case for the other camera classes). This is my bad when I refactored the camera classes. :slightly_frowning_face:

gtsam/gtsam.i
```diff
@@ -2294,10 +2358,14 @@ class Values {
...
+                 gtsam::Cal3Fisheye,
+                 gtsam::Cal3Unified,
+		 gtsam::EssentialMatrix,
                  gtsam::PinholeCameraCal3_S2,
-                 gtsam::PinholeCamera<gtsam::Cal3Bundler>,
-                 gtsam::imuBias::ConstantBias,
+                 gtsam::PinholeCameraCal3Bundler,
+                 gtsam::PinholeCameraCal3Fisheye,
+                 gtsam::PinholeCameraCal3Unified,
+		 gtsam::imuBias::ConstantBias,
```
varunagrawal, line 2368: Maybe make the formatting consistent? I was wondering what was going on here for a second.

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -23,6 +23,89 @@ def test_Cal3Unified(self):
...
+        xi = 1
+        stereographic = gtsam.Cal3Unified(fx, fy, s, u0, v0, k1, k2, p1, p2, xi)
+        x, y, z = 1, 0, 1
+        u, v = stereographic.uncalibrate([x, y])
+        r = np.linalg.norm([x, y, z])
+        # Note: 2*tan(atan2(x, z)/2) = 2/(1+sqrt(x^2+z^2))
+        self.assertAlmostEqual(2*np.tan(np.arctan2(x, z)/2), 2/(1+r))
+        self.assertAlmostEqual(u, 2/(1+r))
+        x2, y2 = stereographic.calibrate([u, v])
+        self.assertAlmostEqual(x2, x)
```
varunagrawal, line 39: any reason these are `AlmostEqual` and not `Equal` (since tolerance is not specified)?

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -23,6 +23,89 @@ def test_Cal3Unified(self):
...
+        self.gtsamAssertEquals(pt2, np.array([u, v]))
+        obj1 = camera.backproject([u, v], z)
+        self.gtsamAssertEquals(obj1, np.array([x, y, z]))
+        r1 = camera.range(np.array([x, y, z]))
+        self.assertEqual(r1, r)
+
+    def test_generic_factor(self):
+        "Evaluate residual using pose and point as state variables"
+        fx, fy, s, u0, v0 = 2, 2, 0, 0, 0
+        k1, k2, p1, p2 = 0, 0, 0, 0
```
varunagrawal, line 66: Put all these set up values in a `setUp` function for the class and just access the variables directly?

python/gtsam/tests/test_Cal3Unified.py
varunagrawal, line 39: Any reason these are `AlmostEqual` and not `Equal` (since tolerance is not specified)?

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -23,6 +23,89 @@ def test_Cal3Unified(self):
...
+        xi = 1
+        objPoint = np.array([1, 0, 1])
+        r = np.linalg.norm(objPoint)
+        imgPoint = np.array([2/(1+r), 0])
```
varunagrawal, line 70: Similar case for `setUp` here since this and the previous test have the exact same values.

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -23,6 +23,89 @@ def test_Cal3Unified(self):
...
+        graph = gtsam.NonlinearFactorGraph()
+        state = gtsam.Values()
+        measured = imgPoint
+        noiseModel = gtsam.noiseModel.Isotropic.Sigma(2, 1)
+        poseKey = gtsam.symbol_shorthand.P(0)
+        pointKey = gtsam.symbol_shorthand.L(0)
```
varunagrawal, line 76: Maybe do `from gtsam.symbol_shorthand import L, P` to make this more readable?

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -23,6 +23,89 @@ def test_Cal3Unified(self):
...
+        self.assertAlmostEqual(score, 0)
+
+    def test_sfm_factor2(self):
+        "Evaluate residual with camera, pose and point as state variables"
+        fx, fy, s, u0, v0 = 2, 2, 0, 0, 0
+        k1, k2, p1, p2 = 0, 0, 0, 0
+        xi = 1
+        objPoint = np.array([1, 0, 1])
+        r = np.linalg.norm(objPoint)
+        imgPoint = np.array([2/(1+r), 0])
```
varunagrawal, line 92: This is a nitpick, but in Python it is advised to use snake case for variables.

python/gtsam/tests/test_Cal3Fisheye.py
```diff
@@ -0,0 +1,105 @@
...
+        self.assertAlmostEqual(u, np.arctan2(x, z))
+        self.assertAlmostEqual(v, 0)
+        self.assertAlmostEqual(x2, x)
+        self.assertAlmostEqual(y2, 0)
+
+    def test_pinhole(self):
+        "Spatial equidistant camera projection"
+        x, y, z = 1.0, 0.0, 1.0
+        u, v = np.arctan2(x, z), 0.0
+        camera = gtsam.PinholeCameraCal3Fisheye()
```
varunagrawal, line 41: A lot of the review comments also apply here. Please ask for another round of reviews after updating these. :slightly_smiling_face:

varunagrawal (changes requested):
This is a fantastic PR! Thanks @roderick-koehle.
I have some comments about style and clarity (because of which I am requesting changes), but functionality-wise this PR is great.

python/gtsam/tests/test_Cal3Fisheye.py
```diff
@@ -0,0 +1,146 @@
...
+    def setUpClass(cls):
+        """
+        Equidistant fisheye projection
+        
+        An equidistant fisheye projection with focal length f is defined
+        as the relation r/f = tan(theta), with r being the radius in the 
+        image plane and theta the incident angle of the object point.
+        """
+        x, y, z = 1.0, 0.0, 1.0
+        # x, y, z = 0.5, 0.0, 2.0
```
varunagrawal, line 32: Kill?

python/gtsam/tests/test_Cal3Fisheye.py
```diff
@@ -0,0 +1,146 @@
...
+        p1 = [-1.0, 0.0, -1.0]
+        p2 = [ 1.0, 0.0, -1.0]
+        q1 = gtsam.Rot3(1.0, 0.0, 0.0, 0.0)
+        q2 = gtsam.Rot3(1.0, 0.0, 0.0, 0.0)
+        obj_point = np.array([0.0, 0.0, 0.0])
+        pose1 = gtsam.Pose3(q1, p1)
+        pose2 = gtsam.Pose3(q2, p2)
+        camera1 = gtsam.PinholeCameraCal3Fisheye(pose1)
+        camera2 = gtsam.PinholeCameraCal3Fisheye(pose2)
+        cameras = gtsam.CameraSetCal3Fisheye([camera1, camera2])
```
varunagrawal, line 125: This and the previous test both have the same test data, so put that in `setUpClass`?

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -14,15 +14,135 @@
...
+        Stereographic fisheye projection
+        
+        An equidistant fisheye projection with focal length f is defined
+        as the relation r/f = 2*tan(theta/2), with r being the radius in the 
+        image plane and theta the incident angle of the object point.
+        """
+        x, y, z = 1.0, 0.0, 1.0
+        r = np.linalg.norm([x, y, z])
+        u, v = 2*x/(z+r), 0.0
+        #u, v = 2*np.tan(np.arctan2(x, z)/2), 0.0
```
varunagrawal, line 34: kill?

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -14,15 +14,135 @@
...
+        """Estimate spatial point from image measurements using rectification"""
+        p1 = [-1.0, 0.0, -1.0]
+        p2 = [ 1.0, 0.0, -1.0]
+        q1 = gtsam.Rot3(1.0, 0.0, 0.0, 0.0)
+        q2 = gtsam.Rot3(1.0, 0.0, 0.0, 0.0)
+        obj_point = np.array([0.0, 0.0, 0.0])
+        pose1 = gtsam.Pose3(q1, p1)
+        pose2 = gtsam.Pose3(q2, p2)
+        camera1 = gtsam.PinholeCameraCal3Unified(pose1, self.stereographic)
+        camera2 = gtsam.PinholeCameraCal3Unified(pose2, self.stereographic)
```
varunagrawal, line 136: Same deal. If the data is the same for multiple tests, make your life easy by putting it in the `setUp` method.

varunagrawal (changes requested):
Almost there!!

varunagrawal (approved):
Beautiful!! Amazing work @roderick-koehle

ProfFan:
Last GO/NOGO needed from @dellaert  :rocket:

dellaert (approved):
very nice! I guess this should be merged before we merge the split of the wrapper files !

dellaert:
Thanks all !!!!

PR #817, opened 2021-07-09
Add ShonanAveraging2 interface to wrapper with BetweenFactorPose2s, that does not require g2o files

johnwlambert:
Currently, the python wrapper includes an interface to `ShonanAveraging2`, but only accepts g2o files as input
```c++
  ShonanAveraging2(string g2oFile);
  ShonanAveraging2(string g2oFile,
                   const gtsam::ShonanAveragingParameters2 &parameters);
```

This PR adds support so that optimization over `Rot2` is a first-class citizen in the wrapper with `Rot3`, since `Rot3` has the following additional support:
```c++
  ShonanAveraging3(string g2oFile);
  ShonanAveraging3(string g2oFile,
                   const gtsam::ShonanAveragingParameters3 &parameters);

  // TODO(frank): deprecate once we land pybind wrapper
  ShonanAveraging3(const gtsam::BetweenFactorPose3s &factors);
  ShonanAveraging3(const gtsam::BetweenFactorPose3s &factors,
                   const gtsam::ShonanAveragingParameters3 &parameters);
```

@dellaert I see a line here about deprecating something after the pybind wrapper is landed -- could you give a bit more context about that? It seems relevant to this PR (in case I'm extending something that was supposed to be deprecated). Thanks!
```c++
ShonanAveraging3::ShonanAveraging3(string g2oFile, const Parameters &parameters)
    : ShonanAveraging<3>(maybeRobust(parseMeasurements<Rot3>(g2oFile),
                                     parameters.getUseHuber()),
                         parameters) {}
// TODO(frank): Deprecate after we land pybind wrapper
// Extract Rot3 measurement from Pose3 betweenfactors
// Modeled after similar function in dataset.cpp
static BinaryMeasurement<Rot3> convert(
    const BetweenFactor<Pose3>::shared_ptr &f) {
  auto gaussian =
      boost::dynamic_pointer_cast<noiseModel::Gaussian>(f->noiseModel());
  if (!gaussian)
    throw std::invalid_argument(
        "parseMeasurements<Rot3> can only convert Pose3 measurements "
        "with Gaussian noise models.");
  const Matrix6 M = gaussian->covariance();
  auto model = noiseModel::Gaussian::Covariance(M.block<3, 3>(3, 3));
  return BinaryMeasurement<Rot3>(f->key1(), f->key2(), f->measured().rotation(),
                                 model);
}
```

### DIFF
--- a/gtsam/sfm/ShonanAveraging.cpp
+++ b/gtsam/sfm/ShonanAveraging.cpp
@@ -944,6 +944,20 @@ ShonanAveraging2::ShonanAveraging2(string g2oFile, const Parameters &parameters)
                                      parameters.getUseHuber()),
                          parameters) {}
 
+static ShonanAveraging2::Measurements extractRot2Measurements(
+    const BetweenFactorPose2s &factors) {
+  ShonanAveraging2::Measurements result;
+  result.reserve(factors.size());
+  for (auto f : factors) result.push_back(convert(f));
+  return result;
+}
+
+ShonanAveraging2::ShonanAveraging2(const BetweenFactorPose2s &factors,
+                                   const Parameters &parameters)
+    : ShonanAveraging<3>(maybeRobust(extractRot2Measurements(factors),
+                                     parameters.getUseHuber()),
+                         parameters) {}    
+    
 /* ************************************************************************* */
 // Explicit instantiation for d=3
 template class ShonanAveraging<3>;
--- a/gtsam/sfm/ShonanAveraging.cpp
+++ b/gtsam/sfm/ShonanAveraging.cpp
@@ -944,11 +944,27 @@ ShonanAveraging2::ShonanAveraging2(string g2oFile, const Parameters &parameters)
                                      parameters.getUseHuber()),
                          parameters) {}
 
+// Extract Rot2 measurement from Pose2 betweenfactors
+// Modeled after similar function in dataset.cpp
+static BinaryMeasurement<Rot2> convertPose2ToBinaryMeasurementRot2(
+    const BetweenFactor<Pose2>::shared_ptr &f) {
+  auto gaussian =
+      boost::dynamic_pointer_cast<noiseModel::Gaussian>(f->noiseModel());
+  if (!gaussian)
+    throw std::invalid_argument(
+        "parseMeasurements<Rot2> can only convert Pose2 measurements "
+        "with Gaussian noise models.");
+  const Matrix6 M = gaussian->covariance();
+  auto model = noiseModel::Gaussian::Covariance(M.block<2, 2>(2, 2));
+  return BinaryMeasurement<Rot2>(f->key1(), f->key2(), f->measured().rotation(),
+                                 model);
+}
+    
 static ShonanAveraging2::Measurements extractRot2Measurements(
     const BetweenFactorPose2s &factors) {
   ShonanAveraging2::Measurements result;
   result.reserve(factors.size());
-  for (auto f : factors) result.push_back(convert(f));
+  for (auto f : factors) result.push_back(convertPose2ToBinaryMeasurementRot2(f));
   return result;
 }
 
--- a/gtsam/sfm/ShonanAveraging.cpp
+++ b/gtsam/sfm/ShonanAveraging.cpp
@@ -954,7 +954,7 @@ static ShonanAveraging2::Measurements extractRot2Measurements(
 
 ShonanAveraging2::ShonanAveraging2(const BetweenFactorPose2s &factors,
                                    const Parameters &parameters)
-    : ShonanAveraging<3>(maybeRobust(extractRot2Measurements(factors),
+    : ShonanAveraging<2>(maybeRobust(extractRot2Measurements(factors),
                                      parameters.getUseHuber()),
                          parameters) {}    
     
--- a/gtsam/sfm/ShonanAveraging.cpp
+++ b/gtsam/sfm/ShonanAveraging.cpp
@@ -954,7 +954,7 @@ static BinaryMeasurement<Rot2> convertPose2ToBinaryMeasurementRot2(
     throw std::invalid_argument(
         "parseMeasurements<Rot2> can only convert Pose2 measurements "
         "with Gaussian noise models.");
-  const Matrix6 M = gaussian->covariance();
+  const Matrix3 M = gaussian->covariance();
   auto model = noiseModel::Gaussian::Covariance(M.block<2, 2>(2, 2));
   return BinaryMeasurement<Rot2>(f->key1(), f->key2(), f->measured().rotation(),
                                  model);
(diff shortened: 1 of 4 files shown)


dellaert:
I think, @ProfFan correct me if I'm wrong, that with pybind we can now just pass a container of `BetweenFactorPose`, i.e., there would be no need to have a special `BetweenFactorPose3s` type

dellaert (approved):

ProfFan:
That is correct, given that they are specialized in `specializations.h`. Performance will be the same if one use this pattern:
```
b = gtsam.BetweenFactorPose3Vector() # this is std::vector
b.append(xxx)
```
but this also works:
```
b = [ xxx, xxx, xxx ]
```
which will involve copies, but should work.

johnwlambert:
> That is correct, given that they are specialized in `specializations.h`. Performance will be the same if one use this pattern:
> 
> ```
> b = gtsam.BetweenFactorPose3Vector() # this is std::vector
> b.append(xxx)
> ```
> 
> but this also works:
> 
> ```
> b = [ xxx, xxx, xxx ]
> ```
> 
> which will involve copies, but should work.

Cool, thanks @dellaert and @ProfFan. Fan, should I remove the "TODO" from the .cpp file? If I understood you correctly, the only change is that on the Python side, the user can now choose to create 

```
b = gtsam.BetweenFactorPose3Vector()
b.append(BetweenFactorPose3(i2, i1, i2Ti1, noise_model))
```
or 
```
b = gtsam.BetweenFactorPose3s()
b.append(BetweenFactorPose3(i2, i1, i2Ti1, noise_model))
```
or
```
b = [BetweenFactorPose3(i2, i1, i2Ti1, noise_model)), BetweenFactorPose3(i2, i1, i2Ti1, noise_model)) ]
```
Is there anything on the c++ side that we want to simplify or clean up here?

gtsam/sfm/ShonanAveraging.cpp
```diff
@@ -944,6 +944,36 @@ ShonanAveraging2::ShonanAveraging2(string g2oFile, const Parameters &parameters)
...
+static BinaryMeasurement<Rot2> convertPose2ToBinaryMeasurementRot2(
+    const BetweenFactor<Pose2>::shared_ptr &f) {
+  auto gaussian =
+      boost::dynamic_pointer_cast<noiseModel::Gaussian>(f->noiseModel());
+  if (!gaussian)
+    throw std::invalid_argument(
+        "parseMeasurements<Rot2> can only convert Pose2 measurements "
+        "with Gaussian noise models.");
+  const Matrix3 M = gaussian->covariance();
+  auto model = noiseModel::Gaussian::Covariance(M.block<1, 1>(2, 2));
```
johnwlambert, line 958: @dellaert or @ProfFan could either of you point me to where the ordering for the Pose2 elements is documented when you get a chance? Not sure if it's (x,y,theta) or (theta,x,y). Here I'm assuming it's (x,y,theta), but I read [here](https://github.com/borglab/gtsam/blob/43e8f1e5aeaf11890262722c1e5e04a11dbf9d75/gtsam/slam/dataset.cpp#L855) that in GTSAM the g2o order (t,R) is the opposite of GTSAM's (R,t) order, so maybe I have it wrong.


@dellaert regarding Pose3, I noticed that [here](https://github.com/borglab/gtsam/pull/817/files#diff-2681e2ffa3c440f21039b01556f84cf32000a6c0c251989f657835ee69622367R1004) the [3:6,3:6] block of the 6x6 covariance matrix is used for R:
```
  const Matrix6 M = gaussian->covariance();
  auto model = noiseModel::Gaussian::Covariance(M.block<3, 3>(3, 3));
  return BinaryMeasurement<Rot3>(f->key1(), f->key2(), f->measured().rotation(),
                                 model);
```

but [here](https://github.com/borglab/gtsam/blob/43e8f1e5aeaf11890262722c1e5e04a11dbf9d75/gtsam/slam/dataset.cpp#L885) it appears to be the opposite: the block [0:3,0:3] is pulled out for R:
```python
  const Matrix6 M = gaussian->covariance();
  return BinaryMeasurement<Rot3>(
      p.key1(), p.key2(), p.measured().rotation(),
      noiseModel::Gaussian::Covariance(M.block<3, 3>(0, 0), true));
}
```
Is that intentional or perhaps a typo? Thanks!

ProfFan: Assuming that you are talking about the tangent space, it's
```c++
Vector3 Pose2::Logmap(const Pose2& p, OptionalJacobian<3, 3> H) {
  if (H) *H = Pose2::LogmapDerivative(p);
  const Rot2& R = p.r();
  const Point2& t = p.t();
  double w = R.theta();
  if (std::abs(w) < 1e-10)
    return Vector3(t.x(), t.y(), w);
  else {
    double c_1 = R.c()-1.0, s = R.s();
    double det = c_1*c_1 + s*s;
    Point2 p = R_PI_2 * (R.unrotate(t) - t);
    Point2 v = (w/det) * p;
    return Vector3(v.x(), v.y(), w);
  }
}
```

python/gtsam/tests/test_ShonanAveraging.py
```diff
@@ -133,6 +145,58 @@ def test_PriorWeights(self):
...
+        
+        noise_model = gtsam.noiseModel.Unit.Create(3)
+        between_factors = gtsam.BetweenFactorPose2s()
+        for (i1, i2), i2Ri1 in i2Ri1_dict.items():
+            i2Ti1 = Pose2(i2Ri1, np.zeros(2))
+            between_factors.append(BetweenFactorPose2(i2, i1, i2Ti1, noise_model))
+        
+        obj = ShonanAveraging2(between_factors, shonan_params)
+        initial = obj.initializeRandomly()
+        result_values, _ = obj.run(initial, min_p=2, max_p=40)
```
johnwlambert, line 196: @dellaert is there any way you could take a look at this ShonanAveraging2 unit test? For some reason it can't seem to converge even when I set pmax all the way up to 40

gtsam/sfm/ShonanAveraging.cpp
johnwlambert: thanks!

johnwlambert:
@ProfFan any idea as to why just one of the 22 checks is failing during compilation?

ProfFan:
@johnwlambert It's OOM caused by wrapper. I think it doesn't matter - after we merge #822 this issue will be nonexistent 

Again kudos @varunagrawal

johnwlambert:
> @johnwlambert It's OOM caused by wrapper. I think it doesn't matter - after we merge #822 this issue will be nonexistent
> 
> Again kudos @varunagrawal

Got it, that makes sense. @ProfFan could you do a review of this PR too? are you ok with merging it even with that CI test failing, since we are deadlocked (Varun was waiting for this PR to merge to fix OOM)

ProfFan (approved):
LGTM!

PR #819, opened 2021-07-11
Replace boost with std

### LINKED ISSUE #812: Compilation error with boost::placeholders
## Description

Using `boost::placeholders` in 16.04 is not supported (out of the box).
Is there a reason why you use `boost::` and not `std::` ?

```
/root/gtsam/gtsam/inference/Symbol.cpp:66:61: error: 'boost::placeholders' has not been declared
   return boost::bind(&Symbol::chr, boost::bind(make, boost::placeholders::_1)) == c;
```

## Steps to reproduce

1. Compile in fresh 16.04 (docker ideally)

## Environment

Ubuntu 16.04

varunagrawal:
Fixes #812

1. Use `std::placeholders` instead of `boost::placeholders`.
2. Use `std::bind` instead of `boost::bind`.
3. Use `std::function` instead of `boost::function`.

Had to make edits to functionals in some places (used lambdas instead) to get things to work. Also the tests can't seem to disambiguate the placeholder variables (guess this is because some other boost library is importing `boost/bind` internally), so I just made them have the full namespace.

@ToniRV FYI

### DIFF
--- a/gtsam/base/Matrix.h
+++ b/gtsam/base/Matrix.h
@@ -29,7 +29,7 @@
 #include <gtsam/config.h>
 
 #include <boost/format.hpp>
-#include <boost/function.hpp>
+#include <functional>
 #include <boost/tuple/tuple.hpp>
 #include <boost/math/special_functions/fpclassify.hpp>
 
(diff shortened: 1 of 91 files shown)


ProfFan (approved):
2nd

dellaert:
Awesomesauce!

PR #821, opened 2021-07-11
Break interface file into multiple files

varunagrawal:
Big PR alert!

This PR allows breaking the GTSAM wrapper into multiple smaller files. The benefits are multifold:

1. Parallel compilation for improved speed.
2. Takes care of the OOM issue in the CI.
3. Smaller files are easier to debug and update.

I have tested this for both Python and Matlab and they work great.


dellaert:
Hi Varun. Close without merge? Is there still an issue?

ProfFan:
@dellaert  New PR is #822

Issue #820, opened 2021-07-11
segmentation fault if too many variables and constraints.


Tahiti410:
## Description
Segmentation fault if too many(about 140k) constraints and states.

I guess it's related to the resources release(recursive) in descrutors. 
## Steps to reproduce

1. check/set the stack size of your machine by `ulimit -s 8192`
2. Replace the code in example/LocalizationExample.cpp by code in Additional information.
2. cd build
2. build an run the demo program by make -j8 LocalizationExample && ./example/LocalizationExample

## Expected behavior
`Segmentation fault (core dumped)`

## Environment
~~Master~~`develop` branch of gtsam
`Linux version 5.4.0-52-generic (buildd@xxx-amd64-022) (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)) #57~18.04.1-Ubuntu SMP Thu Oct 15 14:04:49 UTC 2020`

`ulimit -s` output `8192`

## Additional information
~~~cpp
#include <gtsam/geometry/Pose2.h>
#include <gtsam/inference/Key.h>
#include <gtsam/nonlinear/LevenbergMarquardtOptimizer.h>
#include <gtsam/nonlinear/NonlinearFactor.h>
#include <gtsam/nonlinear/NonlinearFactorGraph.h>
#include <gtsam/nonlinear/Values.h>
#include <gtsam/slam/BetweenFactor.h>

using namespace std;
using namespace gtsam;

class UnaryFactor : public NoiseModelFactor1<Pose2> {
  double mx_, my_;

 public:
  typedef boost::shared_ptr<UnaryFactor> shared_ptr;
  UnaryFactor(Key j, double x, double y, const SharedNoiseModel& model)
      : NoiseModelFactor1<Pose2>(model, j), mx_(x), my_(y) {}

  ~UnaryFactor() override {}
  Vector evaluateError(const Pose2& q, boost::optional<Matrix&> H = boost::none) const override {
    const Rot2& R = q.rotation();
    if (H) (*H) = (gtsam::Matrix(2, 3) << R.c(), -R.s(), 0.0, R.s(), R.c(), 0.0).finished();
    return (Vector(2) << q.x() - mx_, q.y() - my_).finished();
  }
  gtsam::NonlinearFactor::shared_ptr clone() const override {
    return boost::static_pointer_cast<gtsam::NonlinearFactor>(
        gtsam::NonlinearFactor::shared_ptr(new UnaryFactor(*this)));
  }
};

[...]

Issue #809, opened 2021-06-30
#800 causes Levenberg-Marquardt optimisation to crash


HMellor:
## Description

When updating to the latest commit of the `develop` branch my existing codebase that uses GTSAM started crashing when running LM. 

There is no stack-trace, only an error saying the process has died:
```
[ERROR] [fgo-1]: process has died [pid 8821, exit code -11,
    cmd '/home/user/ros_ws/install/fgo/lib/fgo/fgo--ros-args -r __node:=fgo
    --params-file /tmp/launch_params__dkrh3ic --params-file /tmp/launch_params_8v56gi4f
    --params-file /tmp/launch_params_btvjwucu'].
```

I have narrowed it down to 6ee17c2ed7b2a89f4a25ae568009bb9385061ad1 which comes from #800.  Checking out the previous commit causes the program to no longer crash.

It may help to list some classes I'm using:
* NonlinearFactorGraph
* CombinedImuFactor
* GPSFactor

## Expected behavior

Should not crash

## Environment

 * **OS**
    * Distributor ID: Ubuntu
    * Description:    Ubuntu 20.04.2 LTS
    * Release:        20.04
    * Codename:       focal
* **ROS**
    * Foxy
* **TBB Version (relevant to #800)**
    * libtbb-dev/focal,now 2020.1-2 amd64 [installed]
    * libtbb2/focal,now 2020.1-2 amd64 [installed,automatic]
* **Language**
    * C++

ProfFan:
@acxz Could you take a look?

ProfFan:
Also, @HMellor could you provide a minimal reproducer (one cpp file, no other dependencies)?

dellaert:
@acxz this was your PR - could you weigh in? @HMellor thatnks for letting us know the TBB versions you used.

acxz:
This might be an issue similar to why the testGncOptimize test failed once out of 100 runs. I could take a deeper look into this issue with a minimal reproducing example.

In the meanwhile, it may be better to just revert my PR.

PR #800, opened 2021-06-19
replace deprecated tbb functionality

acxz:
resolves #208 continuation of #336 

I was able to replace the creation of the old api tasks with functors being passed into task_groups.

Since this is a change to core functionality, I urge people to review and check unit tests thoroughly.

PR #823, opened 2021-07-15
Revert "replace deprecated tbb functionality"

acxz:
Reverts borglab/gtsam#800

due to issues raised in #809 and #820


ProfFan (approved):
Thanks @acxz :)

PR #832, opened 2021-07-29
Wrapper for basis functions

gchenfc:
Add some fixes needed to make the wrapper for basis functions work, and added `FitBasis` to the wrapper.

Will probably want a squash-merge due to commit noise


gtsam/basis/basis.i
```diff
@@ -64,7 +68,7 @@ class ParameterMatrix {
 #include <gtsam/basis/BasisFactors.h>
 
 template <BASIS = {gtsam::Chebyshev2}>
-class EvaluationFactor : gtsam::NoiseModelFactor {
+virtual class EvaluationFactor : gtsam::NoiseModelFactor {
```
varunagrawal, line 71: Does this have to be `virtual`?

gtsam/basis/basis.i
```diff
@@ -122,3 +126,32 @@ class ManifoldEvaluationFactor : gtsam::NoiseModelFactor {
...
                            double x, double a, double b);
 };
+
+// TODO(gerry): Add `DerivativeFactor`, `VectorDerivativeFactor`, and
+// `ComponentDerivativeFactor`
+
+#include <gtsam/basis/FitBasis.h>
+// We'll allow transparent binding of python dict to Sequence in this
+// compilation unit using pybind11/stl.h.
+// Another alternative would be making Sequence opaque in
```
varunagrawal, line 136: Pick one option and implement that?

varunagrawal (approved):
LGTM

gtsam/basis/basis.i
gchenfc: Just for future reference, ya I was getting a compile (or linker?) error without it:
https://github.com/borglab/wrap/blob/master/DOCS.md
> * Virtual inheritance
>   * ...
>   * Mark with virtual keyword, e.g. virtual class Base {, and also virtual class Derived : ns::Base {.

:)

gtsam/basis/basis.i
gchenfc: Thanks, I was thinking someone might question why I went this route including `pybind11/stl.h` so I wanted to give a reason why, but in the future I'll be less verbose.

gtsam/basis/basis.i
varunagrawal: Yeah I am aware of the docs, but the pybind wrapper doesn't quite use virtual anywhere for subclasses. I may be wrong, but the virtual is supposed to be mainly for base classes.

Gotta look it up and update docs.

gtsam/basis/basis.i
varunagrawal: Verbosity is not the issue, the more verbose the better actually. I am more of the opinion that we should pick one way and stick to it, and the alternatives can be discussed beforehand. :slightly_smiling_face: 

FYI I am adding the stl.h include to the `gtsam.tpl` file directly since includes don't add much overhead.

varunagrawal: Actually no, we don't include `stl.h` since it interferes with other containers where we need reference passing. I'm just adding `Sequence` to specializations.

gchenfc: I got build errors when I tried that, hence only including it in the basis module - but would be helpful if you did succeed :)

I assume related to other folders using opaque types

gchenfc: oops commented at the same time - ya

My reasoning for not adding Sequence to specializations is described in this comment: it will muddy the syntax to do

```python
fit = gtsam.FitBasis(3, gtsam.Sequence({data}), noiseModel)
```
as compared to 
```python
fit = gtsam.FitBasis(3, data, noiseModel)
```

And the advantages of the specialization stuff, namely correct reference passing, isn't as important because this is never going to be a mutable variable.

Maybe I should have been even more verbose in my comment to make this clearer 😂

PR #838, opened 2021-08-04
Fit Basis python wrapper

gchenfc:
FitBasis now works in python, with unit test


varunagrawal:
CI failing?

python/gtsam/preamble/basis.h
```diff
@@ -10,3 +10,11 @@
...
  * mutations on Python side will not be reflected on C++.
  */
+
+// We'll allow transparent binding of python dict to Sequence in this
+// compilation unit using pybind11/stl.h.
+// Another alternative would be making Sequence opaque in
+// python/gtsam/{preamble, specializations}, but std::map<double, double> is
+// common enough that it may cause collisions, and we don't need
+// reference-access anyway.
+#include <pybind11/stl.h>
```
varunagrawal, line 20: I had tried this before and it didn't work for me. I am sure there is something we are missing.

dellaert (approved):

dellaert (changes requested):
Actually, CI does not pass, so can’t land yet

varunagrawal:
FYI I resolved this issue in another (arguably cleaner) way.

PR #843, opened 2021-08-11
Wrap GNC

johnwlambert:
Add GNC to the wrapper.

### DIFF
--- a/gtsam/nonlinear/nonlinear.i
+++ b/gtsam/nonlinear/nonlinear.i
@@ -527,7 +527,7 @@ template<class BaseOptimizerParameters>
 class GncParams {
   GncParams(const BaseOptimizerParameters& baseOptimizerParams);
   GncParams();
-  void print(const std::string& str) const;
+  void print(const string& str) const;
 };
   
 #include <gtsam/nonlinear/NonlinearOptimizer.h>
--- a/gtsam/nonlinear/nonlinear.i
+++ b/gtsam/nonlinear/nonlinear.i
@@ -524,7 +524,7 @@ virtual class DoglegParams : gtsam::NonlinearOptimizerParams {
 
 #include <gtsam/nonlinear/GncParams.h>
 template<PARAMS = {gtsam::GaussNewtonParams, gtsam::LevenbergMarquardtParams}>
-class GncParams {
+virtual class GncParams {
   GncParams(const PARAMS& baseOptimizerParams);
   GncParams();
   void print(const string& str) const;
@@ -563,7 +563,7 @@ virtual class DoglegOptimizer : gtsam::NonlinearOptimizer {
 #include <gtsam/nonlinear/GncOptimizer.h>
 template<PARAMS = {gtsam::GncParams<gtsam::GaussNewtonParams>,
                    gtsam::GncParams<gtsam::LevenbergMarquardtParams>}>
-class GncOptimizer {
+virtual class GncOptimizer {
   GncOptimizer(const gtsam::NonlinearFactorGraph& graph,
                const gtsam::Values& initialValues,
                const PARAMS& params);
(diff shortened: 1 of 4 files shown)


johnwlambert:
@ProfFan hi Fan, any thoughts on why the Python CI fails here?

gtsam/nonlinear/nonlinear.i
```diff
@@ -522,6 +522,14 @@ virtual class DoglegParams : gtsam::NonlinearOptimizerParams {
   void setVerbosityDL(string verbosityDL) const;
 };
 
+#include <gtsam/nonlinear/GncParams.h>
+template<class BaseOptimizerParameters>
```
varunagrawal, line 526: You shouldn't be specifying templates as `template<class T>`, it should be just `template<T>`.
Same deal for `GncOptimizer`.

johnwlambert: Got it, thank you Varun

gtsam/nonlinear/nonlinear.i
```diff
@@ -522,6 +522,14 @@ virtual class DoglegParams : gtsam::NonlinearOptimizerParams {
   void setVerbosityDL(string verbosityDL) const;
 };
 
+#include <gtsam/nonlinear/GncParams.h>
+template<BaseOptimizerParameters>
```
varunagrawal, line 526: Nitpick: Use a smaller uppercase name like `PARAMS` and provide either a default via the template list `PARAMS = {SomeValue}` or use the typedef to declare on (I prefer the former).

Without a template list of typedef, the python module import should error out.

gtsam/nonlinear/nonlinear.i
```diff
@@ -551,6 +559,15 @@ virtual class DoglegOptimizer : gtsam::NonlinearOptimizer {
                   const gtsam::DoglegParams& params);
   double getDelta() const;
 };
+  
+#include <gtsam/nonlinear/GncOptimizer.h>
+template<GncParameters>
```
varunagrawal, line 564: Same as above.

varunagrawal (commented):
I have a feeling this will compile but won't import since the templates don't have instantiations.

gtsam/nonlinear/nonlinear.i
johnwlambert: You're right, i checked and it compiled but import didn't work.

I updated the templates, let me know if this is what you had in mind @varunagrawal. Thanks.

gtsam/nonlinear/nonlinear.i
johnwlambert: should be addressed now

gtsam/nonlinear/nonlinear.i
johnwlambert: weird, even with the updated templates we still get
```bash
ERROR: test_NonlinearOptimizer (unittest.loader._FailedTest)
ImportError: Failed to import test module: test_NonlinearOptimizer
  File "/usr/lib/python3.6/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/runner/work/gtsam/gtsam/python/gtsam/tests/test_NonlinearOptimizer.py", line 18, in <module>
    from gtsam import (DoglegOptimizer, DoglegParams,
ImportError: cannot import name 'GncParams'
```

johnwlambert: I'm guessing this is because we were missing GTSAM_EXPORT calls, and because the class wasn't marked as virtual in the .i files. Do we have those steps documented somewhere in `gtwrap`, along with why they are required? I'm a bit confused about that.

johnwlambert: Thanks @varunagrawal, looks like all the tests pass now.

varunagrawal (approved):
Sweet!

PR #844, opened 2021-08-12
Add python type annotations to some older python files

johnwlambert:
- Add type hints per PEP 484 (https://google.github.io/styleguide/pyguide.html#221-type-annotated-code). Some were previous added as part of docstring. These cannot be used for static analysis, so moved to function signature itself.
- remove dangerous wildcard imports
- fix some erroneous docstrings
- add some keyword args for readability

Files affected: 
- test_Pose3SLAMExample.py
- circlePose3.py
- plot.py
- visual_data_generator.py

### DIFF
--- a/python/gtsam/tests/test_Pose3SLAMExample.py
+++ b/python/gtsam/tests/test_Pose3SLAMExample.py
@@ -15,14 +15,14 @@ import numpy as np
 
 import gtsam
 from gtsam.utils.test_case import GtsamTestCase
-from gtsam.utils.circlePose3 import *
+from gtsam.utils.circlePose3 import circlePose3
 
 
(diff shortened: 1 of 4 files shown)


ProfFan (approved):
It appears that the Python unit tests are failing, but otherwise looks good :)

johnwlambert:
> It appears that the Python unit tests are failing, but otherwise looks good :)

thanks @ProfFan, should be fixed now

ProfFan:
Noice!

varunagrawal:
There's a missing import for `Pose3` that was introduced in this PR. We really need to add more Python tests.

varunagrawal:
Also, @johnwlambert what formatter are you using? We all use yapf in case you use something else.

PR #848, opened 2021-08-15
Fixing the EKF example and change linearization point initialization

ProfFan:
This PR tries to change the linearization point to a linear expolation of the current estimate using the previous linearization.

However I found this to be not as useful as it would be in the intended usage in GTSAM:

1. All GTSAM examples and unit tests for the EKF uses a motion model that does not depend on the future state. i.e. they are all explicit models, thus this change is unnecessary.
2. Adding this does not break anything, but it does add more runtime complexity (though not too much).

In making this change I also found out that the `easyPoint2KalmanFilter` example has a typo, which I also fixed.

If we decide that this is unnecessary, I can keep only the typo fix.


gtsam/nonlinear/ExtendedKalmanFilter.h
```diff
@@ -58,6 +59,8 @@ class ExtendedKalmanFilter {
  protected:
   T x_;                                     // linearization point
   JacobianFactor::shared_ptr priorFactor_;  // Gaussian density on x_
+  GaussianFactorGraph::shared_ptr lastLinearization;
```
dellaert, line 62: docs: what are these things?

gtsam/nonlinear/ExtendedKalmanFilter-inl.h
```diff
@@ -82,20 +83,38 @@ namespace gtsam {
...
-    linearFactorGraph.push_back(priorFactor_);
+    linearFactorGraph->push_back(priorFactor_);
 
     // Linearize motion model and add it to the Kalman Filter graph
     Values linearizationPoint;
-    linearizationPoint.insert(keys[0], x_);
-    linearizationPoint.insert(keys[1], x_); // TODO should this really be x_ ?
-    linearFactorGraph.push_back(motionFactor.linearize(linearizationPoint));
+
+    if (lastLinearization != nullptr) {
```
dellaert, line 94: document! What's happening?

gtsam/nonlinear/ExtendedKalmanFilter-inl.h
```diff
@@ -82,20 +83,38 @@ namespace gtsam {
     const auto keys = motionFactor.keys();
 
     // Create a Gaussian Factor Graph
-    GaussianFactorGraph linearFactorGraph;
+    GaussianFactorGraph::shared_ptr linearFactorGraph = boost::make_shared<GaussianFactorGraph>();
```
dellaert, line 86: Why did you change this to a pointer?

dellaert:
@ProfFan address comments so we can merge

dellaert:
Thanks for the contribution and investigation here. I’m going to close this PR because the example correction was independently merged in #2031, while the remaining extrapolation change is now stale and conflicts with `develop`.

More importantly, the extrapolated linearization can land at a singular future-state Jacobian for valid implicit motion factors, causing an `IndeterminateSystemException` where the current implementation succeeds. The existing tests do not expose this because their motion models are affine in the future state.

If this capability is still needed, please open a fresh, focused PR with an implicit-motion regression test and either an opt-in extrapolation policy or a robust fallback.

PR #851, opened 2021-08-20
Fix utilities and add tests

### LINKED ISSUE #841: gtsam.utilities.extractPoint2() doesn't work
gtsam.utilities.extractPoint2(Values) return empty array, even if Values have Point2 variables. 

## Steps to reproduce

1. `initial = gtsam.Values()`
2.  `initial.insert(1, gtsam.Pose2(0.0, 0.1, 0.1)`
3.   `initial.insert(2, gtsam.Point2(0.0, 0.1))`
4.   `gtsam.utilities.extractPoint2(initial)` returns `array([], shape=(0, 2), dtype=float64)`
5.  Where as `gtsam.utilities.extractPose2(initial)` return correctly which is `array([[0. , 0.1, 0.1]])`
  

Expted behavior of `gtsam.utilities.extractPoint2(initial)` is returning `array([[0. , 0.1]])`

varunagrawal:
Fixes #841

### DIFF
--- a/gtsam/geometry/tests/testUtilities.cpp
+++ b/gtsam/geometry/tests/testUtilities.cpp
@@ -1,0 +1,64 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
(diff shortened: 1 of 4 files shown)


ProfFan (approved):
LGTM

PR #854, opened 2021-08-22
Expand DSF map unit tests

johnwlambert:
Add more checks to make sure wrapped DSFMap behaves as expected.


varunagrawal:
Can you set the target branch to be my branch in #850? That way all the tests should work before and after and we can just merge 850 into develop.

johnwlambert:
thanks for the review Varun. When I changed the base to #850, the file diff became quite unclear in the UI. I think it's clearer for the merge history like this.

varunagrawal (approved):
I guess this branch doesn't base of the latest develop, but that's okay. LGTM, let's get this ball rolling.

PR #857, opened 2021-08-23
Add section about Boost version requirement

varunagrawal:
As per the discussion in #850.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -13,7 +13,8 @@ $ make install
 ## Important Installation Notes
 
 1. GTSAM requires the following libraries to be installed on your system:
-    - BOOST version 1.65 or greater (install through Linux repositories or MacPorts)
+    - BOOST version 1.65 or greater (install through Linux repositories or MacPorts). Please see [Boost Notes](#boost-notes).
+
     - Cmake version 3.0 or higher
     - Support for XCode 4.3 command line tools on Mac requires CMake 2.8.8 or higher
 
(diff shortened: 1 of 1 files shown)


dellaert:
What about allowing 1.58 on 16.04 and printing a warning via cmake?

varunagrawal:
Those kind of edge cases get messy especially since Ubuntu 16 is almost at the end of it's life term.

dellaert:
> Those kind of edge cases get messy especially since Ubuntu 16 is almost at the end of it's life term.

 This is what we had discussed, however. If you disagreed, why not mention it in the other thread?? End of life or not, many robots are indeed still at 16.04.

varunagrawal:
CMake doesn't provide a way to get the underlying OS type and version aka Ubuntu 16.04. We would have to [hack something up](https://stackoverflow.com/questions/26919334/detect-underlying-platform-flavour-in-cmake) and it doesn't seem to be worth the effort.

My opinion is that for 16.04, installing Boost 1.65 via a 3rd party PPA such as [mhier](https://launchpad.net/~mhier/+archive/ubuntu/libboost-latest) is a lot cleaner and something that end users can easily accomplish. It's literally 2 extra lines in their setup scripts.

dellaert:
Sounds good. Maybe add that info to the readme.

varunagrawal:
Already did. I added a line about using alternative channels if the main package manager doesn't have support for the specified version.

dellaert (approved):

PR #858, opened 2021-08-23
Add missing import of Pose3

varunagrawal:
Fixes bug accidentally introduced by #844.

### DIFF
--- a/python/gtsam/utils/plot.py
+++ b/python/gtsam/utils/plot.py
@@ -10,7 +10,7 @@ from matplotlib import patches
 from mpl_toolkits.mplot3d import Axes3D  # pylint: disable=unused-import
 
 import gtsam
-from gtsam import Marginals, Point3, Pose2, Values
+from gtsam import Marginals, Point3, Pose2, Pose3, Values
 
 
 def set_axes_equal(fignum: int) -> None:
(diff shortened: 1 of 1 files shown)


johnwlambert (approved):
Thanks for catching the missing import.

I am using `python black` https://github.com/psf/black. I find it to be the cleanest of all the Python auto-formatters. Could you take a look at it? If we could switch to it, I would be quite happy : - )

Python black is used by 73K repos. Yapf is used by just 22K repos, according to Github

varunagrawal:
Yeah Github stars and users don't mean much since we follow the Google Style for Python (which recommends yapf).

varunagrawal:
I also like black because it is very opinionated. Unfortunately, formatting all the python code is a chore I don't want to undertake any time soon.

ProfFan:
What is the biggest advantage of black over yapf?

johnwlambert:
> What is the biggest advantage of black over yapf?

I think the function signatures are much easier to read with Black. For one example, YAPF formats this as:
```python
def ellipsoid(rx: float, ry: float, rz: float,
              n: int) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
```
but Black formats it as:
```python
def ellipsoid(
    rx: float, ry: float, rz: float, n: int
) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
```

I find the latter a lot easier to read and review.

Black is also quite fast: https://news.ycombinator.com/item?id=17155048

PR #859, opened 2021-08-24
make templates for smart projection factors uniform (all templated on cameras)

lucacarlone:
This is a preliminary PR that will allow using other camera models (e.g., omnidirectional instead of pinhole) in the smart projection factors. This PR involves changing the smart factor template from CALIBRATION to CAMERA (strangely, the template argument is already CAMERA in the base classes, but becomes CALIBRATION in the final factors). Factors to be changed:

- [ ] SmartProjectionPoseFactor
- [ ] possibly simplify all Stereo factors by templating on StereoCamera

### DIFF
--- a/gtsam/base/Lie.h
+++ b/gtsam/base/Lie.h
@@ -17,6 +17,7 @@
  * @author Frank Dellaert
  * @author Mike Bosse
  * @author Duy Nguyen Ta
+ * @author Yotam Stern
  */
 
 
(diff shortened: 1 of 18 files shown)


lucacarlone:
@dellaert : I was planning to start on this PR sometime this week, but I wanted to get your opinion before I start. I don't remember why we templated some smart factors on CALIBRATION (besides the practical reason that all projects used pinhole camera at the time).

PR #860, opened 2021-08-27
Basis - Python unit test

gchenfc:
Python unit test for FitBasis in basis.h, which was added in #403 

Replaces #838 

This tests:
* FourierBasis
* Chebyshev1Basis
* Chebyshev2Basis
* Chebyshev2
* FitBasis<T> where T is the above 4 bases.

I added `#include<pybind11/stl.h>` to be able to construct FitBasisXXX with a python dictionary.  There are other ways to accomplish this but I think this is easiest and most transparent.


gchenfc:
Will merge in as soon as CI passes

PR #863, opened 2021-08-28
Fix Python Tests Cmake

varunagrawal:
Updated the CMake so that the tests are verbose and easy to debug.

I also have some doubts about the working directory for the tests so updated that. Ever since I've done that, I haven't encountered any segfaults (was getting plenty of these before).

@gchenfc I am merging this into your `tests/basis_python` branch so we can debug whatever is going on there.


varunagrawal:
@gchenfc so your python unit test actually unveiled a bug in `FourierBasis`. I fixed the bug and also formatted the tests a bit so that they report which particular case fails a bit better. The CI should now pass!

gchenfc (approved):
Awesome, tysm!

Since this is pointing to `tests/basis_python` I'll merge this in.

PR #869, opened 2021-09-05
ImuFactor Fixes

varunagrawal:
+ wrapped `smart` flag for noise models.


dellaert (approved):
Cool. Would be nice to add a screenshot in the PR comment

varunagrawal:
Plots are exactly the same. I was just getting a lot of terminal noise due to the deprecation warnings so I resolved that.

The wrapping happened more as a consequence of debugging IMU sequences with noise so I could check the various covariances.

PR #874, opened 2021-09-16
CombinedImuFactor: Add bias effect on position jacobian

### LINKED ISSUE #368: CombinedImuFactor with just one measurement leads to degeneracy
## Description
If only 1 measurement is integrated into the **CombinedImuFactor**, the position related covariance terms are all zero.
This only seems to happen for the **CombinedImuFactor** though. If the same single measurement is integrated into an **ImuFactor** all elements of the covariance matrix is filled out. No zero rows/columns.

Unfortunately, these zero rows in the covariance leads to degeneracy and the infamous 'Indeterminant linear system' error.

## Steps to reproduce
1. Create a **PreintegratedCombinedMeasurements** 
2. Integrate 1 IMU measurement into a the measurement
3. Create a **CombinedImuFactor** with the measurement
4. Add the factor and necessary initializations to a **NonlinearFactorGraph**
5. Perform optimization

In general the problem can be reproduced with the [**ImuFactorsExample.cpp**](https://github.com/borglab/gtsam/blob/b1bb0c9ed58f62638c068d7b5332fe7e0e49a29b/examples/ImuFactorsExample.cpp) by ensuring that only 1 measurement is integrated: _preintegrated->integrateMeasurement(imu.head<3>(), imu.tail<3>(), dt);_

## Expected behavior
Even with just one measurement I would expect at the very least all diagonal element to be non-zero of the preintegrated factor covariance, just as is the case if just one measurement is integrated into the **ImuFactor**.

## Additional information
More information can be found here: https://groups.google.com/forum/#!msg/gtsam-users/WskPzqexiNk/UwQrbBe7AAAJ

varunagrawal:
Fixes #368 

TODO
- [x] Add a unit test to capture this.
- [x] Update ImuFactor.pdf with details about the jacobians for CombinedImuFactor
- [x] Add details about Integration Covariance to ImuFactor.pdf.


varunagrawal:
@dellaert I need some help regarding the integration covariance. Can you point me to a resource that can help explain the mathematical basis for it?

dellaert:
> @dellaert I need some help regarding the integration covariance. Can you point me to a resource that can help explain the mathematical basis for it?

No, @lucacarlone and @cfo did all this. The integration covariance was always a bit of a mystery to me as well ;-) Your best bet is @lucacarlone 's TR which I think you already asked for.

varunagrawal:
> > @dellaert I need some help regarding the integration covariance. Can you point me to a resource that can help explain the mathematical basis for it?
> 
> No, @lucacarlone and @cfo did all this. The integration covariance was always a bit of a mystery to me as well ;-) Your best bet is @lucacarlone 's TR which I think you already asked for.

Gotcha! That's the only thing left to document before this PR is ready. :slightly_smiling_face:

gtsam/navigation/tests/testCombinedImuFactor.cpp
```diff
@@ -203,6 +203,48 @@ TEST(CombinedImuFactor, PredictRotation) {
...
+      0, 0, 0, 0, 0, 0, 0, 0.010001, 0, 0, 0, 0, 0, 0, 0,      //
+      0, 0, 0, 0, 0, 0, 0, 0, 0.010001, 0, 0, 0, 0, 0, 0,      //
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0.01, 0, 0, 0, 0, 0,          //
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.01, 0, 0, 0, 0,          //
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.01, 0, 0, 0,          //
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.01, 0, 0,          //
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.01, 0,          //
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.01;
+
+  // regression
```
dellaert, line 244: So, I think a "regression" is not enough here. It's easy to make a mistake, so how do we know the code is correct?

lucacarlone:
> > > @dellaert I need some help regarding the integration covariance. Can you point me to a resource that can help explain the mathematical basis for it?
> > 
> > 
> > No, @lucacarlone and @cfo did all this. The integration covariance was always a bit of a mystery to me as well ;-) Your best bet is @lucacarlone 's TR which I think you already asked for.
> 
> Gotcha! That's the only thing left to document before this PR is ready. 🙂

I shared the report, but I suspect it will not be very useful (it mostly covers the jacobians). I'm happy to write down the math if this is not super-urgent (I cannot do that before sometimes in October).

gtsam/navigation/tests/testCombinedImuFactor.cpp
varunagrawal: That's a great point which I have been wondering about myself. `ImuFactor` has a similar test so the same question can be asked about that factor.

I should add sub-tests that check for these things individually.

varunagrawal:
@lucacarlone yes that works. We can update the document later since this PR is focused on resolving #368. 🙂

varunagrawal:
@dellaert so I've spent some more time trying to test the preintegration covariance, and I've found a lot of inconsistencies.

1. For the same initial conditions and measurements, the `ImuFactor` and the `CombinedImuFactor` give me significantly different covariances (6 orders of magnitude).
2. There are some off-diagonal elements missing in the covariance computation for `CombinedImuFactor`. This is easy enough to add in.

Unfortunately, without understanding the math behind the preintegration covariance, I cannot be certain whether any operation is the correct one to do.

varunagrawal:
This is the preintegration covariance for the ImuFactor
```
3.0625e-10          0          0          0          0          0          0          0          0
         0 3.0625e-10          0          0          0          0          0          0          0
         0          0 3.0625e-10          0          0          0          0          0          0
         0          0          0    2.5e-11          0          0      5e-09          0          0
         0          0          0          0    2.5e-11          0          0      5e-09          0
         0          0          0          0          0    2.5e-11          0          0      5e-09
         0          0          0      5e-09          0          0      1e-06          0          0
         0          0          0          0      5e-09          0          0      1e-06          0
         0          0          0          0          0      5e-09          0          0      1e-06
```

and for the same numbers, we get for CombinedImuFactor
```
0.01  0     0      0           0           0           0        0        0        0    0    0    0    0    0
0     0.01  0      0           0           0           0        0        0        0    0    0    0    0    0
0     0     0.01   0           0           0           0        0        0        0    0    0    0    0    0
0     0     0      2.50025e-07 0           0           0        0        0        0    0    0    0    0    0
0     0     0      0           2.50025e-07 0           0        0        0        0    0    0    0    0    0
0     0     0      0           0           2.50025e-07 0        0        0        0    0    0    0    0    0
0     0     0      0           0           0           0.010001 0        0        0    0    0    0    0    0
0     0     0      0           0           0           0        0.010001 0        0    0    0    0    0    0
0     0     0      0           0           0           0        0        0.010001 0    0    0    0    0    0
0     0     0      0           0           0           0        0        0        0.01 0    0    0    0    0
0     0     0      0           0           0           0        0        0        0    0.01 0    0    0    0
0     0     0      0           0           0           0        0        0        0    0    0.01 0    0    0
0     0     0      0           0           0           0        0        0        0    0    0    0.01 0    0
0     0     0      0           0           0           0        0        0        0    0    0    0    0.01 0
0     0     0      0           0           0           0        0        0        0    0    0    0    0    0.01
```

I know what is causing the difference, I just need to understand the rationale behind it.

dellaert:
> Unfortunately, without understanding the math behind the preintegration covariance, I cannot be certain whether any operation is the correct one to do.

So, is this (a) the manifold or (b) tangent-space formulation that has issues?
If (a), the equations in Forster17tro should describe this.
If (b), the section Noise Propagation in ImuFactor.pdf should describe this.

You should not do any operation/correction unless it is described there. If somethings is done in the code that is *not* described in either document, what is it?

PS It might be easiest to start with tangent-space formulation.

dellaert:
Also, which unit test fails - excluding regression tests?

varunagrawal:
I'm currently working explicitly in the tangent space formulation, but this is an issue in the noise propagation at the factor level so I imagine it will show even in the manifold formulation.

None of the math regarding the covariance of the CombinedImuFactor exists in ImuFactor.pdf. What's confusing is the interaction of the integration covariance with all the Jacobians which we don't have any documentation for yet.

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -42,6 +42,8 @@ typedef ManifoldPreintegration PreintegrationType;
  * L. Carlone, Z. Kira, C. Beall, V. Indelman, F. Dellaert, Eliminating
  * conditionally independent sets in factor graphs: a unifying perspective based
  * on smart factors, Int. Conf. on Robotics and Automation (ICRA), 2014.
+ * 
+ * [3] is available in this repo as "PreintegratedIMUJacobians.pdf".
```
dellaert, line 46: Why not say this below where [3] is defined?

gtsam/navigation/ImuFactor.h
```diff
@@ -44,6 +44,8 @@ typedef ManifoldPreintegration PreintegrationType;
  * C. Forster, L. Carlone, F. Dellaert, D. Scaramuzza, "IMU Preintegration on
  * Manifold for Efficient Visual-Inertial Maximum-a-Posteriori Estimation",
  * Robotics: Science and Systems (RSS), 2015.
+ * 
+ * [3] is available in this repo as "PreintegratedIMUJacobians.pdf".
```
dellaert, line 48: Same: say this below

gtsam/navigation/tests/testCombinedImuFactor.cpp
dellaert: Yeah, especially with the non-trivial changes here, we need to think about how we can test, maybe using numbers also derivatives

dellaert (changes requested):
Need to resolve the testing issue before we can land

varunagrawal:
@lucacarlone yes last PR. Just the one file in `CombinedImuFactor.cpp` needs to be reviewed since that is the main change here.

lucacarlone:
@varunagrawal : to be honest I'm a bit lost among these PRs. the PR says this mostly updates the pdf, but you mention CombinedImuFactor.cpp is what's important?

varunagrawal:
The original PR comment is definitely outdated.

lucacarlone (approved):
I'm approving this, but haven't re-reviewed the math.

varunagrawal:
@lucacarlone thank you! That should be fine since the math is the same from the prior PRs you reviewed and I went from math -> code + added lots of unit tests.

@dellaert just need you to unblock this PR now and we can land it!!! :smile:

varunagrawal:
@dellaert I undid the name change from `biasAccOmegaInt` to `biasAccOmegaInit`. Will make another PR for that once this is merged so we can land this one.

varunagrawal:
@dellaert

dellaert:
Thanks for isolating the name change. Will approve now.

dellaert (approved):

PR #884, opened 2021-10-04
Add constructor for OptionalJacobian

gchenfc:
The fixed-size version of `OptionalJacobian` doesn't have a constructor for pointers to dynamic matrices.  This makes it cumbersome to deal with e.g.:

```c++

double myFunc(double x, gtsam::OptionalJacobian<6, 6> H = boost::none) { ... }

class MyFactor : public gtsam::NoiseModelFactor {

  ...

  Vector unwhitenedError(const Values& x,
        boost::optional<std::vector<Matrix>&> H = boost::none) const override {
    double y = x.at<double>(keys_[0]);
    // myFunc(y, H ? &(*H)[0] : nullptr); //< this doesn't compile

    return H ? myFunc(y, (*H)[0])
             : myFunc(y); //< this does compile but is not as nice
  }
};

```

Frank recommends using the former "modern" syntax but this isn't possible without adding the constructor overload:

```c++
OptionalJacobian(Eigen::MatrixXd* dynamic) {...}
```

For future readers who are less familiar, ternary operator must have same-types for the middle and last parameters, and since the last parameter is a null then the middle parameter (and the whole output-type) has to be a pointer type.

No breaking changes and nothing removed, just added a new constructor overload.

### DIFF
--- a/gtsam/base/OptionalJacobian.h
+++ b/gtsam/base/OptionalJacobian.h
@@ -89,6 +89,13 @@ public:
     usurp(dynamic.data());
   }
 
+  /// Constructor that will resize a dynamic matrix (unless already correct)
+  OptionalJacobian(Eigen::MatrixXd* dynamic) :
+      map_(nullptr) {
+    dynamic->resize(Rows, Cols); // no malloc if correct size
+    usurp(dynamic->data());
+  }
+
 #ifndef OPTIONALJACOBIAN_NOBOOST
 
   /// Constructor with boost::none just makes empty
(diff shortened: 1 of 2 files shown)


varunagrawal:
I'm wondering if we can use boost::none instead of nullptr.

varunagrawal (approved):
LGTM

gchenfc:
Ya, in this case I used `nullptr` to stay consistent with the rest of the file, but I guess I usually use nullptr for brevity - do you think there's enough advantage that I should use `boost::none` in the future?

PR #885, opened 2021-10-04
`Pose3::Adjoint(xi)` Jacobians

### LINKED ISSUE #433: AdjointMap jacobians
## Feature

Lie Groups have `AdjointMap` implemented.

Just as (lowercase) `Pose3.adjointMap` has functions for
```c++
Matrix66 adjointMap() { ... }
Vector6 adjoint(Vector6 y) { ... }
Vector6 adjoint(Vector6 y, OptionalJacobian H) { ... }
```

It would be useful to have the equivalent one with jacobians for (uppercase) `AdjointMap`.  Also, I propose to add another argument:

```c++
Vector Adjoint(Vector6 y, OptionalJacobian H, OptionalJacobian H_y) { ... }
```

to avoid having to calculate the AdjointMap again.

Unless there is a reason against this, I can begin working on this.

## Motivation

Factors which need to convert tangent vectors to different frames will need the derivatives of AdjointMap.  For example, in GTDynamics many factors such as TwistFactor, TwistAccelFactor, and WrenchEquivalenceFactor need this.  Currently, GTDynamics is using an ugly way that would become much cleaner if this were implemented, at least for Pose3.

## Notes

Currently only Pose3 has a function `Adjoint(Vector6 y) { return AdjointMap() * y; }`, and also Pose3's `AdjointMap` has a comment " /// FIXME Not tested - marked as incorrect".

gchenfc:
Closes #433 

This PR implements the Jacobians for `Pose3::Adjoint(x)` := `Pose3::AdjointMap() * x`.

It uses `Rot3.rotate` and `cross(Point3, Point3)` to calculate intermediate derivatives.

This should help in GTDynamics to eliminate the `AdjointMapJacobianQ` function and allow it to be generalized to other joint types.


gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,81 @@ Matrix6 Pose3::AdjointMap() const {
...
+  Rw = R_.rotate(w, H_this ? &Rw_H_R : nullptr, H_xib ? &Rw_H_w : nullptr);
+  const Vector3 Rv =
+      R_.rotate(v, H_this ? &Rv_H_R : nullptr, H_xib ? &Rv_H_v : nullptr);
+  const Vector3 pRw = cross(t_, Rw, H_this ? &pRw_H_t : nullptr,
+                            (H_this || H_xib) ? &pRw_H_Rw : nullptr);
+  result.tail<3>() = pRw + Rv;
+
+  // Jacobians
+  if (H_this) {
+    pRw_H_t = Rw_H_R;  // This is needed to pass the unit tests for some reason
```
varunagrawal, line 88: We should figure this out before landing this.

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -145,6 +145,75 @@ TEST(Pose3, Adjoint_full)
...
+{
+  Matrix6 actualH1, actualH2, expectedH1, expectedH2;
+  std::function<Vector6(const Pose3&, const Vector6&)> Adjoint_proxy =
+      [&](const Pose3& T, const Vector6& xi) {
+        Vector6 res1 = T.AdjointMap() * xi, res2 = T.Adjoint(xi);
+        EXPECT(assert_equal(res1, res2));
+        return res1;
+      };
+
+  gtsam::Vector6 xi =
```
varunagrawal, line 160: Use the syntax `xi(1, 2, 3, 4, 5, 6)`?

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -145,6 +145,75 @@ TEST(Pose3, Adjoint_full)
...
+  Matrix6 actualH1, actualH2, expectedH1, expectedH2;
+  std::function<Vector6(const Pose3&, const Vector6&)> AdjointTranspose_proxy =
+      [&](const Pose3& T, const Vector6& xi) {
+        Vector6 res1 = T.AdjointMap().transpose() * xi,
+                res2 = T.AdjointTranspose(xi);
+        EXPECT(assert_equal(res1, res2));
+        return res1;
+      };
+
+  gtsam::Vector6 xi =
```
varunagrawal, line 195: Same as above

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -145,6 +145,75 @@ TEST(Pose3, Adjoint_full)
...
 
+/* ************************************************************************* */
+// Check Adjoint numerical derivatives
+TEST(Pose3, Adjoint_jacobians)
+{
+  Matrix6 actualH1, actualH2, expectedH1, expectedH2;
+  std::function<Vector6(const Pose3&, const Vector6&)> Adjoint_proxy =
+      [&](const Pose3& T, const Vector6& xi) {
+        Vector6 res1 = T.AdjointMap() * xi, res2 = T.Adjoint(xi);
+        EXPECT(assert_equal(res1, res2));
```
varunagrawal, line 156: It's not super clean to add a check inside a functional. You want to be able to isolate issues with results and Jacobians.

varunagrawal (commented):
Some comments.

gtsam/geometry/Pose3.cpp
dellaert: Uhm, yes. Need some math. In math.pdf?

gtsam/geometry/tests/testPose3.cpp
gchenfc: Thanks!  For some reason I thought this only worked up to Vector3 but good to know I was just totally wrong :)

gchenfc: Actually I'm having trouble getting this to work - do you have a reference?

I can't seem to get the initializer-list constructor to work for some reason:

[Documentation](https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html)
> Matrices and vectors can also be initialized from lists of coefficients. Prior to C++11, this feature is limited to small fixed-size column or vectors up to size 4:
...
> If C++11 is enabled, fixed-size column or row vectors of arbitrary size can be initialized by passing an arbitrary number of coefficients:
... (brace initializer lists)

Trying
```c++
gtsam::Vector6 xi {0.1, 1.2, 2.3, 3.1, 1.4, 4.5};
gtsam::Vector6 xi = {0.1, 1.2, 2.3, 3.1, 1.4, 4.5};
gtsam::Vector6 xi {{0.1, 1.2, 2.3, 3.1, 1.4, 4.5}};
```
but none seem to be working :(

gtsam/geometry/tests/testPose3.cpp
gchenfc: Ya this has bitten me before - I'll switch to copy-pasting this line

gtsam/geometry/tests/testPose3.cpp
gchenfc: Update: I think we must be on an old version of Eigen that doesn't have this initializer - the [latest Eigen Matrix.h file](https://eigen.tuxfamily.org/dox/Matrix_8h_source.html) Lines 286-322 have the initializer_list code but in our 3rdparty folder those lines are missing.

In light of this, I'm not going to change these lines, ok?

Please resolve if this is ok with you.

gtsam/geometry/tests/testPose3.cpp
gchenfc: done

gtsam/geometry/tests/testPose3.cpp
gchenfc: just for posterity (my own notes) initializer list constructor wasn't added until at least 3.3.8, started by this commit 2 years ago: https://gitlab.com/libeigen/eigen-backup/-/commit/db152b9ee6effd3799f70a621f495c427cb3c33f

3.3.7 is 3 years old now.

varunagrawal: Use parentheses not curly braces. That works perfectly.

gchenfc: didn't work for me, is this GTSAM custom constructor or eigen?

Can you confirm that you have tried compiling?  I spent 2 hours this morning trying to make this work by your suggestion and would rather not spend 2 more before having the peace of mind that you confirmed it working on your machine.

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -145,6 +145,81 @@ TEST(Pose3, Adjoint_full)
...
+
+  // Check evaluation sanity check
+  EQUALITY(static_cast<gtsam::Vector>(T.AdjointMap() * xi), T.Adjoint(xi));
+  EQUALITY(static_cast<gtsam::Vector>(T2.AdjointMap() * xi), T2.Adjoint(xi));
+  EQUALITY(static_cast<gtsam::Vector>(T3.AdjointMap() * xi), T3.Adjoint(xi));
+
+  // Check jacobians
+  Matrix6 actualH1, actualH2, expectedH1, expectedH2;
+  std::function<Vector6(const Pose3&, const Vector6&)> Adjoint_proxy =
+      [&](const Pose3& T, const Vector6& xi) { return T.AdjointMap() * xi; };
```
varunagrawal, line 162: I know this is the same result but for readability, this should return `T.Adjoint(xi)`.

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -145,6 +145,81 @@ TEST(Pose3, Adjoint_full)
...
+  EQUALITY(static_cast<Vector>(T2.AdjointMap().transpose() * xi),
+           T2.AdjointTranspose(xi));
+  EQUALITY(static_cast<Vector>(T3.AdjointMap().transpose() * xi),
+           T3.AdjointTranspose(xi));
+
+  // Check jacobians
+  Matrix6 actualH1, actualH2, expectedH1, expectedH2;
+  std::function<Vector6(const Pose3&, const Vector6&)> AdjointTranspose_proxy =
+      [&](const Pose3& T, const Vector6& xi) {
+        return T.AdjointMap().transpose() * xi;
```
varunagrawal, line 201: Same here.

gtsam/geometry/Pose3.cpp
varunagrawal: It seems like the way you're doing the `cross` operation is making the jacobian be incorrect. The jacobian should be `d (p_cross * R * w) / dp`.

By the chain rule for matrix-matrix-products [Matrix Cookbook Eq 37](https://www.math.uwaterloo.ca/~hwolkowi/matrixcookbook.pdf), we get `R * dp_cross/dp * w + p_cross * dR/dp * w = R * dp_cross/dp * w`.

We have `p_cross` as:
```
p_cross = 
 0 -z  y
 z  0 -x
-y  x  0
```

If we differentiate the skew-symmetric matrix with each element, we get:
```
dp_cross / dx = 
0 0  0
0 0 -1
0 1  0

dp_cross / dy = 
0  0 1
0  0 0
-1 1 0

dp_cross / dz = 
0 -1 0
1  0 0
0  0 0
```

Just for `dx` we get
```
R11 R12 R13 * 0 0  0 * wx
R21 R22 R23   0 0 -1   wy
R31 R32 R33   0 1  0   wz

=
0  R13  -R12   wx
0  R23  -R22 * wy
0  R33  -R32   wz

=
R13*wy - R12*wz
R23*wy - R22*wz
R33*wy - R32*wz
```

which is the same as `R * (-w_cross)`.
```
R11 R12 R13 *  0   wz  -wy
R21 R22 R23   -wz  0    wx
R31 R32 R33    wy -wx   0

=
(-R12*wz + R13*wy)
(-R22*wz + R23*wy)
(-R32*wz + R33*wy)
```

This is why using `Rw_H_R` works, since it is just `R * (-w_cross)`.


#### Possible Solutions

1. Compute the jacobian `pRw_H_t` separately as `R * (-w_cross)`.
2. Implement the cross product for `Rot3` with a `Point3` along with the appropriate jacobians.

I personally prefer (1) since it is easier, though we will have to make a note about the reason for it (which can be just the comments from here).

varunagrawal:
I merged the latest `develop` since it has some CI fixes.

varunagrawal (approved):
Fixed various stuff so LGTM.

gtsam/geometry/Pose3.cpp
gchenfc: Thanks for this analysis!

But I have a couple questions:

1) >By the chain rule for matrix-matrix-products Matrix Cookbook Eq 37,...
    
    did you mean product rule?  Eq 37 states \partial{ XY } = (\partial{X})Y + X(\partial{Y})
2) >... we get `R * dp_cross/dp * w + p_cross * dR/dp * w = R * dp_cross/dp * w`.  
    
    I'm not sure I understand this - Eq 37 doesn't switch the order of X and Y so I think it should be (using [] to denote skew symmetric matrix):
    image
    instead of yours (typeset to be easier to read):
    image
    And then I think if you switch the order of \partial{[t]}R then I think we can simplify (abusing the skew-symmetric notation to apply to matrices to indicate "column-wise" and forming a tensor):
    image
    (this is I think the same identity that you maybe noticed: `R[I]w = -RI[w]` where I is 3x3 identity)
    And this expression is the same expression in `Point3::cross`: `if (H1) *H1 << skewSymmetric(-q.x(), -q.y(), -q.z());`
3) `Rw` is an ordinary 3-vector so why wouldn't `t x (Rw)` be able to use the cross product Jacobians?  I think the explanation you give would give different results if you tried to do `\partial (tR) x w` vs `\partial t x (Rw)`, which of course these should give the same result

I suspect the actual problem with this jacobian is that `Pose3::Local` is NOT just `Vector6() << R_.Local(), t_.Local()`.  Instead, it's applying some stuff to `t_` (see `Pose3::Logmap`).  But the problem with this hypothesis is that turning off `GTSAM_POSE3_EXPMAP` does not resolve this, so I'm still working on figuring out what's going on.

PS: the same issue occurs in [`AdjointTranspose` calculating `H_this`](https://github.com/borglab/gtsam/blob/2de623befccc6cce0b064737053e3ca4d9a72389/gtsam/geometry/Pose3.cpp#L132-L136) and I'm not sure if the same reasoning applies to this one because this code does the cross product before doing the rotation.

gchenfc: Following up on my "edit":

> I forgot that T.localCoordinates is between-ing T then applying ChartAtOrigin so this is where a multiplication by R gets introduced.

`-[Rw] * R = -(Rw) x R = R[w]`
That last part can be intuited by splitting up image where `R_i` denotes the i'th column of R.  Then geometrically you can imagine that `-w_i(R_i x R)` will generate something similar to `R[w]` since `R_i x R_j` will be 0 if i==j and `+/-R_k` otherwise (where k =/= i =/= j).  Carefully working through each example we can confirm that indeed it is exactly `R[w]`.

This correction (multiply by R) works for both `Adjoint` and `AdjointTranspose` so I think this was indeed the issue.

In summary, for some arbitrary `X`, `[dX / dT] = [dX/dR ; dX/dt * R]` where the pose `T` has rotation `R` and translation `t`, where the extra `R` is the part I was missing.  Furthermore, some algebraic manipulation can simplify the expressions.

I'll write this up and simplify.

varunagrawal: You're right about the ordering for the product rule (I don't know why I called it the chain rule, maybe because it is a chain of operations?). This [stackexchange answer](https://math.stackexchange.com/a/366931/134918) was what gave me the equation and I just looked up the equation number.

varunagrawal:
I am wondering why you made the tolerances tighter. I made them looser because the CI wasn't passing, and this is because the scale of the values in the matrix were quite different, so it's more of a numerical stability issue than anything else.

gchenfc:
Ya idk why I made the tolerances higher - that was stupid - late night fogginess

gtsam/geometry/Pose3.cpp
gchenfc: > the issue is exactly that the multiplication by R is not happening since the cross product is not aware of that operation having happened.

It doesn't need to be aware of that multiplication?  `Rw` is an ordinary 3-vector so I don't really see why it matters where `Rw` came from when calculating jacobian w.r.t. `t` since `Rw` is independent of `t` (but if there were a dependence, then we would use product rule and include a term `+ pRw_H_Rw * Rw_H_t` and we still wouldn't need to do this "manual" matrix math work).  Basically, I don't think we should ever need to drop into the matrix math as long as we just follow chain rule.  I think this is most clear in commit 33e16aa, but then in f2ac90d I made the tradeoff to be more efficient at the cost of readability by applying the simplification that `pRw_H_t * t_H_xi = pRw_H_t * R = Rw_H_R` by some matrix magic.

> I think we came to the same conclusion but by different means. I'll spend more time understanding your comment and see what is the key difference.

I don't think so - I don't think the issue is related to cross product at all, but also feel free to check out the math.pdf updates I made in Sec. 13.2.  The key sentence for this discussion being "Applying chain rule for the translation ...".

### Simplest Explanation
`H_this` does not equal `[H_R , H_t]`, it actually equals `[H_R , H_t] * [R_H_this ; t_H_this]`.  Basically, I assumed that the "vector representation for Pose3" (aka Local) was just R stacked on top of t (which is false) and therefore I forgot about the partial derivatives w.r.t. R, t.

If you look at [`translation(Hself)` and `rotation(Hself)`](https://github.com/borglab/gtsam/blob/f2ac90d5b9682f4b78d8ffe99be378ed32df11c5/gtsam/geometry/Pose3.cpp#L344-L357) (reproduced below), you will see that actually `translation(Hself)` is _not_ `[0 I]`, it is `[0 R]`.

https://github.com/borglab/gtsam/blob/f2ac90d5b9682f4b78d8ffe99be378ed32df11c5/gtsam/geometry/Pose3.cpp#L344-L357

### Explanation for why `translation(Hself)` is `[0 R]` (optional read if you're curious, but not necessary to understand)
Recall that when we do derivative w.r.t. a manifold object, what we actually mean is something along the lines of:
image
Where `f(a)` is the function we're trying to differentiate (`this.Adjoint() * xi_b` in this case) and `R_a(xi)` denotes the retraction at pose `a` using local coordinates `xi`.  This is **NOT** mathematically precise, just what I thought was a more intuitive version of the formal definition in Section 6.2 of math.pdf:
image

The key being that it's the retraction *at `a`* (aka `this`).  This is significant because the jacobian we calculated was w.r.t. the translation in the world frame, but actually we want the translation in the frame of `this`.   We would be able to see this either in `retract` or `localCoordinates` of `Lie.h`:
https://github.com/borglab/gtsam/blob/9798bfa8151788b31e225be9acce0ca4c4d30b10/gtsam/base/Lie.h#L140-L161

Notice how there is a `compose` and `between` in `retract` / `localCoordinates` respectively, so although the ChartAtOrigin does indeed just stack `xi = [R t]`, we have to transform this into the chart at `a` by rotating the translation by `R.transpose()`.  The transpose goes away because in the jacobian the translation is across columns so instead of `R.transpose * xi` it's more like `xi.transpose() * R` (colloquially - it's hard to write in text but I'm sure you can figure this out).

gchenfc: This is probably the clearest code, from the state of the code at 33e16aa

https://github.com/borglab/gtsam/blob/33e16aa7d2ba47239ec59439da997238d642b8bd/gtsam/geometry/Pose3.cpp#L67-L99

doc/math.lyx
```diff
@@ -5084,6 +5084,185 @@ reference "ex:projection"
 
 \end_layout
 
+\begin_layout Subsection
+Derivative of Adjoint
```
dellaert, line 5088: Did you read section 3 in detail? In this document we are trying to take a different view, eschewing the partial derivative notation. I’m happy to spend some time trying to rewrite it with you…

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,92 @@ Matrix6 Pose3::AdjointMap() const {
   return adj;
 }
 
+/* ************************************************************************* */
+// Calculate AdjointMap applied to xi_b, with Jacobians
+Vector6 Pose3::Adjoint(const Vector6& xi_b, OptionalJacobian<6, 6> H_this,
```
dellaert, line 68: Call this H_pose not H_this?

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,92 @@ Matrix6 Pose3::AdjointMap() const {
...
+    // pRw_H_thisv = pRw_H_t * R = [Rw]x * R = R * [w]x = Rw_H_R
+    //    where [ ]x denotes the skew-symmetric operator.
+    // See docs/math.pdf for more details.
+    const Matrix3 &pRw_H_thisv = Rw_H_R;
+    *H_this = (Matrix6() << Rw_H_R, /*               */ Z_3x3,  //
+               /*        */ pRw_H_Rw * Rw_H_R + Rv_H_R, pRw_H_thisv)
+                  .finished();
+  }
+  if (H_xib) {
+    *H_xib = (Matrix6() << Rw_H_w, /*      */ Z_3x3,  // note: this is Adjoint
```
dellaert, line 98: This is just AdjointMap

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,92 @@ Matrix6 Pose3::AdjointMap() const {
...
+              /*        */ pRw_H_Rw * Rw_H_w, Rv_H_v)
+                 .finished();
+  }
+
+  // Return
+  return result;
```
dellaert, line 104: This is just AdjointMap * xi

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,92 @@ Matrix6 Pose3::AdjointMap() const {
...
+
+/* ************************************************************************* */
+/// The dual version of Adjoint
+Vector6 Pose3::AdjointTranspose(const Vector6& x, OptionalJacobian<6, 6> H_this,
+                                OptionalJacobian<6, 6> H_x) const {
+  //  Ad^T   *   xi   =   [ R^T  R^T.[-t]  .  [w
+  //                         0     R^T ]       v]
+  // Declarations, aliases, and intermediate Jacobians easy to compute now
+  Vector6 result;
+  const Vector3 &w = x.head<3>(), &v = x.tail<3>();
```
dellaert, line 115: Intermediate things should not be references, that is an argument thing

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,92 @@ Matrix6 Pose3::AdjointMap() const {
...
 
+/* ************************************************************************* */
+// Calculate AdjointMap applied to xi_b, with Jacobians
+Vector6 Pose3::Adjoint(const Vector6& xi_b, OptionalJacobian<6, 6> H_this,
+                       OptionalJacobian<6, 6> H_xib) const {
+  //  Ad   *   xi   =   [   R   0    .  [w
+  //                      [t]R  R ]      v]
+  // Declarations, aliases, and intermediate Jacobians easy to compute now
+  Vector6 result;
+  auto Rw = result.head<3>();
```
dellaert, line 74: This is initialized at this point

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,92 @@ Matrix6 Pose3::AdjointMap() const {
...
+  const Vector3 &w = xi_b.head<3>(), &v = xi_b.tail<3>();
+  Matrix3 Rw_H_R, Rv_H_R;
+  const Matrix3 &Rw_H_w = R_.matrix();
```
dellaert, line 77: Matrix() is expensive when quaternion so do only once

dellaert: Intermediate things should not be references, that is an argument thing

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,92 @@ Matrix6 Pose3::AdjointMap() const {
...
+  const Matrix3 &Rv_H_v = R_.matrix();
+  const Matrix3 pRw_H_Rw = skewSymmetric(t_);
+
+  // Calculations
+  Rw = R_.rotate(w, H_this ? &Rw_H_R : nullptr /*, Rw_H_w */);
+  const Vector3 Rv = R_.rotate(v, H_this ? &Rv_H_R : nullptr /*, Rv_H_v */);
+  const Vector3 pRw = cross(t_, Rw /*, pRw_H_t, pRw_H_Rw */);
```
dellaert, line 84: Does cross not have derivatives?

dellaert (changes requested):
If cross does not have derivatives, it probably should have, and they should be used below, and wherever we jump through hoops to compute them in place.

I’d like to sit down for 30 minutes after you have taken into account my comments. See how far you get with the PDF rewrite I’m asking.

gtsam/geometry/Pose3.cpp
gchenfc: Ya I figured we could save a function call and re-use intermediates this way (namely `[t]x`), hoping that the comment `note: this is Adjoint` was sufficient to clarify that this is just the same as `AdjointMap`

Shall I replace it with the marginally less efficient (?) call to AdjointMap or clarify the comment or ?

gtsam/geometry/Pose3.cpp
gchenfc: Again, trying to save computations that we already performed

gtsam/geometry/Pose3.cpp
gchenfc: Can you clarify what you mean in this comment?

`Rw` is meant to be an "output variable" which is first assigned `Rw = R_.rotate(w, ...` then later used.  This is to avoid creating an extra Vector3 and then later copying it over to the result.

gtsam/geometry/Pose3.cpp
gchenfc: Oops thanks, didn't notice the quaternion case - done.

gtsam/geometry/Pose3.cpp
gchenfc: done

doc/math.lyx
gchenfc: I will re-read in more detail.

I think the "eschewing the partial derivative notation" direction was lost on me based on the very first paragraph of "Part II: Practice" which states:
image

which made me assume that this was the preferred notation.  Perhaps we might add a note that this is _not_ the preferred way of thinking about things?

I will go through section 3 and try to emulate that style more, and send you an email when revised to find some time to go over this making sure I have the right ideas etc.

dellaert:
> @dellaert In general I think a lot of these comments are directed at my (maybe misguided) attempts at reducing copy operations, since so there's so many matrices that are the same (e.g. derivatives for all the rotate operations w.r.t. point being rotated are all the same). Also, I tried to keep the chain-rule stuff for readability, which explains why I have so many aliases (`const Matrix &`).
> 
> Would you like to:
> 
> 1. address them one-by-one,
> 2. ditch all these "optimizations" opting instead for readability (i.e. extra copies, recompute things that have already been computed, etc), or
> 3. ditch the chain-rule stuff and just use the algebraic expressions (readable code but unclear where the math came from unless reader reads math.pdf)
> 
> ?

Try to avoid calculating R.matrix many times, and don't use chain rule if it threatens do re-compute R.matrix(). But I would use cross derivatives (I checked and they exist). And then make the math.pdf match the code exactly.

It really is just [R*w; t x R*w + R*v] so:
R = R.matrix()
Rw = R * w
txRw = cross(t, Rw, ...)
if (H_pose) { ... }
return [Rw; t x Rw + R*v]

If you will pass a matrix in a derivative slot, make sure it has right size and is *not* a reference.

varunagrawal:
It's because of a const and the fact that we're not passing in a reference to the matrix.

gtsam/geometry/Pose3.cpp
varunagrawal: That shouldn't be a problem since Eigen's templating logic + the C++ compiler will optimize this to avoid that extra copy. We should optimize for readability and let the library handle the underlying complexities, especially for simple cases like this. :slightly_smiling_face:

doc/math.lyx
```diff
@@ -5082,6 +5082,190 @@ reference "ex:projection"
...
+ & =\left(D_{2}g_{(T,\xi)}(\xi)\right)\hat{\xi}g^{-1}(T,0)+g(T,0)\hat{\xi}\left(D_{2}g_{(T,\xi)}^{-1}(\xi)\right)\\
+ & =T\hat{\xi}\hat{\xi}_{b}T^{-1}-T\hat{\xi}_{b}\hat{\xi}T^{-1}\\
+ & =T\left(\hat{\xi}\hat{\xi}_{b}-\hat{\xi}_{b}\hat{\xi}\right)T^{-1}\\
+ & =-Ad_{T}(ad_{\xi_{b}}\hat{\xi})\\
+D_{1}F_{(T,\xi_{b})} & =-(Ad_{T})(ad_{\hat{\xi}_{b}})
+\end{align*}
+
+\end_inset
+
+An alternative, perhaps more intuitive way of deriving this would be to
```
dellaert, line 5207: I like this. But derivations seem to differ by a sign?

doc/math.lyx
```diff
@@ -5082,6 +5082,190 @@ reference "ex:projection"
...
+
+\begin_layout Standard
+\begin_inset Formula 
+\[
+D_{1}Ad_{T}\hat{\xi}_{b}(\xi)=D_{1}Ad_{T*I}\hat{\xi}_{b}(\xi)=Ad_{T}\left(D_{1}Ad_{I}\hat{\xi}_{b}(\xi)\right)=Ad_{T}\left(ad_{\hat{\xi}_{b}}(\xi)\right)
+\]
+
+\end_inset
+
+It's difficult to apply a similar procedure to compute the derivative of
```
dellaert, line 5232: Explain that this is for Adjoint transpose. Maybe even a new section.

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,47 @@ Matrix6 Pose3::AdjointMap() const {
...
+                       OptionalJacobian<6, 6> H_xib) const {
+  const Matrix6 Ad = AdjointMap();
+
+  // Jacobians
+  // D1 Ad_T(xi_b) = D1 Ad_T Ad_I(xi_b) = Ad_T * D1 Ad_I(xi_b) = Ad_T * ad_xi_b
+  // D2 Ad_T(xi_b) = Ad_T
+  // See docs/math.pdf for more details.
+  // In D1 calculation, we could be more efficient by writing it out, but do not
+  // for readability
+  if (H_pose) *H_pose = -Ad * adjointMap(xi_b);
```
dellaert, line 78: Oh, my god this is beautiful !

gtsam/geometry/Pose3.cpp
```diff
@@ -63,6 +63,47 @@ Matrix6 Pose3::AdjointMap() const {
...
+                                OptionalJacobian<6, 6> H_x) const {
+  const Matrix6 &AdT = AdjointMap().transpose();
+  const Vector6 &AdTx = AdT * x;
+
+  // Jacobians
+  // See docs/math.pdf for more details.
+  if (H_pose) {
+    const auto &w_T_hat = skewSymmetric(AdTx.head<3>()),
+               &v_T_hat = skewSymmetric(AdTx.tail<3>());
+    *H_pose = (Matrix6() << w_T_hat, v_T_hat,  //
```
dellaert, line 96: seems very similar to AdT

dellaert (approved):
Incredible!

gtsam/geometry/Pose3.cpp
varunagrawal: I know right? I showed this to Fan as well and mentioned how amazing this function is now.

doc/math.lyx
gchenfc: Fixed - I mixed up ad_X(Y) vs ad_Y(X)

doc/math.lyx
gchenfc: Done: made a new section

gtsam/geometry/Pose3.cpp
gchenfc: :)

gtsam/geometry/Pose3.cpp
gchenfc: Yes, similar but not quite exactly the same

```
[what, vhat
 vhat, 0   ];
```
vs adT:
```
[what, vhat,
 0,    what];
```
vs AdT:
```
[R', -R*that,
 0,     R'  ];
```

I tried for a while to figure out a cleaner representation and to figure out why it was so close to but not quite adT or AdT, but finally I gave up.  Although maybe there is some simplification, I couldn't find it nor could I figure out a way to derive this without just working through the algebra.  I tried reading up on lie coalgebras but it was just way too dense for me to parse at my current knowledge level.

PR #887, opened 2021-10-07
add failing unit test on axisAngle for Rot3 in c++

johnwlambert:


ProfFan:
Oh dang. It seems that we accidentally disabled all C++ tests? @varunagrawal

varunagrawal:
I misunderstood what the issue is. This is my fault, I updated the CI to be faster and made a small mistake...

gtsam/geometry/SO3.cpp
```diff
@@ -261,25 +261,37 @@ Vector3 SO3::Logmap(const SO3& Q, ChartJacobian H) {
...
+      omega = sgn_w * correction * (M_PI_2 / sqrt(2.0 + 2.0 * R33)) * Vector3(R31 + R13, R32 + R23, 2.0 + 2.0 * R33);
+    } else if (max_elem == 1) {
+      const double sgn_w = (R13 - R31) < 0 ? -1.0 : 1.0;
+      const double r = sqrt(2.0 + 2.0 * R22);
+      const double correction = 1.0 - M_1_PI / r * (R13 - R31);
+      omega = sgn_w * correction * (M_PI_2 / sqrt(2.0 + 2.0 * R22)) * Vector3(R21 + R12, 2.0 + 2.0 * R22, R23 + R32);
+    } else {
       // if(std::abs(R.r1_.x()+1.0) > 1e-5)  This is implicit
-      omega = (M_PI / sqrt(2.0 + 2.0 * R11)) * Vector3(1.0 + R11, R21, R31);
+      const double sgn_w = (R32 - R23) < 0 ? -1.0 : 1.0;
```
johnwlambert, line 279: thanks for working on this @ProfFan. Should we go ahead and check in the latex doc? or add this to math.pdf or something? a bit hard to derived just from looking at the code

ProfFan: I am writing the document now, it's a bit complicated...

johnwlambert: cool, thanks!

ProfFan:
Robust Conversion from Matrix to Axis Angle Form

# 1 Introduction

Robust conversion from matrix to axis-angle form is not trivial because the 
transform is plagued with numeric instabilities. In this document we will 
review the normal form of such a conversion, and provide detailed instructions 
on how to properly do it.

## 1.1 Basic Conversion

It is known that the conversion from a   matrix can be 
written as

Theorem 1: The rotation angle $\theta$ can be calculated as

<span id="eq-orig-angle" class="tm-eqlabel"></span>
 

<div align="center"></div>

*where   is the trace of the rotation matrix.*

As such, the normalized rotation axis is:

Theorem 2: The normalized rotation axis for a rotation matrix $R$ is

<span id="eq-orig-axis" class="tm-eqlabel"></span>
 

<div align="center"></div>

*where   are the elements in the matrix.*

# 2 Problem

sin(x)02\*piCannot process embedded image

Figure 1.$\sin (x)$

It's pretty straightforward that when  ,  
[(undefined label: 'eq-orig-axis')](#eq-orig-axis) will not work because $\sin 
(\theta)$ will be 0. From [(undefined label: 'eq-orig-angle')](#eq-orig-angle) 
we know that, since  ,

 

<div align="center"></div>

Thus we know that there are two singular points, $\- 1$ and $3$, for the 
axis-angle conversion. Next thing is how to deal with these two points.

# 3 Properties of the $\operatorname{SO} (3)$ transform at the singular points

## 3.1  

For   the solution is simple: We simply 
take the Taylor expansion of

\\[ \frac{\theta}{2 \sin (\theta)} \\]

at  . Note this is not a simple Taylor expansion, but a 
limit. We can do this in Maxima:

maximadefault

Maxima 5.45.1 https://maxima.sourceforge.io

using Lisp SBCL 2.1.9

Distributed under the GNU Public License. See the file COPYING.

Dedicated to the memory of William Schelter.

The function bug\_report() provides bug reporting information.

(%i1)θ:acos(t-12) 

(%i2)mag:θ2\*sin(θ) 

(%i3)taylor(mag,t,3,3)

  

Thus the solution around   is

 

<div align="center"></div>

## 3.2 $\theta \rightarrow \pi$

The solution becomes a lot more complicated when at  . 
This is because at this extreme point you have two solutions: you can go from 
both left and right of the unit circle!

Note 3: The trace   is negative.

In this case we can first convert $R$ to quaternion, which is done by

 

<div align="center"></div>

where  ,  ,  .

Normally we want the angle  , thus we need to make sure 
that  ,

and now we can easily get the angle $\theta$ as

 

<div align="center"></div>

and now the magnitude of the vector is

 

<div align="center"></div>

However, in GTSAM, we do the following simplification to avoid doing the atan2 
(expensive).

First note that  . Thus we have 
(around $w = 0$) this Taylor expansion:

maximadefault

Maxima 5.45.1 https://maxima.sourceforge.io

using Lisp SBCL 2.1.9

Distributed under the GNU Public License. See the file COPYING.

Dedicated to the memory of William Schelter.

The function bug\_report() provides bug reporting information.

(%i1)θ:2\*atan2(1-w,w) 

(%i2)scale:θ/sin(θ/2) 

(%i3)taylor(scale,w,0,3)

  

which indicates that we only need a correction term of   to get the 
true magnitude near $\pi$.

PR #888, opened 2021-10-08
Run tests for C++ CI

varunagrawal:
I made an error when improving the performance of the CI due to which the tests were not being run. This PR fixes it.


varunagrawal:
Thanks for merging this Fan! :)

PR #892, opened 2021-10-13
Add unit tests where angle of (axis,angle) repr. exceeds 180 degrees by <0.5 deg

johnwlambert:
Adds specific examples to stress-test the log map.


johnwlambert:
@ProfFan thanks for the updates. just curious about a few things:
(1) why is atan2 expensive? i've always thought of that as a very cheap function
(2) the `maxima` software seems very cool that it can compute those taylor series symbolically. why are we confident that maxima is more accurate than wolfram's Taylor series?

gtsam/geometry/SO3.cpp
```diff
@@ -261,25 +261,46 @@ Vector3 SO3::Logmap(const SO3& Q, ChartJacobian H) {
...
-      const double sgn_w = (R13 - R31) < 0 ? -1.0 : 1.0;
-      const double r = sqrt(2.0 + 2.0 * R22);
-      const double scale = M_PI_2 / r - 0.5 / (r * r) * (R13 - R31);
-      omega = sgn_w * scale * Vector3(R21 + R12, 2.0 + 2.0 * R22, R23 + R32);
+      // R22 is the largest diagonal, a=2, b=3, c=1
+      const double W = R13 - R31;
+      const double Q1 = 2.0 + 2.0 * R22;
+      const double Q2 = R23 + R32;
+      const double Q3 = R12 + R21;
+      const double r = sqrt(Q1);
```
johnwlambert, line 284: this computation of `r` doesnt seem to match the math in https://github.com/borglab/gtsam/files/7333898/AxisAngleConversion.pdf right?

ProfFan: I have a new version

gtsam/geometry/SO3.cpp
```diff
@@ -261,25 +261,46 @@ Vector3 SO3::Logmap(const SO3& Q, ChartJacobian H) {
...
+      omega = sgn_w * scale * Vector3(Q2, Q3, Q1);
     } else if (R22 > R11) {
-      // R22 is the largest diagonal
-      const double sgn_w = (R13 - R31) < 0 ? -1.0 : 1.0;
-      const double r = sqrt(2.0 + 2.0 * R22);
-      const double scale = M_PI_2 / r - 0.5 / (r * r) * (R13 - R31);
-      omega = sgn_w * scale * Vector3(R21 + R12, 2.0 + 2.0 * R22, R23 + R32);
+      // R22 is the largest diagonal, a=2, b=3, c=1
+      const double W = R13 - R31;
+      const double Q1 = 2.0 + 2.0 * R22;
```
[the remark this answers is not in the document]
johnwlambert: @ProfFan thanks for that info. I think we can make this function about 15 lines shorter if we avoid any copy-paste. How does this look? 

```c++
double compute_signed_angle_from_quaternion(const double qx, const double qy, const double qz, const double qw)
{
	const double r = sqrt(qx);
	const double one_over_r = 1 / r;
	const double norm = sqrt(qx*qx + qy*qy + qz*qz + qw*qw);
	const double sgn_w = W < 0 ? -1.0 : 1.0;
	const double mag = M_PI - (2 * sgn_w * W) / norm;
	const double scale = 0.5 * one_over_r * mag;
	signed_scale = sign_w * scale;
	return signed_scale;
}

if (R33 > R22 && R33 > R11) {
	// R33 is the largest diagonal, a=3, b=1, c=2
	const double qw = R21 - R12;
	const double qx = 2.0 + 2.0 * R33;
	const double qy = R31 + R13;
	const double qz = R23 + R32;
	const double signed_scale = compute_signed_angle_from_quaternion(qx, qy, qx, qw);
	omega = signed_scale * Vector3(qy, qz, qx);
} else if (R22 > R11) {
	// R22 is the largest diagonal, a=2, b=3, c=1
	const double qw = R13 - R31;
	const double qx = 2.0 + 2.0 * R22;
	const double qy = R23 + R32;
	const double qz = R12 + R21;
	const double signed_scale = compute_signed_angle_from_quaternion(qx, qy, qx, qw);
	omega = signed_scale * Vector3(qz, qx, qy);
} else {
	// R11 is the largest diagonal, a=1, b=2, c=3
	const double qw = R32 - R23;
	const double qx = 2.0 + 2.0 * R11;
	const double qy = R12 + R21;
	const double qz = R31 + R13;
	const double signed_scale = compute_signed_angle_from_quaternion(qx, qy, qx, qw);
	omega = signed_scale * Vector3(qx, qy, qz);
}
```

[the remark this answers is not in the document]
ProfFan: That adds one more function call. Maybe you can use `inline` for that function though. Also have to make sure that it's still correct. Otherwise looks good to me.

gtsam/geometry/SO3.cpp
ProfFan: Check slack

PR #893, opened 2021-10-18
Clean up the __init__

varunagrawal:
Simple PR to do some wrapper clean up.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -65,10 +65,10 @@ set(interface_headers
     ${PROJECT_SOURCE_DIR}/gtsam/basis/basis.i
 )
 
-set(GTSAM_TARGET gtsam_py)
-set(GTSAM_UNSTABLE_TARGET gtsam_unstable_py)
+set(GTSAM_PYTHON_TARGET gtsam_py)
+set(GTSAM_PYTHON_UNSTABLE_TARGET gtsam_unstable_py)
 
-pybind_wrap(${GTSAM_TARGET} # target
+pybind_wrap(${GTSAM_PYTHON_TARGET} # target
             "${interface_headers}" # interface_headers
             "gtsam.cpp" # generated_cpp
             "gtsam" # module_name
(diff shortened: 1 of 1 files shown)


python/gtsam/__init__.py
```diff
@@ -1,6 +1,12 @@
...
-from .gtsam import *
-from .utils import findExampleDataFile
+"""Module definition file for GTSAM"""
+
+# pylint: disable=import-outside-toplevel, global-variable-not-assigned, possibly-unused-variable, import-error, import-self
+
+import sys
+
+from gtsam import gtsam, utils
+from gtsam.gtsam import *
```
johnwlambert, line 8: thanks for the PR @varunagrawal. While we're updating this, we probably want to remove this wildcard import too.

[Wildcard imports discussed in PEP8](https://www.python.org/dev/peps/pep-0008/)

python/CMakeLists.txt
```diff
@@ -65,8 +65,10 @@ set(interface_headers
     ${PROJECT_SOURCE_DIR}/gtsam/basis/basis.i
 )
 
+set(GTSAM_TARGET gtsam_py)
+set(GTSAM_UNSTABLE_TARGET gtsam_unstable_py)
 
-pybind_wrap(gtsam_py # target
+pybind_wrap(${GTSAM_TARGET} # target
```
ProfFan, line 71: GTSAM_PYTHON_TARGET

python/CMakeLists.txt
```diff
@@ -122,7 +124,7 @@ if(GTSAM_UNSTABLE_BUILD_PYTHON)
             gtsam::CameraSetCal3Fisheye
             gtsam::KeyPairDoubleMap)
             
-    pybind_wrap(gtsam_unstable_py # target
+    pybind_wrap(${GTSAM_UNSTABLE_TARGET} # target
```
ProfFan, line 127: same

python/gtsam/__init__.py
ProfFan: This is needed, and I think is generally OK for pybind11 modules.

PR #899, opened 2021-10-21
add LAGO (for Pose2) to python wrapper

johnwlambert:
Add's LAGO (for Pose2 slam without an initial estimate) to the Python wrapper, and ports the C++ example to Python.

### DIFF
--- a/python/gtsam/examples/Pose2SLAMExample_lago.py
+++ b/python/gtsam/examples/Pose2SLAMExample_lago.py
@@ -25,12 +25,7 @@ from argparse import Namespace
 import numpy as np
 
 import gtsam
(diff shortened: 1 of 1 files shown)


varunagrawal:
The LAGO functions are in their own namespace, so you can do this in the wrapper:

```cpp
#include <gtsam/slam/lago.h>
namespace lago {
  gtsam::Values initialize(const gtsam::NonlinearFactorGraph& graph, const gtsam::Values& initialGuess);
  gtsam::Values initialize(const gtsam::NonlinearFactorGraph& graph, bool useOdometricPath);
}
```

This will allow you to do `gtsam.lago.initialize()`.

PR #901, opened 2021-10-21
Revert @acxz's TBB revert

ProfFan:
I tested and didn't find any actual problem with the code by @acxz , so I think it's safe to revert.

Plus this allows building GTSAM with TBB 2021.x.

### DIFF
--- a/cmake/HandleTBB.cmake
+++ b/cmake/HandleTBB.cmake
@@ -7,9 +7,9 @@ if (GTSAM_WITH_TBB)
     if(TBB_FOUND)
         set(GTSAM_USE_TBB 1)  # This will go into config.h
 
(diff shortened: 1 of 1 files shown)


varunagrawal:
Update on this PR: It seems to be the cause of the recent ILS errors in GTDynamics. By turning off TBB, the errors go away.

@acxz can you take a look at this and help us understand why? You're probably the best at TBB amongst us.

acxz:
I'm afraid I don't really have the time to look at the GTDynamics issues right now.
I'll keep this in the back of my mind so that I can return to this later on in the month.

varunagrawal:
This isn't a GTDynamics issue, but the GTDynamics issue is a symptom of the issue with TBB :)

PR #913, opened 2021-11-02
enable expression double multiply

yetongumich:
Enable * operator for double expressions


dellaert:
@yetongumich feel free to merge after CI passes

PR #915, opened 2021-11-03
Fix warnings raised by GCC -Wpedactic

jlblancoc:
Basically extra `;` here and there. 
In the way, I updated some "old" typedef's by C++11's `using`.


dellaert:
Seems like CI fails?

jlblancoc:
> Seems like CI fails?

Yes, didn't test the unit tests... Now it should pass, let's wait for CI.

jlblancoc:
It now works. I "fixed" the issue that, for *one particular macro*, their instances *sometimes* had an ending semicolon and others they have not (giving raise to warnings if using `-Wpedantic`).
I peek at the sources and "fixed" it by: 
- If the most macro instances already were *with* final semicolon, add it to all remaining cases; 
- If it was the opposite, removing all semicolons; and
- fixing the definition of the macro such that no additional semicolon is generated, i.e. if instances of the macro already had the macro, we don't want a semicolon in the definition, or viceversa.

The only remaining issue is that there is not a 100% consistent convention over the entire codebase, since for some macros the most common case was *having* a semicolon and for others it was *not having* it... for example: 
- Needs a final semicolon: `GTSAM_CONCEPT_POSE_TYPE`, `GTSAM_MAKE_VECTOR_DEFS`, ...
- Does NOT need it: `GTSAM_CONCEPT_TESTABLE_TYPE`, ...

@dellaert , let me know if that's ok (and the PR can be merged), or if you prefer having a uniform convention (all macros with or w/o ";") which one is your preference...

dellaert (approved):
We *could* merge this now, but I agree with your sentiment (between the lines) that we should be consistent. If you have appetite to do so, I guess I prefer the semicolon in the macro, rather than in the test files. It's (a tiny bit) cleaner. I can go either way, though.

dellaert:
Cool. Merge at will after CI

jlblancoc:
> Cool. Merge at will after CI

I recently found, under project settings, an option to enable "automatic merging". Despite the name, it's just a "do the merge after CI passes". Pretty useful...
A new button shows up at PRs and, for your settings it will mean that merge will happen after review + CI passes, automatically.

PR #917, opened 2021-11-06
Use Eigen expressions more effectively and kill & in code.

dellaert:
Quick edit on @gchenfc 's recent PR #433 
Be careful with having references for local variables. That's for arguments, not things on the stack. Also, though counter-intuitive I think doing transpose() twice here is actually more efficient as it will not actually store the transpose internally, which you forced. Read about Eigen expressions and tell me if I'm wrong :-)


gchenfc (approved):
Yep, great on both accounts!  Optional commentary on why I had these mistakes:

1) `const XXX &var = func()` - I read on stackoverflow one time that this is the one special time it is legal to assign a const reference to a temporary or something along those lines, but after doing more reading now I think this was just factually wrong advice.  Because I thought it was legal, I've been using this syntax because I was too lazy to check (or to remember) whether the function returns a `const ref` (e.g. for a member variable getter) or value.  If the function returns a const ref, this will save a copy and if the function returns a value, I figured it doesn't make a difference so it's just easier to always assign the function return to const ref.  But now I know it's illegal and will go back and fix code where I did this.

2) Good catch!  I just missed this.

PR #918, opened 2021-11-06
Add Similarity(2) support in gtsam/geometry

johnwlambert:
We previously had Similarity(3).

This adds a basic interface for Similarity(2) which is quite useful for my current work. Mostly parallels the implementation in Similarity3.h / .cpp, except this is a bare bones interface without wedge, adjointMap, expmap, or logmap for now.


gtsam/geometry/Similarity2.cpp
```diff
@@ -0,0 +1,203 @@
...
+  for (const Point2Pair& abPair : abPointPairs) {
+    Point2 da = abPair.first - centroids.first;
+    Point2 db = abPair.second - centroids.second;
+    d_abPointPairs.emplace_back(da, db);
+  }
+  return d_abPointPairs;
+}
+
+/// Form inner products x and y and calculate scale.
+static const double calculateScale(const Point2Pairs &d_abPointPairs,
```
dellaert, line 42: No need for const in return trype

gtsam/geometry/Similarity2.cpp
```diff
@@ -0,0 +1,203 @@
...
+                                   const Rot2 &aRb) {
+  double x = 0, y = 0;
```
dellaert, line 44: Move da, db declarations into loop, make refs?

gtsam/geometry/Similarity2.cpp
```diff
@@ -0,0 +1,203 @@
...
+  std::cout << std::endl;
+  std::cout << s;
+  rotation().print("\nR:\n");
+  std::cout << "t: " << translation().transpose() << " s: " << scale() << std::endl;
+}
+
+Similarity2 Similarity2::identity() {
+  return Similarity2();
+}
+Similarity2 Similarity2::operator*(const Similarity2& S) const {
```
dellaert, line 125: Missing empty line

gtsam/geometry/Similarity2.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+#include <gtsam/geometry/Pose2.h>
+#include <gtsam/base/Manifold.h>
+#include <gtsam/slam/KarcherMeanFactor-inl.h>
+
+namespace gtsam {
+
+using std::vector;
+
+namespace {
```
dellaert, line 28: Maybe name this “internal” and qualify calls below, so we can more easily see it’s local

gtsam/geometry/Similarity2.cpp
```diff
@@ -0,0 +1,203 @@
...
+  return alignGivenR(abPointPairs, aRb_estimate);
+}
+
+std::ostream &operator<<(std::ostream &os, const Similarity2& p) {
+  os << "[" << p.rotation().theta() << " "
+      << p.translation().transpose() << " " << p.scale() << "]\';";
+  return os;
+}
+
+const Matrix3 Similarity2::matrix() const {
```
dellaert, line 192: No need for const in return type

gtsam/geometry/Similarity2.cpp
```diff
@@ -0,0 +1,203 @@
...
+  Matrix3 T;
+  T.topRows<2>() << R_.matrix(), t_;
+  T.bottomRows<1>() << 0, 0, 1.0 / s_;
+  return T;
+}
+
+Similarity2::operator Pose2() const {
```
dellaert, line 199: That’s a weird cast

gtsam/geometry/Similarity2.h
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /// Return the scale
+  double scale() const {
+    return s_;
+  }
+
+  /// Convert to a rigid body pose (R, s*t)
+  /// TODO(frank): why is this here? Red flag! Definitely don't have it as a cast.
+  GTSAM_EXPORT operator Pose2() const;
```
dellaert, line 167: Interesting that I have comments in this new PR ?? Must have forgotten. In any case, comment holds: remove this cast operator, as I don’t think there is a valid way to cast down, only cast from Pose2 to Sim2.

gtsam/geometry/geometry.i
```diff
@@ -890,6 +890,28 @@ class PinholeCamera {
...
+  Similarity2(const Matrix& T);
+
+  gtsam::Point2 transformFrom(const gtsam::Point2& p) const;
+  gtsam::Pose2 transformFrom(const gtsam::Pose2& T);
+
+  static gtsam::Similarity2 Align(const gtsam::Point2Pairs& abPointPairs);
+  static gtsam::Similarity2 Align(const gtsam::Pose2Pairs& abPosePairs);
+
+  // Standard Interface
+  const Matrix matrix() const;
```
dellaert, line 910: No const in return

dellaert (changes requested):
Cool, but CI does not pass… email me when a new review needed as might not catch a GitHub note

gtsam/geometry/Similarity2.cpp
```diff
@@ -0,0 +1,203 @@
...
+Similarity2::Similarity2() :
+    t_(0,0), s_(1) {
+}
+
+Similarity2::Similarity2(double s) :
+    t_(0,0), s_(s) {
+}
+
+Similarity2::Similarity2(const Rot2& R, const Point2& t, double s) :
+    R_(R), t_(t), s_(s) {
```
johnwlambert, line 95: thanks for the review, @dellaert. Currently we don't have a 2x2 constructor for Rot2, nor a "closestTo" equivalent. We could make a `.closestTo()` using something like the C++ analogue of:

```python
    theta_rad = np.arctan2(R[1, 0], R[0, 0])
    theta_deg = np.rad2deg(theta_rad)
    R_ = rotmat2d(theta_deg)
```

Any particular reason not to have a 2x2 constructor on Rot2, other than the fact that a user could provide an invalid set of numbers (not orthogonal)? They could do that for the 3x3 case for Rot3 as well, though.

dellaert: Rot2 only stores 2 numbers

dellaert:
@johnwlambert take a look at this again? CI fails, too.

gtsam/geometry/Similarity2.cpp
johnwlambert: @dellaert would you be ok with a 2x2 constructor that just extracts those 2 numbers, and then calls the 2-number constructor?

johnwlambert: Otherwise I think that 2-number extraction code would have to be duplicated in a few different places here

dellaert:
@johnwlambert I think we should start thinking about merging this PR? I think the only thing missing is the closestto?  I don’t think it’s appropriate to do that with a constructor, as I mentioned.

gtsam/geometry/Similarity2.h
varunagrawal: This is actually copy-pasted from `Similarity3.h`. We should probably address that there.

gtsam/geometry/Similarity2.cpp
johnwlambert: Fixed by @varunagrawal . Marking as resolved.

gtsam/geometry/Similarity2.cpp
johnwlambert: Fixed by @varunagrawal.

gtsam/geometry/Similarity2.cpp
johnwlambert: Resolved -- by use of a `Rot2.ClosestTo()`, instead of with a 2x2 matrix arg constructor.

gtsam/geometry/Similarity2.cpp
johnwlambert: Didn't quite follow the comment here.

gtsam/geometry/Similarity2.cpp
johnwlambert: Casting `Sim2` to `Pose2` is a bit of an unusual cast, but quite useful for some of my use cases (sometimes ground truth is provided in this form for some datasets).

varunagrawal:
I'm going to leave reviewing to @dellaert (or maybe someone else from gtsfm) since I am a contributor to this PR and hence biased. :)

gtsam/geometry/Rot2.cpp
```diff
@@ -129,6 +129,16 @@ Rot2 Rot2::relativeBearing(const Point2& d, OptionalJacobian<1, 2> H) {
   }
 }
 
+/* ************************************************************************* */
+Rot2 Rot2::ClosestTo(const Matrix2& M) {
```
dellaert, line 133: I applaud the existence of this method!
But not sure this is right. I think it should be done with SVD as in SO3, then c and s can just be read off as below.

gtsam/geometry/Point2.cpp
```diff
@@ -113,6 +113,18 @@ list<Point2> circleCircleIntersection(Point2 c1, double r1, Point2 c2,
   return circleCircleIntersection(c1, c2, fh);
 }
 
+Point2Pair means(const std::vector<Point2Pair> &abPointPairs) {
```
dellaert, line 116: For this PR and my latest PR, I'm starting to think we should "light-deprecate" PointPairs and do the computation in Eigen.
https://eigen.tuxfamily.org/dox/group__TutorialReductionsVisitorsBroadcasting.html

gtsam/geometry/Pose2.h
```diff
@@ -322,6 +322,10 @@ inline Matrix wedge<Pose2>(const Vector& xi) {
 typedef std::pair<Point2,Point2> Point2Pair;
 GTSAM_EXPORT boost::optional<Pose2> align(const std::vector<Point2Pair>& pairs);
 
+// Convenience typedef
+using Pose2Pair = std::pair<Pose2, Pose2>;
```
dellaert, line 326: these I think already exist in Point2, which is where they should be defined I think

gtsam/geometry/Similarity2.cpp
dellaert:
```
{
  Point2 &da, &db;
  std::tie(da, db) = d_abPair;
...
}
```

NOT for this PR but I now think we should have a Matrix version of Rot2, Pose2, and and do
```
DA = A.colwise() - A.mean();  // 2xN
DB = B.colwise() - B.mean();  // 2xN
aRb * DB * DA.transpose();  //  2x2 * 2xN * Nx2 = 2x2
// I think x and y should be found in the matrix above?
```

gtsam/geometry/Similarity2.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * given a known or estimated rotation.
+ * Refer to:
+ * http://www5.informatik.uni-erlangen.de/Forschung/Publikationen/2005/Zinsser05-PSR.pdf
+ * Chapter 3
+ *
+ * @param abPointPairs
+ * @param aRb
+ * @return Similarity2
+ */
+static Similarity2 alignGivenR(const Point2Pairs& abPointPairs,
```
dellaert, line 96: static methods are capitalized in GTSAM.

gtsam/geometry/Similarity2.h
```diff
@@ -0,0 +1,179 @@
...
+
+  GTSAM_EXPORT friend std::ostream& operator<<(std::ostream& os,
+                                               const Similarity2& p);
+
+  /// @}
+  /// @name Group
+  /// @{
+
+  /// Return an identity transform
+  GTSAM_EXPORT static Similarity2 identity();
```
dellaert, line 87: capitalize. Unless the "CONCEPT" wants lowercase (which would be a mistake to correct)

gtsam/geometry/Similarity2.h
```diff
@@ -0,0 +1,179 @@
...
+  Rot2 R_;
+  Point2 t_;
+  double s_;
+
+ public:
+  /// @name Constructors
+  /// @{
+
+  /// Default constructor
+  GTSAM_EXPORT Similarity2();
```
dellaert, line 52: GTSAM_EXPORT should be added to the class, not all methods. Fix in Similarity3 if was copied from there. Goes for all comments

gtsam/geometry/Similarity3.h
```diff
@@ -180,15 +180,15 @@ class Similarity3: public LieGroup<Similarity3, 7> {
   /// @{
 
   /// Calculate 4*4 matrix group equivalent
-  GTSAM_EXPORT const Matrix4 matrix() const;
+  GTSAM_EXPORT Matrix4 matrix() const;
```
dellaert, line 183: GTSAM_EXPORT on class, not methods

gtsam/navigation/tests/testGPSFactor.cpp
```diff
@@ -71,7 +70,7 @@ TEST( GPSFactor, Constructor ) {
   EXPECT(assert_equal(Z_3x1,factor.evaluateError(T),1e-5));
 
   // Calculate numerical derivatives
-  Matrix expectedH = numericalDerivative11<Vector,Pose3>(
+  Matrix expectedH = numericalDerivative11<Vector, Pose3>(
```
dellaert, line 73: Why changes in unrelated files? Maybe merge in develop again? Or move these to new PR

python/gtsam/tests/test_Sim2.py
```diff
@@ -0,0 +1,196 @@
...
+        expected_t = np.array([1, 2])
+        assert np.allclose(expected_t, bSa.translation())
+
+    def test_scale(self) -> None:
+        """Ensure the scale factor is returned properly."""
+        bRa = Rot2()
+        bta = np.array([1, 2])
+        bsa = 3.0
+        bSa = Similarity2(R=bRa, t=bta, s=bsa)
+        assert bSa.scale() == 3.0
```
dellaert, line 192: self.assertEqual

dellaert (commented):
Sorry about the delay, let's try to land this. Note I am thinking we should prefer Matrix versions (and do computation there) for *all* `Align` constructors and add Rot2 and Rot3 versions, but that's for a different day.

Not yet approving as want to review after comments have been addressed (or rebuffed).

gtsam/geometry/Pose2.h
varunagrawal: That's `Point2Pair` so this declaration here is appropriate. :)

gtsam/geometry/Similarity2.cpp
varunagrawal: Even I don't understand the purpose of this. It's doable but why?

varunagrawal: And because it's a reference, initialization of the variables is required, which means the above 2 lines will change to the single line

```cpp
const Point2 &da = d_abPair.first, &db = d_abPair.second;
```

python/gtsam/tests/test_Sim2.py
varunagrawal: Updated this for all assertions.

gtsam/geometry/Similarity3.h
varunagrawal: Done for both Similarity2 and Similarity3

gtsam/navigation/tests/testGPSFactor.cpp
varunagrawal: Merged in latest `develop`.

gtsam/geometry/Point2.cpp
varunagrawal: Should we save this for another PR in favor of landing this one?
We can create an issue to track it.

dellaert (approved):
Cool, ready I think!

varunagrawal:
I couldn't find any references for AdjointMap of Sim(2) and a C++ based unit test file is missing. :(

I'll push a test file with a single constructor test and then merge.

gtsam/geometry/Similarity2.h
varunagrawal: The concept indeed requires lowercase. I'll add a TODO.

Issue #889, opened 2021-10-10
[Deprecation] Properly mark deprecated functions in C++


ProfFan:
We should mark deprecated functions both in Doxygen and **in C++**.

For example,

In `EliminateableFactorGraph-inst.h` Line 39 it's calling deprecated function.

dellaert:
I think this was incorrectly closed

varunagrawal:
What is needed to close this? PR #925 added the macros for conditonal compilation.

PR #920, opened 2021-11-09
Fix minor stuff

varunagrawal:
A bunch of minor fixes I made along the way.

### DIFF
--- a/gtsam/inference/EliminateableFactorGraph-inst.h
+++ b/gtsam/inference/EliminateableFactorGraph-inst.h
@@ -36,17 +36,17 @@ namespace gtsam {
       // no Ordering is provided.  When removing optional from VariableIndex, create VariableIndex
       // before creating ordering.
       VariableIndex computedVariableIndex(asDerived());
-      return eliminateSequential(function, computedVariableIndex, orderingType);
+      return eliminateSequential(orderingType, function, computedVariableIndex);
     }
     else {
       // Compute an ordering and call this function again.  We are guaranteed to have a 
       // VariableIndex already here because we computed one if needed in the previous 'if' block.
       if (orderingType == Ordering::METIS) {
         Ordering computedOrdering = Ordering::Metis(asDerived());
-        return eliminateSequential(computedOrdering, function, variableIndex, orderingType);
+        return eliminateSequential(computedOrdering, function, variableIndex);
       } else {
         Ordering computedOrdering = Ordering::Colamd(*variableIndex);
-        return eliminateSequential(computedOrdering, function, variableIndex, orderingType);
+        return eliminateSequential(computedOrdering, function, variableIndex);
       }
     }
   }
--- a/gtsam/nonlinear/factorTesting.h
+++ b/gtsam/nonlinear/factorTesting.h
@@ -35,8 +35,7 @@ namespace gtsam {
  * zero errors anyway. However, it means that below will only be exact for the correct measurement.
  */
 JacobianFactor linearizeNumerically(const NoiseModelFactor& factor,
-    const Values& values, double delta = 1e-5) {
-
+                                    const Values& values, double delta = 1e-5) {
   // We will fill a vector of key/Jacobians pairs (a map would sort)
   std::vector<std::pair<Key, Matrix> > jacobians;
 
@@ -46,24 +45,24 @@ JacobianFactor linearizeNumerically(const NoiseModelFactor& factor,
 
   // Loop over all variables
   const double one_over_2delta = 1.0 / (2.0 * delta);
-  for(Key key: factor) {
+  for (Key key : factor) {
     // Compute central differences using the values struct.
     VectorValues dX = values.zeroVectors();
     const size_t cols = dX.dim(key);
     Matrix J = Matrix::Zero(rows, cols);
     for (size_t col = 0; col < cols; ++col) {
       Eigen::VectorXd dx = Eigen::VectorXd::Zero(cols);
-      dx[col] = delta;
+      dx(col) = delta;
       dX[key] = dx;
       Values eval_values = values.retract(dX);
       const Eigen::VectorXd left = factor.whitenedError(eval_values);
-      dx[col] = -delta;
+      dx(col) = -delta;
       dX[key] = dx;
       eval_values = values.retract(dX);
       const Eigen::VectorXd right = factor.whitenedError(eval_values);
       J.col(col) = (left - right) * one_over_2delta;
     }
-    jacobians.push_back(std::make_pair(key,J));
+    jacobians.push_back(std::make_pair(key, J));
   }
 
   // Next step...return JacobianFactor
--- a/python/gtsam/examples/SFMExample_bal.py
+++ b/python/gtsam/examples/SFMExample_bal.py
@@ -16,36 +16,37 @@ import sys
 
 import gtsam
 from gtsam import (GeneralSFMFactorCal3Bundler,
-                   PriorFactorPinholeCameraCal3Bundler, PriorFactorPoint3,
-                   readBal)
+                   PriorFactorPinholeCameraCal3Bundler, PriorFactorPoint3)
 from gtsam.symbol_shorthand import C, P
-from gtsam.utils.plot import plot_3d_points, plot_trajectory
+from gtsam.utils import plot
+from matplotlib import pyplot as plt
 
 logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
+DEFAULT_BAL_DATASET = "dubrovnik-3-7-pre"
 
-def plot(scene_data: gtsam.SfmData, result: gtsam.Values):
-    """Plot the trajectory."""
+
+def plot_scene(scene_data: gtsam.SfmData, result: gtsam.Values):
+    """Plot the SFM results."""
     plot_vals = gtsam.Values()
     for cam_idx in range(scene_data.number_cameras()):
         plot_vals.insert(C(cam_idx),
                          result.atPinholeCameraCal3Bundler(C(cam_idx)).pose())
-    for t_idx in range(scene_data.number_tracks()):
-        plot_vals.insert(P(t_idx), result.atPoint3(P(t_idx)))
+    for j in range(scene_data.number_tracks()):
+        plot_vals.insert(P(j), result.atPoint3(P(j)))
 
-    plot_3d_points(0, plot_vals, linespec="g.")
-    plot_trajectory(0, plot_vals, show=True)
+    plot.plot_3d_points(0, plot_vals, linespec="g.")
+    plot.plot_trajectory(0, plot_vals, title="SFM results")
+
+    plt.show()
 
 
 def run(args: argparse.Namespace):
     """ Run LM optimization with BAL input data and report resulting error """
-    if args.input_file:
-        input_file = args.input_file
-    else:
-        input_file = gtsam.findExampleDataFile("dubrovnik-3-7-pre")
+    input_file = args.input_file
 
     # Load the SfM data from file
-    scene_data = readBal(input_file)
+    scene_data = gtsam.readBal(input_file)
     logging.info("read %d tracks on %d cameras\n", scene_data.number_tracks(),
                  scene_data.number_cameras())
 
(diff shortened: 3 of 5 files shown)


gtsam/inference/EliminateableFactorGraph-inst.h
```diff
@@ -36,17 +36,17 @@ namespace gtsam {
       // no Ordering is provided.  When removing optional from VariableIndex, create VariableIndex
       // before creating ordering.
       VariableIndex computedVariableIndex(asDerived());
-      return eliminateSequential(function, computedVariableIndex, orderingType);
+      return eliminateSequential(orderingType, function, computedVariableIndex);
```
johnwlambert, line 39: thanks for the PR, Varun.

Why change the order of variables here? breaks backwards compatibility perhaps?

varunagrawal: It doesn't because this is a function call within another function aka the public API is still the same.

Fan and I discovered that the original function call was just calling this function, so instead of the indirect call, we save on the call stack by making this call directly.

varunagrawal: Another way to think of this is that we're just calling another version of an overloaded function, but with the added benefit of reducing the call stack, thus minor gains in efficiency plus easier to read the stack trace during debugging.

johnwlambert (changes requested):

gtsam/inference/EliminateableFactorGraph-inst.h
johnwlambert: Cool, makes sense. Could you add a link to the old wrapper function and the new function we're calling directly? Which file are they in?

python/gtsam/examples/SFMExample_bal.py
```diff
@@ -7,49 +7,58 @@
...
-logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
+    plot_3d_points(0, plot_vals, linespec="g.")
+    plot_trajectory(0, plot_vals, show=True)
 
-def run(args):
+
+def run(args: argparse.Namespace):
     """ Run LM optimization with BAL input data and report resulting error """
-    input_file = gtsam.findExampleDataFile(args.input_file)
+    if args.input_file:
```
akshay-krishnan, line 42: I think it might be better for this function to check that args.input_file exists and can be read. If not, it can fail. 
The default is probably better provided as a flag default. Are there other reasons why you prefer it this way?

akshay-krishnan (commented):
All comments I have are a subset of John's, so feel free to merge once he approves.

python/gtsam/examples/SFMExample_bal.py
```diff
@@ -60,59 +69,61 @@ def run(args):
...
     # Error drops from ~2764.22 to ~0.046
-    logging.info(f"final error: {graph.error(result)}")
+    logging.info("initial error: %f", graph.error(initial))
+    logging.info("final error: %f", graph.error(result))
 
+    plot(scene_data, result)
 
-if __name__ == "__main__":
+
+def main():
```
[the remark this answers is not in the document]
varunagrawal: Adding a return type declaration for None is questionable, since is it even useful? If the return type is something non-trivial, then it helps with things like linting and autocomplete, but None doesn't help with any of that.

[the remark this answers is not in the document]
johnwlambert: I think it's useful purely for complete static analysis

python/gtsam/examples/SFMExample_bal.py
varunagrawal: `readBal` would fail if the file didn't exist so I am not sure what the ask is here.

varunagrawal: I updated the default flag in a way that should make you happy. :slightly_smiling_face:

gtsam/inference/EliminateableFactorGraph-inst.h
varunagrawal: This is the original one which is deprecated (though not properly).
https://github.com/borglab/gtsam/blob/744db328e7ae537e71329e04cc141b3a28b0d6bd/gtsam/inference/EliminateableFactorGraph.h#L301

This is replaced by this https://github.com/borglab/gtsam/blob/744db328e7ae537e71329e04cc141b3a28b0d6bd/gtsam/inference/EliminateableFactorGraph.h#L117

python/gtsam/examples/SFMExample_bal.py
```diff
@@ -7,112 +7,122 @@
...
+from gtsam.symbol_shorthand import C, P
+from gtsam.utils import plot
+from matplotlib import pyplot as plt
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+DEFAULT_BAL_DATASET = "dubrovnik-3-7-pre"
+
+
+def plot_scene(scene_data: gtsam.SfmData, result: gtsam.Values):
```
johnwlambert, line 29: looks good. nit here:  missing return type

python/gtsam/examples/SFMExample_bal.py
```diff
@@ -7,112 +7,122 @@
...
-    GeneralSFMFactorCal3Bundler,
-    PinholeCameraCal3Bundler,
-    PriorFactorPinholeCameraCal3Bundler,
-    readBal,
-    symbol_shorthand
-)
+from gtsam import (GeneralSFMFactorCal3Bundler,
+                   PriorFactorPinholeCameraCal3Bundler, PriorFactorPoint3)
+from gtsam.symbol_shorthand import C, P
+from gtsam.utils import plot
```
johnwlambert, line 21: I would import this as something like
```python
import gtsam.utils.plot gtsam_plot_utils
```

or 
```python
import gtsam.utils.plot gtsam_plot
```

Otherwise I get confused about whether this is matplotlib

johnwlambert (changes requested):

python/gtsam/examples/SFMExample_bal.py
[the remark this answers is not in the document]
varunagrawal: But does the static analysis report it missing? I have all warning enabled (even for the linter) and I get no warnings.

gtsam/nonlinear/factorTesting.h
```diff
@@ -46,24 +45,24 @@ JacobianFactor linearizeNumerically(const NoiseModelFactor& factor,
...
     Matrix J = Matrix::Zero(rows, cols);
     for (size_t col = 0; col < cols; ++col) {
       Eigen::VectorXd dx = Eigen::VectorXd::Zero(cols);
-      dx[col] = delta;
+      dx(col) = delta;
       dX[key] = dx;
       Values eval_values = values.retract(dX);
       const Eigen::VectorXd left = factor.whitenedError(eval_values);
-      dx[col] = -delta;
+      dx(col) = -delta;
```
johnwlambert, line 59: what's the preference here between square bracket and paren?

python/gtsam/examples/SFMExample_bal.py
varunagrawal: Which is why I had the functions imported directly. :slightly_smiling_face: 
If we are okay with that, I can revert to the function import, since this is the case everywhere in the examples.

python/gtsam/examples/SFMExample_bal.py
varunagrawal: Sorry, I don't see the utility of adding a `None` return type specification. Unless there is a compelling reason for this, we are losing the forest for the trees.

python/gtsam/examples/SFMExample_bal.py
johnwlambert: I think if it's wrong in the other examples, we can slowly make the correct code go viral, starting from here : - )

Importing functions should not be allowed. But if you prefer calling it `plot`, feel free to leave this change as you have it here

gtsam/nonlinear/factorTesting.h
varunagrawal: Parens allow for safer access in Eigen matrices and is the preferred syntax.

Checkout the section `Coefficient accessors` at [this link](https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html).

python/gtsam/examples/SFMExample_bal.py
johnwlambert: For me, it makes it easier to immediately understand the function signature.

If we don't have it, the type annotations are not complete, and we can't rely on mypy or other static analyzers completely.

gtsam/nonlinear/factorTesting.h
johnwlambert: thanks!

python/gtsam/examples/SFMExample_bal.py
varunagrawal: What's the reason for that? Google's style guide mentions `Use import statements for packages and modules only, not for individual classes or functions`, so we're clearly not abiding by this. :slightly_smiling_face: 

Plus, I am of the opinion that these are _guidelines_ and not the rules, so we can not conform to them if it means the code is more readable.

python/gtsam/examples/SFMExample_bal.py
varunagrawal: If it's missing, we should be able to assume default None if the docstring doesn't mention a return type. No return type == None makes sense.

python/gtsam/examples/SFMExample_bal.py
varunagrawal: Also, ironically, calling it `gtsam_plot_utils` is no better than simply doing `gtsam.utils.plot.whatever_func`.

python/gtsam/examples/SFMExample_bal.py
johnwlambert: We can infer that, but mypy can't : - )

python/gtsam/examples/SFMExample_bal.py
johnwlambert: I think many modern Python libraries like Pytorch abide by this. Otherwise it's hard to immediately tell where to go to look for the implementation of a function (do I ctrl-F in this file, or look in another file)

 Interesting that Google style forbids the class import, though.

gtsam/inference/EliminateableFactorGraph-inst.h
johnwlambert: I see. So now we are using overloaded versions with boost "none" defaults for missing args. I wonder if we should make a TODO or note to ** fully ** remove the deprecated one as well

johnwlambert (approved):
LGTM except I prefer the additional type hints. Rest looks good.

python/gtsam/examples/SFMExample_bal.py
varunagrawal: Mypy doesn't report None violations regardless. I added `-> None` to the run function and did `x = run(parser.parse_args())`, then I ran mypy and mypy didn't report any issue here, which is the correct behavior.

This goes back to my original comment, this would be useful for autocomplete. However, we're missing the main point that the wrapper doesn't provide type definitions out of the box, so mypy will fail on a lot of other things related to GTSAM before a missing return type None. Plus, we don't even check for type definitions in the CI, so again, I don't see a compelling reason for this review request.

python/gtsam/examples/SFMExample_bal.py
varunagrawal: Pytorch does something way wackier. Plus they have a bunch of people making 200K a year for this stuff, no 22K a year like yours truly. :slightly_smiling_face: 

If you paid me that much money, I'd be more than happy to accommodate this request. :joy:

gtsam/inference/EliminateableFactorGraph-inst.h
varunagrawal: Yup, that's issue #889

varunagrawal: We don't do TODOs here, we make issues haha

python/gtsam/examples/SFMExample_bal.py
varunagrawal: I added in the type-hints.

PR #921, opened 2021-11-10
CMake fix for Python Tests

varunagrawal:
Update cmake to copy python tests whenever they are updated.


gchenfc (approved):
Great, thanks!  Tested working on my machine :)

PR #922, opened 2021-11-11
Add Jacobian of second argument to `adjoint` and `adjointTranpsose`

gchenfc:
The `adjoint` function is equivalent to
```c++
adjointMap(x) * y
```

The current function can output the jacobian w.r.t. x but not w.r.t. y, for some reason.  This PR adds the jacobian w.r.t. y.

The jacobian w.r.t. y is pretty obviously `adjointMap(x)`, but I think it's convenient and more consistent with other function calls to have the option to include both jacobians.

This also updates the python wrapper, since the new wrapper does not have any issues with the non-underscore versions of the functions.


dellaert (approved):
Cool!

PR #925, opened 2021-11-11
Proper Deprecation of Eliminate Methods

### LINKED ISSUE #889: [Deprecation] Properly mark deprecated functions in C++

varunagrawal:
Found this while looking at elimination in GaussianFactorGraphs for MHS with @ProfFan. Let's get this out of the way so we don't have accidentally cause a dependency on this.

Fixes #889

### DIFF
--- a/gtsam/inference/EliminateableFactorGraph-inst.h
+++ b/gtsam/inference/EliminateableFactorGraph-inst.h
@@ -78,29 +78,31 @@ namespace gtsam {
   }
 
   /* ************************************************************************* */
-  template<class FACTORGRAPH>
-  boost::shared_ptr<typename EliminateableFactorGraph<FACTORGRAPH>::BayesTreeType>
-    EliminateableFactorGraph<FACTORGRAPH>::eliminateMultifrontal(
-    OptionalOrderingType orderingType, const Eliminate& function,
-    OptionalVariableIndex variableIndex) const
-  {
-    if(!variableIndex) {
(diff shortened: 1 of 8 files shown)


dellaert (approved):

ProfFan:
@varunagrawal  Could you also add the deprecation pragmas?

```
#if defined(__GNUC__) || defined(__clang__)
#define GTSAM_DEPRECATED __attribute__((deprecated))
#elif defined(_MSC_VER)
#define GTSAM_DEPRECATED __declspec(deprecated)
#else
#define GTSAM_DEPRECATED
#endif
```

dellaert:
> @varunagrawal Could you also add the deprecation pragmas?
> 
> ```
> #if defined(__GNUC__) || defined(__clang__)
> #define GTSAM_DEPRECATED __attribute__((deprecated))
> #elif defined(_MSC_VER)
> #define GTSAM_DEPRECATED __declspec(deprecated)
> #else
> #define GTSAM_DEPRECATED
> #endif
> ```

I would think that is a different PR. That would be cool though

varunagrawal:
I had to look up what that was but yeah it would be awesome to have that and mark the corresponding methods with it. Next PR :slightly_smiling_face:

PR #931, opened 2021-11-18
Re-factored arc consistency

dellaert:
This is revamping some of the code in gtsam_unstable on solving CSPs. The reason for doing this now is to see whether we can efficiently implement STRIPS-style planning with GTSAM, which I think we can.

In particular, I modernized runArcConsistency, which is an implementation of AC1 constrain propagation. This already worked quite well, but I made it so that it can work with arbitrary keys, not hardcoded Sudoku keys (0...80). Mostly this involved switching from a `vector<Domain>` data structure to a `std::map`.

After AC1, you can theoretically create a smaller factor graph, and I made that actually work. That's the `partiallyApply` function. It works, but the factor graphs are still too complex to solve large sudokus.

It does work for the last Sudoku example (comment out line 254):
```
(base) M1:build dellaert$ make testSudoku.run
[100%] Built target testSudoku
9 5 7 3 4 6 8 1 2 
1 8 4 9 7 2 3 5 6 
6 2 3 5 8 1 4 7 9 
8 7 1 2 9 5 6 4 3 
3 9 6 8 1 4 7 2 5 
5 4 2 7 6 3 9 8 1 
2 3 9 4 5 7 1 6 8 
7 1 5 6 3 8 2 9 4 
4 6 8 1 2 9 5 3 7 
There were no test failures
[100%] Built target testSudoku.run

```
I also:
- Made in/out parameter a pointer, consistent with Google style, and made checkAllDiff function functional (const!). 
- commented tests better and teased them out into smaller tests.

### DIFF
--- a/gtsam/discrete/DiscreteFactorGraph.h
+++ b/gtsam/discrete/DiscreteFactorGraph.h
@@ -101,25 +101,27 @@ public:
 
   /// @}
 
-  template<class SOURCE>
+  // Add single key decision-tree factor.
+  template <class SOURCE>
   void add(const DiscreteKey& j, SOURCE table) {
     DiscreteKeys keys;
     keys.push_back(j);
-    push_back(boost::make_shared<DecisionTreeFactor>(keys, table));
+    emplace_shared<DecisionTreeFactor>(keys, table);
   }
 
-  template<class SOURCE>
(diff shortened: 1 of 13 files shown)


xsj01 (approved):
Looks good to me

PR #932, opened 2021-11-19
Wrap KeyVector methods

varunagrawal:
The title. This is needed for MHS.

We're also going to need to publish the latest version of the python package so that the CI can run successfully.

### DIFF
--- a/gtsam/linear/linear.i
+++ b/gtsam/linear/linear.i
@@ -302,6 +302,7 @@ virtual class JacobianFactor : gtsam::GaussianFactor {
   void print(string s = "", const gtsam::KeyFormatter& keyFormatter =
                                 gtsam::DefaultKeyFormatter) const;
   void printKeys(string s) const;
+  gtsam::KeyVector& keys() const;
   bool equals(const gtsam::GaussianFactor& lf, double tol) const;
   size_t size() const;
   Vector unweighted_error(const gtsam::VectorValues& c) const;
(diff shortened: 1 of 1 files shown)


dellaert:
> The title. This is needed for MHS.
> 
> We're also going to need to publish the latest version of the python package so that the CI can run successfully.

Explain comment about CI, it should not depend on a python package

varunagrawal:
The MHS CI pulls the GTSAM python package for efficiency rather than build the wrapper from scratch. If we wrap something new, that's not already available on PyPI and so we need to push the latest package to PyPI

johnwlambert:
> The MHS CI pulls the GTSAM python package for efficiency rather than build the wrapper from scratch. If we wrap something new, that's not already available on PyPI and so we need to push the latest package to PyPI

Yeah our pypi version is ancient. What's our current versioning policy? Are we on 4.1 or 4.0.1?

ProfFan:
4.1 is still not released, but maybe we should push another pre release

varunagrawal:
Then let's merge this and get to it?

ProfFan (approved):

dellaert:
> The MHS CI pulls the GTSAM python package for efficiency rather than build the wrapper from scratch. If we wrap something new, that's not already available on PyPI and so we need to push the latest package to PyPI

OK, you should probably specify it won't affect *GTSAM* CI. This is a GTSAM PR.

varunagrawal:
Gotcha. I mentioned this was needed for MHS but in my head I figured it was implied.

PR #933, opened 2021-11-20
Modernize discrete factor graphs (a bit)

dellaert:
- use emplace_shared were we can
- bring use of Values up to date with rest of GTSAM (no shared pointer mania)

Technically an API change, but discrete is very seldomly used, if ever, except by me I think.

### DIFF
--- a/gtsam_unstable/discrete/tests/testCSP.cpp
+++ b/gtsam_unstable/discrete/tests/testCSP.cpp
@@ -73,11 +73,11 @@ TEST_UNSAFE(CSP, allInOne) {
   EXPECT(assert_equal(expectedProduct, product));
 
   // Solve
-  CSP::sharedValues mpe = csp.optimalAssignment();
+  auto mpe = csp.optimalAssignment();
   CSP::Values expected;
   insert(expected)(ID.first, 1)(UT.first, 0)(AZ.first, 1);
-  EXPECT(assert_equal(expected, *mpe));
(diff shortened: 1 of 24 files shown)


gtsam_unstable/discrete/tests/testCSP.cpp
```diff
@@ -120,8 +120,8 @@ TEST_UNSAFE(CSP, WesternUS) {
   Ordering ordering;
   ordering += Key(0), Key(1), Key(2), Key(3), Key(4), Key(5), Key(6), Key(7),
       Key(8), Key(9), Key(10);
-  CSP::sharedValues mpe = csp.optimalAssignment(ordering);
-  // GTSAM_PRINT(*mpe);
+  auto mpe = csp.optimalAssignment(ordering);
+  // GTSAM_PRINT(mpe);
   CSP::Values expected;
   insert(expected)(WA.first, 1)(CA.first, 1)(NV.first, 3)(OR.first, 0)(
```
ProfFan, line 126: Sorry for jumping in, but I remember last time you asked me to wrap some discrete in Python, and I realized this is not easy to wrap. Could we drop this syntax (which is boost specific) and use some C++11-only style instead?

dellaert: different PR

xsj01:
Looks good to me

xsj01 (approved):

dellaert:
Thanks @xsj01, but note reviewers don't merge :-)

Issue #927, opened 2021-11-12
Add Deprecation Pragmas


varunagrawal:
## Feature

Add the following deprecation pragmas and use them so that GTSAM warns users during compilation of their projects.
Creating this issue so we can track this.

```
#if defined(__GNUC__) || defined(__clang__)
#define GTSAM_DEPRECATED __attribute__((deprecated))
#elif defined(_MSC_VER)
#define GTSAM_DEPRECATED __declspec(deprecated)
#else
#define GTSAM_DEPRECATED
#endif
```

## Motivation

Deprecation warnings!

jlblancoc:
For the records: This one could be replaced by standard C++14 [deprecated attribute](https://en.cppreference.com/w/cpp/language/attributes/deprecated), if #937 ever got addressed.

PR #934, opened 2021-11-21
Add deprecations

ProfFan:
#927

### DIFF
--- a/gtsam/discrete/DiscreteConditional.cpp
+++ b/gtsam/discrete/DiscreteConditional.cpp
@@ -91,7 +91,7 @@ bool DiscreteConditional::equals(const DiscreteFactor& other,
     return false;
   else {
     const DecisionTreeFactor& f(
-        dynamic_cast<const DecisionTreeFactor&>(other));
+        static_cast<const DecisionTreeFactor&>(other));
     return DecisionTreeFactor::equals(f, tol);
   }
 }
--- a/gtsam/discrete/DiscreteConditional.cpp
+++ b/gtsam/discrete/DiscreteConditional.cpp
@@ -35,7 +35,7 @@ using namespace std;
 namespace gtsam {
 
 // Instantiate base class
-template class Conditional<DecisionTreeFactor, DiscreteConditional> ;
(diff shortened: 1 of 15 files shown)


gtsam/discrete/DiscreteConditional.cpp
```diff
@@ -91,7 +91,7 @@ bool DiscreteConditional::equals(const DiscreteFactor& other,
     return false;
   else {
     const DecisionTreeFactor& f(
-        static_cast<const DecisionTreeFactor&>(other));
+        dynamic_cast<const DecisionTreeFactor&>(other));
```
varunagrawal, line 94: What's up with this? Was this needed?
Generally, static_cast is considered safer than dynamic_cast, hence asking.

gtsam/discrete/Potentials.cpp
```diff
@@ -27,8 +27,8 @@ using namespace std;
 namespace gtsam {
 
 // explicit instantiation
-template class DecisionTree<Key, double>;
-template class AlgebraicDecisionTree<Key>;
+//template class GTSAM_EXPORT DecisionTree<Key, double>;
+//template class GTSAM_EXPORT AlgebraicDecisionTree<Key>;
```
varunagrawal, line 31: Accidentally commented out these two?

varunagrawal (changes requested):
Couple of comments

gtsam/discrete/DiscreteConditional.cpp
ProfFan: Oh yeah this is a clang-tidy false positive. I can revert this one

gtsam/discrete/Potentials.cpp
ProfFan: I make the template GTSAM_EXPORT so these two are not needed. Need to check if Windows is happy with this or not...

varunagrawal: Gotcha. If they are not needed then kill them.

varunagrawal (approved):
LGTM

PR #938, opened 2021-11-24
IMUKittiExampleGPS.m

cxwoaiwojiaa:
Hello, nice to meet you, and thank you very much for providing such a useful example for new users of gtsam. About the program (imukittiexamplegps. M), I have successfully run it. Now I want to test it with different data. As a first step, I try to generate data in the same format from Kitti raw data. I want to know   Kittiequivbiasedimu.txt and attached   Kittigps_converted.txt what do these two data components represent respectively, such as attached   What are the meanings of X, y and Z in kittigps_converted.txt? What are the meanings of X, y and Z in kittiequivbiasedimu.txt. Looking forward to your reply!!!


ProfFan:
Please DO NOT spam the repo! Discussions should be directed to the **mailing list** https://groups.google.com/g/gtsam-users

PR #941, opened 2021-11-27
Wrap additional ISAM2 methods

varunagrawal:
Wrapping additional methods in ISAM2 needed for legged state estimation.

### DIFF
--- a/gtsam/linear/NoiseModel.h
+++ b/gtsam/linear/NoiseModel.h
@@ -177,16 +177,15 @@ namespace gtsam {
         return *sqrt_information_;
       }
 
-    protected:
-
-      /** protected constructor takes square root information matrix */
-      Gaussian(size_t dim = 1, const boost::optional<Matrix>& sqrt_information = boost::none) :
-        Base(dim), sqrt_information_(sqrt_information) {
-      }
 
     public:
 
       typedef boost::shared_ptr<Gaussian> shared_ptr;
+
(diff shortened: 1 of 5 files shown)


dellaert (changes requested):
Confused. Most additions are commented out. Make draft PR until finalized.

varunagrawal:
Okay removed out the commented stuff. That was my mistake, I added those in as part of another change not related to this PR. Thank you for catching them. :slightly_smiling_face:

dellaert (approved):
LGTM, but is that the only method you want to add? Are there others public methods that could easily be added in the same PR?

varunagrawal:
No other methods so far that we need for MHS + State Estimator. If we find something, we can make a PR for then.

dellaert:
My question was not about MHS, but expanding this PR to include other methods now that are easily wrapped.

varunagrawal:
We can. It just depends on the requirements. My comment about MHS is that the current wrapped method is all I need so I didn't spend too much time looking at other methods.

dellaert (changes requested):
OK, you're (choosing) to not hear me :-) For this *GTSAM* PR, please also wrap any other easy wrap-able functions before merging, so GTSAM users have this convenience in the future.

varunagrawal:
Gotcha

dellaert (approved):

dellaert:
Cool, thanks !!!

PR #942, opened 2021-11-29
Enable pickling

varunagrawal:
Enable pickling so we can deep copy a bunch of things in the Python wrapper.

The wrapper expects declaration of the `pickle` method in the interface file and will use the underlying `serialize` method of the class. The wrapper has had this feature for a while, but the classes updated in this PR didn't have the pickling enabled.

### DIFF
--- a/gtsam/linear/NoiseModel.h
+++ b/gtsam/linear/NoiseModel.h
@@ -177,16 +177,15 @@ namespace gtsam {
         return *sqrt_information_;
       }
 
-    protected:
-
-      /** protected constructor takes square root information matrix */
-      Gaussian(size_t dim = 1, const boost::optional<Matrix>& sqrt_information = boost::none) :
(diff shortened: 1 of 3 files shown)


dellaert (changes requested):
How does this PR enable pickling? Does this rely on functionality in wrap? Is that the wrap version currently included with GTSAM? Add more context in PR comment and req-request review?

varunagrawal:
@dellaert added details on pickling with the wrapper.

dellaert (approved):
Not holding up this PR, but why do we have a separate pickle and serialize method? Is not every serializable class pickalable?

varunagrawal:
It is. I noticed this exact problem when making this PR, but this requires a change to the wrapper. I can do it over the next week after higher priority items have been completed. :slightly_smiling_face:

PR #943, opened 2021-11-29
Fix deprecations

varunagrawal:
This PR fixes all the `axpy` deprecation messages that now (correctly) pop up.


dellaert:
I don't understand. Where is it stated axpy is deprecated? I looked but could not find.

varunagrawal:
This was done in #934 in line with #927

dellaert (approved):

PR #945, opened 2021-11-30
Minor fixes to fix some warnings

bricerebsamen:
When I was building gtsam with my build system and my compiler flags I got a bunch or errors:
- return type should not be const `static const double fn()` --> `static double fn()`
- missing shared_ptr includes in 2 files
- missing arg in initializer list when initializing `ParseMeasurement<Pose2>`

I also added `*.swp` to `.gitignore` to ignore vim swp files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .idea
 *.pyc
 *.DS_Store
+*.swp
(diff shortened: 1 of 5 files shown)


varunagrawal (approved):
LGTM

PR #946, opened 2021-12-01
add missing interface for PoseTranslationPrior

StefanKaiser-TomTom:
Change is straight forward.

However, I'm not fully aware how the _pickle()_ works .. I can't call it from Python even though its listed in the interface and some stuff is done in the [wrapper](https://github.com/borglab/wrap/blob/d90abb52badb08c1c1a42b901a15fa5877024370/gtwrap/pybind_wrapper.py#L109)?
Lastly, I'd put up some unit-tests but would need some guidance where to put it, and how you want those interfaces being tested (or maybe you don't want to test them at all for the sake of flexibility?).

### DIFF
--- a/gtsam/slam/slam.i
+++ b/gtsam/slam/slam.i
@@ -168,6 +168,13 @@ template <POSE>
 virtual class PoseTranslationPrior : gtsam::NoiseModelFactor {
   PoseTranslationPrior(size_t key, const POSE& pose_z,
                        const gtsam::noiseModel::Base* noiseModel);
+  POSE measured() const;
+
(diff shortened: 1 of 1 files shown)


varunagrawal:
Thanks for the PR @StefanKaiser-TomTom. Yeah that is an oversight of ours in the wrapper. You should be able to call  `pickle.dump()` on the object to get the serialization to work for you.

We are already working towards removing the need to define `pickle()` in the .i files.

varunagrawal (approved):
LGTM

dellaert:
Great! thanks both!

PR #947, opened 2021-12-01
NoiseModelFactorN - fixed-number of variables >6

gchenfc:
Although this doesn't have an immediate use-case, I just wanted to work on this as a de-stressing exercise.  Will not feel bad if it is rejected.

Currently there is `NoiseModelFactor1`, `NoiseModelFactor2`, ..., `NoiseModelFactor6` but nothing higher, so if you want to implement n-way factors where n is known at compile time but greater than 6, you have to derive from `NoiseModelFactor` and implement `unwhitenedError`.  This is fine, but I thought it'd be nice to have an n-way version that still has the convenient `evaluateError` syntax.

This might also be nice to use in place of existing `NoiseModelFactor1-6` classes in the future since, if the number of arguments for a given factor ever changes, it requires fewer changes (N is figured out by the compiler rather than hard-coded).


varunagrawal:
Dude, I love this!

dellaert (commented):
I love this. Is there a way to replace some of the earlier factors with a typedef  to this new factor? has to be 100% compatible of course.

gchenfc:
Thanks!

> I love this. Is there a way to replace some of the earlier factors with a typedef to this new factor? has to be 100% compatible of course.

I did consider this but decided to wait what you thought first.  Glad you feel similarly :)

I think the only extra things that wouldn't be possible with the typedef would be:
1. `inline Key key1() const { return keys_[0]; }` etc.
2. `typedef VALUE1 X1;` etc.

Because these have literal numbers in their names (and macros are processed before templates so we can't use macro magic to get these names).

After fixing the CI failure, I'll draft a potential solution and see what you think.

dellaert:
We can just add key want to key six and have it fail if N< less than six :-) There might even be template magic to only instantiate these methods up to N.

gchenfc:
Oh true I forgot about that!

I implemented that - it's mostly done, except that the `using` syntax:
```c++
template <class VALUE1, class VALUE2, ...., class VALUE6>
using NoiseModelFactor6 = NoiseModelFactorN<VALUE1, VALUE2, VALUE3, VALUE4, VALUE5, VALUE6>;
```
is causing some serialization issues.  Here's a MWE:
https://godbolt.org/z/WjrbfWMnT
which is perplexing because if you change `using BaseAlias = Base<T>;` to `struct BaseAlias : Base<T> {};` it works, even though `BaseAlias` is still not a type without giving it template parameters.  Or even if you use `Base` which is also templated (and how it is currently in GTSAM) it still works.

I'll think about this more later and maybe ask fan or varun for help.

dellaert:
Ping @gchenfc - I prefer you ask help from someone not Varun or Fan - both racing to deadlines.

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +307,306 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
+  ~NoiseModelFactorN() override {}
 
-  /** Serialization function */
-  friend class boost::serialization::access;
-  template<class ARCHIVE>
-  void serialize(ARCHIVE & ar, const unsigned int /*version*/) {
-    ar & boost::serialization::make_nvp("NoiseModelFactor",
-        boost::serialization::base_object<Base>(*this));
+  /** Methods to retrieve keys */
+#define SUB(Old, New) template <int Old = New>  // to delay template deduction
```
dellaert, line 471: Indent?

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +307,306 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
+#define KEY_IF_TRUE(Enable) typename std::enable_if<(Enable), Key>::type
+  // templated version of `key<I>()`
```
dellaert, line 473: Properly document and separate from old functions, which should be deprecated and put in deprecated section near end of class.

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +307,306 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
-protected:
-
-  typedef NoiseModelFactor Base;
-  typedef NoiseModelFactor1<VALUE> This;
-
-public:
+template <class... VALUES>
+class NoiseModelFactorN
+    : public NoiseModelFactor,
+      public detail::AliasX<VALUES...>,   // using X = VALUE1
```
dellaert, line 404: I strongly dislike this :-) clutters up a *key* class in GTSAM. All to have aliases X1 etc? Do we even use them anywhere?

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +307,306 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
 
 
 /* ************************************************************************* */
+/* We need some helper structs to help us with NoiseModelFactorN - specifically
```
dellaert, line 310: A great exercise in meta-programming which I’d love to see disappear, see comment below :-)

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -29,6 +29,41 @@
 #include <boost/serialization/base_object.hpp>
 #include <boost/assign/list_of.hpp>
 
+#if BOOST_VERSION >= 106600
```
dellaert, line 32: Why do we need this stuff? Document.

gtsam/mainpage.dox
```diff
@@ -16,7 +16,7 @@ To use GTSAM to solve your own problems, you will often have to create new facto
 
 -# The number of variables your factor involves is <b>unknown</b> at compile time - derive from NoiseModelFactor and implement NoiseModelFactor::unwhitenedError()
    - This is a factor expressing the sum-of-squares error between a measurement \f$ z \f$ and a measurement prediction function \f$ h(x) \f$, on which the errors are expected to follow some distribution specified by a noise model (see noiseModel).
--# The number of variables your factor involves is <b>known</b> at compile time and is between 1 and 6 - derive from NoiseModelFactor1, NoiseModelFactor2, NoiseModelFactor3, NoiseModelFactor4, NoiseModelFactor5, or NoiseModelFactor6, and implement <b>\c evaluateError()</b>
+-# The number of variables your factor involves is <b>known</b> at compile time and is between 1 and 6 - derive from NoiseModelFactor1, NoiseModelFactor2, NoiseModelFactor3, NoiseModelFactor4, NoiseModelFactor5, or NoiseModelFactor6, and implement <b>\c evaluateError()</b>.  If the number of variables is greater than 6, derive from NoiseModelFactorN.
```
dellaert, line 19: Seems to conflict with “deprecated” comment below.

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +307,306 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
   }
-}; // \class NoiseModelFactor6
+};  // \class NoiseModelFactorN
 
 /* ************************************************************************* */
+/** @deprecated: use NoiseModelFactorN
+ * Convenient base classes for creating your own NoiseModelFactors with 1-6
+ * variables.  To derive from these classes, implement evaluateError().
+ *
+ * Note: we cannot use `using NoiseModelFactor1 = NoiseModelFactorN<VALUE>` due
```
dellaert, line 593: Can you comment on that ?

dellaert (changes requested):

gtsam/nonlinear/NonlinearFactor.h
gchenfc: I can indent, but clang-format auto un-indents it since it's a macro/preprocessor thing (https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives)

gtsam/nonlinear/NonlinearFactor.h
gchenfc: Another option is to just typedef them to void if they shouldn't exist.  This makes the code cleaner at the expense of less stringent compile-time checks.  For example, something like:

```c++
using X2 = std::conditional< (N > 2), VALUES[2], void >
```

but then someone might get confused if they do e.g. `NoiseModelFactorN<double, double>::X6 myVariable = 3.2` and then the compiler will give some confusing message (as opposed to just "X2 does not exist").




> Do we even use them anywhere?

I don't think it's common, but does occur in some places, e.g.
https://github.com/borglab/gtsam/blob/7f80c906c4cadc32e71acbe5209f2dba90ace779/tests/testExtendedKalmanFilter.cpp#L192-L194
or, strangely, `BoundingConstraint` redefines it:
https://github.com/borglab/gtsam/blob/7f80c906c4cadc32e71acbe5209f2dba90ace779/gtsam/slam/BoundingConstraint.h#L99-L102

But also, backwards compatibility?  What if users use it?

gtsam/nonlinear/NonlinearFactor.h
gchenfc: If you mean can I explain more in detail here, yes:

In c++, a derived class can refer to its base class either _with_ or _without_ template parameters.  This is a syntactical convenience: you can always refer to yourself (or your visible ancestors) from within your class scope.  See [injected-class-name](https://en.cppreference.com/w/cpp/language/injected-class-name) for full description.

Here's a simple example, where `Derived` can either use the `B` form _or_ the `B<T>` form.

```c++
template <typename T>
struct Base {
  using Type = T;
};

template <typename T>
struct Derived : public Base<T>{
  using Type1 = Base::Type;
  using Type2 = Base::<T>::Type;
};
```

As a note, this means instead of the common pattern
```c++
using Base = NoiseModelFactor4<T1, T2, T3, T4>;
```
I'm pretty sure we could instead just do
```c++
using Base = NoiseModelFactor4;
```

The reason this is problematic is because if you do
```c++
template <typename T>
using Derived = Base<T>
```

Just a weird quirk that anyone who inherits from `Derived` does NOT get class name injection so they _must_ use `Base<T>` and _cannot_ use just `Base`.  I'm not sure why the spec was written this way.

In one test file in the repo, someone does use just `NoiseModelFactor6` in a derived class without specifying the template parameters.  We could just change the unit test, but then that would be in conflict with:
> has to be 100% compatible of course.

Also, if we try doing something tricky like

```c++
class NoiseModelFactorN<...> {
  ...
  using NoiseModelFactor4 = This;
};
```

Then the `NoiseModelFactor4<T>` syntax doesn't work in the derived class.  So basically I'm almost certain there isn't a way to be able to use both syntaxes: `NoiseModelFactor4` and `NoiseModelFactor4<T1, T2, T3, T4>` using only `using`.

An alternative is to define structs:

```c++
template <...>
struct NoiseModelFactor4 : public NoiseModelFactorN<...> {
 public:
  using NoiseModelFactorN<...>::NoiseModelFactorN;
  using This = NoiseModelFactor4<...>;
};
```

But I don't see this as being significantly better than the typedefs.

One advantage of structs is that we could then move the deprecated `keyX()` functions and `using X` into there.

gtsam/nonlinear/NonlinearFactor.h
gchenfc: done

gtsam/nonlinear/NonlinearFactor.h
gchenfc: done

gtsam/mainpage.dox
gchenfc: done

jlblancoc:
Might be doable with the `typename... Args` with clearer syntax? See e.g. #444 
Just a first guess, perhaps it's not possible in this case for some reason.

gchenfc:
> Might be doable with the `typename... Args` with clearer syntax? See e.g. #444 Just a first guess, perhaps it's not possible in this case for some reason.

Thanks so much for the tip @jlblancoc!  Would you mind clarifying which aspects you were thinking to include this?  It's currently being used quite a bit here, e.g.
https://github.com/borglab/gtsam/blob/11fd8612269690c9d95eba4a80ac5db705906aa1/gtsam/nonlinear/NonlinearFactor.h#L404-L406
https://github.com/borglab/gtsam/blob/11fd8612269690c9d95eba4a80ac5db705906aa1/gtsam/nonlinear/NonlinearFactor.h#L519-L520

but I'm hung up on all the "custom" additions for backwards compatibility, e.g. `Key key1 const { return keys_[0]; }` and `typedef X2 VALUE2` which are the majority of what is making this so ugly at the moment - would absolutely love any ideas/suggestions for alternatives if you have any!

gtsam/nonlinear/NonlinearFactor.h
gchenfc: @dellaert please advise on recommended course of action.  Options:
1) keep current
2) typedef to `void` if the type should not apply to this set of templates (cleaner using std::conditional).  e.g. `using X1 = std::conditional< (1 <= N), VALUES[0], void >`
3) delete X1, X2, ...  since they're rarely used (backwards compatibility violation)
4) change `#define NoiseModelFactor4 NoiseModelFactorN` to `struct NoiseModelFactor4 : public NoiseModelFactorN` and move these custom typedefs into there
5) other

jlblancoc:
@gchenfc Ah, very good job then, IMO you are on the right track!  :+1: 

> but I'm hung up on all the "custom" additions for backwards compatibility, e.g. `Key key1 const { return keys_[0]; }`

I'm afraid gtsam is so used, it needs that backwards compatibility, so it should be kept despite the "ugliness" ;-)

gchenfc:
Cool, thanks for sharing + advice! much appreciated :)

gchenfc:
Just for posterity, recapping a quick offline discussion with Frank:

We'll get rid of the template meta-programming needed for the key1() key2(), etc. functions and `typedef X1`, X2, etc. and that'll make things much easier to read (but we'll still have the template parameter packs etc. for other NoiseModelFactorN stuff e.g. evaluateError, obviously).  In their place, we'll just have e.g.
```c++
class NoiseModelFactor3 : public NoiseModelFactorN {
 public:
  using X1 = VALUE<0>;
  Key key1() const { return keys_[0]; };
  ...
};
```

This is similar to the state of the code at commit [ea7d6947](https://github.com/borglab/gtsam/pull/947/files/ea7d769476e897f786799c5f774846460e880e19).  For example, [NoiseModelFactor1](https://github.com/borglab/gtsam/pull/947/files/ea7d769476e897f786799c5f774846460e880e19#diff-34f9b717c856740c8ada0fe7c4b1f746e979ddfebb5fedb48f73dc72248007a8R501-R533)

I'll revert to this commit, clean up, add a unit test to make sure serialization still behaves the same way, then should be ready to go.

dellaert:
ping

dellaert:
CI seem to be failing still. Please re-request review when needed.

gchenfc:
Re-requested review, Python CI should pass after #1250

mattking-smith:
@gchenfc Will I be able to test this NoiseModelFactorN in `Matlab`? What is holding up the merge on this PR? I know me and my lab mate have both needed this in the past.

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,439 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
-public:
+template <class... VALUES>
+class NoiseModelFactorN : public NoiseModelFactor {
+ public:
+  /// N is the number of variables (N-way factor)
+  enum { N = sizeof...(VALUES) };
+
+  /** The type of the i'th template param can be obtained as X<I> */
+  template <int I, typename std::enable_if<(I < N), bool>::type = true>
+  using X = typename std::tuple_element<I, std::tuple<VALUES...>>::type;
```
varunagrawal, line 313: Would you be opposed to renaming this from `X` to `TYPE` or `T`? I can envision a scenario where a user sees `X` and thinks it is the symbol shorthand.

From a readability perspective, I prefer `MyFactor::TYPE<1>`.

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,439 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
+ public:
   /// @name Constructors
   /// @{
 
-  /** Default constructor for I/O only */
-  NoiseModelFactor1() {}
-
-  ~NoiseModelFactor1() override {}
+  /**
+   * Default Constructor for I/O
```
varunagrawal, line 334: This can be a one line docstring
```
/// Default Constructor for I/O
```

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,439 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
+    assert(keys.size() == N);
+  }
 
   /// @}
+
+  ~NoiseModelFactorN() override {}
+
+  /** Returns a key. Usage: `key<I>()` returns the I'th key. */
+  template <int I>
+  inline typename std::enable_if<(I < N), Key>::type key() const {
```
varunagrawal, line 367: This is neat!

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,439 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
 
-public:
+/* ************************************************************************* */
+/** @deprecated: use NoiseModelFactorN, replacing .key1() with .key<1> and X1
+ * with X<1>.
+ * A convenient base class for creating your own NoiseModelFactor
+ * with 2 variables.  To derive from this class, implement evaluateError().
+ */
+template <class VALUE1, class VALUE2>
+class NoiseModelFactor2 : public NoiseModelFactorN<VALUE1, VALUE2> {
```
varunagrawal, line 511: Add the `GTSAM_DEPRECATED` attribute to these classes?

```cpp
class GTSAM_DEPRECATED NoiseModelFactor2 ...
```

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,439 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
+};  // \class NoiseModelFactor2
 
+/* ************************************************************************* */
+/** @deprecated: use NoiseModelFactorN, replacing .key1() with .key<1> and X1
+ * with X<1>.
+ * A convenient base class for creating your own NoiseModelFactor
+ * with 3 variables.  To derive from this class, implement evaluateError().
+ */
+template <class VALUE1, class VALUE2, class VALUE3>
+class NoiseModelFactor3 : public NoiseModelFactorN<VALUE1, VALUE2, VALUE3> {
```
varunagrawal, line 547: Same here `GTSAM_DEPRECATED`

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,439 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
-  NoiseModelFactor4(const SharedNoiseModel& noiseModel, Key j1, Key j2, Key j3, Key j4) :
-    Base(noiseModel, cref_list_of<4>(j1)(j2)(j3)(j4)) {}
-
+/** @deprecated: use NoiseModelFactorN, replacing .key1() with .key<1> and X1
+ * with X<1>.
+ * A convenient base class for creating your own NoiseModelFactor
+ * with 4 variables.  To derive from this class, implement evaluateError().
+ */
+template <class VALUE1, class VALUE2, class VALUE3, class VALUE4>
+class NoiseModelFactor4
```
varunagrawal, line 585: Same here `GTSAM_DEPRECATED`

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,439 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
-  NoiseModelFactor5(const SharedNoiseModel& noiseModel, Key j1, Key j2, Key j3, Key j4, Key j5) :
-    Base(noiseModel, cref_list_of<5>(j1)(j2)(j3)(j4)(j5)) {}
-
+/** @deprecated: use NoiseModelFactorN, replacing .key1() with .key<1> and X1
+ * with X<1>.
+ * A convenient base class for creating your own NoiseModelFactor
+ * with 5 variables.  To derive from this class, implement evaluateError().
+ */
+template <class VALUE1, class VALUE2, class VALUE3, class VALUE4, class VALUE5>
+class NoiseModelFactor5
```
varunagrawal, line 626: Same here `GTSAM_DEPRECATED`

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,439 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
-    Base(noiseModel, cref_list_of<6>(j1)(j2)(j3)(j4)(j5)(j6)) {}
-
+/** @deprecated: use NoiseModelFactorN, replacing .key1() with .key<1> and X1
+ * with X<1>.
+ * A convenient base class for creating your own NoiseModelFactor
+ * with 6 variables.  To derive from this class, implement evaluateError().
+ */
+template <class VALUE1, class VALUE2, class VALUE3, class VALUE4, class VALUE5,
+          class VALUE6>
+class NoiseModelFactor6
```
varunagrawal, line 671: Same here `GTSAM_DEPRECATED`

varunagrawal (approved):
@gchenfc this looks great except for some nitpicks. Can you please make the modifications and we can land this?

gtsam/nonlinear/NonlinearFactor.h
varunagrawal: Or even `MyFactor::type<1>` similar to `MyFactor::key<1>` later.

gchenfc: I'm happy to change, e.g. to `T<#>`, but my logic here was to match the old API, which uses `X1`, `X2`, ...
I think this was probably done in the original version since the input variable (e.g. in unwhitenedError, evaluateError, etc.) is always named with `x`, so the type should be capital `X`.  I guess `x` is used to denote an "input" variable, as in f(x).

Can you please re-confirm whether you prefer this change?  Then I'll apply it or resolve convo.

gchenfc:
Thanks @varunagrawal for the review!  I'll go through and add DEPRECATED - I think you created that after I had originally made this PR so I'm glad you reminded me of this!  Can you also confirm the other 2 items and, if you still want those changes, I'll add those changes too.



> @gchenfc Will I be able to test this NoiseModelFactorN in `Matlab`? What is holding up the merge on this PR? I know me and my lab mate have both needed this in the past.

I don't think this can directly be used in Matlab since `NoiseModelFactorN` is templated and you need to supply the types of all the variables during compile-time.  Actually, now that I think of it, I forget how e.g. NoiseModelFactor2 is even used in Matlab, since you can't define custom factors?  Because `NoiseModelFactor1`, ..., `NoiseModelFactorN` are primarily used as base classes for defining custom factors.  But it would be easier (less copy-paste) to define e.g. `NoiseModelFactor8<T1, T2, ..., T8>` if hypothetically you wanted it.  If it's just for linear, we can do that in a different PR.  If it's for nonlinear, I'm not sure how this would help in Matlab since custom factors need to be defined in C++ anyway (unless using Fan's CustomFactor class).
And I think the PR was previously just waiting for review, so thanks varun again :)

gtsam/nonlinear/NonlinearFactor.h
varunagrawal: If the old API had `X` then let's match it.

gtsam/nonlinear/NonlinearFactor.h
varunagrawal: I've generally seen that we use `///` for single line comments. We definitely don't follow the Google Style in this case. 😅

varunagrawal:
`GTSAM_DEPRECATED` was already present even before I started working with GTSAM. @dellaert and I have been deprecating stuff as we revisit components for hybrid estimation (e.g. `EliminateForMPE`) which is how I know about it, so passing that knowledge on. :)

@mattking-smith we'll be able to write a C++ class with 8+ variables (which isn't doable in current `develop`) and then we can wrap that class. We can talk about it offline (especially since now that I better understand what you're doing).

gtsam/nonlinear/NonlinearFactor.h
gchenfc: done

gtsam/nonlinear/NonlinearFactor.h
gchenfc: done

gtsam/nonlinear/NonlinearFactor.h
gchenfc: done

gtsam/nonlinear/NonlinearFactor.h
gchenfc: done

gtsam/nonlinear/NonlinearFactor.h
gchenfc: done

gtsam/nonlinear/NonlinearFactor.h
gchenfc: done

gchenfc:
Thanks!  Will merge once CI completes, assuming that github lets me; we may still need an approval from @dellaert  since he previously requested changes (that have since been fixed I think).

gtsam/nonlinear/NonlinearFactor.h
dellaert: I noticed the mucking in the tests with avoiding collisions with symbol_shorthand::X, so I think we should revisit. I propose to rename

`VALUES -> ValueTypes` 

```
using ValueType = ...
```

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,437 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
+  using X = typename std::tuple_element<I, std::tuple<VALUES...>>::type;
+
+ protected:
+  using Base = NoiseModelFactor;
+  using This = NoiseModelFactorN<VALUES...>;
+
+  /* Like std::void_t, except produces `boost::optional<Matrix&>` instead. Used
+   * to expand fixed-type parameter-packs with same length as VALUES */
+  template <typename T>
+  using optional_matrix_type = boost::optional<Matrix&>;
```
dellaert, line 322: OptionalMatrix, and should it be `int i` rather than `typename T` ?

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,437 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
+
+  /* Like std::void_t, except produces `Key` instead. Used to expand fixed-type
+   * parameter-packs with same length as VALUES */
+  template <typename T>
+  using key_type = Key;
```
dellaert, line 327: KeyType, and same comment.

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -272,503 +273,437 @@ class GTSAM_EXPORT NoiseModelFactor: public NonlinearFactor {
...
+   */
+  NoiseModelFactorN(const SharedNoiseModel& noiseModel,
+                    key_type<VALUES>... keys)
+      : Base(noiseModel, std::array<Key, N>{keys...}) {}
 
   /**
-   *  Constructor
-   *  @param noiseModel shared pointer to noise model
-   *  @param key1 by which to look up X value in Values
+   * Constructor.  Only enabled for n-ary factors where n > 1.
```
dellaert, line 348: why?

dellaert (commented):
Some comments. This PR is a year old now :-) Could we meet and put a pin in it ? I really like the idea and want to have it in 4.2.

dellaert:
@gchenfc

gtsam/nonlinear/NonlinearFactor.h
gchenfc: Changed the name, but I think it should still be templated on `typename T` instead of `int i` because it needs to accept a template parameter of type `ValueType`.

The reasoning I think is best described by cppreference description of [std::void_t](https://en.cppreference.com/w/cpp/types/void_t):
> Utility metafunction that maps a [sequence] of any types to the type void

Except in this case we're mapping any `ValueType` (non-sequence) to a `boost::optional<Matrix&>` type.  You may be able to imagine why this is helpful in `evaluateError`, since the arguments come in pairs of `(ValueType x1, ..., Matrix H1, ...)`.

In anticipation of the next question, `OptionalJacobian<N, gtsam::traits<ValueType>::N>` or something along those lines had issues with inheritance & backwards compatibility, IIRC.  But perhaps this would actually be possible in a future PR.

gtsam/nonlinear/NonlinearFactor.h
gchenfc: change the name, see comment above

gtsam/nonlinear/NonlinearFactor.h
gchenfc: Actually I'm not sure.  I changed it to remove this requirement.

No tests break when N = 1 and I added unit tests to also test the N=1 case so I think I might have just gotten overzealous with SFINAE at the time I was writing this.

gtsam/nonlinear/NonlinearFactor.h
gchenfc: Done.

dellaert:
PS, re-reading some above, we definitely need to check that both wrappers (python and matlab) still work. @mattking-smith might be able to help with latter...

gchenfc:
> PS, re-reading some above, we definitely need to check that both wrappers (python and matlab) still work. @mattking-smith might be able to help with latter...

Tested both and all pass!

Python is tested by CI and I also ran on my machine, all pass.

Matlab I tested on my machine and all pass, after needing to change some instances of `symbol` -> `Symbol` due to my computer having case insensitive filenames.

dellaert (approved):
This is an amazing piece of work, Gerry! Will merge so it’s ready to go into 4.2 :-)

PR #949, opened 2021-12-03
Fix Quaternion on M1 (and possibly other platforms)

ProfFan:
Now it aligns with scipy and the matrix version.


dellaert (approved):
Cool, thanks!

PR #951, opened 2021-12-05
Use System Eigen by Default

ProfFan:
^


dellaert (changes requested):
Should be discussed first

varunagrawal:
One question: why?

dellaert:
No response to questions so I will close this for now, we can discuss it in our weekly meeting.

PR #956, opened 2021-12-08
Add override keyword to functions in PoseToPointFactor

dwisth:
Really minor PR to fix compiler warnings about missing `override` keyword.

### DIFF
--- a/gtsam_unstable/slam/PoseToPointFactor.h
+++ b/gtsam_unstable/slam/PoseToPointFactor.h
@@ -45,9 +45,8 @@ class PoseToPointFactor : public NoiseModelFactor2<POSE, POINT> {
   /** implement functions needed for Testable */
 
   /** print */
-  virtual void print(
-      const std::string& s,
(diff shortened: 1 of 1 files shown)


gtsam_unstable/slam/PoseToPointFactor.h
```diff
@@ -45,8 +45,9 @@ class PoseToPointFactor : public NoiseModelFactor2<POSE, POINT> {
   /** implement functions needed for Testable */
 
   /** print */
-  virtual void print(const std::string& s, const KeyFormatter& keyFormatter =
-                                               DefaultKeyFormatter) const {
+  virtual void print(
```
dellaert, line 48: I think we should not have virtual *and* override. One or the other. Surprised that does not trigger a warning (check clang)

dwisth: good catch, my mistake! I just fixed it

dellaert (approved):

Issue #958, opened 2021-12-10
Unify Values types


varunagrawal:
## Feature

Currently discrete factor types require the use of custom `Values` which is just an alias to the `Assignment` class. We can achieve the same objective using the regular `gtsam::Values`.

We should remove all the Values typedefs in `discrete/` and instead use good ol' `gtsam::Values`.

## Motivation

This unification will help with the newly proposed HybridFactors. Tagging @dellaert for visibility.

## Pitch

Update all the code to use gtsam::Values and the tests should all pass.

PR #959, opened 2021-12-10
Add new insert_or_assign method to Values

varunagrawal:
This PR adds a new method `insert_or_assign` in support of #958.

~~Modern databases support the `upsert` method which tries to perform an update and, if a key doesn't exist, defaults to an insert.~~

`insert_or_assign` is a C++17 `std::map` method so this method aligns us better with that.

This simplifies this common pattern:
```cpp
if (values.exists(key)) {
	values.update(key, value);
} else {
	values.insert(key, value);
}
```

to the more concise:
```cpp
values.insert_or_assign(key, value);
```

### DIFF
--- a/gtsam/nonlinear/Values-inl.h
+++ b/gtsam/nonlinear/Values-inl.h
@@ -391,10 +391,10 @@ namespace gtsam {
     update(j, static_cast<const Value&>(GenericValue<ValueType>(val)));
   }
 
-  // upsert with templated value
+  // insert_or_assign with templated value
   template <typename ValueType>
-  void Values::upsert(Key j, const ValueType& val) {
-    upsert(j, static_cast<const Value&>(GenericValue<ValueType>(val)));
+  void Values::insert_or_assign(Key j, const ValueType& val) {
+    insert_or_assign(j, static_cast<const Value&>(GenericValue<ValueType>(val)));
   }
 
 }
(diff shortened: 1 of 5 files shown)


dellaert:
Already exists as tryInsert ?

varunagrawal:
tryInsert doesn't perform an update if the key already exists. upsert will perform the update.

dellaert:
I’m not in love with this. If we modify Values, I’d prefer to align it more with stl::map, especially the c++11 and c++17 versions.

It’s probably better to do a design-review these PRs to the core data structures via an issue that proposes a change, before spending a lot of time.

If you disagree strongly, let’s discuss off-line.

varunagrawal:
Adding some research I did here on templated operator overloading.
Apparently it is possible. We can define this:
```cpp
template <typename T>
T& operator[](Key k) {
...
}
```

but the operator call would be ugly
```cpp
Values values;
Pose3 pose;
values[key] = pose; // Compiler error due to template mismatch!!
values.operator[]<Pose3>(pose);  // Works, but is more verbose and defeats the purpose.
```

dellaert:
Yeah, that’s really ugly. Did you read the entire c++17 interface to standard map though? Is there another function that already exists that does what the imperative operator does?

varunagrawal:
I did. There's `insert_or_assign` which is exactly what `upsert` is. Should we rename upsert to that?

I guess the cpp standards committee likes verbose function names.

dellaert:
Yeah. I like shorter names as well, but `insert_or_assign` will create less surprise for developers.

varunagrawal:
Cool, I'll update the method name. :)

varunagrawal:
@dellaert updated

dellaert (approved):
Nice !

PR #960, opened 2021-12-10
[WIP] Unify Discrete and Continuous Values dict

varunagrawal:
This PR updates the `discrete` module to use `gtsam::Values` instead of `Assignment` maps which should make hybrid inference much easier and cleaner to code up.

1. Use `gtsam::Values` instead of other classes and typedefs.
2. Since all the values for discrete types are `size_t`, various Values operations are templated where required.
3. Unit tests have been updated and all pass on my local machine. CI might reveal more.


varunagrawal:
The macOS CI was failing because the definition of `size_t` is different for Xcode vs GCC/Clang. I update this PR to replace the platform dependent `size_t` with `uint64_t`.

varunagrawal:
We will have to think about this some more. Definitely needs a design review.

PR #963, opened 2021-12-11
Minimal changes to make DCSAM classes work within GTSAM

dellaert:
Added a unit test and made minimal changes to make DCSAM classes work within GTSAM

The biggest change is the namespace change. @varunagrawal and I decided to just adopt Kevin's class structure to be 100% compatible.


keevindoherty (approved):
This all looks great to me! Appreciate those `const &` additions, much nicer :)

PR #965, opened 2021-12-13
Wrap Update

varunagrawal:
This update adds support to the Matlab wrapper for default arguments.


gchenfc (approved):

varunagrawal:
Thanks for the review @gchenfc. I am going to let @mattking-smith submit a review before I merge. :)

mattking-smith (approved):
@gchenfc This work look terrific. Well cone.

mattking-smith:
> Thanks for the review @gchenfc. I am going to let @mattking-smith submit a review before I merge. :)

I looked through everything. I do not see a problem with this. Feel free to merge @varunagrawal

varunagrawal:
Now we just have to figure out the memory leak and we should be good to go with the Matlab wrapper 😄

PR #966, opened 2021-12-13
DiscreteValues

dellaert:
Added a separate type and header for DiscreteValues. If we *do* change to a unified Values type, it will be easy to do a global replace.

### DIFF
--- a/gtsam/discrete/DiscreteFactor.h
+++ b/gtsam/discrete/DiscreteFactor.h
@@ -18,7 +18,7 @@
 
 #pragma once
 
-#include <gtsam/discrete/Assignment.h>
+#include <gtsam/discrete/DiscreteValues.h>
 #include <gtsam/inference/Factor.h>
 #include <gtsam/base/Testable.h>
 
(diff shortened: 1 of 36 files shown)


gtsam/discrete/DiscreteFactor.h
```diff
@@ -104,6 +94,6 @@ class GTSAM_EXPORT DiscreteFactor: public Factor {
 
 // traits
 template<> struct traits<DiscreteFactor> : public Testable<DiscreteFactor> {};
-template<> struct traits<DiscreteFactor::Values> : public Testable<DiscreteFactor::Values> {};
+template<> struct traits<DiscreteValues> : public Testable<DiscreteValues> {};
```
varunagrawal, line 97: We should move this to `DiscreteValues.h`.

varunagrawal (approved):
Awesome stuff, but I have one comment for something very simple.

gtsam/discrete/DiscreteFactor.h
dellaert: Good catch!!!

PR #969, opened 2021-12-17
Barometric factor

PeterQFR:
Add a factor that handles barometric pressure readings as a constraint on the z height of a Pose3 value. 
The conversion from pressure to height is given by conversions from https://www.grc.nasa.gov/www/k-12/airplane/atmosmet.html

The model is 

measured height in meters = Pose.z() + bias + noise;

With Pose and Bias both being values in the graph. 

There are conversions both ways between pressure and height in meters. 

We are only handling the tropospheric conversion from 0 to 11km.

I have to remove CMakelists changes for installation in a Ament environment.

### DIFF
--- a/gtsam/navigation/BarometricFactor.cpp
+++ b/gtsam/navigation/BarometricFactor.cpp
@@ -23,30 +23,31 @@ using namespace std;
 namespace gtsam {
 
 //***************************************************************************
-void BarometricFactor::print(const string& s, const KeyFormatter& keyFormatter) const {
-  cout << (s.empty() ? "" : s + " ") << "Barometric Factor on " << keyFormatter(key1())
-       << "Barometric Bias on " << keyFormatter(key2()) << "\n";
+void BarometricFactor::print(const string& s,
+                             const KeyFormatter& keyFormatter) const {
+    cout << (s.empty() ? "" : s + " ") << "Barometric Factor on "
+         << keyFormatter(key1()) << "Barometric Bias on "
+         << keyFormatter(key2()) << "\n";
 
-  cout << "  Baro measurement: " << nT_ << "\n";
-  noiseModel_->print("  noise model: ");
+    cout << "  Baro measurement: " << nT_ << "\n";
+    noiseModel_->print("  noise model: ");
 }
 
 //***************************************************************************
-bool BarometricFactor::equals(const NonlinearFactor& expected, double tol) const {
-  const This* e = dynamic_cast<const This*>(&expected);
-  return e != nullptr && Base::equals(*e, tol) && traits<double>::Equals(nT_, e->nT_, tol);
+bool BarometricFactor::equals(const NonlinearFactor& expected,
+                              double tol) const {
+    const This* e = dynamic_cast<const This*>(&expected);
+    return e != nullptr && Base::equals(*e, tol) &&
+           traits<double>::Equals(nT_, e->nT_, tol);
 }
 
 //***************************************************************************
-Vector BarometricFactor::evaluateError(const Pose3& p,
-        const double& bias, boost::optional<Matrix&> H,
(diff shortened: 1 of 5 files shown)


dellaert:
Cool! Could you: 
- format with Google style, e.g. cppformat in vs code.
- Add a test with non-zero, non-identity poses
- un comment or remove other test?
Then please request a review by me.

PeterQFR:
I've put in the non-zero tests, however I get an error with the numerical differentiation for the jacobian. Whenever there is any roll or pitch (Rx Ry) applied to the pose, the numerical derivative seems to use that to calculate the change in z in the local frame. 

I believe the analytic jacobians are correct in the evaluateError() but I would not be suprised if they are not. so would like some help deriving them if they are not correct. 

Ie at the moment, for the pose3 only Tz contributes anything to the error so the jacobian should only be dependent on it (and the bias).

PeterQFR:
@dellaert I've applied the formatting and non-zero tests, but need some help with the Jacobians I am returning for the Pose3 value. 

It does not agree with the numerical derivative in the test when Rx or Ry is applied to the Pose3d it is like the numerical derivative is calculated in the child frame of the pose where as I am using it as per the parent frame (to borrow from ROS conventions). The analytical jacobian of the error I think is correct.

gtsam/navigation/BarometricFactor.cpp
```diff
@@ -0,0 +1,53 @@
...
+    return e != nullptr && Base::equals(*e, tol) &&
+           traits<double>::Equals(nT_, e->nT_, tol);
+}
+
+//***************************************************************************
+Vector BarometricFactor::evaluateError(const Pose3& p, const double& bias,
+                                       boost::optional<Matrix&> H,
+                                       boost::optional<Matrix&> H2) const {
+    if (H2) (*H2) = (Matrix(1, 1) << 1.0).finished();
+    if (H) (*H) = (Matrix(1, 6) << 0., 0., 0., 0., 0., 1.).finished();
```
dellaert, line 49: Respect the order :-) also, this is likely to be incorrect. The translation method will actually calculate the correct Jacobian for you.

PeterQFR: Thanks Frank, I've updated it to use the output of the Jacobian from the translation method. The jacobians have been something that I've found challenging for a while. I think its been mostly around the frames used as I typically have found it difficult to derive those given in all but the simplest of examples. The papers and tutorials I've read don't seem to go into depth on the subject. Are there any good tutorials on derivation of analytic jacobians for these type of systems?

It does seem that gtsam has a good system for dealing with this complexity, but I'd like to be better set up to handle these problems in future. 

@dellaert Thats all updated and for review.

dellaert: Great! I started to workflows for CI. As far as the math, take a look in math.pdf, in the docs folder.

ProfFan:
BTW @dellaert do we need a new subfolder for this factor?

dellaert:
> BTW @dellaert do we need a new subfolder for this factor?

No, navigation is fine. Will merge now.

dellaert (approved):
Looks awesome! Many thanks for your contribution!

mnissov:
Doesn't this also need an implementation for the bias dynamics? As far as I can see this factor has no consideration of the bias random walk, which I think is necessary for this to be used?

PeterQFR:
Thats right, but that is common with a lot of factor implementations. Here the random walk is achieved by adding additional between factor between bias values at each measurement step. 
That is, you need to add an additional Between factor, linking each of your Bias Values in the graph. Its not included in this factor because that would mean this factor would then need to reference the bias of the previous timestep. 
This factor doesn't assume anything about the bias which could be constant among other things.

PR #970, opened 2021-12-17
Add wrapper fixes for Matlab memory leak

varunagrawal:
This PR pulls the latest updates from `gtwrap` which has fixes for a memory leakage in the matlab wrapper.
@mattking-smith has already verified this works well, so it should be an easy approval.


dellaert (approved):
Thanks!!!

PR #971, opened 2021-12-20
Streamlined dot methods

dellaert:
Driven by tests in discrete and nonlinear, and two notebooks, I refactored dot methods:

# inference
- added dot/dot/saveGraph pattern in factor graph, bayes net, bayes tree
- added DotWriter class

# discrete
- added dot for DecisionTree and DiscreteConditional
- added tests for dot functionality
- added two notebooks that "show" the graphs

# nonlinear
- inherit methods from inference
- extra dot/dot/saveGraph in NonlinearFactorGraph
- deprecated saveGraph(ostream) method
- moved GraphvizFormatting separate file
- inherit GraphvizFormatting from DotWriter
- added tests for dot functionality

TODO: Gaussian still needs to conform. It's a bit tricky as GaussianBayesNet has to be made a derived class of BayesNet (again?)

Example output from DiscreteSwitching.ipynb:
[bn.svg.pdf](https://github.com/borglab/gtsam/files/7742830/bn.svg.pdf)
[fg.svg.pdf](https://github.com/borglab/gtsam/files/7742831/fg.svg.pdf)
[bt.svg.pdf](https://github.com/borglab/gtsam/files/7742832/bt.svg.pdf)


gtsam/discrete/DecisionTree-inl.h
```diff
@@ -258,8 +259,10 @@ namespace gtsam {
         }
 
         os << "\"" << this->id() << "\" -> \"" << branch->id() << "\"";
-        if (i == 0) os << " [style=dashed]";
-        if (i > 1) os << " [style=bold]";
+        if (B == 2) {
```
varunagrawal, line 262: What happens when `B != 2`?

gtsam/discrete/tests/testDiscreteBayesTree.cpp
```diff
@@ -26,76 +26,89 @@ using namespace boost::assign;
...
-  bayesNet.add((key[0] | key[8], key[12]) = "2/3 1/4 3/2 4/1");
+/* ************************************************************************* */
+TEST(DiscreteBayesTree, ThinTree) {
+  const TestFixture self;
+  const auto& keys = self.keys;
 
   if (debug) {
-    GTSAM_PRINT(bayesNet);
-    bayesNet.saveGraph("/tmp/discreteBayesNet.dot");
+    GTSAM_PRINT(self.bayesNet);
```
varunagrawal, line 85: Any reason for not just doing `auto bayesNet = self.bayesNet`? `self` is a little confusing since this is not a class, but this is a minor nitpick.

gtsam/inference/BayesTree-inst.h
```diff
@@ -63,20 +63,40 @@ namespace gtsam {
...
+          "the root of Bayes tree has not been initialized!");
+    os << "digraph G{\n";
+    for (const sharedClique& root : roots_) dot(os, root, keyFormatter);
+    os << "}";
+    std::flush(os);
+  }
+
+  /* ************************************************************************* */
+  template <class CLIQUE>
+  std::string BayesTree<CLIQUE>::dot(const KeyFormatter& keyFormatter) const {
```
varunagrawal, line 80: Seeing this repeat in multiple places, do you think it is worth having this as a trait? The `Testable` trait has a requirement for `print` so we could have a similar requirement for `dot(ostream&, KeyFormatter&)`.
Just a thought.

varunagrawal (approved):
This is great stuff! I have a few comments, but they can be addressed now or later. 🙂

gtsam/discrete/tests/testDiscreteBayesTree.cpp
dellaert: I'm just using the variable from the test fixture...

gtsam/discrete/DecisionTree-inl.h
dellaert: Regular style

gtsam/inference/BayesTree-inst.h
dellaert: That's a good idea. I'll add a TODO and issue.

dellaert: Actually, I started on the issue but I don't see the reason, after all. traits are useful if they are used for some functionality. We can make traits if it serves a purpose, later.

PR #973, opened 2021-12-20
Add Discrete-Continuous factor graph machinery

varunagrawal:
This PR adds the following:

1. `DCFactor` and `DCFactorGraph` code ported from `DCSAM` (@keevindoherty has already looked at this part).
2. Split all non-templated code into .h and .cpp files.
3. Move the `expNorm` function to `DiscreteFactor.[h|cpp]`.
4. Move the types in `DCSAM_types.h` i.e. `DCValues` and `DCMarginals` to their own files to conform to our codebase structure.
5. Add a new test assertion `assert_type` which asserts whether an object has a specified type.
6. Use of the `assert_type` for simple constructor tests for DCValues and DCMarginals.


gtsam/hybrid/DCFactor.h
```diff
@@ -0,0 +1,264 @@
...
+   *
+   * @return the dimension (number of rows) in the Jacobian of the `error`
+   * function with respect to continuous variables for
+   */
+  virtual size_t dim() const = 0;
+
+  /*
+   * Return the discrete keys for this factor.
+   */
+  gtsam::DiscreteKeys discreteKeys() const { return discreteKeys_; }
```
dellaert, line 136: Const &

gtsam/hybrid/DCFactor.h
```diff
@@ -0,0 +1,264 @@
...
+   * error function can be called setting each discrete key's variable
+   * individually, and the overall DCFactor can itself be factored as a product
+   * of unary gtsam::DecisionTreeFactors.
+   *
+   * Alternative implementations might consider something like the
+   * gtsam::AllDiff approach here:
+   * https://github.com/borglab/gtsam/blob/43e8f1e5aeaf11890262722c1e5e04a11dbf9d75/gtsam_unstable/discrete/AllDiff.cpp#L43
+   *
+   * @param continuousVals - an assignment to the continuous variables
+   * @param discreteVals -
```
dellaert, line 154: What is the role for discreteVals here? @keevindoherty  should provide this knowledge :-)

gtsam/hybrid/DCFactor.h
```diff
@@ -0,0 +1,264 @@
...
+
+  /**
+   * Calculate a normalizing constant for this DCFactor. Most implementations
+   * will be able to use the helper function
+   * nonlinearFactorLogNormalizingConstant provided below for most of the
+   * calculation.
+   * TODO(Kurran) is this the cleanest way to do this? Seems necessary for the
+   * DCMaxMixtureFactor implementations etc...
+   */
+  virtual double logNormalizingConstant(const gtsam::Values& values) const {
```
dellaert, line 169: To cpp?

gtsam/hybrid/DCMarginals.h
```diff
@@ -0,0 +1,32 @@
+/* ----------------------------------------------------------------------------
+ * Copyright 2020 The Ambitious Folks of the MRG
+ * See LICENSE for the license information
+ * -------------------------------------------------------------------------- */
+
+/**
+ *
```
dellaert, line 7: I don’t remember we had added this, @varunagrawal ?

gtsam/hybrid/tests/testDCMarginals.cpp
```diff
@@ -0,0 +1,41 @@
...
+using namespace gtsam;
+using namespace serializationTestHelpers;
+
+//******************************************************************************
+TEST(DCMarginals, Constructor) {
+  Marginals continuous;
+  DiscreteMarginals discrete;
+
+  DCMarginals dc_marginals(continuous, discrete);
+  EXPECT(assert_type("gtsam::DCMarginals", dc_marginals));
```
dellaert, line 33: I don’t understand. Self-fulfilling prophecy?

gtsam/hybrid/tests/testDCValues.cpp
```diff
@@ -0,0 +1,42 @@
...
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ *  @file  testDCValues.cpp
```
dellaert, line 13: Same: is this something you added Varun? Design review??

gtsam/hybrid/DCMarginals.h
varunagrawal: Yup, I added this. This is point 4 in the PR comment: since DCValues and DCMarginals are structs in DC_types.h, I just moved them to their own file so as to be consistent with the rest of GTSAM.

I took a bit of design liberty here, so I hope that's okay?

gtsam/hybrid/tests/testDCMarginals.cpp
varunagrawal: This is a barebones test for the default constructor. This ensures that everything is included correctly and that a basic operation works.

The idea is that since this is a test, it should assert something, so just assert that the type is as expected.

gtsam/hybrid/tests/testDCValues.cpp
varunagrawal: Yes, same reason as for DCMarginals, just an attempt to be consistent. Again, a bit of liberty taken.

dellaert: OK, but please list Kevin as author if you did not change the functionality, just the location. If you did change any code, explain what it is.

gtsam/hybrid/DCMarginals.h
dellaert: Same comment about authorship...

gtsam/hybrid/tests/testDCMarginals.cpp
dellaert: C++ is strongly typed. Test anything, but type is known. Let's cherry pick `assert_type` in a branch if we ever need it, but please leave out of this PR.

dellaert (commented):
Please convert to WIP - I'd like to play with mixture factor before we merge, to better understand.

gtsam/hybrid/tests/testDCValues.cpp
varunagrawal: Kevin is still the author on `DCValues.h`. I am the author on `testDCValues.cpp` since I wrote that whole file by myself (no copy involved).

Same situation for `DCMarginals.h`.

gtsam/hybrid/DCMarginals.h
varunagrawal: This file has Kevin as the sole author as per your recommendation. I only added myself as author to `testDCMarginals.cpp` since I wrote that file from scatch.

gtsam/hybrid/DCFactor.h
keevindoherty: `discreteVals` is an assignment to the discrete variables. This is useful for a kind of discrete-continuous smart factor where we've eliminated some subset of variables conditioned on a joint discrete and continuous assignment. In our case, we are using these for semantic+geometric measurements with ambiguous associations in semantic SLAM. For max-product elimination, check out [this max-mixture factor](https://github.com/MarineRoboticsGroup/dcsam/blob/main/include/dcsam/DCMaxMixtureFactor.h) for example, or for sum-product, we have [a discrete-continuous EMFactor here](https://github.com/MarineRoboticsGroup/dcsam/blob/main/include/dcsam/DCEMFactor.h). There are perhaps other applications I haven't thought about yet. Not all DCFactors have to use these values, but we pass them in case they are needed.

dellaert:
Were you guys not going to move to a fork?

varunagrawal:
Oof that's my bad. @ProfFan do you want to use my fork of GTSAM? You should have access to it.

ProfFan:
Ah sure, I can push to your fork

dellaert:
I think we should close this in favor of @ProfFan 's PR, right? If you agree please close, otherwise let's talk over email...

dellaert:
Closing this again until Fan’s PR is merged and then we can re-open a substantially modified PR for this. @varunagrawal please don’t reopen without talking to me first about the pans regarding that.

PR #976, opened 2021-12-21
Spherical Camera Constructor Fix

varunagrawal:
The `SphericalCamera` constructor which takes a calibration does not use the passed argument to initialize its calibration variable. This PR fixes that.

As a result, more issues were found due to non-initialization of `EmptyCal` shared pointers, so those have been fixed as well.


lucacarlone:
@varunagrawal : very well done, thanks! I don't think the empty calibration is used anywhere, but still good to pass it down.

lucacarlone (approved):
LGTM - thanks!

PR #979, opened 2021-12-21
Conformize various print statements

varunagrawal:
This PR updates all the print functions to have a common pattern and to be cleaner when printing with or without the optional `s` argument.

It also adds defaults for `string& s` where they were missing.

### DIFF
--- a/gtsam/base/deprecated/LieScalar.h
+++ b/gtsam/base/deprecated/LieScalar.h
@@ -49,8 +49,8 @@ namespace gtsam {
 
     /// @name Testable
     /// @{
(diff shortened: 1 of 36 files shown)


ProfFan:
I really, really hate the extra `endl` for the print methods, but I think there is still chance to fix this since 4.1.1 is out and we are on 4.2 mode?

dellaert (changes requested):
No design review. Please convert to WIP and discuss in our next meeting.

varunagrawal:
Closing since this is an API breaking change, and is low priority.

PR #982, opened 2021-12-22
HybridFactorGraph linearize method

varunagrawal:
This PR adds a `linearize` method to the `HybridFactorGraph` which linearizes the nonlinear factors around a given point.

### DIFF
--- a/gtsam/hybrid/DCFactor.h
+++ b/gtsam/hybrid/DCFactor.h
@@ -105,6 +105,9 @@ class DCFactor : public gtsam::Factor {
       const gtsam::Values& continuousVals,
       const DiscreteValues& discreteVals) const = 0;
 
+  virtual DCFactor::shared_ptr linearize(
+      const Values& continuousVals) const = 0;
+
   /**
    * Returns true when the DCFactor is equal to `other`
    *
--- a/gtsam/hybrid/DCFactor.h
+++ b/gtsam/hybrid/DCFactor.h
@@ -105,6 +105,13 @@ class DCFactor : public gtsam::Factor {
       const gtsam::Values& continuousVals,
       const DiscreteValues& discreteVals) const = 0;
 
+  /**
+   * @brief Linearize all the continuous factors only with respect to the
+   * continuous variables (as given in `keys_`).
+   *
+   * @param continuousVals The continuous variables referenced by `keys_`.
+   * @return DCFactor::shared_ptr
+   */
   virtual DCFactor::shared_ptr linearize(
       const Values& continuousVals) const = 0;
 
--- a/gtsam/hybrid/DCFactor.h
+++ b/gtsam/hybrid/DCFactor.h
@@ -39,7 +39,7 @@ class DCFactor : public gtsam::Factor {
 
  public:
   using Base = gtsam::Factor;
-  using shared_ptr = boost::shared_ptr<Factor>;
+  using shared_ptr = boost::shared_ptr<DCFactor>;
 
   DCFactor() = default;
 
@@ -101,7 +101,7 @@ class DCFactor : public gtsam::Factor {
    * @param discreteVals - Likewise, assignment to the discrete variables in
    * `discreteKeys__`.
    */
-  virtual boost::shared_ptr<gtsam::GaussianFactor> linearize(
+  virtual GaussianFactor::shared_ptr linearize(
       const gtsam::Values& continuousVals,
       const DiscreteValues& discreteVals) const = 0;
 
--- a/gtsam/hybrid/DCFactor.h
+++ b/gtsam/hybrid/DCFactor.h
@@ -109,6 +109,10 @@ class DCFactor : public gtsam::Factor {
    * @brief Linearize all the continuous factors only with respect to the
    * continuous variables (as given in `keys_`).
    *
+   * This `linearize` is different from the `linearize(continuous, discrete)` in
+   * that it assumes no assignment of discrete keys and linearizes all
+   * continuous factors associated with this factor.
+   *
    * @param continuousVals The continuous variables referenced by `keys_`.
    * @return DCFactor::shared_ptr
    */
--- a/gtsam/hybrid/DCGaussianMixtureFactor.h
+++ b/gtsam/hybrid/DCGaussianMixtureFactor.h
@@ -80,10 +80,10 @@ class DCGaussianMixtureFactor : public DCFactor {
     return factors_[0];
   }
 
-  /// Return this linear factor as a DCFactor::shared_ptr
+  /// Return linearized version of this factor.
   virtual DCFactor::shared_ptr linearize(
       const Values& continuousVals) const override {
-    return boost::make_shared<DCGaussianMixtureFactor>(*this);
+    throw std::runtime_error("DCGaussianFactor is already linear");
   }
 
   // TODO(dellaert): implement
--- a/gtsam/hybrid/DCGaussianMixtureFactor.h
+++ b/gtsam/hybrid/DCGaussianMixtureFactor.h
@@ -67,17 +67,17 @@ class DCGaussianMixtureFactor : public DCFactor {
     return 0;
   }
 
-  // TODO(dellaert): Does not make sense
   double error(const gtsam::Values& continuousVals,
                const gtsam::DiscreteValues& discreteVals) const override {
-    return 0;
+    throw std::runtime_error("DCGaussianMixtureFactor::error not implemented");
   }
 
-  // TODO(dellaert): Does not make sense
   GaussianFactor::shared_ptr linearize(
       const gtsam::Values& continuousVals,
       const DiscreteValues& discreteVals) const override {
-    return factors_[0];
+    throw std::runtime_error(
+        "DCGaussianMixtureFactor::linearize(continuous, discrete) not "
+        "implemented");
   }
 
   /// Return linearized version of this factor.
@@ -87,7 +87,9 @@ class DCGaussianMixtureFactor : public DCFactor {
   }
 
   // TODO(dellaert): implement
-  size_t dim() const override { return 0; };
+  size_t dim() const override {
+    throw std::runtime_error("DCGaussianMixtureFactor::dim not implemented");
+  };
 
   /// Testable
   /// @{
--- a/gtsam/hybrid/DCGaussianMixtureFactor.h
+++ b/gtsam/hybrid/DCGaussianMixtureFactor.h
@@ -51,12 +51,9 @@ class DCGaussianMixtureFactor : public DCFactor {
   DCGaussianMixtureFactor() = default;
 
   DCGaussianMixtureFactor(
-      const KeyVector& keys, const DiscreteKeys& dk,
+      const KeyVector& keys, const DiscreteKeys& discreteKeys,
       const std::vector<GaussianFactor::shared_ptr>& factors)
-      : discreteKeys_(dk), factors_(factors) {
-    // Compiler doesn't like `keys_` in the initializer list.
-    keys_ = keys;
-  }
+      : Base(keys, discreteKeys), factors_(factors) {}
 
   /// Discrete key selecting mixture component
   const DiscreteKeys& discreteKeys() const { return discreteKeys_; }
@@ -98,7 +95,7 @@ class DCGaussianMixtureFactor : public DCFactor {
     }
     std::cout << "; " << formatter(discreteKeys_.front().first) << " ]";
     std::cout << "{\n";
-    for (int i = 0; i < factors_.size(); i++) {
+    for (size_t i = 0; i < factors_.size(); i++) {
       auto t = boost::format("component %1%: ") % i;
       factors_[i]->print(t.str());
     }
--- a/gtsam/hybrid/DCGaussianMixtureFactor.h
+++ b/gtsam/hybrid/DCGaussianMixtureFactor.h
@@ -55,6 +55,8 @@ class DCGaussianMixtureFactor : public DCFactor {
       const std::vector<GaussianFactor::shared_ptr>& factors)
       : Base(keys, discreteKeys), factors_(factors) {}
 
+  DCGaussianMixtureFactor(const DCGaussianMixtureFactor& x) = default;
+
   /// Discrete key selecting mixture component
   const DiscreteKeys& discreteKeys() const { return discreteKeys_; }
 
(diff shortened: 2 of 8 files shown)


gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -58,6 +58,16 @@ const gtsam::DiscreteFactorGraph& HybridFactorGraph::discreteGraph() const {
   return discreteGraph_;
 }
 
+HybridFactorGraph HybridFactorGraph::linearize(
+    const Values& continuousValues) const {
+  auto gaussian_factor_graph = nonlinearGraph_.linearize(continuousValues);
+
+  HybridFactorGraph linearizedGraph(this->nonlinearGraph_,
```
dellaert, line 65: I’m assuming some code should go here still :-) checking for mixtures…

varunagrawal:
@ProfFan can you take a look at this? I made some new updates since Frank's previous changes made life easier for us.

gtsam/hybrid/DCFactor.h
```diff
@@ -101,10 +101,13 @@ class DCFactor : public gtsam::Factor {
    * @param discreteVals - Likewise, assignment to the discrete variables in
    * `discreteKeys__`.
    */
-  virtual boost::shared_ptr<gtsam::GaussianFactor> linearize(
+  virtual GaussianFactor::shared_ptr linearize(
       const gtsam::Values& continuousVals,
       const DiscreteValues& discreteVals) const = 0;
 
+  virtual DCFactor::shared_ptr linearize(
```
dellaert, line 108: Document

gtsam/hybrid/DCGaussianMixtureFactor.h
```diff
@@ -69,6 +67,28 @@ class DCGaussianMixtureFactor : public Factor {
...
+  }
+
+  // TODO(dellaert): Does not make sense
+  GaussianFactor::shared_ptr linearize(
+      const gtsam::Values& continuousVals,
+      const DiscreteValues& discreteVals) const override {
+    return factors_[0];
+  }
+
+  /// Return this linear factor as a DCFactor::shared_ptr
```
dellaert, line 83: Also does not make sense: factor is already linear.

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -144,6 +148,17 @@ class DCMixtureFactor : public DCFactor {
...
 
+  /// Linearize all the continuous factors to get a DCGaussianMixtureFactor.
+  DCFactor::shared_ptr linearize(const Values& continuousVals) const override {
+    std::vector<GaussianFactor::shared_ptr> linearized_factors;
+    for (size_t i = 0; i < factors_.size(); i++) {
+      auto linearized = factors_[i].linearize(continuousVals);
+      linearized_factors.push_back(linearized);
+    }
+    DCGaussianMixtureFactor linearized(keys_, dk_, linearized_factors);
+    return boost::make_shared<DCGaussianMixtureFactor>(linearized);
```
dellaert, line 159: Copy! Just pass arguments to make_shared… that’s the point of it :-)

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -22,23 +22,25 @@ void HybridFactorGraph::push_nonlinear(
...
 }
 
-void HybridFactorGraph::print(const std::string &str,
-                              const gtsam::KeyFormatter &keyFormatter) const {
+void HybridFactorGraph::print(const std::string& str,
+                              const gtsam::KeyFormatter& keyFormatter) const {
   std::string nonlinearStr = str + ": NonlinearFactorGraph";
   std::string discreteStr = str + ": DiscreteFactorGraph";
   std::string dcStr = str + ": DCFactorGraph";
+  std::string gStr = str + ": GaussianGraph";
```
dellaert, line 38: Get rid of these intermediate variables

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -58,13 +60,32 @@ const gtsam::DiscreteFactorGraph& HybridFactorGraph::discreteGraph() const {
...
+  DCFactorGraph linearized_DC_factors;
+  for (DCFactor::shared_ptr factor : dcGraph_) {
+    auto dcgf = factor->linearize(continuousValues);
+    linearized_DC_factors.push_back(dcgf);
+  }
+
+  // Construct new linearized HybridFactorGraph
+  HybridFactorGraph linearizedGraph(this->nonlinearGraph_, this->discreteGraph_,
+                                    linearized_DC_factors,
+                                    *gaussian_factor_graph);
```
dellaert, line 78: If the HFG had Gaussian factors, we will lose them here :-) add a check in test that Gaussian factors are left alone.

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -110,6 +188,8 @@ class HybridFactorGraph {
    */
   const gtsam::DiscreteFactorGraph& discreteGraph() const;
 
+  HybridFactorGraph linearize(const Values& continuousValues) const;
```
dellaert, line 191: Document

gtsam/hybrid/tests/testHybridFactorGraph.cpp
```diff
@@ -50,31 +50,36 @@ TEST(HybridFactorGraph, Switching) {
...
+  Values values;
+  for(size_t k=1; k <= K; k++){
+    values.insert<double>(X(k), 0.0);
+  }
+
   // TODO: create 4 linearization points.
 
   // Linearize here:
-  HybridFactorGraph dcmfg;  // TODO: = fg.linearize(values);
+  HybridFactorGraph dcmfg = fg.linearize(values);
```
dellaert, line 82: Actually test stuff :-)

dellaert (commented):
Cool, this is going the right way. I propose to add actual tests, and only then investigate whether we can use decisiontrees instead. I would recommend reading the API of DecisionTree, as it is very powerful.

gtsam/hybrid/DCGaussianMixtureFactor.h
varunagrawal: Yeah the issues is that without an implementation of that method, `DCGaussianMixtureFactor` is an abstract type so we cannot initialize it in DCMixtureFactor. :(

varunagrawal: Maybe we should have a DCFactor and a corresponding DCGaussianFactor?
Or we can just remove the error and linearize methods from the DCFactor class.

gtsam/hybrid/HybridFactorGraph.cpp
varunagrawal: Huh that's a good point. Do we want to append the new gaussian factors to the existing gaussian factor graph? As per this implementation, they will definitely be overwritten,

gtsam/hybrid/tests/testHybridFactorGraph.cpp
varunagrawal: Added a simple size assertion test.

gtsam/hybrid/DCGaussianMixtureFactor.h
dellaert: Just check its type using dynamic pointer cast. Or have a method `isLinear`.

gtsam/hybrid/HybridFactorGraph.cpp
dellaert: Yes, append.

gtsam/hybrid/tests/testHybridFactorGraph.cpp
dellaert: Yeah, there should be much more here.

gtsam/hybrid/DCGaussianMixtureFactor.h
```diff
@@ -69,6 +67,28 @@ class DCGaussianMixtureFactor : public Factor {
     return 0;
   }
 
+  // TODO(dellaert): Does not make sense
+  double error(const gtsam::Values& continuousVals,
+               const gtsam::DiscreteValues& discreteVals) const override {
+    return 0;
```
ProfFan, line 73: throw?

gtsam/hybrid/tests/testDCGaussianElimination.cpp
```diff
@@ -0,0 +1,235 @@
...
+    // return etree.eliminate(function);
+
+    auto bayesNet = boost::make_shared<DCGaussianBayesNet>();
+    auto factors = boost::make_shared<DCGaussianMixtureFactorGraph>();
+    return {bayesNet, factors};
+  }
+};
+
+/* TODO(dellaert) EliminateableFactorGraph does not play well with hybrid:
+/// Speical DCFactorGraph that can be eliminated partially
```
ProfFan, line 163: typo

dellaert (changes requested):
The test is inadequate. Please re-request review when you actually test what linearize is doing as a whole? I look at the test first to understand what is being accomplished, and at this moment I see only one gaussian factor being there. Is that really all?

gtsam/hybrid/HybridFactorGraph.cpp
varunagrawal: Done

gtsam/hybrid/DCGaussianMixtureFactor.h
varunagrawal: Would we be checking just for `DCGaussianMixtureFactor`? Asking since we want to keep the possibility of adding other `DCGaussianXFactor`s in the future I guess?

varunagrawal: Also, since DCGaussianMixtureFactor inherits from DCFactor, we kinda need this method no?

gtsam/hybrid/DCFactor.h
dellaert: Document better: how is this different from linearize above?

gtsam/hybrid/DCGaussianMixtureFactor.h
dellaert: Agreed: `throw std::runtime_error("DCGaussianMixtureFactor::error not implemented");`

gtsam/hybrid/DCGaussianMixtureFactor.h
```diff
@@ -69,6 +67,28 @@ class DCGaussianMixtureFactor : public Factor {
...
+  }
+
+  // TODO(dellaert): Does not make sense
+  GaussianFactor::shared_ptr linearize(
+      const gtsam::Values& continuousVals,
+      const DiscreteValues& discreteVals) const override {
+    return factors_[0];
```
dellaert, line 80: same: throw

gtsam/hybrid/DCGaussianMixtureFactor.h
```diff
@@ -69,6 +67,28 @@ class DCGaussianMixtureFactor : public Factor {
...
+  }
+
+  /// Return linearized version of this factor.
+  virtual DCFactor::shared_ptr linearize(
+      const Values& continuousVals) const override {
+    throw std::runtime_error("DCGaussianFactor is already linear");
+  }
+
+  // TODO(dellaert): implement
+  size_t dim() const override { return 0; };
```
dellaert, line 90: same

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -58,13 +57,49 @@ const gtsam::DiscreteFactorGraph& HybridFactorGraph::discreteGraph() const {
...
+
+HybridFactorGraph HybridFactorGraph::linearize(
+    const Values& continuousValues) const {
+  // linearize the continuous factors
+  auto gaussian_factor_graph = nonlinearGraph_.linearize(continuousValues);
+
+  // linearize the DCFactors
+  DCFactorGraph linearized_DC_factors;
+  for (DCFactor::shared_ptr factor : dcGraph_) {
+    // Check if the factor is a DCGaussianMixtureFactor since we don't linearize
```
dellaert, line 72: get comment on one line :-)

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -58,13 +57,49 @@ const gtsam::DiscreteFactorGraph& HybridFactorGraph::discreteGraph() const {
...
+    // those.
+    auto dcgm = dynamic_cast<const DCGaussianMixtureFactor*>(factor.get());
```
dellaert, line 74: I think you can do
`auto dcgm = dynamic_pointer_cast<const DCGaussianMixtureFactor>(factor);`

gtsam/hybrid/tests/testDCFactorGraph.cpp
```diff
@@ -219,13 +219,14 @@ TEST(DiscreteBayesTree, Switching) {
...
   GTSAM_PRINT(dcmfg);
 
-  // Eliminate partially.
-  Ordering ordering;
-  for (size_t k = 1; k <= K; k++) ordering += X(k);
-  using dc_traits = EliminationTraits<DCGaussianMixtureFactorGraph>;
-  auto result = dc_traits::eliminatePartialSequential(dcmfg, ordering);
-  GTSAM_PRINT(*result.first); // DCGaussianBayesNet
-  GTSAM_PRINT(*result.second); // DCGaussianMixtureFactorGraph
+  //TODO(Varun) uncomment and finish implementing
```
dellaert, line 222: I am doing that next

dellaert (approved):
Cool.  A few more comments but after that should be mergeable.

varunagrawal:
@dellaert we should all probably meet. I am a bit lost as to what the test should actually be testing since we don't have elimination integrated yet.

gtsam/hybrid/HybridFactorGraph.cpp
varunagrawal: Gotcha. I was trying to avoid an extra include.

varunagrawal: Hmm doesn't work. I guess this is because `factor` is a `boost::shared_ptr`?

varunagrawal: Yeah that is it. I see this in the error message: `note:   ‘gtsam::DCFactor’ is not derived from ‘const std::shared_ptr<_Tp>’`

gtsam/hybrid/tests/testDCFactorGraph.cpp
varunagrawal: Cool. I commented it out since it would get CI to pass and we can always uncomment it when requiired.

dellaert:
> @dellaert we should all probably meet. I am a bit lost as to what the test should actually be testing since we don't have elimination integrated yet.

I met with @ProfFan . Let's merge this PR then you guys both (together!) make a new PR for decision tree version? I'll be out all day (Wed 29) working on something else.

PR #983, opened 2021-12-23
HybridFactorGraph emplace variants

dellaert:
emplace variants, automatically deduce the right graph.

Unfortunately emplace variants don't deal well with {}, but it works with auto.

### DIFF
--- a/gtsam/hybrid/HybridFactorGraph.cpp
+++ b/gtsam/hybrid/HybridFactorGraph.cpp
@@ -35,10 +35,12 @@ void HybridFactorGraph::print(const std::string &str,
   std::string nonlinearStr = str + ": NonlinearFactorGraph";
   std::string discreteStr = str + ": DiscreteFactorGraph";
   std::string dcStr = str + ": DCFactorGraph";
+  std::string gStr = str + ": GaussianGraph";
 
   nonlinearGraph_.print(nonlinearStr, keyFormatter);
   discreteGraph_.print(discreteStr, keyFormatter);
   dcGraph_.print(dcStr, keyFormatter);
+  gaussianGraph_.print(gStr, keyFormatter);
 }
 
 gtsam::FastSet<gtsam::Key> HybridFactorGraph::keys() const {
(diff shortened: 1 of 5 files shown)


dellaert:
Review for @ProfFan , cc @varunagrawal

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -35,10 +35,12 @@ void HybridFactorGraph::print(const std::string &str,
   std::string nonlinearStr = str + ": NonlinearFactorGraph";
   std::string discreteStr = str + ": DiscreteFactorGraph";
   std::string dcStr = str + ": DCFactorGraph";
+  std::string gStr = str + ": GaussianGraph";
```
ProfFan, line 38: const

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -53,6 +53,57 @@ class HybridFactorGraph {
...
+
+  /// Construct a factor and add (shared pointer to it) to factor graph.
+  template <class FACTOR, class... Args>
+  IsGaussian<FACTOR> emplace_shared(Args&&... args) {
+    gaussianGraph_.push_back(boost::allocate_shared<FACTOR>(
+        Eigen::aligned_allocator<FACTOR>(), std::forward<Args>(args)...));
+  }
+
+  // TODO(dellaert): from below I think we should only keep shared pointer
+  // versions.
```
ProfFan, line 105: I agree, since I just looked at GFG and NFG and they all use `shared_ptr` internally.

ProfFan (approved):

varunagrawal:
I'll wait for this PR to merge before making any updates since I need some of the changes here. :slightly_smiling_face:

gtsam/hybrid/HybridFactorGraph.cpp
dellaert: not even sure why we have these intermediate variables. Feel free to move them in place after I merge.

dellaert:
merged before checks finish. I think the arg forwarding might fail on some platforms, but we can handle in the linearize PR

PR #986, opened 2021-12-24
Create markdown representation

dellaert:
This PR adds markdown formatting for discrete factors and conditionals. Main new functions are
- `DiscreteConditional::_repr_markdown_`
- `DecisionTreeFactor::_repr_markdown_`

These method names are not very c++ like but work in python notebooks as is. Waiting for a PR in wrap so we can rename them to "markdown", in analogy to "dot". @varunagrawal :-)

An example:

image

Key formatters are supported but need a bit more machinery on python side, e.g.,
```python
class MD:
    """Small class that renders markdown with optional arguments."""
    def __init__(self, obj, *args):
        self.md = obj._repr_markdown_(*args)
    def _repr_markdown_(self):
        return self.md
```


varunagrawal:
Should I only do `markdown` for now and add other methods later? Or is there some subset of functions that you would find immediately useful.

dellaert:
> Should I only do `markdown` for now and add other methods later? Or is there some subset of functions that you would find immediately useful.

It's just a loop over strings, right? Just add all supported formats, including mime.

gtsam/discrete/DiscreteConditional.cpp
```diff
@@ -147,10 +147,10 @@ void DiscreteConditional::solveInPlace(DiscreteValues* values) const {
     keys & dk;
   }
   // Get all Possible Configurations
-  vector<DiscreteValues> allPosbValues = cartesianProduct(keys);
+  const auto allPosbValues = cartesianProduct(keys);
 
   // Find the MPE
-  for(DiscreteValues& frontalVals: allPosbValues) {
+  for(const auto& frontalVals: allPosbValues) {
```
varunagrawal, line 153: Minor comment: `auto&&` will automatically take care of constness.

varunagrawal (approved):
Straightforward PR.

dellaert:
Cool, thanks for joint effort! I will merge after CI passes.

varunagrawal:
Woohoo

PR #989, opened 2021-12-25
dot argument order

dellaert:
Sorry for the churn, but I screwed up argument order in factor graph `dot` methods.
I switched formatter and writer arguments to be more inline with BayesNet etc.


varunagrawal (approved):
LGTM

PR #994, opened 2021-12-30
Use DecisionTree of factors instead of vector

varunagrawal:
Replace the use of `std::vector<FactorType>` for `DCMixtureFactor`s with `DecisionTree<Key, FactorType>`.

This should enable much easier ways for performing discrete key selection and overall elimination.

### DIFF
--- a/gtsam/discrete/AlgebraicDecisionTree.h
+++ b/gtsam/discrete/AlgebraicDecisionTree.h
@@ -30,6 +30,13 @@ namespace gtsam {
   template<typename L>
   class AlgebraicDecisionTree: public DecisionTree<L, double> {
 
+    /**
+     * @brief Default method for comparison of two doubles upto some tolerance.
+     */
+    static bool DefaultComparator(double a, double b, double tol) {
+      return std::abs(a - b) < tol;
+    }
+
   public:
 
     typedef DecisionTree<L, double> Super;
@@ -138,6 +145,12 @@ namespace gtsam {
       return this->combine(labelC, &Ring::add);
     }
 
+    /// Equality method customized to node type `double`.
+    bool equals(const AlgebraicDecisionTree& other, double tol = 1e-9,
+                const std::function<bool(double, double, double)>& comparator =
+                    &DefaultComparator) const {
+      return this->root_->equals(*other.root_, tol, comparator);
+    }
   };
 // AlgebraicDecisionTree
 
--- a/gtsam/discrete/AlgebraicDecisionTree.h
+++ b/gtsam/discrete/AlgebraicDecisionTree.h
@@ -30,14 +30,7 @@ namespace gtsam {
   template<typename L>
   class AlgebraicDecisionTree: public DecisionTree<L, double> {
 
-    /**
-     * @brief Default method for comparison of two doubles upto some tolerance.
-     */
-    static bool DefaultComparator(double a, double b, double tol) {
-      return std::abs(a - b) < tol;
-    }
-
-  public:
+   public:
 
     typedef DecisionTree<L, double> Super;
 
@@ -146,9 +139,11 @@ namespace gtsam {
     }
 
     /// Equality method customized to node type `double`.
-    bool equals(const AlgebraicDecisionTree& other, double tol = 1e-9,
-                const std::function<bool(double, double, double)>& comparator =
-                    &DefaultComparator) const {
+    bool equals(const AlgebraicDecisionTree& other, double tol = 1e-9) const {
+      // lambda for comparison of two doubles upto some tolerance.
+      auto comparator = [](double a, double b, double tol) {
+        return std::abs(a - b) < tol;
+      };
       return this->root_->equals(*other.root_, tol, comparator);
     }
   };
--- a/gtsam/discrete/AlgebraicDecisionTree.h
+++ b/gtsam/discrete/AlgebraicDecisionTree.h
@@ -141,10 +141,10 @@ namespace gtsam {
     /// Equality method customized to node type `double`.
     bool equals(const AlgebraicDecisionTree& other, double tol = 1e-9) const {
       // lambda for comparison of two doubles upto some tolerance.
-      auto comparator = [](double a, double b, double tol) {
+      auto compare = [tol](double a, double b) {
         return std::abs(a - b) < tol;
       };
-      return this->root_->equals(*other.root_, tol, comparator);
+      return this->root_->equals(*other.root_, tol, compare);
     }
   };
 // AlgebraicDecisionTree
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -458,6 +458,14 @@ namespace gtsam {
   }
 
   /*********************************************************************************/
+  template <typename L, typename Y>
+  template <typename X>
+  DecisionTree<L, Y>::DecisionTree(const DecisionTree<L, X>& other,
+                                   std::function<Y(const X&)> op) {
+    root_ = convert(other.root_, op);
+  }
+
+  /*********************************************************************************/
   // Called by two constructors above.
   // Takes a label and a corresponding range of decision trees, and creates a new
   // decision tree. However, the order of the labels needs to be respected, so we
@@ -600,6 +608,38 @@ namespace gtsam {
       functions += converted;
     }
     return LY::compose(functions.begin(), functions.end(), newLabel);
+  }
+
+  /*********************************************************************************/
+  template<typename L, typename Y>
+  template<typename X>
+  typename DecisionTree<L, Y>::NodePtr DecisionTree<L, Y>::convert(
+      const typename DecisionTree<L, X>::NodePtr& f,
+      std::function<Y(const X&)> op) {
+
+    typedef DecisionTree<L, X> LX;
+    typedef typename LX::Leaf LXLeaf;
+    typedef typename LX::Choice LXChoice;
+    typedef typename LX::NodePtr LXNodePtr;
+    typedef DecisionTree<L, Y> LY;
+
+    // ugliness below because apparently we can't have templated virtual functions
+    // If leaf, apply unary conversion "op" and create a unique leaf
+    const LXLeaf* leaf = dynamic_cast<const LXLeaf*> (f.get());
+    if (leaf) return NodePtr(new Leaf(op(leaf->constant())));
+
+    // Check if Choice
+    boost::shared_ptr<const LXChoice> choice = boost::dynamic_pointer_cast<const LXChoice> (f);
+    if (!choice) throw std::invalid_argument(
+        "DecisionTree::Convert: Invalid NodePtr");
+
+    // put together via Shannon expansion otherwise not sorted.
+    std::vector<LY> functions;
+    for(const LXNodePtr& branch: choice->branches()) {
+      LY converted(convert<X>(branch, op));
+      functions += converted;
+    }
+    return LY::compose(functions.begin(), functions.end(), choice->label());
   }
 
   /*********************************************************************************/
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -79,7 +79,7 @@ namespace gtsam {
     bool equals(const Node& q, double tol) const override {
       const Leaf* other = dynamic_cast<const Leaf*> (&q);
       if (!other) return false;
-      return std::abs(double(this->constant_ - other->constant_)) < tol;
+      return this->constant_ == other->constant_;
     }
 
     /** print */
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -108,7 +108,7 @@ namespace gtsam {
 
     /** print */
     void print(const std::string& s,
-               const std::function<std::string(L)> formatter) const override {
+               const std::function<std::string(L)>& formatter) const override {
       bool showZero = true;
       if (showZero || constant_) std::cout << s << " Leaf " << constant_ << std::endl;
     }
@@ -261,7 +261,8 @@ namespace gtsam {
     }
 
     /** print (as a tree) */
-    void print(const std::string& s, const std::function<std::string(L)> formatter) const override {
+    void print(const std::string& s,
+               const std::function<std::string(L)>& formatter) const override {
       std::cout << s << " Choice(";
       std::cout << formatter(label_) << ") " << std::endl;
       for (size_t i = 0; i < branches_.size(); i++)
@@ -675,7 +676,7 @@ namespace gtsam {
   template <typename L, typename Y>
   void DecisionTree<L, Y>::print(
       const std::string& s,
-      const std::function<std::string(L)> formatter) const {
+      const std::function<std::string(L)>& formatter) const {
     root_->print(s, formatter);
   }
 
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -19,7 +19,6 @@
 
 #pragma once
 
-#include <gtsam/base/Testable.h>
 #include <gtsam/discrete/DecisionTree.h>
 
 #include <boost/assign/std/vector.hpp>
@@ -78,8 +77,7 @@ namespace gtsam {
 
     /** equality up to tolerance */
     bool equals(const Node& q, double tol,
-                const std::function<bool(const Y&, const Y&, double)>&
-                    comparator) const override {
+                const ComparatorFunc& comparator) const override {
       const Leaf* other = dynamic_cast<const Leaf*>(&q);
       if (!other) return false;
       return comparator(this->constant_, other->constant_, tol);
@@ -87,7 +85,7 @@ namespace gtsam {
 
     /** print */
     void print(const std::string& s,
-               const std::function<std::string(L)>& formatter) const override {
+               const FormatterFunc& formatter) const override {
       bool showZero = true;
       if (showZero || constant_) std::cout << s << " Leaf " << constant_ << std::endl;
     }
@@ -241,7 +239,7 @@ namespace gtsam {
 
     /** print (as a tree) */
     void print(const std::string& s,
-               const std::function<std::string(L)>& formatter) const override {
+               const FormatterFunc& formatter) const override {
       std::cout << s << " Choice(";
       std::cout << formatter(label_) << ") " << std::endl;
       for (size_t i = 0; i < branches_.size(); i++)
@@ -284,8 +282,7 @@ namespace gtsam {
 
     /** equality up to tolerance */
     bool equals(const Node& q, double tol,
-                const std::function<bool(const Y&, const Y&, double)>&
-                    comparator) const override {
+                const ComparatorFunc& comparator) const override {
       const Choice* other = dynamic_cast<const Choice*>(&q);
       if (!other) return false;
       if (this->label_ != other->label_) return false;
@@ -651,16 +648,14 @@ namespace gtsam {
 
   /*********************************************************************************/
   template <typename L, typename Y>
-  bool DecisionTree<L, Y>::equals(
-      const DecisionTree& other, double tol,
-      const std::function<bool(const Y&, const Y&, double)>& comparator) const {
+  bool DecisionTree<L, Y>::equals(const DecisionTree& other, double tol,
+                                  const ComparatorFunc& comparator) const {
     return root_->equals(*other.root_, tol, comparator);
   }
 
   template <typename L, typename Y>
-  void DecisionTree<L, Y>::print(
-      const std::string& s,
-      const std::function<std::string(L)>& formatter) const {
+  void DecisionTree<L, Y>::print(const std::string& s,
+                                 const FormatterFunc& formatter) const {
     root_->print(s, formatter);
   }
 
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -77,7 +77,7 @@ namespace gtsam {
 
     /** equality up to tolerance */
     bool equals(const Node& q, double tol) const override {
-      const Leaf* other = dynamic_cast<const Leaf*> (&q);
+      const Leaf* other = dynamic_cast<const Leaf*>(&q);
       if (!other) return false;
       return this->constant_ == other->constant_;
     }
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -77,10 +77,10 @@ namespace gtsam {
 
     /** equality up to tolerance */
     bool equals(const Node& q, double tol,
-                const ComparatorFunc& comparator) const override {
+                const CompareFunc& compare) const override {
       const Leaf* other = dynamic_cast<const Leaf*>(&q);
       if (!other) return false;
-      return comparator(this->constant_, other->constant_, tol);
+      return compare(this->constant_, other->constant_);
     }
 
     /** print */
@@ -282,14 +282,14 @@ namespace gtsam {
 
     /** equality up to tolerance */
     bool equals(const Node& q, double tol,
-                const ComparatorFunc& comparator) const override {
+                const CompareFunc& compare) const override {
       const Choice* other = dynamic_cast<const Choice*>(&q);
       if (!other) return false;
       if (this->label_ != other->label_) return false;
       if (branches_.size() != other->branches_.size()) return false;
       // we don't care about shared pointers being equal here
       for (size_t i = 0; i < branches_.size(); i++)
-        if (!(branches_[i]->equals(*(other->branches_[i]), tol, comparator)))
+        if (!(branches_[i]->equals(*(other->branches_[i]), tol, compare)))
           return false;
       return true;
     }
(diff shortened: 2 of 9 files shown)


gtsam/discrete/DecisionTree-inl.h
```diff
@@ -75,15 +77,38 @@ namespace gtsam {
...
+    }
+    /// @}
+
     /** equality up to tolerance */
     bool equals(const Node& q, double tol) const override {
-      const Leaf* other = dynamic_cast<const Leaf*> (&q);
+      const Leaf* other = dynamic_cast<const Leaf*>(&q);
       if (!other) return false;
-      return std::abs(double(this->constant_ - other->constant_)) < tol;
+      return this->equals<Y>(this->constant_, other->constant_, tol);
```
dellaert, line 106: What happened to passing a compare method? I had advised to keep DT generic, independent of GTSAM :-)

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -236,12 +261,11 @@ namespace gtsam {
     }
 
     /** print (as a tree) */
-    void print(const std::string& s) const override {
+    void print(const std::string& s, const std::function<std::string(L)> formatter) const override {
```
dellaert, line 264: Reference? I know this file has them without, should check what is correct way…

gtsam/discrete/DecisionTree.h
```diff
@@ -79,7 +79,13 @@ namespace gtsam {
       const void* id() const { return this; }
 
       // everything else is virtual, no documentation here as internal
-      virtual void print(const std::string& s = "") const = 0;
+      virtual void print(
+          const std::string& s = "",
+          const std::function<std::string(L)> formatter = [](const L& x) {
```
dellaert, line 84: Hmm. Look at how Key does it? This clutters the method declaration…

gtsam/discrete/DecisionTree.h
```diff
@@ -113,14 +119,19 @@ namespace gtsam {
     convert(const typename DecisionTree<M, X>::NodePtr& f, const std::map<M,
         L>& map, std::function<Y(const X&)> op);
 
-    /** Default constructor */
-    DecisionTree();
+    /** Convert only node to a different type */
+    template <typename X>
+    NodePtr convert(const typename DecisionTree<L, X>::NodePtr& f,
```
dellaert, line 124: Where are these convert things used? I frown at the non-functional style, even though I probably did this myself. Do we really need it? If not, we can deprecate them…

gtsam/discrete/DecisionTree.h
```diff
@@ -149,12 +160,23 @@ namespace gtsam {
...
+
     /// @}
     /// @name Testable
     /// @{
 
     /** GTSAM-style print */
-    void print(const std::string& s = "DecisionTree") const;
+    void print(
+        const std::string& s = "DecisionTree",
+        const std::function<std::string(L)> formatter = [](const L& x) {
```
dellaert, line 175: Clutter. Copy/paste clutter?

gtsam/discrete/DecisionTree.h
```diff
@@ -219,12 +241,19 @@ namespace gtsam {
 
   /** free versions of apply */
 
+  //TODO(Varun) where are these templates Y, L and not L, Y?
```
dellaert, line 244: No reason. Pls switch :-)

gtsam/discrete/tests/testAlgebraicDecisionTree.cpp
```diff
@@ -48,7 +48,7 @@ template<> struct traits<ADT> : public Testable<ADT> {};
 #define DISABLE_DOT
 
 template<typename T>
-void dot(const T&f, const string& filename) {
+void DOT(const T&f, const string& filename) {
```
dellaert, line 51: No, no need to do this. Fully qualify other by doing `gtsam::dot`

gtsam/discrete/tests/testDecisionTree.cpp
```diff
@@ -32,13 +32,13 @@ using namespace std;
 using namespace gtsam;
 
 template<typename T>
-void dot(const T&f, const string& filename) {
+void write_dot(const T&f, const string& filename) {
```
dellaert, line 35: Same. Pls revert if my fix works.

gtsam/discrete/tests/testDecisionTree.cpp
```diff
@@ -179,8 +179,8 @@ TEST(DT, example)
 enum Label {
   U, V, X, Y, Z
 };
-typedef DecisionTree<Label, bool> BDT;
-bool convert(const int& y) {
+typedef DecisionTree<Label, int> BDT;
```
dellaert, line 182: Why?? Messing with tests that pass.

gtsam/hybrid/DCGaussianMixtureFactor.h
```diff
@@ -41,19 +40,26 @@ namespace gtsam {
...
- private:
-  std::vector<GaussianFactor::shared_ptr> factors_;
-
  public:
   using Base = DCFactor;
   using shared_ptr = boost::shared_ptr<DCGaussianMixtureFactor>;
+  using FactorDecisionTree = DecisionTree<Key, GaussianFactor::shared_ptr>;
+
+ private:
+  FactorDecisionTree factors_;
```
dellaert, line 49: Document

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -34,56 +33,79 @@ namespace gtsam {
...
                   const std::vector<NonlinearFactorType>& factors,
                   bool normalized = false)
-      : dk_(dk), factors_(factors), normalized_(normalized) {
-    // Compiler doesn't like `keys_` in the initializer list.
-    keys_ = keys;
-
-    // Add `dk` to `dkeys` list.
-    discreteKeys_.push_back(dk);
+      : Base(keys, discreteKeys), normalized_(normalized) {
+    std::vector<boost::shared_ptr<NonlinearFactorType>> factor_pointers;
```
dellaert, line 73: Why was it so much easier for Gaussian mixture factor? I am probably not seeing something. Also, explain this “1-1’ mapping. Look at Signature for an example.

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -34,56 +33,79 @@ namespace gtsam {
...
     this->factors_ = rhs.factors_;
   }
 
   ~DCMixtureFactor() = default;
 
   double error(const Values& continuousVals,
                const DiscreteValues& discreteVals) const override {
-    // Retrieve the assignment to our discrete key.
-    const size_t assignment = discreteVals.at(dk_.first);
+    /********************************************************/
```
dellaert, line 92: Huh?

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -34,56 +33,79 @@ namespace gtsam {
...
+           this->nonlinearFactorLogNormalizingConstant(*factor, continuousVals);
   }
 
   size_t dim() const override {
     // TODO(kevin) Need to modify this? Maybe we take discrete vals as parameter
     // and DCContinuousFactor will pass this in as needed.
-    return (factors_.size() > 0) ? factors_[0].dim() : 0;
+    // return (factors_.size() > 0) ? factors_[0].dim() : 0;
+    // TODO (Varun) find the best way to implement this
+    return 0;
```
dellaert, line 108: Throw? Who’s asking?

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -140,22 +155,25 @@ class DCMixtureFactor : public DCFactor {
...
-    return boost::make_shared<DCGaussianMixtureFactor>(keys_, dk_, linearized_factors);
+    std::function<GaussianFactor::shared_ptr(
+        const boost::shared_ptr<NonlinearFactorType>&)>
+        linearizeDT =
+            [continuousVals](
+                const boost::shared_ptr<NonlinearFactorType>& factor) {
+              return factor->linearize(continuousVals);
+            };
+
+    DecisionTree<Key, GaussianFactor::shared_ptr> linearized_factors(
```
dellaert, line 172: Nice!

gtsam/hybrid/tests/testDCMixtureFactor.cpp
```diff
@@ -21,48 +21,72 @@ using namespace gtsam;
...
-/*
- * Test DCDiscreteFactor using a simple mixture.
- *
- * Construct a single factor (for a single variable) consisting of a
- * discrete-conditional mixture. Here we have a "null hypothesis" consisting of
- * a Gaussian with large variance and an "alternative hypothesis" consisting of
- * a Gaussian with smaller variance.
- */
-TEST(TestSuite, dcdiscrete_mixture) {
+// /*
```
dellaert, line 24: Commented out? Decide if in this PR or not :-) if not, leave it out.

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -19,22 +19,24 @@
 
 #pragma once
 
-#include <gtsam/discrete/DecisionTree.h>
 #include <gtsam/base/Testable.h>
+#include <gtsam/base/VectorSpace.h>
```
dellaert, line 23: Why? Could we make DecisionTree and ADT completely independent of GTSAM? They are nearly so: an elegant, minimal data structure. Maybe even change the namespace. (We are in 4.2 now).

dellaert (commented):
Coool!! Many comments, and I also don’t understand convert vs apply. Have some philosophical issues with making DT and ADT more gtsam interwoven as well. And convert is no -functional. Why did I do that? Did I?

gtsam/discrete/DecisionTree-inl.h
varunagrawal: Yeah I had a question about that and just wanted something working for now. This `equals` is in the `Leaf` class, so we would have to add the comparison method for all node types. Is that intended?

gtsam/discrete/DecisionTree.h
varunagrawal: Added a `defaultFormatter` method in the protected namespace of the class.

gtsam/discrete/DecisionTree.h
varunagrawal: I don't think they are used anywhere other than in the constructor for converting to another type. We can deprecate them but we will still need a mechanism to convert the nodes from one type to another (for the `linearize` method atleast).

gtsam/hybrid/DCMixtureFactor.h
varunagrawal: Oh this was made to demarcate the old version vs the new version so Fan and I knew what were we doing. Let me remove it.

gtsam/discrete/tests/testAlgebraicDecisionTree.cpp
varunagrawal: That doesn't work since the compiler thinks that this `dot` is the Vector dot. Though I may be able to resolve this once the equality method is figured out.

gtsam/discrete/tests/testDecisionTree.cpp
varunagrawal: Ah sorry about that. Reverting.

gtsam/discrete/DecisionTree-inl.h
varunagrawal: Yup, this is because of SFINAE and using the `traits::Equals`. It was also causing the `dot` collision, so I am more than glad to get rid of this.

gtsam/hybrid/DCMixtureFactor.h
varunagrawal: Because in `DCGaussianMixtureFactor` we are passing in a vector of shared pointers, so no need to convert to shared pointers like we're doing here.
The 1-to-1 is a misnomer and a mistake. I'll update it.

varunagrawal:
@dellaert just need a decision on whether I should add a compare method to all the Node classes (Leaf, Choice, DecisionTree). My gut says yes, but I want to ensure we're on the same page before I make major changes.

dellaert:
> @dellaert just need a decision on whether I should add a compare method to all the Node classes (Leaf, Choice, DecisionTree). My gut says yes, but I want to ensure we're on the same page before I make major changes.

No, you would not add a compare *method*. I am suggesting adding a compare *argument* to equals, so derived classes can pass that in.

varunagrawal:
Then what would be the default implementation? It cannot go before `tol` since that violates the `Testable` trait, and it needs a default to go after.
If I do something like `a.equals(b, tol)`, this throws a compiler error for doubles and ints, and `a==b` throws an error for other types. :(

varunagrawal:
Also, when I said compare method, I actually meant a compare _member_.

varunagrawal:
Okay figured it out. :)

gtsam/discrete/DecisionTree-inl.h
varunagrawal: Resolved :D

gtsam/hybrid/tests/testDCMixtureFactor.cpp
varunagrawal: This is the last thing that needs to be done.

gtsam/discrete/tests/testAlgebraicDecisionTree.cpp
varunagrawal: Resolved

gtsam/discrete/AlgebraicDecisionTree.h
```diff
@@ -138,6 +145,12 @@ namespace gtsam {
       return this->combine(labelC, &Ring::add);
     }
 
+    /// Equality method customized to node type `double`.
+    bool equals(const AlgebraicDecisionTree& other, double tol = 1e-9,
+                const std::function<bool(double, double, double)>& comparator =
```
dellaert, line 150: I think function should not be an argument here, just tolerance.

gtsam/discrete/AlgebraicDecisionTree.h
```diff
@@ -30,6 +30,13 @@ namespace gtsam {
   template<typename L>
   class AlgebraicDecisionTree: public DecisionTree<L, double> {
 
+    /**
+     * @brief Default method for comparison of two doubles upto some tolerance.
+     */
+    static bool DefaultComparator(double a, double b, double tol) {
```
dellaert, line 36: just make a lambda in equals that uses tol

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -19,22 +19,23 @@
 
 #pragma once
 
-#include <gtsam/discrete/DecisionTree.h>
 #include <gtsam/base/Testable.h>
+#include <gtsam/discrete/DecisionTree.h>
```
dellaert, line 23: why Testable still?

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -76,14 +77,17 @@ namespace gtsam {
     }
 
     /** equality up to tolerance */
-    bool equals(const Node& q, double tol) const override {
-      const Leaf* other = dynamic_cast<const Leaf*> (&q);
+    bool equals(const Node& q, double tol,
+                const std::function<bool(const Y&, const Y&, double)>&
```
dellaert, line 81: use a typedef?

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -76,14 +77,17 @@ namespace gtsam {
...
+      const Leaf* other = dynamic_cast<const Leaf*>(&q);
       if (!other) return false;
-      return std::abs(double(this->constant_ - other->constant_)) < tol;
+      return comparator(this->constant_, other->constant_, tol);
     }
 
     /** print */
-    void print(const std::string& s) const override {
+    void print(const std::string& s,
+               const std::function<std::string(L)>& formatter) const override {
```
dellaert, line 90: use a typedef?

gtsam/discrete/DecisionTree.h
dellaert: Linearize is functional, though.  I don't understand why we are adding a non-const method. You say constructor but I don't see a new constructor...

gtsam/discrete/tests/testDecisionTree.cpp
```diff
@@ -40,7 +40,19 @@ void dot(const T&f, const string& filename) {
 
 #define DOT(x)(dot(x,#x))
 
-struct Crazy { int a; double b; };
+struct Crazy {
+  int a;
+  double b;
+
+  bool equals(const Crazy& other, double tol = 1e-12) const {
```
dellaert, line 47: why?

gtsam/discrete/tests/testDecisionTree.cpp
```diff
@@ -40,7 +40,19 @@ void dot(const T&f, const string& filename) {
...
+    return a == other.a && std::abs(b - other.b) < tol;
+  }
+
+  bool operator==(const Crazy& other) const {
```
dellaert, line 51: why?

gtsam/hybrid/DCGaussianMixtureFactor.h
```diff
@@ -41,20 +40,46 @@ namespace gtsam {
...
- private:
-  std::vector<GaussianFactor::shared_ptr> factors_;
-
  public:
   using Base = DCFactor;
   using shared_ptr = boost::shared_ptr<DCGaussianMixtureFactor>;
+  using FactorDecisionTree = DecisionTree<Key, GaussianFactor::shared_ptr>;
+
+ private:
+  /// Decision tree of Gaussian factors indexed by a Key.
```
dellaert, line 49: by discrete keys

gtsam/hybrid/DCGaussianMixtureFactor.h
```diff
@@ -41,20 +40,46 @@ namespace gtsam {
...
+   *
+   * @param keys Vector of keys for continuous factors.
+   * @param discreteKeys Vector of discrete assignments.
+   * @param factors Vector of gaussian factor shared pointers.
+   */
   DCGaussianMixtureFactor(
       const KeyVector& keys, const DiscreteKeys& discreteKeys,
       const std::vector<GaussianFactor::shared_ptr>& factors)
-      : Base(keys, discreteKeys), factors_(factors) {}
+      : Base(keys, discreteKeys),
```
dellaert, line 79: delegate to other constructor.

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -34,56 +33,78 @@ namespace gtsam {
...
- private:
-  DiscreteKey dk_;
-  std::vector<NonlinearFactorType> factors_;
-  bool normalized_;
-
  public:
   using Base = DCFactor;
   using shared_ptr = boost::shared_ptr<DCMixtureFactor>;
 
+  using FactorDecisionTree =
```
dellaert, line 40: document

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -34,56 +33,78 @@ namespace gtsam {
...
+      DecisionTree<Key, boost::shared_ptr<NonlinearFactorType>>;
+
+ private:
+  FactorDecisionTree factors_;
```
dellaert, line 44: document with `///< blah` at least

gtsam/hybrid/DCMixtureFactor.h
dellaert: I think this should not be the API. `factors` should be a FactorGraph<NonlinearFactorType>. This conversion costs money, and the rule in GTSAM is: every factor passed around is a shared pointer.

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -116,15 +137,13 @@ class DCMixtureFactor : public DCFactor {
...
-    // component factors in `factors_`.
-    if (factors_.size() != f.factors_.size()) return false;
-
-    // If the number of factors is the same, we compare them individually (they
-    // should be in the same order!). If any fail to match, return false.
-    for (size_t i = 0; i < factors_.size(); i++) {
-      if (!factors_[i].equals(f.factors_[i])) return false;
-    }
+    // Ensure that this DCMixtureFactor and `f` have the same `factors_`.
+    auto comparator = [](const boost::shared_ptr<NonlinearFactorType>& a,
```
dellaert, line 141: nice! *but* check in Testable. There is actually a shared pointer version of Equals called equals_star. Maybe it's just:
```
    if (!factors_.equals(f.factors_, 1e-9, equals_star<NonlinearFactorType>)) return false;
```

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -140,22 +159,25 @@ class DCMixtureFactor : public DCFactor {
...
 
   /// Linearize all the continuous factors to get a DCGaussianMixtureFactor.
   DCFactor::shared_ptr linearize(const Values& continuousVals) const override {
-    std::vector<GaussianFactor::shared_ptr> linearized_factors;
-    for (size_t i = 0; i < factors_.size(); i++) {
-      auto linearized = factors_[i].linearize(continuousVals);
-      linearized_factors.push_back(linearized);
-    }
-    return boost::make_shared<DCGaussianMixtureFactor>(keys_, dk_, linearized_factors);
+    std::function<GaussianFactor::shared_ptr(
```
dellaert, line 168: auto?

dellaert (commented):
Wow, much better! Still some nits/questions. Again, biggest question is: why a non-const `convert`?

gtsam/discrete/DecisionTree.h
varunagrawal: It's in the constructor definitions in `DecisionTree-inl.h`.

gtsam/discrete/tests/testDecisionTree.cpp
varunagrawal: This is needed so that the compiler doesn't complain about the default comparator method. Technically we don't need this, just the == operator overload.

gtsam/discrete/tests/testDecisionTree.cpp
varunagrawal: Same as above, needed for compiling with the default comparator.

gtsam/hybrid/DCMixtureFactor.h
varunagrawal: Gotcha, but this should probably go in the next PR since the vector API is from the original DC-SAM.

gtsam/discrete/DecisionTree.h
dellaert: I see. As long as the converts are private, all is good. If they are not, can you move the declarations to the private block at the end?

gtsam/hybrid/DCMixtureFactor.h
varunagrawal: I though about that, but the comparator requires 3 arguments `(a, b, tol)` whereas `equals_star` takes two.
`tol` is passed in as the constructor argument to `equals_star`.

gtsam/hybrid/DCMixtureFactor.h
varunagrawal: Sadly, I already tried using `auto` but the compiler is having some problems with that. :(

gtsam/discrete/tests/testDecisionTree.cpp
dellaert: I think this points to an issue, same as equals_star issue: the compare function we pass in should not have a tol! That is a detail for the derived classes. For a simple type like this, the default compare should just be ==. If you want a tol, you need to construct a lambda, where tol is passed in as context.

gtsam/discrete/DecisionTree.h
varunagrawal: Yup, they are in `protected` so we should be good. :)

varunagrawal:
Added the test and this is ready for full review!

gtsam/discrete/AlgebraicDecisionTree.h
```diff
@@ -138,6 +138,14 @@ namespace gtsam {
       return this->combine(labelC, &Ring::add);
     }
 
+    /// Equality method customized to node type `double`.
+    bool equals(const AlgebraicDecisionTree& other, double tol = 1e-9) const {
+      // lambda for comparison of two doubles upto some tolerance.
+      auto comparator = [](double a, double b, double tol) {
```
dellaert, line 144: Sorry, but this still does not make sense to me. I think the root issue is that in `DecisionTree.*` we have both tol and compare functions as arguments. I think it should be *only* the compare function, which should *not* have a tol argument. Instead, we should have here:
```
      auto compare = [tol](double a, double b) { return std::abs(a - b) < tol; };
      return this->root_->equals(*other.root_, compare);
```

Also, comparator -> compare. Former is weird.

dellaert (commented):
Sorry that this is such an iterative process, but I think equals in DecisionTree.* is still not there. It should not be "up to a tolerance" anymore, but just take a *binary* compare function. Also, please rename everything from "comparator" to "compare".

gtsam/discrete/AlgebraicDecisionTree.h
varunagrawal: No worries about the iteration. We want this to be kickass! :slightly_smiling_face: 

Also that snippet helps, thank you!

varunagrawal: Also does this mean I should remove `tol` from the `equals` method and update the traits declarations accordingly?

dellaert: We removed Testable from DecisionTree, so not needed there.
For ADT, keep tol and traits if they exist.

dellaert (approved):
AwesomeSauce !!!!

gtsam/discrete/AlgebraicDecisionTree.h
varunagrawal: Ha guess not :smile:

PR #996, opened 2021-12-30
Two small fixes

dellaert:
Two examples did not compile. Don't know why CI did not catch this before.


varunagrawal (approved):
LGTM, though I should look into why the CI didn't catch them.

Issue #995, opened 2021-12-30
Similarity(3) estimation returns a negative scale for some poorly aligned pose graphs


johnwlambert:
For some badly aligned pose graphs, `Similarity3.Align(` returns an object with **negative scale**, e.g. 

```python
Sim(3) estimate: 
R: [
	-0.786299, 0.313693, -0.532288;
	-0.569328, -0.0332175, 0.821439;
	0.239998, 0.948943, 0.204713
]
translation = [786.49152258 275.49551117 122.02794014]
scale =  -0.0009678177616769815
```
This wreaks havoc on reprojection errors, as cheirality exceptions begin to occur everywhere when using camera poses that have been transformed by the Sim(3) object with negative scale.

When the scale is clamped to a positive number, then reprojection errors are unmodified by applying SIm(3) to points and cameras.

According to Ethan Eade (see Section 10.1 of [this doc](https://ethaneade.com/lie_groups.pdf)), Sim(3) scale must be non-negative, so the GTSAM code is currently buggy.

## How can I reproduce it?
Call:
```python
    ab_pairs = Pose3Pairs(list(zip(aTi_list, bTi_list)))
    aSb = Similarity3.Align(ab_pairs)
```
on the following 2 pose graphs (using 17 out of 32 cameras for correspondences):
```python
        poses_gt = [
            Pose3(
                Rot3(
                    [
                        [0.696305769, -0.0106830792, -0.717665705],
                        [0.00546412488, 0.999939148, -0.00958346857],
                        [0.717724415, 0.00275160848, 0.696321772],
                    ]
                ),
                Point3(5.83077801, -0.94815149, 0.397751679),
            ),
            Pose3(
                Rot3(
                    [
                        [0.692272397, -0.00529704529, -0.721616549],
                        [0.00634689669, 0.999979075, -0.00125157022],
                        [0.721608079, -0.0037136016, 0.692291531],
                    ]
                ),
                Point3(5.03853323, -0.97547405, -0.348177392),
            ),
            Pose3(
                Rot3(
                    [
                        [0.945991981, -0.00633548292, -0.324128225],
                        [0.00450436485, 0.999969379, -0.00639931046],
                        [0.324158843, 0.00459370582, 0.945991552],
                    ]
                ),
                Point3(4.13186176, -0.956364218, -0.796029527),
            ),
            Pose3(
                Rot3(
                    [
                        [0.999553623, -0.00346470207, -0.0296740626],
[...]

PR #997, opened 2021-12-30
Force nonnegative scale for Sim(3)

johnwlambert:
As discussed in https://github.com/borglab/gtsam/issues/995, for certain inputs, `Similarity3.Align()` returns an object with negative scale.

According to [my derivation here](https://johnwlambert.github.io/icp/#sim3-align-derivation), based on [Zinnser05prip](http://www5.informatik.uni-erlangen.de/Forschung/Publikationen/2005/Zinsser05-PSR.pdf), the correct scale should not need an absolute value cast.

However, this PR adds an absolute value cast on the scale, and we can likely blame numerical instabilities as a reasonable motivation (the negative values I saw were relatively close to zero).


dellaert (approved):
Cool! Thanks John!

PR #998, opened 2021-12-31
Moved empty to Factor base class.

dellaert:
Because it was inconsistently defined, and this one (from Symbolic) seems to works for all Factor types.


dellaert:
PS needed for DC elimination, @varunagrawal ...

varunagrawal (approved):
Very cool. Simplicity FTW

PR #1000, opened 2022-01-02
DecisionTree refactor

dellaert:
For hybrid elimination, we need DecisionTree to be more powerful. 
- Previously there was a restriction to values that could be printed using <<. This PR adds label formatters (done previously in a branch by Varun) and value formatters (by me, here). 
- In addition, equals no longer takes a tolerance, but a compare function.
- DT is no longer Testable, but AlgebraicDecisionTree is.


gtsam/discrete/DecisionTree-inl.h
```diff
@@ -660,26 +694,34 @@ namespace gtsam {
...
+    dot(os, labelFormatter, valueFormatter, showZero);
     int result = system(
         ("dot -Tpdf " + name + ".dot -o " + name + ".pdf >& /dev/null").c_str());
     if (result==-1) throw std::runtime_error("DecisionTree::dot system call failed");
   }
 
-  template<typename L, typename Y>
-  std::string DecisionTree<L, Y>::dot(bool showZero) const {
+  template <typename L, typename Y>
+  std::string DecisionTree<L, Y>::dot(const LabelFormatter& labelFormatter,
```
varunagrawal, line 720: Three versions of `dot` so definitely need docstrings here.

gtsam/discrete/tests/testDecisionTree.cpp
```diff
@@ -40,32 +40,88 @@ void dot(const T&f, const string& filename) {
...
 #define DOT(x)(dot(x,#x))
 
-struct Crazy { int a; double b; };
-typedef DecisionTree<string,Crazy> CrazyDecisionTree; // check that DecisionTree is actually generic (as it pretends to be)
+struct Crazy {
+  int a;
+  double b;
+};
+
+//   bool equals(const Crazy& other, double tol = 1e-12) const {
```
varunagrawal, line 48: Kill these since you've removed Testable?

gtsam/discrete/AlgebraicDecisionTree.h
```diff
@@ -28,9 +28,15 @@ namespace gtsam {
    * TODO: consider eliminating this class altogether?
    */
   template<typename L>
-  class AlgebraicDecisionTree: public DecisionTree<L, double> {
+  class GTSAM_EXPORT AlgebraicDecisionTree: public DecisionTree<L, double> {
+    /// Default method used by `formatter` when printing.
+    static std::string DefaultFormatter(const L& x) {
```
varunagrawal, line 33: We can have this and the `valueFormatter` in the DecisionTreeFactor be the default methods?

varunagrawal (approved):
LGTM, though hold up on merging since I'll add docstrings and other stuff addressing my comments in a separate PR targeting this PR.

gtsam/discrete/DecisionTree.h
```diff
@@ -108,19 +122,20 @@ namespace gtsam {
...
 
-    /** Convert to a different type */
-    template<typename M, typename X> NodePtr
-    convert(const typename DecisionTree<M, X>::NodePtr& f, const std::map<M,
-        L>& map, std::function<Y(const X&)> op);
+    /// Convert from a DecisionTree<M, X>.
+    template <typename M, typename X>
+    NodePtr convertFrom(const typename DecisionTree<M, X>::NodePtr& f,
+                        std::function<L(const M&)> L_of_M,
+                        std::function<Y(const X&)> Y_of_X) const;
```
varunagrawal, line 129: Also, does it make sense to name these `M_to_L` and `X_to_Y`? That way it reads `functor to convert from M to L`. I can make this change if you agree.

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -625,13 +649,23 @@ namespace gtsam {
 
   template<typename L, typename Y>
   DecisionTree<L, Y> DecisionTree<L, Y>::apply(const Unary& op) const {
+    // It is unclear what should happen if tree is empty:
+    if (empty()) {
+      throw std::runtime_error(
+          "DecisionTree::apply(unary op) undefined for empty tree.");
+    }
```
varunagrawal, line 656: If the tree is empty, should this just be a no-op, which means we can return `this`?

dellaert: No. It is not allowed. *this would be wrong type.

gtsam/discrete/DecisionTree.h
dellaert: No, I really dislike "to" names, as you get

```
to_value = From2To(from_value)
```

instead of the natural flow

```
to_value = ToOfFrom(from_value)
```

That's why matrix convention is the way it is

PR #1002, opened 2022-01-03
DecisionTree Improvements

varunagrawal:
~~1. Moved DefaultFormatter to base class in ADT.~~
2. Replace `Super` with `Base` in ADT.
3. Use `using` instead of `typedef`.
4. Improved docs.

@dellaert this should address all my review comments in #1000.


dellaert (approved):
Cool, thanks ! For future reference:
- node type -> value type (only relevant to leaf nodes)
- node label -> label type (only relevant to choice nodes)

But I will change that.

PR #1004, opened 2022-01-03
Fix documentation on ProjectionFactorPPPC

dwisth:
A couple of small changes to `ProjectionFactorPPPC`:

- Remove excess constructor, use default arguments instead
- Fix construction documentation
- Make some variables `const` where possible


varunagrawal (approved):
LGTM

PR #1005, opened 2022-01-04
More DecisionTree improvements

dellaert:
- Added some depth-first methods: fold, visit, visitWith
- Use template for functions rather than std::functions, which matches against lambdas and function pointers as well.

### DIFF
--- a/gtsam/navigation/ImuBias.h
+++ b/gtsam/navigation/ImuBias.h
@@ -142,10 +142,10 @@ public:
     return q - (*this);
   }
   Vector6 GTSAM_DEPRECATED localCoordinates(const ConstantBias& q) {
-    return between(q).vector();
+    return (q - (*this)).vector();
   }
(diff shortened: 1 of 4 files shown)


gtsam/navigation/ImuBias.h
```diff
@@ -142,10 +142,10 @@ class GTSAM_EXPORT ConstantBias {
     return q - (*this);
   }
   Vector6 GTSAM_DEPRECATED localCoordinates(const ConstantBias& q) {
-    return between(q).vector();
+    return (q - (*this)).vector();
```
ProfFan, line 145: Is this inadvertently added?

ProfFan (approved):
Approving sans one comment.

gtsam/navigation/ImuBias.h
varunagrawal: This is fine. It is just addressing the GTSAM deprecation in ImuBias.h.

varunagrawal (approved):
LGTM

PR #1010, opened 2022-01-05
Subgraph API change

dellaert:
This is the first PR from some refactoring I did over the break, to get rid of shared pointers for *containers*. It is totally appropriate that we use shared pointers for factors (although, maybe we should not!), but it never made sense (I see now) that we used shared pointers for graphs, as they are not really shuffled around like factors.

The big changes that are coming (branches have not been published yet) are mostly in linear, to get a newer and faster API for elimination -- prompted by development in hybrid branch. This PR removes shared pointers from the subgraph API, for starters.

Adding @jlblancoc as a reviewer, partly to get his opinion on above, and @ProfFan as FYI as I know he has a forked branch with lots of shared pointer changes (presumably moving to std::shared_ptr).

### DIFF
--- a/gtsam/linear/SubgraphBuilder.cpp
+++ b/gtsam/linear/SubgraphBuilder.cpp
@@ -446,30 +446,29 @@ SubgraphBuilder::Weights SubgraphBuilder::weights(
 }
 
 /*****************************************************************************/
-GaussianFactorGraph::shared_ptr buildFactorSubgraph(
-    const GaussianFactorGraph &gfg, const Subgraph &subgraph,
-    const bool clone) {
-  auto subgraphFactors = boost::make_shared<GaussianFactorGraph>();
-  subgraphFactors->reserve(subgraph.size());
+GaussianFactorGraph buildFactorSubgraph(const GaussianFactorGraph &gfg,
+                                        const Subgraph &subgraph,
+                                        const bool clone) {
+  GaussianFactorGraph subgraphFactors;
+  subgraphFactors.reserve(subgraph.size());
   for (const auto &e : subgraph) {
     const auto factor = gfg[e.index];
-    subgraphFactors->push_back(clone ? factor->clone() : factor);
+    subgraphFactors.push_back(clone ? factor->clone() : factor);
   }
   return subgraphFactors;
 }
 
 /**************************************************************************************************/
-std::pair<GaussianFactorGraph::shared_ptr, GaussianFactorGraph::shared_ptr>  //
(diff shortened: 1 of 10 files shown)


dellaert:
Some love?

gtsam/linear/SubgraphBuilder.cpp
```diff
@@ -446,30 +446,29 @@ SubgraphBuilder::Weights SubgraphBuilder::weights(
...
-GaussianFactorGraph::shared_ptr buildFactorSubgraph(
-    const GaussianFactorGraph &gfg, const Subgraph &subgraph,
-    const bool clone) {
-  auto subgraphFactors = boost::make_shared<GaussianFactorGraph>();
-  subgraphFactors->reserve(subgraph.size());
+GaussianFactorGraph buildFactorSubgraph(const GaussianFactorGraph &gfg,
+                                        const Subgraph &subgraph,
+                                        const bool clone) {
+  GaussianFactorGraph subgraphFactors;
+  subgraphFactors.reserve(subgraph.size());
```
varunagrawal, line 453: What's the point of doing a `reserve` if `push_back` below will allocate memory for us?

gtsam/linear/linear.i
```diff
@@ -625,7 +625,7 @@ virtual class SubgraphSolverParameters : gtsam::ConjugateGradientParameters {
 
 virtual class SubgraphSolver  {
   SubgraphSolver(const gtsam::GaussianFactorGraph &A, const gtsam::SubgraphSolverParameters &parameters, const gtsam::Ordering& ordering);
-  SubgraphSolver(const gtsam::GaussianFactorGraph &Ab1, const gtsam::GaussianFactorGraph* Ab2, const gtsam::SubgraphSolverParameters &parameters, const gtsam::Ordering& ordering);
+  SubgraphSolver(const gtsam::GaussianFactorGraph &Ab1, const gtsam::GaussianFactorGraph& Ab2, const gtsam::SubgraphSolverParameters &parameters, const gtsam::Ordering& ordering);
```
varunagrawal, line 628: Nit: formatting

varunagrawal (approved):
Looks good to me modulo 2 comments. Would like to hear @jlblancoc's thoughts as well.

gtsam/linear/SubgraphBuilder.cpp
dellaert: Reserve is recommended if you know exactly how many things you will push

gtsam/linear/linear.i
dellaert: Global replace…

PR #1012, opened 2022-01-06
Fix separator keys for hybrid elimination

varunagrawal:
This PR serves as both a fix and a place to do a design review before merging the changes.

It addresses the problem where the discrete keys were not being properly read from the separator keys when running the hybrid eliminate method.


gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -315,10 +315,12 @@ class HybridFactorGraph : protected FactorGraph<Factor>,
   /// The total number of factors in the Gaussian factor graph.
   DiscreteKeys discreteKeys() const {
     DiscreteKeys result;
-    // TODO(Frank): implement!
-    // DiscreteKeys result = discreteGraph_.discreteKeys();
-    // DiscreteKeys dcKeys = dcGraph_.discreteKeys();
-    // result += dcKeys
+    for(auto&& key: discreteGraph_.keys()) {
+      //TODO(Varun) how to get cardinality?
+      result.emplace_back(key, 2);
```
varunagrawal, line 320: This is the place that needs the most thought. I guess we'll have to record the DiscreteKeys in the discrete factors. Currently they only record the keys and discard the cardinalities.

CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ endif()
 set (GTSAM_VERSION_MAJOR 4)
 set (GTSAM_VERSION_MINOR 2)
 set (GTSAM_VERSION_PATCH 0)
-set (GTSAM_PRERELEASE_VERSION "a0")
+set (GTSAM_PRERELEASE_VERSION "a1")
```
dellaert, line 14: How come this is in this PR? Did you merge in develop into the wrong branch?

dellaert: If so, merge developed into the base branch as well, so we only see the relevant diff here.

varunagrawal: Okay done!

varunagrawal:
Okay for some reason the merge isn't working the diffs out. We can either absorb this or I can drop the merge commits with some git magic (though this will require a force push). Your choice 🙂

dellaert:
> Okay for some reason the merge isn't working the diffs out. We can either absorb this or I can drop the merge commits with some git magic (though this will require a force push). Your choice 🙂

Just cherry pick the real commits from this into a new branch, started from base?

PR #1015, opened 2022-01-08
Add HybridFactorGraph::toDecisionTreeFactor method

varunagrawal:
This PR adds a new `toDecisionTreeFactor` which can be used to generate the final factor over all discrete values after the continuous values have been eliminated.

### DIFF
--- a/gtsam/hybrid/HybridFactorGraph.cpp
+++ b/gtsam/hybrid/HybridFactorGraph.cpp
@@ -22,6 +22,8 @@ using namespace std;
 
 namespace gtsam {
 
+using GFG = GaussianFactorGraph;
+
 // Instantiate base classes
 // template class FactorGraph<Factor>;
 template class EliminateableFactorGraph<HybridFactorGraph>;
@@ -122,6 +124,21 @@ Sum HybridFactorGraph::sum() const {
   return sum;
 }
 
+DecisionTreeFactor::shared_ptr HybridFactorGraph::toDecisionTreeFactor() const {
+  // Get the decision tree mapping an assignment to a GaussianFactorGraph
+  Sum sum = this->sum();
+
+  // Get the decision tree with each leaf as the error for that assignment
+  std::function<double(GaussianFactorGraph)> gfgError = [&](const GFG& graph) {
+    VectorValues values = graph.optimize();
+    return graph.error(values);
+  };
+  DecisionTree<Key, double> gfgdt(sum, gfgError);
+
+  auto factor = boost::make_shared<DecisionTreeFactor>(discreteKeys(), gfgdt);
+  return factor;
+}
+
 ostream& operator<<(ostream& os,
                     const GaussianFactorGraph::EliminationResult& er) {
   os << "ER" << endl;
@@ -138,7 +155,6 @@ pair<GaussianMixture::shared_ptr, boost::shared_ptr<Factor>> EliminateHybrid(
   // STEP 1: ELIMINATE
   // Eliminate each sum using conventional Cholesky:
   // We can use this by creating a *new* decision tree:
-  using GFG = GaussianFactorGraph;
   using Pair = GaussianFactorGraph::EliminationResult;
 
   KeyVector keys;
@@ -165,8 +181,16 @@ pair<GaussianMixture::shared_ptr, boost::shared_ptr<Factor>> EliminateHybrid(
   // If there are no more continuous parents, then we should create here a
   // DiscreteFactor, with the error for each discrete choice.
   if (separatorKeys.size() == 0) {
-    auto discreteFactor = boost::make_shared<DecisionTreeFactor>(/*TODO*/);
-    cout << "adding a discrete factor!" << endl;
+    // Get the decision tree with each leaf as the error for that assignment
+    std::function<double(GaussianFactorGraph)> gfgError =
+        [&](const GFG& graph) {
+          VectorValues values = graph.optimize();
+          return graph.error(values);
+        };
+    DecisionTree<Key, double> gfgdt(sum, gfgError);
+
+    auto discreteFactor =
+        boost::make_shared<DecisionTreeFactor>(discreteKeys, gfgdt);
     return {conditional, discreteFactor};
   } else {
     // Create a resulting DCGaussianMixture on the separator.
--- a/gtsam/hybrid/HybridFactorGraph.cpp
+++ b/gtsam/hybrid/HybridFactorGraph.cpp
@@ -22,8 +22,6 @@ using namespace std;
 
 namespace gtsam {
 
-using GFG = GaussianFactorGraph;
-
 // Instantiate base classes
 // template class FactorGraph<Factor>;
 template class EliminateableFactorGraph<HybridFactorGraph>;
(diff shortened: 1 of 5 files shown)


gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -122,6 +124,21 @@ Sum HybridFactorGraph::sum() const {
   return sum;
 }
 
+DecisionTreeFactor::shared_ptr HybridFactorGraph::toDecisionTreeFactor() const {
+  // Get the decision tree mapping an assignment to a GaussianFactorGraph
+  Sum sum = this->sum();
+
+  // Get the decision tree with each leaf as the error for that assignment
+  std::function<double(GaussianFactorGraph)> gfgError = [&](const GFG& graph) {
```
dellaert, line 132: Beautiful. But you should say be able to say `auto`

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -122,6 +124,21 @@ Sum HybridFactorGraph::sum() const {
...
+    VectorValues values = graph.optimize();
+    return graph.error(values);
+  };
+  DecisionTree<Key, double> gfgdt(sum, gfgError);
+
+  auto factor = boost::make_shared<DecisionTreeFactor>(discreteKeys(), gfgdt);
```
dellaert, line 138: Just return

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -22,6 +22,8 @@ using namespace std;
 
 namespace gtsam {
 
+using GFG = GaussianFactorGraph;
```
dellaert, line 25: Let’s just use the type. I know I did this, but now  I think it will be more readable if we don’t typedef

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -165,8 +181,16 @@ pair<GaussianMixture::shared_ptr, boost::shared_ptr<Factor>> EliminateHybrid(
...
-    cout << "adding a discrete factor!" << endl;
+    // Get the decision tree with each leaf as the error for that assignment
+    std::function<double(GaussianFactorGraph)> gfgError =
+        [&](const GFG& graph) {
+          VectorValues values = graph.optimize();
+          return graph.error(values);
+        };
+    DecisionTree<Key, double> gfgdt(sum, gfgError);
+
+    auto discreteFactor =
```
dellaert, line 192: Did you not just make a function for all this upstairs?

gtsam/hybrid/tests/testHybridFactorGraph.cpp
```diff
@@ -335,10 +334,31 @@ TEST(DCGaussianElimination, Eliminate_fully) {
   // EXPECT_LONGS_EQUAL(1, result.second->size());
 }
 
+/* ****************************************************************************/
+/// Test the toDecisionTreeFactor method
+TEST(HybridFactorGraph, AssignmentFactors) {
```
dellaert, line 339: toDecisionTreeFactor

gtsam/hybrid/tests/testHybridFactorGraph.cpp
```diff
@@ -335,10 +334,31 @@ TEST(DCGaussianElimination, Eliminate_fully) {
...
+      self.linearizedFactorGraph.dcGraph(),
+      self.linearizedFactorGraph.gaussianGraph());
+
+  auto decisionTreeFactor = linearizedFactorGraph.toDecisionTreeFactor();
+
+  auto allAssignments =
+      cartesianProduct(linearizedFactorGraph.discreteKeys());
+
+  for (auto&& assignment : allAssignments) {
+    std::cout << (*decisionTreeFactor)(assignment) << std::endl << std::endl;
```
dellaert, line 354: Where is the test?

dellaert (commented):
Awesome, but the test needs to be finished. At least test something

gtsam/hybrid/tests/testHybridFactorGraph.cpp
varunagrawal: Added some regression tests.

varunagrawal:
@dellaert I changed the noise models for the BetweenFactors to `Isotropic::Sigma(1, 0.00000005)` and the new potentials I get are:

```
factor 2: Potentials: Choice(m2)
Potentials: 0 Choice(m1)
Potentials: 0 0 Leaf 0.12
Potentials: 0 1 Leaf -0.031
Potentials: 1 Choice(m1)
Potentials: 1 0 Leaf -0.031
Potentials: 1 1 Leaf 2.5e-11
```

This looks great to me since m1=1, m1=1 gives the lowest error, and the other errors also make sense.

gtsam/hybrid/tests/testHybridFactorGraph.cpp
```diff
@@ -335,10 +335,36 @@ TEST(DCGaussianElimination, Eliminate_fully) {
...
+  HybridFactorGraph linearizedFactorGraph(
+      NonlinearFactorGraph(), DiscreteFactorGraph(),
+      self.linearizedFactorGraph.dcGraph(),
+      self.linearizedFactorGraph.gaussianGraph());
+
+  auto decisionTreeFactor = linearizedFactorGraph.toDecisionTreeFactor();
+
+  auto allAssignments = cartesianProduct(linearizedFactorGraph.discreteKeys());
+
+  // Regressions: test all 4 assignments.
```
dellaert, line 353: So, a regression test is not what is needed here, but something like:
- create nonlinear factor graph for m1=0, m2=1
- linearize, optimize, get true error
- compare with error stored in decision tree

if this works for 01, we can trust it to work for all 4 combinations.

dellaert (approved):
I'll approve, but your regression tests seemed more work than a "real" test. I added guidance on how to make it real.

gtsam/hybrid/tests/testHybridFactorGraph.cpp
varunagrawal: I was thinking of doing this but the errors are currently too close to each other to really make that distinction.

I'll update the noise models so we get error values in the 1e-2 range that can be more easily compared, after that adding this separate graph should make a lot of sense.

dellaert: I'm (reluctantly) fine with it as is, just for future reference, a *real* test like that is always preferred to hardcoding values; how do you *know* those values are correct?

varunagrawal: Yes I completely agree. This test is the equivalent of a strawman argument and wanted to use this time to brainstorm a proper test. I got an idea while I was out running. :)

dellaert: Something like

- create nonlinear factor graph for m1=0, m2=1
- linearize, optimize, get true error
- compare with error stored in decision tree
perhaps ? :-)

varunagrawal: All done! You should like this new test a lot more 😄

gtsam/hybrid/tests/testHybridFactorGraph.cpp
```diff
@@ -335,10 +337,68 @@ TEST(DCGaussianElimination, Eliminate_fully) {
...
+
+  // Get the error of the discrete assignment m1=0, m2=1.
+  double actual = (*decisionTreeFactor)(allAssignments[1]);
+
+  /********************************************/
+  // Create equivalent factor graph for m1=0, m2=1
+  GaussianFactorGraph graph;
+
+  // Create the linearization point.
+  Values linearizationPoint;
```
dellaert, line 368: Cool test! Note that `self.linearizationPoint` exists so no need for this. Can just fold into next PR.

varunagrawal:
Ah crap, I accidentally merged this into the older `hybrid_separator` branch.

PR #1017, opened 2022-01-08
fail fast in linux matrix

dellaert:
Currently we keep running all Linux CIs even if one failed already. This seems wasteful.


varunagrawal (approved):
LGTM, though one thing to think about is getting all the errors to fix, not just one at a time. Though this should be resolved locally fist I guess.

PR #1022, opened 2022-01-09
html renderers in discrete

dellaert:
This PR adds html renderers for all discrete types for which markdown existed, and additionally for `DiscreteValues`. Importantly, it also 
- moves Names to DiscreteKeys
- makes the free version `cartesianProduct` a static method of `Assignment`
- adds a specialization of that in `DiscreteValues`
- markdown in `DiscreteConditional` is significantly re-factored to avoid copy/paste


gtsam/discrete/tests/testDiscreteValues.cpp
```diff
@@ -0,0 +1,76 @@
...
+  values[12] = 1;  // A
+  values[5] = 0;   // B
+  string expected =
+      "<div>\n"
+      "<table class='DiscreteValues'>\n"
+      "  <thead>\n"
+      "    <tr><th>Variable</th><th>value</th></tr>\n"
+      "  </thead>\n"
+      "  <tbody>\n"
+      "    <tr><th>B</th><td>'-</td></tr>\n"
```
ProfFan, line 60: Should be `" <tr><th>B</th><td>-</td></tr>\n"`

gtsam/discrete/tests/testDiscreteValues.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+/* ************************************************************************* */
+// Check html representation with a value formatter.
+TEST(DiscreteValues, htmlWithValueFormatter) {
+  DiscreteValues values;
+  values[12] = 1;  // A
+  values[5] = 0;   // B
+  string expected =
+      "<div>\n"
+      "<table class='DiscreteValues'>\n"
```
ProfFan, line 55: Should be double quotes escaped

gtsam/discrete/DiscreteBayesTree.cpp
```diff
@@ -72,5 +72,23 @@ namespace gtsam {
     return ss.str();
   }
 
+  /* **************************************************************************/
+  std::string DiscreteBayesTree::html(
```
ProfFan, line 76: This is NOT tested

gtsam/discrete/DiscreteValues.cpp
```diff
@@ -0,0 +1,97 @@
...
+  // Print out header row.
+  ss << "    <tr><th>Variable</th><th>value</th></tr>\n";
+
+  // Finish header and start body.
+  ss << "  </thead>\n  <tbody>\n";
+
+  // Print out all rows.
+  for (const auto& kv : *this) {
+    ss << "    <tr>";
+    ss << "<th>" << keyFormatter(kv.first) << "</th><td>\'"
```
ProfFan, line 79:
```suggestion
    ss << "<th>" << keyFormatter(kv.first) << "</th><td>"
```

gtsam/discrete/DiscreteValues.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  return ss.str();
+}
+
+string DiscreteValues::html(const KeyFormatter& keyFormatter,
+                            const Names& names) const {
+  stringstream ss;
+
+  // Print out preamble.
+  ss << "<div>\n<table class=\'DiscreteValues\'>\n  <thead>\n";
```
ProfFan, line 68:
```suggestion
  ss << "<div>\n<table class=\"DiscreteValues\">\n  <thead>\n";
```

ProfFan (approved):
Approving with some comments to unblock

gtsam/discrete/tests/testDiscreteValues.cpp
dellaert: Done

gtsam/discrete/tests/testDiscreteValues.cpp
dellaert: I was wondering. I did some searching and [this](https://docs.microsoft.com/en-us/cpp/c-language/c-string-literals?view=msvc-170) says  "The single quotation mark (') can be represented without an escape sequence."

dellaert: Or are yuo saying html needs double quotes?

dellaert: I think single is fine: https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.2

gtsam/discrete/DiscreteBayesTree.cpp
dellaert: You are right.

gtsam/discrete/DiscreteValues.cpp
dellaert: I will just make them single quotes.

PR #1023, opened 2022-01-09
Fix Issue with Hybrid Elimination

varunagrawal:
@ProfFan and I finished writing up the test that confirms our suspicion about Hybrid Elimination being myopic.

So for a simple factor graph 
```
*-X1-*-X2
    /
   M1
```

when we eliminate X1 and X2 in order, the final elimination for X2 only involves a factor between X2 and M1, and not X1. Because of this, the decision tree formed is only on the gaussian factor graph on X2, thus for either 0/1 assignment of M1 the error is only on X2 (which is 0) hence the tree gets pruned to be just a leaf rather than a choice + 2 leaves.

__UPDATE__

By modifying the factor graph a bit to add a measurement factor on each variable, we are able to show that we do get the correct behavior.

```
    /*  /*
*-X1-*-X2
    /
   M1
```

### DIFF
--- a/gtsam/hybrid/HybridFactorGraph.cpp
+++ b/gtsam/hybrid/HybridFactorGraph.cpp
@@ -108,11 +108,6 @@ static Sum& operator+=(Sum& sum, const GaussianFactor::shared_ptr& factor) {
 }
 
 Sum HybridFactorGraph::sum() const {
-  if (nrNonlinearFactors()) {
-    throw runtime_error(
-        "HybridFactorGraph::sum cannot handle NonlinearFactors.");
-  }
-
   if (nrDiscreteFactors()) {
     throw runtime_error(
         "HybridFactorGraph::sum cannot handle DiscreteFactors.");
--- a/gtsam/hybrid/HybridFactorGraph.cpp
+++ b/gtsam/hybrid/HybridFactorGraph.cpp
@@ -139,7 +139,7 @@ DecisionTreeFactor::shared_ptr HybridFactorGraph::toDecisionTreeFactor() const {
   // Get the decision tree with each leaf as the error for that assignment
   auto gfgError = [&](const GaussianFactorGraph& graph) {
     VectorValues values = graph.optimize();
-    return graph.error(values);
+    return graph.probPrime(values);
   };
   DecisionTree<Key, double> gfgdt(sum, gfgError);
 
--- a/gtsam/hybrid/tests/testHybridFactorGraph.cpp
+++ b/gtsam/hybrid/tests/testHybridFactorGraph.cpp
@@ -172,7 +172,7 @@ struct Switching {
 
   // Create motion models for a given time step
   static std::vector<MotionModel::shared_ptr> motionModels(size_t k,
-                                                    double sigma = 1.0) {
+                                                           double sigma = 1.0) {
     auto noise_model = Isotropic::Sigma(1, sigma);
     auto still =
              boost::make_shared<MotionModel>(X(k), X(k + 1), 0.0, noise_model),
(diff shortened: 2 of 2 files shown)


dellaert:
Can you show the comparison between true errors (and prob prime) and calculated errors (and prob prime) in the PR comment?

dellaert:
Also, could I cancel all workflows on these "internal" PRs? They are clogging up CI and I can't build thw wheels that I promised Seth.

varunagrawal:
Yes here are the numbers:

From the DecisionTreeFactor after elimination:

```
DecisionTreeFactor mode 0 error: 0
DecisionTreeFactor mode 1 error: 0
```

By generating the 2 separate factor graphs (for each mode) and then evaluating error/probability on the true values.
```
error for mode 0: 0.5
error for mode 1: 0
prob for mode 0: 0.778801
prob for mode 1: 1
```

varunagrawal:
Adding comments from today's meeting here.

As per the cholesky factorization, each variable getting eliminated gives us the R,d values for that variable, and the remaining information matrix for the remaining variables.

What this means is that the error values in the matrix should get transferred over to each elimination step until we have eliminated all the variables and have the final `e` vector.

So this means that there is no information loss and we should be able to get the proper error.

I thought about this some more and it seems the major issue is that while the elimination has access to all the necessary info, the final elimination step only has access to parts of the Ab matrix pertaining to the last variable. I'll check this thoroughly, but this is my thought from the information I have working on this.

dellaert:
> I thought about this some more and it seems the major issue is that while the elimination has access to all the necessary info, the final elimination step only has access to parts of the Ab matrix pertaining to the last variable. 

Right, but that *contains* all information from all previous elimination steps.

dellaert:
> But can we access all that information inside the elimination function? That's the primary issue: that M1 only has a factor between itself and X2, so it can't create a decision tree with X1 also involved.

Again, the last elimination step that computes the factor on X2 has all that information in the e part. If you are lucky - but this is maybe not a long term solution - it just sits in the information matrix, in the lower-left entry.

Rather than hypothesizing about this, can you investigate it?

varunagrawal:
Okay I figured it out, though I'm going to need help on doing this correctly.

The error value for the GaussianFactorGraph is computed using 
```
VectorValues values = graph.optimize();
return graph.error(values);
```

This error will always be zero since the `values` are optimized to that specific factor graph. In iMHS we were using the factor graph to compute the error i.e.
```
result = gfg.optimize()
error = factor_graph.error(result)
```

so how would that work here?

dellaert:
> Okay I figured it out, though I'm going to need help on doing this correctly.
> 
> The error value for the GaussianFactorGraph is computed using
> 
> ```
> VectorValues values = graph.optimize();
> return graph.error(values);
> ```
> 
> This error will always be zero since the `values` are optimized to that specific factor graph. In iMHS we were using the factor graph to compute the error i.e.
> 
> ```
> result = gfg.optimize()
> error = nlfg.error(result)
> ```
> 
> so how would that work here?

why would the error be zero? Please look at some code, examples to get intuition, which I think is still lacking.

varunagrawal:
> > But can we access all that information inside the elimination function? That's the primary issue: that M1 only has a factor between itself and X2, so it can't create a decision tree with X1 also involved.
> 
> Again, the last elimination step that computes the factor on X2 has all that information in the e part. If you are lucky - but this is maybe not a long term solution - it just sits in the information matrix, in the lower-left entry.
> 
> Rather than hypothesizing about this, can you investigate it?

Yup you are correct and I looked at the results from elimination to verify this. Which is what lead to my next comment on the error values.

I'm looking into that at a deep level now.

dellaert:
> > > But can we access all that information inside the elimination function? That's the primary issue: that M1 only has a factor between itself and X2, so it can't create a decision tree with X1 also involved.
> > 
> > 
> > Again, the last elimination step that computes the factor on X2 has all that information in the e part. If you are lucky - but this is maybe not a long term solution - it just sits in the information matrix, in the lower-left entry.
> > Rather than hypothesizing about this, can you investigate it?
> 
> Yup you are correct and I looked at the results from elimination to verify this. Which is what lead to my next comment on the error values.
> 
> I'm looking into that at a deep level now.

I think you're looking at a very special case where the error is indeed zero. I would add (conflicting) "measurement" factors on x1 and x2 (i.e., unary Gaussians with means that conflict with the current prior and motion model you have now) and you will see the error is *not* zero. That would be a better test anyway.

varunagrawal:
@dellaert I struck upon the same idea before you mentioned it. It works now! :smile:

varunagrawal:
Okay one test will fail but that should be resolved once I merge in #1027.

varunagrawal:
And we are done!

dellaert (approved):
Very cool!

dellaert (commented):
For a different PR, three potential improvements:
1) you do factors.sum() twice, *and* the graphs are eliminated twice. You can simply get the optimal solution from the conditional using solve.
2) This means that, again, Cholesky has probably calculated the error already, and we don’t even have to anything like that. Print out the augmented hessian please and share the results?
3) make an unzip in decisiontree

PR #1028, opened 2022-01-10
Wrap double for between factor

varunagrawal:
Wrapped double so we can run tests in Python as well for quick and easy examination.


dellaert (commented):
I'd like to see a test where this works.
We might have to do with a Vector.

varunagrawal:
@dellaert added the test. I wouldn't make the PR if I wasn't convinced it would work. :slightly_smiling_face:

dellaert (approved):

PR #1034, opened 2022-01-12
Fix probPrime

### LINKED ISSUE #1032: Bug in probPrime
## Description

`probPrime` in `GaussianFactorGraph` applies a 0.5 constant to the error, however each factor already applies 0.5 when computing its error, so the extra 0.5 is wrong.

## Steps to reproduce

We can reproduce this with a simple unit test on a `JacobianFactor`.

## Expected behavior

The `probPrime` should be correct.

varunagrawal:
Added better tests for probPrime and fixed the function implementation.

Fixes #1032


ProfFan (commented):
LGTM, but I'll leave approval to Frank LOL

dellaert (approved):

PR #1037, opened 2022-01-16
Better DiscreteConditional

dellaert:
This PR has 
- creation of joint densities using operator*. It replaces the very poorly tested and confusing `Compose`, which was not used anywhere.
- extensive testing
- removed toFactor, which is a remnant from when we did not inherit from factors
- exposes creation of a DiscretePrior from a vector of floats.
- added a constructor from an ADT
- some smaller documentation improvements

### DIFF
--- a/python/gtsam/tests/test_DiscreteConditional.py
+++ b/python/gtsam/tests/test_DiscreteConditional.py
@@ -16,6 +16,13 @@ import unittest
 from gtsam import DecisionTreeFactor, DiscreteConditional, DiscreteKeys
 from gtsam.utils.test_case import GtsamTestCase
 
+# Some DiscreteKeys for binary variables:
+A = 0, 2
+B = 1, 2
+C = 2, 2
+D = 4, 2
+E = 3, 2
+
 
 class TestDiscreteConditional(GtsamTestCase):
     """Tests for Discrete Conditionals."""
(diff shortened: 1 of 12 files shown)


python/gtsam/tests/test_DiscreteConditional.py
```diff
@@ -36,6 +43,53 @@ def test_single_value_versions(self):
...
+                self.assertAlmostEqual(
+                    actual(v), AB_given_D(v) * C_given_DE(v))
+
+    def test_marginals(self):
+        conditional = DiscreteConditional(A, [B], "1/2 2/1")
+        prior = DiscreteConditional(B, "1/2")
+        pAB = prior * conditional
+        self.gtsamAssertEquals(prior, pAB.marginal(B[0]))
+
+        pA = DiscreteConditional(A % "5/4")
```
varunagrawal, line 90: What is happening over here? We have `(0, 2) % "5/4"`, not sure if this should be `(0, 2), "5/4"`.

varunagrawal (approved):
LGTM sans one comment

python/gtsam/tests/test_DiscreteConditional.py
dellaert: Yeah, will fix that

varunagrawal: I am making fixes since other stuff is also breaking. 🙂

dellaert:
Thanks! Now I'm behind a CI traffic jam, though, so I'd prefer if you just comment in future? :-)

varunagrawal:
It'll be fast enough. Go enjoy a hot cup of coffee in this weather and it will be done before you know it. :)

dellaert:
> It'll be fast enough. Go enjoy a hot cup of coffee in this weather and it will be done before you know it. :)

pretty please. I need to get this release out.

PR #1039, opened 2022-01-16
Rename DiscretePrior -> DiscreteDistribution

dellaert:
I added the class `DiscretePrior` in 4.2a0 for pedagogical reasons in the [evolving book](https://www.roboticsbook.org/intro.html). It turns out that it is much more useful than just use as a prior, in particular for implementing Bayes law in section 2.4. Hence, here I am renaming it in this pre-release to `DiscreteDistribution`. This PR is almost exclusively a global rename.

Please don't push anything until I released next alpha release, I can't afford to wait for a lengthy CI. TAs need this.

### DIFF
--- a/gtsam/discrete/DiscreteBayesNet.h
+++ b/gtsam/discrete/DiscreteBayesNet.h
@@ -19,7 +19,7 @@
 #pragma once
 
 #include <gtsam/discrete/DiscreteConditional.h>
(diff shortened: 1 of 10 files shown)


ProfFan (approved):

dellaert:
Thanks !!!

PR #1041, opened 2022-01-17
Finish release 4.2a3

dellaert:
merges all commits since last release.


dellaert:
Thanks. Please note there is also https://github.com/borglab/gtsam/pull/1040 to approve

PR #1042, opened 2022-01-18
Added a convenience function for plotting 2D points.

magicbycalvin:
Included a convenience function for plotting 2D points in Python. The new code closely resembles the existing plotting functions available in `gtsam.utils.plot`. Plotting a 2D point is useful for visualizing the location of landmarks.

### DIFF
--- a/python/gtsam/utils/plot.py
+++ b/python/gtsam/utils/plot.py
@@ -125,7 +125,10 @@ def plot_point2_on_axes(axes,
     if P is not None:
         w, v = np.linalg.eig(P)
 
-        # k = 2.296
+        # "Sigma" value for drawing the uncertainty ellipse. 5 sigma corresponds
+        # to a 99.9999% confidence, i.e. assuming the estimation has been
+        # computed properly, there is a 99.999% chance that the true position
+        # of the point will lie within the uncertainty ellipse.
         k = 5.0
 
         angle = np.arctan2(v[1, 0], v[0, 0])
--- a/python/gtsam/utils/plot.py
+++ b/python/gtsam/utils/plot.py
@@ -10,7 +10,7 @@ from matplotlib import patches
 from mpl_toolkits.mplot3d import Axes3D  # pylint: disable=unused-import
 
 import gtsam
-from gtsam import Marginals, Point3, Pose2, Pose3, Values
+from gtsam import Marginals, Point2, Point3, Pose2, Pose3, Values
 
 
 def set_axes_equal(fignum: int) -> None:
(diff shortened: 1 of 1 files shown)


python/gtsam/utils/plot.py
```diff
@@ -108,6 +108,66 @@ def plot_covariance_ellipse_3d(axes,
...
+        axes (matplotlib.axes.Axes): Matplotlib axes.
+        point: The point to be plotted.
+        linespec: String representing formatting options for Matplotlib.
+        P: Marginal covariance matrix to plot the uncertainty of the estimation.
+    """
+    axes.plot([point[0]], [point[1]], linespec, marker='.', markersize=10)
+    if P is not None:
+        w, v = np.linalg.eig(P)
+
+        # k = 2.296
```
varunagrawal, line 128: Kill this line? And maybe add a comment on what `k` is?

varunagrawal (approved):
LGTM but please address my comment before merging.

python/gtsam/utils/plot.py
magicbycalvin: I removed the commented out line and included a comment as to what `k` is (the "sigma" value for the uncertainty ellipse). I specifically chose 5 sigma as that is the same value used in the other functions, such as `plot_pose2_on_axes`.
Thanks!

dellaert:
@varunagrawal I'll let you merge

python/gtsam/utils/plot.py
```diff
@@ -108,6 +108,69 @@ def plot_covariance_ellipse_3d(axes,
...
+        axes (matplotlib.axes.Axes): Matplotlib axes.
+        point: The point to be plotted.
+        linespec: String representing formatting options for Matplotlib.
+        P: Marginal covariance matrix to plot the uncertainty of the estimation.
+    """
+    axes.plot([point[0]], [point[1]], linespec, marker='.', markersize=10)
+    if P is not None:
+        w, v = np.linalg.eig(P)
+
+        # "Sigma" value for drawing the uncertainty ellipse. 5 sigma corresponds
```
senselessDev, line 128: @magicbycalvin This is not correct in 2D, your values seem to be for the 1D case. See `Geometric interpretation` [here](https://en.wikipedia.org/wiki/Multivariate_normal_distribution#Properties) and read [this](https://stackoverflow.com/a/39749274/1959116). Can you fix that in another PR?

dellaert: Actually, I added a note in #1063, and I think k=5 indeed corresponds to 99.9996% in 2D case. Check the note and let me know... That PR could be amended to accommodate a change if needed, if @gchenfc does not merge it first.

dellaert: It's in plot.py, in that PR

magicbycalvin: Just now seeing this discussion. I have since added an additional PR addressing the comments above.

senselessDev: > Actually, I added a note in #1063, and I think k=5 indeed corresponds to 99.9996% in 2D case. Check the note and let me know... That PR could be amended to accommodate a change if needed, if @gchenfc does not merge it first.

While I agree with 99.9996% (see #1067), that is not what was written in the comment:

> to a 99.9999% confidence, i.e. assuming the estimation has been computed properly, there is a 99.999% chance that the true position

PR #1043, opened 2022-01-18
DiscreteBayesNet sample/optimize variants

dellaert:
Two more sample/optimize variants in `DiscreteBayesNet`, that allow us to sample "incomplete" Bayes nets, e.g., resulting from incomplete elimination. Two things to note:
- test is in Python.
- `given` argument is passed by value, no accident. 
- added better docs as well
- I reformatted .cpp file, might want to hide whitespace changes


xsj01 (approved):
Looks good to me.

PR #1045, opened 2022-01-19
Some discrete improvements

dellaert:
Some more changes needed for the book:
- New, more powerful choose, yields a Conditional now
- Wrap push_back methods for DiscreteFactorGraph, and also did some reformatting/docs/using
- Change factor graph plotting default to not confuse people


tests/testNonlinearFactorGraph.cpp
```diff
@@ -341,9 +341,15 @@ TEST(NonlinearFactorGraph, dot) {
...
-      "  var8646911284551352321--var7782220156096217089;\n"
-      "  var8646911284551352322--var7782220156096217089;\n"
+      "  factor1[label=\"\", shape=point];\n"
+      "  var8646911284551352321--factor1;\n"
+      "  var8646911284551352322--factor1;\n"
+      "  factor2[label=\"\", shape=point];\n"
+      "  var8646911284551352321--factor2;\n"
+      "  var7782220156096217089--factor2;\n"
+      "  factor3[label=\"\", shape=point];\n"
+      "  var8646911284551352322--factor3;\n"
```
ProfFan, line 351: is this intended?

dellaert: It is, but now that I think about it, I don't understand how this is portable across different machines. Uses a pointer address to create a unique labels for a factor.

dellaert: But all checks are successful !?

ProfFan (approved):

tests/testNonlinearFactorGraph.cpp
ProfFan: These are actually symbols not pointers, so a formatter would be great

PR #1036, opened 2022-01-16
WIP for incremental discrete

ProfFan:
WIP for incremental discrete elimination.

PR #1047, opened 2022-01-19
GaussianHybridFactorGraph class

varunagrawal:
This PR defines a new class called `GaussianHybridFactorGraph` which is the linearized version of the `HybridFactorGraph`.

As such, I moved all the elimination machinery to `GaussianHybridFactorGraph` and made other updates.

NOTE: Please review this PR carefully since I made a lot of changes to parts which were originally @dellaert's code. I tried my best to understand the overall design and what the meaning of each part is, but I could have easily made a mistake.


dellaert:
So, it seems you copy/pasted?
I think the right thing to do is to make HybridFactorGraph a templated base class, so we have:

```c++
template<class FG> 
HybridFactorGraph {
// storage for DiscreteFG, DCFG, FG
};

GaussianHybridFactorGraph : public HybridFactorGraph<GaussianFactorGraph> {};
NonlinearHybridFactorGraph : public HybridFactorGraph<NonlinearFactorGraph> {};
```

varunagrawal:
Yep pretty much copy-pasted in an effort to minimize time spent on this. That's a great suggestion, I was also thinking of the same thing but wanted to minimize changes in this PR.

dellaert:
> Yep pretty much copy-pasted in an effort to minimize time spent on this. That's a great suggestion, I was also thinking of the same thing but wanted to minimize changes in this PR.

I think this *would* minimize PR changes, and make it easier to review. Only small changes to HFG are needed I think. And the two derived classes would be small.

varunagrawal:
Oh I completely agree with the recommended change. I was just laying out my thought process so that we were all on the same page :)

varunagrawal:
@dellaert if we make HybridFactorGraph a templated class, how would things like `IsNonlinear` and `IsDiscrete` work?

varunagrawal:
Okay so I made `HybridFactorGraph` a templated base class, but I don't think it got us much. I kind of preferred the prior separation of `IsGaussian` vs `IsNonlinear` methods since they made semantic sense inside the individual subclasses.

Plus, by templating, we had to move some of the definitions to the header == larger binary size.

gtsam/discrete/DecisionTreeFactor.cpp
```diff
@@ -167,7 +167,10 @@ namespace gtsam {
   DiscreteKeys DecisionTreeFactor::discreteKeys() const {
     DiscreteKeys result;
     for (auto&& key : keys()) {
-      result.emplace_back(key, cardinality(key));
+      DiscreteKey dkey(key, cardinality(key));
```
dellaert, line 170: Why? This should never be triggered, and we should get to root cause of the cases that *would* trigger this.

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -14,58 +14,24 @@
...
-  /// The function type that does a single elimination step on a variable.
-  static std::pair<GaussianMixture::shared_ptr, sharedFactor> DefaultEliminate(
-      const HybridFactorGraph& factors, const Ordering& ordering) {
-    return EliminateHybrid(factors, ordering);
-  }
-};
-
-class HybridFactorGraph : protected FactorGraph<Factor>,
-                          public EliminateableFactorGraph<HybridFactorGraph> {
+template <typename FG>
```
dellaert, line 24: Doxygen comment with explanation of FG

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -77,6 +43,11 @@ class HybridFactorGraph : protected FactorGraph<Factor>,
   using IsDC =
       typename std::enable_if<std::is_base_of<DCFactor, FACTOR>::value>::type;
 
+  /// Check if FACTOR type is derived from NonlinearFactor.
```
dellaert, line 46: why is that here? This class is not supposed to know about nonlinear factors

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -89,22 +60,19 @@ class HybridFactorGraph : protected FactorGraph<Factor>,
   /**
    * @brief Construct a new Hybrid Factor Graph object.
    *
-   * @param nonlinearGraph A factor graph with continuous factors.
+   * @param discreteGraph A factor graph with factors depending on type of `FG`.
```
dellaert, line 63: factorGraph *of* type* FG

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -227,16 +151,13 @@ class HybridFactorGraph : protected FactorGraph<Factor>,
   /**
    * Simply prints the factor graph.
    */
-  void print(
+  virtual void print(
```
dellaert, line 154: Why does it not override?

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -14,58 +14,24 @@
...
-  static std::pair<GaussianMixture::shared_ptr, sharedFactor> DefaultEliminate(
-      const HybridFactorGraph& factors, const Ordering& ordering) {
-    return EliminateHybrid(factors, ordering);
-  }
-};
-
-class HybridFactorGraph : protected FactorGraph<Factor>,
-                          public EliminateableFactorGraph<HybridFactorGraph> {
+template <typename FG>
+class GTSAM_EXPORT HybridFactorGraph : protected FactorGraph<Factor> {
```
dellaert, line 25: no EXPORT on templates/header only classes -> move to derived.

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -253,54 +174,43 @@ class HybridFactorGraph : protected FactorGraph<Factor>,
...
 
-  /// The total number of factors in the Gaussian factor graph.
-  size_t nrGaussianFactors() const { return gaussianGraph_.size(); }
-
   /**
    * Clears all internal factor graphs
    * TODO(dellaert): Not loving this!
    */
-  void clear();
+  void clear() {
```
dellaert, line 193: Propose to remove this imperative method. Create a new instance!

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -14,58 +14,24 @@
...
  public:
   using shared_ptr = boost::shared_ptr<HybridFactorGraph>;
   using Base = FactorGraph<Factor>;
 
  protected:
   // Separate internal factor graphs for different types of factors
-  NonlinearFactorGraph nonlinearGraph_;
+  FG factorGraph_;
   DiscreteFactorGraph discreteGraph_;
   DCFactorGraph dcGraph_;
```
dellaert, line 34: This could still contain nonlinear things of course :-( Should we template on Factor type instead ? NonlinearFactor | GaussianFactor? I'd say leave for now but something to discuss.

gtsam/hybrid/NonlinearHybridFactorGraph.h
```diff
@@ -0,0 +1,144 @@
...
+  template <typename FACTOR>
+  IsNonlinear<FACTOR> push_nonlinear(
+      const boost::shared_ptr<FACTOR>& nonlinearFactor) {
+    factorGraph_.push_back(nonlinearFactor);
+    Base::Base::push_back(nonlinearFactor);
+  }
+
+  /// Construct a factor and add (shared pointer to it) to factor graph.
+  template <class FACTOR, class... Args>
+  IsGaussian<FACTOR> emplace_gaussian(Args&&... args) {
```
dellaert, line 65: huh ? Remove?

gtsam/hybrid/NonlinearHybridFactorGraph.h
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  /**
+   * @brief Linearize all the continuous factors in the
+   * NonlinearHybridFactorGraph.
+   *
+   * @param continuousValues: Dictionary of continuous values.
+   * @return GaussianHybridFactorGraph
+   */
+  GaussianHybridFactorGraph linearize(const Values& continuousValues) const;
```
dellaert, line 126: I love this !

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -23,190 +23,33 @@
...
-    result.push_back(factor);
-    return result;
-  };
-  sum = sum.apply(add);
-  return sum;
-}
 
-Sum HybridFactorGraph::sum() const {
+template <typename FG>
+Sum HybridFactorGraph<FG>::sum() const {
```
dellaert, line 40: template should be in header. Or just specialize it on GaussianFG and move there.

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -23,190 +23,33 @@
...
-  }
-
-  // Construct new linearized HybridFactorGraph
-  return HybridFactorGraph({}, discreteGraph_, linearized_DC_factors,
-                           *gaussianFactorGraph);
-}
-
-bool HybridFactorGraph::equals(const HybridFactorGraph& other,
-                               double tol) const {
+template <typename FG>
```
dellaert, line 28: should be in header

dellaert: In fact, kill this file??

gtsam/hybrid/GaussianHybridFactorGraph.h
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * @brief Add a single factor shared pointer to the hybrid factor graph.
+   * Dynamically handles the factor type and assigns it to the correct
+   * underlying container.
+   *
+   * @tparam FACTOR The factor type template
+   * @param sharedFactor The factor to add to this factor graph.
+   */
+  template <typename FACTOR>
```
dellaert, line 113: Why is this even a template? Factor is a common base class. But maybe that does not work with shared ptrs?

gtsam/hybrid/GaussianHybridFactorGraph.cpp
```diff
@@ -0,0 +1,166 @@
...
+// Instantiate base classes
+template class EliminateableFactorGraph<GaussianHybridFactorGraph>;
+
+void GaussianHybridFactorGraph::print(
+    const string& str, const gtsam::KeyFormatter& keyFormatter) const {
+  Base::print(str, keyFormatter);
+  factorGraph_.print("GaussianGraph", keyFormatter);
+}
+
+DiscreteKeys GaussianHybridFactorGraph::discreteKeys() const {
```
dellaert, line 39: Why is that here? Expected in HFG...

dellaert (commented):
I *love* this. So much better!
One more iteration before approve?

gtsam/hybrid/HybridFactorGraph.h
varunagrawal: Okay so should I move all the `IsNonlinear` based methods to `NonlinearHybridFactorGraph` and all the `IsGaussian` based methods to `GaussianHybridFactorGraph`? Then my previous comment on the benefits of templating come up since we can't move a lot of things to the .cpp file.

gtsam/discrete/DecisionTreeFactor.cpp
varunagrawal: The lack of this causes a test to fail since we get M1 twice. There is double counting because both keys and discreteKeys in the DCFactorGraph have M1.

gtsam/hybrid/HybridFactorGraph.h
varunagrawal: I further override it in the subclasses but I guess it's not required.

gtsam/hybrid/NonlinearHybridFactorGraph.h
varunagrawal: Ah nice catch

gtsam/hybrid/HybridFactorGraph.cpp
varunagrawal: Killed the file.

gtsam/hybrid/GaussianHybridFactorGraph.h
varunagrawal: We would need the template for SFINAE right? To make sure the push_back is happening to the correct factor graph?

gtsam/hybrid/GaussianHybridFactorGraph.cpp
varunagrawal: It's an override specifically for dealing with factors in `discreteGraph_` which are `DecisionTreeFactor`s.

varunagrawal:
I'm still questioning the mileage we're getting from templating. `factorGraph_` is used in just 3 functions in `HybridFactorGraph`. All the work is still happening in the subclasses, plus we can possibly remove all the SFINAE and just specialize the `emplace_*` and `push_back*` methods to the appropriate class.
That would make the code a lot leaner IMHO.

gtsam/hybrid/GaussianHybridFactorGraph.cpp
dellaert: But that is there in HFG…

gtsam/hybrid/GaussianHybridFactorGraph.h
dellaert: I don’t think this method is an SFINAE. Just dynamic casts

dellaert (commented):
2 things still unresolved I think.

gtsam/hybrid/GaussianHybridFactorGraph.cpp
varunagrawal: I thought there was some difference when I split the code but no test fails so yep I'll remove this. :)

gtsam/hybrid/GaussianHybridFactorGraph.h
varunagrawal: Let me try it out. Also FYI we have a typedef above called `sharedFactor` which I will change to `SharedFactor` so that I can use it here.

varunagrawal: Also you're correct about it not being SFINAE. I was misunderstood what was really happening here and with a fresh mind (+ without the stress of getting something running for the state estimator) this makes a lot of sense.

varunagrawal:
I'm going to wait on #1036 to be merged before merging this so I can update this PR with all the changes and fixes from that PR.

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -7,65 +7,39 @@
...
+ * @author Varun Agrawal
  * @date   December 2021
  */
 
 #pragma once
 
 #include <gtsam/discrete/DiscreteFactorGraph.h>
 #include <gtsam/hybrid/DCFactorGraph.h>
-#include <gtsam/nonlinear/NonlinearFactorGraph.h>
+#include <gtsam/hybrid/DCGaussianMixtureFactor.h>
```
dellaert, line 18: Why is that here?

gtsam/hybrid/HybridFactorGraph.h
dellaert: I like the templated version

gtsam/hybrid/HybridFactorGraph.h
dellaert: Does not seem to be gone

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -310,19 +222,20 @@ class HybridFactorGraph : protected FactorGraph<Factor>,
...
-
-  /**
-   * @brief Convert the DecisionTree of (Key, GaussianFactorGraph) to
-   * (Key, Graph probability).
-   * @return DecisionTreeFactor::shared_ptr
-   */
-  DecisionTreeFactor::shared_ptr toDecisionTreeFactor() const;
-
-  /// @}
+  DCGaussianMixtureFactor::Sum sum() const {
```
dellaert, line 225: Why is this not in GaussianHFG?

dellaert (approved):
I'll approve now, but "sum" still seems to be in the wrong place, I think.
I like the templated version, please keep.

gtsam/hybrid/HybridFactorGraph.h
varunagrawal: Ah sorry about that

gtsam/hybrid/HybridFactorGraph.h
varunagrawal: Oh I see what you mean. I thought create new instance meant for each member factor graph.

gtsam/hybrid/HybridFactorGraph.h
varunagrawal: Moved it to GaussianHFG

PR #1051, opened 2022-01-21
Eliminate Hybrid Machinery

varunagrawal:
Merge all the work done in `HybridFactorGraph` elimination:

- Split HybridFactorGraph in `NonlinearHybridFactorGraph` and `GaussianHybridFactorGraph`.
- Add `HybridEliminate` method to perform elimination on `GaussianHybridFactorGraph`s.
- Refactor `discrete` to have a better API and make it easy for using in `hybrid`.
- Changed the `DCFactor` paradigm so that `keys()` has both the continuous and the discrete keys.
- CI updates for removing warnings in Windows, and updating Python install on MacOS.


dellaert:
Please merge develop into both feature/hybrid and this branch before I review.

gtsam/discrete/DecisionTree.h
```diff
@@ -41,6 +41,7 @@ namespace gtsam {
    */
   template<typename L, typename Y>
   class DecisionTree {
+
```
dellaert, line 44: remove changes from all files in discrete (I count 3 files)

gtsam/hybrid/DCFactor.h
```diff
@@ -50,11 +61,12 @@ class DCFactor : public Factor {
...
    */
   DCFactor(const KeyVector& continuousKeys, const DiscreteKeys& discreteKeys)
-      : Base(continuousKeys), discreteKeys_(discreteKeys) {}
+      : Base(AllKeys(continuousKeys, discreteKeys)),
+        discreteKeys_(discreteKeys) {}
 
   // NOTE unsure if needed?
   explicit DCFactor(const DiscreteKeys& discreteKeys)
-      : discreteKeys_(discreteKeys) {}
+      : DCFactor({}, discreteKeys) {}
```
dellaert, line 69: Let's remove

gtsam/hybrid/DCFactor.h
```diff
@@ -99,10 +111,24 @@ class DCFactor : public Factor {
    * @param discreteVals - Likewise, assignment to the discrete variables in
    * `discreteKeys__`.
    */
-  virtual boost::shared_ptr<GaussianFactor> linearize(
+  virtual GaussianFactor::shared_ptr linearize(
```
dellaert, line 114: revert and forward declare GaussianFactor, so we don't have to include GaussianFactor.h and co here.

gtsam/hybrid/DCFactor.h
```diff
@@ -99,10 +111,24 @@ class DCFactor : public Factor {
...
       const Values& continuousVals,
       const DiscreteValues& discreteVals) const = 0;
 
+  /**
+   * @brief Linearize all the continuous factors only with respect to the
+   * continuous variables (as given in `keys_`).
+   *
+   * This `linearize` is different from the `linearize(continuous, discrete)` in
+   * that it assumes no assignment of discrete keys and linearizes all
```
dellaert, line 123: Comment on the return type. In fact, should it also be a GaussianMixture?

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -33,55 +33,92 @@ namespace gtsam {
...
+                  const std::vector<sharedFactor>& factors,
+                  bool normalized = false)
+      : DCMixtureFactor(keys, discreteKeys,
+                        Factors(discreteKeys, factors)) {}
 
-  DCMixtureFactor& operator=(const DCMixtureFactor& rhs) {
-    Base::operator=(rhs);
-    this->dk_ = rhs.dk_;
-    this->factors_ = rhs.factors_;
+  /**
```
dellaert, line 86: Let's just kill. Caller should create shared pointers without copying

gtsam/hybrid/DCMixtureFactor.h
```diff
@@ -33,55 +33,92 @@ namespace gtsam {
...
+      auto f_ptr = boost::make_shared<NonlinearFactorType>(factor);
+      factor_pointers.push_back(f_ptr);
+    }
+    // Generate decision tree based on discreteKey to factor mapping.
+    factors_ = Factors(discreteKeys, factor_pointers);
   }
 
   ~DCMixtureFactor() = default;
 
   double error(const Values& continuousVals,
```
dellaert, line 105: doxygen

gtsam/hybrid/GaussianMixture.h
```diff
@@ -0,0 +1,116 @@
...
+   * @param discreteKeys - the keys for *discrete* variables
+   * @param conditionals A decision tree of GaussianConditional instances.
+   * TODO(Frank): (possibly wrongly) assumes nrFrontals is one
+   * TODO(Frank): should pass frontal keys, cont. parent keys, discrete parent
+   * keys, cannot be a conditional on a discrete key.
+   * TODO: desired API =
+   * GaussianConditionalMixture(const Conditionals& conditionals, 
+   *                            const DiscreteKeys& discreteParentKeys)
+   */
+  GaussianMixture(size_t nrFrontals,
```
dellaert, line 67: move definition to cpp? Too ugly here

gtsam/hybrid/GaussianMixture.h
```diff
@@ -0,0 +1,116 @@
...
+                          % nrFrontals % p->nrFrontals()).str());
+                return boost::dynamic_pointer_cast<GaussianFactor>(p);
+              })),
+        BaseConditional(nrFrontals) {}
+
+  /// @}
+  /// @name Standard API
+  /// @{
+
+  GaussianConditional::shared_ptr operator()(
```
dellaert, line 89: same -> cpp

gtsam/slam/BetweenFactor.h
```diff
@@ -80,10 +80,11 @@ namespace gtsam {
     /// @{
 
     /// print with optional string
-    void print(const std::string& s, const KeyFormatter& keyFormatter = DefaultKeyFormatter) const override {
-      std::cout << s << "BetweenFactor("
-          << keyFormatter(this->key1()) << ","
-          << keyFormatter(this->key2()) << ")\n";
+    void print(
```
dellaert, line 83: Remove from this PR. Core method of core factor should not be slipped in.

python/gtsam/tests/test_BetweenFactor.py
```diff
@@ -0,0 +1,33 @@
+"""
```
dellaert, line 1: I don't see change to a .i file. ?

dellaert (commented):
One more round and we can merge. Not yet approving as I feel strongly about some things

python/gtsam/tests/test_BetweenFactor.py
varunagrawal: That change is already in develop for some reason.

gtsam/hybrid/DCFactor.h
varunagrawal: DCFactor is the right type since it is the polymorphic base type.

gtsam/hybrid/DCFactor.h
varunagrawal: I don't quite understand this. GaussianFactor.h is not included here, so do you want to remove the #include from the DCGaussianMixtureFactor?

gtsam/hybrid/DCMixtureFactor.h
varunagrawal: I also found a bug in the `vector<sharedFactor>` version of the constructor. `normalized_` was not being set!

gtsam/hybrid/DCFactor.h
dellaert: If it is not included, it should not compile. If it is included, it should be removed from this file.

gtsam/hybrid/DCFactor.h
dellaert: I don't understand this, can it be anything else, ever? Make the type as specific as it can be.

dellaert:
Don't worry about windows now. Will deal with windows when we try to merge feature/hybrid

varunagrawal:
I actually did a normal push but GitHub didn't reflect it (some sort of bug on their end?) so I had to force push to get it to actually reflect.

gtsam/hybrid/DCFactor.h
varunagrawal: Okay then there's nothing to do here since we include `GaussianFactorGraph.h` (which is needed) and that adds `GaussianFactor` as well. :)

gtsam/hybrid/DCFactor.h
varunagrawal: So we can use DCGaussianMixtureFactor as the return type, but we've been using DCFactor since it is the common base class for all the DC factors. If we update this to use DCGaussianMixtureFactor, then it becomes a little messy since the forward declaration doesn't exactly work and there will be circular includes if we import DCGaussianMixtureFactor in DCFactor.

dellaert: Did you actually investigate or are you shooting from the hip? I think it's a valid review comment.
Why can you not just say 
```
class DCGaussianMixtureFactor; // forward declaration
```
above?  It should return a shared pointer, so really `boost::shared_ptr<DCGaussianMixtureFactor>`  should just work.

gtsam/hybrid/DCFactor.h
dellaert: ok

dellaert (approved):
I'll approve so we can merge but that one comment is still valid I think. Code for the now, not the future. If it works now, do it.

gtsam/hybrid/DCFactor.h
varunagrawal: I actually ran it with the forward declaration and got a bunch of compiler erros, but I did a `DCGaussianMixtureFactor::shared_ptr` which is the wrong thing to do. Let me try again and resolve it with full certainty.

varunagrawal: That worked!! Yay

PR #1052, opened 2022-01-21
Fix various minor MPE issues

varunagrawal:
Addresses my review comments to get things moving along. :)


varunagrawal:
CI killed

dellaert:
I will just cherry-pick because there are things here that I did not want. E.g., I am OK with breaking the unstable API.

PR #1055, opened 2022-01-22
Changes in gtsam to support hybrid branches

dellaert:
This PR collect a number of commits to files made in the hybrid development branches.
I made the commits below descriptive.
@varunagrawal we can re-target the PRs or just wait until this is merged.

### DIFF
--- a/gtsam/discrete/DecisionTree.h
+++ b/gtsam/discrete/DecisionTree.h
@@ -38,7 +38,7 @@ namespace gtsam {
    * Y = function range (any algebra), e.g., bool, int, double
    */
   template<typename L, typename Y>
-  class GTSAM_EXPORT DecisionTree {
+  class DecisionTree {
 
    protected:
     /// Default method for comparison of two objects of type Y.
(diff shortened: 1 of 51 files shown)


gtsam/discrete/DecisionTree.h
```diff
@@ -340,4 +340,11 @@ namespace gtsam {
     return f.apply(g, op);
   }
 
+  /// unzip a DecisionTree if its leaves are `std::pair`
+  template<typename L, typename T1, typename T2>
+  std::pair<DecisionTree<L, T1>, DecisionTree<L, T2> > unzip(const DecisionTree<L, std::pair<T1, T2> > &input) {
```
dellaert, line 345: Actually, @ProfFan , this should probably have been a static method and be called `Unzip`. Or a method, and then it would not take an argument, right?

dellaert: BTW, just looking for consensus, I think I would change it to static.

dellaert:
Killed CI while deciding about Unzip

gtsam/discrete/DecisionTree.h
dellaert: Actually, I see now it is neither: it is a free function :-/ Which is correctly named, so never mind! Will re-format is all.

ProfFan: Do you want to reopen CI?

dellaert: Yes, I'll just do the required one and merge

dellaert:
I think this is ready to merge if I can get an approving review from @ProfFan or @varunagrawal

varunagrawal:
Sorry I was out the entire afternoon. I'll review in the next 30 minutes.

varunagrawal (approved):
LGTM! Merge whenever.

dellaert:
I can't merge because there is some CI waiting to be reported. Any ideas?

ProfFan:
I reopened the PR, now it should start to build. @dellaert

ProfFan:
@dellaert CI is ready :)

PR #1059, opened 2022-01-23
Expose GraphvizFormatting and test in python

senselessDev:
This exposes the `GraphvizFormatting` class with direct member access as discussed in #1053. This allows to do some configuration on the `saveGraph` call that was not possible via the bindings before.

It also adds a few basic tests for it.

Lastly, I had the problem that the python-test target was only testing the files that were copied to the build directory before but not any new tests. Adding the dependency fixed that by always copying the tests. I don't know if that is wanted behavior.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -181,5 +181,5 @@ add_custom_target(
           ${CMAKE_COMMAND} -E env # add package to python path so no need to install
           "PYTHONPATH=${GTSAM_PYTHON_BUILD_DIRECTORY}/$ENV{PYTHONPATH}"
           ${PYTHON_EXECUTABLE} -m unittest discover -v -s .
-          DEPENDS ${GTSAM_PYTHON_DEPENDENCIES}
+          DEPENDS ${GTSAM_PYTHON_DEPENDENCIES} ${GTSAM_PYTHON_TEST_FILES}
           WORKING_DIRECTORY "${GTSAM_PYTHON_BUILD_DIRECTORY}/gtsam/tests")
--- a/python/gtsam/tests/test_GraphvizFormatting.py
+++ b/python/gtsam/tests/test_GraphvizFormatting.py
@@ -1,0 +1,139 @@
+"""
+GTSAM Copyright 2010-2021, Georgia Tech Research Corporation,
+Atlanta, Georgia 30332-0415
+All Rights Reserved
+
+See LICENSE for the license information
+
+Unit tests for Graphviz formatting of NonlinearFactorGraph.
+Author: Frank Dellaert & senselessDev
+"""
+
+# pylint: disable=no-member, invalid-name
+
+import unittest
+import textwrap
+
+import numpy as np
+
+import gtsam
+from gtsam.utils.test_case import GtsamTestCase
+
+
+class TestGraphvizFormatting(GtsamTestCase):
+    """Tests for saving NonlinearFactorGraph to GraphViz format."""
+
+    def setUp(self):
+        self.graph = gtsam.NonlinearFactorGraph()
+
+        odometry = gtsam.Pose2(2.0, 0.0, 0.0)
+        odometryNoise = gtsam.noiseModel.Diagonal.Sigmas(
+            np.array([0.2, 0.2, 0.1]))
+        self.graph.add(gtsam.BetweenFactorPose2(0, 1, odometry, odometryNoise))
+        self.graph.add(gtsam.BetweenFactorPose2(1, 2, odometry, odometryNoise))
+
+        self.values = gtsam.Values()
+        self.values.insert_pose2(0, gtsam.Pose2(0., 0., 0.))
+        self.values.insert_pose2(1, gtsam.Pose2(2., 0., 0.))
+        self.values.insert_pose2(2, gtsam.Pose2(4., 0., 0.))
+
+    def test_default(self):
+        """Test with default GraphvizFormatting"""
+        expected_result = """\
+            graph {
+              size="5,5";
+
+              var0[label="0", pos="0,0!"];
+              var1[label="1", pos="0,2!"];
+              var2[label="2", pos="0,4!"];
+
+              factor0[label="", shape=point];
+              var0--factor0;
+              var1--factor0;
+              factor1[label="", shape=point];
+              var1--factor1;
+              var2--factor1;
+            }
+        """
+
+        self.assertEqual(self.graph.dot(self.values),
+                         textwrap.dedent(expected_result))
+
+    def test_swapped_axes(self):
+        """Test with user-defined GraphvizFormatting swapping x and y"""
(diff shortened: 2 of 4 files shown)


python/CMakeLists.txt
```diff
@@ -181,5 +181,5 @@ add_custom_target(
           ${CMAKE_COMMAND} -E env # add package to python path so no need to install
           "PYTHONPATH=${GTSAM_PYTHON_BUILD_DIRECTORY}/$ENV{PYTHONPATH}"
           ${PYTHON_EXECUTABLE} -m unittest discover -v -s .
-          DEPENDS ${GTSAM_PYTHON_DEPENDENCIES}
+          DEPENDS ${GTSAM_PYTHON_DEPENDENCIES} ${GTSAM_PYTHON_TEST_FILES}
```
dellaert, line 184: yes!!!!

python/gtsam/tests/test_GraphvizFormatting.py
```diff
@@ -0,0 +1,139 @@
+"""
+GTSAM Copyright 2010-2021, Georgia Tech Research Corporation,
+Atlanta, Georgia 30332-0415
+All Rights Reserved
+
+See LICENSE for the license information
+
+Unit tests for Graphviz formatting of NonlinearFactorGraph.
+Author: Frank Dellaert & senselessDev
```
dellaert, line 9: I'm not an author :-)

python/gtsam/tests/test_GraphvizFormatting.py
```diff
@@ -0,0 +1,139 @@
...
+              var0--factor0;
+              var1--factor0;
+              factor1[label="", shape=point];
+              var1--factor1;
+              var2--factor1;
+            }
+        """
+
+        self.assertEqual(self.graph.dot(self.values),
+                         textwrap.dedent(expected_result))
```
dellaert, line 60: That's awesome! I learned something!

python/gtsam/tests/test_GraphvizFormatting.py
```diff
@@ -0,0 +1,139 @@
...
+        self.graph.add(gtsam.BetweenFactorPose2(1, 2, odometry, odometryNoise))
+
+        self.values = gtsam.Values()
+        self.values.insert_pose2(0, gtsam.Pose2(0., 0., 0.))
+        self.values.insert_pose2(1, gtsam.Pose2(2., 0., 0.))
+        self.values.insert_pose2(2, gtsam.Pose2(4., 0., 0.))
+
+    def test_default(self):
+        """Test with default GraphvizFormatting"""
+        expected_result = """\
```
dellaert, line 42: Do we need the \ ?

dellaert (approved):
This is awesome! I'll trigger CI so please wait on the results before pushing, so CI only has to run once more if any problems pop up there.

python/gtsam/tests/test_GraphvizFormatting.py
senselessDev: Yes, because otherwise the `expected_result` starts with "\n" which does not equal the output of `dot`

python/gtsam/tests/test_GraphvizFormatting.py
senselessDev: Well, I did not write the copyright section ;)

senselessDev:
Thanks for your positive feedback!

One additional thing I just noticed: in the readme of the python directory it is written it should take CMAKE_INSTALL_PREFIX into account. But according to my observation, it does not. I think we have to add `--prefix ${CMAKE_INSTALL_PREFIX}` to the setup.py install line in the `CMakeLists.txt`.

python/gtsam/tests/test_GraphvizFormatting.py
dellaert: You don't need that copyright here, unless you're affiliated with georgia tech. Feel free to remove or replace.

dellaert:
> Thanks for your positive feedback!
> 
> One additional thing I just noticed: in the readme of the python directory it is written it should take CMAKE_INSTALL_PREFIX into account. But according to my observation, it does not. I think we have to add `--prefix ${CMAKE_INSTALL_PREFIX}` to the setup.py install line in the `CMakeLists.txt`.
> 
> Edit: Obviously, I can push that if you say we want that and the CI is done ;)

@varunagrawal comment? 
The comment might be out of date because as far as I know, we now use the default python install locations, which I think is probably appropriate.

senselessDev:
I actually like the option to install it somewhere else. By default, it wants to install to `/usr/local` which requires root privileges. When you specify CMAKE_INSTALL_PREFIX to `~/.local` for example, it usually is also on PATH/PYTHON_PATH and only a per-user installation which does not require root privileges.

python/gtsam/tests/test_GraphvizFormatting.py
senselessDev: No, I'm not affiliated with GT. That will trigger the CI pipeline once again, right? Is that worth it?

varunagrawal: It is and I recommend adding your full name and email so you can be contacted in the future if need be.

varunagrawal:
I need to check but that part of the readme may indeed be out of date.

python/gtsam/tests/test_GraphvizFormatting.py
senselessDev: Won't do, sorry. But changed it nevertheless.

senselessDev:
Changed copyright section and included --prefix option for setup.py. I think the second point only is beneficial and has no drawbacks.

dellaert (changes requested):
Sorry, but the prefix thing has to be a different PR. We can't risk unintended consequences. Please remove from this PR, but we'd love it if you could submit as a separate PR and explain the behavior. I don;t understand well enough what it does.

python/CMakeLists.txt
```diff
@@ -170,7 +170,7 @@ endif()
 # Add custom target so we can install with `make python-install`
 set(GTSAM_PYTHON_INSTALL_TARGET python-install)
 add_custom_target(${GTSAM_PYTHON_INSTALL_TARGET}
-        COMMAND ${PYTHON_EXECUTABLE} ${GTSAM_PYTHON_BUILD_DIRECTORY}/setup.py install
+        COMMAND ${PYTHON_EXECUTABLE} ${GTSAM_PYTHON_BUILD_DIRECTORY}/setup.py install --prefix ${CMAKE_INSTALL_PREFIX}
```
senselessDev, line 173:
```suggestion
        COMMAND ${PYTHON_EXECUTABLE} ${GTSAM_PYTHON_BUILD_DIRECTORY}/setup.py install
```

dellaert (approved):

dellaert:
Thanks !!!!!

PR #1062, opened 2022-01-25
Abstract Conditional for Hybrid

ProfFan:
This PR implements the idea of making `Conditional` base on a concrete `AbstractConditional` so we can using virtual dispatch.

All tests pass except one - hybrid now actually does hybrid, so the full elimination fails, as we didn't implement discrete elimination in `EliminateHybrid`:

```
testHybridFactorGraph
gtsam/hybrid/tests/testHybridFactorGraph.cpp:377: Failure: "Exception: DecisionTree::convertFrom: Invalid NodePtr" 
There were 1 failures

Process finished with exit code 1
```

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -1,0 +1,13 @@
+//
+// Created by Fan Jiang on 1/24/22.
+//
+
+#include <gtsam/hybrid/HybridBayesNet.h>
+
+namespace gtsam {
+
+GaussianMixture::shared_ptr HybridBayesNet::atGaussian(size_t i) {
+  return boost::dynamic_pointer_cast<GaussianMixture>(factors_.at(i));
+}
+
+}
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -9,5 +9,8 @@ namespace gtsam {
 GaussianMixture::shared_ptr HybridBayesNet::atGaussian(size_t i) {
   return boost::dynamic_pointer_cast<GaussianMixture>(factors_.at(i));
 }
+DiscreteConditional::shared_ptr HybridBayesNet::atDiscrete(size_t i) {
+  return boost::dynamic_pointer_cast<DiscreteConditional>(factors_.at(i));
+}
 
 }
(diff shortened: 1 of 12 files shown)


gtsam/hybrid/tests/testHybridFactorGraph.cpp
```diff
@@ -365,13 +365,50 @@ TEST(HybridFactorGraph, Elimination) {
...
+  auto linearizedFactorGraph = self.linearizedFactorGraph;
+
+  // Create ordering.
+  Ordering ordering;
+  for (size_t k = 1; k <= self.K; k++) ordering += X(k);
+  for (size_t k = 1; k < self.K; k++) ordering += M(k);
+
+  // Eliminate partially.
+  HybridBayesNet::shared_ptr hybridBayesNet =
+      linearizedFactorGraph.eliminateSequential(ordering);
```
ProfFan, line 389: Fails here

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+#include <gtsam/hybrid/HybridBayesNet.h>
+
+namespace gtsam {
+
+GaussianMixture::shared_ptr HybridBayesNet::atGaussian(size_t i) {
+  return boost::dynamic_pointer_cast<GaussianMixture>(factors_.at(i));
+}
+
+}
```
dellaert, line 13: return

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -0,0 +1,13 @@
+//
+// Created by Fan Jiang on 1/24/22.
+//
+
+#include <gtsam/hybrid/HybridBayesNet.h>
+
+namespace gtsam {
+
+GaussianMixture::shared_ptr HybridBayesNet::atGaussian(size_t i) {
```
dellaert, line 9: should this not be an abstract conditional ?

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -0,0 +1,47 @@
...
+using symbol_shorthand::X;
+
+static const DiscreteKey Asia(0, 2);
+
+/* ****************************************************************************/
+// Test creation
+TEST(HybridBayesNet, Creation) {
+  HybridBayesNet bayesNet;
+
+  bayesNet.add(Asia, "99/1");
```
dellaert, line 39: Well, add a test and implement :-)

dellaert (approved):

gtsam/hybrid/HybridBayesNet.cpp
ProfFan: This is `atGaussian`, the `at()` does the abstract one

gtsam/hybrid/tests/testHybridBayesNet.cpp
ProfFan: done

PR #1065, opened 2022-01-25
Feature/add discrete hybrid

ProfFan:
Adds a hacky way to deal with discrete that only works when all continuous are eliminated first.

### DIFF
--- a/gtsam/hybrid/GaussianHybridFactorGraph.cpp
+++ b/gtsam/hybrid/GaussianHybridFactorGraph.cpp
@@ -117,10 +117,7 @@ pair<AbstractConditional::shared_ptr, boost::shared_ptr<Factor>> EliminateHybrid
 
   // TODO(fan): Now let's assume that all continuous will be eliminated first!
   // Here sum is null if remaining are all discrete
-  std::cout << "[EL] " << "key = ";
-  ordering.print("");
   if (sum.empty()) {
-    std::cerr << "LOLOLOL you got empty! let's do discrete!\n";
     // Not sure if this is the correct thing, but anyway!
     DiscreteFactorGraph dfg;
     dfg.push_back(factors.discreteGraph());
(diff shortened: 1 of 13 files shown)


gtsam/hybrid/GaussianHybridFactorGraph.cpp
```diff
@@ -115,6 +115,17 @@ pair<GaussianMixture::shared_ptr, boost::shared_ptr<Factor>> EliminateHybrid(
     return hasNull?GaussianFactorGraph():gfg;
   };
 
+  // TODO(fan): Now let's assume that all continuous will be eliminated first!
+  // Here sum is null if remaining are all discrete
+  std::cout << "[EL] " << "key = "; ordering.print("");
+  if (sum.empty()) {
+    std::cerr << "LOLOLOL you got empty! let's do discrete!\n";
+    // Not sure if this is the correct thing, but anyway!
```
dellaert, line 123: 1) should do maxProduct, check that PR and code in DiscreteFactorGraph
2) You need a unit test.

ProfFan: I did maxProduct, hope this is the correct way of doing it?

I also added unit test to check against partial elimination.

gtsam/hybrid/GaussianHybridFactorGraph.cpp
```diff
@@ -115,6 +115,19 @@ pair<GaussianMixture::shared_ptr, boost::shared_ptr<Factor>> EliminateHybrid(
...
   };
 
+  // TODO(fan): Now let's assume that all continuous will be eliminated first!
+  // Here sum is null if remaining are all discrete
+  if (sum.empty()) {
+    // Not sure if this is the correct thing, but anyway!
+    DiscreteFactorGraph dfg;
+    dfg.push_back(factors.discreteGraph());
+
+    auto dbn = dfg.eliminatePartialSequential(ordering, EliminateForMPE);
```
dellaert, line 125: Actually, *within* eliminateHybrid you should probably just call EliminateFirMPE? Eliminate-sequential is the outer elimination loop, whereas Dliminate does one step (possibly of multiple variables at a a time, but we will only test that when we do multi frontal/Bayes trees)

dellaert (approved):
I’ll approve modulo my comment

gtsam/hybrid/GaussianHybridFactorGraph.cpp
ProfFan: I see, I used that initially but that one has a less easy to use return type, so I switched...

PR #1068, opened 2022-01-26
Added inference module

dellaert:
This fixes the fact that the inference module was forgotten when we split up the wrapper :-)


ProfFan (approved):
LGTM

dellaert:
Thanks for the quick review, @ProfFan !

varunagrawal:
I don't understand how this changes anything. It's not like we do `import gtsam.linear` right?

That being said, I am disliking the multiple #include pybind/stl.h. There should just be one and I'll do that next week since this has been on the wrapper agenda long enough :/

varunagrawal:
Is VariableIndex the big update? Cool!

gtsam/inference/inference.i
```diff
@@ -0,0 +1,163 @@
...
+
+// Headers for overloaded methods below, break hierarchy :-/
+#include <gtsam/linear/GaussianFactorGraph.h>
+#include <gtsam/nonlinear/NonlinearFactorGraph.h>
+#include <gtsam/symbolic/SymbolicFactorGraph.h>
+
+class VariableIndex {
+  // Standard Constructors and Named Constructors
+  VariableIndex();
+  // TODO: Templetize constructor when wrap supports it
```
varunagrawal, line 142: FYI wrapper supports constructor templates.

PR #1069, opened 2022-01-26
Wrap FindKarcherMean for Rot3

johnwlambert:
Wrap `gtsam.FindKarcherMean` with `Rot3Vector` argument (a new type defined in Rot3.h)

### DIFF
--- a/gtsam/slam/slam.i
+++ b/gtsam/slam/slam.i
@@ -324,7 +324,7 @@ virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
 };
 
 template <T = {gtsam::Rot3}>
-const T FindKarcherMean<Rot3>(const std::vector<T>& rotations;
+const T FindKarcherMean<Rot3>(const std::vector<T>& rotations);
 
 #include <gtsam/slam/FrobeniusFactor.h>
 gtsam::noiseModel::Isotropic* ConvertNoiseModel(gtsam::noiseModel::Base* model,
--- a/gtsam/slam/slam.i
+++ b/gtsam/slam/slam.i
@@ -331,7 +331,8 @@ class Rot3Vector {
   gtsam::Rot3 at(size_t i) const;
   void push_back(const gtsam::Rot3& R);
 };
-gtsam::Rot3 FindKarcherMean<gtsam::Rot3>(const Rot3Vector& rotations);
+typedef FindKarcherMean<gtsam::Rot3> FindKarcherMeanRot3;
+gtsam::Rot3 FindKarcherMeanRot3(const Rot3Vector& rotations);
 
 #include <gtsam/slam/FrobeniusFactor.h>
 gtsam::noiseModel::Isotropic* ConvertNoiseModel(gtsam::noiseModel::Base* model,
--- a/gtsam/slam/slam.i
+++ b/gtsam/slam/slam.i
@@ -322,9 +322,7 @@ template <T = {gtsam::Point2, gtsam::Rot2, gtsam::Pose2, gtsam::Point3,
 virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
   KarcherMeanFactor(const gtsam::KeyVector& keys);
 };
-
-template <T = {gtsam::Rot3}>
-const T FindKarcherMean<T>(const std::vector<T>& rotations);
+const gtsam::Rot3 FindKarcherMean<gtsam::Rot3>(const std::vector<gtsam::Rot3>& rotations);
 
 #include <gtsam/slam/FrobeniusFactor.h>
 gtsam::noiseModel::Isotropic* ConvertNoiseModel(gtsam::noiseModel::Base* model,
--- a/gtsam/slam/slam.i
+++ b/gtsam/slam/slam.i
@@ -316,12 +316,17 @@ class InitializePose3 {
   static gtsam::Values initialize(const gtsam::NonlinearFactorGraph& graph);
 };
 
-#include <gtsam/slam/KarcherMeanFactor-inl.h>
+#include <gtsam/slam/KarcherMeanFactor.h>
 template <T = {gtsam::Point2, gtsam::Rot2, gtsam::Pose2, gtsam::Point3,
                gtsam::SO3, gtsam::SO4, gtsam::Rot3, gtsam::Pose3}>
 virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
   KarcherMeanFactor(const gtsam::KeyVector& keys);
 };
+  
+template <class T>
+T FindKarcherMean(const std::vector<T, Eigen::aligned_allocator<T>> &rotations);
+
+template <class T> T FindKarcherMean(std::initializer_list<T> &&rotations);
   
 class Rot3Vector {
   Rot3Vector();
--- a/gtsam/slam/slam.i
+++ b/gtsam/slam/slam.i
@@ -325,12 +325,13 @@ virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
   
 class Rot3Vector {
   Rot3Vector();
+  size_t size() const;
 
   // structure specific methods
   gtsam::Rot3 at(size_t i) const;
   void push_back(const gtsam::Rot3& R);
 };
-const gtsam::Rot3 FindKarcherMean<gtsam::Rot3>(const Rot3Vector& rotations);
+gtsam::Rot3 FindKarcherMean<gtsam::Rot3>(const Rot3Vector& rotations);
 
 #include <gtsam/slam/FrobeniusFactor.h>
 gtsam::noiseModel::Isotropic* ConvertNoiseModel(gtsam::noiseModel::Base* model,
--- a/gtsam/slam/slam.i
+++ b/gtsam/slam/slam.i
@@ -316,24 +316,14 @@ class InitializePose3 {
   static gtsam::Values initialize(const gtsam::NonlinearFactorGraph& graph);
 };
 
-#include <gtsam/slam/KarcherMeanFactor.h>
+#include <gtsam/slam/KarcherMeanFactor-inl.h>
 template <T = {gtsam::Point2, gtsam::Rot2, gtsam::Pose2, gtsam::Point3,
                gtsam::SO3, gtsam::SO4, gtsam::Rot3, gtsam::Pose3}>
 virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
   KarcherMeanFactor(const gtsam::KeyVector& keys);
 };
 
-class Rot3Vector {
-  Rot3Vector();
-  size_t size() const;
-
-  // structure specific methods
-  gtsam::Rot3 at(size_t i) const;
-  void push_back(const gtsam::Rot3& R);
-};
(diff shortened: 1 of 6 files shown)


gtsam/slam/slam.i
```diff
@@ -323,6 +323,9 @@ virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
   KarcherMeanFactor(const gtsam::KeyVector& keys);
 };
 
+template <T = {gtsam::Rot3}>
+const T FindKarcherMean<T>(const std::vector<T>& rotations);
```
johnwlambert, line 327: @varunagrawal @ProfFan question for you all -- do we have support for std::vector yet? or do we need to make a new type alias here such as Rot3Vector?

gtsam/slam/slam.i
```diff
@@ -322,6 +322,15 @@ template <T = {gtsam::Point2, gtsam::Rot2, gtsam::Pose2, gtsam::Point3,
 virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
   KarcherMeanFactor(const gtsam::KeyVector& keys);
 };
+  
+class Rot3Vector {
+  Rot3Vector();
+
+  // structure specific methods
+  gtsam::Rot3 at(size_t i) const;
+  void push_back(const gtsam::Rot3& R);
```
akshay-krishnan, line 331: could you add the size() method? `size_t size() const;`

gtsam/slam/slam.i
```diff
@@ -322,6 +322,15 @@ template <T = {gtsam::Point2, gtsam::Rot2, gtsam::Pose2, gtsam::Point3,
...
+};
+const gtsam::Rot3 FindKarcherMean<gtsam::Rot3>(const Rot3Vector& rotations);
```
akshay-krishnan, line 333: no need for `const ` in the return type: `gtsam::Rot3 FindKarcherMean<gtsam::Rot3>(const Rot3Vector& rotations);`

gtsam/slam/slam.i
ProfFan: No because of MATLAB end

python/gtsam/specializations/slam.h
```diff
@@ -17,3 +17,4 @@ py::bind_vector<
 py::bind_vector<
     std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose2> > > >(
     m_, "BetweenFactorPose2s");
+py::bind_vector<std::vector<gtsam::Rot3>>(m_, "Rot3Vector");
```
ProfFan, line 20: Are you sure you don't need aligned allocator?

ProfFan (approved):
LGTM minus one

python/gtsam/specializations/slam.h
johnwlambert: Thanks for the review @ProfFan and @akshay-krishnan.

Let me check about aligned allocator. Btw I don't love the name `Rot3Vector` so much ... sounds like a vector representation of a Rot3. But `Rot3Set` would not be consistent with `Pose3Vector` nomenclature. So I guess it works.

gtsam/slam/slam.i
dellaert: huh? I thought that std::vector worked with wrapper??

gtsam/slam/slam.i
johnwlambert: sure, i added it

gtsam/slam/slam.i
```diff
@@ -322,6 +322,17 @@ template <T = {gtsam::Point2, gtsam::Rot2, gtsam::Pose2, gtsam::Point3,
...
+class Rot3Vector {
+  Rot3Vector();
+  size_t size() const;
+
+  // structure specific methods
+  gtsam::Rot3 at(size_t i) const;
+  void push_back(const gtsam::Rot3& R);
+};
+typedef FindKarcherMean<gtsam::Rot3> FindKarcherMeanRot3;
+gtsam::Rot3 FindKarcherMeanRot3(const Rot3Vector& rotations);
```
johnwlambert, line 335: @dellaert @ProfFan the CI states: `Cannot find class FindKarcherMean in module!`. Strange. Should we be including `#include <gtsam/slam/KarcherMeanFactor.h>` instead of `#include <gtsam/slam/KarcherMeanFactor-inl.h>` in this `.i` file?

```
  File "/home/runner/work/gtsam/gtsam/wrap/scripts/pybind_wrap.py", line 94, in <module>
    main()
  File "/home/runner/work/gtsam/gtsam/wrap/scripts/pybind_wrap.py", line 85, in main
    wrapper.wrap_submodule(args.src)
  File "/home/runner/work/gtsam/gtsam/wrap/gtwrap/pybind_wrapper.py", line 696, in wrap_submodule
    cc_content = self.wrap_file(content, module_name=module_name)
  File "/home/runner/work/gtsam/gtsam/wrap/gtwrap/pybind_wrapper.py", line 626, in wrap_file
    module = instantiator.instantiate_namespace(module)
  File "/home/runner/work/gtsam/gtsam/wrap/gtwrap/template_instantiator/namespace.py", line 80, in instantiate_namespace
    element = instantiate_namespace(element)
  File "/home/runner/work/gtsam/gtsam/wrap/gtwrap/template_instantiator/namespace.py", line 59, in instantiate_namespace
    typedef_inst.typename)
  File "/home/runner/work/gtsam/gtsam/wrap/gtwrap/interface_parser/namespace.py", line 109, in find_class_or_function
    typename.name))
ValueError: Cannot find class FindKarcherMean in module!
make[2]: *** [python/slam.cpp] Error 1
python/CMakeFiles/pybind_wrap_gtsam.dir/build.make:164: recipe for target 'python/slam.cpp' failed
CMakeFiles/Makefile2:6769: recipe for target 'python/CMakeFiles/pybind_wrap_gtsam.dir/all' failed
make[1]: *** [python/CMakeFiles/pybind_wrap_gtsam.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

dellaert: But you *can* use `std::`. Who is telling you you can't ?

ProfFan: Matlab wrapper needs correct signature, unless @varunagrawal implemented something that handles this?

varunagrawal: If I'm not mistaken, std should be useable. When in doubt, check out or write a test for it in the wrap project. :)

(I spent a whole month writing those tests folks, please make them worth my time 😅)

gtsam/slam/slam.i
```diff
@@ -316,13 +316,18 @@ class InitializePose3 {
...
 template <T = {gtsam::Point2, gtsam::Rot2, gtsam::Pose2, gtsam::Point3,
                gtsam::SO3, gtsam::SO4, gtsam::Rot3, gtsam::Pose3}>
 virtual class KarcherMeanFactor : gtsam::NonlinearFactor {
   KarcherMeanFactor(const gtsam::KeyVector& keys);
 };
   
+template <class T>
+T FindKarcherMean(const std::vector<T, Eigen::aligned_allocator<T>> &rotations);
+
+template <class T> T FindKarcherMean(std::initializer_list<T> &&rotations);
```
dellaert, line 329: Not sure what this would translate to in python.....

gtsam/slam/slam.i
dellaert: > Matlab wrapper needs correct signature, unless @varunagrawal implemented something that handles this?

I don't know what that means. Example?

dellaert:
@johnwlambert , are you planning on still submitting other commits? You should know that on every push the CI runs 34 checks. It is recommended you do everything locally, including make check and make python-install, and then push once you think everything was addressed.

dellaert:
If you are planning another push, kindly cancel the CI jobs for this PR.

johnwlambert:
> If you are planning another push, kindly cancel the CI jobs for this PR.

I've gotten to like the Github.com interface a bit too much : - ) I've cancelled the nonrelevant CI jobs, thanks.

dellaert:
Yeah, please don't develop with our CI as your compute :-) It clogs all other PRs, and I'm trying to build a wheel (which will unfortunately not include this PR).

python/gtsam/specializations/slam.h
```diff
@@ -17,3 +17,4 @@ py::bind_vector<
 py::bind_vector<
     std::vector<boost::shared_ptr<gtsam::BetweenFactor<gtsam::Pose2> > > >(
     m_, "BetweenFactorPose2s");
+py::bind_vector<std::vector<gtsam::Rot3, Eigen::aligned_allocator<gtsam::Rot3>>(m_, "Rot3Vector");
```
akshay-krishnan, line 20: there is a missing ">" at the end here, that should let you wrap Rot3Vector at least.

akshay-krishnan:
It looks like we cannot use `std::initializer_list` or `Eigen::aligned_allocator` in slam.i. 
So the only way I can think of is to have a typedef for `std::vector<Rot3, Eigen::aligned_allocator>` and define another function specifically for Rot3 (using the typedef as the argument). 
This seems a little hacky.  
Alternatively, we could just re-implement `FindKarcherMean ` in python using the `KarcherMeanFactor` which has been wrapped. Its a really small function. @johnwlambert WDYT?

johnwlambert:
Thanks for taking a look @akshay-krishnan. Yeah this seems like we should just use KarcherMeanFactor directly in Python w/ an optimizer of our choice.

dellaert:
Nonono! Let's fix this. Let me take a quick look.

dellaert:
@johnwlambert what were you actually trying to accomplish with this PR? There is no PR comment. And, there is `test_KarcherMeanFactor.py` that just works, I think?

johnwlambert:
> 

Thanks Frank. Yeah we were just trying to avoid repeating that boilerplate from https://github.com/borglab/gtsam/blob/develop/python/gtsam/tests/test_KarcherMeanFactor.py#L24. But we can use it in Python, no problem. A simple unit test would look like 

```python
import gtsam
from gtsam import Rot3, Rot3Vector
from gtsam.utils.test_case import GtsamTestCase


class TestKarcherMean(GtsamTestCase):
    """Tests for DecisionTreeFactors."""

    def test_find_karcher_mean_identity() -> None:
        """Averaging 3 identity rotations should yield the identity."""
        a1Rb1 = Rot3()
        a2Rb2 = Rot3()
        a3Rb3 = Rot3()

        aRb_list = Rot3Vector([a1Rb1, a2Rb2, a3Rb3])
        aRb_expected = Rot3()

        aRb = gtsam.FindKarcherMeanRot3(aRb_list)
        self.assertTrue(aRb.equals(aRb_expected))
```

but there are already some implemented [here](https://github.com/borglab/gtsam/blob/develop/python/gtsam/tests/test_KarcherMeanFactor.py#L51), which seem fine.

dellaert:
@johnwlambert I think I fixed it, could you please update the PR with a good description after looking at the file differences?

johnwlambert:
> @johnwlambert I think I fixed it, could you please update the PR with a good description after looking at the file differences?

Beautiful work Frank. I added a bit to the description you wrote. Thanks!

PR #1070, opened 2022-01-27
Added support for positioning in BayesNet plotting

dellaert:
- added a field `variablePositions` in DotWriter
- made method in `BayesNet` use them, if given
- derived `SymbolicBayesNet` and `GaussianBayesNet` from `BayesNet` so they have this as well
- removed obsolete code as a consequence
- some slight reformatting here and there

This supports graphs in the book that look like this:

image

### DIFF
--- a/gtsam/inference/BayesNet-inst.h
+++ b/gtsam/inference/BayesNet-inst.h
@@ -10,16 +10,16 @@
 * -------------------------------------------------------------------------- */
 
 /**
-* @file    BayesNet.h
-* @brief   Bayes network
-* @author  Frank Dellaert
-* @author  Richard Roberts
-*/
+ * @file    BayesNet.h
+ * @brief   Bayes network
+ * @author  Frank Dellaert
+ * @author  Richard Roberts
+ */
 
 #pragma once
 
-#include <gtsam/inference/FactorGraph-inst.h>
 #include <gtsam/inference/BayesNet.h>
+#include <gtsam/inference/FactorGraph-inst.h>
 
 #include <boost/range/adaptor/reversed.hpp>
 #include <fstream>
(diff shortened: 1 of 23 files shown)


gtsam/inference/BayesNet-inst.h
```diff
@@ -10,41 +10,50 @@
...
+
+  // Create nodes for each variable in the graph
+  for (Key key : this->keys()) {
+    auto position = writer.variablePos(key);
+    writer.drawVariable(key, keyFormatter, position, &os);
+  }
+  os << "\n";
 
-  for (auto conditional : *this) {
+  for (auto conditional : boost::adaptors::reverse(*this)) {
```
yetongumich, line 51: Maybe a comment here why reversing the order? Like following a top-down order in printing?

gtsam/inference/DotWriter.h
```diff
@@ -35,6 +40,22 @@ struct GTSAM_EXPORT DotWriter {
...
   bool binaryEdges;          ///< just use non-dotted edges for binary factors
 
+  /**
+   * Variable positions can be optionally specified and will be included in the
+   * dor file with a "!' sign, so "neato" can use it to render them.
+   */
+  std::map<Key, Vector2> variablePositions;
+
+  /**
+   * The position hints allow one to use symbol character and index to specify
```
yetongumich, line 50: I like this functionality

yetongumich: Do we want to specify factor positions as well?

yetongumich (approved):
LGTM with minor comments

gtsam/inference/BayesNet-inst.h
dellaert: It is because Bayes nets are typically specified in reverse topological sort order, that's how they come out of elimination. Will add comments in the next PR to not restart CI for this one thing

gtsam/inference/DotWriter.h
dellaert: Yeah, forgot about that. will add. Currently in graphVizFormatting, actually, but I need it here.

PR #1073, opened 2022-01-28
Release/4.2a4

dellaert:


johnwlambert (approved):
LGTM. Why are we using `master` instead of `develop` though?

dellaert:
Thanks John. There are *two* PRs, one into master and one into develop. Please approve both :-)
See https://gtsam.org/notes/Contributing.html

johnwlambert:
> Thanks John. There are _two_ PRs, one into master and one into develop. Please approve both :-) See https://gtsam.org/notes/Contributing.html

Got it, both are approved : - )

ProfFan:
Release branches should be kept as is @dellaert

dellaert:
I don't think so... Maybe discuss with me first?

dellaert:
Gitflow convention is to tag and delete the branch, AFAIK.
But I see that we need it for manylinux build :-) That's a different issue, and glad you restored it.

PR #1077, opened 2022-01-31
Wrapped more things

dellaert:
I wrapped a number of methods that were needed when writing about SLAM. Cleaned up a few tiny bits here and there as I was wrapping. Includes:
- loading
- elimination
- printing and inspecting Bayes tree/ Bayes net


dellaert:
@varunagrawal I know CI is broken and already fixed, but will await your review to also push any other comments you have so I don't trash CI again...

python/gtsam/tests/test_GaussianBayesNet.py
```diff
@@ -0,0 +1,53 @@
+"""
+GTSAM Copyright 2010-2019, Georgia Tech Research Corporation,
+Atlanta, Georgia 30332-0415
+All Rights Reserved
+
+See LICENSE for the license information
+
+Unit tests for Linear Factor Graphs.
```
varunagrawal, line 8: update docstring?

varunagrawal (approved):
LGTM sans one comment

PR #1082, opened 2022-02-01
Cleaned up loading SfmData

dellaert:
- made static constructors
- deprecated functions using a non-const reference
- added an example in Python with Balbianello:
image

### DIFF
--- a/gtsam/sfm/SfmData.cpp
+++ b/gtsam/sfm/SfmData.cpp
@@ -18,6 +18,7 @@
 
 #include <gtsam/inference/Symbol.h>
 #include <gtsam/sfm/SfmData.h>
+#include <gtsam/slam/GeneralSFMFactor.h>
 
 #include <fstream>
 #include <iostream>
@@ -99,13 +100,15 @@ Pose3 gtsam2openGL(const Pose3 &PoseGTSAM) {
 }
 
 /* ************************************************************************** */
-bool readBundler(const std::string &filename, SfmData &data) {
+SfmData SfmData::FromBundlerFile(const std::string &filename) {
   // Load the data file
   std::ifstream is(filename.c_str(), std::ifstream::in);
   if (!is) {
-    cout << "Error in readBundler: can not find the file!!" << endl;
-    return false;
+    throw std::runtime_error(
+        "Error in FromBundlerFile: can not find the file!!");
   }
+
+  SfmData sfmData;
 
   // Ignore the first line
   char aux[500];
@@ -133,9 +136,8 @@ bool readBundler(const std::string &filename, SfmData &data) {
 
     // Check for all-zero R, in which case quit
     if (r11 == 0 && r12 == 0 && r13 == 0) {
-      cout << "Error in readBundler: zero rotation matrix for pose " << i
-           << endl;
-      return false;
+      throw std::runtime_error(
+          "Error in FromBundlerFile: zero rotation matrix");
     }
 
     // Get the translation vector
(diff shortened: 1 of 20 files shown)


gtsam/sfm/SfmData.h
```diff
@@ -61,6 +82,31 @@ struct SfmData {
...
+   *
+   * @param model a noise model for projection errors
+   * @param fixedCamera which camera to fix, if any (use boost::none if none)
+   * @param fixedPoint which point to fix, if any (use boost::none if none)
+   * @return NonlinearFactorGraph
+   */
+  NonlinearFactorGraph sfmFactorGraph(
+      const SharedNoiseModel& model = noiseModel::Isotropic::Sigma(2, 1.0),
+      boost::optional<size_t> fixedCamera = 0,
+      boost::optional<size_t> fixedPoint = 0) const;
```
akshay-krishnan, line 108: Nit, feel free to ignore if merging is urgent: 
I think it is unclear what the difference between generalSfmFactors and sfmFactorGraph is to a user from the name itself. Maybe sfmFactorGraphWithPriors is a better name?

akshay-krishnan (approved):
Looks good! I like the new static read methods, avoids quite some boilerplate.

dellaert:
Thanks! The exception does not need to be flagged, it's just a deprecated method and making sure it returns a bool - that's a pretty useless behavior, that's also why deprecated.

gtsam/sfm/SfmData.cpp
```diff
@@ -178,34 +179,34 @@ bool readBundler(const std::string &filename, SfmData &data) {
       track.siftIndices.emplace_back(cam_idx, point_idx);
     }
 
-    data.tracks.push_back(track);
+    sfmData.tracks.push_back(track);
   }
 
-  is.close();
```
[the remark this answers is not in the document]
dellaert: Ah, no, a stream closes when it is destroyed :-)

gtsam/sfm/SfmData.h
dellaert: generalSfmFactors is sorta internal, so maybe I will move that to a different {} group

python/gtsam/utils/visual_data_generator.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, K=Cal3_S2(), nrCameras: int = 3, nrPoints: int = 4) -> None:
         self.cameras = [Pose3()] * nrCameras
         self.points = [Point3(0, 0, 0)] * nrPoints
 
-    def print_(self, s="") -> None:
+    def print(self, s="") -> None:
```
johnwlambert, line 39: Looks great. nit: type hint missing here for `s`

python/gtsam/tests/test_SfmData.py
```diff
@@ -56,24 +62,58 @@ def test_data(self):
...
         self.data.addTrack(track2)
+
         # Number of tracks in SfmData is 2
         self.assertEqual(self.data.numberTracks(), 2)
+        
         # camera idx of first measurement of second track corresponds to i1
         cam_idx, img_measurement = self.data.track(1).measurement(0)
         self.assertEqual(cam_idx, i1)
 
+    def test_Balbianello(self):
```
johnwlambert, line 88: nit: would be nice to add a 1-liner sentence explaining what the test is striving for

python/gtsam/tests/test_SfmData.py
```diff
@@ -17,6 +17,7 @@
 import numpy as np
 
 import gtsam
+from gtsam import SfmData, SfmTrack, Point2, Point3
```
johnwlambert, line 20: nit: sorted alphabetically, these should be 
```python
from gtsam import Point2, Point3, SfmData, SfmTrack
```

johnwlambert (approved):
LGTM, though a few minor python syntax changes could be made

dellaert:
Okidoke!

dellaert:
PS I know why the CI fails, will fix that as well.

python/gtsam/tests/test_SfmData.py
dellaert: Done

python/gtsam/utils/visual_data_generator.py
dellaert: Done

PR #1083, opened 2022-02-02
Fixes to get matlab wrapper working properly

varunagrawal:
Just the title. Recent churn and lack of Matlab CI allowed some bugs to creep in so just solving those bugs.

### DIFF
--- a/python/gtsam/preamble/inference.h
+++ b/python/gtsam/preamble/inference.h
@@ -10,6 +10,3 @@
  * Without this they will be automatically converted to a Python object, and all
  * mutations on Python side will not be reflected on C++.
  */
-
-#include <pybind11/stl.h>
-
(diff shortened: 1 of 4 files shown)


dellaert (approved):
Thanks!!! I guess these came to light when testing MATLAB?

varunagrawal:
Yes

dellaert:
@mattking-smith FYI

python/gtsam/preamble/inference.h
```diff
@@ -10,6 +10,3 @@
  * Without this they will be automatically converted to a Python object, and all
  * mutations on Python side will not be reflected on C++.
  */
-
-#include <pybind11/stl.h>
```
dellaert, line 14: @varunagrawal : why was this removed? I think this broke DotWriter

PR #1086, opened 2022-02-04
Add default constructor to PoseRotationPrior to fix serialization

ScottMcMichael:
This matches the existing code here:  https://github.com/borglab/gtsam/blob/develop/gtsam/slam/PoseTranslationPrior.h


dellaert (approved):
Thanks !!!

PR #1089, opened 2022-02-05
Fix DotWriter wrapping

dellaert:
@varunagrawal , [this commit](https://github.com/borglab/gtsam/pull/1083/commits/134e82836e0ff19cd0a2cfeb91d91f3502d74b10) removed the line

```
#include <pybind11/stl.h>
```

which was needed to access and update the fields in `gtsam.DotWriter`, which I need in [gtbook](https://gtbook.github.io/gtbook/). Admittedly, there was no test in gtsam that showed this line is crucial. This PR fixes that by adding a test and adding the line back in.

PS, this was a pain to debug as the wrapper cmake scripts do *not* copy over files in preamble and specializations after they have been touched. I had to manually update the copy in the build folder. Maybe something to fix in wrap after IROS deadline.

### DIFF
--- a/gtsam/discrete/tests/testDiscreteBayesNet.cpp
+++ b/gtsam/discrete/tests/testDiscreteBayesNet.cpp
@@ -150,7 +150,6 @@ TEST(DiscreteBayesNet, Dot) {
   fragment.add((Either | Tuberculosis, LungCancer) = "F T T T");
 
   string actual = fragment.dot();
-  cout << actual << endl;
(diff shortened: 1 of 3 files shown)


dellaert:
No idea what is going on with CI here...

varunagrawal (approved):
LGTM. Sorry about that, didn't know it would have this big an effect on another application.

varunagrawal:
I'm merging this.

PR #1090, opened 2022-02-06
Get Sim2 working

varunagrawal:
Fixed all the pending stuff in Sim2, plus also did some refactor of Sim3 to reflect comments in Sim2's original PR.


gtsam/geometry/Similarity2.cpp
```diff
@@ -15,21 +15,21 @@
...
+#include <gtsam/geometry/Rot3.h>
+#include <gtsam/geometry/Similarity2.h>
 #include <gtsam/slam/KarcherMeanFactor-inl.h>
 
 namespace gtsam {
 
 using std::vector;
 
-namespace {
+namespace internal {
```
johnwlambert, line 28: thanks for the PR @varunagrawal. Appreciate you taking the time to help out here.

Was curious the rationale for the `internal` namespace (why its needed and the naming)

varunagrawal: I followed Frank's suggestion from your PR.

dellaert: I am assuming @varunagrawal this is because you need Sim3 for your A-> B :-)

varunagrawal: Yup exactly :)

varunagrawal:
@johnwlambert I merged in develop since some tests were failing because of that. It also seems that `test_Sim2.py` has a lot of tests failing because of the semantics not making complete sense. Can you look into that? It should be straightforward to run and test from this branch.

varunagrawal:
Okay I resolved all the tests :tada:

python/gtsam/tests/test_Sim2.py
```diff
@@ -27,10 +27,10 @@ def test_align_poses_along_straight_line(self) -> None:
         Scenario:
            3 object poses
            same scale (no gauge ambiguity)
-           world frame has poses rotated about x-axis (90 degree roll)
+           world frame has poses rotated about 180 degrees.
            world and egovehicle frame translated by 15 meters w.r.t. each other
         """
-        Rx90 = Rot2.fromDegrees(90)
+        R180 = Rot2.fromDegrees(180)
```
johnwlambert, line 33: @varunagrawal everything looks great. Was curious why you changed this value from 90 to 180, though. Both are valid, as I recall, for the purpose of the test.

johnwlambert (approved):
Really nice. LGTM.

python/gtsam/tests/test_Sim2.py
varunagrawal: The translation values don't make sense otherwise. If we rotate by 90 degrees one of the two trajectories should change in the y axis.

PR #1091, opened 2022-02-07
Improvements in linear

dellaert:
Improvements in linear:
- FromMeanAndStddev constructors
- sampling using std random number generators
- wrapping of above
- documentation improvements

### DIFF
--- a/gtsam/linear/GaussianBayesNet.h
+++ b/gtsam/linear/GaussianBayesNet.h
@@ -95,10 +95,27 @@ namespace gtsam {
     /// Version of optimize for incomplete BayesNet, given missing variables
     VectorValues optimize(const VectorValues given) const;
 
-    /// Sample using ancestral sampling
+    /**
+     * Sample using ancestral sampling
+     * Example:
+     *   std::mt19937_64 rng(42);
+     *   auto sample = gbn.sample(&rng);
+     */
+    VectorValues sample(std::mt19937_64* rng) const;
+
+    /**
+     * Sample from an incomplete BayesNet, given missing variables
+     * Example:
+     *   std::mt19937_64 rng(42);
+     *   VectorValues given = ...;
+     *   auto sample = gbn.sample(given, &rng);
+     */
+    VectorValues sample(VectorValues given, std::mt19937_64* rng) const;
+
+    /// Sample using ancestral sampling, use default rng
     VectorValues sample() const;
 
-    /// Sample from an incomplete BayesNet, given missing variables
+    /// Sample from an incomplete BayesNet, use default rng
     VectorValues sample(VectorValues given) const;
 
     /**
--- a/gtsam/linear/GaussianBayesNet.h
+++ b/gtsam/linear/GaussianBayesNet.h
@@ -88,11 +88,18 @@ namespace gtsam {
     /// @name Standard Interface
     /// @{
 
-    /// Solve the GaussianBayesNet, i.e. return \f$ x = R^{-1}*d \f$, by back-substitution
+    /// Solve the GaussianBayesNet, i.e. return \f$ x = R^{-1}*d \f$, by
+    /// back-substitution
     VectorValues optimize() const;
 
-    /// Version of optimize for incomplete BayesNet, needs solution for missing variables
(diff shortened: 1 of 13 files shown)


gtsam/linear/GaussianBayesNet.h
```diff
@@ -88,11 +88,35 @@ namespace gtsam {
...
+    VectorValues sample(std::mt19937_64* rng) const;
+
+    /**
+     * Sample from an incomplete BayesNet, given missing variables
+     * Example:
+     *   std::mt19937_64 rng(42);
+     *   VectorValues given = ...;
+     *   auto sample = gbn.sample(given, &rng);
+     */
+    VectorValues sample(VectorValues given, std::mt19937_64* rng) const;
```
varunagrawal, line 113: Set `rng` defaults here and above so we can remove the two methods below?

gtsam/linear/GaussianConditional.cpp
```diff
@@ -192,7 +224,44 @@ namespace gtsam {
...
+  VectorValues GaussianConditional::sample(const VectorValues& parentsValues,
+                                           std::mt19937_64* rng) const {
+    if (nrFrontals() != 1) {
+      throw std::invalid_argument(
+          "GaussianConditional::sample can only be called on single variable "
+          "conditionals");
+    }
+    if (!model_) {
+      throw std::invalid_argument(
+          "GaussianConditional::sample can only be called if a diagonal noise "
```
varunagrawal, line 237: This says diagonal noise model, but the check is just for a model existing rather than a diagonal one. So perhaps update the error message?

gtsam/linear/GaussianConditional.h
```diff
@@ -125,10 +151,39 @@ namespace gtsam {
...
+     *   auto sample = gbn.sample(given, &rng);
+     */
+    VectorValues sample(const VectorValues& parentsValues,
+                        std::mt19937_64* rng) const;
+
+    /// Sample, use default rng
+    VectorValues sample() const;
+
+    /// Sample with given values, use default rng
+    VectorValues sample(const VectorValues& parentsValues) const;
```
varunagrawal, line 176: Same deal about default arguments. We can get rid of this and the above sample method by specifying defaults.

gtsam/linear/Sampler.h
```diff
@@ -63,15 +63,9 @@ class GTSAM_EXPORT Sampler {
   /// @name access functions
   /// @{
 
-  size_t dim() const {
-    assert(model_.get());
```
varunagrawal, line 67: Was there a reason this assert was added in the first place? Seems weird to have it from before. Sadly no documentation :(

varunagrawal (approved):
A bunch of comments but LGTM.

gtsam/linear/GaussianBayesNet.h
dellaert: Not sure how, as static default rng is only defined in cpp. I don't want to export it in header.

gtsam/linear/GaussianConditional.cpp
dellaert: GaussianConditional inherits from JacobianFactor which only accepts diagonal models.

gtsam/linear/GaussianConditional.h
dellaert: Same issue :-(

gtsam/linear/GaussianBayesNet.h
varunagrawal: Pass in null and check for that? If null then set to the default rng.

varunagrawal, line 113: Okay, I have a better suggestion. You can change this signature to
```
VectorValues sample(VectorValues given, std::mt19937_64 rng = std::mt19937_64(42)) const;
```
and the definition would have us pass `&rng`. I tried it out locally and this works great!

Con: You will have convert from a pointer to a reference.

dellaert: I'm sticking to my solution, as it uses one static rng in cpp.
Don't spend too much time on this.

PR #1095, opened 2022-02-09
New functionality in linear

dellaert:
- Added a likelihood method in GaussianConditional
- Added an html method in VectorValues:

image

### DIFF
--- a/gtsam/linear/GaussianConditional.cpp
+++ b/gtsam/linear/GaussianConditional.cpp
@@ -225,6 +225,50 @@ namespace gtsam {
   }
 
   /* ************************************************************************ */
+  JacobianFactor::shared_ptr GaussianConditional::likelihood(
+      const VectorValues& frontalValues) const {
+    // Error is |Rx - (d - Sy - Tz - ...)|^2
+    // so when we instantiate x (which has to be completely known) we beget:
+    // |Sy + Tz + ... - (d - Rx)|^2
+    // The noise model just transfers over!
+
+    // Get frontalValues as vector
+    const Vector x =
+        frontalValues.vector(KeyVector(beginFrontals(), endFrontals()));
+
+    // Copy the augmented Jacobian matrix:
+    auto newAb = Ab_;
+
+    // Restrict view to parent blocks
+    newAb.firstBlock() += nrFrontals_;
+
+    // Update right-hand-side (last column)
+    auto last = newAb.matrix().cols() - 1;
+    const auto RR = R().triangularView<Eigen::Upper>();
+    newAb.matrix().col(last) -= RR * x;
+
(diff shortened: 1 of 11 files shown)


gtsam/linear/GaussianConditional.cpp
```diff
@@ -224,6 +224,50 @@ namespace gtsam {
...
+    auto last = newAb.matrix().cols() - 1;
+    const auto RR = R().triangularView<Eigen::Upper>();
+    newAb.matrix().col(last) -= RR * x;
+
+    // The keys now do not include the frontal keys:
+    KeyVector newKeys;
+    newKeys.reserve(nrParents());
+    for (auto&& key : parents()) newKeys.push_back(key);
+
+    // Hopefully second newAb copy below is optimized out...
```
gchenfc, line 255: By this comment, do you mean hopefully the compiler optimizes it out?

Because indeed I _think_ the lines:
`auto newAb = Ab_`
and
`return boost::make_shared<JacobianFactor>(newKeys, newAb, model_);`
are both making copies.

Although I guess since the `JacobianFactor` constructor is copying anyway, I can't think of a functional-style way to avoid the extra copy.

gchenfc (approved):
Looks good to me.  Minor comment about extra copies, but probably not that significant anyway.

gtsam/linear/GaussianConditional.cpp
dellaert: The first one is for sure, but maybe the second one is optimized out? A la RVO?

gchenfc: Hmm I don't _think_ RVO can optimize-out a copy into a function call argument, only for a return value out (i.e. shared_ptr<JacobianFactor> won't have extra copy), but ofc I'm not very knowledgable here.

Though I tried making a quick mock-up:
https://godbolt.org/z/Wze8EW5x3

I think the place the copy is actually happening is the JacobianFactor constructor:
```c++
  JacobianFactor::JacobianFactor(
    const KEYS& keys, const VerticalBlockMatrix& augmentedMatrix, const SharedDiagonal& model) :
  Base(keys), Ab_(augmentedMatrix)
```
attention: `Ab_(matrix)`
The `const VerticalBlockMatrix &` ensures there's not a copy there, and I don't know if the standard allows introspecting that far to check to optimize-out a copy

Well in any case, I don't know the copy can be easily avoided regardless, since `newAb` has to get modified at some point without changing `Ab_`.

ProfFan: Note I think if you do `Obj(m): m(std::move(m)) {}` the move is guaranteed.

dellaert:
Thanks @ProfFan . But how do we say that we want to move this copy?

PR #1097, opened 2022-02-14
EXPORT statements to help Windows build. In Constraint.h, there was a…

mikesheffler:
* In Constraint.h, there was a GTSAM_EXPORT that should have been GTSAM_UNSTABLE_EXPORT, and
* In DiscreteKey.h, there should be a GTSAM_EXPORT in front of an operator definition for a class that is being exported.

This does **not** fix _all_ of the `gtsam_unstable` build issues on MSVC / Windows, but it fixes all of the issues stemming from `gtsam_unstable\Discrete`


dellaert (approved):
Thanks!!

PR #1102, opened 2022-02-15
Fix MSVC / Windows Build of GTSAM_UNSTABLE

mikesheffler:
Per https://github.com/borglab/gtsam/blob/develop/Using-GTSAM-EXPORT.md , classes with no methods defined in a .cpp file shouldn't have the GTSAM_EXPORT or GTSAM_UNSTABLE_EXPORT modifier. This was causing problems with the building of gtsam_unstable on MSVC / Windows.

What I don't like about this PR is that it is reverses several changes @varunagrawal made in #904. I think that the way it is done in this PR is probably correct, but I think @varunagrawal should weigh in, just in case I'm missing something.

This is in service of #1087, but it isn't a resolution, since that issue is about MATLAB. I started this little cul-de-sac as a waypoint on the way to examining the MATLAB build, and I'll take a look at that soon.

Two requests from @dellaert are ignored here:

1) **_Silence some / all of the C4834 warnings about ignored returns from [[nodiscard]] functions:_** Doing this file-by-file with `#pragma warning(disable : 4834)` resulted in, I think, 85 files with changes. I made a patch of that change just in case we _do_ want it, but I thought that changing that many files was probably a sign it was not the right approach.

2) **_Add more actions to the Windows CI:_** There are six targets in `.github/workflows/build-windows.yml`:
```
          cmake --build build -j 4 --config ${{ matrix.build_type }} --target gtsam
          cmake --build build -j 4 --config ${{ matrix.build_type }} --target gtsam_unstable
          cmake --build build -j 4 --config ${{ matrix.build_type }} --target wrap
          cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.base
          cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.base_unstable
          cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.linear
```
 and it turns out that that is all of the the `check`s that _will_ pass right now. I can modify the list to include more targets, but they will fail. I haven't tried to look at them and see _why_ they're failing.

PR #1107, opened 2022-02-16
Fix issues with grouped tests

### LINKED ISSUE #91: GTSAM_SINGLE_TEST_EXE option causes compile errors
## Description
If the GTSAM_SINGLE_TEST_EXE option is enabled, then the unit tests fail to compile.  For example, check.geometry will not compile under either Linux or Windows. It generates a "multiple definition" error

## Steps to reproduce

1. Turn on the GTSAM_SINGLE_TEST_EXE flag in cmake
2. Run `make check.geometry`

## Expected behavior
Expect the unit tests to compile without errors.

## Environment
This occurs in both Windows and Linux.  That said, the GTSAM_SINGLE_TEST_EXE is turned on by default in Windows, but is off by default in Linux.  So, this is far more "noticeable" in Windows.

## Additional information
Two questions really. 
1.  Is there a reason this is turned on by default in Windows? Or a reason for this flag at all?
2.  There are a lot of duplicate tests as you go across the individual programs in "check.geometry" (and lots of other groups of "check" programs).  Is there a reason for these duplicates?  We can go through and make each test individually names, or we could remove some of the duplicates.

dellaert:
Fixes #91 and hopefully addresses some issues in #1102


dellaert:
Hey @mikesheffler, this works now. If you review I can merge, and then you can merge in develop to see the effect on Windows.

varunagrawal (approved):
All we had to do was add namespace guards to all the global variables in the tests??? Sheesh, I'm kinda sad that #91 took this long to resolve. LGTM!!!

mikesheffler:
> Hey @mikesheffler, this works now. If you review I can merge, and then you can merge in develop to see the effect on Windows.

Sorry, I've been tied up all day.

In plain English, what's the goal here? Are we trying to get all of the unit tests working in this build, or just a select subset? I'm fine either way, but if the goal is to get _everything_ going, I wonder if there might be some remaining issues. I pulled and then tried building everything, but I'm still seeing a bunch of build errors. I swear that this isn't sarcastic, but: Am I doing something wrong?

I'm ...

1) Running CMake like I always do
2) Opening the `GTSAM.sln` that pops out
3) Clicking on the little triangle next to `Unit tests` in Solution Explorer
4) Right-click on `check.geometry`
5) Selecting `Build`
6) Seeing a bunch of errors in the console

That process _does_ work for e.g. `check.base`

I thought maybe the GUI was hiding what I wanted, so I ...

1) Opened the `x64 Native Tools Command Prompt for VS 2019`
2) `cd`ed to my GTSAM build directory
3) Used a spell similar to [this workflow file](https://github.com/borglab/gtsam/actions/runs/1856871555/workflow): `cmake --build . -j 4 --config Debug --target check.geometry`
4) Saw a bunch of errors in the console

Like the GUI-based approach, this does work for `check.base`.

I saw in one of the commits above that @dellaert 's commit message was `fix check.sam`, but that's one of the ones that I can't build. This is at 55ad184, btw -- I'm at the current `HEAD` of `develop`.

dellaert:
> Sorry, I've been tied up all day.

I get that :-)

> In plain English, what's the goal here? 

Trying to resolve #91 and #1102, ultimately getting MATLAB and python wrappers working on windows. Borrowing from your little bird analogy, I propose using `make check` as the canary in the coal mine: this would suss out all compile and linking errors in gtsam proper, make our windows CI better, and move on to any issues in building the MATLAB wrapper. Makes sense?

No, I don't think you're doing anything wrong if it works with base. What are the errors with check.geometry? Maybe paste them in #1102 , as we now closed this PR. I'll repeat this comment there.

varunagrawal:
My $0.02: I made some great progress on fixing most of the dllexport issues, and only have geometry, nonlinear, sam and slam not passing.

Except for geometry, every other failure is test failure which should be easier to debug than the dllexport issues. I'll create a PR so y'all can take a look.

PR #1108, opened 2022-02-17
Header discipline in base

dellaert:
Trying to cut down on compile times, so
- moved serialization of Vector and Matrix into separate headers, sparsely included
- moved various other includes out of base header files
- fixed issues related to those

### DIFF
--- a/gtsam/base/Matrix.cpp
+++ b/gtsam/base/Matrix.cpp
@@ -25,6 +25,7 @@
 
 #include <boost/tuple/tuple.hpp>
 #include <boost/tokenizer.hpp>
+#include <boost/format.hpp>
 
 #include <cstdarg>
 #include <cstring>
--- a/gtsam/base/Matrix.cpp
+++ b/gtsam/base/Matrix.cpp
@@ -35,7 +35,6 @@
 #include <limits>
 #include <iostream>
 #include <iostream>
-#include <vector>
 
 using namespace std;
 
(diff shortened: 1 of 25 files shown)


gtsam/base/Matrix.cpp
```diff
@@ -33,6 +34,7 @@
 #include <fstream>
 #include <limits>
 #include <iostream>
+#include <iostream>
```
varunagrawal, line 37: duplicate

dellaert: If you reviewed everything and you trust me to fix the above, approve??

varunagrawal (commented):
There is this cool tool called [include-what-you-use](https://include-what-you-use.org/) which we should probably leverage.

gtsam_unstable/slam/tests/testSerialization.cpp
```diff
@@ -18,12 +15,16 @@
...
-#include <sstream>
+#include <CppUnitLite/TestHarness.h>
+
 #include <boost/assign/std/vector.hpp>
 #include <boost/filesystem.hpp>
 
+#include <iostream>
+#include <cstdlib>
+#include <fstream>
+#include <sstream>
```
varunagrawal, line 26: Pretty sure fstream includes iostream for us, but I like being explicit so yay!

varunagrawal (approved):
Took a deeper look and LGTM

dellaert:
> There is this cool tool called [include-what-you-use](https://include-what-you-use.org/) which we should probably leverage.

Yeah, tried that at one point. Its not so great for large repos with boost and Eigen. Maybe we should try it again.

PR #1112, opened 2022-02-19
Add missing header for timing target

dellaert:
that's it :-)

### DIFF
--- a/gtsam/basis/Chebyshev.h
+++ b/gtsam/basis/Chebyshev.h
@@ -21,8 +21,6 @@
 #include <gtsam/base/Manifold.h>
 #include <gtsam/basis/Basis.h>
 
-#include <unsupported/Eigen/KroneckerProduct>
-
 namespace gtsam {
 
 /**
--- a/timing/timeSFMBAL.h
+++ b/timing/timeSFMBAL.h
@@ -17,6 +17,7 @@
  */
 
 #include <gtsam/sfm/SfmData.h>
+#include <gtsam/slam/dataset.h>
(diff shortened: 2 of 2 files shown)


dellaert:
Tagging @ayushbaid as reviewer so he knows timing code exists in GTSAM :-)

dellaert:
@varunagrawal added you on review as well because I also removed a forgotten obsolete header that prevents double expresso from working for me. It was not needed anymore as you recall.

gtsam/basis/Chebyshev.h
```diff
@@ -21,8 +21,6 @@
 #include <gtsam/base/Manifold.h>
 #include <gtsam/basis/Basis.h>
 
-#include <unsupported/Eigen/KroneckerProduct>
```
varunagrawal, line 24: Whoops sorry about that. But great catch!

varunagrawal (approved):
LGTM

varunagrawal:
Failure is because of Azure crapping out. Just need to re-run it again.

dellaert:
OK, thanks. I'll restart that CI run.

dellaert:
Although, that would restart a bunch. I'll just ignore that failure if all others pass.

ayushbaid:
I did not get this. Does including the header file means we will automatically start timing all the functions in that file?

dellaert:
> I did not get this. Does including the header file means we will automatically start timing all the functions in that file?

No, absolutely not. There are a number of timing scripts in GTSAM, but you have to enable timing with the mechanisms described in the header.

PR #1116, opened 2022-02-21
Include guards

acxz:
This PR adds missing include guards via `#pragma once` to various headers which are missing them.


acxz:
The header files under `python/` and `wrap/` still have missing guards. I wasn't sure whether to add them or not. Let me know if we should and I'll add them rq.

dellaert (approved):

varunagrawal:
`python` headers don't have guards since they serve a difference purpose. Though I guess adding guards would help, we would have to ensure the wrapping process doesn't break.

If you want to include guards in `wrap`, then do so [here](https://github.com/borglab/wrap).

PR #1117, opened 2022-02-21
Allow more functionality through system METIS

acxz:
This is a follow up PR to #570 as when it was merged I was not available to provide my review.

There is no reason to specify the full local `metis.h` include path, with how metis is treated in cmake, `metis.h` works for both local and system metis.

See: https://github.com/borglab/gtsam/pull/570/files#diff-0106693c254359c8aceceb57d5b6c0643760cc8cce6760ccfd2d31bc6d0be2dbR26
where only metis.h is used without the ifdef.

More importantly this PR enables metis's internal api ("metislib.h") via extern C for SYSTEM_METIS. We already use the internal API for our local metis install and since we carry the header files of metis with us I do not see a problem in enabling `extern "C"` for system metis as well. The only problem that can occur with this is a desync between the local headers we carry for metis and what metis is compiled under at the system level. Considering metis is a very old and stable package that hasn't seen updates for a long time, I do not see this issue occurring and if it does should be easily fixable, by just updating our local metis copy.

Tagging @jlblancoc and @varunagrawal  for reviews. Also @berndpfrommer ig

### DIFF
--- a/cmake/HandleMetis.cmake
+++ b/cmake/HandleMetis.cmake
@@ -22,7 +22,7 @@ if(GTSAM_USE_SYSTEM_METIS)
 
     add_library(metis-gtsam-if INTERFACE)
     target_include_directories(metis-gtsam-if BEFORE INTERFACE ${METIS_INCLUDE_DIR}
-      # gtsam_unstable/partition/FindSeparator-inl.h uses internel metislib.h API
+      # gtsam_unstable/partition/FindSeparator-inl.h uses internal metislib.h API
       # via extern "C"
       $<BUILD_INTERFACE:${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis/libmetis>
       $<BUILD_INTERFACE:${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis/GKlib>
@@ -37,7 +37,7 @@ else()
   target_include_directories(metis-gtsam BEFORE PUBLIC
     $<INSTALL_INTERFACE:include/gtsam/3rdparty/metis/>
     $<BUILD_INTERFACE:${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis/include>
-    # gtsam_unstable/partition/FindSeparator-inl.h uses internel metislib.h API
+    # gtsam_unstable/partition/FindSeparator-inl.h uses internal metislib.h API
     # via extern "C"
     $<BUILD_INTERFACE:${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis/libmetis>
     $<BUILD_INTERFACE:${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis/GKlib>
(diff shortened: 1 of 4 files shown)


dellaert:
What specific issue are we fixing? Was there a problem that prompted this PR? Is there an issue associated with the problem?

acxz:
> What specific issue are we fixing?

Unnecessary ifdef when including <metis.h>
FindSeparator tests not run under system metis

> Was there a problem that prompted this PR?

From my [review comment](https://github.com/borglab/gtsam/pull/570#issuecomment-729964071) in the last METIS PR:
> Even if that is changed, system metis does not provide metislib.h which means we won't be able to use the functionality of it. Thus rendering the SYSTEM_METIS flag broken. If its going to be broken when merged in, we should just not merge it in.

Currently the ifdef guard for SYSTEM_METIS just disables the associated test. We should fix that by allowing tests to run under system metis.

> Is there an issue associated with the problem?

Here is the issue: https://github.com/borglab/gtsam/issues/292#issuecomment-656276457
This PR implements @jlblancoc 's approach in https://github.com/borglab/gtsam/issues/292#issuecomment-656495923

dellaert:
Awesome. Requesting José as a reviewer, who can merge after approving.

cmake/HandleMetis.cmake
```diff
@@ -21,7 +21,12 @@ if(GTSAM_USE_SYSTEM_METIS)
     mark_as_advanced(METIS_LIBRARY)
 
     add_library(metis-gtsam-if INTERFACE)
-    target_include_directories(metis-gtsam-if BEFORE INTERFACE ${METIS_INCLUDE_DIR})
+    target_include_directories(metis-gtsam-if BEFORE INTERFACE ${METIS_INCLUDE_DIR}
+      # gtsam_unstable/partition/FindSeparator-inl.h uses internel metislib.h API
```
varunagrawal, line 25: typo: internal

cmake/HandleMetis.cmake
```diff
@@ -30,10 +35,12 @@ else()
   add_subdirectory(${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis)
 
   target_include_directories(metis-gtsam BEFORE PUBLIC
+    $<INSTALL_INTERFACE:include/gtsam/3rdparty/metis/>
     $<BUILD_INTERFACE:${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis/include>
+    # gtsam_unstable/partition/FindSeparator-inl.h uses internel metislib.h API
```
varunagrawal, line 40: typo: internal

varunagrawal (approved):
Minor nits but LGTM

cmake/HandleMetis.cmake
acxz:
```suggestion
      # gtsam_unstable/partition/FindSeparator-inl.h uses internal metislib.h API
```

cmake/HandleMetis.cmake
acxz:
```suggestion
    # gtsam_unstable/partition/FindSeparator-inl.h uses internal metislib.h API
```

PR #1120, opened 2022-02-23
Using a random number generator in translation averaging

akshay-krishnan:
Updated translation averaging to use a random number generator (RNG) just like Shonan averaging. 
- Added a new method that accepts an RNG
- Using a static RNG with the existing method. 
- fixed a typo in the method name "initializeRandomly()" 
- updating a comment in Shonan averaging.


akshay-krishnan:
I just realized there might have been an existing branch called "ta-seed", but looks like nothing has been changed on it.

gtsam/sfm/TranslationRecovery.h
```diff
@@ -99,10 +99,18 @@ class TranslationRecovery {
 
   /**
    * @brief Create random initial translations.
-   *
+   * 
+   * @param rng random number generator
+   * @return Values
+   */
+  Values initializeRandomly(std::mt19937 &rng) const;
```
dellaert, line 106: Actually, I think it should be `std::mt19937 *rng`: use a pointer for i/o, not a reference. This should also be fixed in all other cases, but let's not add another case.

dellaert (approved):

PR #1121, opened 2022-02-27
Added more tests for basis

dellaert:
- moved tests for BasisFactors to basis folder
- added tests for EvaluationFactor and ManifoldEvaluationFactor
- added test for ManifoldEvaluationFunctor in testChebyshev2


varunagrawal (approved):
LGTM but we should probably add some dedicated tests for `FunctorizedFactor` (which IMO was one of the key innovations that got this stuff working!).
The BasisFactor tests currently serve more as integration tests rather than unit tests and I see this as my shortcoming in hindsight.

dellaert:
I just moved them to the right place and added some more :-)

PR #1125, opened 2022-03-02
Decision Tree `prune` method 

varunagrawal:
Added a new method to prune decision trees in general based on a predicate function.

This also required modifying some other parts to remove some assumptions about the code.

### DIFF
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -702,7 +702,7 @@ namespace gtsam {
     // If leaf, check the predicate and return accordingly.
     if (auto leaf = boost::dynamic_pointer_cast<const Leaf>(node)) {  
       if(predicate(leaf->constant())) {
-        return nullptr;
+        return NodePtr();
       }
       return NodePtr(new Leaf(leaf->constant()));
     }
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -697,8 +697,6 @@ namespace gtsam {
   template <typename L, typename Y>
   template <typename Func>
   typename DecisionTree<L, Y>::NodePtr DecisionTree<L, Y>::prune(const NodePtr& node, Func predicate) {
-    // Prune<L, Y> prune(f);
-    // prune(root_);
     using LY = DecisionTree<L, Y>;
     // If leaf, check the predicate and return accordingly.
     if (auto leaf = boost::dynamic_pointer_cast<const Leaf>(node)) {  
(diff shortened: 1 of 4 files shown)


varunagrawal:
I may have to close this PR since this method is buggy and handling the pointers correctly doesn't seem to be a trivial task.

varunagrawal:
Added a new test that demonstrates all that is wrong with this method.

dellaert:
> I may have to close this PR since this method is buggy and handling the pointers correctly doesn't seem to be a trivial task.

So, please share the printouts? And share what you think is going wrong?

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -309,7 +312,7 @@ namespace gtsam {
       }
 #endif
       size_t index = x.at(label_);
-      NodePtr child = branches_[index];
+      NodePtr child = branches_.at(index);
```
dellaert, line 315: Why did this need to be changed?

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -556,15 +559,6 @@ namespace gtsam {
     if (labelC == end) {
       // Base case: only one key left
       // Create a simple choice node with values as leaves.
-      if (size != nrChoices) {
```
dellaert, line 559: Why are you removing this code?

dellaert (commented):
I left some questions. Maybe we should meet.

gtsam/discrete/DecisionTree-inl.h
varunagrawal: Was getting segfaults over here. This change means we now get a proper error message about a vector indexing problem.

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -171,10 +171,13 @@ namespace gtsam {
       if (f->allSame_) {
         assert(f->branches().size() > 0);
         NodePtr f0 = f->branches_[0];
-        assert(f0->isLeaf());
-        NodePtr newLeaf(
+        if (f0->isLeaf()) {
```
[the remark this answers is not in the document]
varunagrawal: This is needed to get pruning to work. I may be doing something wrong though since the prune method is buggy.

gtsam/discrete/DecisionTreeFactor.cpp
```diff
@@ -167,7 +167,13 @@ namespace gtsam {
     // Construct unordered_map with values
     std::vector<std::pair<DiscreteValues, double>> result;
     for (const auto& assignment : assignments) {
-      result.emplace_back(assignment, operator()(assignment));
+      try {
```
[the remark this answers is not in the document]
varunagrawal: If a leaf has been pruned away, that particular assignment is no longer valid, which is what causes the exception. I see that this should have been the comment in the code.

dellaert:
I think we need to talk about the semantics of pruning and how it affects what invariants hold in the decision tree. This is a case where we could break the semantics of the class by not thinking it through

PR #1126, opened 2022-03-04
Decision Tree Constructor

varunagrawal:
This PR demonstrates and tries to fix a shortcoming in the decision tree constructor.

Essentially, the constructor assumes that the number of nodes passed in is equal to `numberOfLabels * cardinality` (here cardinality is assumed equal for each label but this holds in general).

#### Detailed Look
I create a DecisionTree that looks like this:

```
 Choice(C) 
 0 Choice(B) 
 0 0 Choice(A) 
 0 0 0 Leaf 0
 0 0 1 Leaf 1
 0 1 Choice(A) 
 0 1 0 Leaf 2
 0 1 1 Leaf -1
 1 Choice(B) 
 1 0 Choice(A) 
 1 0 0 Leaf 4
 1 0 1 Leaf 5
 1 1 Leaf -1
```

As one can see, the final `1 1` node is all `-1` because the decision tree combines equal branches into a single branch. When trying to create a new tree from this result, the constructor complains with:
```
Trying to create DD on A
DecisionTree::create: expected 2 values but got 1 instead
/home/varun/borglab/gtsam/gtsam/discrete/tests/testDecisionTree.cpp:417: Failure: "Exception: DecisionTree::create invalid argument" 
```
which is basically saying that it expected 2 values since the cardinality of `A` is 2.

#### Why is this an issue?

We want to enable pruning of the decision tree, so we should be able to say that a leaf has value 0 or a nullptr, and check for that. 

This would involve going to each leaf, and based on some conditions, set that leaf to null. However, we want to be functional, so we should create a new tree that has the corresponding structure and null assignments.

This scheme works fine for the first time we do it, but if all leaves in a subtree are null, then the tree will automatically collapse all the branches into one. When this happen, a subsequent traversal and tree creation step will fail with the above error.

#### Potential Fix

The general way to fix this is to make the constructor aware of the assignment for each node.

One current fix I though of is to create a constructor that uses the same mechanics as `convertFrom` but passes in the assignment as well to the functional.

So instead of
```
template <typename X, typename Func>
    DecisionTree(const DecisionTree<L, X>& other, std::function<Y(X)> Y_of_X);
```
we should have
```
template <typename X, typename Func>
    DecisionTree(const DecisionTree<L, X>& other, std::function<Y(X, Assignment<L>)> Y_of_X);
```
and call a new method `convertFromWithChoice` that accounts for the assignment as well.

### DIFF
--- a/gtsam/discrete/tests/testDecisionTree.cpp
+++ b/gtsam/discrete/tests/testDecisionTree.cpp
@@ -20,12 +20,11 @@
 // #define DT_DEBUG_MEMORY
 // #define DT_NO_PRUNING
 #define DISABLE_DOT
-#include <gtsam/discrete/DecisionTree-inl.h>
-
-#include <gtsam/base/Testable.h>
-#include <gtsam/discrete/Signature.h>
-
 #include <CppUnitLite/TestHarness.h>
+#include <gtsam/base/Testable.h>
+#include <gtsam/discrete/DecisionTree-inl.h>
+#include <gtsam/discrete/DiscreteValues.h>
+#include <gtsam/discrete/Signature.h>
 
 #include <boost/assign/std/vector.hpp>
 using namespace boost::assign;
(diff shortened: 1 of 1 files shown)


varunagrawal:
@dellaert I have a fix ready for whenever you want it. :slightly_smiling_face:

dellaert:
Hmmm. I am not sure there is an issue. The constructor that takes keys and a set of values *does* expect a value for every assignment. That is the documented behavior, right?

dellaert:
Maybe we should meet again...

varunagrawal:
> Hmmm. I am not sure there is an issue. The constructor that takes keys and a set of values *does* expect a value for every assignment. That is the documented behavior, right?

True, but then the notion of a pruned tree isn't supported.

dellaert:
> > Hmmm. I am not sure there is an issue. The constructor that takes keys and a set of values _does_ expect a value for every assignment. That is the documented behavior, right?
> 
> True, but then the notion of a pruned tree isn't supported.

There are two notions of pruning:
- internal to DecisionTree, which is an implementation detail that saves on memory if many leaves have the same value.
- in hybrid inference, where we set some values to 0 (or infinity).

The constructor in question is orthogonal to both, and is a constructor that takes a value for every assignment, as documented. It is not broken.

Perhaps you are proposing we should have some other constructor?

dellaert:
PS only propose it if we actually need it. You told me in person that pruning works, so I assuming it is not even using this constructor, so the whole issue might be moot?

varunagrawal:
Closing in favor of updating `apply` as discussed on our call.

PR #1135, opened 2022-03-18
Fix `visitWith` for Pruned Tree

varunagrawal:
Added a unit test demonstrating a bug in partial assignments of `visitWith` and added a fix for it.

For example, if we have a pruned tree like below:

```
 Choice(C)
 0 Choice(B)
 0 0 Leaf 0
 0 1 Choice(A)
 0 1 0 Leaf 2
 0 1 1 Leaf 3
 1 Choice(B)
 1 0 Leaf 4
 1 1 Choice(A)
 1 1 0 Leaf 6
 1 1 1 Leaf 7
```

if we look at all the assignments that get generated by `visitWith`, we get

```
>>: (B, 0)(C, 0)
>>: (A, 0)(B, 1)(C, 0)
>>: (A, 1)(B, 1)(C, 0)
>>: (A, 1)(B, 0)(C, 1)  // This is WRONG!
>>: (A, 0)(B, 1)(C, 1)
>>: (A, 1)(B, 1)(C, 1)
```

and we should instead get

```
>>: (B, 0)(C, 0)
>>: (A, 0)(B, 1)(C, 0)
>>: (A, 1)(B, 1)(C, 0)
>>: (B, 0)(C, 1)  // This is CORRECT.
>>: (A, 0)(B, 1)(C, 1)
>>: (A, 1)(B, 1)(C, 1)
```

### DIFF
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -668,9 +668,11 @@ namespace gtsam {
       for (size_t i = 0; i < choice->nrChoices(); i++) {
         choices[choice->label()] = i;  // Set assignment for label to i
 
-        VisitWith<L, Y> visit(f);
-        visit.choices = choices;
-        visit(choice->branches()[i]);  // recurse!
+        (*this)(choice->branches()[i]);  // recurse!
+
+        // Remove the choice so we are backtracking
+        auto choice_it = choices.find(choice->label());
+        choices.erase(choice_it);
       }
     }
   };
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -666,8 +666,11 @@ namespace gtsam {
       if (!choice)
         throw std::invalid_argument("DecisionTree::VisitWith: Invalid NodePtr");
       for (size_t i = 0; i < choice->nrChoices(); i++) {
-        choices[choice->label()] = i;    // Set assignment for label to i
-        (*this)(choice->branches()[i]);  // recurse!
+        choices[choice->label()] = i;  // Set assignment for label to i
+
+        VisitWith<L, Y> visit(f);
+        visit.choices = choices;
+        visit(choice->branches()[i]);  // recurse!
       }
     }
   };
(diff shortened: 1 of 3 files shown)


gtsam/discrete/tests/testDecisionTree.cpp
```diff
@@ -20,13 +20,11 @@
 // #define DT_DEBUG_MEMORY
 // #define DT_NO_PRUNING
 #define DISABLE_DOT
-#include <gtsam/discrete/DecisionTree-inl.h>
-
+#include <CppUnitLite/TestHarness.h>
```
dellaert, line 23: Include order convention in gtsam is near -> far. It was already correct except all discrete includes should precede base includes.

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -666,8 +666,11 @@ namespace gtsam {
       if (!choice)
         throw std::invalid_argument("DecisionTree::VisitWith: Invalid NodePtr");
       for (size_t i = 0; i < choice->nrChoices(); i++) {
-        choices[choice->label()] = i;    // Set assignment for label to i
-        (*this)(choice->branches()[i]);  // recurse!
+        choices[choice->label()] = i;  // Set assignment for label to i
+
+        VisitWith<L, Y> visit(f);
```
dellaert, line 671: That seems rather heavy handed. Can we not fix the issue simply by removing the choice after the recursive call? That we way we are not copying binary trees all over the place. Worried that the complexity of this method just became O(n^2) rather than O(n).

dellaert (commented):
Good find! Please see whether my suggested fix works, as I think that will not touch the complexity.

gtsam/discrete/tests/testDecisionTree.cpp
varunagrawal: Auto-format oopsie

gtsam/discrete/DecisionTree-inl.h
varunagrawal: Gotcha. The `VisitWith` struct is a pretty lightweight datastructure so I figured a functional recursive implementation would be alright. Let me update it.

dellaert (approved):
Cool ! Thanks for finding the bug and fixing it!

PR #1136, opened 2022-03-19
Wrap ParameterMatrix as argument for Values

varunagrawal:
As the title says. This should solve your issues, @dellaert.


dellaert (approved):
Thanks!!

PR #1137, opened 2022-03-19
Apply DecisionTree `op` with Assignment

varunagrawal:
This PR adds a new variant of the `DecisionTree::apply` method where the function being passed in takes both the leaf value and the assignment as input. 
This allows the function to use closures to index into other trees as an example, but is more general that that since you can choose what to do with the assignment. I have added a test case which demonstrates this via a simple pruning example.

Best part is that this is a fully functional operation (aka functional style, lambdas, closures, all the good stuff).


dellaert (approved):
Cool

PR #1138, opened 2022-03-19
Various Fixes

### LINKED ISSUE #1057: Clang `-march=native` fails on Apple M1
When I try to build from source on my M1 Mac, I get this error:
`clang: error: the clang compiler does not support '-march=native'`

Is there a quick workaround for this?

Thanks!

varunagrawal:
1. Added a second test for `DecisionTree::apply` to show it doesn't enumerate all leaf options (so pruning actually works as expected).
2. Check for Apple Silicon before setting `-march=native`. If Apple system and architecture is arm64, then we don't set `-march=native`. Fixes #1057.
3. Remove spurious `using` statement so that the corresponding warning is resolved.

### DIFF
--- a/cmake/GtsamBuildTypes.cmake
+++ b/cmake/GtsamBuildTypes.cmake
@@ -191,7 +191,7 @@ endif()
 
 if (NOT MSVC)
   option(GTSAM_BUILD_WITH_MARCH_NATIVE  "Enable/Disable building with all instructions supported by native architecture (binary may not be portable!)" ON)
-  if(GTSAM_BUILD_WITH_MARCH_NATIVE)
+  if(GTSAM_BUILD_WITH_MARCH_NATIVE AND (APPLE AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64"))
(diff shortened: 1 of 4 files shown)


varunagrawal:
Added a fix for 1057 because I got annoyed at that on the Macbook

dellaert (approved):

stefangachter (commented):
Hi Varun, is this change not too strict? Actually, it limits the `GTSAM_BUILD_WITH_MARCH_NATIVE` flag to Apple architecture only. I was running into the issue under Ubuntu 18.04 (WSL2, Intel CPU). I expected to compile gtsam with `-march=native` when setting `GTSAM_BUILD_WITH_MARCH_NATIVE=ON`, but it did not add this compiler option. Thus, I got a segmentation fault when using GTSAM together in VIO package which was compiled with `-march=native` . To compile GTSAM properly, I set `CMAKE_CXX_FLAGS="-march=native"` and my issue was solved.

varunagrawal:
The flag is by default off now but I see your point.

Can you please make a PR for the fix? I don't have cycles at the moment, but I'll gladly help land the fix.

PR #1139, opened 2022-03-20
Fixed custom factors in doc/gtsam.lyx/pdf

dellaert:
Fixed math to agree with code and added explanation:

image


doc/gtsam.lyx
```diff
@@ -1620,6 +1622,110 @@ H=\left[\begin{array}{ccc}
...
+ matrix representation for 2D poses, and hence 
+\begin_inset Formula 
+\[
+h(qe^{\hat{\xi}})\approx h\left(\left[\begin{array}{ccc}
+\cos(q_{\theta}) & -\sin(q_{\theta}) & q_{x}\\
+\sin(q_{\theta}) & \cos(q_{\theta}) & q_{y}\\
+0 & 0 & 1
+\end{array}\right]\left[\begin{array}{ccc}
+1 & -\delta\theta & \delta x\\
+\delta\theta & 1 & \delta x\\
```
xsj01, line 1712: Shouldn't it be \delta y instead of \delta x here?

doc/gtsam.lyx
```diff
@@ -1620,6 +1622,110 @@ H=\left[\begin{array}{ccc}
...
+The exponential map for 
+\begin_inset Formula $SE(2)$
+\end_inset
+
+ can be approximated to first order as
+\begin_inset Formula 
+\[
+\exp\hat{\xi}\approx\left[\begin{array}{ccc}
+1 & -\delta\theta & \delta x\\
+\delta\theta & 1 & \delta x\\
```
xsj01, line 1692: Shouldn't it be \delta y instead of \delta x here?

roger-:
Tangentially related question: is there a standard mathematical notation or function for extracting a pose from a transformation matrix? Something like the skew operator in the Lie algebra but for the group?

Like `(x, y, theta)^T = somefunc(q)` for matrix `q`?

doc/gtsam.lyx
```diff
@@ -1599,11 +1604,11 @@ q_{y}
 \begin_inset Formula $q=\left(q_{x},q_{y},q_{\theta}\right)$
```
xsj01, line 1604: Should matrix representation be used here to avoid confusion? This definition of q seems inconsistent with q in the equation at line 1710

doc/gtsam.lyx
```diff
@@ -1618,6 +1623,171 @@ H=\left[\begin{array}{ccc}
...
+\end_inset
+
+when using the 
+\begin_inset Formula $3\times3$
+\end_inset
+
+ matrix representation for 2D poses, and hence 
+\begin_inset Formula 
+\[
+h(qe^{\hat{\xi}})\approx h\left(\left[\begin{array}{ccc}
```
xsj01, line 1710:
```suggestion
h(q\exp\hat{\xi})\approx h\left(\left[\begin{array}{ccc}
```

dellaert: I think it's well known that `exp(x) = e^x`

doc/gtsam.lyx
dellaert: See my comment above. I understand it's confusing if you want this document to be a primer on SE(2) and the fact that they can be represented as 3x3 matrices, but the audience here is (perhaps wrongly?) assumed knowledgeable about it.

dellaert:
> LGTM, except for the comments above

Sorry to rebut both those comments :-/

roger-:
@dellaert 
Above you define the error as:

image

Could we also define it as:

image

where 
image
just converts from homogenous to Cartesian coordinates.

I have to double check my math, but I believe this gives a Jacobian similar to the original formulation (I got a negative identity matrix followed by a non-zero last column).

Don't think this is advantageous, but was just curious.

I was also wondering if this error could be defined in the tangent space, but I guess that doesn't make sense since `m` has no angle.

PR #1143, opened 2022-03-24
Print parents when printing GaussianConditional

varunagrawal:
Currently, `GaussianConditional`s don't print the parents, only the frontals. This PR fixes that.

For the following conditional
```
auto conditional =
      GaussianConditional::FromMeanAndStddev(X(0), A1, X(1), b, sigma);
```
which is just `P(X0 | X1)`.

#### Before
```
GaussianConditional  Conditional density [x0] 
  R = [ 1 0 ]
      [ 0 1 ]
  S[x1] = [ -1 -2 ]
          [ -3 -4 ]
  d = [ 20 40 ]
isotropic dim=2 sigma=3
```

#### After
```
Conditional density [x0 | x1]
  R = [ 1 0 ]
      [ 0 1 ]
  S[x1] = [ -1 -2 ]
          [ -3 -4 ]
  d = [ 20 40 ]
isotropic dim=2 sigma=3
```

### DIFF
--- a/gtsam/linear/GaussianConditional.cpp
+++ b/gtsam/linear/GaussianConditional.cpp
@@ -89,7 +89,7 @@ namespace gtsam {
 
   /* ************************************************************************ */
   void GaussianConditional::print(const string &s, const KeyFormatter& formatter) const {
-    cout << s << "Conditional density [";
+    cout << s << " p(";
     for (const_iterator it = beginFrontals(); it != endFrontals(); ++it) {
       cout << (boost::format("%1%")%(formatter(*it))).str() << " ";
     }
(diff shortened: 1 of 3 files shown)


varunagrawal:
**NOTE**: I've converted the default print pre-text to empty string rather than `GaussianConditional` (since it is already printing `Conditional density`). Please let me know if I should undo this.

dellaert:
> **NOTE**: I've converted the default print pre-text to empty string rather than `GaussianConditional` (since it is already printing `Conditional density`). Please let me know if I should undo this.

I would prefer if you keep the default, but remove the redundant `Conditional density`. Also, maybe actually *do* print `p(x0 | x1)`. Note *lowercase* p because it's a density, not a probability mass function.

dellaert (approved):
Cool beans!

PR #1145, opened 2022-03-24
Fix `python-install`

varunagrawal:
Added a quick fix to the CMake for `python-install` so it no longer needs to install to access protected directories. Also modernized it to use `pip install` rather than the now deprecated `python setup.py install`.


ProfFan (approved):
Thanks Varun!

PR #1147, opened 2022-03-25
Improve Gaussian Conditional printing

varunagrawal:
I am still not satisfied with the printing of `GaussianConditional` since if we provide a value for the preface string `s`, it no longer prints out "GaussianConditional". This is pretty problematic in the hybrid decision tree.

I also updated the printing for when there are no parents. Before it was
```
GaussianConditional p(x0 |)
```

but now it is
```
GaussianConditional p(x0)
```


dellaert (approved):
OK, but please undo change in default.

PR #1148, opened 2022-03-26
Pose3 improvements

dellaert:
- add Matrix versions of transformTo/From
- Align, matrix version
- expose Align in wrapper

### DIFF
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -473,6 +473,17 @@ boost::optional<Pose3> Pose3::Align(const Point3Pairs &abPointPairs) {
   return Pose3(aRb, aTb);
 }
 
+boost::optional<Pose3> Pose3::Align(const Matrix& a, const Matrix& b) {
+  if (a.rows() != 3 || b.rows() != 3 || a.cols() != b.cols()) {
+    throw std::invalid_argument(
+      "Pose3:Align expects 3*N matrices of equal shape.");
+  }
+  Point3Pairs abPointPairs;
+  for (size_t j=0; j < a.cols(); j++) {
+    abPointPairs.emplace_back(a.col(j), b.col(j));
+  }
+  return Pose3::Align(abPointPairs);
+}
 boost::optional<Pose3> align(const Point3Pairs &baPointPairs) {
   Point3Pairs abPointPairs;
   for (const Point3Pair &baPair : baPointPairs) {
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -358,11 +358,8 @@ Matrix Pose3::transformFrom(const Matrix& points) const {
   if (points.rows() != 3) {
     throw std::invalid_argument("Pose3:transformFrom expects 3*N matrix.");
   }
-  Matrix result(3, points.cols());
-  for (size_t j=0; j < points.cols(); j++) {
-    result.col(j) = transformFrom(Point3(points.col(j)));
-  }
-  return result;
+  const Matrix3 R = R_.matrix();
+  return (R * points).colwise() + t_;  // Eigen broadcasting!
 }
 
 /* ************************************************************************* */
@@ -389,11 +386,8 @@ Matrix Pose3::transformTo(const Matrix& points) const {
   if (points.rows() != 3) {
     throw std::invalid_argument("Pose3:transformTo expects 3*N matrix.");
   }
-  Matrix result(3, points.cols());
-  for (size_t j=0; j < points.cols(); j++) {
-    result.col(j) = transformTo(Point3(points.col(j)));
-  }
-  return result;
+  const Matrix3 Rt = R_.transpose();
+  return Rt * (points.colwise() - t_);  // Eigen broadcasting!
 }
 
 /* ************************************************************************* */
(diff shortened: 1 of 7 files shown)


gtsam/geometry/Pose3.cpp
```diff
@@ -354,6 +354,17 @@ Point3 Pose3::transformFrom(const Point3& point, OptionalJacobian<3, 6> Hself,
...
 
+Matrix Pose3::transformFrom(const Matrix& points) const {
+  if (points.rows() != 3) {
+    throw std::invalid_argument("Pose3:transformFrom expects 3*N matrix.");
+  }
+  Matrix result(3, points.cols());
+  for (size_t j=0; j < points.cols(); j++) {
+    result.col(j) = transformFrom(Point3(points.col(j)));
+  }
+  return result;
```
varunagrawal, line 365: Why not use homogenous coordinates and make this a simple matrix-matrix multiplication? It will be vectorized and much faster. Makes sense since you're not asking for Jacobians.

gtsam/geometry/Pose3.cpp
```diff
@@ -374,6 +385,17 @@ Point3 Pose3::transformTo(const Point3& point, OptionalJacobian<3, 6> Hself,
...
 
+Matrix Pose3::transformTo(const Matrix& points) const {
+  if (points.rows() != 3) {
+    throw std::invalid_argument("Pose3:transformTo expects 3*N matrix.");
+  }
+  Matrix result(3, points.cols());
+  for (size_t j=0; j < points.cols(); j++) {
+    result.col(j) = transformTo(Point3(points.col(j)));
+  }
+  return result;
```
varunagrawal, line 396: Same deal about making this a matrix-matrix multiplication.

gtsam/geometry/Pose3.cpp
dellaert: That's a good point actually. I don't *need* to embed in homogeneous coordinates, I can just do `R*P+t`, although I don't know how to do the '+t' so it vectorizes, so homogeneous might the way to go anyway.

dellaert:
Used Eigen broadcasting: https://eigen.tuxfamily.org/dox/group__TutorialReductionsVisitorsBroadcasting.html#title7

jerredchen (approved):
After the Eigen broadcasting everything looks good with me.

python/gtsam/tests/test_Pose3.py
```diff
@@ -30,13 +30,34 @@ def test_between(self):
...
         """Test transformTo method."""
-        transform = Pose3(Rot3.Rodrigues(0, 0, -1.570796), Point3(2, 4, 0))
-        actual = transform.transformTo(Point3(3, 2, 10))
+        pose = Pose3(Rot3.Rodrigues(0, 0, -math.pi/2), Point3(2, 4, 0))
+        actual = pose.transformTo(Point3(3, 2, 10))
         expected = Point3(2, 1, 10)
         self.gtsamAssertEquals(actual, expected, 1e-6)
 
+        # multi-point version
+        points = np.stack([Point3(3, 2, 10), Point3(3, 2, 10)]).T
```
varunagrawal, line 41: Just an FYI, you can use np.hstack here to stack along the horizontal axis.

python/gtsam/tests/test_Pose3.py
```diff
@@ -30,13 +30,34 @@ def test_between(self):
...
+        actual_array = pose.transformTo(points)
+        self.assertEqual(actual_array.shape, (3, 2))
+        expected_array = np.stack([expected, expected]).T
+        np.testing.assert_allclose(actual_array, expected_array, atol=1e-6)
+
+    def test_transformFrom(self):
+        """Test transformTo method."""
```
varunagrawal, line 48: Please update the docstring?

varunagrawal (approved):
One docstring needs updating but LGTM

gtsam/geometry/Pose3.cpp
```diff
@@ -354,6 +354,14 @@ Point3 Pose3::transformFrom(const Point3& point, OptionalJacobian<3, 6> Hself,
   return R_ * point + t_;
 }
 
+Matrix Pose3::transformFrom(const Matrix& points) const {
+  if (points.rows() != 3) {
+    throw std::invalid_argument("Pose3:transformFrom expects 3*N matrix.");
```
johnwlambert, line 359: Thanks for the PR, Frank. I personally would find it more intuitive to have (N,3) as input, rather than (3,N).

dellaert: I respect that, and realize many will feel that way as it's the numpy way. I don't understand why. Points are column vectors in all texts I know of, and replicating columns is much more natural, e.g.,  to multiply with matrices as the code above shows.  It is also used the convention elsewhere in GTSAM (unless it slipped in somewhere, in which case I'd like to deprecate it).

python/gtsam/tests/test_Pose3.py
dellaert: Done

python/gtsam/tests/test_Pose3.py
dellaert: That actually does *not* work. Dang numpy and it's weird treatment of vectors and matrices.

PR #1154, opened 2022-03-30
Record number of assignments for each leaf

varunagrawal:
This PR adds a new variable to `DecisionTree::Leaf` nodes so we can keep track of the number of assignments in each leaf.
Particularly useful when pruning is enabled as it avoids the need to inspect the internal structure of the decision tree.


varunagrawal:
@dellaert in case you didn't get a notification email.

dellaert (approved):

PR #1155, opened 2022-03-31
DecisionTree Refactor

varunagrawal:
Refactored `DecisionTree` to:

1. Make a distinction between leaves and assignments. A tree always has a full number of assignments (which each gives us a value) but can have a smaller number of leaves (the data structure) due to pruning.
2. Updated the documentation accordingly.
3. Renamed `DT_NO_PRUNING` to `GTSAM_DT_NO_PRUNING` in the event we decide to expose it via CMake.

### DIFF
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -59,7 +59,7 @@ namespace gtsam {
     /** constant stored in this leaf */
     Y constant_;
 
-    /** The number of assignments contained within this leaf
+    /** The number of assignments contained within this leaf.
      * Particularly useful when leaves have been pruned.
      */
     size_t nrAssignments_;
@@ -68,7 +68,7 @@ namespace gtsam {
     Leaf(const Y& constant, size_t nrAssignments = 1)
         : constant_(constant), nrAssignments_(nrAssignments) {}
 
-    /** return the constant */
+    /// Return the constant
     const Y& constant() const {
       return constant_;
     }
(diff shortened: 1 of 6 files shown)


varunagrawal:
@dellaert hope this PR is up to your expectations.

varunagrawal:
@dellaert gentle reminder to take a look at this when you can :)

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -678,7 +680,14 @@ namespace gtsam {
   }
 
   /****************************************************************************/
-  // Functor performing depth-first visit without Assignment<L> argument.
+  /**
+   * Functor performing depth-first visit without Assignment<L> argument.
+   *
+   * NOTE: We differentiate between leaves and assignments. Concretely, a 3
```
dellaert, line 686: So, what does this Functor do? Leaves, right? And should it not now pass the number of assignments captured by the leaf?

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -678,7 +680,14 @@ namespace gtsam {
   }
 
   /****************************************************************************/
-  // Functor performing depth-first visit without Assignment<L> argument.
+  /**
+   * Functor performing depth-first visit without Assignment<L> argument.
```
dellaert, line 684: What *is* the argument then?

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -763,12 +775,14 @@ namespace gtsam {
   }
 
   /****************************************************************************/
-  // labels is just done with a visit
+  // Get (partial) labels by performing a visit.
```
dellaert, line 778: Explain more. Maybe with example.

gtsam/discrete/DecisionTree.h
```diff
@@ -234,11 +233,13 @@ namespace gtsam {
 
     /**
      * @brief Visit all leaves in depth-first fashion.
-     * 
-     * @param f side-effect taking a value.
-     * 
-     * @note Due to pruning, leaves might not exhaust choices.
-     * 
+     *
```
dellaert, line 236: I think these functions passed in  should now also take an argument for the number of assignments captured in a particular leaf.

gtsam/discrete/DecisionTree-inl.h
varunagrawal: The `VisitLeaf` functor passes the `Leaf` object which has the number of assignments recorded in it.

gtsam/discrete/DecisionTree.h
varunagrawal: I added `visitLeaf` as an alternative to that. Should I remove that and update the rest to do this (will be API breaking)?

dellaert (approved):
Approved, and I will merge...

gtsam/discrete/DecisionTree.h
dellaert: Shoot, I merged too soon! I just noticed this new visitLeaf, which in this PR seems identical??

dellaert: And the examples are wrong I think... We can talk about it in our meeting.

varunagrawal: Okay! Though I don't want to spend more than 5 minutes on this since it is secondary to the workshop paper.

PR #1156, opened 2022-03-31
New `maxNrAssignment` scheme for pruning

varunagrawal:
As discussed in our meeting yesterday, this PR updates the `DecisionTreeFactor::prune` method to use the new `maxNrAssignments` scheme.

Here, we use the record of assignments for each leaf to compute the number of duplicate values in the tree without enumerating all the (exponential) number of assignments, and threshold on the top `maxNrAssignments` values. To facilitate this, I added a new `visitLeaf` method to `DecisionTree` which follows the same idea as `visit` in that we visit all the leaves (and not assignments), but `visitLeaf` passes the full leaf object to the functional, allowing us to do `leaf.nrAssignments()` and `leaf.constant()` quite easily.

**NOTE** I have one question: should the functional receive the leaf object or the leaf pointer? It is a const either way, but I've made it as a leaf object const-ref so there shouldn't be a memory hit and it is convenient to use (since we don't declare a `LeafPtr` typedef).

### DIFF
--- a/gtsam/discrete/AlgebraicDecisionTree.h
+++ b/gtsam/discrete/AlgebraicDecisionTree.h
@@ -160,7 +160,7 @@ namespace gtsam {
               const typename Base::LabelFormatter& labelFormatter =
                   &DefaultFormatter) const {
       auto valueFormatter = [](const double& v) {
-        return (boost::format("%4.4g") % v).str();
+        return (boost::format("%4.8g") % v).str();
       };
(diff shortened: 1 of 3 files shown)


dellaert (approved):

ProfFan:
@varunagrawal  Shall we merge this?

varunagrawal:
I'm waiting for approval on #1155

PR #1157, opened 2022-03-31
InvDepthFactor3: Fix wrong derivative H3

d-vo:
This PR fixes a wrong assignment in `gtsam_unstable/slam/InvDepthFactor3.h` where H2 is set to the wrong dimensions and H3 is left unassigned in case of a projection outside of the camera field-of-view.

### DIFF
--- a/gtsam_unstable/slam/InvDepthFactor3.h
+++ b/gtsam_unstable/slam/InvDepthFactor3.h
@@ -92,7 +92,7 @@ public:
     } catch( CheiralityException& e) {
       if (H1) *H1 = Matrix::Zero(2,6);
       if (H2) *H2 = Matrix::Zero(2,5);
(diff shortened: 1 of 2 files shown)


varunagrawal (approved):
LGTM! Thanks @d-vo

varunagrawal:
@d-vo can you please do us a favor and add a unit test for this? It should be pretty straightforward since we already have machinery for computing and verifying against numerical derivatives.

d-vo:
Sure. I can add some unit tests for the derivatives, but might take a few days before I get to it. We can leave the PR open until I get to it, or we can merge and I will add them separately - as you prefer.

varunagrawal:
I would prefer to keep this open since it helps with the git history. Looking forward to the tests!

dellaert:
Started CI, @varunagrawal or I can merge when it passes.

Issue #1092, opened 2022-02-08
std::out_of_range with map::at exception with gtsam 4.1.1 (rtabmap)


cdb0y511:
Hi, I am using rtabmp with gtsam 4.1.1. It occasionally occurs terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at.
The details are similar to 
https://github.com/introlab/rtabmap/issues/823
I hope it could be fixed soon. Maybe you could work it out together @matlabbe.
I give a reprint here.
> With GTSAM version built from source (revision: [4.1.1](https://github.com/borglab/gtsam/releases/tag/4.1.1)) during loop closure optimization:
> 
> ```shell
> [DEBUG] (2022-02-03 16:13:45.949) OptimizerGTSAM.cpp:101::optimize() Optimizing graph...
> [DEBUG] (2022-02-03 16:13:45.949) OptimizerGTSAM.cpp:146::optimize() hasGPSPrior=false
> [DEBUG] (2022-02-03 16:13:45.949) OptimizerGTSAM.cpp:164::optimize() fill poses to gtsam... rootId=1 (priorsIgnored=1 landmarksIgnored=0)
> [DEBUG] (2022-02-03 16:13:45.949) OptimizerGTSAM.cpp:223::optimize() fill edges to gtsam...
> [DEBUG] (2022-02-03 16:13:45.953) OptimizerGTSAM.cpp:479::optimize() create optimizer
> [DEBUG] (2022-02-03 16:13:45.955) OptimizerGTSAM.cpp:504::optimize() GTSAM optimizing begin (max iterations=20, robust=0)
> [DEBUG] (2022-02-03 16:13:45.965) OptimizerGTSAM.cpp:584::optimize() iteration 1 error =2850609649897726489254101293091235944356044801931291559347145600874420166397741503058701851780329932591817167121596483299665242156038849992125986899196008202587784806682300342868079360100003408894353952842967492154479751175590697972670210540378883501996026547213724436053723262923791400960.000000
> [DEBUG] (2022-02-03 16:13:45.976) OptimizerGTSAM.cpp:584::optimize() iteration 2 error =5016100078967479096064413010768809019148602854242979119938905283661092705639338472477011700353588127562436548176456729989570331508193090845168052534407557742985955853652363314364564935033288790078552428268421593627697933614414369113679293174995444909290919633032349218707196942858868340438922690560.000000
[...]

PR #1158, opened 2022-04-05
Fix the race!

ProfFan:
This fixes the TBB race by adding proper locks.

### DIFF
--- a/gtsam/inference/ClusterTree-inst.h
+++ b/gtsam/inference/ClusterTree-inst.h
@@ -15,6 +15,10 @@
 #include <gtsam/base/timing.h>
 #include <gtsam/base/treeTraversal-inst.h>
 
+#ifdef GTSAM_USE_TBB
+#include <mutex>
+#endif
+
 namespace gtsam {
 
 /* ************************************************************************* */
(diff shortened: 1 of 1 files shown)


ProfFan:
This fixes #1092. @DanMcGann @acxz @cdb0y511 @matlabbe could you all take a look and see if this fixed your problem?

varunagrawal:
Can we get some performance numbers please?
Want to make sure the mutexes aren't preventing full parallelization.

ProfFan:
I did tracing locally, the lock is only hit about 5 times per optimization and the performance impact is neglectable. @varunagrawal

varunagrawal:
Having full performance metrics would still be great. Think about when something changes in the future and someone else comes and looks at this PR.

ProfFan:
Before: 1.33s per optimization
After: 1.35s per optimization for the benchmark `testTBBFail` posted in `fix/tbb_regression` (experiment branch)

ProfFan:
@dellaert It would be great if this can be merged before 4.2a6, since this is very important for TBB to be actually used in practice.

ProfFan:
Thanks Varun!

dellaert:
Please note 4.2a6 is already set in stone: Linux wheels have been out. We can't have 2 different 4.2a6 versions!
Waiting for Fan to create mac wheels and upload them to pypi.

PR #1159, opened 2022-04-05
added wrapper for ProjectionFactorRollingShutter

shteren1:
@varunagrawal i added python wrapper for a factor in gtsam_unstable, while i tried to compile and test the wrapper i found out the python wrapper requires cmake>=3.15 since a change you made in December 2021 add POP_FRONT to list which only exists in cmake>=3.15, so i added that to the python wrapper readme, i hope that is ok.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@
 # for QtCreator:
 CMakeLists.txt.user*
 xcode/
+/Dockerfile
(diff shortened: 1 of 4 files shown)


varunagrawal:
Cool. Can you please add python unit tests as well to ensure the wrapping works as expected?

shteren1:
@varunagrawal sure i will, is there any easy way to check that the tests work locally? what i did now was modify the Dockerfile for ubuntu-gtsam-python with my modified source and new cmake to see that it compiles. but i'm not sure how to run the tests.

varunagrawal:
You can run `make python-test` from the build directory.

shteren1:
@varunagrawal i added tests for constructors and for the error, i checked it locally and it passed.

ProfFan (approved):
Thanks!

PR #1168, opened 2022-04-12
Fix slow dynamic cast

### LINKED ISSUE #1165: optimization slowdown since 4.1.0 with python
## Description

Since gtsam=4.1.0 i notice slow down of factor x2-3 when running a LM optimization on standard Visual Slam problem.

## Steps to reproduce

1. create two new python virtual environment, one with gtsam=4.0.3 pip package, and one with any of the packages >=4.1.0
2. copy the following file somewhere locally: https://github.com/shteren1/gtsam/blob/python_timing/python/gtsam_unstable/tests/python_timing.py
3. run the script and see the timing, i got ~9 seconds using gtsam=4.0.3 and 22 seconds using 4.2a5

This script generate 20 uniform poses on a circle of radius 20m all facing towards the center, it then generate 1000 landmarks around the center of the circle and generate measurements for them in each of the cameras represented by the poses.
it adds the generated landmarks positions + some random noise and poses position + random noise as initial guess and optimizes.
The script is compatible with both versions of gtsam before and after 4.1

## Expected behavior

runtime should be nearly the same, or ideally even better with newer versions

## Environment

tested on ubuntu 18.04 using clean python virtual environment
EDIT:
added priors on the first pose and landmark that were missing, the timing diff is x4 now instead of the x2.5 it was before. and its possible to use GN and DL optimizers also, without the prior they were throwing indeterminant linear system error.

ProfFan:
Closes #1165 

Before:
`optimizer ran: 63 iterations, time took: 38.46747899055481 seconds, initial error: 7505.995184644436, final error: 5742.671242796094`

After
`optimizer ran: 63 iterations, time took: 4.811146020889282 seconds, initial error: 7505.995184644436, final error: 5742.671242796094`

on my M1 max mac

### DIFF
--- a/gtsam/nonlinear/Values-inl.h
+++ b/gtsam/nonlinear/Values-inl.h
@@ -294,7 +294,7 @@ namespace gtsam {
    // Handle dynamic matrices
    template <int M, int N>
    struct handle_matrix<Eigen::Matrix<double, M, N>, true> {
-     Eigen::Matrix<double, M, N> operator()(Key j, const Value* const pointer) {
+     inline Eigen::Matrix<double, M, N> operator()(Key j, const Value* const pointer) {
        try {
          // value returns a const Matrix&, and the return makes a copy !!!!!
          return dynamic_cast<const GenericValue<Eigen::Matrix<double, M, N>>&>(*pointer).value();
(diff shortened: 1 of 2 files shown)


ProfFan:
@shteren1

ProfFan:
Ping @dellaert for 2nd round review.

gtsam/nonlinear/Values-inl.h
```diff
@@ -294,11 +294,12 @@ namespace gtsam {
    // Handle dynamic matrices
    template <int M, int N>
    struct handle_matrix<Eigen::Matrix<double, M, N>, true> {
-     Eigen::Matrix<double, M, N> operator()(Key j, const Value* const pointer) {
-       try {
+     inline Eigen::Matrix<double, M, N> operator()(Key j, const Value* const pointer) {
```
dellaert, line 297: Since we're fixing this, one more `try catch` above (in `struct handle`) this functor should be done the same way...

dellaert (approved):
I'll approve, but before merging, search in this file for all dynamic casts and handle them similarly? I have indicated one such case in the comments but there might be more.

Also, CI has to pass of course

dellaert:
Awesome !!!!

PR #1169, opened 2022-04-14
Fix robust deserialization

### LINKED ISSUE #1164: Problem with deserialize GenericProjectionFactor with python and Cauchy
## Description
When a GenericProjectionFactor is filled with serialized data  and Cauchy robust loss it returns wrong error, using python wrapper.
I tested with Huber loss and it works fine. also notice that unweightenedError returns correct result.

## Steps to reproduce

1. see the attached python code:

```
import gtsam
import numpy as np

point_noise = gtsam.noiseModel.Diagonal.Sigmas(np.ones(2))
point_noise = gtsam.noiseModel.Robust.Create(gtsam.noiseModel.mEstimator.Cauchy.Create(1), point_noise)
factor = gtsam.GenericProjectionFactorCal3_S2(np.ones(2), point_noise, 0, 1, gtsam.Cal3_S2())
values = gtsam.Values()
values.insert(0, gtsam.Pose3())
values.insert(1, np.array([2, 0, 15]))
#init new factor with something, since there is no empty constructor for GenerincProjectionFactor
factor1 = gtsam.GenericProjectionFactorCal3_S2(np.ones(2), point_noise, 0, 2, gtsam.Cal3_S2())
factor1.deserialize(factor.serialize())
print(f"original factor error: {factor.error(values)}, deserialized factor error: {factor1.error(values)}")
```
output:
original factor error: 0.24628311578204365, deserialized factor error: 0.003451520129521184

## Expected behavior

The error returned will be similar

## Environment

tested on ubuntu 18.04 with python 3.7 and gtsam pip package version 4.2a5

ProfFan:
Closes #1164

### DIFF
--- a/gtsam/linear/LossFunctions.h
+++ b/gtsam/linear/LossFunctions.h
@@ -223,6 +223,7 @@ class GTSAM_EXPORT Cauchy : public Base {
   void serialize(ARCHIVE &ar, const unsigned int /*version*/) {
     ar &BOOST_SERIALIZATION_BASE_OBJECT_NVP(Base);
     ar &BOOST_SERIALIZATION_NVP(k_);
+    ar &BOOST_SERIALIZATION_NVP(ksquared_);
   }
 };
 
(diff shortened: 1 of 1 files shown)


varunagrawal (approved):
LGTM

varunagrawal:
Though I'm wondering why don't we have a unit test for this. We have tools for testing serialization.

ProfFan:
Yeah I think we just didn't (or can't) test all combinations of these models...

varunagrawal:
We can. It just needs some time and effort.

Issue #875, opened 2021-09-17
Custom Factor in Python with JIT


ProfFan:
Currently the CustomFactor incurs about 100x performance drop because of the following:
1. Pybind11 translation costs
2. Python GIL, thus all multi-threading goes bust (TBB)

We should have a way to use JIT engines like Numba, and directly generate a C++ function on the fly.

Need to think:
1. How to provide access to C++ structures and functions?
2. How to manage C++ memory?

ProfFan:
https://labs.quansight.org/blog/2021/10/cxx-numba-interoperability/

ProfFan:
```
# Author: Pearu Peterson
# Created: October 2021

import os
import re
import sys
import argparse
import subprocess


class Node:
    """Node represents a pair of key and value.
 
    Node may have children nodes.
    All non-root nodes have parents.
    """

    def __init__(self, parent, prefix, key, value):
        assert isinstance(parent, (Node, type(None)))
        assert isinstance(key, str), key
        assert isinstance(value, str), value
        self.parent = parent
        self.prefix = prefix  # used only when parsing clang dump output
        self.key = key
        self._value = value   # original value, unused
        self.loc = get_path(value)

        if key == 'TranslationUnitDecl':
            value = ''
        elif key in ['NamespaceDecl', 'AccessSpecDecl', 'LinkageSpecDecl']:
            # take last word
            value = value.rsplit(None, 1)[-1]
        elif key in ['TypedefDecl', 'CXXMethodDecl', 'CXXConstructorDecl', 'CXXDestructorDecl',
                     'ParmVarDecl', 'TypeAliasDecl', 'EnumConstantDecl', 'FunctionDecl',
                     'VarDecl', 'FieldDecl', 'IndirectFieldDecl', 'UnresolvedUsingValueDecl',
                     ]:
            # name 'signature' rest
            # warning: pre-name part may be relevant as well
            i = value.find("'")
            j = value.rfind("'")
            assert -1 not in {i, j}, (key, value)
            name = value[:i].rstrip().rsplit(None, 1)[-1]
            if key == 'ParmVarDecl' and ':' in name:
                name = ''
            sig = value[i:j+1]
            rest = value[j+1:].lstrip()
            value = f'{name} {sig} {rest}'
        elif key == 'CXXRecordDecl':
            m = re.match(r'.*\b(struct|class)\b\s+(.*)\s+definition', value)
            if m is not None:
                value = '%s %s' % m.groups()
            else:
[...]

PR #1171, opened 2022-04-15
Add camera and Calibration Jacobians in Python

### LINKED ISSUE #1170: Jacobian access in python wrapper
## Feature

Hi I propose to have the jacobians of all gtsam calculations accessible through the python wrappers.

## Motivation

Having jacobian access in python will help with testing a debugging challenging optimization from python, but most of all, this will improve the custom factors building from python considerably, currently one cannot use other gtsam calculation inside a custom factor, for instance projecting a 3d point onto a camera cannot be easily used in a python custom factor because there is no way to get the jacobians updated by the projection operation.

## Pitch

All gtsam routines that accept optional jacobians will acccept numpy matrices in the correct shape from the python wrappers

## Alternatives

Since some gtsam routine support getting a pointer  for a non initialized jacobian matrix and initializes it within this might not propagate well back into python, an alternative can be some intermediate wrapper function that collects the jacobian from the cpp routine and returns it as an output.
for instance:
point2, Hcamera, Hlandmark = camera.Project(landmark, return_jacobians=True)

## Additional context

I'm aware that it might be super complicated to pull off but this feature can make gtsam a very strong tool for the robotics and ML community working with python, being an alternative for other tools like JAX or pytorch for some type of optimization problems.

ProfFan:
Closes #1170

### DIFF
--- a/gtsam/base/OptionalJacobian.h
+++ b/gtsam/base/OptionalJacobian.h
@@ -20,6 +20,7 @@
 #pragma once
 #include <gtsam/config.h>      // Configuration from CMake
 #include <Eigen/Dense>
+#include <stdexcept>
 
 #ifndef OPTIONALJACOBIAN_NOBOOST
 #include <boost/optional.hpp>
@@ -94,6 +95,20 @@ public:
       map_(nullptr) {
     dynamic->resize(Rows, Cols); // no malloc if correct size
     usurp(dynamic->data());
+  }
+
+  /**
+   * @brief Constructor from an Eigen::Ref *value*. Will not usurp if dimension is wrong
+   * @note This is important so we don't overwrite someone else's memory!
+   */
+  OptionalJacobian(Eigen::Ref<Eigen::MatrixXd> dynamic_ref) :
+      map_(nullptr) {
+    if (dynamic_ref.rows() == Rows && dynamic_ref.cols() == Cols && !dynamic_ref.IsRowMajor) {
+      usurp(dynamic_ref.data());
+    } else {
+      // It's never a good idea to throw in the constructor
+      throw std::invalid_argument("OptionalJacobian called with wrong dimensions or storage order.\n");
+    }
   }
 
 #ifndef OPTIONALJACOBIAN_NOBOOST
--- a/gtsam/base/OptionalJacobian.h
+++ b/gtsam/base/OptionalJacobian.h
@@ -21,6 +21,7 @@
 #include <gtsam/config.h>      // Configuration from CMake
 #include <Eigen/Dense>
 #include <stdexcept>
+#include <string>
 
 #ifndef OPTIONALJACOBIAN_NOBOOST
 #include <boost/optional.hpp>
@@ -106,8 +107,11 @@ public:
     if (dynamic_ref.rows() == Rows && dynamic_ref.cols() == Cols && !dynamic_ref.IsRowMajor) {
       usurp(dynamic_ref.data());
     } else {
-      // It's never a good idea to throw in the constructor
-      throw std::invalid_argument("OptionalJacobian called with wrong dimensions or storage order.\n");
+      throw std::invalid_argument(
+          std::string("OptionalJacobian called with wrong dimensions or "
+                      "storage order.\n"
+                      "Expected: ") +
+          "(" + std::to_string(Rows) + ", " + std::to_string(Cols) + ")");
     }
   }
 
--- a/gtsam/base/OptionalJacobian.h
+++ b/gtsam/base/OptionalJacobian.h
@@ -102,7 +102,8 @@ public:
    * @brief Constructor from an Eigen::Ref *value*. Will not usurp if dimension is wrong
    * @note This is important so we don't overwrite someone else's memory!
    */
-  OptionalJacobian(Eigen::Ref<Eigen::MatrixXd> dynamic_ref) :
+  template<class MATRIX>
+  OptionalJacobian(Eigen::Ref<MATRIX> dynamic_ref) :
       map_(nullptr) {
     if (dynamic_ref.rows() == Rows && dynamic_ref.cols() == Cols && !dynamic_ref.IsRowMajor) {
       usurp(dynamic_ref.data());
--- a/gtsam/geometry/geometry.i
+++ b/gtsam/geometry/geometry.i
@@ -824,6 +824,7 @@ template <CALIBRATION>
 class PinholeCamera {
   // Standard Constructors and Named Constructors
   PinholeCamera();
+  PinholeCamera(const gtsam::PinholeCamera<CALIBRATION> other);
   PinholeCamera(const gtsam::Pose3& pose);
   PinholeCamera(const gtsam::Pose3& pose, const CALIBRATION& K);
   static This Level(const CALIBRATION& K, const gtsam::Pose2& pose,
@@ -850,6 +851,7 @@ class PinholeCamera {
   static gtsam::Point2 Project(const gtsam::Point3& cameraPoint);
   pair<gtsam::Point2, bool> projectSafe(const gtsam::Point3& pw) const;
   gtsam::Point2 project(const gtsam::Point3& point);
+  gtsam::Point2 project(const gtsam::Point3& point, Eigen::Ref<Eigen::MatrixXd> Dpose, Eigen::Ref<Eigen::MatrixXd> Dpoint, Eigen::Ref<Eigen::MatrixXd> Dcal);
   gtsam::Point3 backproject(const gtsam::Point2& p, double depth) const;
   double range(const gtsam::Point3& point);
   double range(const gtsam::Pose3& pose);
--- a/gtsam/geometry/geometry.i
+++ b/gtsam/geometry/geometry.i
@@ -584,7 +584,13 @@ class Cal3_S2 {
 
   // Action on Point2
   gtsam::Point2 calibrate(const gtsam::Point2& p) const;
+  gtsam::Point2 calibrate(const gtsam::Point2& p,
+                          Eigen::Ref<Eigen::MatrixXd> Dcal,
+                          Eigen::Ref<Eigen::MatrixXd> Dp) const;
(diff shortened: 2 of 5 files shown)


ProfFan:
```python3
cam1 = gtsam.PinholeCameraCal3Bundler()
cam1.deserialize('22 serialization::archive 19 1 0\n0 0 0 0 0 1 0\n1 1 0\n2 9.99727427959442139e-01 6.30191620439291000e-03 2.24814359098672867e-02 5.97546668723225594e-03 -9.99876141548156738e-01 1.45585928112268448e-02 2.25703977048397064e-02 -1.44202867522835732e-02 -9.99641239643096924e-01 0 0 -5.81446531765312802e-02 -3.64078342172925590e-02 -5.63949743097517997e-01 1 0\n3 0 0 5.18692016601562500e+02 5.18692016601562500e+02 0.00000000000000000e+00 0.00000000000000000e+00 0.00000000000000000e+00 -1.14570140838623047e-01 -3.44798192381858826e-02 1.00000000000000008e-05\n')

Dpose = np.zeros((2,3), order='F')
Dpoint = np.zeros((2,3), order='F')
Dcal = np.zeros((2,6), order='F')
cam1.project(wRb.matrix().T @ np.array([0.39, 1.94,0.0]), Dpose, Dpoint, Dcal)

```
image

gtsam/base/OptionalJacobian.h
```diff
@@ -96,6 +97,20 @@ class OptionalJacobian {
...
+  /**
+   * @brief Constructor from an Eigen::Ref *value*. Will not usurp if dimension is wrong
+   * @note This is important so we don't overwrite someone else's memory!
+   */
+  OptionalJacobian(Eigen::Ref<Eigen::MatrixXd> dynamic_ref) :
+      map_(nullptr) {
+    if (dynamic_ref.rows() == Rows && dynamic_ref.cols() == Cols && !dynamic_ref.IsRowMajor) {
+      usurp(dynamic_ref.data());
+    } else {
+      // It's never a good idea to throw in the constructor
```
dellaert, line 109: Ok, then what?

gtsam/base/OptionalJacobian.h
```diff
@@ -96,6 +97,20 @@ class OptionalJacobian {
     usurp(dynamic->data());
   }
 
+  /**
+   * @brief Constructor from an Eigen::Ref *value*. Will not usurp if dimension is wrong
```
dellaert, line 101: Nice!

gtsam/geometry/geometry.i
```diff
@@ -850,6 +851,7 @@ class PinholeCamera {
   static gtsam::Point2 Project(const gtsam::Point3& cameraPoint);
   pair<gtsam::Point2, bool> projectSafe(const gtsam::Point3& pw) const;
   gtsam::Point2 project(const gtsam::Point3& point);
+  gtsam::Point2 project(const gtsam::Point3& point, Eigen::Ref<Eigen::MatrixXd> Dpose, Eigen::Ref<Eigen::MatrixXd> Dpoint, Eigen::Ref<Eigen::MatrixXd> Dcal);
```
dellaert, line 854: Ok, that’s very cool!

dellaert (commented):
That’s amazing! Add a test in python?
Should we merge this after adding a large number of overloads?

gtsam/base/OptionalJacobian.h
ProfFan: Since we don't allocate in this constructor it's probably fine though...

ProfFan:
@dellaert I added some overloads, but it's too cumbersome to add them all. I think maybe @shteren1 could send a PR with wrapping of those he needs?

python/gtsam/tests/test_PinholeCamera.py
```diff
@@ -0,0 +1,58 @@
...
+from math import pi
+
+import numpy as np
+
+import gtsam
+from gtsam.utils.test_case import GtsamTestCase
+
+
+class TestPinholeCamera(GtsamTestCase):
+    def test_jacobian(self):
```
dellaert, line 21: add docstring saying what is tested and where the data comes from

python/gtsam/tests/test_PinholeCamera.py
```diff
@@ -0,0 +1,58 @@
...
+        cam1 = gtsam.PinholeCameraCal3Bundler()
+        cam1.deserialize(
```
dellaert, line 23: uhg? This makes it a regression test, not a unit test.
For purposes of the new functionality (and documentation on how to use) a simple example with round numbers would be preferred. It suffices even to just check one number as a regression, as this is not supposed to test the derivatives (done nicely in c++) but the passing of derivatives to python.

python/gtsam/tests/test_PinholeCamera.py
```diff
@@ -0,0 +1,58 @@
...
+            '22 serialization::archive 19 1 0\n0 0 0 0 0 1 0\n1 1 0\n2 9.99727427959442139e-01 6.30191620439291000e-03 2.24814359098672867e-02 5.97546668723225594e-03 -9.99876141548156738e-01 1.45585928112268448e-02 2.25703977048397064e-02 -1.44202867522835732e-02 -9.99641239643096924e-01 0 0 -5.81446531765312802e-02 -3.64078342172925590e-02 -5.63949743097517997e-01 1 0\n3 0 0 5.18692016601562500e+02 5.18692016601562500e+02 0.00000000000000000e+00 0.00000000000000000e+00 0.00000000000000000e+00 -1.14570140838623047e-01 -3.44798192381858826e-02 1.00000000000000008e-05\n'
+        )
+
+        # order is important because Eigen is column major!
+        Dpose = np.zeros((2, 6), order='F')
```
dellaert, line 28: Good catch!

dellaert (commented):
See comments. Regarding wrapping everything:
Could we just do *all* camera derivatives as promised by PR? That's not all derivatives in whole of GTSAM, but a good start/example. Just *part* of the camera derivatives is neither here nor there I think.

python/gtsam/tests/test_PinholeCamera.py
ProfFan: done

gtsam/geometry/geometry.i
```diff
@@ -584,7 +584,13 @@ class Cal3_S2 {
 
   // Action on Point2
   gtsam::Point2 calibrate(const gtsam::Point2& p) const;
+  gtsam::Point2 calibrate(const gtsam::Point2& p,
```
dellaert, line 587: Cal3_S2 is **not** a camera, so thank you for the extra work :-) But please update PR comment (do that anyway, it's very terse) to say you also provide calibration derivatives.

PS, speaking about extra work: why not just do this in the base class? Is that because it is virtual and we can't handle?

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -139,6 +139,17 @@ def test_jacobian(self):
         self.gtsamAssertEquals(z, np.zeros(2))
         self.gtsamAssertEquals(H @ H.T, 4*np.eye(2))
 
+        Dcal = np.zeros((2, 10), order='F')
+        Dp = np.zeros((2, 2), order='F')
+        camera.calibrate(img_point, Dcal, Dp)
+        
+        self.gtsamAssertEquals(Dcal, np.array(
```
dellaert, line 146: numpy arrays should be tested with `np.testing.assert...`

python/gtsam/tests/test_Cal3Unified.py
```diff
@@ -139,6 +139,17 @@ def test_jacobian(self):
         self.gtsamAssertEquals(z, np.zeros(2))
         self.gtsamAssertEquals(H @ H.T, 4*np.eye(2))
 
+        Dcal = np.zeros((2, 10), order='F')
```
dellaert, line 142: Now that we *have* derivatives, we no longer have to jump through the linearize hoop above for the earlier test. Can you read and update the test?

python/gtsam/tests/test_PinholeCamera.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_jacobian(self):
+        cam1 = gtsam.PinholeCameraCal3Bundler()
+
+        # order is important because Eigen is column major!
+        Dpose = np.zeros((2, 6), order='F')
+        Dpoint = np.zeros((2, 3), order='F')
+        Dcal = np.zeros((2, 3), order='F')
+        cam1.project(np.array([1, 1, 1]), Dpose, Dpoint, Dcal)
+
+        self.gtsamAssertEquals(Dpoint, np.array([[1, 0, -1], [0, 1, -1]]))
```
dellaert, line 33: np.testing...

dellaert (approved):
Approving modulo some style comments.

As far as modernizing the test, should be easy, but I understand if you're strapped for time.

python/gtsam/tests/test_Cal3Unified.py
ProfFan: internally gtsamAssertEquals also deal with NP arrays

python/gtsam/tests/test_Cal3Unified.py
ProfFan: I think they test for different derivatives though

gtsam/geometry/geometry.i
ProfFan: Most are done in the Base class, but some of them actually are different than the base class. I decided that I should not touch anything that works, so I just kept all the original and added overloads.

ProfFan:
I'll merge first, don't really have time for another round...

dellaert:
Okidoke. Could you announce on the Google group, though? This is pretty cool and I’d like you to get the credit :-)

shteren1:
@ProfFan Thank you very much! I will create a PR soon with additional jacobians wrappings for python.
I think GTSAM is a great alternative for scipy.optimize.least_squares or JAX now when it comes to general NLSQ optimizations in python.
and it'll only get better if you implement the JIT mechanism mentioned at #875.

Issue #1101, opened 2022-02-15
ISAM2 Marginalization Issues


gradyrw:
## Description

I have been trying to use ISAM2's marginalizeLeaves function and encountered a few issues, particularly on larger problems. I've tracked the issues down as best I can and I believe there are 3 different bugs in the marginalizeLeaves function which are causing problems:
1. There is a missing `nullptr` check which is required in cases where a leaf is also a root (e.g. marginalizing the only node in a graph).
   
2. The vertical block matrix of the clique's conditional is not getting updated properly in cases where the entire clique is not marginalized
   
3. The logic for splitting a clique assumes that keys are ordered in a way such that all the marginalized keys appear before any keys that are being kept. This is not always true.
   
I think there are simple solutions for the first two issues, but I'm not sure about the final one.

## Steps to reproduce

I'm using this code to force the keys that I want to marginalize to be leaves and to then update ISAM2 and marginalize:
```
boost::optional<FastMap<Key, int> > createOrderingConstraints(const ISAM2& isam, const KeyVector& new_keys,
                                                              const KeySet& marginalizable_keys)
{
  if (marginalizable_keys.empty()) {
    return boost::none;
  } else {
    FastMap<Key, int> constrained_keys = FastMap<Key, int>();
    // Generate ordering constraints so that the marginalizable variables will be eliminated first
    // Set all existing and new variables to Group1
    for (const auto& [key, val] : isam.getDelta()) {
      constrained_keys.emplace(key, 1);
    }
    for (const auto& key : new_keys) {
      constrained_keys.emplace(key, 1);
    }
    // And then re-assign the marginalizable variables to Group0 so that they'll all be leaf nodes
    for (const auto& key : marginalizable_keys) {
      constrained_keys.at(key) = 0;
    }
    return constrained_keys;
  }
}

void markAffectedKeys(const Key& key, const ISAM2Clique::shared_ptr& root_clique, KeyList& additional_keys)
{
  std::stack<ISAM2Clique::shared_ptr> frontier;
  frontier.push(root_clique);
  // Basic DFS to find additional keys
  while (!frontier.empty()) {
    // Get the top of the stack
    const ISAM2Clique::shared_ptr clique = frontier.top();
    frontier.pop();
    // Check if we have more keys and children to add
    if (std::find(clique->conditional()->beginParents(), clique->conditional()->endParents(), key) !=
        clique->conditional()->endParents()) {
      for (Key i : clique->conditional()->frontals()) {
[...]

PR #1105, opened 2022-02-15
Adding failing tests for ISAM2 marginalization

gradyrw:
Adding unit tests in ISAM2 which trigger the failures described in https://github.com/borglab/gtsam/issues/1101

PR #1172, opened 2022-04-16
ISAM2 marginalization fixes

gradyrw:
Fixes https://github.com/borglab/gtsam/issues/1101

This PR addresses 3 separate problems with the ISAM2 marginalization.

1. The first issue is that a missing null pointer check would cause a segfault if a root was attempted to be marginalized. A test was added in https://github.com/borglab/gtsam/pull/1105 ( [TEST(ISAM2, MarginalizeRoot)](https://github.com/borglab/gtsam/pull/1105/files#:~:text=TEST(ISAM2%2C%20MarginalizeRoot)) ) which triggers this behavior. The fix added here is to add a null pointer check before the addition of the marginal factor. This works since, if we are marginalizing a root, we don't need to keep track of the resulting marginal factor as the entire tree is discarded. The new test passes with the fix included here.
2. The second issue is that the clique conditional's block matrix was being incorrectly updated. When we remove variables from a clique, the column start of the vertical block matrix needs to be incremented by the number of variables removed (it was being set equal to the number of variables instead). Triggering this behavior can be difficult: you need to repeatedly marginalize variables from the same clique, but not remove the entire clique. A test was added in https://github.com/borglab/gtsam/pull/1105 ([TEST(ISAM2, marginalizeLeaves6)](https://github.com/borglab/gtsam/pull/1105/files#:~:text=TEST(ISAM2%2C%20marginalizeLeaves6)) which manages to trigger this behavior. That test now passes with these fixes.
3. The last issue is that `marginalizeLeaves` ignores the `findUnusedFactorSlots` parameter, and always pushes new marginal factors onto the back of the graph instead of inserting them into unused slots. This was fixed by replacing the call to `push_back` with a call to `add_factors` which takes into account the `findUnusedFactorSlots` param, and then using the `newFactorIndices` returned by that call to insert new factors into the graph. Another related change is removing factors _before_ inserting the new marginal factors into the graph. Together, these changes prevent the graph from growing when variables are marginalized. A new test which failed previously (and now passes) had been added in this PR ([TEST(ISAM2, marginalizationSize)](https://github.com/borglab/gtsam/pull/1172/files#:~:text=TEST(ISAM2%2C%20marginalizationSize))

### DIFF
--- a/gtsam/nonlinear/ISAM2.cpp
+++ b/gtsam/nonlinear/ISAM2.cpp
@@ -552,9 +552,12 @@ void ISAM2::marginalizeLeaves(
         // We do not need the marginal factors associated with this clique
         // because their information is already incorporated in the new
         // marginal factor.  So, now associate this marginal factor with the
-        // parent of this clique.
-        marginalFactors[clique->parent()->conditional()->front()].push_back(
-            marginalFactor);
+        // parent of this clique. If the clique is a root and has no parent, then
+        // we can discard it without keeping track of the marginal factor.
+        if (clique->parent()) {
+            marginalFactors[clique->parent()->conditional()->front()].push_back(
+                    marginalFactor);
+        }
         // Now remove this clique and its subtree - all of its marginal
         // information has been stored in marginalFactors.
         trackingRemoveSubtree(clique);
@@ -632,7 +635,7 @@ void ISAM2::marginalizeLeaves(
 
         // Make the clique's matrix appear as a subset
         const DenseIndex dimToRemove = cg->matrixObject().offset(nToRemove);
-        cg->matrixObject().firstBlock() = nToRemove;
+        cg->matrixObject().firstBlock() += nToRemove;
         cg->matrixObject().rowStart() = dimToRemove;
 
         // Change the keys in the clique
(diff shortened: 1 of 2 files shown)


dellaert:
Amazingly detailed issue, and great PR. It’s been a while since we wrote this code, so I am not super confident just from reading the code. Is there a way to make the test cases you highlighted in the issue into additional unit tests, or you can you provide context/evidence in the PR comments that those 3 issues indeed have been fixed?

gradyrw:
> Amazingly detailed issue, and great PR. It’s been a while since we wrote this code, so I am not super confident just from reading the code. Is there a way to make the test cases you highlighted in the issue into additional unit tests, or you can you provide context/evidence in the PR comments that those 3 issues indeed have been fixed?

The two failing tests that were added previously now pass, as does the third test added in this PR. If we're still unsure then it might be a good idea to do a more extensive refactor, breaking down this function into smaller components and adding unit tests for each of them.

mhkabir:
@gradyrw I believe that the markAffectedKeys function in `IncrementalFixedLagSmoother` also suffers from the same issue as your implementation initially did - should that also be fixed in this PR?

dellaert:
> @gradyrw I believe that the markAffectedKeys function in `IncrementalFixedLagSmoother` also suffers from the same issue as your implementation initially did - should that also be fixed in this PR?

I think it should be a separate PR. I'm sorry I lost track of this PR, I will merge it now.

dellaert:
Thanks @gradyrw !!!!!

PR #1175, opened 2022-04-19
noise models: Add getters for all model parameters

jlblancoc:
Add missing "getters" for noise models. Otherwise, it's impossible to find out their internal parameters after construction time.

They were needed to allow the development of [this new](https://github.com/MRPT/gtsam2mrpt_serial) non-intrusive alternative serialization library.

### DIFF
--- a/gtsam/linear/LossFunctions.cpp
+++ b/gtsam/linear/LossFunctions.cpp
@@ -350,8 +350,8 @@ double DCS::weight(double distance) const {
   const double e2 = distance*distance;
   if (e2 > c_)
   {
-    const double w = c_/(c_ + e2);
-    return 2.0*w*w;
+    const double w = 2.0*c_/(c_ + e2);
+    return w*w;
   }
 
   return 1.0;
--- a/gtsam/linear/LossFunctions.cpp
+++ b/gtsam/linear/LossFunctions.cpp
@@ -350,8 +350,8 @@ double DCS::weight(double distance) const {
   const double e2 = distance*distance;
   if (e2 > c_)
   {
-    const double w = 2.0*c_/(c_ + e2);
-    return w*w;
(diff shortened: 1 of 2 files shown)


jlblancoc:
PS: I didn't add a new virtual method `modelParameter()` to robust Base for two reasons: 
- ABI break.
- Not all robust models have 1 parameters ("Null"), and, perhaps, one day somebody creates a model with 2 or N parameters. 
-

dellaert (approved):
I think this is OK and I'll approve it.
If time/resources were not an issue, I'd probably would have preferred a more "meaningful" name for each parameter, and - while we're at it -- better docs for each model (e.g., a doxygen block with the formula and a name for the parameter).
But, unless you have the time (or a student with time) I think this a good compromise.

jlblancoc:
@dellaert I know... I also felt the same about the docs but didn't want to extend the scope of the PR too far away :-) 
Take a look now, I'm much happier with the docs now, let me know your thoughts, or feel free of merging.

While documenting and going through the derivatives, I *think* I fixed a wrong "x 2" factor in the DCS weight, [here](https://github.com/borglab/gtsam/pull/1175/commits/42b795c769e66dacc24cbe8b13309cdfb3506fc2), with the equations  justifying the change here:

https://github.com/borglab/gtsam/blob/017e3cdb179a784bc28452e1ded40d0442bfd9a4/gtsam/linear/LossFunctions.h#L377-L389

Please, confirm whether it's right now or not... I think the "2" should go on the second line, otherwise it would become a "4". Not a big deal probably anyhow...

jlblancoc:
Ok, I just reverted the last "fix" of the DCS equation to make unit tests happy. If you think it was a real bug, we might open a issue ticket and handle it in a separate PR.

PR #1181, opened 2022-04-29
Fix warnings and docs

varunagrawal:
1. Fix the warnings due to type mismatch in newly introduced methods.
2. Update docs for IMU factors clarifying what frame the measurements are in.

### DIFF
--- a/gtsam/geometry/Pose2.cpp
+++ b/gtsam/geometry/Pose2.cpp
@@ -365,7 +365,7 @@ boost::optional<Pose2> Pose2::Align(const Matrix& a, const Matrix& b) {
       "Pose2:Align expects 2*N matrices of equal shape.");
   }
   Point2Pairs ab_pairs;
-  for (size_t j=0; j < a.cols(); j++) {
+  for (size_t j=0; j < size_t(a.cols()); j++) {
     ab_pairs.emplace_back(a.col(j), b.col(j));
   }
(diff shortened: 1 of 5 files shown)


gtsam/geometry/Pose2.cpp
```diff
@@ -365,7 +365,7 @@ boost::optional<Pose2> Pose2::Align(const Matrix& a, const Matrix& b) {
       "Pose2:Align expects 2*N matrices of equal shape.");
   }
   Point2Pairs ab_pairs;
-  for (size_t j=0; j < a.cols(); j++) {
+  for (size_t j=0; j < size_t(a.cols()); j++) {
```
jlblancoc, line 368: Thanks for the pr, I was also thinking of fixing them :-)
IMO, for C++ the C-type cast used here might be replaced by the more explicit static_cast.
That's in general. For these pair of warnings in particular I think it may make more sense to replace the loop counter variable from size_t to Eigen::Index, that should avoid casts at all and fix warnings, but haven't tried it yet...

varunagrawal: Great suggestion. I'll try them out and update!

varunagrawal: `Eigen::Index` works great!

jlblancoc (approved):
LGTM

PR #1183, opened 2022-05-02
Fix Ellipses

varunagrawal:
Merge the ellipse plotting fixes by @magicbycalvin into `develop`.


dellaert:
What was this all about??

varunagrawal:
The original fixes were to another branch and not to develop, so made a PR from that branch to develop. Will let you handle this.

PR #1187, opened 2022-05-05
Some fixes

varunagrawal:
1. Based on my meeting with @dellaert, this PR removes the key formatting when writing the vertices in the dot file. This prevents syntax errors when using keys from GTDynamics for example.
2. I was looking into the `PriorFactor` as part of the legged robot state estimator and found the tests lacking. Added more tests so I can be more confident in what is happening.


dellaert (approved):
Approved so you can merge but CI seems to fail.

varunagrawal:
Oof my bad. Updated the unit tests to capture the correct thing.

PR #1190, opened 2022-05-06
Adding options to translation averaging to support betweenfactor translation priors

akshay-krishnan:
These Point3 "between translations" (not using the term "relative translations" since this was already used for the direction measurements), can come for other sensors in a SLAM problem.


dellaert:
Cool. But what's this?
image
Did you not start from develop?

dellaert:
Two suggestions before I review:
- CI is failing, see what's up?
- I suggest you merge in develop

akshay-krishnan:
I created a branch called ta-add-methods, but it looks like there was one already, and I started working on an existing branch. Pulled from develop now.

tests/testTranslationRecovery.cpp
```diff
@@ -265,6 +265,74 @@ TEST(TranslationRecovery, ThreePosesWithZeroTranslation) {
...
+  poses.insert<Pose3>(1, Pose3(Rot3(), Point3(2, 0, 0)));
+  poses.insert<Pose3>(3, Pose3(Rot3(), Point3(1, -1, 0)));
+
+  auto relativeTranslations =
+      TranslationRecovery::SimulateMeasurements(poses, {{0, 1}, {0, 3}, {1, 3}});
+
+  TranslationRecoveryParams params;
+  std::vector<BinaryMeasurement<Point3>> betweenTranslations;
+  betweenTranslations.emplace_back(0, 1, Point3(2, 0, 0), noiseModel::Constrained::All(3, 1e2));
+  params.setBetweenTranslations(betweenTranslations);
```
dellaert, line 326: Hmmm, I think these are not mere parameters, but *data*, and hence should be given as an (optional) argument to `run`

gtsam/sfm/TranslationRecovery.cpp
```diff
@@ -71,36 +74,63 @@ TranslationRecovery::TranslationRecovery(
...
+                                                  edge.noiseModel());
+    }
+  }
+
   return graph;
 }
 
 void TranslationRecovery::addPrior(
-    const double scale, NonlinearFactorGraph *graph,
+    const double scale, const boost::shared_ptr<NonlinearFactorGraph> graph,
```
dellaert, line 96: I think `*` is right. In `.i` file use `@` to indicated an output parameter.

gtsam/sfm/TranslationRecovery.h
```diff
@@ -28,6 +28,40 @@
 
 namespace gtsam {
 
+// Parameters for the Translation Recovery problem.
+class TranslationRecoveryParams {
```
dellaert, line 32: parameters are typically structs, with fields *without* `_` at end. This is now supported in wrapper, as well. That being said, I think betweenTranslations don't belong here.

gtsam/sfm/TranslationRecovery.h
```diff
@@ -93,7 +125,8 @@ class TranslationRecovery {
    * @param graph factor graph to which prior is added.
    * @param priorNoiseModel the noise model to use with the prior.
    */
-  void addPrior(const double scale, NonlinearFactorGraph *graph,
+  void addPrior(const double scale,
```
dellaert, line 128: should remain *

gtsam/sfm/TranslationRecovery.h
```diff
@@ -131,5 +165,25 @@ class TranslationRecovery {
...
+ private:
+  /**
+   * @brief Gets the key of the variable being optimized among multiple input
+   * variables that have the same translation.
+   *
+   * @param i key of input variable.
+   * @return Key of optimized variable - same as input if it does not have any
+   * zero-translation edges.
+   */
+  Key getUniqueKey(const Key i) const;
```
dellaert, line 178: I have no idea what this does. we can chat about it.

dellaert (commented):
Cool, BUT, I think the params way is not right. We could have *only* betweenTranslations.

tests/testTranslationRecovery.cpp
akshay-krishnan: moved it to a `run` argument.

gtsam/sfm/TranslationRecovery.cpp
akshay-krishnan: changed all shared_ptr back to *, this change was to wrap these methods, but later I decided not to wrap them.

gtsam/sfm/TranslationRecovery.h
akshay-krishnan: Yeah, I would prefer struct too, I used the class for support in wrapper. Good to know that structs are supported now :)

gtsam/sfm/TranslationRecovery.cpp
dellaert: This works in wrapper... Use @

gtsam/sfm/TranslationRecovery.h
akshay-krishnan: If there are multiple input nodes that share the same translation, i.e have a zero unit translation edge between them, only one of them will be added to the factor graph. This function will return the value of the Key that was added to the graph for an input node. If there is no zero-translation edge, the output is the same as input. 

It is used each time we add the betweenFactors or any other priors, we need to find that "root" node that was added to the graph. I renamed the method, hopefully its more clear.

gtsam/sfm/sfm.i
```diff
@@ -252,14 +262,21 @@ class MFAS {
...
-      const gtsam::BinaryMeasurementsUnit3&
-          relativeTranslations);  // default LevenbergMarquardtParams
-  gtsam::Values run(const double scale) const;
+      const gtsam::TranslationRecoveryParams& lmParams);
+  TranslationRecovery(const gtsam::BinaryMeasurementsUnit3&
+                          relativeTranslations);  // default params
+  gtsam::Values run(const gtsam::BinaryMeasurementsPoint3& betweenTranslations,
+                    const double scale) const;
+  // default empty betweenTranslations
+  // gtsam::Values run(const double scale) const;
```
akshay-krishnan, line 279: @dellaert is there a way to wrap the method this way? using the default value of the first betweenTranslations argument?

dellaert: Actually, I think run should have *both* `relativeTranslations` and `betweenTranslations`. The constructor should just take parameters, not data, so you can "run" multiple datasets.
If you agree, you could make scale the first argument and provide two declarations in the .i file:

```c++
gtsam::Values run(const double scale, const gtsam::BinaryMeasurementsUnit3& relativeTranslations) const;
gtsam::Values run(const double scale, const gtsam::BinaryMeasurementsUnit3& relativeTranslations, const gtsam::BinaryMeasurementsPoint3& betweenTranslations) const;
```

akshay-krishnan:
That was not as simple as it should have been, due to the handling of nodes with zero-edges between them. But its done now, moved the relativeTranslations to `run()` . 
The handling of zero-edges now completely happens inside run(), using some helper functions that are not class methods.

dellaert (approved):

gtsam/sfm/TranslationRecovery.h
```diff
@@ -28,6 +28,15 @@
 
 namespace gtsam {
 
+// Parameters for the Translation Recovery problem.
+struct TranslationRecoveryParams {
+  // LevenbergMarquardtParams for optimization.
+  LevenbergMarquardtParams lmParams;
+
+  // Initial values, random intialization will be used if not provided.
+  Values initial;
```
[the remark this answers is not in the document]
akshay-krishnan: updated that

gtsam/sfm/TranslationRecovery.cpp
```diff
@@ -56,94 +56,148 @@ TranslationRecovery::TranslationRecovery(
...
+      removeSameTranslationNodes(relativeTranslations, sameTranslationDSFMap);
+  const std::vector<BinaryMeasurement<Point3>> nonzeroBetweenTranslations =
+      removeSameTranslationNodes(betweenTranslations, sameTranslationDSFMap);
+
+  // Create graph of translation factors.
+  NonlinearFactorGraph graph = buildGraph(nonzeroRelativeTranslations);
+
+  // Add global frame prior and scale (either from betweenTranslations or
+  // scale).
+  addPrior(nonzeroRelativeTranslations, scale, nonzeroBetweenTranslations,
```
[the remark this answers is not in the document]
akshay-krishnan: we do not add the scale prior, only a prior on the first translation to lie at (0, 0, 0). I updated the addPrior documention to be more clear about this.

PR #1193, opened 2022-05-11
All uncertainty ellipses now contain 95%

dellaert:
After many weeks, finally a PR to correct the plotting of ellipses.

Many thanks to @senselessDev and @magicbycalvin 

Fixes #1067


varunagrawal (approved):
LGTM

PR #1194, opened 2022-05-11
Wrap triangulateSafe

dellaert:
Wrap the more sophisticated triangulation code from triangulation.h.

I also slightly modified the `TriangulationResult` class.


akshay-krishnan (approved):

akshay-krishnan:
Looks like the wrapper is failing to build. Please feel free to merge once that is fixed.

PR #1195, opened 2022-05-11
Remove user flag for pip install

varunagrawal:
This PR fixes an issue where the python package is not installed correctly to the conda environment under consideration.

I created a conda env called `py38-test` and verified that removing the flag ensures the install location is under that environment.

I also added printing of the python interpreter's patch version since it is helpful if we have two versions such as `3.8.11` and `3.8.13`.


dellaert (approved):
Thanks !!!

PR #1203, opened 2022-05-21
Hybrid Factor Graph Implementation

ProfFan:
Original draft PR: https://github.com/varunagrawal/gtsam/pull/42


dellaert:
Thanks @ProfFan . Maybe add @keevindoherty as a reviewer?

gtsam/hybrid/GaussianMixture.h
```diff
@@ -0,0 +1,75 @@
...
+  GaussianMixture(const KeyVector &continuousFrontals,
+                  const KeyVector &continuousParents,
+                  const DiscreteKeys &discreteParents,
+                  const Conditionals &conditionals);
+
+  using Sum = DecisionTree<Key, GaussianFactorGraph>;
+
+  const Conditionals &conditionals();
+
+  /* *******************************************************************************/
```
dellaert, line 58: These dividers should not be in .h

gtsam/hybrid/GaussianMixtureFactor.cpp
```diff
@@ -0,0 +1,86 @@
...
+          return {"nullptr"};
+        return rd.str();
+      });
+}
+
+const GaussianMixtureFactor::Factors &GaussianMixtureFactor::factors() {
+  return factors_;
+}
+
+/* *******************************************************************************/
```
dellaert, line 64: Add these dividers to all functions in .cpp

dellaert (commented):
Could you add *at least* a one-liner oxygen to all methods functions declared in .h files?

gtsam/hybrid/GaussianMixture.h
```diff
@@ -0,0 +1,75 @@
...
+
+#pragma once
+
+#include <gtsam/discrete/DecisionTree.h>
+#include <gtsam/hybrid/HybridFactor.h>
+#include <gtsam/inference/Conditional.h>
+#include <gtsam/linear/GaussianConditional.h>
+
+namespace gtsam {
+class GaussianMixture : public HybridFactor,
```
varunagrawal, line 28: Sorry about this but it got a bit confusing when we have `GaussianMixture` and `GaussianMixtureFactor`. Do we want to rename this to `GaussianMixtureConditional`?

gtsam/hybrid/HybridBayesNet.h
```diff
@@ -0,0 +1,43 @@
...
+ * @author  Frank Dellaert
+ * @date    December 2021
+ */
+
+#pragma once
+
+#include <gtsam/hybrid/HybridConditional.h>
+#include <gtsam/inference/BayesNet.h>
+
+#include <iostream>  // TODO!
```
varunagrawal, line 23: Can we kill this? Neither this nor the .cpp file has printing code.

gtsam/hybrid/HybridDiscreteFactor.h
```diff
@@ -0,0 +1,51 @@
...
+#pragma once
+
+#include <gtsam/discrete/DecisionTreeFactor.h>
+#include <gtsam/discrete/DiscreteFactor.h>
+#include <gtsam/hybrid/HybridFactor.h>
+
+namespace gtsam {
+
+/**
+ * A HybridDiscreteFactor is a wrapper for DiscreteFactor, so we hide the
```
varunagrawal, line 27: In terms of wording, should we use `container` instead of `wrapper`?

gtsam/hybrid/HybridEliminationTree.h
```diff
@@ -0,0 +1,64 @@
...
+   * @param factorGraph The factor graph for which to build the elimination tree
+   */
+  HybridEliminationTree(const HybridFactorGraph& factorGraph,
+                        const Ordering& order);
+
+  /** Test whether the tree is equal to another */
+  bool equals(const This& other, double tol = 1e-9) const;
+
+ private:
+  friend class ::EliminationTreeTester;
```
varunagrawal, line 61: Interesting. Are we using GTest for this? I don't believe CppUnitLite needs friend class declarations.

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -0,0 +1,420 @@
...
+      return result;
+    };
+    sum = sum.apply(add);
+  }
+  return sum;
+}
+
+/* ************************************************************************ */
+std::pair<HybridConditional::shared_ptr, HybridFactor::shared_ptr>  //
+EliminateHybrid(const HybridFactorGraph &factors, const Ordering &frontalKeys) {
```
varunagrawal, line 84: I'll take care of breaking this down to smaller functions when I update the pruning. :slightly_smiling_face:

varunagrawal (commented):
Some minor things, most of which honestly I can take care of since they're nitpicks.

I guess the primary thing is to get CI passing and then we can merge this in.

gtsam/hybrid/HybridEliminationTree.h
ProfFan: TBH I don't know, this is the case for all EliminationTrees

gtsam/hybrid/HybridFactorGraph.cpp
ProfFan: Thank you!

ProfFan:
Almost good... Just need to get MSVC happy...

gtsam/hybrid/GaussianMixture.h
ProfFan: done :)

ProfFan:
@varunagrawal probably also need to include `GaussianJunctionTree.h`

varunagrawal:
@dellaert will wait for your approval before merging.

@keevindoherty FYI!

gtsam/hybrid/HybridConditional.h
```diff
@@ -0,0 +1,135 @@
...
+  void print(
+      const std::string& s = "Hybrid Conditional: ",
+      const KeyFormatter& formatter = DefaultKeyFormatter) const override;
+
+  /// GTSAM-style equals
+  bool equals(const HybridFactor& other, double tol = 1e-9) const override;
+
+  /// @}
+
+  friend std::pair<HybridConditional::shared_ptr, HybridFactor::shared_ptr>  //
```
dellaert, line 125: put in section and add doc

gtsam/hybrid/HybridConditional.h
```diff
@@ -0,0 +1,135 @@
...
+    if (!isHybrid_) throw std::invalid_argument("Not a mixture");
+    return boost::static_pointer_cast<GaussianMixtureConditional>(inner);
+  }
+
+  DiscreteConditional::shared_ptr asDiscreteConditional() {
+    if (!isDiscrete_) throw std::invalid_argument("Not a discrete conditional");
+    return boost::static_pointer_cast<DiscreteConditional>(inner);
+  }
+
+  boost::shared_ptr<Factor> getInner() { return inner; }
```
dellaert, line 109: public?

gtsam/hybrid/HybridConditional.h
```diff
@@ -0,0 +1,135 @@
...
+  boost::shared_ptr<Factor> inner;
+
+ public:
+  /// @name Standard Constructors
+  /// @{
+
+  /// Default constructor needed for serialization.
+  HybridConditional() = default;
+
+  HybridConditional(const KeyVector& continuousKeys,
```
dellaert, line 82: Add docs to at least all public methods/constructors in .h

gtsam/hybrid/HybridDiscreteFactor.cpp
```diff
@@ -0,0 +1,48 @@
...
+  return false;
+}
+
+void HybridDiscreteFactor::print(const std::string &s,
+                                 const KeyFormatter &formatter) const {
+  HybridFactor::print(s, formatter);
+  inner->print("inner: ", formatter);
+};
+
+}  // namespace gtsam
```
dellaert, line 48: Add newline

gtsam/hybrid/HybridDiscreteFactor.cpp
```diff
@@ -0,0 +1,48 @@
...
+               ->discreteKeys()) {
+  inner = other;
+}
+
+HybridDiscreteFactor::HybridDiscreteFactor(DecisionTreeFactor &&dtf)
+    : Base(dtf.discreteKeys()),
+      inner(boost::make_shared<DecisionTreeFactor>(std::move(dtf))) {}
+
+bool HybridDiscreteFactor::equals(const HybridFactor &lf, double tol) const {
+  return false;
```
dellaert, line 39: Is this correct?

gtsam/hybrid/HybridDiscreteFactor.h
```diff
@@ -0,0 +1,52 @@
...
+  DiscreteFactor::shared_ptr inner;
+
+  // Implicit conversion from a shared ptr of DF
+  HybridDiscreteFactor(DiscreteFactor::shared_ptr other);
+
+  // Forwarding constructor from concrete DecisionTreeFactor
+  HybridDiscreteFactor(DecisionTreeFactor &&dtf);
+
+ public:
+  virtual bool equals(const HybridFactor &lf, double tol) const override;
```
dellaert, line 46: add section `Testable`

gtsam/hybrid/HybridEliminationTree.h
```diff
@@ -0,0 +1,64 @@
...
+   * this precomputed, use the other constructor instead.
+   * @param factorGraph The factor graph for which to build the elimination tree
+   */
+  HybridEliminationTree(const HybridFactorGraph& factorGraph,
+                        const Ordering& order);
+
+  /** Test whether the tree is equal to another */
+  bool equals(const This& other, double tol = 1e-9) const;
+
+ private:
```
dellaert, line 60: private at top

gtsam/hybrid/HybridFactor.h
```diff
@@ -0,0 +1,107 @@
...
+                                 const DiscreteKeys &key2);
+
+/**
+ * Base class for hybrid probabilistic factors
+ * Examples:
+ *  - HybridGaussianFactor
+ *  - HybridDiscreteFactor
+ *  - GaussianMixtureFactor
+ *  - GaussianMixture
+ */
```
dellaert, line 42: Add comment why not "type-erased"

gtsam/hybrid/HybridFactor.h
```diff
@@ -0,0 +1,107 @@
...
+class GTSAM_EXPORT HybridFactor : public Factor {
+ public:
+  // typedefs needed to play nice with gtsam
+  typedef HybridFactor This;  ///< This class
+  typedef boost::shared_ptr<HybridFactor>
+      shared_ptr;       ///< shared_ptr to this class
+  typedef Factor Base;  ///< Our base class
+
+  bool isDiscrete_ = false;
```
dellaert, line 51: All these data members should be private or protected, no?

gtsam/hybrid/HybridFactor.h
```diff
@@ -0,0 +1,107 @@
...
+  DiscreteKeys discreteKeys_;
+
+ public:
+  /// @name Standard Constructors
+  /// @{
+
+  /** Default constructor creates empty factor */
+  HybridFactor();
+
+  /** Construct from container of keys.  This constructor is used internally
```
dellaert, line 66: add docs all methods and deleted commented out code

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -0,0 +1,425 @@
...
+  // NOTE(fan): Because we are in the Conditional Gaussian regime there are only
+  // a few cases:
+  // continuous variable, we make a GM if there are hybrid factors;
+  // continuous variable, we make a GF if there are no hybrid factors;
+  // discrete variable, no continuous factor is allowed (escapes CG regime), so
+  // we panic, if discrete only we do the discrete elimination.
+
+  // However it is not that simple. During elimination it is possible that the
```
dellaert, line 96: Could you revise the comment to be clear about what was actually implemented. It still reads as "speculative", we "could" etc... It confuses me.

gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -0,0 +1,425 @@
...
+  std::unordered_map<Key, DiscreteKey> mapFromKeyToDiscreteKey;
+  std::set<DiscreteKey> discreteSeparatorSet;
+  std::set<DiscreteKey> discreteFrontals;
+
+  KeySet separatorKeys;
+  KeySet allContinuousKeys;
+  KeySet continuousFrontals;
+  KeySet continuousSeparator;
+
+  if (DEBUG) {
```
dellaert, line 140: I propose to remove all DEBUG code. Also, this function is *way* too large to read and comprehend. Could we isolate the different cases into private methods?

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -0,0 +1,110 @@
...
+                   boost::shared_ptr<FactorType> >
+  DefaultEliminate(const FactorGraphType& factors, const Ordering& keys) {
+    return EliminateHybrid(factors, keys);
+  }
+};
+
+/**
+ * Hybrid Factor Graph
+ * -----------------------
+ * This is the linear version of a hybrid factor graph. Everything inside needs
```
dellaert, line 69: What does "linear" mean in this context? Consider re-phrasing.

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -0,0 +1,110 @@
...
+
+  using Values = gtsam::Values;  ///< backwards compatibility
+  using Indices = KeyVector;     ///> map from keys to values
+
+ public:
+  HybridFactorGraph() = default;
+
+  //  /** Construct from container of factors (shared_ptr or plain objects) */
+  //  template <class CONTAINER>
+  //  explicit HybridFactorGraph(const CONTAINER& factors) : Base(factors) {}
```
dellaert, line 89: remove

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -0,0 +1,110 @@
...
+   * */
+  template <class DERIVEDFACTOR>
+  HybridFactorGraph(const FactorGraph<DERIVEDFACTOR>& graph) : Base(graph) {}
+
+  using FactorGraph::add;
+
+  /// Add a factor directly using a shared_ptr.
+  void add(JacobianFactor&& factor);
+
+  void add(DecisionTreeFactor&& factor);
```
dellaert, line 103: docs and doxygen sections.

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -0,0 +1,51 @@
...
+  GaussianFactor::shared_ptr inner;
+
+  // Explicit conversion from a shared ptr of GF
+  explicit HybridGaussianFactor(GaussianFactor::shared_ptr other);
+
+  // Forwarding constructor from concrete JacobianFactor
+  explicit HybridGaussianFactor(JacobianFactor &&jf);
+
+ public:
+  virtual bool equals(const HybridFactor &lf, double tol) const override;
```
dellaert, line 45: add `Testable` section

gtsam/hybrid/HybridISAM.h
```diff
@@ -0,0 +1,59 @@
...
+  /// @name Standard Constructors
+  /// @{
+
+  /** Create an empty Bayes Tree */
+  HybridISAM();
+
+  /** Copy constructor */
+  HybridISAM(const HybridBayesTree& bayesTree);
+
+  void updateInternal(
```
dellaert, line 44: these are in the `Constructors`  section but (a) don't seem to be constructors, and (b) don't have docs.

gtsam/hybrid/HybridJunctionTree.cpp
```diff
@@ -0,0 +1,191 @@
...
+      HybridConstructorTraversalData& parentData) {
+    // On the pre-order pass, before children have been visited, we just set up
+    // a traversal data structure with its own JT node, and create a child
+    // pointer in its parent.
+    HybridConstructorTraversalData myData =
+        HybridConstructorTraversalData(&parentData);
+    myData.myJTNode = boost::make_shared<Node>(node->key, node->factors);
+    parentData.myJTNode->addChild(myData.myJTNode);
+
+#ifdef GTSAM_HYBRID_JUNCTIONTREE_DEBUG
```
dellaert, line 62: kill all debugging code in this file as well.

gtsam/hybrid/tests/testHybridFactorGraph.cpp
```diff
@@ -0,0 +1,608 @@
...
+#include <boost/stacktrace.hpp>
+
+void my_signal_handler(int signum) {
+  ::signal(signum, SIG_DFL);
+  std::cout << boost::stacktrace::stacktrace();
+  ::raise(SIGABRT);
+}
+#endif
+
+/* ************************************************************************* */
```
dellaert, line 74: Add these dividers between all tests as is (debatable but) standard in GTSAM

gtsam/hybrid/tests/testHybridFactorGraph.cpp
```diff
@@ -0,0 +1,608 @@
...
+  /*
+  Explanation: the Junction tree will need to reeliminate to get to the marginal
+  on X(1), which is not possible because it involves eliminating discrete before
+  continuous. The solution to this, however, is in Murphy02. TLDR is that this
+  is 1. expensive and 2. inexact. neverless it is doable. And I believe that we
+  should do this.
+  */
+}
+
+/**
```
dellaert, line 237: Doxygen is for API in header files. Here just use regular c++ comments

gtsam/inference/BayesTree.h
```diff
@@ -272,24 +273,25 @@ namespace gtsam {
...
 
-    BayesTreeOrphanWrapper(const boost::shared_ptr<CliqueType>& clique) :
-      clique(clique)
-    {
-      // Store parent keys in our base type factor so that eliminating those parent keys will pull
-      // this subtree into the elimination.
-      this->keys_.assign(clique->conditional()->beginParents(), clique->conditional()->endParents());
+    BayesTreeOrphanWrapper(const boost::shared_ptr<CliqueType>& clique)
+        : clique(clique) {
+      // Store parent keys in our base type factor so that eliminating those
```
dellaert, line 286: make doxygen comment

python/gtsam/preamble/discrete.h
```diff
@@ -13,4 +13,3 @@
 
 #include <pybind11/stl.h>
 
-PYBIND11_MAKE_OPAQUE(gtsam::DiscreteKeys);
```
dellaert, line 16: Make sure all unit tests still work in python

python/gtsam/preamble/inference.h
```diff
@@ -10,5 +10,3 @@
  * Without this they will be automatically converted to a Python object, and all
  * mutations on Python side will not be reflected on C++.
  */
-
-#include <pybind11/stl.h>
```
dellaert, line 14: add newline

python/gtsam/tests/test_HybridFactorGraph.py
```diff
@@ -0,0 +1,57 @@
...
+import unittest
+
+import gtsam
+import numpy as np
+from gtsam.symbol_shorthand import X, C
+from gtsam.utils.test_case import GtsamTestCase
+
+
+class TestHybridFactorGraph(GtsamTestCase):
+    def test_create(self):
```
dellaert, line 24: add docstring with intent of test

varunagrawal:
Amazing review. I'll address everything in a PR on top of this to make reviewing easier.

keevindoherty:
Really awesome work! Thanks @varunagrawal and @ProfFan for putting together a much more mature implementation of the hybrid factor graph concepts. And thanks for keeping us in the loop.

It's a little hard to parse changes with respect to the DC-SAM implementation, I wonder if there is a brief synopsis of the difference in API (i.e. what if anything needs to be adapted on the DC-SAM side to ensure our solver remains "plug-and-play" with the latest API on this branch). I'd also be happy to chat about this sometime if you'd like :)

varunagrawal:
Yeah we should probably get on a call to discuss this. There should be a 1-1 mapping from your elements to these elements, with some name changed and updates to leverage the internals of GTSAM.

keevindoherty:
Really fantastic, thanks @varunagrawal  :D

gtsam/hybrid/GaussianMixtureConditional.h
```diff
@@ -0,0 +1,92 @@
...
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file   GaussianMixtureConditional.h
```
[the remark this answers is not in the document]
varunagrawal: In one of my reviews I asked it to be renamed to `GaussianMixtureConditional` so that it is obvious it is a conditional and contrasts with `GaussianMixtureFactor`. I can undo that if you'd like.

gtsam/hybrid/GaussianMixtureConditional.h
```diff
@@ -0,0 +1,92 @@
...
+   * @param continuousParents the continuous parents.
+   * @param discreteParents the discrete parents. Will be placed last.
+   * @param conditionals a decision tree of GaussianConditionals.
+   */
+  GaussianMixtureConditional(const KeyVector &continuousFrontals,
+                             const KeyVector &continuousParents,
+                             const DiscreteKeys &discreteParents,
+                             const Conditionals &conditionals);
+
+  using Sum = DecisionTree<Key, GaussianFactorGraph>;
```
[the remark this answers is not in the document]
varunagrawal: I'm not a fan of the typedef `Sum`. It is not readable and would prefer another name such as `GaussianFactorGraphTree` or succinctly, `GFGTree`.

gtsam/hybrid/GaussianMixtureFactor.h
```diff
@@ -0,0 +1,76 @@
...
+
+class GaussianMixtureFactor : public HybridFactor {
+ public:
+  using Base = HybridFactor;
+  using This = GaussianMixtureFactor;
+  using shared_ptr = boost::shared_ptr<This>;
+
+  using Factors = DecisionTree<Key, GaussianFactor::shared_ptr>;
+
+  Factors factors_;
```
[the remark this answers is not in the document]
varunagrawal: Cool. I was thinking `Factors` can be in the public scope since we can then use `GaussianMixtureFactor::Factors` as a type but we can think about that later.

gtsam/hybrid/GaussianMixtureFactor.h
```diff
@@ -0,0 +1,76 @@
...
+  void print(
+      const std::string &s = "HybridFactor\n",
+      const KeyFormatter &formatter = DefaultKeyFormatter) const override;
+
+ protected:
+  /**
+   * @brief Helper function to return factors and functional to create a
+   * DecisionTree of Gaussian Factor Graphs.
+   *
+   * @return Sum (DecisionTree<Key, GaussianFactorGraph)
```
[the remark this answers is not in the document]
varunagrawal: Woah. Took me a minute to find the typo myself.

gtsam/hybrid/HybridDiscreteFactor.cpp
varunagrawal: Nope. I missed this one.

gtsam/hybrid/HybridBayesTree.h
```diff
@@ -0,0 +1,102 @@
...
+    CLIQUE, typename std::enable_if<
+                boost::is_same<CLIQUE, HybridBayesTreeClique>::value> >
+    : public CLIQUE::ConditionalType {
+ public:
+  typedef CLIQUE CliqueType;
+  typedef typename CLIQUE::ConditionalType Base;
+
+  boost::shared_ptr<CliqueType> clique;
+
+  BayesTreeOrphanWrapper(const boost::shared_ptr<CliqueType>& clique)
```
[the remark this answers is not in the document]
varunagrawal: Can't move the implementation since the class is templated :(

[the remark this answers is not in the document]
ProfFan: Yes, this is deliberate because `BayesTreeOrphanWrapper` needs special implementation specific to hybrid, which is guarded by `std::enable_if<
                 boost::is_same<CLIQUE, HybridBayesTreeClique>::value>`

gtsam/hybrid/HybridConditional.h
varunagrawal: Moved to protected scoped.

varunagrawal: Nvm the method has to be public so it can be called by the factor graph.

varunagrawal: What I am doing is converting the variable from `inner` to `inner_` and adding the `inner()` method to all the classes so we're functional.

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -0,0 +1,16 @@
...
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file   HybridBayesNet.cpp
+ * @brief  A bayes net of Gaussian Conditionals indexed by discrete keys.
+ * @author Fan Jiang
+ * @date   January 2022
+ */
+
+#include <gtsam/hybrid/HybridBayesNet.h>
```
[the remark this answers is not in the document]
varunagrawal: We don't but could use it in the future? Seems fine to have it.

gtsam/hybrid/HybridConditional.h
varunagrawal: I believe we don't need this friend function declaration, so I am removing it.

gtsam/hybrid/HybridFactorGraph.h
varunagrawal: I think this may not be right. We don't have a nonlinear/linear dichotomy for the Hybrid Factor Graph. Seems like `HybridFactorGraph` should be renamed to `GaussianHybridFactorGraph`.

python/gtsam/preamble/discrete.h
varunagrawal: Yup they all work!

gtsam/hybrid/HybridFactor.h
varunagrawal: Umm I don't understand. It is not supposed to be type-erased so we only specify why when a class is type-erased?

gtsam/hybrid/HybridFactorGraph.cpp
varunagrawal: I made these changes in https://github.com/varunagrawal/gtsam/pull/46 so maybe @ProfFan can also take a look at it?

gtsam/hybrid/GaussianMixtureConditional.h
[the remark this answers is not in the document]
dellaert: I would like you to undo it, yes.

gtsam/hybrid/HybridBayesNet.cpp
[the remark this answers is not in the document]
dellaert: Kill until we need it.

gtsam/hybrid/HybridFactor.h
dellaert: The other class is type-erased so I’d like a comment on how this is different.

gtsam/hybrid/GaussianMixtureConditional.h
```diff
@@ -0,0 +1,92 @@
...
+
+#pragma once
+
+#include <gtsam/discrete/DecisionTree.h>
+#include <gtsam/hybrid/HybridFactor.h>
+#include <gtsam/inference/Conditional.h>
+#include <gtsam/linear/GaussianConditional.h>
+
+namespace gtsam {
+class GaussianMixtureConditional
```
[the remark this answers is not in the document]
dellaert: The math in the comment is too generic. I think this class is a mixture of Gaussian, so expect to see that detail :-) Please check GaussianConditional.h for an example.

gtsam/hybrid/GaussianMixtureConditional.h
```diff
@@ -0,0 +1,92 @@
...
+  Conditionals conditionals_;
+
+ public:
+  /**
+   * @brief Construct a new Gaussian Mixture object
+   *
+   * @param continuousFrontals the continuous frontals.
+   * @param continuousParents the continuous parents.
+   * @param discreteParents the discrete parents. Will be placed last.
+   * @param conditionals a decision tree of GaussianConditionals.
```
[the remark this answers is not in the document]
dellaert: Not resolved

gtsam/hybrid/HybridFactorGraph.cpp
dellaert: Why cant’t we do this in this PR? It will make the pruning PR *also* easier to review.

gtsam/hybrid/HybridFactorGraph.cpp
dellaert: I’d love this important PR to be self-contained, so could the comment be adapted?

gtsam/hybrid/HybridISAM.h
```diff
@@ -0,0 +1,69 @@
...
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file HybridISAM.h
```
dellaert, line 13: HybridGaissianISAM, to be consistent.

gtsam/hybrid/HybridJunctionTree.cpp
```diff
@@ -0,0 +1,169 @@
...
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file HybridJunctionTree.cpp
```
dellaert, line 13: HybridGaussianJunctionTree?

dellaert (commented):
Fantastic progress. Some more comments (and some important comments were added before I did a re-review) because I still don’t see eye to eye yet on some naming etc…. Also, I think we should go for `HybraidGaussianXXX` rather than `GaussianHybridXXX` because everything in this folder should mimick what is in linear and non-linear, except preceded by `Hybrid`.

gtsam/hybrid/HybridFactorGraph.h
varunagrawal: We used `GaussianHybridFactorGraph` in the previous code. It made sense since we have `HybridFactorGraph - GaussianHybridFactorGraph - NonlinearHybridFactorGraph`. I am not sure why @ProfFan set the name to `HybridGaussianFactor`, so what should we do?

gtsam/hybrid/GaussianMixture.h
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * @brief A conditional of gaussian mixtures indexed by discrete variables, as
+ * part of a Bayes Network.
+ *
+ * Represents the conditional density P(X | M, Z) where X is a continuous random
+ * variable, M is the selection of discrete variables corresponding to a subset
+ * of the Gaussian variables and Z is parent of this node
+ *
+ * The negative log-probability is given by \f$ \sum_{m=1}^M \pi_m \frac{1}{2}
```
dellaert, line 37: Nit: this is probably wrong. The mixture is `\sum N(x; R,S,T,...d) = \sum exp -0.5 |...|^2` so the negative log would be `-log \sum exp -0.5 |...|^2` which is (unfortunately) *not* `\sum 0.5 |...|^2`. But, in Conditionals we don't really care about the neg log, so we can just say

```
The probability P(x|y,z,...) is proportional to \f$ \sum_i k_i \exp - \frac{1}{2} |R_i x - (d_i - S_i y - T_i z - ...)|^2 \f$ where i indexes the components and k_i is a component-wise normalization constant.
```

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -0,0 +1,367 @@
...
+      if (p) {
+        gfg.push_back(boost::static_pointer_cast<GaussianConditional>(p));
+      } else {
+        // It is an orphan wrapped conditional
+      }
+    }
+  }
+
+  auto result = EliminatePreferCholesky(gfg, frontalKeys);
+  return std::make_pair(
```
dellaert, line 99: does just

```
return {result.first, result.second};
```

not work?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -0,0 +1,367 @@
...
+        dfg.push_back(boost::static_pointer_cast<DiscreteConditional>(p));
+      } else {
+        // It is an orphan wrapper
+      }
+    }
+  }
+
+  auto result = EliminateDiscrete(dfg, frontalKeys);
+
+  return std::make_pair(
```
dellaert, line 125: same

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -0,0 +1,367 @@
...
+}
+
+/* ************************************************************************ */
+std::pair<HybridConditional::shared_ptr, HybridFactor::shared_ptr>
+continuousElimination(const HybridGaussianFactorGraph &factors,
+                      const Ordering &frontalKeys) {
+  GaussianFactorGraph gfg;
+  for (auto &fp : factors) {
+    auto ptr = boost::dynamic_pointer_cast<HybridGaussianFactor>(fp);
+    if (ptr) {
```
dellaert, line 86: I think you can do (here and below)

```
    if (auto hgf = boost::dynamic_pointer_cast<HybridGaussianFactor>(fp)) {...
```

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -0,0 +1,367 @@
...
+hybridElimination(const HybridGaussianFactorGraph &factors,
+                  const Ordering &frontalKeys,
+                  const KeySet &continuousSeparator,
+                  const std::set<DiscreteKey> &discreteSeparatorSet) {
+  // NOTE: since we use the special JunctionTree,
+  // only possiblity is continuous conditioned on discrete.
+  DiscreteKeys discreteSeparator(discreteSeparatorSet.begin(),
+                                 discreteSeparatorSet.end());
+
+  // sum out frontals, this is the factor on the separator
```
dellaert, line 141: Can we make this a separate function as well?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -0,0 +1,367 @@
...
+    }
+  }
+
+  for (auto &f : deferredFactors) {
+    sum = addGaussian(sum, f);
+  }
+
+  gttoc(sum);
+
+  using EliminationPair = GaussianFactorGraph::EliminationResult;
```
dellaert, line 182: Comment on what's happening...

dellaert (approved):
Awesome. Still some nits, and I am not sure all classes have the right name yet.
Some HybridXX need to become HybridGaussianXX?
Since this is still in alpha I think we can leave to next PR...

gtsam/hybrid/GaussianMixture.h
varunagrawal: I was thinking about this today morning. Let me redo it.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: Not sure. Most of the comments in this round is from Fan's code that I didn't feel needed any updates. :)

varunagrawal:
> Awesome. Still some nits, and I am not sure all classes have the right name yet. Some HybridXX need to become HybridGaussianXX? Since this is still in alpha I think we can leave to next PR...

I was wondering about this. We have `GaussianEliminationTree`, `SymbolicEliminationTree` and `DiscreteEliminationTree`, so should this be just `HybridEliminationTree`? Same question for `HybridJunctionTree` since these two only come into play __after__ linearization so `Gaussian` is implied.

gtsam/hybrid/GaussianMixture.h
varunagrawal: Also, github now supports latex:

$\sum_i k_i \exp - \frac{1}{2} |R_i x - (d_i - S_i y - T_i z - ...)|^2$

gtsam/hybrid/HybridGaussianFactorGraph.cpp
ProfFan: No, CI will fail (I initially did {})

gtsam/hybrid/HybridFactor.h
ProfFan: Comment: HybridFactor is the base class, there is no need to be type erased as it is the root of the entire hybrid inheritance hierarchy.

gtsam/hybrid/HybridFactorGraph.h
ProfFan: I think this can be renamed to GaussianHybridFactorGraph

ProfFan:
Thanks @varunagrawal!

gtsam/hybrid/HybridGaussianFactorGraph.cpp
ProfFan: Main culprit is old GCC, we will be free of these problems once we move to C++14/17.

gtsam/hybrid/GaussianMixture.h
```diff
@@ -0,0 +1,133 @@
...
+ * of the Gaussian variables and Z is parent of this node
+ *
+ * The probability P(x|y,z,...) is proportional to
+ * \f$ \sum_i k_i \exp - \frac{1}{2} |R_i x - (d_i - S_i y - T_i z - ...)|^2 \f$
+ * where i indexes the components and k_i is a component-wise normalization
+ * constant.
+ *
+ */
+class GTSAM_EXPORT GaussianMixture
+    : public HybridFactor,
```
varunagrawal, line 44: @ProfFan @dellaert why is this inheriting from `HybridFactor` and not `HybridGaussianMixtureFactor` like in the previous scheme? Do we not need to record the conditionals as factors in the base class anymore?

dellaert:
> > Awesome. Still some nits, and I am not sure all classes have the right name yet. Some HybridXX need to become HybridGaussianXX? Since this is still in alpha I think we can leave to next PR...
> 
> I was wondering about this. We have `GaussianEliminationTree`, `SymbolicEliminationTree` and `DiscreteEliminationTree`, so should this be just `HybridEliminationTree`? Same question for `HybridJunctionTree` since these two only come into play **after** linearization so `Gaussian` is implied.
> 
> @dellaert once this question is resolved, I can merge.

OK, that makes sense. Yes, `HybridEliminationTree` and `HybridJunctionTree` I guess.

gtsam/hybrid/GaussianMixture.h
dellaert: I lost track of this - @ProfFan might have the answer

ProfFan: I think this can be done that way, just what was on my mind is that I want to keep the two things separate, as they don't necessarily need to have inheritance relationship. We can revisit this later.

varunagrawal: Sorry but this is an early design decision that needs to be resolved ASAP.

My understanding was that eliminating a HybridMixtureFactor should result in a GaussianMixture conditional, so there already exists an intrinsic relationship between the two.

My main concern is what happens when we only eliminate a partial set of the keys? If the factor is F(X1, X2, M1) and we only eliminate X1, shouldn't the gaussian factors for X1 still be captured in the conditional until even X2 is eliminated?

dellaert: I actually agree with Varun: a conditional *is* a factor. No sense to have extra storage in this conditional for the Gaussian components, if that is going on.

However, I don't fully understand how the elimination example is relevant. We don't eliminate conditionals, we eliminate factors, so not sure how it bears on this discussion.

varunagrawal: The GaussianMixtureFactor is a container for other factors via the decision tree with Gaussian FGs as the leaves. So if we eliminate one variable, I don't remember how that affects the other variables within the tree. What is the expected behavior?

My understanding is that it should be a conditional on the variable with the remaining variables as parents, so does this mean the conditional would be dependent on variables still a part of factors? In that case we should be recording the factors. However the final Bayes network will only have conditionals, so the GaussianMixtures will have to be retroactively updated to reflect the new conditionals for parents.

varunagrawal: I'll go back and look at the behavior from the previous scheme.

PR #1207, opened 2022-05-26
TableFactor for DiscreteFactorGraph.

ywkim0606:
1. TableFactor implementation using Eigen's SparseVector.
    - lazy_cp stands for [Lazy Cartesian Product](http://phrogz.net/lazy-cartesian-product#:~:text=Lazy%20Cartesian%20Product%3A%20Iterating%20combinations,without%20calculating%20them%20all%20first.)
    - operator*(const TableFactor& f) can be implemented more elegantly using custom interator.
3. Unittests for TableFactor.
4. Added tests in testDiscreteFactorGraph to verify if inference still works.
    - Note that TableFactor gets converted to DecisionTreeFactor.
Can be useful if the factor is very sparse (CSP).


gtsam/discrete/tests/testDiscreteFactorGraph.cpp
```diff
@@ -54,6 +55,31 @@ TEST_UNSAFE(DiscreteFactorGraph, debugScheduler) {
   EXPECT(assert_equal(mpe, actualMPE));
 }
 
+/* ************************************************************************* */
+TEST_UNSAFE(DiscreteFactorGraph, debugSchedulerTable) {
```
dellaert, line 59: remove UNSAFE

gtsam/discrete/tests/testDiscreteFactorGraph.cpp
```diff
@@ -100,6 +126,57 @@ TEST_UNSAFE( DiscreteFactorGraph, DiscreteFactorGraphEvaluationTest) {
   EXPECT_DOUBLES_EQUAL( 1.944, product(values), 1e-9);
 }
 
+/* ************************************************************************* */
+/// Test the () operator of DiscreteFactorGraph
+TEST_UNSAFE(DiscreteFactorGraph, DiscreteFactorGraphEvaluationTestTable) {
```
dellaert, line 131: remove UNSAFE

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,383 @@
...
+    return DiscreteKey(j, cardinalities_.at(j));
+  }
+
+ public:
+  // typedefs needed to play nice with gtsam
+  typedef TableFactor This;
+  typedef DiscreteFactor Base;  ///< Typedef to base class
+  typedef boost::shared_ptr<TableFactor> shared_ptr;
+
+  /// @}
```
dellaert, line 57: remove this line. This ends a section but there is no section defined yet.

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  /// @}
+};
+
+// traits
+template <>
+struct traits<TableFactor> : public Testable<TableFactor> {};
+
+}  // namespace gtsam
```
dellaert, line 383: add newline.

dellaert (changes requested):
Good PR. Please add a .cpp file and move all implementations there. Then I'll do a thorough review on .h file. Some methods are public that i don't think should be public (lazy_cp), but I'll take a look at that later.
Finally, it would be nice to get some timing results vs. DT version.

dellaert:
Sorry, was at ICRA - what's the status/plan on this?

ywkim0606:
> Sorry, was at ICRA - what's the status/plan on this?
Status:
- Rewrote max function which eliminates variables according to the ordering given. (needs review)
- Implemented maxAssignemnt function which returns the assignments of the maximum value. 
- Both has been unittested
Plan:
- Create a new branch to edit DiscreteFactorGraph. On this branch, DiscreteFactorGraph still uses DecisionTreeFactor to optimize.
- Comparing timing results with DT on classical planning task.

dellaert:
- Why the new max functions? Not questioning just curious what the issue was.
- Will review, but in meantime check what CI failure is on windows? Seems related to your code.

gtsam/discrete/TableFactor.cpp
```diff
@@ -0,0 +1,340 @@
...
+  std::vector<double> result;
+  for (std::map<size_t, double>::iterator it = cs.begin(); it != cs.end();
+       ++it) {
+    result.push_back(it->second);
+  }
+  return boost::make_shared<TableFactor>(target_key, result);
+}
+
+/* ************************************************************************ */
+TableFactor TableFactor::max(size_t nrFrontals) const {
```
dellaert, line 155: can you not call the method below?

gtsam/discrete/TableFactor.cpp
```diff
@@ -0,0 +1,340 @@
...
+/* ************************************************************************ */
+TableFactor TableFactor::max(const Ordering& frontalKeys) const {
+  // Create map to keep track of summed values
+  if (frontalKeys.size() > size())
+    throw invalid_argument(
+        (boost::format("TableFactor::max: invalid number of frontal "
+                       "keys %d, nr.keys=%d") %
+         frontalKeys.size() % size())
+            .str());
+  TableFactor result(*this);
```
dellaert, line 179: What is happening here? How is eliminate a max ? Don't just explain in reply, add a comment.

gtsam/discrete/TableFactor.cpp
```diff
@@ -0,0 +1,340 @@
...
+  for (size_t i = 0; i < frontalKeys.size(); i++) {
+    Key j = frontalKeys[i];
+    result = result.eliminate(j);
+  }
+  return result;
+}
+
+/* ************************************************************************ */
+TableFactor TableFactor::eliminate(const Key key) const {
```
dellaert, line 188: We prefer calling "up" i.e. in this case eliminate should be defined (and declared in .h) *before* max.

dellaert, line 188: Add some comments on how this function works.

gtsam/discrete/TableFactor.cpp
```diff
@@ -0,0 +1,340 @@
...
+    }
+  }
+  return eliminated_f;
+}
+
+/* ************************************************************************ */
+TableFactor TableFactor::fromDecisionTreeFactor(
+    const DecisionTreeFactor& f) const {
+  DiscreteKeys tree_dkeys = f.discreteKeys();
+  std::vector<std::pair<DiscreteValues, double>> enumerated = f.enumerate();
```
dellaert, line 223: auto?

gtsam/discrete/TableFactor.cpp
```diff
@@ -0,0 +1,340 @@
...
+      eliminated_f.sparse_table_.coeffRef(index) = it.value();
+    }
+  }
+  return eliminated_f;
+}
+
+/* ************************************************************************ */
+TableFactor TableFactor::fromDecisionTreeFactor(
+    const DecisionTreeFactor& f) const {
+  DiscreteKeys tree_dkeys = f.discreteKeys();
```
dellaert, line 222: declare closer to where you use this.

gtsam/discrete/TableFactor.cpp
```diff
@@ -0,0 +1,340 @@
...
+    index += assignment.at(key);
+  }
+  return index;
+}
+
+/* ************************************************************************ */
+DiscreteValues TableFactor::maxAssignment() const {
+  DiscreteValues assignments;
+  Eigen::Index nnz = sparse_table_.nonZeros();
+  Eigen::Index rowIdx;
```
dellaert, line 257: `rowIdx` -> `maxRowIndex`

gtsam/discrete/TableFactor.cpp
```diff
@@ -0,0 +1,340 @@
...
+  for (const std::pair<const Key, size_t>& key : cs) {
+    keys.emplace_back(key);
+  }
+  // Sort
+  std::sort(keys.begin(), keys.end());
+  return keys;
+}
+
+/* ************************************************************************ */
+std::vector<DiscreteValues> TableFactor::project(
```
dellaert, line 283: add some comments on how this function will proceed

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,199 @@
...
+
+  /** Default constructo for I/O */
+  TableFactor();
+
+  /** Constructor from Eigen::SparseVector */
+  TableFactor(const DiscreteKeys& keys,
+              const Eigen::SparseVector<double>& table);
+
+  /** Convert table to SparseVector */
+  static Eigen::SparseVector<double> Convert(const std::vector<double>& table);
```
dellaert, line 68: These should be private I think

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,199 @@
...
+  /// Finds value for the key at index
+  size_t lazy_cp(Key target_key, size_t index) const;
+
+  /// Find assignments with maximum value
+  DiscreteValues maxAssignment() const;
+
+  /// Find index for given assignment of values
+  size_t findIndex(const DiscreteValues& assignment) const;
+
+  /// Find indicies for given partial assignment of values
```
dellaert, line 146: I do not understand the comment, or why this is eliminate.

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,199 @@
...
+  /**
+   * @brief Render as markdown table
+   *
+   * @param keyFormatter GTSAM-style Key formatter.
+   * @param names optional, category names corresponding to choices.
+   * @return std::string a markdown string.
+   */
+  std::string markdown(const KeyFormatter& keyFormatter = DefaultKeyFormatter,
+                       const Names& names = {}) const override {
+    return "";
```
dellaert, line 177: Move to cpp file and fill in :-)

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,199 @@
...
+  /**
+   * @brief Render as html table
+   *
+   * @param keyFormatter GTSAM-style Key formatter.
+   * @param names optional, category names corresponding to choices.
+   * @return std::string a html string.
+   */
+  std::string html(const KeyFormatter& keyFormatter = DefaultKeyFormatter,
+                   const Names& names = {}) const override {
+    return "";
```
dellaert, line 189: same

dellaert (changes requested):
Not quite there yet.

gtsam/discrete/TableFactor.h
ywkim0606: Sorry, that was a depreciated comment. Will be updated on the upcoming commit.

ywkim0606:
> * Why the new max functions? Not questioning just curious what the issue was.
> * Will review, but in meantime check what CI failure is on windows? Seems related to your code.

I miss interpreted the comment in DecisionTreeFactor "Create new factor by maximizing over all values with the same separator". What I did was maximize over a single key looking at just its cardinality.

gtsam/discrete/TableFactor.cpp
ywkim0606: comment added

gtsam/discrete/TableFactor.cpp
ywkim0606: This function would max by eliminating the variables in incremental order without having to explicitly give an ordering. I thought this was why we had a separate function in DecisionTreeFactor so I added the same feature in TableFactor too. We can call the method below by explicitly giving the ordering in incremental order.

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,195 @@
...
+  size_t cardinality(Key j) const { return cardinalities_.at(j); }
+
+  /// Generate DecisionTreeFactor from TableFactor
+  DecisionTreeFactor toDecisionTreeFactor() const override;
+
+  /// Create new factor by summing all values with the same separator values
+  shared_ptr sum(size_t nrFrontals) const;
+
+  /// Create new TableFactor where the input key is eliminated
+  TableFactor eliminate(const Key key) const;
```
dellaert, line 125: just `(Key key)`, as is passed by value hence implied.

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,195 @@
...
+  /// Get the cardinality of key j
+  size_t cardinality(Key j) const { return cardinalities_.at(j); }
+
+  /// Generate DecisionTreeFactor from TableFactor
+  DecisionTreeFactor toDecisionTreeFactor() const override;
+
+  /// Create new factor by summing all values with the same separator values
+  shared_ptr sum(size_t nrFrontals) const;
+
+  /// Create new TableFactor where the input key is eliminated
```
dellaert, line 124: `/// Create new TableFactor where the variable with given key is eliminated.`

But I don't understand this method! Whenever we *eliminate* there is a typically *both* a conditional and a factor that is computed. In addition, elimination functions are typically free-standing functions, see, e.g., `EliminateDiscrete` and `EliminateForMPE` in `DiscreteFactorGraph.cpp`. 

If the above is not clear, I think we might have to meet about this.

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,195 @@
...
+
+  /// @}
+  /// @name Advanced Interface
+  /// @{
+
+  /// Generate TableFactor from DecisionTreeFactor
+  TableFactor fromDecisionTreeFactor(const DecisionTreeFactor& f) const;
+
+  /// Finds value for the key at index
+  size_t lazy_cp(Key target_key, size_t index) const;
```
dellaert, line 141: Naming convention (`lazyCartesianProduct` ?) and should this not be private?

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,195 @@
...
+
+  /// Find assignments with maximum value
+  DiscreteValues maxAssignment() const;
+
+  /// Find index for given assignment of values
+  size_t findIndex(const DiscreteValues& assignment) const;
+
+  /// Create union of keys
+  DiscreteKeys unionKeys(const TableFactor& f) const;
```
dellaert, line 150: private?

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,195 @@
...
+
+  /// Project assignment that agrees with the given assignment
```
dellaert, line 152: Add example here. I suggested this function but from the comment I can for the life of me not understand what this method is for or what it produces.

gtsam/discrete/TableFactor.cpp
```diff
@@ -0,0 +1,415 @@
...
+TableFactor TableFactor::operator*(const TableFactor& f) const {
+  DiscreteKeys keys = unionKeys(f);
+  size_t cardinality = 1;
+  // Create new TableFactor
+  for (DiscreteKey key : keys) cardinality *= key.second;
+  Eigen::SparseVector<double> new_sparse_table_(cardinality);
+  new_sparse_table_.reserve(sparse_table_.nonZeros() *
+                            (cardinality / sparse_table_.size()));
+  TableFactor multiplied_f(keys, new_sparse_table_);
+  for (Eigen::SparseVector<double>::InnerIterator it(f.sparse_table_); it;
```
dellaert, line 110: maybe say `using Eigen::SparseVector<double>::InnerIterator` at top of this file so you can just use `InnerIterator`....

gtsam/discrete/tests/testTableFactor.cpp
```diff
@@ -0,0 +1,323 @@
...
+    }
+  }
+  EXPECT(actual == expected);
+}
+
+/* ************************************************************************* */
+// Check markdown representation looks as expected.
+TEST(TableFactor, markdown) {
+  DiscreteKey A(12, 3), B(5, 2);
+  TableFactor f(A & B, "1 2  3 4  5 6");
```
dellaert, line 254: How about making it sparse :-)

dellaert (changes requested):
Awesome progress. Very important class, so let's iterate once more. Especially confusion around eliminate and project at the moment.

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,195 @@
...
+
+  /** Constructor from doubles */
+  TableFactor(const DiscreteKeys& keys, const std::vector<double>& table)
+      : TableFactor(keys, Convert(table)) {}
+
+  /** Constructor from string */
+  TableFactor(const DiscreteKeys& keys, const std::string& table)
+      : TableFactor(keys, Convert(table)) {}
+
+  /// Single-key specialization
```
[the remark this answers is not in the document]
ywkim0606: This constructor is written to construct a factor for one variable (DiscreteKey instead of DiscreteKeys). Should I rephrase to "Constructor for single-key"?

gtsam/discrete/TableFactor.h
ywkim0606: We can make this private but we won't be able to run unittest on this function anymore. However, since the function has been tested already I'll make it private.

gtsam/discrete/TableFactor.cpp
ywkim0606: "Error: a class qualified name is not allowed". Instead, I `typedef` it: "typedef Eigen::SparseVector<double>::InnerIterator sparse_it;"

dellaert:
@ywkim0606  What's the status here?

ywkim0606:
> @ywkim0606 What's the status here?

I'll schedule a meeting this week regarding elimination.

PR #1219, opened 2022-06-07
Fix / Move `Rot3::quaternion` to deprecated block

### LINKED ISSUE #1209: Rot3 Quaternion BUG
## Description

There is an issue when using `Rot3().quaternion().x()`, `Rot3().quaternion().y()`, `Rot3().quaternion().z()`, or `Rot3().quaternion().w()` to access the components of a quaternion. For example, when using `Rot3().quaternion().w()` to access `qw`, it actually returns `qz`.  

## Steps to reproduce

Using the flowing code to reproduce the bug:
```C++
double qw=0.782240, qx=0.589819, qy=0.199939, qz=0.000000;
std::printf("Correct quaternion:\n");
std::printf(
        "qw: %f  qx: %f  qy: %f  qz: %f\n",
        qw, qx, qy, qz
);

std::printf("\nIn GTSAM Rot3:\n");
gtsam::Rot3 a_rotation(qw, qx, qy, qz);
std::printf(
        "qw: %f  qx: %f  qy: %f  qz: %f\n",
        a_rotation.quaternion().w(),
        a_rotation.quaternion().x(),
        a_rotation.quaternion().y(),
        a_rotation.quaternion().z()
);
std::printf(
        "q[0]: %f  q[1]: %f  q[2]: %f  q[3]: %f\n",
        a_rotation.quaternion()[0],
        a_rotation.quaternion()[1],
        a_rotation.quaternion()[2],
        a_rotation.quaternion()[3]
);
std::printf("\n\n\n\n");

```

## Expected behavior
Screenshot from 2022-06-01 22-05-55

From the first 2 lines of the result above we can see that, the correct components of the quaternion should be 
`qw: 0.782240  qx: 0.589819  qy: 0.199939  qz: 0.000000`

However, when I construct a `Rot3()` instance with the quaternion constructor and access the components through the getters such as `quaternion().x()`, it returns the wrong component(it returns quaternion().w()). 

Though using the index I can get the component in the right order(see the last line of the result), where `quaternion()[0],[1],[2],[3]` corresponding to qw, qx, qy, and qz. 

Additionally, after checking the source code, I'm sure the quaternion passed to the `Rot3()` constructor is correct, as shown below:
https://github.com/borglab/gtsam/blob/342f30d148fae84c92ff71705c9e50e0a3683bda/gtsam/geometry/Rot3.h#L121

## Environment
1. GTSAM 4.0.2
2. Ubuntu 18.04
3. GCC & G++ 7.5.0

HViktorTsoi:
This PR fixes #1209. More details about this issue can be found in the group discussion https://groups.google.com/g/gtsam-users/c/AxbAAEgMHTA.

Basically, this issue is because `Rot3::quaternion()` returns a `Vector` as `[qw, qx, qy, qz]`, not a quaternion. `Vector` is a typedef for `Eigen::VectorXd`, for which `.x()` returns the first element (instead of qw), `.y()` the second (so here qx), etc... Therefore this `Rot3::quaternion()` API is very non-intuitive and it could lead to a bug in user code if they don't pay attention enough. 

This fix modifies the `Rot3::quaternion()` API and makes sure the ordering of the returned `[qw,qx,qy,qz]` are consistent with  `gtsam::Quaternion`.

### DIFF
--- a/gtsam_unstable/timing/timeShonanAveraging.cpp
+++ b/gtsam_unstable/timing/timeShonanAveraging.cpp
@@ -79,9 +79,9 @@ void saveG2oResult(string name, const Values& values, std::map<Key, Pose3> poses
         myfile << "VERTEX_SE3:QUAT" << " ";
         myfile << i << " ";
         myfile << poses[i].x() << " " << poses[i].y() << " " << poses[i].z() << " ";
-        Vector quaternion = Rot3(R).quaternion();
-        myfile << quaternion(3) << " " << quaternion(2) << " " << quaternion(1)
-               << " " << quaternion(0);
+        Quaternion quaternion = Rot3(R).toQuaternion();
+        myfile << quaternion.x() << " " << quaternion.y() << " " << quaternion.z()
+               << " " << quaternion.w();
         myfile << "\n";
     }
     myfile.close();
--- a/gtsam/geometry/Rot3.cpp
+++ b/gtsam/geometry/Rot3.cpp
@@ -228,6 +228,7 @@ double Rot3::yaw(OptionalJacobian<1, 3> H) const {
 }
 
 /* ************************************************************************* */
+#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V42
 Vector Rot3::quaternion() const {
   gtsam::Quaternion q = toQuaternion();
   Vector v(4);
@@ -237,6 +238,7 @@ Vector Rot3::quaternion() const {
   v(3) = q.z();
   return v;
 }
+#endif
 
 /* ************************************************************************* */
 pair<Unit3, double> Rot3::axisAngle() const {
--- a/gtsam/geometry/Rot3.h
+++ b/gtsam/geometry/Rot3.h
@@ -515,11 +515,16 @@ class GTSAM_EXPORT Rot3 : public LieGroup<Rot3, 3> {
      */
     gtsam::Quaternion toQuaternion() const;
 
+#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V42
     /**
      * Converts to a generic matrix to allow for use with matlab
      * In format: w x y z
+     * @deprecated: use Rot3::toQuaternion() instead.
+     * If still using this API, remind that in the returned Vector `V`,
+     * `V.x()` denotes the actual `qw`, `V.y()` denotes 'qx', `V.z()` denotes `qy`, and `V.w()` denotes 'qz'.
      */
-    Vector quaternion() const;
+    Vector GTSAM_DEPRECATED quaternion() const;
+#endif
 
     /**
      * @brief Spherical Linear intERPolation between *this and other
--- a/gtsam/geometry/geometry.i
+++ b/gtsam/geometry/geometry.i
@@ -355,7 +355,7 @@ class Rot3 {
   double yaw() const;
   pair<gtsam::Unit3, double> axisAngle() const;
   gtsam::Quaternion toQuaternion() const;
-  Vector quaternion() const;
+  // Vector quaternion() const; // @deprecated, see https://github.com/borglab/gtsam/pull/1219
   gtsam::Rot3 slerp(double t, const gtsam::Rot3& other) const;
 
   // enabling serialization functionality


dellaert:
Note you could just re-use this PR.

gtsam_unstable/timing/timeShonanAveraging.cpp
```diff
@@ -79,9 +79,9 @@ void saveG2oResult(string name, const Values& values, std::map<Key, Pose3> poses
         myfile << "VERTEX_SE3:QUAT" << " ";
         myfile << i << " ";
         myfile << poses[i].x() << " " << poses[i].y() << " " << poses[i].z() << " ";
-        Vector quaternion = Rot3(R).quaternion();
-        myfile << quaternion(3) << " " << quaternion(2) << " " << quaternion(1)
-               << " " << quaternion(0);
+        Quaternion quaternion = Rot3(R).toQuaternion();
+        myfile << quaternion.x() << " " << quaternion.y() << " " << quaternion.z()
```
dellaert, line 83: Oh, wow, so this old code was incorrect? As I think this is not equivalent, right?

dellaert (approved):
Awesome, thanks ! Please respond to my comment and then I will merge after CI passes.

gtsam_unstable/timing/timeShonanAveraging.cpp
HViktorTsoi: Yes... according to the [g2o documentation](https://github.com/RainerKuemmerle/g2o/wiki/File-format-slam-3d#3d-robot-pose), the quaternion written to a `.g2o` file should be in the order of `qx,qy,qz,qw`,  while in the old code, it's actually written in the order of `qz,qy,qx,qw`(q3,q2,q1,q0). So I guess that's why the `saveG2oResult()` function in `timeShonanAveraging.cpp` was commented out because it didn't output a g2o file in the correct format? 

BTW, I've watched your lecture 'Pushing Factor Graphs beyond SLAM' before, which also mentioned the Shonan Averaging, very insightful talk!

HViktorTsoi:
Seems all **Python CI** failed, because in the CI configuration `.github/python.sh`, the `GTSAM_ALLOW_DEPRECATED_SINCE_V42` is set to `OFF`, therefore, the `Rot3::quaternion` API that I put into a `GTSAM_ALLOW_DEPRECATED_SINCE_V42` block is ignored when compiling, while the `wrap` Module can still find that API from `geometry.i`, and then generates the `geometry.cpp` with it.

Currently, I have 2 possible solutions for this, and may need some advice from the project maintainers:

1. Remove the `GTSAM_ALLOW_DEPRECATED_SINCE_V42` macro around the `quaternion()` in this fix, and only add a `GTSAM_DEPRECATED` before `Rot3::quaternion`. This is less harmful, but when using the python/matlab bindings, the user can only get the deprecated warning once during compiling;

2. Keep the `GTSAM_ALLOW_DEPRECATED_SINCE_V42` macro, but remove the `Rot3::quaternion()` API from [`geometry.i`](https://github.com/borglab/gtsam/blob/9a1eb022a95368fc5892ca49891fb43ab14abf85/gtsam/geometry/geometry.i#L358). This may cause some API-breaking issues when using then python/matlab bindings (but is trackable).

varunagrawal:
I'm sorry but I am not seeing the argument for this. I just updated some code to use `toQuaternion` and it is less readable.

Previously I could do 
```python
q = wTc.rotation().quaternion()
```

but now we have to do

```python
q = wTb.rotation().toQuaternion()
q = np.asarray([q.w(), q.x(), q.y(), q.z()])
```

which isn't functional since we can't chain methods together anymore e.g.:
```python
do_something(wTb.rotation().quaternion())
```

Moreover, the impression that the current API is confusing doesn't have much evidence other than the single data point.

johnwlambert:
We also rely on the prior functionality in GTSFM, which I believe is quite clean. See: https://github.com/borglab/gtsfm/blob/master/gtsfm/utils/io.py#L374-L377

HViktorTsoi:
> I'm sorry but I am not seeing the argument for this. I just updated some code to use `toQuaternion` and it is less readable.
> 
> Previously I could do
> 
> ```python
> q = wTc.rotation().quaternion()
> ```
> 
> but now we have to do
> 
> ```python
> q = wTb.rotation().toQuaternion()
> q = np.asarray([q.w(), q.x(), q.y(), q.z()])
> ```
> 
> which isn't functional since we can't chain methods together anymore e.g.:
> 
> ```python
> do_something(wTb.rotation().quaternion())
> ```
> 
> Moreover, the impression that the current API is confusing doesn't have much evidence other than the single data point.

As you suggest, for the python/matlab users, the `quaternion` API may be handy in certain circumstances, for it allows the users to use it as an np array; 

However, things are different in C++, because the `quaternion()` actually returns an Eigen Vector with the similar getters `w()` `x()` `y()` `z()` as a quaternion, but the definition is totally different from `qw,qx,qy,qz`, and that is confusing. 

Suppose the following use case: the users want to extract the pose estimation from gtsam and feed it to a ROS odometry message. Without checking the documentation, their IDE gives a coding completion that says there is a `quaternion` method in the Rot3 object and it allows them to access the rotation data. If they don't know the difference between `quaternion()` and `toQuaternion()`, or they don't know how a quaternion is stored inside the object, it is VERY likely to write the following buggy code:
```C++
// pose estimation from previous step
gtsam::Point3 wtc;
gtsam::Rot3 wRc;

nav_msgs::Odometry odom;
// wrong quaternion component order!
odom.pose.orientation.w = wRc.quaternion().w();
odom.pose.orientation.x = wRc.quaternion().x();
odom.pose.orientation.y = wRc.quaternion().y();
odom.pose.orientation.z = wRc.quaternion().z();
```
Everything seems correct, but the ordering of the quaternion doesn't match. And for the debugging process, it will be a nightmare...

varunagrawal:
@HViktorTsoi I see your argument, but the docstring for `Rot3::quaternion` states that it returns an Eigen matrix object.
```
/**
 * Converts to a generic matrix to allow for use with matlab
 * In format: w x y z
 */
```

Moreover, a user should be familiar that a Matrix/Vector is not semantically the same as a quaternion since a vector doesn't enforce the structure of a quaternion. Thus they should already be wary of calling `.w()` on a vector object and double check the documentation.

On the other hand, `toQuaternion` returns a `gtsam::Quaternion` object so this follows the semantics of a quaternion, and thus `.w()`, `.x()` etc would work as expected.

varunagrawal:
> Without checking the documentation, their IDE gives a coding completion that says: there is a quaternion method with w(),x(),y(),z() getters in the Rot3 object

This is the gist of the issue here. The object is not a Rot3 but an Eigen vector so one cannot assume the getters for the vector are the same as the getters for a Rot3/Quaternion object.

HViktorTsoi:
> I'm sorry but I am not seeing the argument for this. I just updated some code to use `toQuaternion` and it is less readable.
> 
> Previously I could do
> 
> ```python
> q = wTc.rotation().quaternion()
> ```
> 
> but now we have to do
> 
> ```python
> q = wTb.rotation().toQuaternion()
> q = np.asarray([q.w(), q.x(), q.y(), q.z()])
> ```
> 
> which isn't functional since we can't chain methods together anymore e.g.:
> 
> ```python
> do_something(wTb.rotation().quaternion())
> ```
> 
> Moreover, the impression that the current API is confusing doesn't have much evidence other than the single data point.

Maybe one can still chain methods without `quaternion()`? If they need a quaternion with the`wxzy` order, it can be written with `toQuaternion().coeffs()` like this:

```python
do_something(wTb.rotation().toQuaternion().coeffs()[[3,0,1,2]])
```
It look a little bit tricky though.

varunagrawal:
Yeah I thought about that but because Eigen uses [x, y, z, w], that is also error prone, exactly like the original issue you faced.

I am terribly sorry you are having difficulties with debugging and other issues, but hopefully we can figure something out that works for everyone!

dellaert:
The previous behavior is *definitely* problematic. Eigen (for better or worse) provides x/y/z/w on vectors, and they conflict with the order in the quaternion method. Just bad.  I think the solution in this PR is good. @johnwlambert , let's just fix GTSFM to not use the deprecated method.

dellaert:
@varunagrawal are we ok to merge?

varunagrawal:
I'm not so sure. @HViktorTsoi has done a brilliant job with this PR but @johnwlambert, myself and others would like to see some way to just get the quaternion vector directly without the 2 step process. :(

varunagrawal:
> The previous behavior is *definitely* problematic. Eigen (for better or worse) provides x/y/z/w on vectors, and they conflict with the order in the quaternion method. Just bad.  I think the solution in this PR is good. @johnwlambert , let's just fix GTSFM to not use the deprecated method.

Okay if you feel this is the correct approach then let's go for it. We should probably decide on a fixed convention and make that the de facto (aka update constructors and other methods).

Thanks for the brilliant PR @HViktorTsoi. You've made our lives so much easier! 😊

PR #1224, opened 2022-06-18
Fix the failure in TA when adding betweenTranslations without any measurements. 

akshay-krishnan:
We were initializing the values for optimization only if they have a relativeTranslation between them. 
So if betweenTranslations contain a node that did not have a relativeTranslation, we would get the "attempting to fetch uninitialized Value" error during optimization. 

Now we initialize the nodes in betweenTranslations as well.


dellaert (approved):
I trust ya :-)

PR #1234, opened 2022-07-04
added jacobians for all pose3 methods in python wrappers

shteren1:
This PR adds jacobian access for all of Pose3 methods in the python wrappers.
I also added Pose3::interp to access Lie group interpolate from the python wrappers.

### DIFF
--- a/python/gtsam/tests/test_Pose3.py
+++ b/python/gtsam/tests/test_Pose3.py
@@ -237,5 +237,4 @@ class TestPose3(GtsamTestCase):
 
 
 if __name__ == "__main__":
-    # unittest.main()
-    TestPose3().test_slerp()
+    unittest.main()


varunagrawal:
Thanks for the PR @shteren1. Is there any reason we can't wrap interpolate directly? The python wrapper supports inheritance.

shteren1:
@varunagrawal Ohh I was not aware of that, I didn't see any wrappings of other Lie group classes.

varunagrawal:
I also recommend adding unit tests in python to verify the wrapping works as expected.

varunagrawal:
Both! And yes numerical derivatives should suffice as a high level check.

gtsam/geometry/Pose3.h
```diff
@@ -379,6 +379,14 @@ class GTSAM_EXPORT Pose3: public LieGroup<Pose3, 6> {
     return std::make_pair(0, 2);
   }
 
+    /**
+   * @brief Spherical Linear interpolation between *this and other
+   * @param s a value between 0 and 1.5
+   * @param other final point of interpolation geodesic on manifold
+   */
+  Pose3 interp(double t, const Pose3& other, OptionalJacobian<6, 6> Hx = boost::none,
```
varunagrawal, line 387: On second thought, I actually prefer this since `Lie::interpolate` is a free function. Can you please rename this to `slerp` similar to `Rot3`?

shteren1: Hmm, isn't "slerp" referring to interpolation of rotations on the quaternion unit sphere specifically?

shteren1:
@varunagrawal added tests for jacobians on some pose3 methods for python.
and changes interp name to slerp

shteren1:
@varunagrawal i'm not sure how they will work, the way gtsam calculates the jacobians of manifold actions on Pose3 class is not trivial, autograds packages will would probably have operated on the euclidean space.

I had to use the retract and localCoordinates actions to get the same results.

and I think JAX is the go to package these days for auto differentiating on numpy and scipy objects, outside of pytorch.

varunagrawal:
Yes I am expecting numerical derivatives on the tangent space directly in the case of numpy.
I'm curious whether Jax supports Lie Groups. I know Tim Barfoot's group has an excellent python based Lie Groups library with Jacobian support.

varunagrawal (approved):
Awesome sauce!

Issue #1174, opened 2022-04-19
Drafting 4.2 Release


ProfFan:
Since now that the annoying TBB bug has been finally resolved (#1158) I think it's probably a good time for a 4.2 release.

Outstanding issues and PRs:
- #1173 
- #292 
- #368
- #1129 
- #1122 
- #1101 
- #1098 
- #846
- #1175 

@varunagrawal @jlblancoc @gchenfc @dellaert @ayushbaid @acxz and all,
feel free to add the issues and PRs you think are important to this list :)

dellaert:
We are finalizing the book draft v1.0 by the end of the semester, so I might need one more alpha release to make sure everything works there.

I will work on #1098 as well.

jlblancoc:
Now that you are considering API break, would it be a good time to merge these two commits, which fixed commonly-found crashes? 
- https://github.com/borglab/gtsam/commit/a29b4dbd7418b6be4748023fd381648b59872314
- https://github.com/borglab/gtsam/commit/07bace5e057d305674b0b17c47a2bc56bb7926a9

(cc: as mentioned https://github.com/borglab/gtsam/issues/292#issuecomment-1009977935 )

@ProfFan You could add https://github.com/borglab/gtsam/pull/1175 to the list, I think it's fairly straightforward and will fix the linked side project.

It would be great to make progress on https://github.com/borglab/gtsam/issues/292 too, but at present I lack spare time for this, sorry...

So good to see the big changes that 4.2 is bringing, btw! :-)

dellaert:
@jlblancoc yeah, feel free to add a small PR with these two commits, although I'd love to have some discussion on native behavior in PR, if only to document.

varunagrawal:
Closing this since 4.2 has been released.

PR #1237, opened 2022-07-07
Jose's updates

varunagrawal:
Adding the updates @jlblancoc made as per https://github.com/borglab/gtsam/issues/1174#issuecomment-1103605148

### DIFF
--- a/cmake/GtsamBuildTypes.cmake
+++ b/cmake/GtsamBuildTypes.cmake
@@ -190,8 +190,9 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
 endif()
 
 if (NOT MSVC)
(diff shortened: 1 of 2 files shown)


ProfFan (changes requested):
I think `march=native` on is a good default, errors related to this are probably deeper issues like different compilation flags that is misunderstood to be a march issue.

I think the Eigen change is good :)

varunagrawal:
> I think `march=native` on is a good default, errors related to this are probably deeper issues like different compilation flags that is misunderstood to be a march issue.

I am no expert on architecture specific code generation, but @jlblancoc seemed to think setting `march=native` off was a good idea.

From a more practical standpoint, I have noticed issues when compiling downstream repos (e.g. GTDynamics, GPMP2 etc) when the `march=native` flag is on, and these are incredibly hard to debug. For this reason, I still think the default should be OFF and those who want the extra boost from architecture specific machine code can turn it on (aka generality over performance).

ProfFan (approved):

PR #1239, opened 2022-07-07
Expose GNC params to python

amadoantonini:


varunagrawal:
@amadoantonini thanks for the PR. Can you please add some python unit tests to verify everything works as expected?

amadoantonini:
@varunagrawal added some tests, let me know if those are ok.

python/gtsam/tests/test_NonlinearOptimizer.py
```diff
@@ -83,6 +84,59 @@ def test_graduated_non_convexity(self):
...
 
+    def test_gnc_params(self):
+        base_params = LevenbergMarquardtParams()
+        # Test base params
+        for base_max_iters in (50, 100):
+            base_params.setMaxIterations(base_max_iters)
+            params = GncLMParams(base_params)
+            self.assertEqual(params.baseOptimizerParams.getMaxIterations(), base_max_iters)
+        # Test printing
+        str(params)
```
varunagrawal, line 95: Is this not testing something?

python/gtsam/tests/test_NonlinearOptimizer.py
```diff
@@ -83,6 +84,59 @@ def test_graduated_non_convexity(self):
...
+        # Test each parameter
+        for loss_type in (GncLossType.TLS, GncLossType.GM):
```
varunagrawal, line 97: Maybe run the yapf formatter on this file? Also please add newlines between each group so that it is easier to read.

python/gtsam/tests/test_NonlinearOptimizer.py
amadoantonini: It was testing that the params object could actually be printed. This would throw an error if not for the default string I added.

Anyways I made the test a little more explicit now.

python/gtsam/tests/test_NonlinearOptimizer.py
amadoantonini: Ran yapf and added some new lines. Let me know if it's not like you expected.

varunagrawal (approved):
Awesome stuff @amadoantonini

varunagrawal:
@amadoantonini can you please resolve the CI failures?

amadoantonini:
I can try. Any ideas as to why those seemingly unrelated tests fail?

python/gtsam/preamble/nonlinear.h
```diff
@@ -9,4 +9,5 @@
  * automatic STL binding, such that the raw objects can be accessed in Python.
  * Without this they will be automatically converted to a Python object, and all
  * mutations on Python side will not be reflected on C++.
- */
\ No newline at end of file
+ */
+#include <pybind11/stl.h>
```
ProfFan, line 13: This is likely the culprit. Automatic STL binding also caused some issues for me in another PR.

amadoantonini: Makes sense. And how have you solved them?

python/gtsam/preamble/nonlinear.h
```diff
@@ -9,4 +9,7 @@
...
```
varunagrawal, line 13: Is this needed? I ran the code in this PR locally and it works without updating this file.

varunagrawal: Yup just get rid of all the changes to this file and you should be golden. This is why unit testing is important. :slightly_smiling_face:

amadoantonini: Interesting, it doesn't work for me without it... it fails [here](https://github.com/borglab/gtsam/pull/1239/files#diff-757771908deb6caea6b6e19556f82810b4cf28d344426be172f961f8c266f214R132). I am running Linux 20.04, though, which I just noticed is not one of the supported platforms. Maybe that's the difference?

I'm happy to remove this from the PR if it works on all the currently supported platforms.

varunagrawal: Do you mean Ubuntu 20.04? I am using the same system and ran it locally to great effect, so it might be some other setting? What compiler are you using? GCC-7?

amadoantonini: Yeah but with gcc-9 (9.4.0 to be specific)

ProfFan:
I think this is the problem:
```
[1981](https://github.com/borglab/gtsam/runs/7333219428?check_suite_focus=true#step:7:1982)
  File "/Users/runner/work/gtsam/gtsam/python/gtsam/tests/test_NonlinearOptimizer.py", line 128, in test_gnc_params
[1982](https://github.com/borglab/gtsam/runs/7333219428?check_suite_focus=true#step:7:1983)
    params.setKnownInliers(inl)
[1983](https://github.com/borglab/gtsam/runs/7333219428?check_suite_focus=true#step:7:1984)
TypeError: setKnownInliers(): incompatible function arguments. The following argument types are supported:
[1984](https://github.com/borglab/gtsam/runs/7333219428?check_suite_focus=true#step:7:1985)
    1. (self: gtsam.gtsam.GncLMParams, knownIn: std::__1::vector<unsigned long, std::__1::allocator<unsigned long> >) -> None
```

`std::__1::vector<unsigned long, std::__1::allocator<unsigned long> >` need to have a type alias and wrapped. This is also essential for the MATLAB wrapper to work.

python/gtsam/preamble/nonlinear.h
ProfFan: Removed it and used the stl_bind instead

ProfFan:
P.S. I have 0 idea why the automatic binding code cause so many hard-to-debug issues. I encountered an infinity loop issue that is caused by `stl.h` when wrapping the hybrid code, took me a week to debug. Should we try upgrade the pybind11 lib and see if problems disappear? @varunagrawal

varunagrawal:
This looks like the Jacobians are not being copied back to the C++ side of things. I guess we have to look at how the opacity is being set.

This is also a good time to revisit my PR on using the stl bindings fully.

varunagrawal:
@amadoantonini the issue seems to be tangential to your fix. @ProfFan and I are working on figuring this out, so we thank you for your patience!

varunagrawal:
@amadoantonini I opened a PR to fix your issue https://github.com/openspacelabs/gtsam/pull/1
Can you please merge it in? It should automatically reflect here.

varunagrawal:
@amadoantonini I am starting to think this was a wild goose chase. You can remove `#include <pybind11/stl.h>` from `nonlinear.h` and everything should work. It seems like Pybind already recognizes `vector<size_t>` which I guess is from `vector<Key>` since Key is just a uint64_t.

Anyway, please make the above change and CI should pass!

varunagrawal:
Ahhh I'm going to have to look into this more.

amadoantonini:
Thanks @varunagrawal! Let me know if I can help.

varunagrawal:
@amadoantonini can you please merge in the latest develop? We made some CI changes to fix issues we've been seeing lately (this PR included).

amadoantonini:
That seems to have fixed it on my side!

varunagrawal:
@amadoantonini gentle reminder about https://github.com/openspacelabs/gtsam/pull/2

varunagrawal:
@amadoantonini can you please merge in the latest develop? Github Actions is being funky.

amadoantonini:
@varunagrawal I had to resolve a conflict. I think I did the right thing based on the order of your changes. Please verify.

varunagrawal:
Yup looks good to me.

varunagrawal:
@amadoantonini please merge this https://github.com/openspacelabs/gtsam/pull/3 and please allow maintainers to make edits so I can push here directly.

varunagrawal:
Yay finally merged!

amadoantonini:
Thanks for all the support @varunagrawal !

PR #1240, opened 2022-07-07
Wrap BearingRange Factors

### LINKED ISSUE #912: Bearing range 3D factor in python
hi,

Thanks for the awesome work!

I wonder if the BearingRangeFactor for Pose3 and Point3 is also ported to Python? I install gtsam via pip and it doesn't have BearingRangeFactor3D.

varunagrawal:
Fixes #912


ProfFan (approved):
Thanks @varunagrawal !

PR #1248, opened 2022-07-16
Clear ConcurrentMap before loading archive data

ScottMcMichael:
Adding the clear() call causes the contents of the ConcurrentMap to be replaced by the archive data, without it the archive data is appended to the current object contents.


varunagrawal:
@ScottMcMichael thanks for the PR! Can you please add a small unit test to ensure we can catch this in the future? You'll be able to find other tests for serialization in the repo.

varunagrawal (approved):
LGTM! Thanks @ScottMcMichael

PR #1252, opened 2022-07-23
replace addtogroup with ingroup for the SLAM group

agilemapper:
I was recently browing GTSAM doxygen documentation and I noticed that the detailed description for some classes is not present in the generated website. For instance here for [CombinedImuFactor](https://gtsam.org/doxygen/a04060.html#a5edf3d620bbec01e8c9f24146a79a251).

Instead the docs are kind of collated into the separate [SLAM](https://gtsam.org/doxygen/a01590.html) page, due to the `@addtogroup` command. Not an expert on doxygen, but it looks like an `@ingroup` is a better suited command, since it retains the detailed description with the class itself, and only adds a link to the group page.

With these changes, the new version looks like this:

combined-imu-factor

And the group page looks like this:

slam-page

Hope that improves the docs a bit.

### DIFF
--- a/gtsam/navigation/CombinedImuFactor.h
+++ b/gtsam/navigation/CombinedImuFactor.h
@@ -123,7 +123,7 @@ public:
  * it is received from the IMU) so as to avoid costly integration at time of
  * factor construction.
  *
- * @addtogroup SLAM
+ * @ingroup SLAM
  */
 class GTSAM_EXPORT PreintegratedCombinedMeasurements : public PreintegrationType {
 
@@ -252,7 +252,7 @@ public:
  *    the correlation between the bias uncertainty and the preintegrated
  *    measurements uncertainty.
  *
- * @addtogroup SLAM
+ * @ingroup SLAM
  */
 class GTSAM_EXPORT CombinedImuFactor: public NoiseModelFactor6<Pose3, Vector3, Pose3,
     Vector3, imuBias::ConstantBias, imuBias::ConstantBias> {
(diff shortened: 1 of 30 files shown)


varunagrawal:
Interesting. I am wondering if we should use this opportunity to set the groups correctly, since the IMU factors should be in Navigation?

varunagrawal:
@agilemapper I want to do some checking on the generated doxygen docs and verify if we're doing the right things wrt doxygen, so while this PR looks great to me, I'll approve it once I get some better understanding, since I want to use this is a discussion platform to get our docs in good shape.

Thanks for your PR and the help!

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -123,7 +123,7 @@ struct GTSAM_EXPORT PreintegrationCombinedParams : PreintegrationParams {
  * it is received from the IMU) so as to avoid costly integration at time of
  * factor construction.
  *
- * @addtogroup SLAM
+ * @ingroup SLAM
```
varunagrawal, line 126: Can you please update the group to match the directory name?
E.g. for this file, it should be `@ingroup Navigation`.

gtsam/nonlinear/PriorFactor.h
```diff
@@ -24,7 +24,7 @@ namespace gtsam {
 
   /**
    * A class for a soft prior on any Value type
-   * @addtogroup SLAM
+   * @ingroup SLAM
```
varunagrawal, line 27: Similarly this belongs to `nonlinear` so it should be `@ingroup Nonlinear`.

varunagrawal (approved):
@agilemapper this works amazingly well. I just have a small request: can you please update the group names to match the directory? I have given some examples in my review comments.

varunagrawal:
I'm going to merge this since I have an idea to improve upon this.

PR #1254, opened 2022-07-26
Break up Python CI steps so reading logs is easier

varunagrawal:
This PR makes the Python CI run smaller steps so we're not inundated with information when debugging the CI.

### DIFF
--- a/.github/scripts/python.sh
+++ b/.github/scripts/python.sh
@@ -100,6 +100,7 @@ function test()
 case $1 in
   -d)
     install_dependencies
+    ;;
   -b)
     build
     ;;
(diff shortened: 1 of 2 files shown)


gchenfc:
To clarify, this is purely cosmetic, i.e. only separates the output into 2 different steps so that we can read it easier, but does not change the computation that occurs?

Because one of the things I was thinking about was to split up the cpp compiling from the wrapper compiling so that cpp could be -j2 then switch wrapper to -j1, or something like that.  But maybe that's not necessary since the swap space thing seems to have done the trick.

varunagrawal:
Yeah any computation changes would be a follow up PR. 🙂

PR #1255, opened 2022-07-26
Doxygen Updates

varunagrawal:
Based on #1252, I made a bunch of improvements to doxygen related doc generation.

1. Defined different groups corresponding to the directories under `gtsam`.
2. Made `GTSAM_UNSTABLE` a `page` since doxygen doesn't like having 2 `mainpage`s.
3. Fixed a whole bunch of grouping issues.
4. Fixed a whole bunch of formatting issues.
5. Replaced `addtogroup` with `ingroup` so that the docs render nicely.
6. Upgraded the `Doxyfile` config file.
7. Updated CMake to have the correct directory listing to look for docstrings.
8. Fixed some docstring typos.
9. Fixed incorrect filenames.


gtsam/navigation/CombinedImuFactor.h
```diff
@@ -252,7 +252,7 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurements : public PreintegrationType
  *    the correlation between the bias uncertainty and the preintegrated
  *    measurements uncertainty.
  *
- * @ingroup SLAM
+ * @ingroup navigation
```
ProfFan, line 255: does case matter in the final doc?

varunagrawal: No but those are labels and I liked the labels to be lower case. Helps differentiate between the actual titles in the docs.

jlblancoc (approved):
LGTM

PR #1109, opened 2022-02-17
Windows Tests

varunagrawal:
Thanks to the awesome work by @mikesheffler, I was finally able to figure out what was the issue behind all the "undefined" errors.

This PR adds/removes all the necessary `GTSAM_EXPORT` declarations and updates the CI to run all `GTSAM` test targets. Unfortunately some targets have test failures due to bugs creeping in over time and the CI not being enabled for them, but those can be tackled later.

I am leaving `GTSAM_UNSTABLE` test targets for another day.

PR #1257, opened 2022-07-31
Fix exports on Windows

Tobias-Fischer:
Hi there,

This is the first of a few PRs which emerged from our efforts to package gtsam in conda-forge (see https://github.com/conda-forge/gtsam-feedstock/tree/main/recipe).

Related: #1109 

This one is relatively simple - there is no associated cpp file that exports symbols, so GTSAM_EXPORT needs to be removed in these two instances.


varunagrawal (approved):
Great catch! Thanks @Tobias-Fischer

PR #1258, opened 2022-07-31
Link to interface library in case system metis is used

Tobias-Fischer:
This test links to `metis-gtsam`, which only exists if the vendored metis is being built. If the system metis is used, only `metis-gtsam-if` exists (and it also exists in the vendor case, so all good to change).

Related: #1257 - this issue was found as part of packaging gtsam for conda-forge.


varunagrawal:
I'll approve once CI passes so we can merge with confidence :slightly_smiling_face:

varunagrawal (approved):
LGTM

PR #1265, opened 2022-08-04
Add Zenodo for citations

varunagrawal:
This PR adds a link to GTSAM's Zenodo page where citation text is auto-generated. This way, one can cite GTSAM easily in academic work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,24 @@ We provide support for [MATLAB](matlab/README.md) and [Python](python/README.md)
 
 ## Citation
 
-If you are using GTSAM for academic work, please use the citation available from Zenodo below:
+If you are using GTSAM for academic work, please use the following citation:
 
-[![DOI](https://zenodo.org/badge/86362856.svg)](https://zenodo.org/badge/latestdoi/86362856)
+```
+@software{gtsam,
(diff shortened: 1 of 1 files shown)


gchenfc:
This is cool but, personally, I think it's best @dellaert review this (not sure if you already coordinated with him independently or not), and make sure that this citation format:
1. will be maintained long-term; I'm not sure how established zenodo is or not.
2. is compatible / links with existing gtsam citations
3. will citation trackers link different releases to the same citations?  I feel like it would be strange to have 10 doi's with 10 citations each that don't get merged into one reference with 100 citations, but idk.

But these are just my personal thoughts.  In any case, having an easy "please use this citation", either in text/bibtex/some form would be great.

varunagrawal:
1. Zenodo is used by CERN and lots of organizations in the sciences to archive software. GitHub itself endorses it, so I don't think it's going away anytime soon. I wouldn't have suggested using it if it wasn't well-established already.
2. I believe current citations just use Frank's book? There has never been a citation for the software repo itself (unless you crafted one manually).
3. Zenodo has something called Concept version which is a DOI to all the versions cumulatively. This should ensure we don't have issues with different versions having different citations. I'll double check and update the PR to ensure this is the case.

gchenfc (approved):

PR #1270, opened 2022-08-15
Linear HybridBayesNet optimization

xsj01:
This PR adds the following:

1. optimize() in HybridBayesNet. It finds the argmax by back-substitution.
2. HybridValues class to store the optimization results.


varunagrawal:
@xsj01 can you please run a formatter on your changed code? That should make reading and reviewing it easier.

gtsam/hybrid/HybridLookupDAG.cpp
```diff
@@ -0,0 +1,77 @@
...
+    // solve the resulting GaussianConditional.
+    auto conditional =
+        boost::static_pointer_cast<GaussianMixture>(inner_)->conditionals()(
+            values->discrete);
+    values->continuous.insert(conditional->solve(values->continuous));
+  }
+}
+
+// /* **************************************************************************
+// */
```
ProfFan, line 59: Fix comment style

gtsam/hybrid/HybridLookupDAG.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+HybridValues HybridLookupDAG::argmax(HybridValues result) const {
+  // Argmax each node in turn in topological sort order (parents first).
+  for (auto lookupTable : boost::adaptors::reverse(*this))
+    lookupTable->argmaxInPlace(&result);
+  return result;
+}
+/* ************************************************************************** */
+
+}  // namespace gtsam
```
ProfFan, line 77: newline

gtsam/hybrid/HybridValues.h
```diff
@@ -0,0 +1,127 @@
...
+#include <vector>
+
+namespace gtsam {
+
+/**
+ * HybridValues represents a collection of DiscreteValues and VectorValues.  It
+ * is typically used to store the variables of a HybridGaussianFactorGraph.
+ * Optimizing a HybridGaussianBayesNet returns this class.
+ */
+class GTSAM_EXPORT HybridValues {
```
ProfFan, line 38: for thoughts: I always wondered can we just get rid of `DiscreteKey` and stuff, and store the cardinality of variables somewhere else, like along with the value itself.

gtsam/hybrid/HybridValues.h
```diff
@@ -0,0 +1,127 @@
...
+ public:
+  // DiscreteValue stored the discrete components of the HybridValues.
+  DiscreteValues discrete;
+
+  // VectorValue stored the continuous components of the HybridValues.
+  VectorValues continuous;
```
ProfFan, line 44: Should we name the class `HybridVectorValues`? This class is linear only. (comment only)

gtsam/hybrid/hybrid.i
```diff
@@ -4,6 +4,22 @@
 
 namespace gtsam {
 
+#include <gtsam/hybrid/HybridValues.h>
+class HybridValues {
```
ProfFan, line 8: Can you add some unit tests in Python?

ProfFan (changes requested):

gtsam/hybrid/HybridValues.h
xsj01: Maybe I can change VectorValues to Values?

ProfFan: I think this is more like a design question: Is this limited to linear hybrid systems? If so we should name it Gaussian.

xsj01: I think we will extend that for nonlinear hybrid later, but currently it's only for linear hybrid system. 
If we rename this to HybridVectorValues, then we will also need HybridValues for nonlinear hybrid system later. Is it necessary to have both HybridVectorValues and HybridValues?

ProfFan: This is the spirit for `linear` and `nonlinear` though. because VectorValues and values are different we should probably have the same

ProfFan (approved):
LGTM
tested locally, seems to be working

ProfFan:
@xsj01

varunagrawal:
I think this was merged too soon. I didn't get to take a good look at it. :(
Also, @xsj01 please delete branches after merging.

ProfFan:
We can change this later, btw @xsj01 there is also `optimize` for `HybridBayesTree`, which is harder :)

gtsam/hybrid/HybridLookupDAG.h
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * @brief Construct a new Hybrid Lookup Table object form a HybridConditional.
+   *
+   * @param conditional input hybrid conditional
+   */
+  HybridLookupTable(HybridConditional& conditional) : Base(conditional){};
+
+  /**
+   * @brief Calculate assignment for frontal variables that maximizes value.
```
varunagrawal, line 55: @xsj01 What would be the frontal variables here? They should all be continuous variables since we eliminate those first, so then this doesn't make sense.

gtsam/hybrid/HybridValues.h
varunagrawal, line 38: I don't understand the use of this class. Can't we simply have `optimize` run MPE on the discrete tree and then run a regular optimization on the continuous values corresponding to the assignment? That way we can just return a `pair<Values, DiscreteValues>`.

gtsam/hybrid/HybridLookupDAG.h
```diff
@@ -0,0 +1,119 @@
...
+   * @param (in/out) parentsValues Known assignments for the parents.
+   */
+  void argmaxInPlace(HybridValues* parentsValues) const;
+};
+
+/** A DAG made from hybrid lookup tables, as defined above. Similar to
+ * DiscreteLookupDAG */
+class GTSAM_EXPORT HybridLookupDAG : public BayesNet<HybridLookupTable> {
```
varunagrawal, line 63: The docs for this class are severely lacking. @xsj01 you have to be better than just copy-pasting the same docstrings, since now the motivation for this class is lost.

varunagrawal (commented):
Some thoughts and questions. Unfortunately this PR was merged too soon. 😞

varunagrawal:
> We can change this later, btw @xsj01 there is also `optimize` for `HybridBayesTree`, which is harder :)

Not quite the point. There has to be coherence so that everyone working on hybrid estimation understands the overall design and there is less wasted effort. Going back and forth won't solve issues with design.

I had requested a PR from @xsj01 so I could see his work and understand his requirements, with the hope that it would lead to better overall structure.

varunagrawal:
Also the tests don't have any docs. :(

varunagrawal:
@xsj01 can you please respond with a plan to tackle these issues?

xsj01:
Sorry for the delay @varunagrawal, I will address the comments. Just wondering shall we revert this PR and merge it later?

gtsam/hybrid/HybridValues.h
xsj01: Yeah, I think that's also an option. But I feel like the api is not very user friendly while accessing/assigning values. We can talk about it in our meeting.

gtsam/hybrid/HybridLookupDAG.h
xsj01: Yeah, it seems the HybridLookupTable is not necessary. Just had some discussions with Fan, I will submit a new PR to fix this.

PR #1273, opened 2022-08-21
Hybrid ISAM

varunagrawal:
This PR gets the last non-trivial test for Hybrid GaussianISAM to pass, and adds necessary updates to avoid crashes and incorrect results.

After this, hybrid elimination should be good to go!

### DIFF
--- a/gtsam/hybrid/HybridGaussianFactorGraph.cpp
+++ b/gtsam/hybrid/HybridGaussianFactorGraph.cpp
@@ -96,8 +96,15 @@ GaussianMixtureFactor::Sum sumFrontals(
       }
 
     } else if (f->isContinuous()) {
-      deferredFactors.push_back(
-          boost::dynamic_pointer_cast<HybridGaussianFactor>(f)->inner());
+      // Check if f is HybridConditional or HybridGaussianFactor.
+      if (auto hc = boost::dynamic_pointer_cast<HybridConditional>(f)) {
+        auto conditional =
+            boost::dynamic_pointer_cast<GaussianConditional>(hc->inner());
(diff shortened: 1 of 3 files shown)


gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -96,8 +109,15 @@ GaussianMixtureFactor::Sum sumFrontals(
       }
 
     } else if (f->isContinuous()) {
-      deferredFactors.push_back(
-          boost::dynamic_pointer_cast<HybridGaussianFactor>(f)->inner());
+      // Check if f is HybridConditional or HybridGaussianFactor.
+      if (auto hc = boost::dynamic_pointer_cast<HybridConditional>(f)) {
```
ProfFan, line 113: This should just work without explicit casting as `HybridConditional` is a subclass of HybridGaussianConditional

varunagrawal: Yeah I don't remember why I did this. Guess I was debugging something.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -77,6 +77,19 @@ static GaussianMixtureFactor::Sum &addGaussian(
 }
 
 /* ************************************************************************ */
+/* Function to eliminate variables **under the following assumptions**:
```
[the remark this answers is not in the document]
varunagrawal: Oh okay. This was the function where we had the long discussion on the other PR so I figured this is where you wanted it.
Should we add another comment for `sumFrontals`?

PR #1277, opened 2022-08-22
Additional Nonlinear Hybrid

varunagrawal:
Fix some weird PR juju. These changes have already been approved in a previous PR.


gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -195,7 +208,10 @@ hybridElimination(const HybridGaussianFactorGraph &factors,
     if (graph.empty()) {
       return {nullptr, nullptr};
     }
-    auto result = EliminatePreferCholesky(graph, frontalKeys);
```
ProfFan, line 198: I think auto will just work since we are on GCC7

ProfFan (approved):

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: I prefer the readability. Made debugging easier.

PR #1280, opened 2022-08-26
Hybrid Bayes Net/Tree Optimize

varunagrawal:
This PR adds `optimize` methods for the Hybrid Bayes Net and the Hybrid Bayes Tree.

The idea is that if we know the optimal discrete assignment, we can pick out the Gaussian Bayes Net and just optimize that.

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -125,8 +125,14 @@ GaussianBayesNet HybridBayesNet::choose(
     const DiscreteValues &assignment) const {
   GaussianBayesNet gbn;
   for (size_t idx = 0; idx < size(); idx++) {
-    GaussianMixture gm = *this->atGaussian(idx);
-    gbn.push_back(gm(assignment));
+    try {
+      GaussianMixture gm = *this->atGaussian(idx);
+      gbn.push_back(gm(assignment));
+
+    } catch (std::exception &exc) {
+      // if factor at `idx` is discrete-only, just continue.
+      continue;
+    }
   }
   return gbn;
 }
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -16,8 +16,8 @@
  */
 
 #include <gtsam/hybrid/HybridBayesNet.h>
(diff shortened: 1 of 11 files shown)


ProfFan:
Optimal discrete assignment needs to be solved by using MPE, CC @xsj01

gtsam/hybrid/HybridBayesTree.cpp
```diff
@@ -68,6 +68,13 @@ VectorValues HybridBayesTree::optimize(const DiscreteValues& assignment) const {
...
   }
+  // If TBB is enabled, the bayes net order gets reversed,
+  // so we pre-reverse it
+#ifdef GTSAM_USE_TBB
+  auto reversed = boost::adaptors::reverse(gbn);
+  gbn = GaussianBayesNet(reversed.begin(), reversed.end());
+#endif
+
   // Return the optimized bayes net.
   return gbn.optimize();
```
ProfFan, line 79: Does this mean that `gbn.optimize` does not work when the bayes net has non sequential order?

varunagrawal: The GBN needs to be topologically sorted so that it optimizes parents before children, and for some reason, the GBN is reverse topologically sorted only when TBB is enabled.

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -112,22 +112,27 @@ HybridBayesNet HybridBayesNet::prune(
...
   GaussianBayesNet gbn;
   for (size_t idx = 0; idx < size(); idx++) {
-    GaussianMixture gm = *this->atGaussian(idx);
-    gbn.push_back(gm(assignment));
+    try {
+      GaussianMixture gm = *this->atGaussian(idx);
+      gbn.push_back(gm(assignment));
+
+    } catch (std::exception &exc) {
+      // if factor at `idx` is discrete-only, just continue.
```
ProfFan, line 133: assert that idx is indeed discrete only

ProfFan (approved):

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: Made the assertion in #1282

PR #1281, opened 2022-08-26
Add wrapping for hybrid nonlinear

ProfFan:
Add wrapping for hybrid nonlinear so we can do inference in Python.


varunagrawal (approved):
LGTM

PR #1285, opened 2022-08-31
Rewrap KeyVector for Matlab wrapper

varunagrawal:
Wrap the KeyVector for Matlab since I accidentally removed it in a previous commit.


mattking-smith:
Yeah I am getting stuck with the following error in MATLAB: 
image

So we are going to need to debug this a little further.

mattking-smith (approved):
Tested and verified :). Looks good to go.

PR #1287, opened 2022-09-02
Fix HybridBayesTree Optimize

varunagrawal:
This PR fixes an issue in the HybridBayesTree optimization by traversing the tree rather than looping over all the nodes.
Doing this ensures we don't optimize a child before its parent aka we maintain the reverse topological sort of the BayesTree.

### DIFF
--- a/gtsam/hybrid/HybridBayesTree.cpp
+++ b/gtsam/hybrid/HybridBayesTree.cpp
@@ -24,6 +24,7 @@
 #include <gtsam/hybrid/HybridBayesTree.h>
 #include <gtsam/inference/BayesTree-inst.h>
 #include <gtsam/inference/BayesTreeCliqueBase-inst.h>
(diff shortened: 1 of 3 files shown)


gtsam/hybrid/HybridBayesTree.cpp
```diff
@@ -24,6 +24,7 @@
 #include <gtsam/hybrid/HybridBayesTree.h>
 #include <gtsam/inference/BayesTree-inst.h>
 #include <gtsam/inference/BayesTreeCliqueBase-inst.h>
```
ProfFan, line 26: move inst to bottom

ProfFan (approved):
Looks great!

gtsam/hybrid/HybridBayesTree.cpp
varunagrawal: I tried but the formatter keeps moving it back 😕

Issue #1297, opened 2022-09-27
gtsam should use eigen cmake config file


OznOg:
## Feature

gtsam uses a custom FindEigen3.cmake but eigen3 already provides a Eigen3Config.cmake for eigen detection.

## Motivation

Using vanilla cmake detection will prevent maintaining code and overall prevent bad usage and detection of the library:

for example, path to the eigen lib eventually get embeded in the gtsamConfig.cmake which should not be the case (gtsam should use find_dependency() macro instead so it would not break when using packaged version of eigen)

Moreover, gtsam shoud use Eigen3::eigen library names instead of forcing eigen include path.

## Alternatives

have an hybrid approach trying to use the vanilla eigenConfig of the system and fallback on the custom one if nothing is found

## Additional context

Useful when working with system eigen and trying to package gtsam (rpm, deb, conan etc...)

ProfFan:
Yeah but we don't have bandwidth for this, but contributions are always welcome!

OznOg:
what is the oldest eigen version supposed to be supported by current gtsam?

OznOg:
looks like to be 3.3.0, any one can confirm, please?

ProfFan:
3.3.4-4 as in Ubuntu 18.04

varunagrawal:
@OznOg has this been a problem for you? This seems to be more of an opinion rather than a bug or issue.

OznOg:
This is not, I'm working to integrate gtsam on yocto and I had to patch it for this, this is why I'm looking forward having the code patched upstream.
Moreover *any* packager will encounter the very same problem as the path put in the includes of the target references the build tree which it should not.

varunagrawal:
Awesome. Thanks for the info and the PR @OznOg :)

OznOg:
welcome! If I write more patches for yocto tht make sens in the mainline, I'll open another ticket.

PR #1304, opened 2022-10-07
Use eigen3 config.cmake

OznOg:
This fixes bug #1297


cmake/HandleEigen.cmake
```diff
@@ -14,10 +14,13 @@ endif()
...
 # Switch for using system Eigen or GTSAM-bundled Eigen
 if(GTSAM_USE_SYSTEM_EIGEN)
-    find_package(Eigen3 REQUIRED) # need to find again as REQUIRED
+    find_package(Eigen3 CONFIG REQUIRED) # need to find again as REQUIRED
 
-    # Use generic Eigen include paths e.g. <Eigen/Core>
-    set(GTSAM_EIGEN_INCLUDE_FOR_INSTALL "${EIGEN3_INCLUDE_DIR}")
+    # The actual include directory (for BUILD cmake target interface):
+    # Node Eigen cmake is completelly broken on some version,
+    # EIGEN3_INCLUDE_DIR is just unusable as it points to some random location.
```
varunagrawal, line 21: Does this mean there is no guarantee this will work everywhere and for everyone?

OznOg: This means it is really ugly, and should not be necessary on some versions. This works since eigen 3.3.0. Prior Eigen will not, but is not supported anyway AFAIK.

varunagrawal: Then we should probably replace the text with a comment stating that it will work for Eigen 3.3.0 onwards but may not for prior versions. Basically what you said in your comment. 🙂

OznOg: This is also in the commit message, but I'll update the comment.

OznOg: done

varunagrawal:
@OznOg this PR is great. I just need to test it out with the Matlab wrapper (we don't have a CI for it) and if that works, I'll approve and merge. :)

OznOg:
> @OznOg this PR is great. I just need to test it out with the Matlab wrapper (we don't have a CI for it) and if that works, I'll approve and merge. :)

Ok, good news

OznOg:
Did you have time to test with matlab? any issue?

varunagrawal:
@OznOg I've been having issues with my Matlab machine which should be resolved by this weekend. Apologies for the delay, this is my fault.

dellaert:
Letting @varunagrawal continue the review, but one question I had: there are two Eigen versions people can use: the one included with GTSAM, and system. Will this PR work with both? I know both options are being used. Not sure whether our CI is good at testing both :-/

varunagrawal:
> Letting @varunagrawal continue the review, but one question I had: there are two Eigen versions people can use: the one included with GTSAM, and system. Will this PR work with both? I know both options are being used. Not sure whether our CI is good at testing both :-/

@dellaert yes the CI tests for System Eigen as well. It's in `build-special.yml` under the `ubuntu-system-libs` workflow. :slightly_smiling_face:

varunagrawal:
Matlab tests pass! Approving.

varunagrawal (approved):
LGTM

dellaert:
Uhm, now can't compile anymore:

```
In file included from /Users/dellaert/git/github/gtsam/base/Matrix.h:28:
/Users/dellaert/git/github/gtsam/base/Vector.h:74:1: error: static_assert failed due to requirement '3 == 3 && 3 == 4' "Error: GTSAM was built against a different version of Eigen"
static_assert(
```

What's the fix?

jlblancoc:
Hi Frank! Was it building the gtsam lib itself, or an example/app?
Have you tried checking what's the directory shown by `cd BuildDir && grep Eigen CMakeCache.txt`?

(Note: I've not tested the merged version yet!)

jlblancoc:
@OznOg @varunagrawal : Is there any reason to leave a copy of Eigen3 under `gtsam/3rdparty`? I think it's useless "as is" right now, because it must be processed by cmake first to produce a usable `EigenConfig.cmake` file... Perhaps it could be removed and installation instructions updated to remember Eigen is a mandatory prerequisite, just like Boost?

varunagrawal:
Seems like CMake is either picking up a different version of Eigen (3.4 rather than the provided 3.3). Are you on a M1 Mac?

varunagrawal:
> @OznOg @varunagrawal : Is there any reason to leave a copy of Eigen3 under `gtsam/3rdparty`? I think it's useless "as is" right now, because it must be processed by cmake first to produce a usable `EigenConfig.cmake` file... Perhaps it could be removed and installation instructions updated to remember Eigen is a mandatory prerequisite, just like Boost?

I believe the idea is that by including Eigen under `3rdparty`, we can compile out of the box and have some control over the Eigen version. @dellaert prefers this approach since it requires no wasted effort in integration.

varunagrawal:
@dellaert FWIW I compiled (and cleaned and compiled) the latest develop on my M1 mac and I don't seem to have this problem. You may want to do a full rebuild?

varunagrawal:
I think I found the issue. Line 5 in `HandleEigen.cmake` sets `USE_SYSTEM_EIGEN_INITIAL_VALUE` to ON which tells Cmake to use the system Eigen. Looks like @OznOg's PR unearthed a bug in the cmake. 🙂 

@dellaert you should be able to compile by setting the `GTSAM_USE_SYSTEM_EIGEN` flag to false.

gtsam/CMakeLists.txt
```diff
@@ -117,12 +117,9 @@ set_target_properties(gtsam PROPERTIES
     VERSION             ${gtsam_version}
     SOVERSION           ${gtsam_soversion})
 
-# Append Eigen include path, set in top-level CMakeLists.txt to either
 # system-eigen, or GTSAM eigen path
-target_include_directories(gtsam PUBLIC
-  $<BUILD_INTERFACE:${GTSAM_EIGEN_INCLUDE_FOR_BUILD}>
-  $<INSTALL_INTERFACE:${GTSAM_EIGEN_INCLUDE_FOR_INSTALL}>
-)
+target_link_libraries(gtsam PUBLIC Eigen3::Eigen)
```
varunagrawal, line 121: Found the issue. It's this line making GTSAM link against the system Eigen always instead of the provided Eigen even when the system Eigen flag is set to OFF.

dellaert:
> I think I found the issue. Line 5 in `HandleEigen.cmake` sets `USE_SYSTEM_EIGEN_INITIAL_VALUE` to ON which tells Cmake to use the system Eigen. Looks like @OznOg's PR unearthed a bug in the cmake. 🙂
> 
> @dellaert you should be able to compile by setting the `GTSAM_USE_SYSTEM_EIGEN` flag to false.

That does not work, but #1314 does fix it for me. I'll approve and merge that one...

OznOg:
Hi guys, is there something I can/need to do to help, or is everything ok now?

varunagrawal:
Everything has been figured out. Thanks for following up. :)

PR #1306, opened 2022-10-13
Hybrid Updates

varunagrawal:
1. Move pruner functor to `GaussianMixture` so we remove duplication and use the same API for both the `HybridBayesTree` and the `HybridBayesNet`.
2. Fix issue where pruning doesn't happen for subset of discrete keys (based on @ProfFan's idea).
3. Various formatting and niggles (`GaussianMixtureFactor.cpp`, `HybridFactor.cpp`, `HybridGaussianFactorGraph.cpp`, `HybridGaussianISAM.cpp`).
4. Add key formatter to `HybridNonlinearISAM::print`.
5. Fix compile time warnings from `testDiscreteBayesTree.cpp`.


gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -128,25 +128,81 @@ void GaussianMixture::print(const std::string &s,
...
+/**
+ * @brief Helper function to get the pruner functional.
+ *
+ * @param decisionTree The probability decision tree of only discrete keys.
+ * @return std::function<GaussianConditional::shared_ptr(
+ * const Assignment<Key> &, const GaussianConditional::shared_ptr &)>
+ */
+std::function<GaussianConditional::shared_ptr(
+    const Assignment<Key> &, const GaussianConditional::shared_ptr &)>
+GaussianMixture::prunerFunc(const DecisionTreeFactor &decisionTree) {
```
ProfFan, line 149: I feel that this should be using the full discrete bayes tree if one wants to be more eager in pruning

ProfFan (approved):

gtsam/hybrid/GaussianMixture.cpp
varunagrawal: The discrete Bayes tree/net is what is being passed in.
Or am I misunderstanding your comment?

PR #1308, opened 2022-10-15
CombinedImuFactor fixes

varunagrawal:
1. Move `PreintegratedCombinedParams` to its own file so it has the same structure as `PreintegratedImuParams`. This is fully backwards compatible.
2. Rename some of the jacobian variables in noise propagation to better match the math in ImuFactor.pdf.
3. Update `PreintegratedCombinedMeasurements::resetIntegration` to set the initial bias covariance either from the last step or from the user specified value.


gtsam/navigation/CombinedImuFactor.cpp
```diff
@@ -72,6 +72,17 @@ bool PreintegratedCombinedMeasurements::equals(
 
 //------------------------------------------------------------------------------
 void PreintegratedCombinedMeasurements::resetIntegration() {
+  // Set the initial bias covariance to
+  // the estimated covariance from the last step.
+  p().biasAccOmegaInt = preintMeasCov_.block<6, 6>(9, 9);
```
raabuchanan, line 77: I don't think this is correct. The propagated noise is not the same as the uncertainty on the bias at t1. Ideally we would use the marginal covariance on the bias but we can't expect the information to be available. So I think when resetting without any arguments the biasAccOmegaInt should be left alone.

gtsam/navigation/CombinedImuFactor.cpp
```diff
@@ -72,6 +72,17 @@ bool PreintegratedCombinedMeasurements::equals(
...
+  PreintegrationType::resetIntegration();
+  preintMeasCov_.setZero();
+}
+
+//------------------------------------------------------------------------------
+void PreintegratedCombinedMeasurements::resetIntegration(
+    const gtsam::Matrix6& Q_init) {
+  p().biasAccOmegaInt = Q_init;
```
raabuchanan, line 85: Shouldn't this line go after the one below it? Otherwise biasAccOmegaInt just gets reset again according to your changes to resetIntegration() ?

gtsam/navigation/tests/testCombinedImuFactor.cpp
```diff
@@ -316,6 +317,38 @@ TEST(CombinedImuFactor, Accelerating) {
...
+
+  CombinedScenarioRunner runner(scenario, testing::Params(), T / 10);
+
+  PreintegratedCombinedMeasurements pim = runner.integrate(T);
+  // Make copy for testing different conditions
+  PreintegratedCombinedMeasurements pim2 = pim;
+
+  // Test default method
+  pim.resetIntegration();
+  Matrix6 expected = Z_6x6;
```
raabuchanan, line 343: Do we expect zero? I thought you were setting it to `preintMeasCov_.block<6, 6>(9, 9);` ?

python/gtsam/examples/CombinedImuFactorExample.py
```diff
@@ -0,0 +1,251 @@
+"""
```
raabuchanan, line 1: Is this a brand new python example?

varunagrawal: Yup. I wanted to add one so we can do a quick comparison with the ImuFactor example.

gtsam/navigation/CombinedImuFactor.cpp
varunagrawal: When propagating the noise, shouldn't this be the new (propagated) biasAccOmegaInit after the last preintegration step?
Please correct me if I am wrong, but my understanding was that we should automatically be picking the updated/propagated biasAccOmegaInit from the last time step (t_k, not t1) and setting it for the next step so we don't use the original biasAccOmegaInit set at the very beginning.

gtsam/navigation/CombinedImuFactor.cpp
varunagrawal: I looked at this again. It's actually correct since it is calling `PreintegrationType::resetIntegration` aka the `resetIntegration` of the base class. I'll still make this change since it will make the code easier to read.

varunagrawal: Also `PreintegrationType::resetIntegration` resets the `PreintegratedMeasurements` and not the preintegration params which is where the covariances are stored.

gtsam/navigation/tests/testCombinedImuFactor.cpp
varunagrawal: Okay made a bunch of updates to enable this. I actually like the added flexibility.

varunagrawal:
@raabuchanan gentle reminder :)

gtsam/navigation/CombinedImuFactor.cpp
dellaert: Indeed, I think this might be incorrect. @varunagrawal , we should talk.

plusk01: Hi @varunagrawal, would it be possible to report back on the resolution of this to improve understanding? It might also be helpful to update the `ImuFactor.pdf` on the role of `\eps_init` in the model (eq 15, 16) --- is this just a zero-mean Gaussian prior on the bias when the IMU turned on? In which case, what is an example use case / motivation for wanting to set a new `biasAccOmegaInt` upon `resetIntegration`?

plusk01: also, it is probably worth renaming `biasAccOmegaInt` --> `biasAccOmegaInit` to be more consistent and to avoid potential confusion with the integration noise.

python/gtsam/examples/CombinedImuFactorExample.py
```diff
@@ -0,0 +1,251 @@
...
+GTSAM Copyright 2010-2019, Georgia Tech Research Corporation,
+Atlanta, Georgia 30332-0415
+All Rights Reserved
+
+See LICENSE for the license information
+
+A script validating and demonstrating the CobiendImuFactor inference.
```
plusk01, line 8: typo

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -184,6 +127,15 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurements : public PreintegrationType
   /// Re-initialize PreintegratedCombinedMeasurements
   void resetIntegration() override;
 
+  /**
+   * @brief Re-initialize PreintegratedCombinedMeasurements with initial bias
+   * covariance estimate.
+   *
+   * @param Q_init The initial bias covariance estimates as 6x6 matrix. If not
+   * provided, it uses the last values from the preintMeasCov.
```
plusk01, line 135: outdated?

python/gtsam/examples/CombinedImuFactorExample.py
```diff
@@ -0,0 +1,251 @@
...
+                plt.title("Ground Truth Trajectory")
+
+                # create IMU factor every second
+                factor = gtsam.CombinedImuFactor(X(i), V(i), X(i + 1),
+                                                 V(i + 1), B(i), B(i + 1), pim)
+                graph.push_back(factor)
+
+                if verbose:
+                    print(factor)
+                    print(pim.predict(initial_state_i, self.actualBias))
```
plusk01, line 195: should  there be a `initial_state_i = self.scenario.navState(i)` somewhere? currently `initial_state_i` is only for `i=0`

gtsam/navigation/CombinedImuFactor.cpp
varunagrawal: @plusk01 yes it is a gaussian for when we start the IMU. The idea for setting a new `biasAccOmegaInit` is that when we reset the integration between $t_k$ and $t_{k+1}$, we are still using the old `biasAccOmegaInit` rather than a new covariance for $t_{k+1}$. As @raabuchanan mentioned, we could compute this new value from the marginal on the bias.

Regarding the naming of `biasAccOmegaInt` to `biasAccOmegaInit`, I'll need @dellaert's approval for that.

varunagrawal:
I've updated the PR to fix all comments and issues. @raabuchanan please review and let's land this? :)

gtsam/navigation/CombinedImuFactor.cpp
raabuchanan: @plusk01 `biasAccOmegaInt` is a typo and should be `biasAccOmegaInit`. It is the same variable as Q_init in `ImuFactor.pdf` except you pass a single value to the factor and it is converted to a diagonal matrix.

raabuchanan:
Looks good to me now

raabuchanan (approved):

varunagrawal:
Merging this in. Some CI elements are failing due to memory issues on the servers.

PR #1309, opened 2022-10-20
Hybrid Pruning

varunagrawal:
Got hybrid pruning to work and this PR is the result of it!

### DIFF
--- a/gtsam/hybrid/HybridSmoother.cpp
+++ b/gtsam/hybrid/HybridSmoother.cpp
@@ -47,8 +47,6 @@ void HybridSmoother::update(HybridGaussianFactorGraph graph,
 
   // Add the partial bayes net to the posterior bayes net.
   hybridBayesNet_.push_back<HybridBayesNet>(*bayesNetFragment);
-
-  tictoc_print_();
 }
 
 /* ************************************************************************* */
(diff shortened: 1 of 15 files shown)


varunagrawal:
@dellaert FYI this PR gets the 1D robot example working extremely well!

gtsam/hybrid/HybridSmoother.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    // Conditionals to remove from the bayes net
+    // since the conditional will be updated.
+    std::vector<HybridConditional::shared_ptr> conditionals_to_erase;
+
+    // New conditionals to add to the graph
+    gtsam::HybridBayesNet newConditionals;
+
+    // NOTE(Varun) Using a for-range loop doesn't work since some of the
+    // conditionals are invalid pointers
```
ProfFan, line 73: Interesting, I don't think this is possible in a Hybrid Bayes Net?

ProfFan (commented):
One question

gtsam/hybrid/HybridSmoother.cpp
varunagrawal: It is! I was able to get a simple example up and running with this.

varunagrawal: I have no other explanation why a for-range loop doesn't work, unfortunately...

ProfFan (approved):

PR #1311, opened 2022-10-22
Update Switching fixture to be 0 indexed

varunagrawal:
Based on a recent conversation between @dellaert and myself, this PR converts the Switching test fixture from a 1-indexed scheme (e.g. `X(1)`) to a 0-indexed scheme (`X(0)`) since it better matches the iMHS paper.

This will ensure we have consistency of notation across papers and documents we write.


dellaert (approved):
LGTM

PR #1312, opened 2022-10-23
Dsf-gtsfm-refactor

dellaert:
More minimal C++ port of John's DSF track generation code.

1. Does not have NamedSfmMeasurement
2. Makes SfmTrack2d a gtsam type, that SfmTrack inherits from
3. Puts code in cpp file, broken up in easier to grok functions


gtsam/sfm/SfmTrack.h
```diff
@@ -78,6 +76,46 @@ struct GTSAM_EXPORT SfmTrack {
   /// Get the SIFT feature index corresponding to the measurement at `idx`
   const SiftIndex& siftIndex(size_t idx) const { return siftIndices[idx]; }
 
+  /**
+   * @brief Validates the track by checking that no two measurements are from
```
johnwlambert, line 80: nit: "...are from the same camera."

gtsam/sfm/SfmTrack.h
```diff
@@ -35,28 +35,26 @@ typedef std::pair<size_t, Point2> SfmMeasurement;
...
-struct GTSAM_EXPORT SfmTrack {
-  Point3 p;       ///< 3D position of the point
-  float r, g, b;  ///< RGB color of the 3D point
-
+struct GTSAM_EXPORT SfmTrack2d {
   /// The 2D image projections (id,(u,v))
   std::vector<SfmMeasurement> measurements;
 
-  /// The feature descriptors
+  /// The feature descriptors (optional)
```
johnwlambert, line 46: just curious, what is a siftIndex? I see it marked as a descriptor, so i would expect something like a 128-d vector, but that doesnt match the type (pair of size_t's). So it seems like it is supposed to represent 2d detections? Or maybe I'm misunderstanding the data structure?

gtsam/sfm/DsfTrackGenerator.h
```diff
@@ -0,0 +1,91 @@
...
+#include <optional>
+#include <vector>
+
+namespace gtsam {
+
+namespace gtsfm {
+
+typedef Eigen::MatrixX2i CorrespondenceIndices;  // N x 2 array
+
+// struct Keypoints;
```
johnwlambert, line 34: nit: i think this line contains some commented out code we could remove

johnwlambert (approved):
Thank you very much Frank! Much cleaner. LGTM

gtsam/sfm/SfmTrack.h
```diff
@@ -78,6 +76,46 @@ struct GTSAM_EXPORT SfmTrack {
...
+  }
+};
+
+using SfmTrack2dVector = std::vector<SfmTrack2d>;
+
+/**
+ * @brief An SfmTrack stores SfM measurements grouped in a track
+ * @addtogroup sfm
+ */
+struct GTSAM_EXPORT SfmTrack : SfmTrack2d {
```
akshay-krishnan, line 101: Nice!

gtsam/sfm/sfm.i
```diff
@@ -4,10 +4,22 @@
 
 namespace gtsam {
 
-#include <gtsam/nonlinear/NonlinearFactorGraph.h>
-#include <gtsam/nonlinear/Values.h>
 #include <gtsam/sfm/SfmTrack.h>
-class SfmTrack {
+class SfmTrack2d
+{
```
akshay-krishnan, line 9: nit: is this formatted? shouldnt it be `class SfmTrack2d {` ?

gtsam/sfm/sfm.i
```diff
@@ -4,10 +4,22 @@
...
+  std::vector<pair<size_t, gtsam::Point2>> measurements;
+
+  SfmTrack2d();
+  SfmTrack2d(std::vector<gtsam::SfmMeasurement> &measurements);
```
akshay-krishnan, line 13: const std::vector ?

gtsam/sfm/sfm.i
```diff
@@ -4,10 +4,22 @@
...
+  size_t numberMeasurements() const;
+  pair<size_t, gtsam::Point2> measurement(size_t idx) const;
+  pair<size_t, size_t> siftIndex(size_t idx) const;
+  void addMeasurement(size_t idx, const gtsam::Point2& m);
+  gtsam::SfmMeasurement measurement(size_t idx) const;
+  bool hasUniqueCameras();
```
akshay-krishnan, line 19: nit: const method?

gtsam/sfm/sfm.i
```diff
@@ -310,4 +317,45 @@ class TranslationRecovery {
...
+
+  size_t size() const;
+  bool empty() const;
+  void clear();
+  gtsam::gtsfm::CorrespondenceIndices at(const pair<size_t, size_t>& keypair) const;
+};
+
+
+class Keypoints
+{
```
akshay-krishnan, line 336: some formatting inconsistency here too

gtsam/sfm/DsfTrackGenerator.h
```diff
@@ -0,0 +1,91 @@
...
+   * Creates a disjoint-set forest (DSF) and 2d tracks from pairwise matches.
+   * We create a singleton for union-find set elements from camera index of a
+   * detection and the index of that detection in that camera's keypoint list,
+   * i.e. (i,k).
+   *
+   * @param Map from (i1,i2) image pair indices to (K,2) matrix, for K
+   *        correspondence indices, from each image.
+   * @param Length-N list of keypoints, for N images/cameras.
+   */
+  static std::vector<SfmTrack2d> tracksFromPairwiseMatches(
```
akshay-krishnan, line 84: why do we need a class? 🤔

gtsam/sfm/DsfTrackGenerator.cpp
```diff
@@ -0,0 +1,136 @@
...
+      tracks2d.begin(), tracks2d.end(), std::back_inserter(validTracks),
+      [](const SfmTrack2d& track2d) { return track2d.hasUniqueCameras(); });
+
+  if (verbose) {
+    size_t erroneous_track_count = tracks2d.size() - validTracks.size();
+    double erroneous_percentage = static_cast<float>(erroneous_track_count) /
+                                  static_cast<float>(tracks2d.size()) * 100;
+
+    // TODO(johnwlambert): restrict decimal places to 2 decimals.
+    std::cout << "DSF Union-Find: " << erroneous_percentage;
```
akshay-krishnan, line 125: that TODO should be easy to address now

gtsam/sfm/DsfTrackGenerator.h
dellaert: We do not - fixed

gtsam/sfm/SfmTrack.h
dellaert: Don't remember and not on my A->B at this time.

akshay-krishnan (approved):

gtsam/sfm/DsfTrackGenerator.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <iostream>
+
+namespace gtsam {
+
+namespace gtsfm {
+
+typedef DSFMap<IndexPair> DSFMapIndexPair;
+
+/// Generate the DSF to form tracks.
+static DSFMapIndexPair generateDSF(const MatchIndicesMap& matches) {
```
akshay-krishnan, line 31: is there a reason for these functions to be static?

dellaert:
@johnwlambert added a vectorized interface for extra speed. If the DSF is still a bottleneck, we could also try a DSFBase with `10000000*i1 + k1` as key. DSFBase uses vector rather than map and can be significantly faster.

gtsam/sfm/DsfTrackGenerator.cpp
dellaert: Yes. They are internal to a cpp file; it is customary to make them static.

dellaert:
PS, those two vectorized methods in SfmTrack2d were almost completely written by Github CoPilot. Pretty amazing :-)

varunagrawal:
Was this tested with the Matlab wrapper? :slightly_frowning_face: I seem to be having compilation issues here...

akshay-krishnan:
> Was this tested with the Matlab wrapper? 🙁 I seem to be having compilation issues here...

The CI does not run matlab wrapper tests?

varunagrawal:
Nope. Matlab needs a license so we can't install it on the CI.

gtsam/sfm/sfm.i
```diff
@@ -310,4 +318,38 @@ class TranslationRecovery {
...
+#include <gtsam/sfm/DsfTrackGenerator.h>
+
+class MatchIndicesMap {
+  MatchIndicesMap();
+  MatchIndicesMap(const gtsam::gtsfm::MatchIndicesMap& other);
+
+  size_t size() const;
+  bool empty() const;
+  void clear();
+  gtsam::gtsfm::CorrespondenceIndices at(const pair<size_t, size_t>& keypair) const;
```
varunagrawal, line 332: This line breaks the Matlab wrapper because `std::map` doesn't have an `at` method that takes a `std::pair` as its argument, rather it takes the key type.

[Docs](https://en.cppreference.com/w/cpp/container/map/at)

@dellaert there is no unit test for this.
@johnwlambert @akshay-krishnan reviewers need to do a better job. :slightly_smiling_face:

akshay-krishnan: My wrapper knowledge is very limited, sorry for missing that!

varunagrawal: No worries man. Just something to keep in mind going forward. :slightly_smiling_face:

varunagrawal:
@johnwlambert @akshay-krishnan can I expect a fix from one of you two soon? @mattking-smith and I need the Matlab wrapper for a paper he's working on and needs to submit soon.

gtsam/sfm/sfm.i
akshay-krishnan: what would the fix be? just changing this to `const IndexPair& keypair`? 
if we do that, can we do `.at((i, j))` in. python? or will we need to construct `IndexPair(i, j)` ?

akshay-krishnan:
> @johnwlambert @akshay-krishnan can I expect a fix from one of you two soon? @mattking-smith and I need the Matlab wrapper for a paper he's working on and needs to submit soon.

how urgent is it? I would be able to send a fix tonight, and I doubt @johnwlambert will be able to get to this sooner.

johnwlambert:
Unfortunately I don’t own a Matlab license so I won’t be able to test it out. Could we set up an octave CI? Would the octave install instructions match the matlab ones 1:1? Alternatively, a matlab license key could be stored in the gtsam GitHub Actions secrets file to install it during a CI run.

varunagrawal:
> > @johnwlambert @akshay-krishnan can I expect a fix from one of you two soon? @mattking-smith and I need the Matlab wrapper for a paper he's working on and needs to submit soon.
> 
> how urgent is it? I would be able to send a fix tonight, and I doubt @johnwlambert will be able to get to this sooner.

I would say it is pretty urgent. If you make the PR, I can review it and test it out on my machine which has the Matlab wrappers built.

@johnwlambert Octave won't work since the bindings don't match up exactly. Uploading a Matlab license might be a legal hassle. I don't have the cycles for this currently, so if you want to take this up, please feel free.

PR #1314, opened 2022-10-25
Fix CMake so that Eigen discovery and linking is consistent

varunagrawal:
This fixes the issue that was encountered after merging #1304 since that PR unearthed this subtle bug. Kudos to @OznOg for helping with this.

The issue is that Eigen is being `find_package`d twice, and if the first one is the system Eigen, it keeps using that even if the USE_SYSTEM_EIGEN flag is set to OFF.

As a result, the version set by CMake is the 3rdparty one, but GTSAM is compiled against the system Eigen, hence version mismatch.

Easy fix is to just not do the first `find_package` since that is being used to set a default flag value (which is now OFF directly). KISS

### DIFF
--- a/cmake/HandleEigen.cmake
+++ b/cmake/HandleEigen.cmake
@@ -1,10 +1,6 @@
 ###############################################################################
 # Option for using system Eigen or GTSAM-bundled Eigen
-# Default: Use system's Eigen if found automatically:
-find_package(Eigen3 CONFIG QUIET)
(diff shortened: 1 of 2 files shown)


dellaert (approved):
I'll approve and merge because GTSAM is broken after $1304, but @jlblancoc please take a second look to see if you agree with the changes? We can always do another PR if not.

PR #1316, opened 2022-10-26
Fix for matlab wrapper in sfm.i

akshay-krishnan:
Changes from pair<size_t, size_t> to IndexPair

### DIFF
--- a/gtsam/sfm/sfm.i
+++ b/gtsam/sfm/sfm.i
@@ -329,7 +329,7 @@ class MatchIndicesMap {
   size_t size() const;
   bool empty() const;
   void clear();
-  gtsam::gtsfm::CorrespondenceIndices at(const IndexPair& keypair) const;
+  gtsam::gtsfm::CorrespondenceIndices at(const gtsam::IndexPair& keypair) const;
 };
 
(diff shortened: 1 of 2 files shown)


akshay-krishnan:
@varunagrawal would this fix the matlab wrapper? I don't have matlab myself, so wont be able to test it.

johnwlambert:
Thanks Akshay. Looks fine to me, but I'll defer to the CI and to anyone with a MATLAB license...

I don't understand why this change is necessary, though : - ) [IndexPair](https://github.com/borglab/gtsam/blob/develop/gtsam/base/DSFMap.h#L117) should compile to a pair of size_t's, looking at its definition:
```c++
class IndexPair : public std::pair<size_t,size_t> {
```

varunagrawal:
@akshay-krishnan compiles great.

@johnwlambert this seems to be something going on at the compiler level and my guess is it has to do with template matching at compile time. You're specifying the template as `IndexPair` but using `std::pair<size_t, size_t>` so perhaps the match is incorrect?

akshay-krishnan:
Thanks! We can merge once CI completes.

PR #1317, opened 2022-10-28
Wrapper Update

varunagrawal:
Update the wrapper to support Matlab properties and add a GTSAM test.


mattking-smith (approved):
Tested on my machine and looks good.

PR #1318, opened 2022-10-31
Hybrid Mixture Error

varunagrawal:
Add `error()` methods to both the `MixtureFactor` and the `GaussianMixtureFactor`.

This will be used when computing the error terms to create the discrete probability tree.


varunagrawal:
Also added computing error for the `HybridBayesNet` and the `GaussianMixture` (aka the Hybrid Conditional). Now we can get the error tree for the full hybrid bayes net.

varunagrawal:
Can I please get some love on this PR? It's just adds to the API.

dellaert:
I will take a look now

gtsam/hybrid/GaussianMixture.h
```diff
@@ -144,6 +144,26 @@ class GTSAM_EXPORT GaussianMixture
...
+
+  /**
+   * @brief Compute the error of this Gaussian Mixture given the continuous
+   * values and a discrete assignment.
+   *
+   * @param continuousVals The continuous values at which to compute the error.
+   * @param discreteValues The discrete assignment for a specific mode sequence.
+   * @return double
+   */
+  double error(const VectorValues &continuousVals,
```
dellaert, line 164: `continuousValues`

gtsam/hybrid/GaussianMixtureFactor.h
```diff
@@ -126,6 +130,26 @@ class GTSAM_EXPORT GaussianMixtureFactor : public HybridFactor {
    */
   Sum add(const Sum &sum) const;
 
+  /**
+   * @brief Compute error of the GaussianMixtureFactor as a tree.
+   *
+   * @param continuousVals The continuous VectorValues.
+   * @return AlgebraicDecisionTree<Key> A decision tree with corresponding keys
+   * as the factor but leaf values as the error.
```
dellaert, line 138: as the factor??

gtsam/hybrid/GaussianMixtureFactor.h
```diff
@@ -126,6 +130,26 @@ class GTSAM_EXPORT GaussianMixtureFactor : public HybridFactor {
...
+
+  /**
+   * @brief Compute the error of this Gaussian Mixture given the continuous
+   * values and a discrete assignment.
+   *
+   * @param continuousVals The continuous values at which to compute the error.
+   * @param discreteValues The discrete assignment for a specific mode sequence.
+   * @return double
+   */
+  double error(const VectorValues &continuousVals,
```
dellaert, line 150: `continuousValues` (also in cpp, of course)

gtsam/hybrid/HybridBayesNet.h
```diff
@@ -124,6 +124,37 @@ class GTSAM_EXPORT HybridBayesNet : public BayesNet<HybridConditional> {
...
+   * @brief Compute conditional error for each discrete assignment,
+   * and return as a tree.
+   *
+   * @param continuousValues Continuous values at which to compute the error.
+   * @return AlgebraicDecisionTree<Key>
+   */
+  AlgebraicDecisionTree<Key> error(const VectorValues &continuousValues) const;
+
+  /**
+   * @brief Compute unnormalized probability for each discrete assignment,
```
dellaert, line 148: Add formulas as in gtsam, so we can clearly see difference in the docs?

gtsam/hybrid/HybridBayesNet.h
```diff
@@ -124,6 +124,37 @@ class GTSAM_EXPORT HybridBayesNet : public BayesNet<HybridConditional> {
...
+   * for a specific discrete assignment.
+   *
+   * @param continuousValues Continuous values at which to compute the error.
+   * @param discreteValues Discrete assignment for a specific mode sequence.
+   * @return double
+   */
+  double error(const VectorValues &continuousValues,
+               const DiscreteValues &discreteValues) const;
+
+  /**
```
dellaert, line 138: Could you (as in the other files, move the tree version *before* the non-tree version? In gtsam, the convention is to call "up", and I prefer the API to be structured that way as well.

gtsam/hybrid/HybridGaussianFactorGraph.h
```diff
@@ -170,6 +170,26 @@ class GTSAM_EXPORT HybridGaussianFactorGraph
     }
   }
 
+  /**
+   * @brief Compute error for each discrete assignment,
```
dellaert, line 174: formulas for this and next function?

gtsam/hybrid/MixtureFactor.h
```diff
@@ -107,8 +107,12 @@ class MixtureFactor : public HybridFactor {
       std::copy(f->keys().begin(), f->keys().end(),
                 std::inserter(factor_keys_set, factor_keys_set.end()));
 
-      nonlinear_factors.push_back(
-          boost::dynamic_pointer_cast<NonlinearFactor>(f));
+      if (auto nf = boost::dynamic_pointer_cast<NonlinearFactor>(f)) {
```
dellaert, line 110: Why is this templated? Can't we just have boost::shared_ptr<NonlinearFactor> anyway? which, BTW, is known as a `NonlinearFactorGraph` ? Then the implementation can also be moved to cpp.

PS, hoping the above is a rhetorical question but maybe there is a reason I don't understand.

gtsam/hybrid/MixtureFactor.h
```diff
@@ -149,7 +169,7 @@ class MixtureFactor : public HybridFactor {
 
   /// print to stdout
   void print(
-      const std::string& s = "MixtureFactor",
+      const std::string& s = "",
```
dellaert, line 172: Why?

gtsam/hybrid/tests/testGaussianMixture.cpp
```diff
@@ -78,15 +78,51 @@ TEST(GaussianMixture, Equals) {
...
+      vector<GaussianConditional::shared_ptr>{conditional0, conditional1});
+  GaussianMixture mixture({X(1)}, {X(2)}, {m1}, conditionals);
+
+  VectorValues values;
+  values.insert(X(1), Vector2::Ones());
+  values.insert(X(2), Vector2::Zero());
+  auto error_tree = mixture.error(values);
+
+  std::vector<DiscreteKey> discrete_keys = {m1};
+  std::vector<double> leaves = {0.5, 4.3252595};
```
dellaert, line 119: Always comment on numerical values in tests, either "regression" (you copied from output and it's not a real test), or "calculated by hand" or something else.

gtsam/hybrid/tests/testGaussianMixture.cpp
```diff
@@ -78,15 +78,51 @@ TEST(GaussianMixture, Equals) {
...
+  AlgebraicDecisionTree<Key> expected_error(discrete_keys, leaves);
+
+  // regression
+  EXPECT(assert_equal(expected_error, error_tree, 1e-6));
+}
```
dellaert, line 124: What about the non-tree version?

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -151,6 +151,45 @@ TEST_UNSAFE(GaussianMixtureFactor, GaussianMixture) {
...
+
+  VectorValues continuousVals;
+  continuousVals.insert(X(1), Vector2(0, 0));
+  continuousVals.insert(X(2), Vector2(1, 1));
+
+  // error should return a tree of errors, with nodes for each discrete value.
+  AlgebraicDecisionTree<Key> error_tree = mixtureFactor.error(continuousVals);
+
+  std::vector<DiscreteKey> discrete_keys = {m1};
+  std::vector<double> errors = {1, 4};
```
dellaert, line 181: regression? comment...

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -183,6 +183,61 @@ TEST(HybridBayesNet, OptimizeMultifrontal) {
...
+
+  std::vector<double> pruned_leaves = {2e50, 3.3973404e-31, 2e50, 0.0097568009};
+  AlgebraicDecisionTree<Key> expected_pruned_error(discrete_keys,
+                                                   pruned_leaves);
+
+  // regression
+  EXPECT(assert_equal(expected_pruned_error, pruned_error_tree, 1e-9));
+
+  // Verify error computation and check for specific error value
+  DiscreteValues discrete_values;
```
dellaert, line 218: Check tests in discrete for more elegant ways to construct discrete values

gtsam/hybrid/tests/testMixtureFactor.cpp
```diff
@@ -0,0 +1,108 @@
...
+// Check iterators of empty mixture.
+TEST(MixtureFactor, Constructor) {
+  MixtureFactor factor;
+  MixtureFactor::const_iterator const_it = factor.begin();
+  CHECK(const_it == factor.end());
+  MixtureFactor::iterator it = factor.begin();
+  CHECK(it == factor.end());
+}
+
+
```
dellaert, line 44: separator + comment

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -207,4 +207,28 @@ void GaussianMixture::prune(const DecisionTreeFactor &decisionTree) {
   conditionals_.root_ = pruned_conditionals.root_;
 }
 
+/* *******************************************************************************/
+AlgebraicDecisionTree<Key> GaussianMixture::error(
+    const VectorValues &continuousVals) const {
+  // functor to convert from GaussianConditional to double error value.
```
dellaert, line 213: convert? -? calculate

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -207,4 +207,28 @@ void GaussianMixture::prune(const DecisionTreeFactor &decisionTree) {
...
+  auto errorFunc =
+      [continuousVals](const GaussianConditional::shared_ptr &conditional) {
+        if (conditional) {
+          return conditional->error(continuousVals);
+        } else {
+          // return arbitrarily large error
```
dellaert, line 219: Comment on when this would happen?

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -207,4 +207,28 @@ void GaussianMixture::prune(const DecisionTreeFactor &decisionTree) {
...
+          return 1e50;
+        }
+      };
+  DecisionTree<Key, double> errorTree(conditionals_, errorFunc);
+  return errorTree;
+}
+
+/* *******************************************************************************/
+double GaussianMixture::error(const VectorValues &continuousVals,
```
dellaert, line 228: Why does above not call this one? And, then my comment about calling "up" would have to be reversed: this is the one that should be declared and defined first...

gtsam/hybrid/GaussianMixtureFactor.cpp
```diff
@@ -95,4 +95,24 @@ GaussianMixtureFactor::Sum GaussianMixtureFactor::asGaussianFactorGraphTree()
...
+  // functor to convert from sharedFactor to double error value.
+  auto errorFunc = [continuousVals](const GaussianFactor::shared_ptr &factor) {
+    return factor->error(continuousVals);
+  };
+  DecisionTree<Key, double> errorTree(factors_, errorFunc);
+  return errorTree;
+}
+
+/* *******************************************************************************/
+double GaussianMixtureFactor::error(
```
dellaert, line 111: same comment

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -232,4 +232,51 @@ VectorValues HybridBayesNet::optimize(const DiscreteValues &assignment) const {
...
+    } else if (factors_.at(idx)->isDiscrete()) {
+      // If factor at `idx` is discrete-only, we skip.
+      continue;
+    }
+  }
+
+  return error_tree;
+}
+
+AlgebraicDecisionTree<Key> HybridBayesNet::probPrime(
```
dellaert, line 276: separator

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -232,4 +232,51 @@ VectorValues HybridBayesNet::optimize(const DiscreteValues &assignment) const {
...
 
+/* ************************************************************************* */
+double HybridBayesNet::error(const VectorValues &continuousValues,
+                             const DiscreteValues &discreteValues) const {
+  GaussianBayesNet gbn = this->choose(discreteValues);
+  return gbn.error(continuousValues);
+}
+
+/* ************************************************************************* */
+AlgebraicDecisionTree<Key> HybridBayesNet::error(
```
dellaert, line 243: Add comments below, I can't figure out what's going on without reading the code.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -423,4 +423,53 @@ const Ordering HybridGaussianFactorGraph::getHybridOrdering() const {
   return ordering;
 }
 
+/* ************************************************************************ */
+AlgebraicDecisionTree<Key> HybridGaussianFactorGraph::error(
```
dellaert, line 427: same: add comments below and I will review. As is can;t do it.

dellaert (changes requested):
Some "tough love" :-) :-)

dellaert:
PS also merge in master before pushing a new version? And ping me on email or I might forget to look at it.

gtsam/hybrid/GaussianMixture.h
```diff
@@ -144,6 +144,26 @@ class GTSAM_EXPORT GaussianMixture
   /// Getter for the underlying Conditionals DecisionTree
   const Conditionals &conditionals();
 
+  /**
+   * @brief Compute error of the GaussianMixture as a tree.
+   *
+   * @param continuousVals The continuous VectorValues.
+   * @return AlgebraicDecisionTree<Key> A decision tree with corresponding keys
+   * as the factor but leaf values as the error.
```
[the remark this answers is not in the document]
varunagrawal: It should read "keys corresponding to the factors involved", or "__the same keys__ as the factor".

Since this is a GaussianMixture, I should be using "conditionals" and not "factors".

gtsam/hybrid/HybridBayesNet.h
varunagrawal: Semantically, the non-tree version should call the tree version. I don't do that in practice since it is cheaper to index via the discrete values and then compute the error, rather than create the whole tree and index into the tree.

gtsam/hybrid/MixtureFactor.h
varunagrawal: The reason is that boost shared_pointers don't seem to allow type hierarchy. For example, if I remove the template, this works

```cpp
using MotionModel = BetweenFactor<double>;

auto noise_model = noiseModel::Isotropic::Sigma(1, 1.0);
auto still = boost::make_shared<MotionModel>(X(0), X(1), 0.0, noise_model),
     moving = boost::make_shared<MotionModel>(X(0), X(1), 1.0, noise_model);

std::vector<NonlinearFactor::shared_ptr> components = {still, moving};
auto dcFactor = boost::make_shared<MixtureFactor>(
    contKeys, DiscreteKeys{gtsam::DiscreteKey(M(1), 2)}, components);
```

but this doesn't
```cpp
using MotionModel = BetweenFactor<double>;

auto noise_model = noiseModel::Isotropic::Sigma(1, 1.0);
auto still = boost::make_shared<MotionModel>(X(0), X(1), 0.0, noise_model),
     moving = boost::make_shared<MotionModel>(X(0), X(1), 1.0, noise_model);

std::vector<MotionModel::shared_ptr> components = {still, moving};
auto dcFactor = boost::make_shared<MixtureFactor>(
    contKeys, DiscreteKeys{gtsam::DiscreteKey(M(1), 2)}, components);
```

The only difference is the type of `components`. 😕

varunagrawal: I guess this becomes a matter of API convenience. Being able to pass in `vector<SpecificType>` rather than `vector<NonlinearFactor>` seems cleaner to me. Same reason for not using `NonlinearFactorGraph` even though they are semantically the same. I can update this if you'd still like.

varunagrawal: Also, I can't do this little neat trick: `NonlinearFactorGraph components = {still, moving};`. Maybe this indicates something missing from the `FactorGraph` class.

gtsam/hybrid/MixtureFactor.h
varunagrawal: It anyway prints `MixtureFactor` below on line 176.

gtsam/hybrid/tests/testGaussianMixture.cpp
varunagrawal: Ah, I added "regression" to the `EXPECT` call right below. Let me move it up.

gtsam/hybrid/GaussianMixture.cpp
varunagrawal: I thought about this. Indexing (`O(1)`) and computing the error is cheaper than building the tree (`O(M)`).

gtsam/hybrid/GaussianMixtureFactor.cpp
varunagrawal: Cheaper to directly index than build the whole tree first.

varunagrawal: I've added a comment to both methods specifying this.

dellaert (approved):
LGTM, thanks for addressing comments! Merge at will :-)

PR #1319, opened 2022-11-01
Proper Hybrid Elimination

varunagrawal:
This PR adds a unit test showing that the elimination scheme isn't correct since the probabilities of the continuous variables `P(X | M, Z)` doesn't match the case when we individually construct each particular modality.

Basically, I have added a unit test that calls the method `getSpecificProblem` which returns a specific (nonlinear) factor graph corresponding to a specific discrete mode sequence. The idea is that the error of this factor graph should equal the error of the hybrid factor graph for the same discrete value assignment.


keevindoherty:
Hi @varunagrawal - thanks for keeping me posted in these PRs! I had some questions related to #1318 and this PR.

I may be misunderstanding, but when I read this, I was confused about a few things:
- I wasn't sure the relationship between the conditional p(X | M, Z) and the elimination scheme in this context. I was thinking if we eliminate discrete variables 'M' from the joint we should have p(X | Z). Are we eliminating a subset of X from the conditional? Or when you say 'elimination' here, are you referring specifically to a chordal Bayes net specifying p(X | M, Z) induced by a particular elimination order, rather than the resulting marginal?
- Accepting for the moment that we are dealing with the conditional p(X | M, Z), it wasn't obvious to me that (the negative log of) this conditional p(X | M, Z) should correspond to the value of the hybrid factor graph error for the same discrete assignment (this certainly need not be the case in general). Specifically, we know that p(X | M = M, Z = Z) satisfies int_X p(X | M = M, Z = Z) = 1, whereas the factor graph for a specific choice of M encodes p(X, M = M, Z = Z). All we know is that this is _proportional to_ p(X | M = M, Z = Z), but it is not necessarily normalized w.r.t. X. (I'm abusing notation a little bit here to make explicit the dependence on here on _fixed_ assignments to M and Z, hopefully that makes sense?)

Let me know if I am missing something?

varunagrawal:
> I wasn't sure the relationship between the conditional p(X | M, Z) and the elimination scheme in this context. I was thinking if we eliminate discrete variables 'M' from the joint we should have p(X | Z). Are we eliminating a subset of X from the conditional? Or when you say 'elimination' here, are you referring specifically to a chordal Bayes net specifying p(X | M, Z) induced by a particular elimination order, rather than the resulting marginal?

Hey Kevin. Yes we eliminated as per a specific ordering (continuous followed by discrete) so we will always eliminate subsets of X before eliminating any M.

> Accepting for the moment that we are dealing with the conditional p(X | M, Z), it wasn't obvious to me that (the negative log of) this conditional p(X | M, Z) should correspond to the value of the hybrid factor graph error for the same discrete assignment (this certainly need not be the case in general). Specifically, we know that p(X | M = M, Z = Z) satisfies int_X p(X | M = M, Z = Z) = 1, whereas the factor graph for a specific choice of M encodes p(X, M = M, Z = Z). All we know is that this is proportional to p(X | M = M, Z = Z), but it is not necessarily normalized w.r.t. X. (I'm abusing notation a little bit here to make explicit the dependence on here on fixed assignments to M and Z, hopefully that makes sense?)

Yes you are correct. I am using P(X | M, Z) to mean the unnormalized property, and the way I am considering it is akin to a Gaussian Mixture Model where M is our indexing variable. The idea is then that given a particular Gaussian index, we get the Gaussian P(X | Z). In fact, that is exactly what my unit test in this PR is doing, generating each Gaussian separately and verifying that elimination gives us the equivalent numbers. If I am misunderstanding your question, please let me know.

ProfFan:
I think what Kevin refers to here is that since our hybrid factors just chooses one of a bunch continuous factors, they do not necessarily have the same normalization constant thus you will have a different scale for the branches, which will make your estimation biased. Please correct me if I am speaking nonsense, @keevindoherty  :)

varunagrawal:
I agree that each leaf will have a different normalization constant, but that shouldn't make any of the leaves biased.

The way to think of this is maintaining N different chains, one corresponding to each leaf. If we perform MLE on each of the chains, we can pick the chain that is best supported by the measurement data. This is a similar idea to particle filtering, where each particle represents a particular discrete mode sequence.

varunagrawal:
@ProfFan @keevindoherty @dellaert I managed to figure out the correct way to perform elimination and set the discrete probabilities such that the probability `P'(Continuous | Discrete)` is the same if we have M different (continuous-only) factor graphs, where M is the total number of discrete modes.

To do this, I had to create a new method `eliminateHybridSequential` and the resulting tests are in `testHybridEstimation.cpp`.

The bad news is that we will have to overload the the `eliminateSequential` methods in the `HybridGaussianFactorGraph` class, since otherwise some of the prior tests fail.

varunagrawal:
Unit tests for this PR will pass once we merge #1323 into this branch. 🙂

dellaert (approved):
I think once #1356 is merged and you merge in develop, taking care of conflicts, we can merge this once CI passes. Don't merge in the "Model Selection" things in yet.

PR #1323, opened 2022-11-10
Hybrid Multifrontal

varunagrawal:
Perform correct multifrontal elimination for hybrid factor graphs.

This follows the same general scheme as #1319 but updates the `HybridBayesTree` to take care of the needed operations. I overloaded the `eliminateMultifrontal` methods so that the elimination would be done correctly and without the need to change any major code in the unit tests.


gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -640,25 +719,99 @@ boost::shared_ptr<HybridGaussianFactorGraph::BayesNetType>
...
+
+  discreteGraph->add(DecisionTreeFactor(discrete_keys, probPrimeTree));
+
+  auto updatedBayesTree =
+      discreteGraph->BaseEliminateable::eliminateMultifrontal(discrete_ordering,
+                                                              function);
+
+  auto discrete_clique = (*updatedBayesTree)[discrete_ordering.at(0)];
+
+  std::set<HybridBayesTreeClique::shared_ptr> clique_set;
```
varunagrawal, line 771: I particularly want a review on this part. Am I creating the Bayes Tree with the discrete root correctly? Unit tests seem to say yes, but if there is a better way, I am all ears.

gtsam/hybrid/tests/testHybridGaussianISAM.cpp
```diff
@@ -177,10 +178,10 @@ TEST(HybridGaussianElimination, IncrementalInference) {
 
   // Test if the probability values are as expected with regression tests.
```
dellaert, line 179: "as expected" and "regression test" are mutually incompatible. If this is a unit test, say how you computed 0.1666667. If it is a regression tests, we'll assume you copied from output to safeguard against future regressions.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: without documentation (e.g. with math, I don;t even know what is happeingin). In general true for this entire piece of code. Now that you have the math, migth as well add it?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -640,25 +607,99 @@ boost::shared_ptr<HybridGaussianFactorGraph::BayesNetType>
...
+  for (auto node : bayesTree->nodes()) {
+    clique_set.insert(node.second);
+  }
+
+  // Set the root of the bayes tree as the discrete clique
```
dellaert, line 664: more mutating? Very suspicious.

gtsam/hybrid/HybridGaussianFactorGraph.h
```diff
@@ -220,34 +220,89 @@ class GTSAM_EXPORT HybridGaussianFactorGraph
...
    *
+   * @tparam BAYES  Template on the type of Bayes graph, either a bayes net or a
+   * bayes tree.
    * @param discrete_keys The discrete keys which form all the modes.
    * @param continuousBayesNet The Bayes Net representing the continuous
    * eliminated variables.
    * @return AlgebraicDecisionTree<Key>
    */
+  template <typename BAYES>
   AlgebraicDecisionTree<Key> continuousProbPrimes(
```
dellaert, line 261: This is not a great name. It refers to math that is not in the documentation. Suggest `unnormalizedProbabilities` documenting the fact that it is `exp(-E)`. On a side note: I question the wisdom of exponentiating small numbers before they will somehow be normalized. Typically we avoid doing this until we know the minimum error and then subtract the minimum error so the corresponding maximum prob' == 1.0.

gtsam/hybrid/HybridGaussianFactorGraph.h
```diff
@@ -259,16 +314,44 @@ class GTSAM_EXPORT HybridGaussianFactorGraph
...
       OptionalVariableIndex variableIndex = boost::none) const;
 
+  /// Sequential elimination overload for hybrid
   boost::shared_ptr<BayesNetType> eliminateSequential(
       const Ordering& ordering,
       const Eliminate& function = EliminationTraitsType::DefaultEliminate,
       OptionalVariableIndex variableIndex = boost::none) const;
 
+  /**
+   * @brief Custom elimination function which computes the correct
```
dellaert, line 330: OK, this is my biggest beef. More in the final review comment.

And add override to all functions you are overriding.

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -319,6 +320,85 @@ TEST(HybridEstimation, Probability) {
...
+  // This is the correct sequence
+  // std::vector<size_t> discrete_seq = {1, 1, 0};
+
+  double between_sigma = 1.0, measurement_sigma = 0.1;
+
+  std::vector<double> expected_errors, expected_prob_primes;
+  for (size_t i = 0; i < pow(2, K - 1); i++) {
+    std::vector<size_t> discrete_seq = getDiscreteSequence<K>(i);
+
+    GaussianFactorGraph::shared_ptr linear_graph = specificProblem(
```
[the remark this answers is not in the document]
varunagrawal: Renamed to `specificModesFactorGraph`. This is a separate function with its own docstring (which is hopefully okay).

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -640,25 +607,99 @@ boost::shared_ptr<HybridGaussianFactorGraph::BayesNetType>
...
+  // Eliminate continuous
+  HybridBayesTree::shared_ptr bayesTree;
+  HybridGaussianFactorGraph::shared_ptr discreteGraph;
+  std::tie(bayesTree, discreteGraph) =
+      BaseEliminateable::eliminatePartialMultifrontal(continuous_ordering,
+                                                      function, variableIndex);
+
+  // Get the last continuous conditional which will have all the discrete
+  Key last_continuous_key =
+      continuous_ordering.at(continuous_ordering.size() - 1);
```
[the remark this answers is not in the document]
varunagrawal: Very cool!

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -640,25 +607,99 @@ boost::shared_ptr<HybridGaussianFactorGraph::BayesNetType>
...
+  auto last_conditional = (*bayesTree)[last_continuous_key]->conditional();
```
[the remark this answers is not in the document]
varunagrawal: It should be a `GaussianMixture` since it is a continuous variable conditioned on discrete variables.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -640,25 +607,99 @@ boost::shared_ptr<HybridGaussianFactorGraph::BayesNetType>
...
+
+  // If not discrete variables, return the eliminated bayes net.
+  if (discrete_keys.size() == 0) {
+    return bayesTree;
+  }
+
+  AlgebraicDecisionTree<Key> probPrimeTree =
+      this->continuousProbPrimes(discrete_keys, bayesTree);
+
+  discreteGraph->add(DecisionTreeFactor(discrete_keys, probPrimeTree));
```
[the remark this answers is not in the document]
varunagrawal: This is the addition of the model selection term `P(M|Z)`.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -640,25 +607,99 @@ boost::shared_ptr<HybridGaussianFactorGraph::BayesNetType>
...
+
+  auto updatedBayesTree =
```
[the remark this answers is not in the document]
varunagrawal: Can't be const since we add the discrete clique as the root so that the Bayes tree is valid.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: I wasn't sure how else to update the BayesTree so set the discrete clique to be the root. That is a requirement for the structure of the HybridBayesTree.

gtsam/hybrid/HybridGaussianFactorGraph.h
varunagrawal: The problem I am trying to tackle is computing the correct `q(\mu)` terms for each mode sequence. This requires having all the continuous conditionals as a BayesNet that we can use.

Unfortunately, the elimination method in `EliminateableFactorGraph` calls `DefaultEliminate` (a static method) with one key at a time. By the time we get to the last continuous variable to eliminate, we only have access to the one conditional and not all the conditionals needed to make the BayesNet. Since `DefaultEliminate` is a static method, we also can't do something like `this->conditionals()`.

Hence the need to override, so we can eliminate all the continuous variables first to get the continuous BayesNet and get the `P(X | M, Z)` posterior as well as the terms in the `P(M | Z)` posterior (e.g. the `q(\mu)` term and the determinant term). This way we can perform all these distinct operations together in sequence when a user calls `HybridFactorGraph::eliminateSequential` or `HybridFactorGraph::eliminateMultifrontal`.

If you have a better way of doing this that doesn't involve overriding, I am all ears. 🙂

gtsam/hybrid/HybridGaussianFactorGraph.h
```diff
@@ -220,34 +220,89 @@ class GTSAM_EXPORT HybridGaussianFactorGraph
    * @brief Compute the VectorValues solution for the continuous variables for
```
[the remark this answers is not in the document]
varunagrawal: Updated

gtsam/hybrid/HybridGaussianFactorGraph.h
varunagrawal: These are unnormalized probabilities only on the continuous variables and `probPrime` is the method name for computing the unnormalized probabilities, so it seemed to make sense. If you still think I should rename it to `unnormalizedProbabilities`, will gladly do so.

Regarding the exponentiation, the minimum error term will always be 0 due to pruning.

gtsam/hybrid/HybridGaussianFactorGraph.h
varunagrawal: Also, I can't mark these methods as override since the base class functions are not virtual. I can update that on `EliminateableFactorGraph` but wanted to let you know first.

gtsam/hybrid/tests/testHybridBayesTree.cpp
```diff
@@ -32,6 +32,25 @@ using noiseModel::Isotropic;
...
 
+/* ****************************************************************************/
+// Test multifrontal optimize
+TEST(HybridBayesTree, OptimizeMultifrontal) {
+  Switching s(4);
+
+  Ordering hybridOrdering = s.linearizedFactorGraph.getHybridOrdering();
+  HybridBayesTree::shared_ptr hybridBayesTree =
+      s.linearizedFactorGraph.eliminateMultifrontal(hybridOrdering);
+  HybridValues delta = hybridBayesTree->optimize();
```
dellaert, line 43: Why would I only get continuous values here?

dellaert (commented):
Assuming this is WIP?

gtsam/hybrid/tests/testHybridBayesTree.cpp
varunagrawal: You should be getting both. Probably should test for `delta.discrete()` as well.

gtsam/hybrid/HybridGaussianFactorGraph.h
varunagrawal: Update: this function will get deleted.

varunagrawal:
I'm merging this sow we get a holistic view in #1319.

PR #1324, opened 2022-11-13
Remove unnecessary functions in ImuFactorsExample.cpp

Taeyoung96:
@varunagrawal   
This is a simple PR to improve code readability.  

In `ImuFactorsExample.cpp`, when optimizing using isam2, the `isam2->update();` function does not affect the results.  
I think it's better to remove this function to avoid confusion.  

If this PR helps, I'd appreciate it if you merge. :smile:  
Thanks,

### DIFF
--- a/examples/ImuFactorsExample.cpp
+++ b/examples/ImuFactorsExample.cpp
@@ -267,7 +267,6 @@ int main(int argc, char* argv[]) {
 
       if (use_isam) {
         isam2->update(*graph, initial_values);
-        isam2->update();
         result = isam2->calculateEstimate();
 
         // reset the graph


varunagrawal:
@Taeyoung96 thanks for the PR! We always appreciate help from wonderful community members like yourself. 🙂 

So just a quick clarifying question: the second `isam->update` method runs another optimization step, so you can think of it as running gradient descent for 2 timesteps instead of 1. Does removing the 2nd update call significantly affect the final results?

I would love to see some plots of the before and after and ensure they both look the same (along with the numerical values) since we don't have unit tests for this thing. If they are indeed the same, I will gladly approve and merge this PR.

Taeyoung96:
@varunagrawal  Thanks for your quick reply!  

After checking the results, the update() function did not affect the overall results.  
The result of the `.csv` file has the same numerical values. :smiley:   
(I performed it about 5 times, and I was able to get the same result.)  

Here are my result `.csv` files below.  
[imuFactorExampleResults(withUpdate).csv](https://github.com/borglab/gtsam/files/10000228/imuFactorExampleResults.withUpdate.csv)  
[imuFactorExampleResults(withoutUpdate).csv](https://github.com/borglab/gtsam/files/10000231/imuFactorExampleResults.withoutUpdate.csv)  

And the result of the plotting is as follows.  

<p align="center"></p>

varunagrawal:
Amazing! Thanks @Taeyoung96

varunagrawal (approved):
LGTM

Issue #1325, opened 2022-11-13
fisheye camera distortion issue


huanglilong:
https://github.com/borglab/gtsam/blob/256c664aaaa27a3204411e461d4d93c0e168bd6b/gtsam/geometry/Cal3Fisheye.cpp#L57

@ghaggin according to https://arxiv.org/pdf/1807.08957.pdf and https://users.aalto.fi/~kannalj1/publications/tpami2006.pdf

`const double s = scaling * K.dot(T);` should be `const double s = scaling * K.dot(T) * t;`

because K.dot(T)  is not equals to d(t) in (27)
![image](https://user-images.githubusercontent.com/7278867/201515735-ac20669d-292f-43ea-99a6-4e5eb9e5967c.png)

cc: @dellaert @varunagrawal

huanglilong:
the issue is from this pull https://github.com/borglab/gtsam/pull/307 @ghaggin @dellaert 
![image](https://user-images.githubusercontent.com/7278867/201516409-b9e2dd0b-606c-4e14-a6bf-fcd0341d3bdc.png)

varunagrawal:
That's interesting. Looks like we don't have the correct unit test coverage for this.

huanglilong:
from kalibr: https://github.com/ethz-asl/kalibr/blob/master/aslam_cv/aslam_cameras/include/aslam/cameras/implementation/EquidistantDistortion.hpp#L23
<img width="586" alt="image" src="https://user-images.githubusercontent.com/7278867/201530182-879c3193-f2b9-4f27-8885-f4aae5eb0309.png">

huanglilong:
@varunagrawal @ghaggin @roderick-koehle @dellaert please review https://github.com/borglab/gtsam/pull/1326

PR #1326, opened 2022-11-14
fix fisheye model dirtortion theta_d and scaling

huanglilong:
detail in https://github.com/borglab/gtsam/issues/1325
@ghaggin @dellaert  @varunagrawal


gtsam/geometry/Cal3Fisheye.cpp
```diff
@@ -53,9 +41,10 @@ Point2 Cal3Fisheye::uncalibrate(const Point2& p, OptionalJacobian<2, 9> H1,
   Vector5 K, T;
   K << 1, k1_, k2_, k3_, k4_;
   T << 1, t2, t4, t6, t8;
-  const double scaling = Scaling(r);
-  const double s = scaling * K.dot(T);
-  const double xd = s * xi, yd = s * yi;
+  T *= t;
+  const double theta_d = K.dot(T);
+  const double scaling = (r > 1e-8) ? theta_d / r : 1.0;
```
huanglilong, line 46: ref: https://arxiv.org/pdf/1807.08957.pdf
scaling = d(theta) / r
image

gtsam/geometry/Cal3Fisheye.cpp
```diff
@@ -110,21 +103,19 @@ Point2 Cal3Fisheye::uncalibrate(const Point2& p, OptionalJacobian<2, 9> H1,
...
-  
-  // Provide initial guess for the Gauss-Newton search.
+
+  // Provide initial guess for the Newton search.
   // The angular coordinates given by (xd, yd) are mapped back to
   // the focal plane of the perspective undistorted projection pi.
-  // See Cal3Unified.calibrate() using the same pattern for the 
+  // See Cal3Unified.calibrate() using the same pattern for the
   // undistortion of omnidirectional fisheye projection.
-  const double scale = (theta > 0) ? tan(theta) / theta : 1.0;
```
huanglilong, line 126: this scale don't need!
because 
theta_d = sqrt(xd * xd + yd * yd) in (31)
theta = theta_d inverse in (32)
image

gtsam/geometry/Cal3Fisheye.cpp
huanglilong: scaling = d(theta) / r 
in (24)

dellaert:
Also, seems CI is failing...

huanglilong:
> 1. the
image
Hi,  @dellaert  for Scaling function actually not need, but we need taylor expansion, because
in origin `Scaling(r)` function, is return `atan(r) / r `means `atan(sqrt(x^2 + y^2) / z) / sqrt(x^2 + y^2)`, so `atan(r)` is the incoming ray angle `theta` ,  in (24), the scaling should be `theta_d / r` , they are not equal, and they are close only when theta is very small (no distortion in center (u0, v0)
Note: z = 1

huanglilong:
image

huanglilong:
previous diff has bug and i has fixed
image

berndpfrommer:
Please see my comments regarding #1325

huanglilong:
@berndpfrommer @dellaert  thanks a lots, **the origin code is correct!** my mistake for `scaling` and `theta_d / r`

thanks for @berndpfrommer  comments https://github.com/borglab/gtsam/issues/1325#issuecomment-1315958657

because we are using `atan2` or `atan` calculates `theta`, this means Fisheye model only support `FOV < 180`
image

PR #1330, opened 2022-11-17
Add an expression for normalize()

nkhedekar:
Exposed `normalize()` as an expression and added a unit test for it


ProfFan (approved):
Thank you!

PR #1332, opened 2022-11-21
Added max_angle parameter for Random() in Rot3

nkhedekar:
Added the ability to provide the maximum angle as a parameter for the generated random `Rot3` instead of necessarily having the max_angle as a value in `[-M_PI, M_PI]`. A default value for `max_angle` is provided and set to `M_PI` to ensure backward compatibility.

### DIFF
--- a/gtsam/geometry/Rot3.cpp
+++ b/gtsam/geometry/Rot3.cpp
@@ -37,9 +37,9 @@ void Rot3::print(const std::string& s) const {
 }
 
 /* ************************************************************************* */
-Rot3 Rot3::Random(std::mt19937& rng) {
+Rot3 Rot3::Random(std::mt19937& rng, double max_angle) {
   Unit3 axis = Unit3::Random(rng);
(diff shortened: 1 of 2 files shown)


hovavalon (approved):

varunagrawal:
I don't see a problem with the code, but I am not sure whether this makes sense mathematically. By setting a `max_angle` we're no longer ensuring full support on the distribution we're sampling for. Can you please explain an example use case where this change might be useful?

nkhedekar:
The specific use case that I had in mind for this was to randomly sample a small rotation that can be used to perturb a pose. Happy to do it some other way if its already available but this seemed to be the simplest way to have this functionality.

dellaert:
Hi, thanks a bunch for a PR and proposed feature. I think your API makes sense, but you probably want this to be a function in your own code. I propose you copy the two lines of code this takes. Your use case is interesting but very specific, and the Rot3::Random has a different intent: to give a totally random rotation (which I think it does).

PR #1346, opened 2022-12-23
Sampling test for Hybrid Posterior

varunagrawal:
This test checks for correctness of the hybrid posterior using a sampling based approach.

Specifically, given a simple factor graph as below
```
 M0\
X0 -*- X1 
```

we compute the posterior `P(x0, m0, x1| z0, z1)` given measurements `z0, z1`.

The test, at a high level performs the following steps:
1. Start with the factor graph and linearize `FG` .
2. Eliminate into a bayes net `BN`.
3. Sample from the bayes net.
4. Check `BN(x)/FG(x)` equals the same value for all samples.

Since we don't have a sample method for the hybrid bayes net, we perform a sampling scheme similar to a mixture model where,
1. We sample from a discrete distribution weighted by the mixing weights $\pi$.
2. Given the discrete sample value, pick the corresponding bayes net and factor graph.
3. Sample as described above.

## Sample Method Update

Since the linearized factor graph and the eliminated bayes net __do not__ record the noise model, sampling from the Bayes Net throws an error. To overcome this, I've added an optional parameter to the sample method `SharedDiagonal model`, and if the conditional to sample from doesn't have a noise model associated with it, then it will sample from the provided model. This model should be the same one that is provided to the original factor in the Nonlinear/Gaussian Factor Graph, making it a convenient addition.


dellaert:
Can I ask to create a separate PR (to develop) with sampling for hybrid BNs? Seeing that in isolation will be easier to review.

dellaert:
Also, convert to draft PR? Don’t think we should run CI as long as we don’t target develop

varunagrawal:
> Can I ask to create a separate PR (to develop) with sampling for hybrid BNs? Seeing that in isolation will be easier to review.

So a `HybridBayesNet::sample` method? Can do!

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,6 +432,69 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
   EXPECT(assert_equal(discrete_seq, hybrid_values.discrete()));
 }
 
+/****************************************************************************/
+/**
+ * Test for correctness via sampling.
+ *
+ * Given the conditional P(x0, m0, x1| z0, z1)
```
dellaert, line 439: Nit: not given, computed in step 2

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,6 +432,69 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
...
+ * with meaasurements z0, z1, we:
+ * 1. Start with the corresponding Factor Graph `FG`.
+ * 2. Eliminate the factor graph into a Bayes Net `BN`.
+ * 3. Sample from the Bayes Net.
+ * 4. Check that the ratio `BN(x)/FG(x) = constant` for all samples `x`.
+ */
+TEST(HybridEstimation, CorrectnessViaSampling) {
+  HybridNonlinearFactorGraph nfg;
+
+  auto noise_model = noiseModel::Diagonal::Sigmas(Vector1(1.0));
```
dellaert, line 449: I thought there was a 1D constructor Sigma

dellaert, line 449: Const auto everywhere - susses out non-const methods/arguments

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,6 +432,69 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
   EXPECT(assert_equal(discrete_seq, hybrid_values.discrete()));
 }
 
+/****************************************************************************/
+/**
+ * Test for correctness via sampling.
+ *
+ * Given the conditional P(x0, m0, x1| z0, z1)
+ * with meaasurements z0, z1, we:
```
dellaert, line 440: Spelling. (install vs code spell-checking) and maybe insert comments inline below

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,6 +432,69 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
...
+ */
+TEST(HybridEstimation, CorrectnessViaSampling) {
+  HybridNonlinearFactorGraph nfg;
+
+  auto noise_model = noiseModel::Diagonal::Sigmas(Vector1(1.0));
+  auto zero_motion =
+      boost::make_shared<BetweenFactor<double>>(X(0), X(1), 0, noise_model);
+  auto one_motion =
+      boost::make_shared<BetweenFactor<double>>(X(0), X(1), 1, noise_model);
+  std::vector<NonlinearFactor::shared_ptr> factors = {zero_motion, one_motion};
```
dellaert, line 454: In-line in emplace call with {}

gtsam/hybrid/tests/testHybridEstimation.cpp
dellaert, line 449: Insert comment that you are creating nonlinear graph first

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,6 +432,69 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
...
+
+  Values initial;
+  double z0 = 0.0, z1 = 1.0;
+  initial.insert<double>(X(0), z0);
+  initial.insert<double>(X(1), z1);
+
+  // 1. Create the factor graph from the nonlinear factor graph.
+  HybridGaussianFactorGraph::shared_ptr fg = nfg.linearize(initial);
+  // 2. Eliminate into BN
+  Ordering ordering = fg->getHybridOrdering();
```
dellaert, line 467: Const…

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,6 +432,69 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
...
+  HybridBayesNet::shared_ptr bn = fg->eliminateSequential(ordering);
+
+  // Set up sampling
+  std::random_device rd;
```
dellaert, line 471: Where is rd used ? Answer with code comment

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,6 +432,69 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
...
+  std::mt19937_64 gen(11);
```
dellaert, line 472: Gen -> rng

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,6 +432,69 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
...
+    DiscreteValues assignment = sample.discrete();
+    double log_ratio = bn->error(sample.continuous(), assignment) -
+                       fg->error(sample.continuous(), assignment);
+    double ratio = exp(-log_ratio);
+
+    // Store the ratio for post-processing
+    ratios.push_back(ratio);
+  }
+
+  // 4. Check that all samples == 1.0 (constant)
```
dellaert, line 492: This is too clever for me. Explain in comment

dellaert (approved):
Many comments…

gtsam/hybrid/tests/testHybridEstimation.cpp
varunagrawal: Nope, that's only for Isotropic. Which I could use here as well, so I'll update this.

gtsam/hybrid/tests/testHybridEstimation.cpp
varunagrawal: Oh good catch. I've already added comments inline with the corresponding bullet numbers.

gtsam/hybrid/tests/testHybridEstimation.cpp
varunagrawal: This is when I wish we were using ML or Lisp. Const by default.

gtsam/hybrid/tests/testHybridEstimation.cpp
varunagrawal: Copy-pasta

gtsam/hybrid/tests/testHybridEstimation.cpp
dellaert: Why not just do the EXPECT_DOUBLES_EQUAL in the loop above? Simpler than to check the sum - which also might accidentally sum to 1. (very unlikely I know). You then also don't have to store the ratios.

dellaert (approved):

gtsam/hybrid/tests/testHybridEstimation.cpp
dellaert: Actually, it is also wrong: ratios should be the same, not 1. Advice to make a lambda to compute log-ratio, then call it once to establish desired ratio, then 10 times to check.

varunagrawal: FYI model selection with the normalizing term is not in this PR. There's another open PR for that. That is why we're seeing the ratio as 1 instead of the determinant value.

I'll retarget that PR to this one and update the model selection code (aka change max product to sum product in elimination). That should give us the desired results.

varunagrawal: I added a lambda `compute_ratio` to compute the ratio both for the singular time as well as for `num_samples=10` samples.

dellaert (approved):
Let’s merge

PR #1347, opened 2022-12-23
Hybrid Bayes Net sampling

varunagrawal:
This PR adds a `sample` method to the `HybridBayesNet`, which operates as follows:

1. Compute the discrete bayes net.
2. Sample discrete assignment from the discrete bayes net.
3. Select the Gaussian bayes net corresponding to the sampled assignment.
4. Sample continuous `VectorValues` from the selected Gaussian bayes net.
5. Return both samples as `HybridValues`.

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -20,6 +20,9 @@
 #include <gtsam/hybrid/HybridBayesNet.h>
 #include <gtsam/hybrid/HybridValues.h>
 
+// In Wrappers we have no access to this so have a default ready
+static std::mt19937_64 kRandomNumberGenerator(42);
+
 namespace gtsam {
 
 /* ************************************************************************* */
@@ -230,6 +233,43 @@ HybridValues HybridBayesNet::optimize() const {
 VectorValues HybridBayesNet::optimize(const DiscreteValues &assignment) const {
   GaussianBayesNet gbn = this->choose(assignment);
   return gbn.optimize();
+}
+
+/* ************************************************************************* */
+HybridValues HybridBayesNet::sample(VectorValues given, std::mt19937_64 *rng,
+                                    SharedDiagonal model) const {
+  DiscreteBayesNet dbn;
+  for (size_t idx = 0; idx < size(); idx++) {
+    if (factors_.at(idx)->isDiscrete()) {
+      // If factor at `idx` is discrete-only, we add to the discrete bayes net.
+      dbn.push_back(this->atDiscrete(idx));
+    }
+  }
+  // Sample a discrete assignment.
+  DiscreteValues assignment = dbn.sample();
+  // Select the continuous bayes net corresponding to the assignment.
+  GaussianBayesNet gbn = this->choose(assignment);
+  // Sample from the gaussian bayes net.
+  VectorValues sample = gbn.sample(given, rng, model);
+  return HybridValues(assignment, sample);
+}
+
+/* ************************************************************************* */
+HybridValues HybridBayesNet::sample(std::mt19937_64 *rng,
+                                    SharedDiagonal model) const {
+  VectorValues given;
+  return sample(given, rng, model);
+}
+
+/* ************************************************************************* */
+HybridValues HybridBayesNet::sample(VectorValues given,
+                                    SharedDiagonal model) const {
+  return sample(given, &kRandomNumberGenerator, model);
+}
+
+/* ************************************************************************* */
+HybridValues HybridBayesNet::sample(SharedDiagonal model) const {
+  return sample(&kRandomNumberGenerator, model);
 }
 
 /* ************************************************************************* */
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -236,7 +236,7 @@ VectorValues HybridBayesNet::optimize(const DiscreteValues &assignment) const {
 }
 
 /* ************************************************************************* */
-HybridValues HybridBayesNet::sample(HybridValues &given,
+HybridValues HybridBayesNet::sample(const HybridValues &given,
                                     std::mt19937_64 *rng) const {
   DiscreteBayesNet dbn;
   for (auto &&conditional : *this) {
@@ -261,7 +261,7 @@ HybridValues HybridBayesNet::sample(std::mt19937_64 *rng) const {
 }
 
 /* ************************************************************************* */
-HybridValues HybridBayesNet::sample(HybridValues &given) const {
+HybridValues HybridBayesNet::sample(const HybridValues &given) const {
   return sample(given, &kRandomNumberGenerator);
 }
 
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -8,7 +8,7 @@
 
 /**
  * @file   HybridBayesNet.cpp
- * @brief  A bayes net of Gaussian Conditionals indexed by discrete keys.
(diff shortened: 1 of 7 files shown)


gtsam/linear/GaussianBayesNet.cpp
```diff
@@ -59,27 +59,30 @@ namespace gtsam {
   }
 
   /* ************************************************************************ */
-  VectorValues GaussianBayesNet::sample(std::mt19937_64* rng) const {
+  VectorValues GaussianBayesNet::sample(std::mt19937_64* rng,
+                                        const SharedDiagonal& model) const {
```
dellaert, line 63: Nono ! Does not make sense to me :-)

dellaert (changes requested):
Two comments:
- given should be HybridValues
- Model should not be needed

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -232,6 +235,43 @@ VectorValues HybridBayesNet::optimize(const DiscreteValues &assignment) const {
   return gbn.optimize();
 }
 
+/* ************************************************************************* */
+HybridValues HybridBayesNet::sample(VectorValues given, std::mt19937_64 *rng,
+                                    SharedDiagonal model) const {
```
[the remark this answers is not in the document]
varunagrawal: Interesting. I followed the signature in the other sample methods, so maybe update there as well?

Also it can't be `const` since DiscreteBayesNet does a `sampleInPlace(&given)` and `GaussianBayesNet` has the line `given.insert(sampled);`. Would you like me to update those to make them functional?

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -232,6 +235,43 @@ VectorValues HybridBayesNet::optimize(const DiscreteValues &assignment) const {
...
+
+/* ************************************************************************* */
+HybridValues HybridBayesNet::sample(std::mt19937_64 *rng,
+                                    SharedDiagonal model) const {
+  VectorValues given;
+  return sample(given, rng, model);
+}
+
+/* ************************************************************************* */
+HybridValues HybridBayesNet::sample(VectorValues given,
```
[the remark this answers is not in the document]
varunagrawal: Can't be const for reasons outlined above. This would require a change to `DiscreteConditional` and `GaussianConditional`, which I can do once I get this working properly. 🙂

gtsam/hybrid/HybridBayesNet.cpp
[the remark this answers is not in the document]
dellaert: Oh, now I remember! It is passed by value deliberately, to allow for updating it in-place.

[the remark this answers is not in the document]
dellaert: Should add that in comment so we don’t forget again

gtsam/hybrid/HybridBayesNet.h
```diff
@@ -120,7 +120,47 @@ class GTSAM_EXPORT HybridBayesNet : public BayesNet<HybridConditional> {
...
+   */
+  HybridValues sample(std::mt19937_64 *rng) const;
+
+  /**
+   * @brief Sample from an incomplete BayesNet, use default rng.
+   *
+   * @param given Values of missing variables.
+   * @return HybridValues
+   */
+  HybridValues sample(HybridValues& given) const;
```
dellaert, line 155: Const

gtsam/linear/GaussianConditional.cpp
```diff
@@ -299,14 +299,22 @@ double GaussianConditional::logDeterminant() const {
...
     if (!model_) {
-      throw std::invalid_argument(
-          "GaussianConditional::sample can only be called if a diagonal noise "
-          "model was specified at construction.");
+      // Initialize model_ to Unit noiseModel if not initialized.
+      // Unit noise model specifies sigmas as 1, since
+      // the noise information should already be in information matrix A.
+      // Dim should be number of rows since
+      // noise model dimension should match (Ax - b)
+      _model = noiseModel::Unit::Create(rows());
```
dellaert, line 311: No, sorry, you can’t do that. This method is and should stay const. Just pass the a vector of one’s in sampleDiagonal below.

dellaert (changes requested):

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -173,35 +176,35 @@ HybridBayesNet HybridBayesNet::prune(size_t maxNrLeaves) {
 
 /* ************************************************************************* */
 GaussianMixture::shared_ptr HybridBayesNet::atMixture(size_t i) const {
-  return factors_.at(i)->asMixture();
+  return at(i)->asMixture();
```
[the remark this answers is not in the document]
varunagrawal: It performs indexing followed by conversion, so the `at` makes that apparent when reading.

gtsam/linear/GaussianConditional.cpp
varunagrawal: It is const. I am assigning a new model `_model` (not `model_`) based on the conditions. This is functional since there is no side-effect. 🙂

varunagrawal: Okay made it a lot cleaner. Hopefully it's also good for you.

gtsam/hybrid/HybridBayesNet.cpp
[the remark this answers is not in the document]
varunagrawal: I updated `GaussianBayesNet::sample` to be functional. So instead of `given.insert(sampled)`, it is now `result.insert(sampled)` with `result` initialized as `VectorValues result(given);`.
Perfectly backwards compatible. :)

gtsam/linear/GaussianBayesNet.cpp
```diff
@@ -64,8 +64,9 @@ namespace gtsam {
     return sample(result, rng);
   }
 
-  VectorValues GaussianBayesNet::sample(VectorValues result,
+  VectorValues GaussianBayesNet::sample(const VectorValues& given,
```
dellaert, line 67: Ok, for this case, the question is whether it should be value (which will automatically make the copy you inserted below) or by const&, which you modified this code to. I chose by value when implementing this so I would get the automatic copy, but maybe it does not make a difference and it is more “clever than smart”.  The code that was here was correct, though.

gtsam/linear/GaussianBayesNet.cpp
```diff
@@ -79,7 +80,7 @@ namespace gtsam {
     return sample(&kRandomNumberGenerator);
   }
 
-  VectorValues GaussianBayesNet::sample(VectorValues given) const {
+  VectorValues GaussianBayesNet::sample(const VectorValues& given) const {
```
dellaert, line 83: This, for sure, is a positive change.

dellaert (approved):
Ok, I think one of your changes in GC was not strictly needed, but I *think* it does not make a performance difference, so let’s go with it.

gtsam/linear/GaussianBayesNet.cpp
varunagrawal: I agree it was correct. I felt this made it clearer to read that you're `given` some VectorValues and you're getting back the `result` of sampling.
I can undo this if you'd like.

gtsam/hybrid/HybridBayesNet.cpp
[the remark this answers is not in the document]
dellaert: Yeah, not really :-) the argument was passed *by value* which makes a copy, functionally equivalent to what you changed it to. Modulo possible compiler optimizations that might do something different - no idea which one is best.

PR #1348, opened 2022-12-24
Fix wrapper warnings

varunagrawal:
Fix some warnings from `gtsam_unstable.i` due to the update to `NoiseModelFactorN`.


varunagrawal:
Fixed some additional deprecation warnings as well which are internal to factor implementations, so it shouldn't affect the external API.

dellaert (approved):

gchenfc (approved):
Thanks!  My bad missing these, I think I had turned off unstable for wrappers

Clever use of template in gtsam_unstable.i : IMUFactor 😂

varunagrawal:
Hey man, I'm not a GTSAM wrapper expert without reason. 😎

varunagrawal:
Also no worries. This is why we have labmates to support one other. :)

PR #1349, opened 2022-12-26
Some refactoring and new test

dellaert:
@varunagrawal I refactored the test (and a bit of the elimination code). I expected a normalized DiscreteConditional on m after elimination, rather than an unnormalized factor. I propose you approve and merge this PR and then let's discuss about why that's not true.

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -99,6 +99,7 @@ std::function<double(const Assignment<Key> &, double)> prunerFunc(
 }
 
 /* ************************************************************************* */
+// TODO(dellaert): what is this non-const method used for? Abolish it?
 void HybridBayesNet::updateDiscreteConditionals(
     const DecisionTreeFactor::shared_ptr &prunedDecisionTree) {
   KeyVector prunedTreeKeys = prunedDecisionTree->keys();
@@ -150,9 +151,7 @@ HybridBayesNet HybridBayesNet::prune(size_t maxNrLeaves) {
 
   // Go through all the conditionals in the
   // Bayes Net and prune them as per decisionTree.
-  for (size_t i = 0; i < this->size(); i++) {
-    HybridConditional::shared_ptr conditional = this->at(i);
-
+  for (auto &&conditional : *this) {
     if (conditional->isHybrid()) {
       GaussianMixture::shared_ptr gaussianMixture = conditional->asMixture();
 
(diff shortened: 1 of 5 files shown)


varunagrawal:
> I expected a normalized DiscreteConditional on m after elimination, rather than an unnormalized factor. 

Working on this, #1343 is the WIP PR. I'm having a bit of trouble initializing the updated `HessianFactor` (with Info matrix multiplied by normalized value) since there is no constructor for the Hessian factor that takes just a matrix, and the one for the `SymmetricBlockMatrix` needs the block dimensions which we don't have access to (these are protected methods).

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -432,42 +432,77 @@ TEST(HybridEstimation, ProbabilityMultifrontal) {
...
   nfg.emplace_hybrid<MixtureFactor>(
-      KeyVector{X(0), X(1)}, DiscreteKeys{DiscreteKey(M(0), 2)},
+      KeyVector{X(0), X(1)}, DiscreteKeys{m},
       std::vector<NonlinearFactor::shared_ptr>{zero_motion, one_motion});
 
+  return nfg;
+}
+
+/*********************************************************************************
+  // Create a hybrid nonlinear factor graph f(x0, x1, m0; z0, z1)
```
varunagrawal, line 462: This docstring should say "Create a hybrid linear/gaussian factor graph", right?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -344,68 +344,71 @@ EliminateHybrid(const HybridGaussianFactorGraph &factors,
...
+  // NOTE: We should really defer the product here because of pruning
 
-  // Case 2: we are only dealing with discrete
-  if (allContinuousKeys.empty()) {
+  if (discrete_only) {
+    // Case 1: we are only dealing with discrete
     return discreteElimination(factors, frontalKeys);
-  }
-
+  } else {
```
varunagrawal, line 399: Any particular reason for this nesting? Even an `else if` will only evaluate if `discrete_only` is false.

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -99,6 +99,7 @@ std::function<double(const Assignment<Key> &, double)> prunerFunc(
 }
 
 /* ************************************************************************* */
+// TODO(dellaert): what is this non-const method used for? Abolish it?
```
varunagrawal, line 102: This is used internally to update the BayesNet with the pruned conditionals after pruning them based on the `prunedDecisionTree`.

PR #1351, opened 2022-12-28
Hybrid Model Selection 2

varunagrawal:
Made a separate PR that deals with creating the hybrid sum-product since it has some issues.


dellaert (changes requested):
We should not merge this yet

varunagrawal:
@dellaert and I figured out a better way to incorporate the normalization constant.

PR #1355, opened 2022-12-29
HybridValues specify continuous then discrete

varunagrawal:
Updated `HybridValues` to take continuous values first, and then discrete so that we are consistent.


xsj01 (approved):
LGTM

varunagrawal:
@dellaert FYI

PR #1357, opened 2022-12-30
Test on ratio

dellaert:
@varunagrawal this shows the ratio is broken from the very start :-/ 
I propose you merge this one and previous one, and then do a PR with your improvements (after merging in develop then) (not the one with HessianMatrix changes). Then I'll review so I know what's up and suggest a course of action.
Timing-wise: I have time to look at your PR at around 10:30pm ATL time.

### DIFF
--- a/gtsam/hybrid/HybridValues.h
+++ b/gtsam/hybrid/HybridValues.h
@@ -118,6 +118,12 @@ class GTSAM_EXPORT HybridValues {
    */
   Vector& at(Key j) { return continuous_.at(j); };
 
+  /** For all key/value pairs in \c values, replace values with corresponding keys in this class
+  *   with those in \c values.  Throws std::out_of_range if any keys in \c values are not present
+  *   in this class. */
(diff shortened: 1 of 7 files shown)


gtsam/hybrid/HybridValues.h
```diff
@@ -118,6 +118,12 @@ class GTSAM_EXPORT HybridValues {
    */
   Vector& at(Key j) { return continuous_.at(j); };
 
+  /** For all key/value pairs in \c values, replace values with corresponding keys in this class
+  *   with those in \c values.  Throws std::out_of_range if any keys in \c values are not present
+  *   in this class. */
+  void update(const VectorValues& values) { continuous_.update(values); }
```
varunagrawal, line 124: There's a TODO from @xsj01 that we should probably kill.

dellaert:
Since all of these are small comments and this is an ongoing process, let's resolve all comments in follow up PRs. Still think all this noise is better in a fork.

varunagrawal (approved):

PR #1363, opened 2023-01-03
Fix hybrid tests

varunagrawal:
Fixes the tests after changes due to #1362.

### DIFF
--- a/gtsam/hybrid/HybridConditional.cpp
+++ b/gtsam/hybrid/HybridConditional.cpp
@@ -117,6 +117,22 @@ bool HybridConditional::equals(const HybridFactor &other, double tol) const {
     return other != nullptr && dc->equals(*other, tol);
   }
   return inner_->equals(*(e->inner_), tol);
+
+  if (inner_) {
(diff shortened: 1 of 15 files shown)


gtsam/hybrid/HybridConditional.cpp
```diff
@@ -117,6 +117,22 @@ bool HybridConditional::equals(const HybridFactor &other, double tol) const {
     return other != nullptr && dc->equals(*other, tol);
   }
   return inner_->equals(*(e->inner_), tol);
+
+  if (inner_) {
```
dellaert, line 121: Maybe
```
return inner_ ? (e->inner_ ? inner_->equals(*(e->inner_), tol) : false) : !(e->inner_);
```

dellaert (approved):
Awesome. The one comment can be done in follow up PR. Merge at will.

varunagrawal:
> I hinted at not needing another CO2 wasting CI run :-)

True, but I wanted to keep this self-contained...

PR #1368, opened 2023-01-03
Hybrid Serialization

### LINKED ISSUE #1366: Discrete Serialization Support
Currently, `DecisionTree` doesn't support serialization and testing is not easy since the general round-trip based methods don't work due to `DecisionTree` not implementing the `Testable` traits and its `equals` and `print` method not following the same signature.

This is important since it prevents serialization of hybrid factors and associated containers.

varunagrawal:
Fixed serialization for both discrete and hybrid.
Fixes #1366

### DIFF
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -64,6 +64,9 @@ namespace gtsam {
      */
     size_t nrAssignments_;
 
+    /// Default constructor for serialization.
+    Leaf() {}
+
     /// Constructor from constant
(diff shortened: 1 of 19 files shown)


dellaert (approved):
Cool! But CI fails?

varunagrawal:
Ah the serialization key export is clashing when tests are grouped. I'll figure this out!

varunagrawal:
Easy fix is to move to a common `testSerializationDiscrete.cpp` file. I prefer this since it is cleaner and keeps all the serialization logic together.

PR #1370, opened 2023-01-05
Un-deprecate `NoiseModelFactor1-6` and move `X1` and `key1-6` shortcuts to `NoiseModelFactorN`

gchenfc:
Following the discussion from #1350, this PR replaces #1350 and undoes some changes from #1344:

1. Moves `X1`-`X6` aliases into `NoiseModelFactorN` (not deprecated)
2. Un-deprecates `key1()`-`key6()` shortcuts in `NoiseModelFactorN`
    * and we can also undo some particularly ugly changes needed in #1344 due to templates on dependent types.  The code should still work fine without undo-ing, but I think it's easier to read this way.
3. `#define NoiseModelFactor1 NoiseModelFactorN`.  Now that `NoiseModelFactorN`'s API is a superset of `NoiseModelFactor1-6`, we can simply define it so.

As for 3, `template <T> using NoiseModelFactor1 = NoiseModelFactorN<T>` doesn't work because some detail regarding templates and class name injection. See https://github.com/borglab/gtsam/pull/947#discussion_r765478636

Primary change is in `NonlinearFactor.h`.  The rest of the files are just undo-ing some changes from #1344 and un-disabling deprecation warnings in `testNonlinearFactor.h`.

---

Copy-pasting pros and cons from [previous discussion](https://github.com/borglab/gtsam/pull/1350#issuecomment-1371653077) for reference:

Pros:
* Makes migrating from `NoiseModelFactor1-6` to `NoiseModelFactorN` way easier, since no calling-side code has to change at all.  Just find-all-replace `: public NoiseModelFactor[1-6]` -> `: public NoiseModelFactorN`
* keep things shorter/more readable.  `MyFactor::X3` is (subjectively) cleaner than `MyFactor::ValueType<3>`.
* Still extensible: in the event someone needs e.g. `X7`, then they can just use `ValueType<7>`.
* There's one particular case that makes `ValueType<1>` annoying to use: when you are defining a templated factor **and** you want to use `ValueType<1>` **inside** the definition of the templated factor, you need to do `typename MyFactor::template ValueType<1>` instead of just `ValueType<1>`.  So in these cases, it's much nicer to just use `X1`.

Cons:
* clutters up implementation of `NoiseModelFactorN` making it slightly difficult to read (now inherits from `detail::NoiseModleFactorAliases<ValueTypes>`).
* since we would do `#define NoiseModelFactor1 NoiseModelFactorN`, some compilers may not support deprecating the old classes, and generally `#define` is kind of bad practice (?)

### DIFF
--- a/gtsam/inference/graph-inl.h
+++ b/gtsam/inference/graph-inl.h
@@ -200,8 +200,8 @@ boost::shared_ptr<Values> composePoses(const G& graph, const PredecessorMap<KEY>
     boost::shared_ptr<Factor> factor = boost::dynamic_pointer_cast<Factor>(nl_factor);
     if (!factor) continue;
 
-    KEY key1 = factor->template key<1>();
-    KEY key2 = factor->template key<2>();
+    KEY key1 = factor->key1();
+    KEY key2 = factor->key2();
 
     PoseVertex v1 = key2vertex.find(key1)->second;
     PoseVertex v2 = key2vertex.find(key2)->second;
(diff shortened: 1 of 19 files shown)


dellaert (approved):
Oh, wow, this is much better !!!!

dellaert:
Feel free to merge when CI passes

PR #1374, opened 2023-01-07
Big Re-"Factor" of Hybrid

dellaert:
- made Factor the common base class
- this allowed removing the (cumbersome) wrapper classes

Most changes are in `HybridGaussianFactorGraph`, and the rest is mainly removing files and boilerplate. We also make good use now of `emplace_shared`. 

The dispatch is ugly and possibly slower than vtables (although probably negligible?), so I am also thinking about streamlining even more


varunagrawal:
I'm going to need some time to look into this in detail. A bit lost since I thought we need those wrapper classes for type erasure so whatever sorcery happened to eliminate that needs to be well understood for me.

dellaert:
> I'm going to need some time to look into this in detail. A bit lost since I thought we need those wrapper classes for type erasure so whatever sorcery happened to eliminate that needs to be well understood for me.

It was actually very simple: a HybridFactorGraph is now simply a FG of Factors. You can stick whatever Factor type in there - the idea being that in GTSAM 5 the most general FG is “hybrid”. In the new scheme, the `HybridFactor` class is just the base class for factors that *actually* have both discrete and continuous arguments.

The (annoying, boilerplate generating) sorcery was only needed for our older factors to “pretend” to be Hybrid.

Please don’t take too much time? I want to finish this and try to add an error function in the Factor base class in a next PR. And, I also know how we can get rid of the `GraphAndConstant` thing.

dellaert:
Just removed remaining `emplace_discrete` and `emplace_hybrid`, which are now redundant: `emplace_shared` can be used for all factor types.

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -59,6 +59,8 @@ GaussianMixture::GaussianMixture(
                       Conditionals(discreteParents, conditionals)) {}
 
 /* *******************************************************************************/
+// TODO(dellaert): This is copy/paste: GaussianMixture should be derived from
+// GaussianMixtureFactor, no?
```
ProfFan, line 63: Yes-GM should be a GMF with canonical parameters

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -130,49 +128,54 @@ GaussianFactorGraphTree HybridGaussianFactorGraph::assembleGraphTree() const {
...
-    } else if (auto p = boost::static_pointer_cast<HybridConditional>(factor)) {
-      auto discrete_conditional =
-          boost::static_pointer_cast<DiscreteConditional>(p->inner());
-      dfg.push_back(discrete_conditional);
+  for (auto &f : factors) {
+    if (auto dtf = dynamic_pointer_cast<DecisionTreeFactor>(f)) {
+      dfg.push_back(dtf);
+    } else if (auto orphan = dynamic_pointer_cast<OrphanWrapper>(f)) {
+      // Ignore orphaned clique.
+      // TODO(dellaert): is this correct? If so explain here.
```
ProfFan, line 165: OrphanWrapper is one of the things that seems a bit "non-canonical" to me in the elimination algorithm.

ProfFan (approved):

gtsam/hybrid/GaussianMixture.cpp
varunagrawal: When you say canonical parameters, do you mean canonical parameters of the Gaussian or something else?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -130,49 +128,54 @@ GaussianFactorGraphTree HybridGaussianFactorGraph::assembleGraphTree() const {
...
                       const Ordering &frontalKeys) {
   GaussianFactorGraph gfg;
-  for (auto &fp : factors) {
-    if (auto ptr = boost::dynamic_pointer_cast<HybridGaussianFactor>(fp)) {
-      gfg.push_back(ptr->inner());
-    } else if (auto ptr = boost::static_pointer_cast<HybridConditional>(fp)) {
-      gfg.push_back(
-          boost::static_pointer_cast<GaussianConditional>(ptr->inner()));
+  for (auto &f : factors) {
+    if (auto gf = dynamic_pointer_cast<GaussianFactor>(f)) {
```
varunagrawal, line 136: I think I see why we're doing `FactoGraph<Factor>` and using `Factor` everywhere instead of `HybridFactor`.

We can make our lives much easier by having the hierarchy:

```
Factor
- SymbolicFactor
- HybridFactor
  - GaussianFactor
  - DiscreteFactor 
```

I mentioned this in the meeting we had. `SymbolicFactor` is not handled by the machinery here so it would have to be a sibling of `HybridFactor` until we provide that capability. Else merging HybridFactor and Factor is also an option.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: Define non-canonical please?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -428,35 +424,9 @@ EliminateHybrid(const HybridGaussianFactorGraph &factors,
   }
 }
 
-/* ************************************************************************ */
```
varunagrawal, line 431: I was super doubtful when these were first added since `GaussianFactorGraph` and `DiscreteFactorGraph` don't have these, but I guess that PR was merged before I could question this.

Glad to see we're cutting down on the sugar. 😉

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -469,33 +439,25 @@ AlgebraicDecisionTree<Key> HybridGaussianFactorGraph::error(
...
-  for (size_t idx = 0; idx < size(); idx++) {
+  for (auto &f : factors_) {
     // TODO(dellaert): just use a virtual method defined in HybridFactor.
     AlgebraicDecisionTree<Key> factor_error;
 
-    if (factors_.at(idx)->isHybrid()) {
-      // If factor is hybrid, select based on assignment.
-      GaussianMixtureFactor::shared_ptr gaussianMixture =
-          boost::static_pointer_cast<GaussianMixtureFactor>(factors_.at(idx));
+    if (auto gaussianMixture = dynamic_pointer_cast<GaussianMixtureFactor>(f)) {
```
varunagrawal, line 446: This makes me sad. `f->isHybrid()` is much cleaner and self-documenting. This is just hard to read and more intimidating (e.g. if I was a new contributor).

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -505,8 +467,17 @@ AlgebraicDecisionTree<Key> HybridGaussianFactorGraph::error(
 /* ************************************************************************ */
 double HybridGaussianFactorGraph::error(const HybridValues &values) const {
   double error = 0.0;
-  for (auto &factor : factors_) {
-    error += factor->error(values);
+  for (auto &f : factors_) {
+    if (auto hf = dynamic_pointer_cast<GaussianFactor>(f)) {
```
varunagrawal, line 471: More sadness.

gtsam/hybrid/HybridNonlinearFactorGraph.cpp
```diff
@@ -50,47 +43,38 @@ void HybridNonlinearFactorGraph::print(const std::string& s,
...
-        }
-        // Finally if nothing else, we are discrete-only which doesn't need
-        // lineariztion.
-      } else {
-        linearFG->push_back(factor);
-      }
-
-    } else {
+    if (!f) {
+      // TODO(dellaert): why?
```
varunagrawal, line 57: Good question. I can look into this.

gtsam/hybrid/GaussianMixture.cpp
ProfFan: > 

Yeah

varunagrawal (approved):
To be honest, I am not a fan of this update. We're sacrificing a lot of readability and code-convenience just for the sake of removing a few "add/insert" methods which seems like a skewed pros-to-cons ratio.

I think the correct way forward that helps maintain all the readability is to make HybridFactor a superclass of DiscreteFactor, GaussianFactor and NonlinearFactor (basically every factor that the HybridFactorGraph is capable of handling currently). This will result in clean code but with the desired API design.

This is important going forward since if Hybrid is to be the de-facto mechanism, we don't want contributors balking at the current complexity and difficulty in parsing the code (especially if they're not proficient in C++ at your level, which is pretty darn high).

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -79,48 +88,37 @@ static GaussianFactorGraphTree addGaussian(
...
-      // BayesTreeOrphanWrapper!
-      auto orphan = boost::dynamic_pointer_cast<
-          BayesTreeOrphanWrapper<HybridBayesTree::Clique>>(f);
-      if (!orphan) {
-        auto &fr = *f;
-        throw std::invalid_argument(
-            std::string("factor is discrete in continuous elimination ") +
-            demangle(typeid(fr).name()));
-      }
+      // TODO(dellaert): there was an unattributed comment here: We need to
```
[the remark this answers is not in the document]
dellaert: Normally, TODOs come with a username so we can ask who wrote the comment. I have no idea whether something needs to happen here or not, or it should be ignored.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -130,49 +128,54 @@ GaussianFactorGraphTree HybridGaussianFactorGraph::assembleGraphTree() const {
 }
 
 /* ************************************************************************ */
-static std::pair<HybridConditional::shared_ptr, HybridFactor::shared_ptr>
+static std::pair<HybridConditional::shared_ptr, boost::shared_ptr<Factor>>
```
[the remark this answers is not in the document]
dellaert: HybridFactor is now only for *truly* hybrid factors. A HybridGaussianFactorGraph needs to also store GaussianFactors, *and* DiscreteFactors. The common base class (supertype) is Factor.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Yeah, I opted for simplicity and with an eye towards GTSAM 5.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: It should be handled with virtual methods. Next PR

gtsam/hybrid/HybridFactorGraph.h
```diff
@@ -11,50 +11,40 @@
...
 
 /**
  * Hybrid Factor Graph
- * -----------------------
- * This is the base hybrid factor graph.
- * Everything inside needs to be hybrid factor or hybrid conditional.
+ * Factor graph with utilities for hybrid factors.
  */
-class HybridFactorGraph : public FactorGraph<HybridFactor> {
+class HybridFactorGraph : public FactorGraph<Factor> {
```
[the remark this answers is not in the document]
dellaert: I agree, FactorGraph is already hybrid. In this case, HybridFactorGraph is really just a few utility methods on top of FactorGraph. This is the entire epiphany we discussed yesterday.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
ProfFan: Seems like a hack

dellaert:
> To be honest, I am not a fan of this update. We're sacrificing a lot of readability and code-convenience just for the sake of removing a few "add/insert" methods which seems like a skewed pros-to-cons ratio.

I disagree. I think everything become much simpler, and simpler to understand. The new style: just use `emplace_shared`. It names the thing it is inserting, and avoids a copy. Examples abound in the unit tests, which is what people copy, anyway. The previous code was baroque.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
ProfFan: `SymbolicFactor` is really *in my opinion* a way to reuse the elimination templates for symbolic factorization. Thus a `SymbolicFactor` in some sense is not a `Factor`...

gtsam/hybrid/GaussianMixture.cpp
dellaert: Don't worry about it, I will resolve in a subsequent PR. I think the copy/pasta is a consequence of the type erasure dance.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: I don’t disagree. In hindsight, maybe a better option is to do the symbolic elimination just with the keys in the factors. But, not having to worry about making “working” numerical factor graphs to test out elimination methods is very useful. One option is to make the Factor base class non-abstract so it can “take over” for symbolic.

varunagrawal:
> I disagree. I think everything become much simpler, and simpler to understand. The new style: just use `emplace_shared`. It names the thing it is inserting, and avoids a copy. Examples abound in the unit tests, which is what people copy, anyway. The previous code was baroque.

I agree there was a lot of extra fat in the previous code, and this PR is a wonderful for addressing that! But all the pointer up- and down-casting is what I am referring to as being really confusing and less readable, while there exists an alternative that gives us the best of both worlds.

dellaert:
> > I disagree. I think everything become much simpler, and simpler to understand. The new style: just use `emplace_shared`. It names the thing it is inserting, and avoids a copy. Examples abound in the unit tests, which is what people copy, anyway. The previous code was baroque.
> 
> I agree there was a lot of extra fat in the previous code, and this PR is a wonderful for addressing that! But all the pointer up- and down-casting is what I am referring to as being really confusing and less readable, while there exists an alternative that gives us the best of both worlds.

I hear you. But (a) that code is mostly internal and pretty localized, (b) I have a follow-up PR that at least removes some of the up-casting. Note virtual methods are just cleaner upcasting.

PR #1376, opened 2023-01-09
More Expression wrappers

BrettRD:
This pull request adds Expressions for:
* Range<Point2, Point2>
* Pose3::transformPoseTo
* Pose3::translation
* Unit3::point3
* OrientedPlane3::distance
* OrientedPlane3::normal
* PinholeCamera::getPose

This pull request also adds Jacobians to the getter functions OrientedPlane3::distance and OrientedPlane3::normal

Let me know if any of these need tests, I couldn't spot any existing test patterns for these kinds of wrappers
I have some examples for these that I can share as well

### DIFF
--- a/gtsam/geometry/OrientedPlane3.h
+++ b/gtsam/geometry/OrientedPlane3.h
@@ -125,12 +125,14 @@ public:
   }
 
   /// Return the normal
-  inline Unit3 normal() const {
+  inline Unit3 normal(OptionalJacobian<2, 3> H = boost::none) const {
+    if (H) *H << I_2x2, Z_2x1;
     return n_;
   }
 
   /// Return the perpendicular distance to the origin
-  inline double distance() const {
+  inline double distance(OptionalJacobian<1, 3> H = boost::none) const {
+    if (H) *H << 0,0,1;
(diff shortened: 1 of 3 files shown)


dellaert:
Awesome! No need to test the expressions, but the new derivatives in Unit3 should have a test :-) Are you up for adding those?

BrettRD:
I can give it a whirl,
Do you mean the new derivatives in `OrientedPlane3.h`?
Would the test end up `testExpressionFactor.cpp`?

dellaert:
Yep, those. They should be tested in `testOrientedPlane.cpp`, look at `TEST(OrientedPlane3, jacobian_retract)` for an example on using functions from `numericalDerivative.h`.

BrettRD:
I was just thinking the explicit-value tests looked superfluous,
I'll try rebuilding those and actually refer to the tangent space

BrettRD:
I think the new tests are what you're after.
I left in the test that compares to retract(zero) because it's fast and readable.

dellaert (approved):
Looks awesome ! Will merge when CI passes. Thanks !!!

PR #1377, opened 2023-01-10
New-style Errors

dellaert:
In an effort to get past the Windows compilation problems on #1375, and, because "friends don't let friends inherit from STL containers", here is an attempt to define `Errors` as a typedef to `FastList<Vector>`.

This might still run into the same problem of needing a compare operator for `merge` on Windows, we'll see in CI...

Thanks @gchenfc for your help in identifying the issue.

### DIFF
--- a/gtsam/linear/Errors.h
+++ b/gtsam/linear/Errors.h
@@ -20,59 +20,54 @@
 #pragma once
 
 #include <gtsam/base/FastList.h>
-#include <gtsam/base/Vector.h>
 #include <gtsam/base/Testable.h>
+#include <gtsam/base/Vector.h>
 
 #include <string>
 
 namespace gtsam {
 
-  // Forward declarations
-  class VectorValues;
+// Forward declarations
+class VectorValues;
 
-  /** vector of errors */
-  class Errors : public FastList<Vector> {
+/// Errors is a vector of errors.
+using Errors = FastList<Vector>;
 
-  public:
+/// Break V into pieces according to its start indices.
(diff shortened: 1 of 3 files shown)


gtsam/linear/Errors.h
```diff
@@ -20,59 +20,54 @@
...
-    /** Addition */
-    GTSAM_EXPORT Errors operator+(const Errors& b) const;
+/// Print an Errors instance.
+GTSAM_EXPORT void print(const Errors& e, const std::string& s = "Errors");
 
-    /** subtraction */
-    GTSAM_EXPORT Errors operator-(const Errors& b) const;
+// Check equality for unit testing.
+GTSAM_EXPORT bool equality(const Errors& actual, const Errors& expected,
+                           double tol = 1e-9);
```
gchenfc, line 44: These three functions break backward compatibility for people not using `traits`, though I'm sure you were already aware.

gchenfc (approved):

gchenfc:
CI failures appear to just be wrapper API change: https://github.com/borglab/gtsam/blob/16136ea86adcf97e26edb750bdd044b4c6ca18f4/gtsam/linear/linear.i#L632-L637

unclear how to best handle this.

(Linux CI failure appears to just be github hiccup)

dellaert:
I have killed the wrapper in #1375 , it was not used and very unlikely to be used by anyone outside. I will close this PR as I incorporated changes in #1375 the moment windows succeeded :-) thanks again for your help!

PR #1379, opened 2023-01-11
replace list_of with initializer list in TimeTBB.cpp

kartikarcot:


dellaert (approved):
Woohoo! Thanks, I missed those :-)

dellaert:
Thanks @kartikarcot !

PR #1380, opened 2023-01-11
Harmonized error/logProbability in entire Factor/Conditional hierarchy

dellaert:
Cleaning up before releasing 4.2 and implementing last hybrid PR
- renamed logDensity to logProbability, as density is continuous only
- we now have logProbability(x) = log(evaluate(x)) = K - error(x) for all conditionals
- probPrime(x) = exp(-error(x)) as always, but is only supposed to be used in factors

My advice for reviewing (seems daunting but most is rote):
- start with Factor.h, Conditional.h and the rest of inference - read the updated docs (and please flag issues).
- Then hybrid.
- Then linear, discrete, symbolic

Note that some methods are gone as now implemented in bottom base classes.

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -257,29 +257,7 @@ VectorValues HybridBayesNet::optimize(const DiscreteValues &assignment) const {
 
 /* ************************************************************************* */
 double HybridBayesNet::evaluate(const HybridValues &values) const {
-  const DiscreteValues &discreteValues = values.discrete();
-  const VectorValues &continuousValues = values.continuous();
-
-  double error = 0.0, probability = 1.0;
-
-  // Iterate over each conditional.
-  for (auto &&conditional : *this) {
-    // TODO: should be delegated to derived classes.
-    if (auto gm = conditional->asMixture()) {
-      const auto component = (*gm)(discreteValues);
-      error += component->error(continuousValues);
-
-    } else if (auto gc = conditional->asGaussian()) {
-      // If continuous only, evaluate the probability and multiply.
-      error += gc->error(continuousValues);
-
-    } else if (auto dc = conditional->asDiscrete()) {
-      // Conditional is discrete-only, so return its probability.
-      probability *= dc->operator()(discreteValues);
-    }
-  }
-
-  return probability * exp(-error);
+  return exp(-error(values));
 }
 
 /* ************************************************************************* */
@@ -318,12 +296,6 @@ HybridValues HybridBayesNet::sample() const {
 }
 
 /* ************************************************************************* */
-double HybridBayesNet::error(const HybridValues &values) const {
-  GaussianBayesNet gbn = choose(values.discrete());
-  return gbn.error(values.continuous());
-}
-
-/* ************************************************************************* */
 AlgebraicDecisionTree<Key> HybridBayesNet::error(
     const VectorValues &continuousValues) const {
   AlgebraicDecisionTree<Key> error_tree(0.0);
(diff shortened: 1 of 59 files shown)


gtsam/inference/Factor.h
```diff
@@ -128,6 +145,12 @@ typedef FastSet<FactorIndex> FactorIndexSet;
    /** Iterator at end of involved variable keys */
    const_iterator end() const { return keys_.end(); }
 
+  /**
+   * All factor types need to implement an error function.
+   * In factor graphs, this is the negative log-likelihood.
+   */
+  virtual double error(const HybridValues& c) const;
```
ProfFan, line 152: indent

gtsam/symbolic/SymbolicConditional.h
```diff
@@ -17,10 +17,10 @@
 
 #pragma once
 
-#include <gtsam/symbolic/SymbolicFactor.h>
-#include <gtsam/inference/Conditional.h>
 #include <gtsam/base/Testable.h>
 #include <gtsam/base/types.h>
+#include <gtsam/inference/Conditional-inst.h>
```
ProfFan, line 22: `inst` headers should not be included?

ProfFan (approved):
Most looks great to me!

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -318,33 +291,28 @@ HybridValues HybridBayesNet::sample() const {
...
-double HybridBayesNet::error(const HybridValues &values) const {
-  GaussianBayesNet gbn = choose(values.discrete());
-  return gbn.error(values.continuous());
-}
-
-/* ************************************************************************* */
-AlgebraicDecisionTree<Key> HybridBayesNet::error(
+AlgebraicDecisionTree<Key> HybridBayesNet::logProbability(
     const VectorValues &continuousValues) const {
   AlgebraicDecisionTree<Key> error_tree(0.0);
```
varunagrawal, line 296: Rename `error_tree` to `log_prob_tree`?

gtsam/hybrid/HybridValues.h
```diff
@@ -37,12 +37,15 @@ namespace gtsam {
...
  private:
-  // VectorValue stored the continuous components of the HybridValues.
+  /// Continuous multi-dimensional vectors for \class GaussianFactor.
   VectorValues continuous_;
 
-  // DiscreteValue stored the discrete components of the HybridValues.
+  /// Discrete values for \class DiscreteFactor.
   DiscreteValues discrete_;
 
+  /// Continuous, differentiable manifold values for \class NonlinearFactor.
```
varunagrawal, line 46: We should have `HybridVectorValues` and `HybridValues`. This is very unclean.

gtsam/inference/BayesNet.h
```diff
@@ -86,6 +87,13 @@ class BayesNet : public FactorGraph<CONDITIONAL> {
                  const KeyFormatter& keyFormatter = DefaultKeyFormatter,
                  const DotWriter& writer = DotWriter()) const;
 
+  /// @}
+  /// @name HybridValues methods
+  /// @{
+
+  double logProbability(const HybridValues& x) const;
```
varunagrawal, line 94: Documentation is missing.

python/gtsam/tests/test_DiscreteBayesNet.py
```diff
@@ -11,13 +11,15 @@
 
 # pylint: disable=no-name-in-module, invalid-name
 
+import math
```
varunagrawal, line 14: You don't need math. Rather use numpy.
E.g. `np.log`.

python/gtsam/tests/test_GaussianBayesNet.py
```diff
@@ -12,13 +12,15 @@
 
 from __future__ import print_function
 
+import math
```
varunagrawal, line 15: Use `numpy` instead.

varunagrawal (changes requested):

gtsam/symbolic/SymbolicConditional.h
dellaert: Had linking errors that I could not resolve for the life of me.

gtsam/hybrid/HybridBayesNet.cpp
dellaert: Sure

gtsam/hybrid/HybridValues.h
dellaert: I disagree

gtsam/inference/BayesNet.h
dellaert: Will fix

gtsam/hybrid/HybridBayesNet.cpp
dellaert: Renamed to `result`. Note that naming convention for variables in gtsam is camlCase.

varunagrawal (approved):

PR #1386, opened 2023-01-14
Test and fix all conditionals

dellaert:
Turns out GaussianMixture error() was wrong! I added very thorough checking of all conditionals with all supported arguments, in both C++ and python, and made sure all compiled and worked. In the process I found out that GaussianMixture did *not* satisfy its invariants. The resolution is documented in GaussianMixture.h:

```
   * @brief Compute the error of this Gaussian Mixture.
   *
   * This requires some care, as different mixture components may have
   * different normalization constants. Let's consider p(x|y,m), where m is
   * discrete. We need the error to satisfy the invariant:
   *
   *    error(x;y,m) = K - log(probability(x;y,m))
   *
   * For all x,y,m. But note that K, for the GaussianMixture, cannot depend on
   * any arguments. Hence, we delegate to the underlying Gaussian
   * conditionals, indexed by m, which do satisfy:
   * 
   *    log(probability_m(x;y)) = K_m - error_m(x;y)
   * 
   * We resolve by having K == 0.0 and
   * 
   *    error(x;y,m) = error_m(x;y) - K_m
```

This was a lot of work but I think this level of care is needed to ensure correctness in hybrid inference.

### DIFF
--- a/gtsam/hybrid/GaussianMixture.h
+++ b/gtsam/hybrid/GaussianMixture.h
@@ -176,7 +176,21 @@ class GTSAM_EXPORT GaussianMixture
   /**
    * @brief Compute the error of this Gaussian Mixture.
    *
-   * log(probability(x)) = K - error(x)
+   * This requires some care, as different mixture components may have
+   * different normalization constants. Let's consider p(x|y,m), where m is
+   * discrete. We need the error to satisfy the invariant:
+   *
+   *    error(x;y,m) = K - log(probability(x;y,m))
+   *
+   * For all x,y,m. But note that K, for the GaussianMixture, cannot depend on
+   * any arguments. Hence, we delegate to the underlying Gaussian
+   * conditionals, indexed by m, which do satisfy:
+   * 
+   *    log(probability_m(x;y)) = K_m - error_m(x;y)
+   * 
+   * We resolve by having K == 0.0 and
+   * 
+   *    error(x;y,m) = error_m(x;y) - K_m
    *
    * @param values Continuous values and discrete assignment.
    * @return double
--- a/gtsam/hybrid/GaussianMixture.h
+++ b/gtsam/hybrid/GaussianMixture.h
@@ -174,8 +174,17 @@ class GTSAM_EXPORT GaussianMixture
       const VectorValues &continuousValues) const;
 
   /**
-   * @brief Compute the logProbability of this Gaussian Mixture given the
-   * continuous values and a discrete assignment.
+   * @brief Compute the error of this Gaussian Mixture.
+   * 
+   * log(probability(x)) = K - error(x)
+   *
+   * @param values Continuous values and discrete assignment.
+   * @return double
+   */
+  double error(const HybridValues &values) const override;
+
+  /**
+   * @brief Compute the logProbability of this Gaussian Mixture.
    *
    * @param values Continuous values and discrete assignment.
    * @return double
--- a/gtsam/hybrid/GaussianMixture.h
+++ b/gtsam/hybrid/GaussianMixture.h
@@ -175,7 +175,7 @@ class GTSAM_EXPORT GaussianMixture
 
   /**
    * @brief Compute the error of this Gaussian Mixture.
-   * 
+   *
    * log(probability(x)) = K - error(x)
    *
    * @param values Continuous values and discrete assignment.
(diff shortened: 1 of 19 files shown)


dellaert:
For reviewing, look at `Conditional-inst.h`, the new file `testHybridConditional.cpp`, `test_HybridBayesNet.py` and `GaussianMixture.*`. Almost everything else is just small additions to make things compile.

gtsam/inference/Conditional.h
```diff
@@ -172,7 +181,17 @@ namespace gtsam {
     /** Mutable iterator pointing past the last parent key. */
     typename FACTOR::iterator endParents() { return asFactor().end(); }
 
+    template <class VALUES>
+    static bool CheckInvariants(const DERIVEDCONDITIONAL& conditional,
```
varunagrawal, line 185: Docs please?

gtsam/hybrid/GaussianMixture.h
```diff
@@ -174,20 +174,44 @@ class GTSAM_EXPORT GaussianMixture
...
   /**
-   * @brief Compute the logProbability of this Gaussian Mixture given the
-   * continuous values and a discrete assignment.
+   * @brief Compute the error of this Gaussian Mixture.
+   *
+   * This requires some care, as different mixture components may have
+   * different normalization constants. Let's consider p(x|y,m), where m is
+   * discrete. We need the error to satisfy the invariant:
+   *
+   *    error(x;y,m) = K - log(probability(x;y,m))
```
varunagrawal, line 183: $K$ is not defined. It is defined in `GaussianConditional.cpp` but a reader should not have to go there for this.

gtsam/hybrid/GaussianMixture.h
```diff
@@ -174,20 +174,44 @@ class GTSAM_EXPORT GaussianMixture
...
+   *
+   * For all x,y,m. But note that K, for the GaussianMixture, cannot depend on
```
varunagrawal, line 185: I am not sure I follow the reasoning for this (other than the fact that the K is the inverse of the sum of all the conditionals by definition). Can you please elaborate?

varunagrawal (approved):
Approving, but I would really like to understand why K cannot be dependent on any arguments. Or should it be "K should not depend on any arguments"?

gtsam/hybrid/GaussianMixture.h
dellaert: I'll add.

gtsam/hybrid/GaussianMixture.h
dellaert: I think you reviewed and read comment in `Conditional.h`? Is it not clear in light of that?

dellaert:
> Approving, but I would really like to understand why K cannot be dependent on any arguments. Or should it be "K should not depend on any arguments"?

See my comment above. It is predicated on invariants set forth in Conditional.h. I will do some more commits in this PR as I think GaussianMixtureFactor is still wrong as well.

gtsam/hybrid/GaussianMixture.h
```diff
@@ -174,20 +174,44 @@ class GTSAM_EXPORT GaussianMixture
       const VectorValues &continuousValues) const;
 
   /**
-   * @brief Compute the logProbability of this Gaussian Mixture given the
-   * continuous values and a discrete assignment.
+   * @brief Compute the error of this Gaussian Mixture.
+   *
+   * This requires some care, as different mixture components may have
+   * different normalization constants. Let's consider p(x|y,m), where m is
```
ProfFan, line 180: Yes, in the past I mental-modeled this as a discrete prior

gtsam/hybrid/GaussianMixture.h
varunagrawal: I think it should say "K should not depend" and not "K cannot depend", since the "cannot" contradicts the statement in `Conditionals.h` about overriding.

ProfFan (approved):
I think this looks great.

To explain the previous behavior: I put the normalization constants as a DiscreteFactor, which should be somewhat equivalent, but of course a lot more hacky.

ProfFan:
Yeah, GMF will need the same modifications

gtsam/hybrid/GaussianMixture.h
dellaert: i think that is slightly different and is still useful. A "prior" would choose probabilistically between different components, even if they had the same normalization constant (which depends on the noise).

I would actually prefer to rename this class, `P(x|m)`, to GaussianComponent, and reserver GaussianMixture for `P(x|m)P(m)`.

dellaert:
I started addressing issues with GMF in a different PR, and I also made the requested changes there…

PR #1387, opened 2023-01-15
Some Hybrid Improvements

varunagrawal:
1. `HybridFactorGraph::discreteKeys` returns a discrete key vector aka `DiscreteKeys` since they are the main workhorse in discrete.
2. Made `discreteKeySet` more efficient by using a STL-style container transform instead of a loop. This also makes the transformation more _functional_.
3. Implement `MixtureFactor::error(HybridValues)` since it was low hanging fruit.
4. Pass in the formatter when printing the mean of a `GaussianConditional`. Found this while debugging since I use GTD keys.

### DIFF
--- a/gtsam/hybrid/HybridFactorGraph.cpp
+++ b/gtsam/hybrid/HybridFactorGraph.cpp
@@ -25,7 +25,7 @@
 namespace gtsam {
 
 /* ************************************************************************* */
-std::set<DiscreteKey> HybridFactorGraph::discreteKeys() const {
+DiscreteKeys HybridFactorGraph::discreteKeys() const {
   std::set<DiscreteKey> keys;
   for (auto& factor : factors_) {
     if (auto p = boost::dynamic_pointer_cast<DecisionTreeFactor>(factor)) {
@@ -39,7 +39,7 @@ std::set<DiscreteKey> HybridFactorGraph::discreteKeys() const {
       }
     }
   }
-  return keys;
+  return DiscreteKeys(keys.begin(), keys.end());
 }
 
 /* ************************************************************************* */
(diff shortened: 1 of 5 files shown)


gtsam/hybrid/HybridFactorGraph.cpp
```diff
@@ -25,7 +25,7 @@
 namespace gtsam {
 
 /* ************************************************************************* */
-std::set<DiscreteKey> HybridFactorGraph::discreteKeys() const {
+DiscreteKeys HybridFactorGraph::discreteKeys() const {
```
dellaert, line 28: I disagree with this. There is no possible use for this method, as I can’t think of a use case where we rely on the order of a set of concatenation keys.

dellaert (commented):
I’m requesting to leave the keys methods as they are.

gtsam/hybrid/HybridFactorGraph.cpp
varunagrawal: I disagree. There are multiple methods that take `DiscreteKeys` as an argument in `discrete`. Having a convenient way to get the discrete key vector makes sense.

dellaert: It would then have to be private, as we would not want to expose an "intermediate-only" method in the API. But I see you reverted this change, so I think we're good now.

dellaert (approved):

gtsam/hybrid/HybridFactorGraph.cpp
varunagrawal: Why would it be intermediate?

Take CartesianProduct for example. It takes DiscreteKeys and is used for enumerating through all the values of a DiscreteTree.

PR #1388, opened 2023-01-17
Hybrid/simplify

dellaert:
This PR implements the scheme in the (included) [doc/Hybrid.pdf](https://github.com/borglab/gtsam/files/10431114/Hybrid.pdf), where `GaussianMixture::likelihood` returns a single `GaussianMixtureFactor` instance, that encodes the log-normalization constant as well - by adding a row to A and b, respectively. This is only needed when the noise is mode-dependent.

We also now enforce error>0, and define the normalization constant for the `GaussianMixture` as the max of all normalization constants of its components.

With this scheme, there is no more need for `FactorAndConstant` and `GraphAndConstant`, which have been removed.

To make the elimination work, I use the correct formula in the hybrid elimination: 
```
q(μ;m) = exp(-error(μ;m)) * sqrt(det(2π Σ_m))
```

There are two tests I had to disable:
1. a pruning test that now yields infinity. @varunagrawal look into that?
2. a complex example where both noise model and motion model are governed by discrete variables, fails the ratio test after elimination. Please take a detailed look there whether I made a silly mistake? Hoping this one does not point to yet another issue with elimination.


dellaert:
OK @varunagrawal , @ProfFan I fixed the last issue. One of the things GraphAndConstant did was to correct for the newly created GaussianConditional normalization factors. I now did this in a second path in the hybrid elimination step, by using `HessianFactor::constantTerm()`.

Could you review? A bit of a deadline, as I want to land 4.2, and this was the missing piece. Also, @varunagrawal , consult on the pruning? I had to change the regression tests in `testHybridBayesNet`, but I believe the results to be correct now (and not before).

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -276,39 +259,46 @@ hybridElimination(const HybridGaussianFactorGraph &factors,
...
+    // Otherwise, we create a resulting GaussianMixtureFactor on the separator,
+    // taking care to correct for conditional constant.
+
+    // Correct for the normalization constant used up by the conditional
+    auto correct = [&](const Result &pair) -> GaussianFactor::shared_ptr {
+      const auto &factor = pair.second;
+      if (!factor) return factor;  // TODO(dellaert): not loving this.
+      auto hf = boost::dynamic_pointer_cast<HessianFactor>(factor);
+      if (!hf) throw std::runtime_error("Expected HessianFactor!");
+      hf->constantTerm() += 2.0 * pair.first->logNormalizationConstant();
```
dellaert, line 291: @varunagrawal @ProfFan this is the correction that was missing.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -276,39 +259,46 @@ hybridElimination(const HybridGaussianFactorGraph &factors,
...
+    // DiscreteFactor here, with the error for each discrete choice.
+
+    // Integrate the probability mass in the last continuous conditional using
+    // the unnormalized probability q(μ;m) = exp(-error(μ;m)) at the mean.
+    //   discrete_probability = exp(-error(μ;m)) * sqrt(det(2π Σ_m))
+    auto probability = [&](const Result &pair) -> double {
+      static const VectorValues kEmpty;
+      // If the factor is not null, it has no keys, just contains the residual.
+      const auto &factor = pair.second;
+      if (!factor) return 1.0;  // TODO(dellaert): not loving this.
```
ProfFan, line 273: How could this branch be taken?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -276,39 +259,46 @@ hybridElimination(const HybridGaussianFactorGraph &factors,
...
+      return exp(-factor->error(kEmpty)) / pair.first->normalizationConstant();
```
ProfFan, line 274: Maybe a `inverseNormalizationConstant` is better, idk.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
ProfFan: Ah! I think this is the same method in Murphy02thesis, using the HessianFactor (which is in the canonical form of a Gaussian).

I regret totally missing this...

ProfFan (approved):

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -63,28 +74,22 @@ GaussianMixture::GaussianMixture(
...
-  auto lambda = [](const GaussianConditional::shared_ptr &conditional) {
-    GaussianFactorGraph result;
-    result.push_back(conditional);
-    if (conditional) {
-      return GraphAndConstant(result, conditional->logNormalizationConstant());
-    } else {
-      return GraphAndConstant(result, 0.0);
-    }
+  auto wrap = [](const GaussianConditional::shared_ptr &gc) {
+    return GaussianFactorGraph{gc};
```
varunagrawal, line 90: Why remove the check for null?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Yeah, whenever we do BayesTree there is some way (that I don't yet understand) that nullptrs are used, and we are dealing with all over the place. Can you explain? Or @varunagrawal ?

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -210,67 +223,62 @@ TEST(HybridBayesNet, Optimize) {
...
-                       hybridBayesNet->logProbability(hybridValues), 1e-9);
 }
 
 /* ****************************************************************************/
 // Test Bayes net pruning
 TEST(HybridBayesNet, Prune) {
   Switching s(4);
 
-  HybridBayesNet::shared_ptr hybridBayesNet =
+  HybridBayesNet::shared_ptr posterior =
```
varunagrawal, line 275: umm why? It would read nicely as `hybridBayesNet` and would be very readable.

varunagrawal (commented):
This could be broken down into smaller PRs...

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Interesting idea, but we already have too many methods floating around.

gtsam/hybrid/tests/testHybridBayesNet.cpp
dellaert: trying to get away from using types as variable names if there is a clear meaning.

gtsam/hybrid/GaussianMixture.cpp
dellaert: The only reason was to make sure we could call `logNormalizationConstant`, which we no longer need to do.

varunagrawal:
@dellaert I'm a bit confused. Have we gone from using probabilities in the discrete trees to logProbabilities? If that's the case, then the pruning methods need to be updated.

dellaert:
> @dellaert I'm a bit confused. Have we gone from using ~probabilities~ errors in the discrete trees to logProbabilities? If that's the case, then the pruning methods need to be updated.

Yeah, this is why I asked your consult. I'm looking at it now.

dellaert:
> > @dellaert I'm a bit confused. Have we gone from using ~probabilities~ errors in the discrete trees to logProbabilities? If that's the case, then the pruning methods need to be updated.
> 
> Yeah, this is why I asked your consult. I'm looking at it now.

Unfortunately, I am not able to decode the pruning code. Can we chat?

varunagrawal:
> > > @dellaert I'm a bit confused. Have we gone from using ~probabilities~ errors in the discrete trees to logProbabilities? If that's the case, then the pruning methods need to be updated.
> > 
> > 
> > Yeah, this is why I asked your consult. I'm looking at it now.
> 
> Unfortunately, I am not able to decode the pruning code. Can we chat?

Yeah the fix is easy: the pruning code sets the probability to 0, so `log(0) = -inf`. We have to just update the regression parameter.

varunagrawal:
@dellaert I'm making a PR for this now.

dellaert:
> @dellaert I'm making a PR for this now.

@varunagrawal  No please don't, I'd like to chat to get on the same page first.

dellaert:
OK,

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -190,12 +189,13 @@ discreteElimination(const HybridGaussianFactorGraph &factors,
 // If any GaussianFactorGraph in the decision tree contains a nullptr, convert
 // that leaf to an empty GaussianFactorGraph. Needed since the DecisionTree will
 // otherwise create a GFG with a single (null) factor.
+// TODO(dellaert): still a mystery to me why this is needed.
```
varunagrawal, line 192: @dellaert about this TODO, the explanation is given above. The `assembleGraphTree` method causes an empty `GaussianFactorGraph` to be created when there is a nullptr, hence we need to explicitly check and remove any factor graph with nullptrs (by setting it to nullptr).

PR #1390, opened 2023-01-17
Correction to docs and lambdas

varunagrawal:
- Updated the docstrings to properly reflect what's happening.
- Updated some lambda names since we're computing (log-)probabilities and not errors.
- Add a TODO about `reorder_relinearize`. It doesn't take pruning into account so relinearization in HybridNonlinearISAM is _very_ slow.


dellaert (approved):

dellaert:
Great. I thought you were  going to patch that branch I shared with you? It has the fix to the test.

varunagrawal:
> Great. I thought you were going to patch that branch I shared with you? It has the fix to the test.

I'll do that today.

PR #1392, opened 2023-01-18
verdant/typedef-optional-references

kartikarcot:
# Overview 

This PR is part of a push to upgrade GTSAM to C++17, and remove boost dependencies from GTSAM. In this PR we want to update our factor interfaces and the factor implementations that use arguments of the type boost::optional<A&> to raw pointers of the type A*. We wish to do that because the semantics of raw pointers are the same as that of an optional to a reference. This is in accordance to google style guide’s recommendation for function input parameters, [link here](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs).

# Details

## Retaining `evaluateError` interface to accept both pointers and matrices
This was achieved by some templating and conditional compilation in `NoiseModelFactorN` class. A side-effect of this is that when a user inherits NoiseModelFactorN and implements it, they have to do something like this-
`using Base::evaluateError;` for all of the evaluateError overloads to be visible for the end user.

## Replacing std::binds with lambdas
std::bind had a problem inferring which evaluateError overload to use through template deduction. I replaced it with lambdas instead.

##  `OptionalJacobian`
I have replaced `boost::optional<Matrix&>` in functions whose, input argument sizes are known, with OptionalJacobians. This enables us to use that function with both pointers and l-value references of matrices.

# TODO
- [x] Need to still replace `boost::optional<Matrix&>` in the following lines of code after discussing with Frank:
```
 gtsam/geometry/CameraSet.h
112:      boost::optional<Matrix&> E = boost::none) const {
141:      boost::optional<Matrix&> E = boost::none) const {

gtsam/base/tests/testOptionalJacobian.cpp
35:  boost::optional<Matrix&> optional(dynamic);

gtsam/slam/SmartFactorBase.h
207:      boost::optional<Matrix&> E = boost::none) const {
244:      boost::optional<Matrix&> E = boost::none) const {}

gtsam_unstable/slam/SmartStereoProjectionFactor.h
457:      boost::optional<Matrix&> E = boost::none) const override {
```
- [ ] Need a review on `Pose3Upright.cpp` 
- [ ]  Need a review on `timePose2.cpp` (Deleted a test which looked like a copy paste to me)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,9 +59,9 @@ endif()
 include(cmake/HandleGeneralOptions.cmake)   # CMake build options
 
 # Libraries:
-if (NOT GTSAM_NO_BOOST_CPP17)
+# if (NOT GTSAM_NO_BOOST_CPP17)
   include(cmake/HandleBoost.cmake)            # Boost
-endif()
+# endif()
   
 include(cmake/HandleCCache.cmake)           # ccache
 include(cmake/HandleCPack.cmake)            # CPack
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ endif()
 
 option(GTSAM_NO_BOOST_CPP17 "Require and use boost" ON)
 add_definitions(-Wno-deprecated-declarations)
+add_definitions(-ftemplate-backtrace-limit=0)
 
   set(CMAKE_CXX_STANDARD 17)
 if (GTSAM_NO_BOOST_CPP17)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ option(GTSAM_NO_BOOST_CPP17 "Require and use boost" ON)
 add_definitions(-Wno-deprecated-declarations)
 add_definitions(-ftemplate-backtrace-limit=0)
 
-  set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 17)
 if (GTSAM_NO_BOOST_CPP17)
   add_definitions(-DNO_BOOST_CPP17)
 endif()
@@ -99,7 +99,7 @@ add_subdirectory(timing)
 
 # Build gtsam_unstable
 if (GTSAM_BUILD_UNSTABLE)
-	# add_subdirectory(gtsam_unstable)
+	add_subdirectory(gtsam_unstable)
 endif()
 
 # This is the new wrapper
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,9 +9,9 @@ endif()
 option(GTSAM_NO_BOOST_CPP17 "Require and use boost" ON)
 add_definitions(-Wno-deprecated-declarations)
 
+  set(CMAKE_CXX_STANDARD 17)
 if (GTSAM_NO_BOOST_CPP17)
   add_definitions(-DNO_BOOST_CPP17)
-  set(CMAKE_CXX_STANDARD 17)
 endif()
 
 # Set the version number for the library
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,9 +6,6 @@ if(NOT DEFINED CMAKE_MACOSX_RPATH)
   set(CMAKE_MACOSX_RPATH 0)
 endif()
 
-add_definitions(-Wno-deprecated-declarations)
-add_definitions(-ftemplate-backtrace-limit=0)
-
 set(CMAKE_CXX_STANDARD 17)
 
 # Set the version number for the library
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,14 +6,10 @@ if(NOT DEFINED CMAKE_MACOSX_RPATH)
   set(CMAKE_MACOSX_RPATH 0)
 endif()
 
-option(GTSAM_NO_BOOST_CPP17 "Require and use boost" ON)
 add_definitions(-Wno-deprecated-declarations)
 add_definitions(-ftemplate-backtrace-limit=0)
 
 set(CMAKE_CXX_STANDARD 17)
-if (GTSAM_NO_BOOST_CPP17)
-  add_definitions(-DNO_BOOST_CPP17)
-endif()
 
 # Set the version number for the library
 set (GTSAM_VERSION_MAJOR 4)
@@ -60,10 +56,7 @@ endif()
 include(cmake/HandleGeneralOptions.cmake)   # CMake build options
 
 # Libraries:
-# if (NOT GTSAM_NO_BOOST_CPP17)
-  include(cmake/HandleBoost.cmake)            # Boost
-# endif()
-  
+include(cmake/HandleBoost.cmake)            # Boost
 include(cmake/HandleCCache.cmake)           # ccache
 include(cmake/HandleCPack.cmake)            # CPack
 include(cmake/HandleEigen.cmake)            # Eigen3
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,7 +98,7 @@ add_subdirectory(timing)
 
 # Build gtsam_unstable
 if (GTSAM_BUILD_UNSTABLE)
-    add_subdirectory(gtsam_unstable)
+	# add_subdirectory(gtsam_unstable)
 endif()
 
 # This is the new wrapper
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,12 @@ if(NOT DEFINED CMAKE_MACOSX_RPATH)
   set(CMAKE_MACOSX_RPATH 0)
 endif()
 
+option(GTSAM_NO_BOOST_CPP17 "Require and use boost" ON)
+
+if (GTSAM_NO_BOOST_CPP17)
+  set(CMAKE_CXX_STANDARD 17)
+endif()
+
 # Set the version number for the library
 set (GTSAM_VERSION_MAJOR 4)
 set (GTSAM_VERSION_MINOR 3)
@@ -51,7 +57,11 @@ endif()
 include(cmake/HandleGeneralOptions.cmake)   # CMake build options
 
 # Libraries:
-include(cmake/HandleBoost.cmake)            # Boost
+if (GTSAM_NO_BOOST_CPP17)
+  include(cmake/HandleBoost.cmake)            # Boost
+  add_definitions(-DNO_BOOST_CPP17)
+endif()
+  
 include(cmake/HandleCCache.cmake)           # ccache
 include(cmake/HandleCPack.cmake)            # CPack
 include(cmake/HandleEigen.cmake)            # Eigen3
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,8 +7,10 @@ if(NOT DEFINED CMAKE_MACOSX_RPATH)
 endif()
 
 option(GTSAM_NO_BOOST_CPP17 "Require and use boost" ON)
+add_definitions(-Wno-deprecated-declarations)
 
 if (GTSAM_NO_BOOST_CPP17)
+  add_definitions(-DNO_BOOST_CPP17)
   set(CMAKE_CXX_STANDARD 17)
 endif()
 
@@ -57,9 +59,8 @@ endif()
 include(cmake/HandleGeneralOptions.cmake)   # CMake build options
 
 # Libraries:
-if (GTSAM_NO_BOOST_CPP17)
+if (NOT GTSAM_NO_BOOST_CPP17)
   include(cmake/HandleBoost.cmake)            # Boost
-  add_definitions(-DNO_BOOST_CPP17)
 endif()
   
 include(cmake/HandleCCache.cmake)           # ccache
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,7 @@ add_subdirectory(timing)
 
 # Build gtsam_unstable
 if (GTSAM_BUILD_UNSTABLE)
-	add_subdirectory(gtsam_unstable)
+  add_subdirectory(gtsam_unstable)
 endif()
 
 # This is the new wrapper
--- a/doc/Code/LocalizationFactor.cpp
+++ b/doc/Code/LocalizationFactor.cpp
@@ -3,6 +3,7 @@ class UnaryFactor: public NoiseModelFactor1<Pose2> {
   
 public:
   using gtsam::NoiseModelFactor1<Pose2>::evaluateError;
+
   UnaryFactor(Key j, double x, double y, const SharedNoiseModel& model):
     NoiseModelFactor1<Pose2>(model, j), mx_(x), my_(y) {}
 
--- a/doc/Code/LocalizationFactor.cpp
+++ b/doc/Code/LocalizationFactor.cpp
@@ -2,11 +2,11 @@ class UnaryFactor: public NoiseModelFactor1<Pose2> {
   double mx_, my_; ///< X and Y measurements
   
 public:
+  using gtsam::NoiseModelFactor1<Pose2>::evaluateError;
   UnaryFactor(Key j, double x, double y, const SharedNoiseModel& model):
     NoiseModelFactor1<Pose2>(model, j), mx_(x), my_(y) {}
 
-  Vector evaluateError(const Pose2& q, 
-      OptionalMatrixType H = OptionalNone) const override {
+  Vector evaluateError(const Pose2& q, OptionalMatrixType H) const override {
     const Rot2& R = q.rotation();
     if (H) (*H) = (gtsam::Matrix(2, 3) <<
             R.c(), -R.s(), 0.0,
--- a/doc/Code/LocalizationFactor.cpp
+++ b/doc/Code/LocalizationFactor.cpp
@@ -6,7 +6,7 @@ public:
     NoiseModelFactor1<Pose2>(model, j), mx_(x), my_(y) {}
 
   Vector evaluateError(const Pose2& q, 
-      boost::optional<Matrix&> H = boost::none) const override {
+      OptionalMatrixType H = OptionalNone) const override {
     const Rot2& R = q.rotation();
     if (H) (*H) = (gtsam::Matrix(2, 3) <<
             R.c(), -R.s(), 0.0,
--- a/doc/Code/LocalizationFactor.cpp
+++ b/doc/Code/LocalizationFactor.cpp
@@ -2,6 +2,8 @@ class UnaryFactor: public NoiseModelFactor1<Pose2> {
   double mx_, my_; ///< X and Y measurements
   
 public:
+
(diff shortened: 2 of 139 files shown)


dellaert:
Seems like python tests are failing?

jlblancoc:
I guess you might have already discussed this in person and discarded it but just in case:

    using optional_ref_t=std::optional<std::reference_wrapper<T>>;

Would be pure C++, avoid null pointers, have a clear semantics, and probably be equally efficient (not tested yet in compiler explorer, though...). 
I think Google's style is fine, but it was defined before C++17

CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ if(NOT DEFINED CMAKE_MACOSX_RPATH)
   set(CMAKE_MACOSX_RPATH 0)
 endif()
 
+set(CMAKE_CXX_STANDARD 17)
```
dellaert, line 9: @jlblancoc I think this the right way and place to do so, right?
@kartikarcot I will probably do this in a PR before this one, so this diff will disappear from this PR.

kartikarcot:
@jlblancoc we did consider it but decided to defer to google style guidelines about optional input output parameters
> Generally, use std::optional to represent optional by-value inputs, and use a const pointer when the non-optional form would have used a reference. Use non-const pointers to represent optional outputs and optional input/output parameters.

[Google Style Guide](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs)

CMakeLists.txt
kartikarcot: Ok will do that

dellaert: No, you don’t have to remove it. It will be in both PRs

dellaert: Will just disappear from diff the moment i merge the other one

python/gtsam/tests/test_DiscreteBayesNet.py
```diff
@@ -130,7 +130,9 @@ def test_fragment(self):
 
         # Now sample from fragment:
         values = fragment.sample(given)
-        self.assertEqual(len(values), 5)
+        # TODO(kartikarcot): Resolve the len function issue. Probably
```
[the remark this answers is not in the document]
kartikarcot: Pasting the stacktrace here for visibility
```
ERROR: test_fragment (test_DiscreteBayesNet.TestDiscreteBayesNet)
Test evaluate/sampling/optimizing for Asia fragment.
  File "/home/runner/work/gtsam/gtsam/python/gtsam/tests/test_DiscreteBayesNet.py", line 133, in test_fragment
    self.assertEqual(len(values), 5)
TypeError: __len__(): incompatible function arguments. The following argument types are supported:
    1. (self: Dict[int, int]) -> int

Invoked with: DiscreteValues{0: 0, 3: 0, 4: 0, 5: 0, 6: 0}

```

CMakeLists.txt
ProfFan: `target_compile_features(${TARGET_NAME} PRIVATE cxx_std_17)` is the right way to do it I think.

kartikarcot: Got it

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -29,11 +29,35 @@
...
+ * Had to use the static_cast of a nullptr, because the compiler is not able to
+ * deduce the type of the nullptr when expanding the evaluateError templates.
+ */
+#define OptionalNone static_cast<Matrix*>(nullptr)
+
+/** This typedef will be used everywhere boost::optional<Matrix&> reference was used
+ * previously. This is used to indicate that the Jacobian is optional. In the future
+ * we will change this to OptionalJacobian
+ */
+using OptionalMatrixType = Matrix*;
```
ProfFan, line 53: Why raw pointers? Isn't `std::optional<std::reference_wrapper<T>>` working?

@jlblancoc

ProfFan (changes requested):
I don't think it's wise to replace these with raw pointers, as raw pointers do not encode lifetime and ownership information.

ProfFan (approved):
Change to approval in case you need to unblock fast.

CMakeLists.txt
dellaert: Not sure - if there is a special CMAKE flag, will it not do the above automatically (and tailored to different compilers, if needed) ?

python/gtsam/tests/test_DiscreteBayesNet.py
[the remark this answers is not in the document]
dellaert: @ProfFan any ideas?

CMakeLists.txt
jlblancoc: > @jlblancoc I think this the right way and place to do so, right?

Yes. But I agree with @ProfFan in that I would prefer to set it as a compile feature, but I think it must be public, not private (otherwise, users not using c++17 will for sure have build errors). 
So, it would be: 

    target_compile_features(${TARGET_NAME} PUBLIC cxx_std_17)

for the target `gtsam`, and perhaps `gtsam_unstable` although the latter would be redundant since it inherits the public flags from `gtsam`.

PS: From my own testing, I found [this additional command](https://github.com/MOLAorg/mola-common/blob/5b2001fb9b0503a3e38508cf0c5839c17047dfdb/mola_cmake_functions.cmake#L117-L123) was also needed to enable C++17 on MSVC, but give it a try first with `target_compile_features()` alone since it probably was a buggy cmake/msvc version by ~2020.

gtsam/nonlinear/NonlinearFactor.h
jlblancoc: It's up to you all, library API designers, both policies may make sense, despite my personal preference for not using pointers.

PS: Ownership and lifetime is neither ensured with the `std::optional<...>` type, btw... The only way to enforce and clarify both are (AFIK) smart pointers, so, dynamic memory, killing performance... I guess you must rely on well-designed code, validate with sanitizers, etc. Just as done so far :-)

CMakeLists.txt
dellaert: What is $TARGET_NAME supposed to be here? `gtsam` ?

ProfFan: I think so

doc/Code/LocalizationFactor.cpp
```diff
@@ -2,11 +2,14 @@ class UnaryFactor: public NoiseModelFactor1<Pose2> {
   double mx_, my_; ///< X and Y measurements
   
 public:
+
+  // Provide access to the Matrix& version of evaluateError:
+  using gtsam::NoiseModelFactor1<Pose2>::evaluateError;
```
ProfFan, line 7: Why is this necessary?

CMakeLists.txt
jlblancoc: Yes.

jlblancoc: Ah, of course then that line wouldn't be at the root cmake file anymore, but after defining the gtsam library target.

python/gtsam/tests/test_DiscreteBayesNet.py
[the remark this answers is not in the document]
ProfFan: Probably an ODR violation. Need to check if there includes of `<pybind11/stl.h>` and whether `std::map<Key, size_t>` got accidentally wrapped automatically.

gtsam/nonlinear/NonlinearFactor.h
```diff
@@ -587,19 +660,31 @@ class NoiseModelFactorN
...
+   * e.g. `const Vector error = factor.evaluateError(pose, point, &Hpose);`
+   */
+  template <typename... OptionalJacArgs, typename = IndexIsValid<sizeof...(OptionalJacArgs) + 1>>
+  inline AreAllMatrixPtrs<Vector, OptionalJacArgs...> evaluateError(const ValueTypes&... x,
+                                                                    OptionalJacArgs&&... H) const {
+    // If they are pointer version, ensure to cast them all to be Matrix* types
+    // This will ensure any arguments inferred as std::nonetype_t are cast to (Matrix*) nullptr
+    // This guides the compiler to the correct overload which is the one that takes pointers
+    return evaluateError(x..., 
+        std::forward<OptionalJacArgs>(H)..., static_cast<OptionalMatrixType>(OptionalNone));
```
gchenfc, line 687: May or may not be worth doing, but I think the pointer versions of these can be replaced with default args.  i.e.

```c++
template <typename... OptionalJacArgs, typename = IndexIsValid<sizeof...(OptionalJacArgs) + 1>>
virtual Vector evaluateError(const ValueTypes&... x, OptionalMatrixTypeT H1 = 0, OptionalJacArgs*... Hs) = 0;
```
or something along those lines (untested).  Or maybe I'm wrong, because otherwise I'm not sure why we did these weird overloading tricks previously.  Maybe I didn't realize that if you provide one default arg, the remaining template params automatically get default args.

gtsam/nonlinear/PriorFactor.h
```diff
@@ -32,6 +32,10 @@ namespace gtsam {
   public:
     typedef VALUE T;
 
+    // Provide access to the Matrix& version of evaluateError:
+    using NoiseModelFactor1<VALUE>::evaluateError;
```
gchenfc, line 36: Just a comment that I don't think this is so bad, because usually this won't be called directly anyway (but rather from a `NonlinearFactorGraph`) except in unit tests, and even then it's usually either (a) without the output args or (b) using factorTesting macros, so I don't think it will actually affect too much old code.

gtsam/nonlinear/PriorFactor.h
```diff
@@ -91,7 +95,7 @@ namespace gtsam {
     /** implement functions needed to derive from Factor */
 
     /** vector of errors */
-    Vector evaluateError(const T& x, boost::optional<Matrix&> H = boost::none) const override {
+    Vector evaluateError(const T& x, OptionalMatrixType H) const override {
```
gchenfc, line 98: Just for reference, is it fair to say that this (and a lot of other instances) could be migrated by doing global find-replace: `boost::optional<Matrix&>` -> `OptionalMatrixType` ?

gtsam/slam/EssentialMatrixConstraint.cpp
```diff
@@ -43,16 +43,20 @@ bool EssentialMatrixConstraint::equals(const NonlinearFactor& expected,
...
   // convert to EssentialMatrix
   Matrix D_hx_1P2;
-  EssentialMatrix hx = EssentialMatrix::FromPose3(_1P2_,
-      (Hp1 || Hp2) ? boost::optional<Matrix&>(D_hx_1P2) : boost::none);
+  EssentialMatrix hx;
+  if (Hp1 || Hp2) {
+    hx = EssentialMatrix::FromPose3(_1P2_, D_hx_1P2);
+  } else {
+      hx = EssentialMatrix::FromPose3(_1P2_, OptionalNone);
+  }
```
gchenfc, line 59: Why did this have to change?  Could this stay as
```c++
EssentialMatrix hx = EssentialMatrix::FromPose3(_1P2_, (Hp1 || Hp2) ? &D_hx_1P2 : 0);
```
?

gtsam/slam/EssentialMatrixFactor.h
```diff
@@ -284,18 +294,21 @@ class EssentialMatrixFactor3 : public EssentialMatrixFactor2 {
...
     } else {
       // Version with derivatives
       Matrix D_e_cameraE, D_cameraE_E;  // 2*5, 5*5
       EssentialMatrix cameraE = E.rotate(cRb_, D_cameraE_E);
-      Vector e = Base::evaluateError(cameraE, d, D_e_cameraE, Dd);
+	  // Had to do this since the only overloaded function EssentialMatrixFactor2
+	  // uses the type OptionalMatrixType. Which would be a pointer when we are
+	  // not using boost. There is no way to redirect that call to the top (NoiseModelFactorN)
+	  // dereference it and bring it back to the Base (EssentialMatrixFactor2)
+      Vector e = Base::evaluateError(cameraE, d, &D_e_cameraE, Dd);
```
gchenfc, line 311: I'm not sure I understand this comment - I think there might be a typo ("since the only overloaded function ... uses the type") and some missing comma's, and I'm not clear what "this" and "that" is referring to.  Sorry, not trying to be aggressive, just trying to give constructive feedback on how I interpreted when reading. :)

I am indeed curious because I would have expected the entire if-else block to be replaced by something like

```c++
Vector evaluateError( /* args */) {
  Matrix D_e_cameraE, D_cameraE_E; // 2*5, 5*5

  EssentialMatrix cameraE = E.rotate(cRb_, DE ? &D_cameraE_E : 0);
  Vector e = Base::evaluateError(cameraE, d, DE ? &D_e_cameraE : 0, Dd);
  if (DE) *DE = D_e_cameraE * D_cameraE_E;
  return e;
}
```

gtsam_unstable/slam/tests/testTSAMFactors.cpp
```diff
@@ -121,18 +129,26 @@ TEST( OdometryFactorBase, all ) {
...
   // Use numerical derivatives to calculate the Jacobians
   Matrix H1Expected, H2Expected, H3Expected, H4Expected;
+  // using lambdas to replace bind
   H1Expected = numericalDerivative11<Vector3, Pose2>(
-      std::bind(&OdometryFactorBase::evaluateError, &factor, std::placeholders::_1, pose1, base2,
-          pose2, boost::none, boost::none, boost::none, boost::none), base1);
+      [&factor, &pose1, &pose2, &base2](const Pose2& pose_arg) {
+        return factor.evaluateError(pose_arg, pose1, base2, pose2);
+      },
+      base1);
```
gchenfc, line 137: FYI for future reference I think it works if we do `boost::none` -> `(OptionalMatrixType*) 0` (or static_cast), but lambda looks better anyway :)

gchenfc (commented):
Didn't read through entire thing in detail yet, but aside from the mentioned backward compatibility stuff, lgtm :)

gchenfc:
> > using optional_ref_t=std::optional<std::reference_wrapper>;
> 
> That's interesting, I forgot about this possibility, but I actually like/prefer the `&` at the call site, to make it clear it's an output argument. We have a plan to switch to OptionalJacobian, to support fixed-size Jacobians, and will see whether we can use it there.

Is there a reason this PR can't already start using `OptionalJacobian` instead of `Matrix*` ?  Mainly for google style guide to use pointers for output arguments?

doc/Code/LocalizationFactor.cpp
kartikarcot: This is to help compiler find the right overload of `evaluateError`. Since a function with the same name is overridden in the derived class (the pointer version), the compiler function name lookup logic just stops in the derived class and complains that it could not find the right overload. By declaring that we use other signatures from the parent class, we are able to access the matrix reference style of `evaluateError`, defined in the parent class, as well.

gtsam/nonlinear/NonlinearFactor.h
kartikarcot: Any remaining optional arguments are added by the line 686, recursively, right?
```
return evaluateError(x..., 
        std::forward<OptionalJacArgs>(H)..., static_cast<OptionalMatrixType>(OptionalNone));
```

gtsam/nonlinear/PriorFactor.h
kartikarcot: Yea that's right as long as the the operator overloads for `*` and `->` have the same semantics, if I got that right

gtsam/slam/EssentialMatrixConstraint.cpp
kartikarcot: Yes it can, I think I changed it because I found it confusing. I will revert it.

gtsam/slam/EssentialMatrixFactor.h
kartikarcot: I have made the comment more clearer. Also by rephrasing it like this would mean cameraE will be computed even if DE is `nullptr`

gtsam_unstable/slam/tests/testTSAMFactors.cpp
kartikarcot: Ah I see

gtsam/slam/EssentialMatrixConstraint.cpp
kartikarcot: Ah I changed it because `OptionalNone` and `D_hx_1P2` are of different types.

dellaert: Well, could you instead use nullptr?

dellaert:
> > > using optional_ref_t=std::optionalstd::reference_wrapper;
> > 
> > 
> > That's interesting, I forgot about this possibility, but I actually like/prefer the `&` at the call site, to make it clear it's an output argument. We have a plan to switch to OptionalJacobian, to support fixed-size Jacobians, and will see whether we can use it there.
> 
> Is there a reason this PR can't already start using `OptionalJacobian` instead of `Matrix*` ? Mainly for google style guide to use pointers for output arguments?

I might migrate to OptionalJacobian in a future PR. This one is the first cut at getting rid of boost::optional.

gtsam/slam/EssentialMatrixConstraint.cpp
kartikarcot: No nullptr and D_hx_1P2 (matrix) are different types too. They can’t be used with a ternary operator. OptionalNone typecasts to nullptr

dellaert: Check the original code, it’s a pointer. This idiom is used all over GTSAM so I’m a bit worried it’s broken.

dellaert: I'll resolve - might fix in follow up PR

gtsam/nonlinear/NonlinearFactor.h
gchenfc: Ya sorry I wasn't very clear, I meant to say that I think it might be more readable with the default arguments rather than the lots-of-templates version.  But nbd :)

gtsam/navigation/tests/testAttitudeFactor.cpp
```diff
@@ -47,11 +49,11 @@ TEST( Rot3AttitudeFactor, Constructor ) {
   Rot3 nRb;
   EXPECT(assert_equal((Vector) Z_2x1,factor.evaluateError(nRb),1e-5));
 
+  auto err_fn = [&factor](const Rot3& r){
+    return factor.evaluateError(r, OptionalNone);
```
varunagrawal, line 53: This seems messy just to be a bit more aligned to the Google style guide...

gchenfc: I suspect this was just an oversight and that the old syntax could be used (and not a change driven by google style).  Although, personally, I actually find the lambda syntax easier to read than std::bind, but that's personal preference.

https://github.com/borglab/gtsam/pull/1392#discussion_r1083141328

PR #1394, opened 2023-01-19
Switch pruning test to probabilities.

dellaert:
@varunagrawal because this was merged after the split, we also need to merge in release/4.2


dellaert:
Actually, does not need a review - will just merge.

PR #1395, opened 2023-01-19
Switch to version 4.3a0 and to c++17.

dellaert:
- patched CMakeLists.txt to Switch to version 4.3a0 and to c++17.
- added docs to that effect.


CMakeLists.txt
```diff
@@ -6,11 +6,13 @@ if(NOT DEFINED CMAKE_MACOSX_RPATH)
   set(CMAKE_MACOSX_RPATH 0)
 endif()
 
+set(CMAKE_CXX_STANDARD 17)
```
ProfFan, line 9: Commented in another PR

jlblancoc: cc: https://github.com/borglab/gtsam/pull/1392#discussion_r1081969554

cmake/GtsamBuildTypes.cmake
```diff
@@ -138,16 +138,22 @@ else()
...
 
-# Enable C++11:
+# Enable C++17:
 if (NOT CMAKE_VERSION VERSION_LESS 3.8)
-    set(GTSAM_COMPILE_FEATURES_PUBLIC "cxx_std_11" CACHE STRING "CMake compile features property for all gtsam targets.")
+    set(GTSAM_COMPILE_FEATURES_PUBLIC "cxx_std_17" CACHE STRING "CMake compile features property for all gtsam targets.")
     # See: https://cmake.org/cmake/help/latest/prop_tgt/CXX_EXTENSIONS.html
-    # This is to enable -std=c++11 instead of -std=g++11
+    # TODO(dellaert): is following line still needed or was that only for c++11?
     set(CMAKE_CXX_EXTENSIONS OFF)
```
jlblancoc, line 146: Yes, it's better to keep it to make compilation not to depend on compiler-specific features... so it's less likely that it builds ok on gcc and not on clang or the other way around.

cmake/GtsamBuildTypes.cmake
```diff
@@ -138,16 +138,22 @@ else()
...
+      # NOTE(jlblanco): seems to be required in addition to the cxx_std_17 above?
+      # TODO(dellaert): is this the right syntax below?
+      list_append_cache(GTSAM_COMPILE_OPTIONS_PUBLIC /std:c++latest)
+    endif()
 else()
   # Old cmake versions:
   if (NOT MSVC)
-    list_append_cache(GTSAM_COMPILE_OPTIONS_PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-std=c++11>)
+    # TODO(dellaert): I just changed 11 to 17 below, hopefully that works
+    list_append_cache(GTSAM_COMPILE_OPTIONS_PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-std=c++17>)
```
jlblancoc, line 156: That will only be active on old distros with old cmake versions, but in principle yes, it's ok. 
Anyhow, the whole test matrix passes, so it looks good enough for the most common OS/compiler versions! 👍

jlblancoc (approved):

PR #1396, opened 2023-01-20
Eigen, cuSparse and SuiteSparse solvers

ProfFan:
Replacing #111 for clean diff

This PR adds a common interface for other kinds of linear solvers. Not perfect but is indeed useful.

tested with yetong and it seems to be accelerating stuff for a bit


dellaert:
Could you please edit PR comment to say what this is about, with some context?

dellaert:
Also, CI fails :-)

dellaert:
Can you merge in develop and resolve conflicts?

dellaert:
I propose we close in favor of new CUDA effort. we can always re-open.

Issue #1400, opened 2023-01-21
Build failing: ‘class gtsam::Values’ has no member named ‘filter’


husseinalijaafar:
## Description

Trying to build GTSAM with the depreciated flag below, but the build is failing. Both myself and a colleague are having similar issues.  
```
[ 84%] Building CXX object examples/CMakeFiles/StereoVOExample_large.dir/StereoVOExample_large.cpp.o
/home/hussein/gbp/gtsam/examples/StereoVOExample_large.cpp: In function ‘int main(int, char**)’:
/home/hussein/gbp/gtsam/examples/StereoVOExample_large.cpp:116:31: error: ‘class gtsam::Values’ has no member named ‘filter’
  116 |   Values pose_values = result.filter<Pose3>();
      |                               ^~~~~~
/home/hussein/gbp/gtsam/examples/StereoVOExample_large.cpp:116:43: error: expected primary-expression before ‘>’ token
  116 |   Values pose_values = result.filter<Pose3>();
      |                                           ^
/home/hussein/gbp/gtsam/examples/StereoVOExample_large.cpp:116:45: error: expected primary-expression before ‘)’ token
  116 |   Values pose_values = result.filter<Pose3>();
      |                                             ^
```
## Steps to reproduce

```
git clone https://github.com/borglab/gtsam.git
cd gtsam
mkdir build && cd build
cmake -DGTSAM_ALLOW_DEPRECATED_SINCE_V42:=OFF .. 
sudo make install
```

## Expected behavior

Build successfuly

## Environment

Xubuntu 20.04, cmake version 3.24.0-rc3

## Additional information

I had previously successfully built gtsam a few days ago. This may be due recent changes.

dellaert:
Thanks! Fixed in #1399 in develop which should be merged once reviewed by @jlblancoc or @varunagrawal , but maybe you are trying to compile release/4.2 ? I should make the changes there, too, PR coming!

husseinalijaafar:
Thank you for the prompt responses Dr!

PR #1402, opened 2023-01-21
Fix examples that were using `filter`

dellaert:
Cherry-picked from c71d07bb 
Fixes issue #1400

### DIFF
--- a/examples/StereoVOExample_large.cpp
+++ b/examples/StereoVOExample_large.cpp
@@ -27,6 +27,7 @@
 #include <gtsam/geometry/Pose3.h>
 #include <gtsam/geometry/Cal3_S2Stereo.h>
 #include <gtsam/nonlinear/Values.h>
+#include <gtsam/nonlinear/utilities.h>
 #include <gtsam/nonlinear/NonlinearEquality.h>
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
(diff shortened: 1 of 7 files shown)


varunagrawal (approved):
LGTM

gchenfc (approved):
This might have been obvious to others, but I just realized that the reason #1398 (and #1397) passed CI check (special cases CI - deprecated) is because CI is never compiling examples.  Even though the ci scripts have a flag to compile examples (`-b` ):
https://github.com/borglab/gtsam/blob/6787bbe0488b1224e8ac58b0f0e16e7316994708/.github/scripts/unix.sh#L134-L141

https://github.com/borglab/gtsam/blob/6787bbe0488b1224e8ac58b0f0e16e7316994708/.github/scripts/unix.sh#L91-L94

https://github.com/borglab/gtsam/blob/6787bbe0488b1224e8ac58b0f0e16e7316994708/.github/scripts/unix.sh#L112-L115

---
all the CI's are actually only using `-t`:
https://github.com/borglab/gtsam/blob/6787bbe0488b1224e8ac58b0f0e16e7316994708/.github/workflows/build-special.yml#L141-L143

Perhaps we should consider compiling the examples as well in CI, at least in e.g. release branches.

dellaert:
Agreed !

PR #1403, opened 2023-01-21
Verdant/replace boost optional vals

kartikarcot:
> Reopening this PR since feature/remove_deprecated_code branch is now merged

This PR will completely remove boost::optional usage from gtsam.

All boost::optional usage is not replaced with std::optional
Look at std_optional_serialization.h which is a mini implementation for serializing std::optional
The testStdOptionalSerialization.cpp has some simple tests for the above feature. Let me know if I should
test on more cases.
All boost::none are replaced with {}
In OptionalJacobian class removed the constructors that accepted boost::optional<Matrix&>
Another note-worthy diff is in EliminateableFactorGraph.h where a boost::optional<const VariableIndex&> was used.
    /// Typedef for an optional variable index as an argument to elimination functions
   -  typedef boost::optional<const VariableIndex&> OptionalVariableIndex;
   + typedef const VariableIndex* OptionalVariableIndex;


dellaert:
Messing with the basics often trips up CI, esp. gcc. Might take a few iterations.

dellaert (approved):
Already reviewed in other PR

gtsam/base/std_optional_serialization.h
```diff
@@ -0,0 +1,95 @@
+// Functionality to serialize std::optional<T> to boost::archive
+// Following this:
+// PR: https://github.com/boostorg/serialization/pull/148/files#
```
gchenfc, line 3: I think this should be boostorg/serialization#163, not 148 (which is for `std::variant` and a bit harder to read).

Also, I think the [boost software license](https://www.boost.org/LICENSE_1_0.txt) is supposed to be linked.

Otherwise, LGTM! :)

PR #1405, opened 2023-01-22
Upgraded to Eigen 3.4.0

dellaert:
This is a straight up replacement of Eigen. No changes to default version to use, no using any special features, although there are plenty: https://eigen.tuxfamily.org/index.php?title=3.4

Any of these changes I'll make in a follow-up PR after discussion.


dellaert:
@ProfFan no idea why this fails. cmake and build work perfectly fine on my Mac with clang, and even Mac CIs seem to fail here.

ProfFan:
Indeed you are missing one file

Did you use the release tarball or copied the folder from the Git repo? You need to use https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.zip

in order to have a portable Eigen release, the Git tag one is broken

dellaert:
> Indeed you are missing one file
> 
> Did you use the release tarball or copied the folder from the Git repo? You need to use [gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.zip](https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.zip)
> 
> in order to have a portable Eigen release, the Git tag one is broken

Actually, it seems I *did* use the one you pointed to.

dellaert:
I figured it out. I missed many files because of an issue with .gitignore on Mac :-/ Let's see what CI says.

ProfFan (approved):

PR #1407, opened 2023-01-22
Deprecated boost transform_iterators

dellaert:
- Deprecated boost iterators in `Values.*`, fixed all downstream issues.
- Also moved some Eigen methods to cpp to cut down on warnings (could be a separate PR but don't want to block CI)

### DIFF
--- a/gtsam/nonlinear/Values.cpp
+++ b/gtsam/nonlinear/Values.cpp
@@ -25,8 +25,6 @@
 #include <gtsam/nonlinear/Values.h>
 #include <gtsam/linear/VectorValues.h>
 
-#include <boost/iterator/transform_iterator.hpp>
-
 #include <list>
 #include <memory>
 #include <sstream>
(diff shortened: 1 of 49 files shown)


ProfFan (approved):

gtsam/nonlinear/Values.cpp
```diff
@@ -102,17 +100,44 @@ namespace gtsam {
...
+    for (key_delta = delta.begin(); key_value != values_.end();
+         ++key_value, ++key_delta) {
+      assert(key_value->first == key_delta->first);
+#endif
+      Key var = key_value->first;
+      assert(static_cast<size_t>(delta[var].size()) == key_value->second->dim());
+      assert(delta[var].allFinite());
+      if (mask.exists(var)) {
+        Value* retracted = key_value->second->retract_(delta[var]);
+        // TODO(dellaert): can we use std::move here?
```
ProfFan, line 122: `GenericValue` does not have move constructor

PR #1406, opened 2023-01-22
Verdant/smart ptrs

kartikarcot:
This PR will replace all boost smart pointer usage with STL smart pointers.
* It is a simple find and replace of all boost smart ptr usage with stl
* Things that changed were:
   - `dynamic_pointer_cast`
   - `shared_ptr`
   - `weak_ptr`
   - `make_shared`
   - `allocate_shared`

PR #1408, opened 2023-01-23
Removed deprecated boost transform_iterators

dellaert:
Quick review, same PR as #1407 , but in 4.3 we completely remove the code.


dellaert:
I canceled CI as I anticipate having to resolve conflicts with #1406

ProfFan (approved):

PR #1417, opened 2023-01-26
Iterator for 4.2

dellaert:
Added a const_iterator back in.

PR #1418, opened 2023-01-26
Added a const_iterator back in

dellaert:
Following #1417, I "undeprecated" begin() and end(), as well as find, upper bound, and lower_bound.


varunagrawal (approved):
LGTM

PR #1421, opened 2023-01-27
Update ROS package.xml format to v3

jlblancoc:
Update package.xml to format v3. 
Also, update the version to 4.1.1 since that seems to be the last stable release... is this correct? Anyhow, that `<version>` line should be updated just before the next release so its contents remain consistent when tagging the repository.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>gtsam</name>
-  <version>4.1.0</version>
+  <version>4.1.1</version>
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Thanks @jlblancoc . Yes, 4.1.1 is still the latest official release, although I am hoping that will change. When you see the PR for that and you don't see the version chnage, please remind me :-)

dellaert:
Before I merge, though: should this not be a patch on 4.1.1 in master? And in release/4.2 it should be 4.2.0, and in develop 4.3.0?

jlblancoc:
Yes, it makes sense. Will open the alternative PRs for each branch. 
Note that the tags will be pointing to the versions w/o the corrected package.xml file... not a big deal since gtsam is not listed yet on rosdistro.

dellaert:
Let's then punt on master - soon it will have the latest 4.3a0 tag.

PR #1423, opened 2023-01-28
Fix (issue #1336) dangling pointer access violation.

oicchris:
Fixes (https://github.com/borglab/gtsam/issues/1336) a dangling pointer access violation when Expression<>::valueAndJacobianMap() is called using a MSVC compiler. Deallocation of the pointer occurs before internal::ExecutionTrace trace's dtor is called. Inside the dtor the dangling pointer is dereferenced.

Cherry pick from the `develop` branch, commit: a41b4e02e5458a4ceda4aece73252f9278c95dca

### DIFF
--- a/gtsam/nonlinear/Expression-inl.h
+++ b/gtsam/nonlinear/Expression-inl.h
@@ -19,6 +19,13 @@
 
 #pragma once
 
+// The MSVC compiler workaround for the unsupported variable length array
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Awesome !!!!

dellaert:
CI already fully succeeded on develop so merged here as well. If an issue arises we'll fix it in post :-)

PR #1426, opened 2023-01-28
Patch ros package (branch: master)

jlblancoc:
Follow up from #1421


dellaert:
Let's not do this with the wrong version. Soon it will have the latest 4.3a0 tag that will come from develop. The only PRs into master are supposed to be the ones coming from release branches anyway,

PR #1428, opened 2023-01-28
Update ROS package.xml format to v3 and update gtsam version (develop branch)

jlblancoc:
Follow up of #1421

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>gtsam</name>
-  <version>4.1.0</version>
+  <version>4.3.0</version>
   <description>gtsam</description>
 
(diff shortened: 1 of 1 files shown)


package.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>gtsam</name>
-  <version>4.1.0</version>
+  <version>4.3.0</version>
```
dellaert, line 5: Should this be 4.3a0 ? That and 4.2  will be the tags going to master soon...

dellaert (approved):
Thanks ! One Q about versioning, as 4.3.0 will not be stable for a while. 4.3a0 is the pre-release that will be merged into master once boost is gone.

dellaert:
PS, don't push to the branch until we agree on right version above - I am trying to fix something for gtbook and I'm blocked on CI cycles.

dellaert:
I ran required CI and merged. If the version is wrong let me know and I can sneak it into my next PR.

PR #1437, opened 2023-02-04
c++17 style pairs everywhere

dellaert:
This started out by rescuing some code in an experimental PR, and now pretty much uses c++17 pattern matching and {} initialization wherever we can. No issues expected.

### DIFF
--- a/gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
+++ b/gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
@@ -270,9 +270,7 @@ TEST(HybridGaussianFactorGraph, eliminateFullMultifrontalTwoClique) {
   auto ordering_full =
       Ordering::ColamdConstrainedLast(hfg, {M(0), M(1), M(2), M(3)});
 
-  HybridBayesTree::shared_ptr hbt;
-  HybridGaussianFactorGraph::shared_ptr remaining;
-  std::tie(hbt, remaining) = hfg.eliminatePartialMultifrontal(ordering_full);
+  const auto [hbt, remaining] = hfg.eliminatePartialMultifrontal(ordering_full);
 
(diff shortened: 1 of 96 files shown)


gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
```diff
@@ -327,10 +325,9 @@ TEST(HybridGaussianFactorGraph, Switching) {
     std::transform(naturalX.begin(), naturalX.end(), std::back_inserter(ordX),
                    [](int x) { return X(x); });
 
-    KeyVector ndX;
-    std::vector<int> lvls;
-    std::tie(ndX, lvls) = makeBinaryOrdering(ordX);
+    auto [ndX, lvls] = makeBinaryOrdering(ordX);
     std::copy(ndX.begin(), ndX.end(), std::back_inserter(ordering));
+    // TODO(dellaert): this has no effect!
```
dellaert, line 330: @ProfFan I think this might be you, no idea what you were trying to achieve here as `lvls` is not visible below.

gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
```diff
@@ -388,10 +381,9 @@ TEST(HybridGaussianFactorGraph, SwitchingISAM) {
...
```
dellaert, line 386: @ProfFan same

dellaert:
@varunagrawal I checked CI and it is OOM, other TBB runs work...

varunagrawal (approved):
Not a fan of replacing all the type information with `auto` but otherwise looks good.

dellaert:
> Not a fan of replacing all the type information with `auto` but otherwise looks good.

Yeah, I'm on the fence myself. I held off in the examples for this reason, except in for loops where the win is just so big.

PR #1438, opened 2023-02-04
Getting rid of misc. boost usage

dellaert:
This is really a number of cherry-picks from [this branch](https://github.com/kartikarcot/gtsam/tree/verdant/boost-misc) by @kartikarcot, with some fixes and additional commits by me. It gets rid most of the "easy" remaining boost usage. Other PRs are coming which require a closer review. 

@ProfFan seeing whether you can do a quick review on this one?

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -55,8 +55,8 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          # LLVM (clang) 9 is not in Bionic's repositories so we add the official LLVM repository.
-          if [ "${{ matrix.compiler }}" = "clang" ] && [ "${{ matrix.version }}" = "9" ]; then
+          # LLVM (clang) 9/14 is not in Bionic's repositories so we add the official LLVM repository.
+          if [ "${{ matrix.compiler }}" = "clang" ]; then
             # (ipv4|ha).pool.sks-keyservers.net is the SKS GPG global keyserver pool
             # ipv4 avoids potential timeouts because of crappy IPv6 infrastructure
             # 15CF4D18AF4F7421 is the GPG key for the LLVM apt repository
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -51,7 +51,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Install Dependencies
         run: |
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -20,19 +20,15 @@ jobs:
         # Github Actions requires a single row to be added to the build matrix.
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
         name: [
-          ubuntu-20.04-gcc-7,
           ubuntu-20.04-gcc-9,
           ubuntu-20.04-clang-9,
+          ubuntu-22.04-gcc-11,
+          ubuntu-22.04-clang-14,
         ]
 
         build_type: [Debug, Release]
         build_unstable: [ON]
         include:
-          - name: ubuntu-20.04-gcc-7
-            os: ubuntu-20.04
-            compiler: gcc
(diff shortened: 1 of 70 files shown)


dellaert:
Windows fails with

> C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\include\type_traits(1003,19): error C2338: You've instantiated std::aligned_storage<Len, Align> with an extended alignment (in other words, Align > alignof(max_align_t)). Before VS 2017 15.8, the member "type" would non-conformingly have an alignment of only alignof(max_align_t). VS 2017 15.8 was fixed to handle this correctly, but the fix inherently changes layout and breaks binary compatibility (*only* for uses of aligned_storage with extended alignments). Please define either (1) _ENABLE_EXTENDED_ALIGNED_STORAGE to acknowledge that you understand this message and that you actually want a type with an extended alignment, or (2) _DISABLE_EXTENDED_ALIGNED_STORAGE to silence this message and get the old non-conforming behavior. [D:\a\gtsam\gtsam\build\gtsam\gtsam.vcxproj]


on this line in ExecutionTrace.h:
```c++
static const unsigned TraceAlignment = 32;
typedef std::aligned_storage<1, TraceAlignment>::type ExecutionTraceStorage;
```

I will just define `_ENABLE_EXTENDED_ALIGNED_STORAGE` unless a windows user (@mikesheffler ?) objects...

mikesheffler:
> I will just define _ENABLE_EXTENDED_ALIGNED_STORAGE unless a windows user (@mikesheffler ?) objects...

Seems reasonable to me. :)

dellaert:
@ProfFan and anyone, does it seem reasonable to stop support for GCC 7? Lots of c+=17 warnings and only experimental support for filesystem, which now throws linking errors… default compiler on Ubuntu 20 is gcc 9 I think.

dellaert:
@mikesheffler adding you as a reviewer to spread the load :-) should be fairly straightforward…

dellaert:
Dang @mikesheffler @ProfFan , setting the define does not help, and setting alignment to 16 (Eigen default anyway) does not help. At my wit's end with this...

ProfFan:
Yeah I see no problem with dropping GCC7. On the windows problem, I think adding a definition in cmake:
```
target_compile_definitions(gtsam PUBLIC _ENABLE_EXTENDED_ALIGNED_STORAGE)
```
should work

ProfFan (approved):

dellaert:
> ```
> target_compile_definitions(gtsam PUBLIC _ENABLE_EXTENDED_ALIGNED_STORAGE)
> ```

Let me try. Should I put it here in GtsamBuildTypes.cmake ?
```
if(MSVC)
  # Common to all configurations:
  list_append_cache(GTSAM_COMPILE_DEFINITIONS_PRIVATE
    WINDOWS_LEAN_AND_MEAN
    NOMINMAX
    )
```

ProfFan:
I think this should probably be in `GTSAM_COMPILE_DEFINITIONS_PUBLIC`.

dellaert:
> I think this should probably be in `GTSAM_COMPILE_DEFINITIONS_PUBLIC`.

Why, and where do I define those? (file, line ?)

dellaert:
> > I think this should probably be in `GTSAM_COMPILE_DEFINITIONS_PUBLIC`.
> 
> Why, and where do I define those? (file, line ?)

I think I see why, ChatGPT had a good explanation :-) Trying this:
```
  list_append_cache(GTSAM_COMPILE_DEFINITIONS_PUBLIC
    _ENABLE_EXTENDED_ALIGNED_STORAGE
  )
```

dellaert:
Windows passed !!!!! Thanks @ProfFan ! I will merge despite failing TBB as that actually slipped into develop. Will fix in upcoming PR.

dellaert:
Thanks again @kartikarcot !

mikesheffler:
I was a little slow in getting to this, but it looks like it already merged; that's great!

kartikarcot:
Yay! I’m glad we don’t support gcc7 anymore 😅

PR #1439, opened 2023-02-04
replace boost phoenix parser with SignatureParser

dellaert:
Thanks a lot @kartikarcot for this new parser that replaces the boost phoenix parser. 
I refactored a bit, I don;t like to use non-const references so using std::optional now.

@varunagrawal this is not as quick but Kartik added a bunch of tests and all hybrid tests work as well.

### DIFF
--- a/gtsam/discrete/Signature.cpp
+++ b/gtsam/discrete/Signature.cpp
@@ -16,55 +16,15 @@
  * @date Feb 27, 2011
  */
 
-#include <sstream>
-
+#include "gtsam/discrete/SignatureParser.h"
 #include "Signature.h"
 
-#include <boost/spirit/include/qi.hpp> // for parsing
-#include <boost/spirit/include/phoenix.hpp> // for qi::_val
+#include <sstream>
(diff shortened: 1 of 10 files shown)


gtsam/discrete/Signature.cpp
```diff
@@ -16,178 +16,129 @@
...
-    if (table_) {
-      const size_t nrStates = table_->at(0).size();
-      for (size_t j = 0; j < nrStates; j++) {
-        for (const Row& row : *table_) {
-          assert(row.size() == nrStates);
-          cpt.push_back(row[j]);
-        }
+using namespace std;
+
+ostream& operator<<(ostream& os, const Signature::Row& row) {
```
dellaert, line 30: Seems like diff is screwing up here. This file is almost identical expect for removing parser.

dellaert:
CI failing will be resolved soon in base branch. Note I will wait and merge directly in develop after base branch lands.

varunagrawal (approved):

Issue #1433, opened 2023-01-31
Stack overflow for large graph


wanmeihuali:
## Description

Because GTSAM uses shared_ptr to manage Tree Structures, stack overflow caused by recursive destruction is a common issue when releasing these data structures(see [link](https://stackoverflow.com/questions/17804235/shared-pointers-delete-recursive-data-structures-recursively-and-the-stack-overf)). 

The behavior is usually a segment fault:
```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x00007f7487cdd085 in gtsam::ClusterTree<gtsam::GaussianFactorGraph>::Cluster::~Cluster (this=0x477b22a0, __in_chrg=<optimized out>) at /root/development/gtsam/gtsam/inference/ClusterTree.h:49
49          virtual ~Cluster() {}
(gdb) bt
#0  0x00007f7487cdd085 in gtsam::ClusterTree<gtsam::GaussianFactorGraph>::Cluster::~Cluster (this=0x477b22a0, __in_chrg=<optimized out>) at /root/development/gtsam/gtsam/inference/ClusterTree.h:49
#1  0x00007f7487cdd279 in __gnu_cxx::new_allocator<gtsam::ClusterTree<gtsam::GaussianFactorGraph>::Cluster>::destroy<gtsam::ClusterTree<gtsam::GaussianFactorGraph>::Cluster> (this=0x477b22a0, __p=0x477b22a0) at /usr/include/c++/10/ext/new_allocator.h:156
#2  0x00007f7487cdd23d in std::allocator_traits<std::allocator<gtsam::ClusterTree<gtsam::GaussianFactorGraph>::Cluster> >::destroy<gtsam::ClusterTree<gtsam::GaussianFactorGraph>::Cluster> (__a=..., __p=0x477b22a0) at /usr/include/c++/10/bits/alloc_traits.h:531
#3  0x00007f7487cdcf57 in std::_Sp_counted_ptr_inplace<gtsam::ClusterTree<gtsam::GaussianFactorGraph>::Cluster, std::allocator<gtsam::ClusterTree<gtsam::GaussianFactorGraph>::Cluster>, (__gnu_cxx::_Lock_policy)2>::_M_dispose (this=0x477b2290) at /usr/include/c++/10/bits/shared_ptr_base.h:560
```
Although simple solution such as increasing stack size exists:
```
ulimit -s unlimited
```
[...]

PR #1441, opened 2023-02-05
Fixing stack overflow issue for large tree

wanmeihuali:
This is a patch for [Issue-1433](https://github.com/borglab/gtsam/issues/1433). 
- The root cause for [Issue-1433](https://github.com/borglab/gtsam/issues/1433) is that smart pointers, especially shared_ptr, is not a good choice to manage data structures such as tree or graph, because they can easily cause stack overview during destruction. 
- In this patch, we manually release the root node of the tree data structure during a BFS when a destructor is called. In this way, the previous recursive release for tree nodes is now sequential. The algorithm is carefully designed so that it won't break the edges between nodes, and in this way keeps the same behavior even when some tree nodes are held by extra codes.
- We tested the modified code with the given example in [Issue-1433], and the issue is fixed with the patch.
- Does it solve all similar issues? We only fixed the three tree data structures we used in our use case(which is a bit more complex version of the bug reproduction code in the issue) in this patch. Considering there are tons of similar tree data structures in Gtsam, it is likely that a similar issue will happen again when other people use other tree data structures, and these issues should be able to be fixed in a similar way. However, fixing all data structures is a large work and we won't do it in this patch.
- To fully get rid of the stack overflow issue, one suggestion is to adopt some garbage collection library and have a global replacement for shared_ptr/make_shared. Of course, it is also necessary to make sure the garbage collection library does not use any recursive implementation. In this way, all similar issues will be solved without a large modification of the existing code, Considering that the garbage collection solution may cause some performance loss, a compiling flag can be used to control whether to use a smart pointer or garbage collection and let the user make the decision based on their use case.


dellaert:
Thanks @wanmeihuali !!! @ProfFan let's wait to run CI here as I'm trying to debug some other issues now.

ProfFan:
CI is failing, @wanmeihuali could you rebase to develop and see if it still fails?

wanmeihuali:
Yep, seems like a network or dep version issue. Just merged the latest develop. XD

dellaert:
Very interesting! Note, there are not tons of more tree data structures in gtsam, these are really the most important ones :-). I must say, though, I don’t fully understand the mechanism yet. Pushing the nodes into the BFS queue is one thing, but I’m assuming there is also an assumption that something will happen when the queue is deleted? That’s not really explained.

I also don’t understand why you do a clear on nodes_ first. Will  that not be a problem if reference counts go to zero? You probably have very good answer is for all these questions, it’s just a matter of getting them into the code comments for future reference :-)

wanmeihuali:
> Very interesting! Note, there are not tons of more tree data structures in gtsam, these are really the most important ones :-). I must say, though, I don’t fully understand the mechanism yet. Pushing the nodes into the BFS queue is one thing, but I’m assuming there is also an assumption that something will happen when the queue is deleted? That’s not really explained.
> 
> I also don’t understand why you do a clear on nodes_ first. Will that not be a problem if reference counts go to zero? You probably have very good answer is for all these questions, it’s just a matter of getting them into the code comments for future reference :-)

Yep it's quite tricky, I write some mermaid show a tree is released by this algorithm. The releasing is actually happening when a node is popped from the queue and then leaves the scope. For BFS, the queue is always empty after all. @dellaert 
Initial state:
```mermaid
graph TD
    T[Tree Data Structure]
    T-->A((A))
    A-->B((B))
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z[Context]
    Z--hold by other data structure-->D
```
Expect behavior:
- A, B, and E are released without recursive.
- D and G are not released, and G is still the child of D.

```mermaid
graph TD
    Q[BFS Queue]
    T[Tree Data Structure]
    T-->A((A))
    A-->B((B))
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z[Context]
    Z--hold by other data structure-->D
```

```mermaid
graph TD
    Q[BFS Queue]
    T[Tree Data Structure]
    Q==push into queue==>A((A))
    T-.set to nullptr.->A
    A-->B((B))
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z[Context]
    Z--hold by other data structure-->D
```


```mermaid
graph TD
    Q[BFS Queue]
    T[Tree Data Structure]
    Q-->A((A))
    A-->B((B))
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z[Context]
    Z--hold by other data structure-->D
```

```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    Q-.popping front.->A((A))
    Z==getting front==>A
    A-->B((B))
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z--hold by other data structure-->D
```

```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    Z-->A((A))
    A-->B((B))
    Q==BFS: adding children to queue==>B
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z--hold by other data structure-->D
```

```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    Z-.leaving scope-.->A((A, ref=0))
    A-->B((B))
    Q-->B
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z--hold by other data structure-->D
```


```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    A((A))-.A get released.->B((B))
    style A stroke:#333,stroke-width:4px
    Q--holds B so B will not be released recursively-->B
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z--hold by other data structure-->D
```


```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    Q-.popping front.->B((B))
    Z==getting front==>B
    B-->D((D))
    B-->E((E))
    D-->G((G))
    Z--hold by other data structure-->D
```


```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    Z-->B
    B((B))-->D((D))
    B-->E((E))
    D-->G((G))
    Q==BFS: adding children to queue==>D
    Q==BFS: adding children to queue==>E
    Z--hold by other data structure-->D
```


```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    Z-.leaving scope.->B
    B((B, ref=0))-->D((D))
    B-->E((E))
    D-->G((G))
    Q==BFS: adding children to queue==>D
    Q==BFS: adding children to queue==>E
    Z--hold by other data structure-->D
```


```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    B((B))-.B released.->D((D))
    style B stroke:#333,stroke-width:4px
    B-.B released.->E((E))
    D-->G((G))
    Q-->D
    Q-->E
    Z--hold by other data structure-->D
```


```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    D((D))-->G((G))
    Q-.pop front.->D
    Q-->E((E))
    Z--hold by other data structure-->D
    Z==getting front==>D
```

```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    D((D))-->G((G))
    Q-->E((E))
    Q==BFS: adding children to queue==>G
    Z--hold by other data structure-->D
    Z-->D
```

```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    D((D, ref=1))-->G((G))
    Q-->E((E))
    Q-->G
    Z--hold by other data structure-->D
    Z-.leaving scope.->D
```

```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    D((D))-->G((G))
    Q-.pop front.->E((E))
    Z==getting front==>E
    Q-->G
    Z--hold by other data structure-->D
```


```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    D((D))-->G((G))
    Z-.leaving scope.->E((E, ref=0))
    style E stroke:#333,stroke-width:4px
    Q-->G
    Z--hold by other data structure-->D
```

```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    D((D))-->G((G))
    Q-.pop front.->G
    Z==getting front==>G
    Z--hold by other data structure-->D
```

```mermaid
graph TD
    Q[BFS Queue]
    Z[Context]
    D((D))-->G((G, ref=1))
    Z-.leaving scope.->G
    Z--hold by other data structure-->D
```

dellaert:
Oh, that helps a lot! I will approve the PR, but I will also leave some  comments as to where you could add some of that explanation in the code.

gtsam/inference/BayesTree.h
```diff
@@ -27,6 +27,7 @@
 #include <gtsam/base/FastVector.h>
 
 #include <string>
+#include <queue>
```
dellaert, line 30: I would prefer that we add include headers and the code in the -inl.h file, rather than in this .h file.

gtsam/inference/BayesTree.h
```diff
@@ -111,6 +112,28 @@ namespace gtsam {
     /** Copy constructor */
     BayesTree(const This& other);
 
+    ~BayesTree() {
+      nodes_.clear();  // clear the map manually, as we cannot control the order of destruction in the map
```
dellaert, line 116: Add comment to say this reduces the ref count but not to zero yet

gtsam/inference/BayesTree.h
```diff
@@ -111,6 +112,28 @@ namespace gtsam {
     /** Copy constructor */
     BayesTree(const This& other);
 
+    ~BayesTree() {
```
dellaert, line 115: Add function level comment with link to this PR for the cool graphs…

gtsam/inference/BayesTree.h
```diff
@@ -111,6 +112,28 @@ namespace gtsam {
...
+        // do a BFS on the tree, for each node, add its children to the queue, and then delete it from the queue
+        // So if the reference count of the node is 1, it will be deleted, and because its children are in the queue,
+        // the deletion of the node will not trigger a recursive deletion of the children.
+        while (!bfs_queue.empty()) {
+          auto current = bfs_queue.front();
+          bfs_queue.pop();
+          for (auto child: current->children) {
+            bfs_queue.push(child);
+          }
+        }
```
dellaert, line 133: Add a comment here to say share_ptr will drop to zero when current goes out of scope

gtsam/inference/ClusterTree.h
```diff
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include <queue>
```
dellaert, line 12: All same comments for this file

gtsam/inference/EliminationTree.h
```diff
@@ -19,6 +19,7 @@
 
 #include <utility>
 #include <memory>
+#include <queue>
```
dellaert, line 22: And for this file

dellaert (changes requested):
Amazing!! I do feel strongly that things are not cluttering the header if possible, so will “request changes” for now and re-review on push…

dellaert (approved):
Very cool, a great addition to enable GTSAM to do large scale inference!

ProfFan (approved):
Thanks @wanmeihuali :)

ShuangLiu1992:
I wonder if the same fix should be applied to BayesTreeCliqueBase?

ShuangLiu1992:
current develop branch seems to have missed this commit for gtsam/inference/BayesTreeCliqueBase.h

https://github.com/wanmeihuali/gtsam/commit/1d06f1e7667d3f9c9761161cbbf9e47aa3955559

wanmeihuali:
> current develop branch seems to have missed this commit for gtsam/inference/BayesTreeCliqueBase.h
> 
> [wanmeihuali@1d06f1e](https://github.com/wanmeihuali/gtsam/commit/1d06f1e7667d3f9c9761161cbbf9e47aa3955559)

@ShuangLiu1992  Yep, As I mentioned I only fixed the issue in my use case... Because Clique Tree is much smaller than the original tree, it did not trigger stack overflow for my case.
You may ask @ProfFan and @dellaert for another patch.

wanmeihuali:
@ShuangLiu1992 I just got some time to check it, gtsam/inference/BayesTreeCliqueBase.h is the base class for the node in BayesTree, so the patch should already fix the issue if you are using BayesTree.h... You can check the image I attached for how it works. If you still got a stackoverflow, maybe you are using another tree data structure with BayesTreeCliqueBase?

ShuangLiu1992:
@wanmeihuali Thanks for taking the time to check. I will double check. On some platforms we are constrained with a stack size of 1mb so the stackover flow might have came from somewhere else in the library. Debug builds are too big to fit on the target machine and Release builds crashes without giving any useful information.....

ProfFan:
1MB of stack is a bit too small for current GTSAM but you can try changing some stack variables used to `unique_ptr`s

PR #1442, opened 2023-02-05
Refactor elimination...

dellaert:
... to no longer use boost::variant. Unfortunately, we can't use std::variant (a solution first proposed and implemented by @kartikarcot) as it creates problems in the python wrapper, so I just made overloads. There is a bit of code duplication (mostly comments really) which at one point we could try to sanitize, but in the end this is less complex, more readable, and works with the wrapper.

Also added a set of tests in `SymbolicFactorGraph.cpp` to make sure everything behaves correctly.

This PR also includes changes from @kartikarcot to remove boost::variant altogether, used in some places to specify parameters.


gtsam/inference/EliminateableFactorGraph-inst.h
```diff
@@ -236,29 +236,53 @@ namespace gtsam {
...
+      return marginalMultifrontalBayesNet(std::cref(marginalVarsOrdering), marginalizationOrdering, function, variableIndex);
+    }
+  }
 
+  /* ************************************************************************* */
+  template<class FACTORGRAPH>
+  std::shared_ptr<typename EliminateableFactorGraph<FACTORGRAPH>::BayesNetType>
+    EliminateableFactorGraph<FACTORGRAPH>::marginalMultifrontalBayesNet(
+    const KeyVector& variables,
+    const Eliminate& function, OptionalVariableIndex variableIndex) const
```
gchenfc, line 258: Could this function get an optional/default argument (or template argument) for `forceOrder` and then the `Ordering` version calls this with `forceOrder=true`?

I guess maybe you didn't want to make the public interface confusing

gtsam/inference/EliminateableFactorGraph.h
```diff
@@ -217,60 +215,108 @@ namespace gtsam {
 
     /** Compute the marginal of the requested variables and return the result as a Bayes net.  Uses
      *  COLAMD marginalization ordering by default
-     *  @param variables Determines the variables whose marginal to compute, if provided as an
-     *         Ordering they will be ordered in the returned BayesNet as specified, and if provided
-     *         as a KeyVector they will be ordered using constrained COLAMD.
-     *  @param function Optional dense elimination function, if not provided the default will be
-     *         used.
+     *  @param variables Determines the *ordered* variables whose marginal to compute, 
+     *         will be ordered in the returned BayesNet.
```
gchenfc, line 219: I think it would be clearer as "will be ordered in the returned BayesNet _as specified_" (like in the original docstring)

gtsam/inference/EliminateableFactorGraph.h
```diff
@@ -217,60 +215,108 @@ namespace gtsam {
...
     std::shared_ptr<BayesNetType> marginalMultifrontalBayesNet(
-      boost::variant<const Ordering&, const KeyVector&> variables,
+      const Ordering& variables,
+      const Eliminate& function = EliminationTraitsType::DefaultEliminate,
+      OptionalVariableIndex variableIndex = {}) const;
+
+    /** Compute the marginal of the requested variables and return the result as a Bayes net.  Uses
+     *  COLAMD marginalization ordering by default
+     *  @param variables Determines the variables whose marginal to compute,
+     *         will be ordered using constrained COLAMD.
```
gchenfc, line 232: Might be worth extra-clarifying that this version will NOT use the key vector ordering, since KeyVector kind of implies an ordering (unlike KeySet), e.g.
```c++
//  @param variables Determines _which_ variables whose marginal to compute,
//           will be ordered using constrained COLAMD.  Will NOT use
//           the KeyVector ordering; use `Ordering(variables)` to specify the variable ordering.
```

Or something like that (same with the other overloads)

gtsam/symbolic/tests/testSymbolicFactorGraph.cpp
```diff
@@ -124,14 +124,40 @@ TEST(SymbolicFactorGraph, eliminatePartialMultifrontal) {
...
       *simpleTestGraph2.marginalMultifrontalBayesNet(Ordering{0, 1, 2, 3});
-  EXPECT(assert_equal(expectedBayesNet, actual1));
+  EXPECT(assert_equal(expectedBayesNet, actual));
+}
+
+TEST(SymbolicFactorGraph, MarginalMultifrontalBayesNetKeyVector) {
+  auto expectedBayesNet = SymbolicBayesNet({0, 1, 2})({2, 1, 3})({1, 3})({3});
+
+  SymbolicBayesNet actual =
+      *simpleTestGraph2.marginalMultifrontalBayesNet(KeyVector{0, 1, 2, 3});
```
gchenfc, line 139: Can you add comment or rename that this is testing COLAMD ordering?  When reviewing I initially read this file and got confused until I read EliminateableFactorGraph and remembered that KeyVector is treated unordered.

gtsam/linear/VectorValues.cpp
```diff
@@ -92,7 +92,7 @@ namespace gtsam {
       // Use this trick to find the value using a hint, since we are inserting
       // from another sorted map
       size_t oldSize = values_.size();
-      hint = values_.emplace_hint(hint, key, value);
+      hint = values_.insert(hint, {key, value});
```
gchenfc, line 95: Is this related to the PR or just random?

gchenfc (approved):
All good - only nitpicks but LGTM.

As you mentioned, a future PR might reduce code duplication.

gtsam/inference/EliminateableFactorGraph-inst.h
dellaert: Mostly I want to stay backwards compatible. I thought about the option you suggested, in an extra variant. But that should be part of a de-duplication PR....

gtsam/inference/EliminateableFactorGraph.h
dellaert: I like that, will do.

gtsam/inference/EliminateableFactorGraph.h
dellaert: Also good!

gtsam/inference/EliminateableFactorGraph-inst.h
```diff
@@ -301,29 +341,53 @@ namespace gtsam {
...
 
       // Split up ordering
-      const size_t nVars = variablesOrOrdering->size();
+      const size_t nVars = variables.size();
       Ordering marginalizationOrdering(totalOrdering.begin(), totalOrdering.end() - nVars);
       Ordering marginalVarsOrdering(totalOrdering.end() - nVars, totalOrdering.end());
 
       // Call this function again with the computed orderings
-      return marginalMultifrontalBayesTree(marginalVarsOrdering, marginalizationOrdering, function, variableIndex);
+      return marginalMultifrontalBayesTree(std::cref(marginalVarsOrdering), marginalizationOrdering, function, variableIndex);
```
kartikarcot, line 382: We can remove `cref` since we don't use reference_wrapper anymore

gtsam/inference/EliminateableFactorGraph-inst.h
```diff
@@ -236,29 +236,53 @@ namespace gtsam {
...
 
       // Split up ordering
-      const size_t nVars = variablesOrOrdering->size();
+      const size_t nVars = variables.size();
       Ordering marginalizationOrdering(totalOrdering.begin(), totalOrdering.end() - nVars);
       Ordering marginalVarsOrdering(totalOrdering.end() - nVars, totalOrdering.end());
 
       // Call this function again with the computed orderings
-      return marginalMultifrontalBayesNet(marginalVarsOrdering, marginalizationOrdering, function, variableIndex);
+      return marginalMultifrontalBayesNet(std::cref(marginalVarsOrdering), marginalizationOrdering, function, variableIndex);
```
kartikarcot, line 277: We can remove `cref` since we don't use reference_wrapper anymore

gtsam/linear/VectorValues.cpp
dellaert: It fails in develop !!! It snuck in so I put the fix in here to save on CI..

gtsam/inference/EliminateableFactorGraph-inst.h
dellaert: Good catch !!!

gtsam/inference/EliminateableFactorGraph-inst.h
dellaert: Yep, will fix !

PR #1447, opened 2023-02-06
Update CppUnitLite to not rely on boost

varunagrawal:
Also made some CMake updates and fixed some warnings.


dellaert (approved):
Thanks !!!! Will merge in the PR. Although the CI seems to be broken for a non-GTSAM reason.

Issue #1448, opened 2023-02-07
Incorrect MPE for Long Chains of Discrete Factor Graphs


pcl15423:
## Description
Hi team, thanks for the wonderful library!

I've run into an issue where I have a fairly large Discrete Factor Graph that I believe is running into an underflow type of error. I've copied a minimal toy example below that shows the same symptom. I'm not sure where to turn for documentation of python bindings or how to debug the optimization further. Thank you for the help!

## Steps to reproduce
Toy example is a hidden markov model with 3 possible states and 200 Observations. In this example, the transition matrix is mostly identity. The desired state outputs any number [0, 199] with equal probability. All other states output 200 with probability 1. I set each measurement equal to the observation index (so only [0,200] are observed and the desired state should be preferred always).

``` python
import gtsam
import numpy as np

import gtbook
from gtbook.discrete import Variables
VARIABLES = Variables()

def toy_problem(num_states, num_obs, desired_idx):
    states = list(range(num_states))
    X = VARIABLES.discrete_series("X", range(num_obs), states)
    Z = VARIABLES.discrete_series("Z", range(num_obs), range(num_obs+1))
    graph = gtsam.DiscreteFactorGraph()

    # Mostly identity transition matrix
    transitions = np.eye(num_states)
    
    # Needed otherwise mpe is always state 0?
    transitions += 0.1/(num_states)
    
    transition_cpt = []
    for i in range(0, num_states):
        transition_row = "/".join([str(x) for x in transitions[i]])
        transition_cpt.append(transition_row)
    transition_cpt = " ".join(transition_cpt)

    for i in reversed(range(1, num_obs)):
        transition_conditional = gtsam.DiscreteConditional(X[i], [X[i-1]], transition_cpt)
        graph.push_back(transition_conditional)

    # Contrived example such that the desired state gives measurements [0, num_obs) with equal probability
    #   but all other states always give measurement num_obs
    obs = np.zeros((num_states, num_obs+1))
    obs[:,-1] = 1
    obs[DESIRED_STATE,0: -1] = 1
    obs[DESIRED_STATE,-1] = 0
    obs_cpt_list = []
    for i in range(0, num_states):
        obs_row = "/".join([str(z) for z in obs[i]])
        obs_cpt_list.append(obs_row)
    obs_cpt = " ".join(obs_cpt_list)
    
    # Contrived example where each measurement is its own index
    for i in range(0, num_obs):
[...]

PR #1449, opened 2023-02-07
Add normalization to max-product, avoiding underflow.

keevindoherty:
This is a quick attempted fix for #1448 where it was observed that performing max-product elimination on long chains resulted in numerical underflow. The fix is to normalize the intermediate products before applying the max operator.

I'll update this PR in a few minutes once I get the proposed Python test case in #1448 working. There was an additional dependency in that toy problem on `gtbook` that it seems prudent to avoid in the base `gtsam` repo, so I'm spinning my own minimal `discrete_series` to mimic the behavior of the function of the same name in the `gtbook.Variables` class.

### DIFF
--- a/gtsam/discrete/DiscreteFactorGraph.cpp
+++ b/gtsam/discrete/DiscreteFactorGraph.cpp
@@ -121,6 +121,11 @@ namespace gtsam {
     for (auto&& factor : factors) product = (*factor) * product;
     gttoc(product);
 
+    // Normalize the product factor to prevent underflow
+    Ordering ordering = Ordering::Colamd(DiscreteFactorGraph{product});
+    DecisionTreeFactor::shared_ptr normalization = product.sum(ordering);
+    product = product / (*normalization);
+
     // max out frontals, this is the factor on the separator
     gttic(max);
     DecisionTreeFactor::shared_ptr max = product.max(frontalKeys);
--- a/gtsam/discrete/DiscreteFactorGraph.cpp
+++ b/gtsam/discrete/DiscreteFactorGraph.cpp
@@ -121,9 +121,10 @@ namespace gtsam {
     for (auto&& factor : factors) product = (*factor) * product;
     gttoc(product);
 
-    // Normalize the product factor to prevent underflow
-    Ordering ordering = Ordering::Colamd(DiscreteFactorGraph{product});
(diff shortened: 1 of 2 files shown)


keevindoherty:
The added test is passing on my machine. I removed the `gtbook` dependency from the toy problem provided by @pcl15423. One thing to note is that I kept the added `numpy` import, but I wasn't sure if that was ok with you @dellaert.

In any case, I think this should be all clear for review now?

gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -121,6 +121,11 @@ namespace gtsam {
     for (auto&& factor : factors) product = (*factor) * product;
     gttoc(product);
 
+    // Normalize the product factor to prevent underflow
+    Ordering ordering = Ordering::Colamd(DiscreteFactorGraph{product});
```
dellaert, line 125: Can’t we get keys directly?

dellaert (commented):
I will check out on my machine as I want to understand a bit better

gtsam/discrete/DiscreteFactorGraph.cpp
keevindoherty: Yes, I don’t see why not 👍

keevindoherty:
> I will check out on my machine as I want to understand a bit better

Please let me know any insights - as I mentioned, this was kind of a quick and dirty hack that I haven’t had time to properly verify myself. I wanted to convince myself this was the correct approach before making a PR, but it seems that demand for discrete factor graphs has outpaced me and this became relevant sooner than anticipated..

gtsam/discrete/DiscreteFactorGraph.cpp
dellaert: I understand the code now. I tried on my machine, and the following works (and is more efficient):
```c++
    // Sum all potentials by pretending all keys are frontal:
    auto normalization = product.sum(product.size());

    // Normalize the product factor to prevent underflow.
    product = product / *normalization;
```

dellaert (approved):
Awesome hotfix! Can you use my proposed modification and merge?

keevindoherty:
Done, and will merge once the CI completes. Sum-product may also be having issues - I'll follow up on #1448.

PR #1453, opened 2023-02-10
Add normalization to sum-product, avoiding underflow.

### LINKED ISSUE #1448: Incorrect MPE for Long Chains of Discrete Factor Graphs

keevindoherty:
This quick fix should actually resolve #1448 now. Adds the same product normalization from #1449 to the `DiscreteFactorGraph::DiscreteEliminate` function. This broke an existing test in `testDiscreteFactorGraph.cpp` due to an expected result being unnormalized, where the result from `DiscreteEliminate` was normalized, so I modified the test to expect a normalized result. I also added a "visual test" to `test_DiscreteFactorGraph.py` to verify the expected results in #1448.

Couple notes for reviewer(s):
- *Highest priority:* The test I added in `test_DiscreteFactorGraph.py` is not a real test, i.e. it does not computationally verify the result and report back, just "prints some stuff." It should not be merged in its current state, but I don't think that simply verifying the number of assignments in a sum-product result is the right move here. If there is a suggestion for a more robust test, I'm happy to cook something up and modify accordingly, but I want to make sure we're on the same page before I try throwing something together.
- *Low priority:* The patch I added in `testDiscreteFactorGraph.cpp` normalizes the expect result before comparison, but it may be worth thinking longer term about mechanisms for normalizing discrete factor graph results (e.g. should discrete conditionals simply be normalized be default? or not, in which case one option would be to add a new assert for discrete to compare "up to scale" where appropriate?). This seems not urgent but perhaps worth earmarking. The latter seems particularly useful if we want to keep constructing "expected" results for future tests without normalizing "by hand." Part of my thinking on this is: I'm not actually sure personally how to implement the same patch in `testDiscreteFactorGraph.cpp` if we had to do it using the GTSAM Python API; is it possible? I guess the last possibility would be to add something like a `DecisionTreeFactor& DecisionTreeFactor::normalize() const` that returns a normalized copy of a DecisionTreeFactor?

Let me know what you think?

### DIFF
--- a/gtsam/discrete/DiscreteFactorGraph.cpp
+++ b/gtsam/discrete/DiscreteFactorGraph.cpp
@@ -210,6 +210,12 @@ namespace gtsam {
     for (auto&& factor : factors) product = (*factor) * product;
     gttoc(product);
 
+    // Sum all the potentials by pretending all keys are frontal:
+    auto normalization = product.sum(product.size());
+
+    // Normalize the product factor to prevent underflow.
+    product = product / (*normalization);
+
     // sum out frontals, this is the factor on the separator
     gttic(sum);
     DecisionTreeFactor::shared_ptr sum = product.sum(frontalKeys);
--- a/gtsam/discrete/DiscreteFactorGraph.cpp
+++ b/gtsam/discrete/DiscreteFactorGraph.cpp
@@ -121,8 +121,8 @@ namespace gtsam {
     for (auto&& factor : factors) product = (*factor) * product;
     gttoc(product);
 
-    // Sum all the potentials by pretending all keys are frontal:
-    auto normalization = product.sum(product.size());
+    // Max over all the potentials by pretending all keys are frontal:
+    auto normalization = product.max(product.size());
 
     // Normalize the product factor to prevent underflow.
     product = product / (*normalization);
@@ -210,8 +210,8 @@ namespace gtsam {
     for (auto&& factor : factors) product = (*factor) * product;
     gttoc(product);
 
-    // Sum all the potentials by pretending all keys are frontal:
-    auto normalization = product.sum(product.size());
+    // Max over all the potentials by pretending all keys are frontal:
+    auto normalization = product.max(product.size());
 
     // Normalize the product factor to prevent underflow.
     product = product / (*normalization);
--- a/python/gtsam/tests/test_DiscreteFactorGraph.py
+++ b/python/gtsam/tests/test_DiscreteFactorGraph.py
@@ -14,7 +14,7 @@ Author: Frank Dellaert
 import unittest
 
 import numpy as np
-from gtsam import DiscreteConditional, DiscreteFactorGraph, DiscreteKeys, DiscreteValues, Ordering, Symbol
+from gtsam import DecisionTreeFactor, DiscreteConditional, DiscreteFactorGraph, DiscreteKeys, DiscreteValues, Ordering, Symbol
 from gtsam.utils.test_case import GtsamTestCase
 
 OrderingType = Ordering.OrderingType
@@ -250,7 +250,7 @@ class TestDiscreteFactorGraph(GtsamTestCase):
 
         # Ensure that the stationary distribution is positive and normalized
         stationary_dist /= np.sum(stationary_dist)
-        expected = stationary_dist.flatten()
+        expected = DecisionTreeFactor(X[chain_length-1], stationary_dist.flatten())
 
         # The transition matrix parsed by DiscreteConditional is a row-wise CPT
         transitions = transitions.T
@@ -271,16 +271,8 @@ class TestDiscreteFactorGraph(GtsamTestCase):
         # Get the DiscreteConditional representing the marginal on the last factor
         last_marginal = sum_product.at(chain_length - 1)
 
-        # Extract the actual marginal probabilities
-        assignment = DiscreteValues()
-        marginal_probs = []
-        for i in range(num_states):
-            assignment[X[chain_length - 1][0]] = i
-            marginal_probs.append(last_marginal(assignment))
-        marginal_probs = np.array(marginal_probs)
-
         # Ensure marginal probabilities are close to the stationary distribution
-        self.gtsamAssertEquals(expected, marginal_probs)
+        self.gtsamAssertEquals(expected, last_marginal)
 
 if __name__ == "__main__":
     unittest.main()
--- a/python/gtsam/tests/test_DiscreteFactorGraph.py
+++ b/python/gtsam/tests/test_DiscreteFactorGraph.py
@@ -216,5 +216,70 @@ class TestDiscreteFactorGraph(GtsamTestCase):
 
         self.assertEqual(vals, [desired_state]*num_obs)
 
+    def test_sumProduct_chain(self):
+        """
+        Test for numerical underflow in EliminateDiscrete on long chains.
+        Adapted from the toy problem of @pcl15423
+        Ref: https://github.com/borglab/gtsam/issues/1448
+        """
+        num_states = 3
+        num_obs = 200
+        desired_state = 1
+        states = list(range(num_states))
+
+        # Helper function to mimic the behavior of gtbook.Variables discrete_series function
+        def make_key(character, index, cardinality):
+            symbol = Symbol(character, index)
+            key = symbol.key()
+            return (key, cardinality)
+
+        X = {index: make_key("X", index, len(states)) for index in range(num_obs)}
+        Z = {index: make_key("Z", index, num_obs + 1) for index in range(num_obs)}
+        graph = DiscreteFactorGraph()
+
+        # Mostly identity transition matrix
+        transitions = np.eye(num_states)
+
+        # Needed otherwise mpe is always state 0?
+        transitions += 0.1/(num_states)
+
+        transition_cpt = []
+        for i in range(0, num_states):
+            transition_row = "/".join([str(x) for x in transitions[i]])
+            transition_cpt.append(transition_row)
+        transition_cpt = " ".join(transition_cpt)
+
+        for i in reversed(range(1, num_obs)):
+            transition_conditional = DiscreteConditional(X[i], [X[i-1]], transition_cpt)
+            graph.push_back(transition_conditional)
+
+        # Contrived example such that the desired state gives measurements [0, num_obs) with equal probability
+        #   but all other states always give measurement num_obs
+        obs = np.zeros((num_states, num_obs+1))
+        obs[:,-1] = 1
+        obs[desired_state,0: -1] = 1
+        obs[desired_state,-1] = 0
+        obs_cpt_list = []
+        for i in range(0, num_states):
+            obs_row = "/".join([str(z) for z in obs[i]])
+            obs_cpt_list.append(obs_row)
+        obs_cpt = " ".join(obs_cpt_list)
+
+        # Contrived example where each measurement is its own index
+        for i in range(0, num_obs):
+            obs_conditional = DiscreteConditional(Z[i], [X[i]], obs_cpt)
+            factor = obs_conditional.likelihood(i)
+            graph.push_back(factor)
+
+        mpe = graph.optimize()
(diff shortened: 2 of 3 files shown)


dellaert:
> * _Highest priority:_ The test I added in `test_DiscreteFactorGraph.py` is not a real test,...If there is a suggestion for a more robust test, I'm happy to cook something up

So, I have been thinking of HMMs, as you know, and a single sweep of an HMM is very easy to code up iteratively, hence a good test would be to compare a long HMM with the last marginal result of elimination.

Alternatively, even easier is a measurement-free Markov chain: The marginal at the end should be T^n which will be the stationary distribution, i.e. Eigenvector of T.

> * _Low priority:_ The patch I added in `testDiscreteFactorGraph.cpp` normalizes the expect result before comparison, but it may be worth thinking longer term about mechanisms for normalizing discrete factor graph results

Discrete factors are not normalized (and can even be confsing if we do). Maybe "normalizing" by max would be better to not confuse peeople.

DiscreteConditionals should already be normalized.

keevindoherty:
> Alternatively, even easier is a measurement-free Markov chain: The marginal at the end should be T^n which will be the stationary distribution, i.e. Eigenvector of T.

I took a first crack at implementing this test in the latest commit. Maybe you or Varun can take a look when you get a chance?

> DiscreteConditionals should already be normalized.

Ah, yep, sorry, I got my wires crossed here. The issue was that the comparison in `testDiscreteFactorGraph.cpp` was comparing two *DecisionTreeFactor* objects, which are _not_ normalized (and probably should not be?), so some extra care needs to be taken when constructing an "expected" DecisionTreeFactor to ensure normalization for testing discrete.

dellaert:
> DecisionTreeFactor

No, DecisionTreeFactors are not normalized as a rule, unless they are the base class of a Conditional

gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -210,6 +210,12 @@ namespace gtsam {
     for (auto&& factor : factors) product = (*factor) * product;
     gttoc(product);
 
+    // Sum all the potentials by pretending all keys are frontal:
+    auto normalization = product.sum(product.size());
```
dellaert, line 214: Wondering whether we could use `max` here to avoid the impression the factor needs to be normalized…

dellaert (approved):
Awesome!!! One comment about max. I think sum creates a wrong expectation…

keevindoherty:
Yeah, max is reasonable too. “Normalizing” w.r.t. the max seems to be the usual technique for dealing with this when performing computations in log-space, too (Koller and Friedman discuss it, if I remember correctly).

I also was wondering if there was a nicer way to extract the probabilities from a DiscreteConditional in the Python API than the “brute force” method I used (see test case)?

python/gtsam/tests/test_DiscreteFactorGraph.py
```diff
@@ -216,5 +216,71 @@ def make_key(character, index, cardinality):
...
+        # Get the DiscreteConditional representing the marginal on the last factor
+        last_marginal = sum_product.at(chain_length - 1)
+
+        # Extract the actual marginal probabilities
+        assignment = DiscreteValues()
+        marginal_probs = []
+        for i in range(num_states):
+            assignment[X[chain_length - 1][0]] = i
+            marginal_probs.append(last_marginal(assignment))
+        marginal_probs = np.array(marginal_probs)
```
keevindoherty, line 280: I wasn't sure if there was a nicer way to extract these probabilities from the discrete conditional (i.e. simpler than 'brute force' iterating over each assignment)

dellaert:
> Yeah, max is reasonable too. “Normalizing” w.r.t. the max seems to be the usual technique for dealing with this when performing computations in log-space, too (Koller and Friedman discuss it, if I remember correctly).
> 
> I also was wondering if there was a nicer way to extract the probabilities from a DiscreteConditional in the Python API than the “brute force” method I used (see test case)?

I would prefer to go with max: the existence of a 1.0 in the decision tree somewhere would be additional verification we are talking about a factor and not a conditional

python/gtsam/tests/test_DiscreteFactorGraph.py
dellaert: Hmm, yes, there must be. Let me think...

dellaert: So, there does not seem to be a nice method for this, but a trick used in test_DiscreteConditional.py is to go the opposite way: create an *expected* marginal by specifying a vector of doubles, and then use gtsamAssertEquals:
```python
        expected = DecisionTreeFactor(Y, "0.8 0.6 0.5")
        self.gtsamAssertEquals(actual1, expected, 1e-9)
```

dellaert: Note you can also use a vector<double>

keevindoherty:
> I would prefer to go with max: the existence of a 1.0 in the decision tree somewhere would be additional verification we are talking about a factor and not a conditional

Sounds good to me. I can make this change tomorrow or Monday so we can get this merged. Just to check - we’ll normalize to the max in EliminateForMPE (max product) too, right? And then is it OK with you if that’s on this same PR? Should just be a one line change, I think.

python/gtsam/tests/test_DiscreteFactorGraph.py
keevindoherty: This looks good to me, and looks a bit cleaner, I’ll do that.

dellaert:
> > I would prefer to go with max: the existence of a 1.0 in the decision tree somewhere would be additional verification we are talking about a factor and not a conditional
> 
> Sounds good to me. I can make this change tomorrow or Monday so we can get this merged. Just to check - we’ll normalize to the max in EliminateForMPE (max product) too, right? And then is it OK with you if that’s on this same PR? Should just be a one line change, I think.

Yes, absolutely. Note I don't know whether there already is a `max`, but should be easy enough, cribbing from sum.

python/gtsam/tests/test_DiscreteFactorGraph.py
```diff
@@ -216,5 +216,63 @@ def make_key(character, index, cardinality):
...
+        transitions /= np.sum(transitions, axis=0)
+
+        # The stationary distribution is the eigenvector corresponding to eigenvalue 1
+        eigvals, eigvecs = np.linalg.eig(transitions)
+        stationary_idx = np.where(np.isclose(eigvals, 1.0))
+        stationary_dist = eigvecs[:, stationary_idx]
+
+        # Ensure that the stationary distribution is positive and normalized
+        stationary_dist /= np.sum(stationary_dist)
+        expected = DecisionTreeFactor(X[chain_length-1], stationary_dist.flatten())
```
dellaert, line 253: Awesome !!!!

dellaert:
Killing all but required CI and will merge after required passes.

dellaert:
Thanks Kevin, this is awesome !

PR #1456, opened 2023-02-11
Fixed warnings that arise from stricter compiler flags

dellaert:
These are a number of small, mostly inconsequential edits that triggered warnings in a rather strict (but still not -Werror) environment.

### DIFF
--- a/cmake/GtsamBuildTypes.cmake
+++ b/cmake/GtsamBuildTypes.cmake
@@ -130,6 +130,8 @@ else()
     $<$<CXX_COMPILER_ID:GNU>:-Wreturn-local-addr -Werror=return-local-addr>            # Error: return local address
     $<$<CXX_COMPILER_ID:Clang>:-Wreturn-stack-address   -Werror=return-stack-address>  # Error: return local address
     $<$<CXX_COMPILER_ID:Clang>:-Wno-misleading-indentation>  # Eigen triggers a ton!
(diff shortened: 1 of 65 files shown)


cmake/GtsamBuildTypes.cmake
```diff
@@ -130,6 +130,8 @@ else()
     $<$<CXX_COMPILER_ID:GNU>:-Wreturn-local-addr -Werror=return-local-addr>            # Error: return local address
     $<$<CXX_COMPILER_ID:Clang>:-Wreturn-stack-address   -Werror=return-stack-address>  # Error: return local address
     $<$<CXX_COMPILER_ID:Clang>:-Wno-misleading-indentation>  # Eigen triggers a ton!
+    $<$<CXX_COMPILER_ID:Clang>:-Wno-weak-template-vtables>  # TODO(dellaert): don't know how to resolve
+    $<$<CXX_COMPILER_ID:Clang>:-Wno-weak-vtables>  # TODO(dellaert): don't know how to resolve
```
gchenfc, line 134: Can you mention which file these occurred in, just for reference?

gchenfc (approved):
Lgtm

cmake/GtsamBuildTypes.cmake
dellaert: Don’t remember all, but basically it occurred in several .cpp files where a template was explicitly instantiated.

dellaert:
Thanks @gchenfc !!

PR #1459, opened 2023-02-12
No-op concepts

dellaert:
My attempt at quickly hacking a replacement was rather ill-advised, as in debugging expression storage I found it added a large overhead to every data structure including the concepts. I have made GTSAM_CONCEPT_ASSERT a no-op for now, and reverted the semicolon change done in fix/warnings.


dellaert:
The semicolons *are* needed. Will revert last commit.

PR #1461, opened 2023-02-12
Fix wrongly formatted link to repo in citation

shervinkoushan:


dellaert:
Thanks !!! I will close this PR because I don't want to run CI for this small change, but I am wrapping it into an updated README landing shortly !

PR #1462, opened 2023-02-12
Patch fixes from 4.3

dellaert:
Adding a number of fixes from develop to 4.2 release branch.


dellaert:
This was mostly about improvements to expressions memory handling that I wanted to back-port to 4.2, but it does not compile out of the box and I could not root-cause it even after spending some time with it. It's one of these things that is also subtly different on windows so hard to debug. Closing for now and unlikely to pick up again...

PR #1464, opened 2023-02-12
Allow override of BUILD_SHARED_LIBS

Ibarria:
The cmake option BUILD_SHARED_LIBS is a cmake built-in option
    to control if libraries are by default SHARED or STATIC.

    For large projects, it is desired to be able to design gtsam as a
    shared or static library regardless of the value of BUILD_SHARED_LIBS.
    This change is unobtrusive, two new cmake options are created, to
    force gtsam to be a shared or static library. If neither option is
    set (this is the default), the behavior of gtsam remains unchanged
    which is to use BUILD_SHARED_LIBS decision.


dellaert (approved):
LGTM !

dellaert:
I will merge after CI passes.

dellaert:
Again, windows fails. It is weird, this issue was fixed by defining `_ENABLE_EXTENDED_ALIGNED_STORAGE` in GtsamBuildTypes.cmake, but you did not modify that...

dellaert:
I will merge this as many Linux CI's actually succeed, so I think it's more OOM stuff.

dellaert:
Dang, close by accident. Will run CI again, but only required...

PR #1481, opened 2023-03-01
Add constant term for hybrid models post-hoc

varunagrawal:
I added 2 tests for hybrid estimation where I manually add the constant term relating to the noise model differences directly to the `HybridGaussianFactorGraph`, done via the function `addConstantTerm`.

This provides an easy alternative to specifying conditionals and going back to a factor graph via the likelihoods which may not be straightforward to do (e.g. legged state estimation). The good news is that on analysis, all we need is the noise model sigmas, the dimensionality of the noise model and the index/indices of the looser model.


gtsam/hybrid/GaussianMixtureFactor.h
```diff
@@ -143,6 +143,9 @@ class GTSAM_EXPORT GaussianMixtureFactor : public HybridFactor {
    */
   double error(const HybridValues &values) const override;
 
+  /// Getter for GaussianFactor decision tree
+  Factors factors() const { return factors_; }
```
dellaert, line 147: Make this return const& or you’ll make a copy!

gtsam/hybrid/HybridSmoother.cpp
```diff
@@ -45,7 +45,7 @@ Ordering HybridSmoother::getOrdering(
   std::copy(allDiscrete.begin(), allDiscrete.end(),
             std::back_inserter(newKeysDiscreteLast));
 
-  const VariableIndex index(newFactors);
+  const VariableIndex index(factors);
```
dellaert, line 48: Seems significant. Why did it ever work?

gtsam/hybrid/tests/testHybridEstimation.cpp
```diff
@@ -503,6 +503,188 @@ TEST(HybridEstimation, CorrectnessViaSampling) {
...
+      double factor_log_constant = -0.5 * d * log2pi + log(1.0 / noise_loose);
+
+      GaussianFactorGraph _gfg;
+      _gfg.push_back(gf);
+      Vector c(d);
+      for (size_t i = 0; i < d; i++) {
+        c(i) = std::sqrt(2.0 * (logConstant - factor_log_constant));
+      }
+
+      auto constantFactor = std::make_shared<JacobianFactor>(c);
```
dellaert, line 533: Emplace_shared is preferred (everywhere)

dellaert (approved):
Maybe the PR comment is out of date, but I did not find a “ addConstantTerm” anywhere. Main comment: you say “easy” but we have found through substantial pain that there is *nothing* easy about specifying the correct constant term. That’s why I strongly prefer the likelihood route. PS, since you have no ratio tests here either, all bets are off whether you did correctly here… I’ll approve but it’s your pain :-)

varunagrawal:
> That’s why I strongly prefer the likelihood route.

But this route doesn't provide prescriptive way to use likelihoods in the definition of the graph. What would be the likelihood function for the contact factor? What would be the measurement value we would use to convert it the likelihood to a factor?

> PS, since you have no ratio tests here either, all bets are off whether you did correctly here… I’ll approve but it’s your pain :-)

I used a graph which is already shown to be correct via a ratio test. By transitivity, we can assert correctness.

PR #1483, opened 2023-03-02
 A few minor fixes to allow compiling 'boost-free' gtsam without compi…

ankur-rc:
A few minor fixes to allow compiling 'boost-free' gtsam without  'gtsam_unstable'.


varunagrawal:
Some comments:
1. `FixedLagSmoother` has been moved to `gtsam/nonlinear` so we should probably work to move `IncrementalFixedLagSmoother` to `gtsam/nonlinear` as well rather than moving the example back to `gtsam_unstable`.
2. How does your cmake update to `examples/CMakeLists.txt` work if `GTSAM_BUILD_UNSTABLE=ON`?
3. I'm a bit surprised by the change to the `HybridSmoother`. It has worked well for me, so the fact that we're using a (suboptimal) loop to add the factors is strange to me. @ankur-rc can you please provide more details?
4. Also please provide comments on the extra headers? I generally compile GTSAM without the unstable, so the fact that headers needed to be added is again strange to me. I am curious to see what's the difference.

examples/CMakeLists.txt
```diff
@@ -20,4 +20,4 @@ if (NOT GTSAM_USE_BOOST_FEATURES)
       )
 endif()
 
-gtsamAddExamplesGlob("*.cpp" "${excluded_examples}" "gtsam;gtsam_unstable;${Boost_PROGRAM_OPTIONS_LIBRARY}")
+gtsamAddExamplesGlob("*.cpp" "${excluded_examples}" "gtsam;${Boost_PROGRAM_OPTIONS_LIBRARY}")
```
varunagrawal, line 23: There should be an if-else condition here based on `GTSAM_BUILD_UNSTABLE`.

gtsam/hybrid/HybridSmoother.cpp
```diff
@@ -27,7 +27,9 @@ namespace gtsam {
 Ordering HybridSmoother::getOrdering(
     const HybridGaussianFactorGraph &newFactors) {
   HybridGaussianFactorGraph factors(hybridBayesNet());
-  factors += newFactors;
+  for(auto& factor: newFactors){
```
varunagrawal, line 30: I don't know why the original didn't work, but you could alternatively do `factors.push_back(newFactors)`. The `push_back` should be much more efficient than looping thanks to `std` magic.

gtsam/base/Matrix.cpp
```diff
@@ -30,6 +30,7 @@
 #include <fstream>
 #include <limits>
 #include <iostream>
+#include <iterator>
```
varunagrawal, line 33: Can you please explain the need for this header and other new headers? Seems strange it is needed now in such a core file.

varunagrawal (commented):
Please note that we don't build the examples in the CI (look at the `test()` function in `unix.sh`) so CI will succeed.

examples/CMakeLists.txt
varunagrawal: Here's what I would do:
```cmake
if (GTSAM_BUILD_UNSTABLE)
  set(GTSAM_UNSTABLE_LIB "gtsam_unstable")
else()
  set(GTSAM_UNSTABLE_LIB "")
endif (GTSAM_BUILD_UNSTABLE)
gtsamAddExamplesGlob("*.cpp" "${excluded_examples}" "gtsam;${GTSAM_UNSTABLE_LIB};${Boost_PROGRAM_OPTIONS_LIBRARY}")
```

ankur-rc: Frank and I discussed and we moved examples that depend on `gtsam_unstable` to `gtsam_unstable/examples`.

gtsam/hybrid/HybridSmoother.cpp
ankur-rc: Thanks @varunagrawal I wasn't aware, I'll change it to what you suggested.

gtsam/base/Matrix.cpp
ankur-rc: Our system failed to compile without the header.

dellaert:
@ankur please merge in master before pushing the one change. That will make CI succeed thanks to @varunagrawal

dellaert:
Also, because of @varunagrawal changes to CI, you’ll have to resolve a conflict in yml file

varunagrawal:
I believe @dellaert means develop and not master. Merging master will put your through quite a bit of pain 😅

dellaert (approved):
Awesome, will merge!

PR #1484, opened 2023-03-03
CI Updates

varunagrawal:
1. Install Boost via apt-get which saves us about 7 minutes.
2. Build special cases as shared libraries so we don't OOM.
3. Don't print warnings to CI, this should also help in saving time as well as readability of the output.
4. Improvements to the CI in general.


dellaert (approved):
Awesome! Will merge

PR #1486, opened 2023-03-06
cmake: explicitly specifying the linker language for gtsam

jlblancoc:
Integrating upstream from: https://salsa.debian.org/science-team/gtsam/-/blob/master/debian/patches/0004-I-m-explicitly-specifying-the-linker-language-for-gt.patch

cc: original author @dkogan


ProfFan (approved):

dellaert:
Cool, will merge...

PR #1487, opened 2023-03-06
Allow using system pybind11 if it exists.

jlblancoc:
Integrated upstream from https://salsa.debian.org/science-team/gtsam/-/blob/master/debian/patches/0003-Using-the-system-pybind11.patch with the additional fix for the cmake policy.

cc: @dkogan


ProfFan (approved):

dellaert:
I’ll merge.

PR #1488, opened 2023-03-06
Go back to default shared libraries, and fix unhandled case.

jlblancoc:
Although I like and agree with the spirit of PR #1464, it broke the so-far expected GTSAM behavior of defaulting to shared libs if not instructed otherwise by the user.

This PR:
- Goes back to defaulting to shared libs by defaulting GTSAM_FORCE_SHARED_LIB to On.
- Handles one previously unhandled case: if BUILD_SHARED_LIBS is not defined, PR #1464 assumed that static libraries were ok. However, formerly, not defining BUILD_SHARED_LIBS was leading to shared libs instead (!), so let's warn the user to make the decision clear.

cc: @berndpfrommer this was the source of the mistery new builds as static libraries! :-)


dellaert:
Thanks, and sorry for the trouble! I did not foresee it when approving that PR. If @varunagrawal does not answer soon please feel free to make the choice you think is right. I will approve now, feel free to merge when you are satisfied. Thanks for being one of our resident cmake Gurus!!!!

dellaert (approved):

varunagrawal:
@jlblancoc yes go ahead and remove those changes. Since shared library building was turned off by default, it was causing OOM issues in the CI, particularly for the special cases. Now that you're defaulting to building shared libraries, we can remove that change since it is the default behavior.

tl;dr yes undo it, since that will be the default behavior thanks to this PR.

varunagrawal:
I hope you don't mind me doing it. That way CI should have run by the time you look at this PR again. :)

jlblancoc:
> I hope you don't mind me doing it. That way CI should have run by the time you look at this PR again. :)

Sure, thanks, go on and do it yourself, and feel free to merge once CI is hopefully happy again! :-) 

> Thanks for being one of our resident cmake Gurus!!!!

My pleasure :+1:

varunagrawal:
Looks like CI is happy. Merging.

PR #1493, opened 2023-03-10
GTSAM_CONCEPT_CHECK Fix

### LINKED ISSUE #1490: ISAM2 throws segfault in external libraries
## Description

With the recent `develop` of GTSAM, if the class definition in an external project has an ISAM2 member object, then the class throws a segmentation fault on construction.

For example, if we add ISAM2 to the class `Statics` in `gtdynamics` like below,
```cpp
class Statics : public Kinematics {
 protected:
  const StaticsParameters p_;  // overrides Base::p_
  gtsam::ISAM2 isam_;

 public:
```

running a test in `testStatics.run` like below will throw a segfault
```cpp
TEST(Statics, Construction) {
  Statics s1;
}
```

I see similar issues in my own downstream code so I am guessing it won't be long before others also encounter this problem.
My guess is that classes in GTSAM other than ISAM2 may also cause this because of the RCA detailed below.

## Root Cause Analysis

I am on MacOS Monterrey (M1 chip) with AppleClang version 14.0.0.14000029, and I haven't tested this on other machines so this might be an Apple specific issue, but it is an issue that needs to be addressed (since @dellaert prefers Apple silicon).

On inspection, it looks like the change that is causing this issue is 9a50233426dd5cc0e10ff513d66c1c1b7c3a6009 where the `GTSAM_CONCEPT_ASSERT` macro was updated to fix `extra-semi` warnings.
If I revert only that single line https://github.com/borglab/gtsam/commit/9a50233426dd5cc0e10ff513d66c1c1b7c3a6009#diff-3e02cab63180b3fbf323a1f46f34c773a3a8eed23d050e379035dc7c0f3b2781R22, I no longer see the segfault behavior so it is indeed the issue.

I am wondering if there is another way to fix the warning without causing the segfault behavior?

varunagrawal:
This PR fixes #1490.

Instead of `void*(concept)`, I use `static_assert(true, "")` as a no-op.

I also went ahead and removed extra semi-colons, cleaning up a bunch of warnings that popped up, and removed an unnecessary header from `PreintergratedCombinedParams` that throws errors when Boost is disabled.

### DIFF
--- a/CppUnitLite/Test.h
+++ b/CppUnitLite/Test.h
@@ -80,7 +80,7 @@ protected:
 #define TEST_UNSAFE(testGroup, testName)\
   class testGroup##testName##Test : public Test \
   { public: testGroup##testName##Test () : Test (#testName "Test", __FILE__, __LINE__, false) {} \
-            virtual ~testGroup##testName##Test () {};\
+            virtual ~testGroup##testName##Test () {} \
             void run (TestResult& result_) override;} \
     testGroup##testName##Instance; \
   void testGroup##testName##Test::run (TestResult& result_)
--- a/gtsam/base/concepts.h
+++ b/gtsam/base/concepts.h
@@ -19,7 +19,7 @@
 // This does something sensible:
 #define BOOST_CONCEPT_USAGE(concept) void check##concept()
 // These just ignore the concept checking for now:
-#define GTSAM_CONCEPT_ASSERT(concept) void*(concept)
+#define GTSAM_CONCEPT_ASSERT(concept) static_assert(true, "")
 #define GTSAM_CONCEPT_REQUIRES(concept, return_type) return_type
 #endif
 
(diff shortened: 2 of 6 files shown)


mattking-smith:
@varunagrawal does this PR fix the CI failures seen on this [`GPMP2` PR](https://github.com/borglab/gpmp2/pull/13#issue-1620398219)?

mattking-smith:
@dellaert @ankur-rc what is the ETA on merging this PR? It would be nice to be able to verify upgrades to `GPMP2` functionality with the current `GTSAM repo` instead of pulling this PR.

ProfFan (approved):

dellaert:
Thanks @varunagrawal 
FYI @ankur-rc

varunagrawal:
This PR has resolved all the CI errors in GTDynamics and GPMP2!

PR #1508, opened 2023-04-11
Fix Windows CI

varunagrawal:
This PR fixes the CI issue with Windows by changing from `/std:c++latest` to `/std:c++17` for Visual Studio.

From my online reading, `/std:c++latest` uses the latest standard for C++ on Windows and this may cause unexpected issues as we experienced. Setting to the C++17 standard explicitly makes things more predictable.


dellaert (approved):
Awesomesauce!!!!!

PR #1511, opened 2023-04-13
Fix wrong jacobian to line in Line3::transformTo

zubingtan:
In Line3::transformTo, the jacobian to `wL` should be 
```
if (Dline) {
   Dline->setIdentity();
   (*Dline)(3, 0) = -t[2];
   (*Dline)(2, 1) = t[2];
}
```
instead of 
```
if (Dline) {
  Dline->setIdentity();
  (*Dline)(0, 3) = -t[2];
  (*Dline)(1, 2) = t[2];
}
```

Maybe this is a  typo. The UT cannot find this error because the author used a simple all-zeros translation to test it, which also is fixed.


dellaert:
Thanks!!!

PR #1514, opened 2023-04-14
Windows fix python CMake for copying test files

mrw2ee:
For cmake version 3.22.1, existing code worked on Linux, but failed on Windows 10 (?!?).  Clarifying relative paths fixed the issue and worked on both systems.


varunagrawal (approved):
LGTM. CI passes so this looks good.

Issue #1503, opened 2023-03-27
Can the maintainer provide a clang-format file for better code consistency?


zubingtan:
I have found that there are several different coding styles in gtsam's repo at present. Can we provide a clang-format file to ensure the consistency of the code format? Such as 

.clang-format
```
BasedOnStyle: Google

BinPackArguments: false
BinPackParameters: false
ColumnLimit: 100
DerivePointerAlignment: false
IncludeBlocks: Preserve
PointerAlignment: Left
```

ProfFan:
That would be a pretty good opportunity for a PR :)

zubingtan:
Hi @dellaert . We have just merged the '.clang-format 'configuration file in https://github.com/borglab/gtsam/pull/1506, which will allow all developers to format code using the `clang-format` tool. In your opinion, will developers be forced to format their code in future PR reviews, or will a so-called readability review be required?

dellaert:
Every review to some extent reviews readability. Guideline for formatting is: only format the edits you make using “format selection” to cut down on review time and repo churn. If formatting entire files, make that a separate PR, perhaps to prep another PR. In general, though, it tends to obfuscate git-blame so it’s not necessarily desired.

dellaert:
Thanks for contributing!! I’ll close the issue.

PR #1517, opened 2023-04-19
add useLOST to triangulateSafe

dllu:
adds a way to use the new LOST triangulation in `triangulateSafe` by setting the `useLOST` parameter in `TriangulationParameters` accordingly.

By the way, what version of `clang-format` do people use? for some reason when I ran `clang-format -i gtsam/geometry/triangulation.h` it resulted in many changed lines (using `clang-format version 10.0.0-4ubuntu1`).

### DIFF
--- a/gtsam/geometry/geometry.i
+++ b/gtsam/geometry/geometry.i
@@ -1146,11 +1146,13 @@ class TriangulationParameters {
   bool enableEPI;
   double landmarkDistanceThreshold;
   double dynamicOutlierRejectionThreshold;
+  bool useLOST;
   gtsam::SharedNoiseModel noiseModel;
   TriangulationParameters(const double rankTolerance = 1.0,
                           const bool enableEPI = false,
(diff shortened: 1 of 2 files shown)


zubingtan:
If I understand correctly and as discussed in https://github.com/borglab/gtsam/issues/1503, we should only format the section you modify. If you are using VScode, you can select the section and click the right menu button and use `Format Section`.  cc @dellaert

akshay-krishnan:
Thanks @dllu , it would be great if you could also add a unit test for triangulateSafe with useLOST=True here: https://github.com/borglab/gtsam/blob/7a3b95551fc5efd319a52586aa8a9b4cf6d01b34/gtsam/geometry/tests/testTriangulation.cpp#L515

varunagrawal:
@dllu gentle reminder

dellaert (approved):
I'm just going to approve and merge.

PR #1520, opened 2023-04-25
Add missing option in the CMake for Pose2 Expmap

ProfFan:
This PR adds the missing option in CMake, the code exists already just not exposed.


varunagrawal (approved):
LGTM, but CI fails?

cmake/HandleGeneralOptions.cmake
```diff
@@ -30,6 +30,7 @@ option(GTSAM_ALLOW_DEPRECATED_SINCE_V43     "Allow use of methods/functions depr
 option(GTSAM_SUPPORT_NESTED_DISSECTION      "Support Metis-based nested dissection" ON)
 option(GTSAM_TANGENT_PREINTEGRATION         "Use new ImuFactor with integration on tangent space" ON)
 option(GTSAM_SLOW_BUT_CORRECT_BETWEENFACTOR "Use the slower but correct version of BetweenFactor" OFF)
+option(GTSAM_SLOW_BUT_CORRECT_EXPMAP        "Use slower but correct expmap"  ON)
```
varunagrawal, line 33: I think this should be OFF by default. That fixes the CI.

ProfFan: Yup, a lot of stuff in CI assumes otherwise...

varunagrawal:
Looks great! Merging.

PR #1522, opened 2023-04-29
Fix JacobianFactor and HessianFactor on platforms where Eigen::Index is not the same size as gtsam::Key  Eigen::Index is defined as std::ptrdiff_t (size_t), which is not always the same size as gtsam::Key (uint64)

ShuangLiu1992:
Fix JacobianFactor and HessianFactor on platforms where Eigen::Index is not the same size as gtsam::Key

Eigen::Index is defined as std::ptrdiff_t (size_t), which is not always the same size as gtsam::Key (uint64)

On platform where they are different, buildDamped for example will add Factors with wrong keys which dangles and cannot be eliminate, causing solver to fail.

Leaving tests unchanged since they have their own private definition using Dims = std::vector<Eigen::Index>; which doesn't break.

### DIFF
--- a/gtsam/linear/HessianFactor.cpp
+++ b/gtsam/linear/HessianFactor.cpp
@@ -38,7 +38,7 @@ using namespace std;
 namespace gtsam {
 
 // Typedefs used in constructors below.
-using Dims = std::vector<Eigen::Index>;
+using Dims = std::vector<Key>;
 
 /* ************************************************************************* */
 void HessianFactor::Allocate(const Scatter& scatter) {
--- a/gtsam/linear/HessianFactor.cpp
+++ b/gtsam/linear/HessianFactor.cpp
@@ -72,7 +72,7 @@ HessianFactor::HessianFactor() :
 
 /* ************************************************************************* */
 HessianFactor::HessianFactor(Key j, const Matrix& G, const Vector& g, double f)
-    : GaussianFactor(KeyVector{j}), info_(Dims{G.cols(), 1}) {
+    : GaussianFactor(KeyVector{j}), info_(Dims{static_cast<Key>(G.cols()), 1}) {
   if (G.rows() != G.cols() || G.rows() != g.size())
     throw invalid_argument(
         "Attempting to construct HessianFactor with inconsistent matrix and/or vector dimensions");
@@ -85,7 +85,7 @@ HessianFactor::HessianFactor(Key j, const Matrix& G, const Vector& g, double f)
 // error is 0.5*(x-mu)'*inv(Sigma)*(x-mu) = 0.5*(x'*G*x - 2*x'*G*mu + mu'*G*mu)
 // where G = inv(Sigma), g = G*mu, f = mu'*G*mu = mu'*g
 HessianFactor::HessianFactor(Key j, const Vector& mu, const Matrix& Sigma)
-    : GaussianFactor(KeyVector{j}), info_(Dims{Sigma.cols(), 1}) {
+    : GaussianFactor(KeyVector{j}), info_(Dims{static_cast<Key>(Sigma.cols()), 1}) {
   if (Sigma.rows() != Sigma.cols() || Sigma.rows() != mu.size())
     throw invalid_argument(
         "Attempting to construct HessianFactor with inconsistent matrix and/or vector dimensions");
@@ -99,7 +99,7 @@ HessianFactor::HessianFactor(Key j1, Key j2, const Matrix& G11,
     const Matrix& G12, const Vector& g1, const Matrix& G22, const Vector& g2,
     double f) :
     GaussianFactor(KeyVector{j1,j2}), info_(
-        Dims{G11.cols(),G22.cols(),1}) {
+        Dims{static_cast<Key>(G11.cols()),static_cast<Key>(G22.cols()),1}) {
   info_.setDiagonalBlock(0, G11);
   info_.setOffDiagonalBlock(0, 1, G12);
   info_.setDiagonalBlock(1, G22);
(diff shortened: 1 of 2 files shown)


varunagrawal:
Wow this is great. @dellaert you should probably also look at this.

gtsam/linear/HessianFactor.cpp
```diff
@@ -38,7 +38,7 @@ using namespace std;
 namespace gtsam {
 
 // Typedefs used in constructors below.
-using Dims = std::vector<Eigen::Index>;
+using Dims = std::vector<Key>;
```
ProfFan, line 41: This looks very suspicious.

gtsam/linear/HessianFactor.cpp
```diff
@@ -72,7 +72,7 @@ HessianFactor::HessianFactor() :
 
 /* ************************************************************************* */
 HessianFactor::HessianFactor(Key j, const Matrix& G, const Vector& g, double f)
-    : GaussianFactor(KeyVector{j}), info_(Dims{G.cols(), 1}) {
+    : GaussianFactor(KeyVector{j}), info_(Dims{static_cast<Key>(G.cols()), 1}) {
```
ProfFan, line 75: `Dims` is consumed by
```c++
    /// Construct from a container of the sizes of each block.
    template<typename CONTAINER>
    SymmetricBlockMatrix(const CONTAINER& dimensions, bool appendOneDimension = false) :
      blockStart_(0)
    {
      fillOffsets(dimensions.begin(), dimensions.end(), appendOneDimension);
      matrix_.resize(variableColOffsets_.back(), variableColOffsets_.back());
      assertInvariants();
    }
```
which is templated and will be inferred to ` SymmetricBlockMatrix(const vector<Key>&)`.

which is then
```c++
    template<typename ITERATOR>
    void fillOffsets(ITERATOR firstBlockDim, ITERATOR lastBlockDim, bool appendOneDimension)
    {
      variableColOffsets_.resize((lastBlockDim-firstBlockDim) + 1 + (appendOneDimension ? 1 : 0));
      variableColOffsets_[0] = 0;
      DenseIndex j=0;
      for(ITERATOR dim=firstBlockDim; dim!=lastBlockDim; ++dim) {
        variableColOffsets_[j+1] = variableColOffsets_[j] + *dim;
        ++ j;
      }
      if(appendOneDimension)
      {
        variableColOffsets_[j+1] = variableColOffsets_[j] + 1;
        ++ j;
      }
    }
```
where everything becomes `Eigen::DenseIndex` which should be the same as `Eigen::Index`.

ProfFan (changes requested):
Thank you for the contribution @ShuangLiu1992 ! However I think a few questions need to be answered before we merge this.

1. What is your target platform?
2. Can you check if you can add your target platform to the test matrix?
3. It seems to me that everything gets converted to `Eigen::DenseIndex` in the end. Are these not the same type on your platform?

gtsam/linear/HessianFactor.cpp
ShuangLiu1992: In the current develop branch, for HessianFactor and JacobianFactor, when you pass a gtsam::Key, their constructor will cast it to Eigen::Index. If someone is ```using gtsam::symbol_shorthand;``` or ```gtsam::Symbol```, these values might be larger than what Eigen::Index can hold.

```
static const size_t keyBits = sizeof(Key) * 8;
static const size_t chrBits = sizeof(unsigned char) * 8;
static const size_t indexBits = keyBits - chrBits;
static const Key chrMask = Key(UCHAR_MAX)  << indexBits; // For some reason, std::numeric_limits<unsigned char>::max() fails
static const Key indexMask = ~chrMask;
```

```
Symbol(unsigned char c, std::uint64_t j)
```


In the Macros.h of Eigen, you can find
```
#define EIGEN_DEFAULT_DENSE_INDEX_TYPE std::ptrdiff_t
typedef EIGEN_DEFAULT_DENSE_INDEX_TYPE Index;
```

For std::ptrdiff_t, it's not guaranteed to be 64 bit.
ref: https://en.cppreference.com/w/cpp/types/ptrdiff_t

Say for example 32bit ARM device or Emscripten. On these platform std::ptrdiff_t is smaller than from the definition in types.h
```
  typedef std::uint64_t Key;
```

It's not very straightforward to add these platforms for testing, as they are cross compiled and might not be able to run on the build system.

I haven't notcied any issue in ```SymmetricBlockMatrix.h``` but for safety it should be changed to gtsam::Key as well.

dellaert: I alos don't understand the change above. A dimension is not a key. And as Fan says, it will go to Eigen::Index.

gtsam/linear/JacobianFactor.cpp
```diff
@@ -40,8 +40,8 @@ using namespace std;
 namespace gtsam {
 
 // Typedefs used in constructors below.
-using Dims = std::vector<Eigen::Index>;
-using Pairs = std::vector<std::pair<Eigen::Index, Matrix>>;
+using Dims = std::vector<Key>;
+using Pairs = std::vector<std::pair<Key, Matrix>>;
```
dellaert, line 44: This here might have to be a Key, I agree, but not Dims.

ProfFan:
@ShuangLiu1992 Could you add a test case that demonstrate the problem on your platform, and we merge that first?

ShuangLiu1992:
> @ShuangLiu1992 Could you add a test case that demonstrate the problem on your platform, and we merge that first?

Hi I have opted to not use gtsam::symbol_shorthand or gtsam::Symbol to circumvent my problem. But I think the problem is well explained and fixes/tests should be pretty stragithforward without a runtime test.

dellaert:
@ProfFan do you know how to fix the type problem in the right way so we can land this?

ProfFan:
@dellaert Yes but no bandwidth right now, the old one is definitely not correct on non 64-bit architectures so we may merge this first to unblock @dkogan, and then determine the right type hierarchy.

ProfFan (approved):

PR #1523, opened 2023-04-30
Upgrade TBB

### LINKED ISSUE #1445: testSmartStereoProjectionFactorPP unstable
## Description

testSmartStereoProjectionFactorPP.cpp is unstable after PR #1442, i.e. fails in *some* workflows.

## Steps to reproduce

Unfortunately: run CI on #1442....

## Expected behavior

Should fail deterministically, or succeed on all platforms.

## Environment

Seems special cases workflow, especially TBB, fails.
https://github.com/borglab/gtsam/actions/runs/4099700133/jobs/7069846556

## Additional information

I turned GTSAM_SINGLE_TEST_EXE off as we look into it, we should turn it back ON if a PR fixes this issue.

I have no idea what's up, but SmartStereoProjectionFactorPP is a complex class that does some funky things, so my hunch is that it is there. For now I will exclude it from the CI.

varunagrawal:
This PR updates TBB to the default Ubuntu latest in the CI (2021.5 for Ubuntu 22.04). The effect is it fixes #1445 effectively meaning that issue we saw was most likely a bug in TBB 4.4.

### DIFF
--- a/.github/scripts/python.sh
+++ b/.github/scripts/python.sh
@@ -9,32 +9,13 @@ set -x -e
 # install TBB with _debug.so files
 function install_tbb()
 {
-  TBB_BASEURL=https://github.com/oneapi-src/oneTBB/releases/download
-  TBB_VERSION=4.4.5
-  TBB_DIR=tbb44_20160526oss
-  TBB_SAVEPATH="/tmp/tbb.tgz"
-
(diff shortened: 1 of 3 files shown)


dellaert (approved):
Awesome !

PR #1524, opened 2023-05-02
Wrap Unit3 methods with Jacobians

varunagrawal:
Exposing Jacobians for some Unit3 methods.


travisdriver (approved):
Added a couple more; everything looks good to me!

PR #1528, opened 2023-05-17
Added TableFactor, a discrete factor optimized for sparsity.

ywkim0606:
## Summary
TableFactor is a discrete probabilistic factor that can replace DecisionTreeFactor.
It should support all features that the current DecisionTreeFactor has such as multiplication, division, sum, and max.
It was developed for faster MAP inference on CSP and Classical Planning problems where discrete state space is very sparse.
Unittest includes a runtime benchmark between TableFactor and DecisionTreeFactor given various sparsity.
Supports cardinality up to 18446744073709551615.

### What is the solution?
Relies on Eigen3's SparseVector data structure. Multiplication and division use the tensor contraction method proposed in [SPARTA](https://dl.acm.org/doi/pdf/10.1145/3437801.3441581.).

### What areas of GTSAM does it impact?
gtsam/discrete

### How to test?
run provided unit test (testTableFactor.cpp) 

## Other Notes
In-depth implementation details are in [thesis](https://ywkim0606.github.io/files/Yoonwoo_thesis.pdf)'s chapter 4.

### DIFF
--- a/gtsam/discrete/TableFactor.h
+++ b/gtsam/discrete/TableFactor.h
@@ -1,0 +1,333 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file TableFactor.h
+ * @date May 4, 2023
+ * @author Yoonwoo Kim
+ */
+
+#pragma once
+
+#include <gtsam/discrete/DiscreteFactor.h>
+#include <gtsam/discrete/DiscreteKey.h>
+#include <gtsam/inference/Ordering.h>
+
+#include <Eigen/Sparse>
+#include <algorithm>
+#include <memory>
+#include <map>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace gtsam {
+
+  class SparseDiscreteConditional;
+  class HybridValues;
+
+  /**
+   * A discrete probabilistic factor optimized for sparsity.
+   *
+   * @ingroup discrete
+   */
+  class GTSAM_EXPORT TableFactor : public DiscreteFactor {
+   protected:
+    std::map<Key, size_t> cardinalities_;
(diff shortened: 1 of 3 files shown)


ywkim0606:
@dellaert

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,333 @@
...
+#include <vector>
+
+namespace gtsam {
+
+  class SparseDiscreteConditional;
+  class HybridValues;
+
+  /**
+   * A discrete probabilistic factor optimized for sparsity.
+   *
```
dellaert, line 40: Please explain in some detail the approach and refer to member variables below.

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,333 @@
...
+   * @ingroup discrete
+   */
+  class GTSAM_EXPORT TableFactor : public DiscreteFactor {
+   protected:
+    std::map<Key, size_t> cardinalities_;
```
dellaert, line 45: Provide at least one line comments that go with class documentaion
```
std::map<Key, size_t> cardinalities_; ///< blah...
```

dellaert (commented):
Some comments

dellaert:
I don't see any changes yet - send me an email when I should start CI again.

gtsam/discrete/TableFactor.h
ywkim0606: I added:
Uses sparse_table_ to store only the non-zero probabilities.
Computes the assigned value for the key using the ordering which the non-zero probabilities are stored in.

gtsam/discrete/TableFactor.h
```diff
@@ -0,0 +1,333 @@
...
+    // typedefs needed to play nice with gtsam
+    typedef TableFactor This;
+    typedef DiscreteFactor Base;  ///< Typedef to base class
+    typedef std::shared_ptr<TableFactor> shared_ptr;
+    typedef Eigen::SparseVector<double>::InnerIterator SparseIt;
+    typedef std::vector<std::pair<DiscreteValues, double>> AssignValList;
+    using Binary = std::function<double(const double, const double)>;
+
+   public:
+     /** The Real ring with addition and multiplication */
```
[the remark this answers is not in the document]
ywkim0606: Its in AlgebraicDecisionTree. I didn't want to include it into TableFactor but I we can use the one implemented in AlgebraicDecisionTree.

PR #1533, opened 2023-05-31
Fix geometry wrapper

varunagrawal:
I introduced a bug in the Matlab wrapper when wrapping code in #1524.
This PR fixes it and adds an additional matlab test.


mattking-smith (approved):
LGTM

mattking-smith (approved):
LGTM

PR #1534, opened 2023-06-02
LOST Triangulation Underconstrained

travisdriver:
I encountered an instance where the iterated DLT solution would correctly triangulate an input point, while the LOST algorithm would return NaNs. 

This seems to be related to a rank deficiency in the QR factorization. The rank tolerance was being set for the DLT methods, but it was not being set for LOST and was instead being [computed automatically by Eigen](https://eigen.tuxfamily.org/dox/classEigen_1_1ColPivHouseholderQR.html#af2983fc27ce1fe5a791fc998f9e598ce). 

Therefore, I added the `rank_tol` variable to the `triangulateLOST` method so that the user could explicitly set the tolerance. However, this change didn't seem to fix the rank deficiency issue. 

Below is code to reproduce the issue. The measurements and camera data were derived from images and ancillary data from the Hayabusa mission to Asteroid 25143 Itokawa.

```python
import numpy as np
import gtsam

from read_write_model import read_model

cameras, images, points3d = read_model("lost_underconstrained_data")

noisemodel = gtsam.noiseModel.Isotropic.Sigma(2, 1e-4)
error = []
for jj in points3d.keys():
    if len(points3d[jj].image_ids) < 5: 
        continue
    kpt_vec, cam_vec = [], []
    for idx, kk in enumerate(points3d[jj].image_ids):
        # Add keypoint measurements.
        pp = points3d[jj].point2D_idxs[idx];  # keypoint index
        kpt_vec.append(gtsam.Point2(images[kk].xys[pp].T.astype("double")))

        # Add cameras.
        R_BC = images[kk].qvec2rotmat().T
        r_CB_B = -R_BC @ images[kk].tvec[..., None]
        fx, fy, cx, cy = cameras[images[kk].camera_id].params
        cam_vec.append(
            gtsam.PinholeCameraCal3_S2(
                gtsam.Pose3(gtsam.Rot3(R_BC), gtsam.Point3(r_CB_B.flatten())),
                  gtsam.Cal3_S2(fx, fy, 0.0, cx, cy),
            )
        )

    # Triangulate.        
    kpt_vec = gtsam.Point2Vector(kpt_vec)
    cam_vec = gtsam.CameraSetCal3_S2(cam_vec)
    xyz_est = gtsam.triangulatePoint3(cam_vec, kpt_vec, rank_tol=1e-9, optimize=False, model=noisemodel, useLOST=True)
    # xyz_est = gtsam.triangulatePoint3(cam_vec, kpt_vec, rank_tol=1e-9, optimize=True, model=noisemodel)
    error.append(np.linalg.norm(points3d[jj].xyz - xyz_est))
    print(xyz_est)

print("Mean error:", np.mean(error))
print(error)
```

Data in COLMAP format: 
[lost_underconstrained_data.zip](https://github.com/borglab/gtsam/files/11639889/lost_underconstrained_data.zip)

The `read_write_model` module is defined [here](https://github.com/colmap/colmap/blob/dev/scripts/python/read_write_model.py).

### DIFF
--- a/gtsam/geometry/triangulation.cpp
+++ b/gtsam/geometry/triangulation.cpp
@@ -25,8 +25,8 @@
 namespace gtsam {
 
 Vector4 triangulateHomogeneousDLT(
-    const std::vector<Matrix34, Eigen::aligned_allocator<Matrix34>>& 
-    projection_matrices,
+    const std::vector<Matrix34, Eigen::aligned_allocator<Matrix34>>&
+        projection_matrices,
     const Point2Vector& measurements, double rank_tol) {
   // number of cameras
   size_t m = projection_matrices.size();
@@ -56,8 +56,8 @@ Vector4 triangulateHomogeneousDLT(
 }
 
 Vector4 triangulateHomogeneousDLT(
-    const std::vector<Matrix34, Eigen::aligned_allocator<Matrix34>>& 
-    projection_matrices,
+    const std::vector<Matrix34, Eigen::aligned_allocator<Matrix34>>&
+        projection_matrices,
     const std::vector<Unit3>& measurements, double rank_tol) {
   // number of cameras
   size_t m = projection_matrices.size();
@@ -68,7 +68,7 @@ Vector4 triangulateHomogeneousDLT(
   for (size_t i = 0; i < m; i++) {
     size_t row = i * 2;
     const Matrix34& projection = projection_matrices.at(i);
-    const Point3& p = 
+    const Point3& p =
         measurements.at(i)
             .point3();  // to get access to x,y,z of the bearing vector
 
@@ -130,7 +130,7 @@ Point3 triangulateLOST(const std::vector<Pose3>& poses,
     // Note: Setting q_i = 1.0 gives same results as DLT.
     const double q_i = num_i / (measurementNoise->sigma() * den_i);
 
-    const Matrix23 coefficientMat = 
+    const Matrix23 coefficientMat =
         q_i * skewSymmetric(calibratedMeasurements[i]).topLeftCorner(2, 3) *
         wTi.rotation().matrix().transpose();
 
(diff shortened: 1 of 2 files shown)


akshay-krishnan (approved):
Thank you for catching and fixing this!
On the gtsam repo, it is actually not encouraged to format entire files. Its also hard for reviewers to spot the relevant change. Other than setting the rank tolerance for Eigen, did I miss anything? 

If you have a small example with only 2 cameras where this happens, it might be worth adding a unit test too.

travisdriver:
The issue was due to the baseline vector coinciding with the jth measurement vector (`wZj`), which caused the denominator to be zero. I added a simple check that loops through other values for j if this occurs.

gtsam/geometry/triangulation.cpp
```diff
@@ -95,16 +95,37 @@ Point3 triangulateLOST(const std::vector<Pose3>& poses,
...
-
-    const Point3 wZi = wTi.rotation().rotate(calibratedMeasurements[i]);
-    const Point3 wZj = wTj.rotation().rotate(calibratedMeasurements[j]);
+    Point3 d_ij = wTj.translation() - wTi.translation();
+    Point3 wZi = wTi.rotation().rotate(calibratedMeasurements[i]);
+    Point3 wZj = wTj.rotation().rotate(calibratedMeasurements[j]);
+    double num_i = wZi.cross(wZj).norm();
+    double den_i = d_ij.cross(wZj).norm();
+
+    // Handle q_i = 0 (or NaN), which arises if the measurement vectors, wZi and wZj, coincide
```
akshay-krishnan, line 107: is this line exceeding the max line length?

akshay-krishnan (commented):
could you revert the formatting changes from the first commit?

travisdriver:
> could you revert the formatting changes from the first commit?

Reverted all of the formatting.

varunagrawal:
Hi @travisdriver. Sorry for the extra work, but can you please add a unit test as well. You can copy-paste the post-processed values from the Hayabusa mission to serve as your test fixture. This will allow us to maximize code coverage and guarantee correctness in the future as well.

dellaert:
I'm going to merge - a unit test would be nice but we have those in GTSFM, and the code does have a set of unit tests already

PR #1542, opened 2023-06-10
Decision Tree Improvements

varunagrawal:
1. `Unique` is now bottom-up, since I found another test case where it was incorrect and this fixes it.
2. Print the number of assignments inside square brackets when printing a `DecisionTree`.
3. Change `GTSAM_DT_NO_PRUNING` to `GTSAM_DT_MERGING` so it disambiguates from Hybrid pruning.
4. Expose `GTSAM_DT_MERGING` via CMake.


gtsam/discrete/DecisionTree-inl.h
```diff
@@ -549,15 +544,15 @@ namespace gtsam {
   template<typename L, typename Y>
   template<typename Iterator> DecisionTree<L, Y>::DecisionTree(
       Iterator begin, Iterator end, const L& label) {
-    root_ = compose(begin, end, label);
+    root_ = Choice::Unique(compose(begin, end, label));
```
dellaert, line 547: Why do all these have a Unique now? That was not detailed in the PR comment

dellaert (commented):
LGTM but see my question. email me when I should look again

gtsam/discrete/DecisionTree-inl.h
varunagrawal: Okay I undid that. Looks like it was a remnant of debugging.

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -199,47 +200,41 @@ namespace gtsam {
     }
 
     /// If all branches of a choice node f are the same, just return a branch.
```
varunagrawal, line 202: TODO(Varun): Elaborate more on how this works.

varunagrawal:
@dellaert so I am really struggling with implementing this function correctly in the decision tree
```cpp
// Applying binary operator to two leaves results in a leaf
  NodePtr apply_g_op_fL(const Leaf& fL, const Binary& op) const override {
    // fL op gL
    // TODO(Varun) Is the nrAssignments setting correct?
    NodePtr h(new Leaf(op(fL.constant_, constant_), fL.nrAssignments()));
    return h;
  }
```

I only need `nrAssignments` for pruning in the `DecisionTreeFactor`, so I replaced that with a loop over all the assignments in the cartesian product (which is a bit slower)
```
auto allValues = DiscreteValues::CartesianProduct(this->discreteKeys());
for (auto&& val : allValues) {
  double prob = (*this)(val);
  probabilities.push_back(prob);
}
```

It seems that there is no time improvement, since the test case is now taking 131.93 seconds compared to 117 seconds with the cartesian product taking up 12.8236 seconds (which accounts for the time difference).

My conclusion is that perhaps the `DecisionTreeFactor` is inherently non-scalable, which means I should perhaps move to the `TableFactor`.

varunagrawal:
I am contemplating removing the `nrAssignments` variable since I introduced it just for pruning and it is not being correctly computed. :(

PR #1545, opened 2023-06-15
4.2: Fix an invalid include, restore C++11 compatibility

valgur:
C++11 compatibility in 4.2a9 is currently broken due to this single minor bug.

Cropped up in https://github.com/conan-io/conan-center-index/pull/17933


varunagrawal (approved):
Huh interesting that this was missed. Thanks for catching it @valgur. LGTM

@dellaert we might need to update the 4.2 release file.

dellaert:
Thanks, will merge !

PR #1546, opened 2023-06-15
Pybind11 STL Support

varunagrawal:
This PR adds full STL support to the Python wrapper. The major benefit is that this will make the wrapper behave in a more "pythonic" fashion.

E.g.
Instead of
```python
poses = gtsam.Pose3Vector([pose1, pose2])
cameras = gtsam.CameraSetCal3Fisheye([camera1, camera2])
measurements = gtsam.Point2Vector([k.project(origin) for k in cameras])
```

we can now do
```python
poses = [pose1, pose2]
cameras = [camera1, camera2]
measurements = [k.project(origin) for k in cameras]
```

This will also lead to improved reflection capabilities (e.g. when calling `type()` on an object) and other benefits.


python/gtsam/preamble/geometry.h
```diff
@@ -10,14 +10,8 @@
...
-PYBIND11_MAKE_OPAQUE(
-    std::vector<gtsam::Point2, Eigen::aligned_allocator<gtsam::Point2>>);
-PYBIND11_MAKE_OPAQUE(gtsam::Point2Pairs);
-PYBIND11_MAKE_OPAQUE(gtsam::Point3Pairs);
-PYBIND11_MAKE_OPAQUE(gtsam::Pose2Pairs);
-PYBIND11_MAKE_OPAQUE(gtsam::Pose3Pairs);
-PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Pose3>);
-PYBIND11_MAKE_OPAQUE(gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3Bundler>>);
 PYBIND11_MAKE_OPAQUE(gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3_S2>>);
+PYBIND11_MAKE_OPAQUE(gtsam::CameraSet<gtsam::PinholeCamera<gtsam::Cal3Bundler>>);
```
johnwlambert, line 15: nit: could we add a small comment here explaining why List's can be converted automatically, but Set's cannot?

python/gtsam/tests/test_DsfTrackGenerator.py
```diff
@@ -81,16 +81,16 @@ def test_track_generation(self) -> None:
...
 
-    def test_sfm_track_2d_constructor(self) -> None:
-        """Test construction of 2D SfM track."""
-        measurements = SfmMeasurementVector()
-        measurements.append((0, Point2(10, 20)))
-        track = SfmTrack2d(measurements=measurements)
-        track.measurement(0)
-        assert track.numberMeasurements() == 1
+#     def test_sfm_track_2d_constructor(self) -> None:
+#         """Test construction of 2D SfM track."""
```
johnwlambert, line 88: nit: why comment these out?

johnwlambert (approved):
Thanks Varun!

python/gtsam/tests/test_DsfTrackGenerator.py
```diff
@@ -81,16 +81,16 @@ def test_track_generation(self) -> None:
         np.testing.assert_allclose(track2.indexVector(), [1, 2])
 
 
-class TestSfmTrack2d(GtsamTestCase):
-    """Tests for SfmTrack2d."""
+# class TestSfmTrack2d(GtsamTestCase):
```
akshay-krishnan, line 84: why is this commented out?

python/gtsam/tests/test_Sim2.py
```diff
@@ -46,7 +46,7 @@ def test_align_poses_along_straight_line(self) -> None:
 
         wToi_list = [wTo0, wTo1, wTo2]
 
-        we_pairs = Pose2Pairs(list(zip(wToi_list, eToi_list)))
+        we_pairs = list(zip(wToi_list, eToi_list))
```
akshay-krishnan, line 49: does pair translate to list? or tuple? or either?

akshay-krishnan (approved):
This is great, thanks Varun!

python/gtsam/preamble/geometry.h
varunagrawal: Sets can be, and `CameraSet` is actually derived from `std::vector` (and poorly named IMO). Let me update the `test_Triangulation.py` file tests and see if I can get rid of these opaque markers.

python/gtsam/tests/test_Sim2.py
varunagrawal: Pair should now be a tuple.

Issue #1163, opened 2022-04-09
More basis wrapper


dellaert:
## Feature

Expand the usefulness of the wrapper for `basis` module.

## Motivation

`basis.i` is missing many things:
- `FitBasis` cannot take an interval [a,b]
- For BasisFactors, only M=3,4,12 is supported
- For BasisFactors, only `Chebyshev2` is supported 

## Pitch

- fix FitBasis to take [a,b] and expose in wrapper
- make BasisFactors *dynamic*, if that is possible, then all M are supported
- add templates for all currently implemented bases

varunagrawal:
@dellaert to make `BasisFactors` dynamic, I would have to change `ParameterMatrix` to not be templated. When we discussed this design choice, we chose to make ParameterMatrix templated so we would get the compile time checking, but this comes at the cost of flexibility in the wrapper.

dellaert:
I'm ok with dynamic

dellaert:
I think this was done.

PR #1548, opened 2023-06-20
Make Basis factors dynamic

varunagrawal:
This PR removes the templating from all the Basis factors and functors, instead opting to pass the erstwhile template parameter as a constructor argument.

This will allow a user to define factors for arbitrarily sized vectors and not be limited to what template parameters are specified in the `basis.i` file.

Fixes almost everything in #1163 except for `FitBasis cannot take an interval [a,b]`. This last part I still need to understand how to do (and may not be high up on my current TODO list 😞).

### DIFF
--- a/gtsam/basis/Basis.cpp
+++ b/gtsam/basis/Basis.cpp
@@ -1,0 +1,33 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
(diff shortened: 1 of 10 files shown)


dellaert (approved):
This is awesome! Wrapper is much cleaner now and supports all dimensions. In another PR, we could get rid of ParameterMatrix altogether now. Simplicity FTW!

PR #1549, opened 2023-06-20
Python Wrapper Backwards Compatibility

varunagrawal:
Added typedefs to the `gtsam/__init__.py` file and reverted relevant parts of the python tests to ensure backwards compatibility of the python wrapper in the face of the new pybind11 STL support.

### DIFF
--- a/python/gtsam/__init__.py
+++ b/python/gtsam/__init__.py
@@ -4,9 +4,49 @@
 
 import sys
 
+from gtsam.utils import findExampleDataFile
+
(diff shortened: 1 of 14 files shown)


dellaert:
All right! Now, you're going to kill me, but I think that - since people use unit tests as guidelines and copy/paste from them - all those unit tests should use the *new* style. Instead of reverting these tests, how about adding a one "backwards compatibility" test that asserts all the uses in these tests?

dellaert (approved):
Great !!!

PR #1554, opened 2023-06-28
Add explanation of the StereoPoint2 constructor

IshitaTakeshi:
The constructor arguments of StereoPoint2 was unclear for me so I added the explanation.
It would be better to say the assumption that the stereo images are rectified.


varunagrawal (approved):
Awesome. Thanks @IshitaTakeshi!

PR #1556, opened 2023-06-28
TableFactor Improvements

varunagrawal:
- Adds a new constructor to the `TableFactor` which takes a `DiscreteConditional`. The API for the TableFactor should now mimic that of the `DecisionTreeFactor`.
- Move `cardinalities_` and `discreteKeys()` to the `DiscreteFactor` class to reduce code duplication.
- Renamed all "tree" based variables in hybrid elimination to names which are agnostic of the underlying factor type. **No change in functionality**.

### DIFF
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -33,16 +33,13 @@ namespace gtsam {
 
   /* ************************************************************************ */
   DecisionTreeFactor::DecisionTreeFactor(const DiscreteKeys& keys,
-                                        const ADT& potentials)
-      : DiscreteFactor(keys.indices()),
-        ADT(potentials),
-        cardinalities_(keys.cardinalities()) {}
+                                         const ADT& potentials)
+      : DiscreteFactor(keys.indices(), keys.cardinalities()), ADT(potentials) {}
 
   /* ************************************************************************ */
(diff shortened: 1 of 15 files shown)


dellaert:
@varunagrawal this PR shows changes in 15 files, just for adding a constructor? Are you sure it targets the right branch?

varunagrawal:
I need to update the PR comments. I also moved a lot of common functionality up to the DisreteFactor class to keep it DRY.

dellaert:
> I need to update the PR comments. I also moved a lot of common functionality up to the DisreteFactor class to keep it DRY.

Hmm, yes, but TableFactor is a big change, and I'd rather review that in a separate PR if possible.

varunagrawal:
I need to update the PR comments. I also moved a lot of common functionality up to the DisreteFactor class to keep it DRY.

varunagrawal:
Using TableFactor in elimination will be a separate PR for sure.

varunagrawal:
@dellaert I've updated the PR comment to better reflect the changes. Please note that there is no functionality change in hybrid elimination in this PR, only variable name changes.

dellaert:
OK. Any reason this is not targeting develop?

varunagrawal:
> OK. Any reason this is not targeting develop?

I chained PRs together since I wanted functionality from different PRs and was not getting review responses to unblock me.

PR #1558, opened 2023-06-29
Overload `+=`, `,` operator

varunagrawal:
Overloaded the `+=` operator in the `Ordering` class so we can append keys as:
```cpp
Ordering o;
o += key1;
```

Also overloaded the comma `,` operator, so we can append in a style similar to Boost's Assign module which was the previous way:
```cpp
Ordering o;
o += key1, key2, key3;
```

Similarly overloaded operators for `FactorGraph`. Added relevant unit tests for both of these classes.

### DIFF
--- a/gtsam/base/Group.h
+++ b/gtsam/base/Group.h
@@ -22,13 +22,6 @@
 
 #include <gtsam/base/Testable.h>
 
-#ifdef GTSAM_USE_BOOST_FEATURES
-#include <boost/concept_check.hpp>
-#include <boost/concept/requires.hpp>
-#include <boost/type_traits/is_base_of.hpp>
-#include <boost/static_assert.hpp>
-#endif
-
 #include <utility>
 
 namespace gtsam {
--- a/gtsam/inference/FactorGraph.h
+++ b/gtsam/inference/FactorGraph.h
@@ -29,10 +29,6 @@
 
 #include <Eigen/Core>  // for Eigen::aligned_allocator
 
-#ifdef GTSAM_USE_BOOST_FEATURES
-#include <boost/assign/list_inserter.hpp>
-#endif
-
 #ifdef GTSAM_ENABLE_BOOST_SERIALIZATION
 #include <boost/serialization/nvp.hpp>
 #include <boost/serialization/vector.hpp>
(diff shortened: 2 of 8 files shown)


gtsam/inference/FactorGraph.h
```diff
@@ -53,45 +49,6 @@ class BayesTree;
 
 class HybridValues;
 
-/** Helper */
```
varunagrawal, line 56: Got rid of all of these since they were needed for `list_inserter` only.

gtsam/base/Group.h
```diff
@@ -22,13 +22,6 @@
 
 #include <gtsam/base/Testable.h>
 
-#ifdef GTSAM_USE_BOOST_FEATURES
```
varunagrawal, line 25: Did some cleaning up.

dellaert (commented):
Very cool, but before I approve, is it still backwards compatible with boost? The list inserter pattern was in our code for a long time and, while we eradicated, users might have copied it in their code.

varunagrawal:
@dellaert it should be backwards compatible with boost. From my understanding, `list_inserter` is a way to reconcile all the different "insert" methods across STL containers (insert vs push_back vs something else).

That being said, since we're killing boost in the current version, users would anyway have to update their code for this new release and this is just 1 extra, albeit small, change.

dellaert:
> @dellaert it should be backwards compatible with boost. From my understanding, `list_inserter` is a way to reconcile all the different "insert" methods across STL containers (insert vs push_back vs something else).
> 
> That being said, since we're killing boost in the current version, users would anyway have to update their code for this new release and this is just 1 extra, albeit small, change.

Can you verify it is? We are not killing boost, we are allowing people to compile without boost. Users that are OK with boost should not have to update their code, as I remember.

varunagrawal:
@dellaert I ran all unit tests with `GTSAM_USE_BOOST_FEATURES` set to both ON and OFF and they all pass.

dellaert (commented):
Cool, good additions !!!

dellaert (approved):

dellaert:
@varunagrawal please merge and delete branch at your leisure.

PR #1565, opened 2023-07-01
Cleaner ci

talregev:
Cleaner ci

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -1,6 +1,12 @@
 name: Linux CI
 
-on: [pull_request]
+on: 
+  push:
+    branches: 
+      - develop
+  pull_request:
+    branches: 
+      - "*" # Pull request for all branches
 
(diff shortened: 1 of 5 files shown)


varunagrawal:
I don't think this is necessarily cleaner. Also we have special actions for push to develop which involve building the nightly packages.

talregev:
> I don't think this is necessarily cleaner. Also we have special actions for push to develop which involve building the nightly packages.

It easier to develop and add new functionality when the CI is cleaner. (The same work for code).
I can bring the old functionality and leave it as only cleaner yaml code.
We can discuss it on different PR if you open minded to hear my side.
If you insist. I can close this PR.

varunagrawal:
Fair point but the CI isn't the focus of improvement right now. Also since we're open source, we don't get paid CI instances so it's a premium resource.

We prefer to run all tests on a local machine first to get things working and then push a PR for a collective check.

Hope that makes sense. :)

talregev:
I am not sure which of the option I suggested you wanted. For now I am closing this PR.

PR #1567, opened 2023-07-03
Handle latest tbb

talregev:
Handle latest tbb

@cdb0y511
Let me know if it work for you.


ProfFan:
We should determine whether we still want to ship our own `FindTBB.cmake` or default to the official (by not using our own) implementation. @dellaert @jlblancoc @varunagrawal as we retire support for old Ubuntu

talregev:
@ProfFan which version is old ubuntu? which are you currently support?

jlblancoc:
@ProfFan As far as I can see [1], not even the latest cmake-data package (next Ubuntu 23.10) comes with a FindTbb.cmake file... :-S

[1] https://packages.ubuntu.com/mantic/all/cmake-data/filelist

talregev:
As far I understand it not find it with FindTbb module. It doing something else.
Even at their source code there isn't `FindTbb.cmake` as you can see here:
https://gitlab.kitware.com/cmake/cmake/-/tree/master/Modules

dellaert:
@ProfFan in light of comments, will you approve?

ProfFan:
@talregev @jlblancoc

> @ProfFan As far as I can see [1], not even the latest cmake-data package (next Ubuntu 23.10) comes with a FindTbb.cmake file... :-S
> 
> [1] https://packages.ubuntu.com/mantic/all/cmake-data/filelist

This is because it's in the `libtbb-dev` package instead: https://packages.ubuntu.com/focal/amd64/libtbb-dev/filelist . As you can see it exists even in `focal` (20.04).

@dellaert 
I would personally support this change, on macOS Hombrew provides this CMake file as well. Also we can remove `FindTBB.cmake` from `cmake` folder as well.

ProfFan (approved):

dellaert:
Cool, thanks all! I'll merge.

jlblancoc:
> This is because it's in the `libtbb-dev` package instead: https://packages.ubuntu.com/focal/amd64/libtbb-dev/filelist . As you can see it exists even in `focal` (20.04).

Ah, that explains everything. Well done.

PR #1572, opened 2023-07-10
Enable GNC Optimizer

varunagrawal:
~Added our own implementation of Halley's Iteration to solve for the chi square inverse function. This allows us to use the `Chi2inv` function without relying on Boost, thus enabling the use of GNC as an optimizer again.~

~I need some feedback on the following points:~
~1. I have named the file with Halley's iteration as `GncHelpers.h`. Should I rename this to something else that is more generic? Most of the code is for computing the gamma distribution inverse, which is used to compute the chi square inverse.~
~2. Should the above file be within `nonlinear/internal`? Currently it is in `nonlinear`.~
~3. Any documentation or other aspects that would be needed? I don't expect this code to be user facing and I would welcome replacing this function with another library that is better tested and documented.~

Implement `chi_squared_quantile` (aka the `chi2inv` function) using the incomplete gamma inverse function from `cephes`.

I added `cephes` to `3rdparty` and added the necessary CMake files to build and link against GTSAM. This makes the change for us really simple.


lucacarlone:
@varunagrawal : this looks excellent. I'll leave the question of (re)naming of GncHelper to @dellaert . My only comment/suggestion would be to add a few more unit tests. This is a quite basic functionality and it we get it wrong, it might be very hard to debug in downstream applications.

dellaert (commented):
Awesome!. I'm fine with the location of the file. Agreed with Luca on unit tests, but my main request is to add some information in GncHelpers as to where this code comes from? Is this de-novo code? Was it copied from boost? Also, if it only implements chi-square, why not name it as such? Finally, this is not a class I think so the file should start with a lowercase letter.

varunagrawal:
I looked at various implementations from both Boost and GCEM to understand how Halley's algorithm worked and how to deal with different datatypes effectively, but I wrote this pretty much on my own.

After some thought, I am considering these changes:
1. Rename `GncHelpers` to `chiSquaredInverse.hpp`.
2. Move it to `nonlinear/internal` since it is mostly for internal usage.
3. Unit test each individual function. I read up on how Boost does this and will do similar things.
4. Improve docs as I go along.

dellaert:
> I looked at various implementations from both Boost and GCEM to understand how Halley's algorithm worked and how to deal with different datatypes effectively, but I wrote this pretty much on my own.
> 
> After some thought, I am considering these changes:
> 
> 1. Rename `GncHelpers` to `chiSquaredInverse.hpp`.
> 2. Move it to `nonlinear/internal` since it is mostly for internal usage.
> 3. Unit test each individual function. I read up on how Boost does this and will do similar things.
> 4. Improve docs as I go along.

Sounds good. Please do document the above. Having straight copies of code from boost where possible would be preferable as those functions would not have to be unit-tested, and it is acceptable under their license.

dellaert:
...with attribution, of course

dellaert (approved):
Thanks !!!

varunagrawal:
I wanted to give a quick update on this:

1. Adding code from Boost was getting really unwieldy, resulting in copy-paste of over a dozen files for computing the incomplete gamma function (this is used to compute the chi-squared inverse easily). This code was also getting far too complicated for long-term maintenance.
2. I did some looking around and found [cephes](https://www.netlib.org/cephes/) which gives us everything we need in a really lightweight C library. Moreover `scipy` uses `cephes` for their implementation of the incomplete gamma inverse function so it is production grade.
3. Cephes is released under a very liberal license which makes bundling it under `3rdparty` really easy.
4. As a result, I added `cephes` to `3rdparty` and got it compiling. The overall PR now is much simpler with much of the work being offloaded to `cephes`.

lucacarlone (approved):
LGTM!

varunagrawal:
Cephes license can be found here: https://smath.com/en-US/view/CephesMathLibrary/license

PR #1575, opened 2023-07-12
Hybrid Elimination Improvements

varunagrawal:
- Override `apply` with `UnaryAssignment` for `DecisionTreeFactor`.
- Improve discrete elimination by eliminating the joint distribution and re-segregating back into original conditionals.
- Make hybrid code use common parent `DiscreteFactor` where applicable.
- Update the eliminate method to check the factors as mentioned in @dellaert's TODO.
- Templetize methods in `Switching.h` to remove duplication.

### DIFF
--- a/gtsam/discrete/DecisionTree.h
+++ b/gtsam/discrete/DecisionTree.h
@@ -300,6 +300,42 @@ namespace gtsam {
     size_t nrLeaves() const;
 
     /**
+     * @brief This is a convenience function which returns the total number of
+     * leaf assignments in the decision tree.
+     * This function is not used for anymajor operations within the discrete
+     * factor graph framework.
+     *
+     * Leaf assignments represent the cardinality of each leaf node, e.g. in a
+     * binary tree each leaf has 2 assignments. This includes counts removed
+     * from implicit pruning hence, it will always be >= nrLeaves().
+     *
+     * E.g. we have a decision tree as below, where each node has 2 branches:
+     *
+     * Choice(m1)
+     * 0 Choice(m0)
+     * 0 0 Leaf 0.0
+     * 0 1 Leaf 0.0
+     * 1 Choice(m0)
+     * 1 0 Leaf 1.0
+     * 1 1 Leaf 2.0
+     *
+     * In the unpruned form, the tree will have 4 assignments, 2 for each key,
+     * and 4 leaves.
+     *
+     * In the pruned form, the number of assignments is still 4 but the number
+     * of leaves is now 3, as below:
+     *
+     * Choice(m1)
+     * 0 Leaf 0.0
+     * 1 Choice(m0)
+     * 1 0 Leaf 1.0
+     * 1 1 Leaf 2.0
+     *
+     * @return size_t
+     */
+    size_t nrAssignments() const;
+
+    /**
      * @brief Fold a binary function over the tree, returning accumulator.
      *
      * @tparam X type for accumulator.
--- a/gtsam/hybrid/tests/testHybridBayesNet.cpp
+++ b/gtsam/hybrid/tests/testHybridBayesNet.cpp
@@ -231,7 +231,7 @@ TEST(HybridBayesNet, Pruning) {
   auto prunedTree = prunedBayesNet.evaluate(delta.continuous());
 
   // Regression test on pruned logProbability tree
-  std::vector<double> pruned_leaves = {0.0, 20.346113, 0.0, 19.738098};
+  std::vector<double> pruned_leaves = {0.0, 32.713418, 0.0, 31.735823};
   AlgebraicDecisionTree<Key> expected_pruned(discrete_keys, pruned_leaves);
   EXPECT(assert_equal(expected_pruned, prunedTree, 1e-6));
 
@@ -248,8 +248,10 @@ TEST(HybridBayesNet, Pruning) {
   logProbability +=
       posterior->at(4)->asDiscrete()->logProbability(hybridValues);
 
+  // Regression
   double density = exp(logProbability);
-  EXPECT_DOUBLES_EQUAL(density, actualTree(discrete_values), 1e-9);
+  EXPECT_DOUBLES_EQUAL(density,
+                       1.6078460548731697 * actualTree(discrete_values), 1e-6);
   EXPECT_DOUBLES_EQUAL(density, prunedTree(discrete_values), 1e-9);
   EXPECT_DOUBLES_EQUAL(logProbability, posterior->logProbability(hybridValues),
                        1e-9);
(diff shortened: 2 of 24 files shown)


varunagrawal:
I changed the way we prune the discrete probabilities by pruning the joint distribution rather than the conditionals. This gives a 3x speedup. Maybe we should be pruning before discrete-only elimination.

#### Before
Number of timesteps: K = 16

```
-Total: 0 CPU (0 times, 0 wall, 0.65 children, min: 0 max: 0)
|   -SmootherEstimation: 3.65 CPU (1 times, 3.70242 wall, 0.65 children, min: 0 max: 0)
|   |   -SmootherUpdate: 3.63 CPU (15 times, 3.67252 wall, 0.65 children, min: 0 max: 0)
|   |   |   -assembleGraphTree: 0.01 CPU (30 times, 0.00976 wall, 0.01 children, min: 0 max: 0)
|   |   |   -hybrid continuous eliminate: 0 CPU (168 times, 0.006431 wall, 0 children, min: 0 max: 0)
|   |   |   -HybridBayesNet PruneDiscreteConditionals: 0.64 CPU (15 times, 0.640958 wall, 0.64 children, min: 0 max: 0)
|   |   |   -HybridBayesNet UpdateDiscreteConditionals: 2.8 CPU (15 times, 2.84436 wall, 0 children, min: 0 max: 0)
|   |   |   |   -HybridBayesNet MakeConditional: 0 CPU (120 times, 0.001385 wall, 0 children, min: 0 max: 0)
|   |   |   -HybridBayesNet PruneMixtures: 0 CPU (15 times, 0.009233 wall, 0 children, min: 0 max: 0)
```

#### After
Number of timesteps: K = 16

```
-Total: 0 CPU (0 times, 0 wall, 0.71 children, min: 0 max: 0)
|   -SmootherEstimation: 0.91 CPU (1 times, 0.904348 wall, 0.71 children, min: 0 max: 0)
|   |   -SmootherUpdate: 0.88 CPU (15 times, 0.882961 wall, 0.71 children, min: 0 max: 0)
|   |   |   -assembleGraphTree: 0 CPU (30 times, 0.009388 wall, 0 children, min: 0 max: 0)
|   |   |   -hybrid continuous eliminate: 0 CPU (168 times, 0.004795 wall, 0 children, min: 0 max: 0)
|   |   |   -HybridBayesNet PruneDiscreteConditionals: 0.64 CPU (15 times, 0.632673 wall, 0.64 children, min: 0 max: 0)
|   |   |   -HybridBayesNet UpdateDiscreteConditionals: 0.07 CPU (15 times, 0.071286 wall, 0.07 children, min: 0 max: 0)
|   |   |   -HybridBayesNet PruneMixtures: 0 CPU (15 times, 0.00871 wall, 0 children, min: 0 max: 0
```

dellaert:
> I changed the way we prune the discrete probabilities by pruning the joint distribution rather than the conditionals. This gives a 3x speedup. Maybe we should be pruning before discrete-only elimination.

Does this PR make *both* these changes? I'd prefer to review a PR that *just* does the tablefactor and shows the speedup...

varunagrawal:
> > I changed the way we prune the discrete probabilities by pruning the joint distribution rather than the conditionals. This gives a 3x speedup. Maybe we should be pruning before discrete-only elimination.
> 
> Does this PR make _both_ these changes? I'd prefer to review a PR that _just_ does the tablefactor and shows the speedup...

This PR only updates the discrete elimination to prune the joint distribution. `TableFactor` is not yet used, and will come in a subsequent PR. :)

Updated PR description to reflect the changes happening here.

dellaert:
I'd still like to request to split into TableFactor related changes and other, and PR straight to develop? Otherwise the base branch will become an un-reviewable kitchen-sink PR. PS CI seems to fail, so splitting might help there as well.

varunagrawal:
In that case I'm going to have to do some cherry picking and force pushing.

varunagrawal:
> I'd still like to request to split into TableFactor related changes and other, and PR straight to develop? Otherwise the base branch will become an un-reviewable kitchen-sink PR. PS CI seems to fail, so splitting might help there as well.

It won't be a kitchen sink PR if we merge in the parent PRs first.

varunagrawal:
@dellaert I split the changes to TableFactor into #1580.

gtsam/discrete/DecisionTree.h
```diff
@@ -299,6 +299,42 @@ namespace gtsam {
     /// Return the number of leaves in the tree.
     size_t nrLeaves() const;
 
+    /**
+     * @brief This is a convenience function which returns the total number of
```
dellaert, line 303: Spelling. And why are we adding it ? And why is the implementation recursive.
I would just as well delete it unless it has a purpose.

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -231,7 +231,7 @@ TEST(HybridBayesNet, Pruning) {
   auto prunedTree = prunedBayesNet.evaluate(delta.continuous());
 
   // Regression test on pruned logProbability tree
-  std::vector<double> pruned_leaves = {0.0, 20.346113, 0.0, 19.738098};
+  std::vector<double> pruned_leaves = {0.0, 32.713418, 0.0, 31.735823};
```
dellaert, line 234: What's happening here? Why are regressions changing ?

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -248,8 +248,10 @@ TEST(HybridBayesNet, Pruning) {
   logProbability +=
       posterior->at(4)->asDiscrete()->logProbability(hybridValues);
 
+  // Regression
   double density = exp(logProbability);
-  EXPECT_DOUBLES_EQUAL(density, actualTree(discrete_values), 1e-9);
+  EXPECT_DOUBLES_EQUAL(density,
```
dellaert, line 253: This was *not* a regression but suddenly there is an arbitrary mult factor here?

gtsam/hybrid/tests/testMixtureFactor.cpp
```diff
@@ -63,8 +63,8 @@ TEST(MixtureFactor, Printing) {
       R"(Hybrid [x1 x2; 1]
 MixtureFactor
  Choice(1) 
- 0 Leaf Nonlinear factor on 2 keys
- 1 Leaf Nonlinear factor on 2 keys
+ 0 Leaf [1]Nonlinear factor on 2 keys
```
dellaert, line 66: spacing is weird in this case.

dellaert (commented):
Some comments. This PR changed too many things at once to be sure of anything, though.

gtsam/hybrid/tests/testHybridBayesNet.cpp
varunagrawal: The decision tree is normalizing the values based on the introduced zeros from pruning. Since I changed the way we're pruning (on the joint rather than the conditionals), the normalizing factor has changed.

gtsam/discrete/DecisionTree.h
varunagrawal: I'm a bit lost on which word is misspelled. The purpose is to help with testing and ensure correctness as a convenience method.

gtsam/hybrid/tests/testHybridBayesNet.cpp
varunagrawal: That is the multiplicative factor for the different normalization constant.

gtsam/hybrid/tests/testMixtureFactor.cpp
varunagrawal: Done

dellaert (approved):
OK, merge this at will :-)

PR #1578, opened 2023-07-15
Show remaining keys in InconsistentEliminationRequested

varunagrawal:
I found this feature to be extremely useful when debugging the hybrid elimination and I figured others would too, hence making a PR.

Generally, when the `InconsistentEliminationRequested` exception is thrown, it shows an error message that is helpful but not informative, i.e. it doesn't say where the elimination inconsistency is coming from. Printing out the keys at the end helps with that.

Now the error message looks like this (note the keys printed at the end)
```
 Failure: "Exception: An inference algorithm was called with inconsistent arguments.  The
factor graph, ordering, or variable index were inconsistent with each
other, or a full elimination routine was called with an ordering that
does not include all of the variables.
Leftover keys after elimination: 2, x3"
```

### DIFF
--- a/gtsam/inference/inferenceExceptions.h
+++ b/gtsam/inference/inferenceExceptions.h
@@ -29,7 +29,7 @@ namespace gtsam {
  * full elimination routine was called with an ordering that does not include
  * all of the variables. */
 class InconsistentEliminationRequested : public std::exception {
-  KeySet keys_;
+  KeyVector keys_;
   const KeyFormatter& keyFormatter = DefaultKeyFormatter;
 
  public:
@@ -38,21 +38,26 @@ class InconsistentEliminationRequested : public std::exception {
   InconsistentEliminationRequested(
       const KeySet& keys,
       const KeyFormatter& key_formatter = DefaultKeyFormatter)
-      : keys_(keys), keyFormatter(key_formatter) {}
+      : keys_(keys.begin(), keys.end()), keyFormatter(key_formatter) {}
 
   ~InconsistentEliminationRequested() noexcept override {}
   const char* what() const noexcept override {
     // Format keys for printing
     std::stringstream sstr;
-    for (auto key : keys_) {
-      sstr << keyFormatter(key) << ", ";
+    size_t nrKeysToDisplay = std::min(size_t(4), keys_.size());
+    for (size_t i = 0; i < nrKeysToDisplay; i++) {
+      sstr << keyFormatter(keys_.at(i));
+      if (i < nrKeysToDisplay - 1) {
+        sstr << ", ";
+      }
     }
+    if (keys_.size() > nrKeysToDisplay) {
+      sstr << ", ... (total " << keys_.size() << " keys)";
+    }
+    sstr << ".";
     std::string keys = sstr.str();
-    // remove final comma and space.
-    keys.pop_back();
-    keys.pop_back();
 
-    static std::string msg =
+    std::string msg =
         "An inference algorithm was called with inconsistent "
         "arguments.  "
         "The\n"
@@ -64,7 +69,8 @@ class InconsistentEliminationRequested : public std::exception {
         "that\n"
         "does not include all of the variables.\n";
     msg += ("Leftover keys after elimination: " + keys);
-    return msg.c_str();
+    // `new` to allocate memory on heap instead of stack
+    return (new std::string(msg))->c_str();
   }
 };
 }  // namespace gtsam
--- a/gtsam/inference/inferenceExceptions.h
+++ b/gtsam/inference/inferenceExceptions.h
@@ -18,9 +18,9 @@
 #pragma once
 
 #include <gtsam/global_includes.h>
+#include <gtsam/inference/Key.h>
 
 #include <exception>
-#include <sstream>
 
 namespace gtsam {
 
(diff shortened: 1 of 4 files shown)


dellaert:
@varunagrawal , nice, but what happens if we are eliminating 300k variables?

varunagrawal:
It will only show the keys leftover after elimination which need not be all 300K. The idea is that it's always a few keys that cause the issue (as @travisdriver recently discovered) so having that info can help debug.

dellaert:
> It will only show the keys leftover after elimination which need not be all 300K. The idea is that it's always a few keys that cause the issue (as @travisdriver recently discovered) so having that info can help debug.

Can you see there might be a bit of a problem with printing out 150k keys if this occurs halfway :-) ?

dellaert:
> It will only show the keys leftover after elimination which need not be all 300K. The idea is that it's always a few keys that cause the issue (as @travisdriver recently discovered) so having that info can help debug.

Can you see there might be a bit of a problem with printing out 150k keys if this occurs halfway :-) ?

travisdriver:
> > It will only show the keys leftover after elimination which need not be all 300K. The idea is that it's always a few keys that cause the issue (as @travisdriver recently discovered) so having that info can help debug.
> 
> 
> 
> Can you see there might be a bit of a problem with printing out 150k keys if this occurs halfway :-) ?

I still think, in this very specific scenario, too much information is better than none :-)

travisdriver:
Otherwise, maybe this could be part of some verbosity setting?

varunagrawal:
> > It will only show the keys leftover after elimination which need not be all 300K. The idea is that it's always a few keys that cause the issue (as @travisdriver recently discovered) so having that info can help debug.
> 
> Can you see there might be a bit of a problem with printing out 150k keys if this occurs halfway :-) ?

I figured they would first debug with a smaller example in that case? Then this PR really shines, especially if the smaller example is informed from the first few keys printed out from the larger example.

ProfFan:
I think having all keys in the exception thrown is OK. However in `::what()` we should print only the first few keys, for example:
```
keys 1, 2, 3, 4, ... (total 11451 keys)
```

dellaert:
> > I think having all keys in the exception thrown is OK. However in `::what()` we should print only the first few keys, for example:
> > ```
> > keys 1, 2, 3, 4, ... (total 11451 keys)
> > ```
> 
> I like this idea!

Me three !

varunagrawal:
Added a test to verify that when a large number of keys exist (in this case 1001), we get the following exception message:

```
An inference algorithm was called with inconsistent arguments.  The
factor graph, ordering, or variable index were inconsistent with each
other, or a full elimination routine was called with an ordering that
does not include all of the variables.
Leftover keys after elimination: 2, 3, 4, 5, ... (total 999 keys).
```

gtsam/inference/inferenceExceptions.h
```diff
@@ -12,30 +12,65 @@
...
- * @author  Richard Roberts
+ * @author  Richard Roberts, Varun Agrawal
  * @date    Apr 25, 2013
  */
 #pragma once
 
 #include <gtsam/global_includes.h>
+
 #include <exception>
+#include <sstream>
```
dellaert, line 23: This will now be included in *all* files that include these exceptions. I propose moving to a cpp file

gtsam/inference/inferenceExceptions.h
```diff
@@ -12,30 +12,65 @@
...
+ public:
+  InconsistentEliminationRequested() noexcept {}
+
+  InconsistentEliminationRequested(
+      const KeySet& keys,
+      const KeyFormatter& key_formatter = DefaultKeyFormatter)
+      : keys_(keys.begin(), keys.end()), keyFormatter(key_formatter) {}
+
+  ~InconsistentEliminationRequested() noexcept override {}
+  const char* what() const noexcept override {
```
dellaert, line 44: Move implementation to .cpp file please?

gtsam/linear/tests/testGaussianFactorGraph.cpp
```diff
@@ -18,16 +18,16 @@
...
-#include <gtsam/linear/GaussianFactorGraph.h>
-#include <gtsam/linear/GaussianConditional.h>
-#include <gtsam/linear/GaussianBayesNet.h>
-#include <gtsam/inference/VariableSlots.h>
-#include <gtsam/inference/VariableIndex.h>
-#include <gtsam/base/debug.h>
-#include <gtsam/base/VerticalBlockMatrix.h>
-
-#include <gtsam/base/TestableAssertions.h>
 #include <CppUnitLite/TestHarness.h>
```
dellaert, line 21: The include order was already correct : please restore.

dellaert (commented):
Thanks !!! But some comments ...

gtsam/linear/tests/testGaussianFactorGraph.cpp
varunagrawal: Autoformat rearranged this. I'll restore it but wanted to let you know.

gtsam/inference/inferenceExceptions.h
varunagrawal: Done

gtsam/linear/tests/testGaussianFactorGraph.cpp
varunagrawal: Done

dellaert (approved):
Great, thanks for this cool feature!

PR #1583, opened 2023-07-19
Added some python and c++ examples with BinaryMeasurements

dellaert:
- in response to GTSAM group question
- also added a check on pMin, which segfaulted if pMin < d :-(


varunagrawal (approved):
LGTM

PR #1581, opened 2023-07-18
Fix more windows tests

talregev:
Fix more windows tests

PR #1553, opened 2023-06-21
Fix Windows Tests

varunagrawal:
My attempt at fixing #1541.

1. Add fix for compilation in Visual Studio. This seems to be a weird issue with how `size_t` is interpreted since the `multiple symbols defined` error comes from additional declarations of `std::map<Key, size_t>`. This might be a bug in MSVC 16.
2. Fixed some more tests across the repo.
3. Updated the Windows CI to run additional GTSAM tests as well as some `gtsam_unstable` tests.

There is still quite a bit of work to do to get all the tests passing on Windows, but hopefully this is a good start point.

PR #1584, opened 2023-07-20
Fix windows tests

### LINKED ISSUE #1541: GTSAM_UNSTABLE Multiple Defined Symbols
When compiling GTSAM on Windows, `GTSAM_UNSTABLE` throws a compilation error due to multiple defined symbols in `LPInitSolver.obj`.

This could indicate some badness in the `GTSAM_UNSTABLE_EXPORT` in the files related to `LPInitSolver.obj`. Below is the output from the MSVC compiler:

```sh
Build started...
1>------ Build started: Project: gtsam_unstable, Configuration: Release x64 ------
1>gtsam.lib(gtsam.dll) : error LNK2005: "public: __cdecl std::map<unsigned __int64,unsigned __int64,struct std::less<unsigned __int64>,class std::allocator<struct std::pair<unsigned __int64 const ,unsigned __int64> > >::~map<unsigned __int64,unsigned __int64,struct std::less<unsigned __int64>,class std::allocator<struct std::pair<unsigned __int64 const ,unsigned __int64> > >(void)" (??1?$map@_K_KU?$less@_K@std@@V?$allocator@U?$pair@$$CB_K_K@std@@@2@@std@@QEAA@XZ) already defined in LPInitSolver.obj
1>   Creating library C:/Mike/Work/Build/gtsam/lib/Release/gtsam_unstable.lib and object C:/Mike/Work/Build/gtsam/lib/Release/gtsam_unstable.exp
1>C:\Mike\Work\Build\gtsam\gtsam_unstable\bin\Release\gtsam_unstable.dll : fatal error LNK1169: one or more multiply defined symbols found
1>Done building project "gtsam_unstable.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 69 up-to-date, 0 skipped ==========
```

talregev:
Fix #1541
Minimum changes for fix windows tests.
This PR is aiming to fix compilation errors. 
Although it not fix all the compilation errors, it solve the majority of them. 
@dellaert Can you take a look of this?

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -99,64 +99,65 @@ jobs:
           cmake -B build -S . -DGTSAM_BUILD_EXAMPLES_ALWAYS=OFF -DBOOST_ROOT="${env:BOOST_ROOT}" -DBOOST_INCLUDEDIR="${env:BOOST_ROOT}\boost\include" -DBOOST_LIBRARYDIR="${env:BOOST_ROOT}\lib"
 
       - name: Build
-        shell: cmd
         run: |
-          :: Since Visual Studio is a multi-generator, we need to use --config
-          :: https://stackoverflow.com/a/24470998/1236990
+          # Since Visual Studio is a multi-generator, we need to use --config
+          # https://stackoverflow.com/a/24470998/1236990
           cmake --build build -j4 --config ${{ matrix.build_type }} --target gtsam
           cmake --build build -j4 --config ${{ matrix.build_type }} --target gtsam_unstable
 
-          :: Target doesn't exist
-          :: cmake --build build -j4 --config ${{ matrix.build_type }} --target wrap
+          # Target doesn't exist
+          # cmake --build build -j4 --config ${{ matrix.build_type }} --target wrap
 
       - name: Test
-        shell: cmd
         run: |
-          :: Run GTSAM tests
+          # Run GTSAM tests
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.base
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.basis
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.discrete
-          cmake --build build -j4 --config ${{ matrix.build_type }} --target check.geometry
+
+          # Compilation error
+          # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.geometry
+
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.inference
 
-          :: Compile. Fail with exception
-          :: cmake --build build -j4 --config ${{ matrix.build_type }} --target check.linear
+          # Compile. Fail with exception
+          cmake --build build -j4 --config ${{ matrix.build_type }} --target check.linear
           
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.navigation          
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.sam
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.sfm          
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.symbolic
 
-          :: Compile. Fail with exception
-          :: cmake --build build -j4 --config ${{ matrix.build_type }} --target check.hybrid
+          # Compile. Fail with exception
+          cmake --build build -j4 --config ${{ matrix.build_type }} --target check.hybrid
 
-          :: Compile. Fail with exception
-          :: cmake --build build -j4 --config ${{ matrix.build_type }} --target check.nonlinear
+          # Compile. Fail with exception
+          # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.nonlinear
 
-          :: Compilation error
-          :: cmake --build build -j4 --config ${{ matrix.build_type }} --target check.slam
+          # Compilation error
+          # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.slam
 
 
-          :: Run GTSAM_UNSTABLE tests
+          # Run GTSAM_UNSTABLE tests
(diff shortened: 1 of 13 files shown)


ProfFan:
Why is the linear tests failing?

talregev:
> Why is the linear tests failing?

image
image

dellaert (approved):
Awesome! Most users use Mac/Linux or python so it's good we have some windows intel improving that path! @ProfFan since all tests that are requested to run do run, I have no issue merging this PR, do you?

.github/workflows/build-windows.yml
```diff
@@ -99,26 +99,64 @@ jobs:
...
-          cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.navigation
-          #cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.nonlinear
-          #cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.sam
-          cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.sfm
-          #cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.slam
-          cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.symbolic
-
-          # Run GTSAM_UNSTABLE tests
-          #cmake --build build -j 4 --config ${{ matrix.build_type }} --target check.base_unstable
+          :: Since Visual Studio is a multi-generator, we need to use --config
```
varunagrawal, line 104: Can you please explain the reason you've changed the syntax for the comment?

varunagrawal (approved):
I don't quite understand why these changes were not simply pushed to #1581 but this looks great to me!

varunagrawal:
@dellaert we can merge this in once we get some clarification on the change in syntax for the YAML comments. I can't find a reference to that syntax anywhere, and for the sake of other developers, I would like to use standard syntax.

.github/workflows/build-windows.yml
talregev: Yes. I changed shell to be cmd and not PowerShell because I want it to fails on first error to discover errors early.
This is one way to have can comment on cmd the close one and similar to `#`

ProfFan (changes requested):
Thank you for the contribution! However I think we should fix the linear check (which regressed) before merging.

talregev:
> Thank you for the contribution! However I think we should fix the linear check (which regressed) before merging.

This error was before. It is was hides because the ci wasn't fails on the first error.
This PR is aiming to fix compilation error only.

talregev:
> Thank you for the contribution! However I think we should fix the linear check (which regressed) before merging.

As you can see from my changes, I didn't change any cpp file or the logic.
This error cause from the correctness of the tests.
This PR is aiming to fix compilation errors only.
Feel free to take these changes and explore why it happening.

talregev:
> Thank you for the contribution! However I think we should fix the linear check (which regressed) before merging.

In the screenshot I compiled the tests with ninja that give automatic debug assertion. 
So the test is not failing without debug assertion, but it not meaning is that ok. It mean there is hidden problem you should address. 
So for now the linear check that you care about, it uncomment  and running and pass in the ci. But I would recommended that you check why the debug assertion cause this issue. This PR I am aiming only solve a completion errors. 
I just uncomment to show you that the state of the tests is the same as before. (And also improve cause now more tests are compiled).

talregev:
I switch back to powershell. cmd seems not to fail on the first error. Not sure how. Need to investigate it on different PR.
So Now you back the powershell and familiar `#` for comment. 

Also because the cmd is not fail on the first error, I missed that the geometry fail on compilation error. So I comment it as well.

.github/workflows/build-windows.yml
talregev: I changed it back to `#`

talregev:
I will switch to shell: bash on windows that it will fail on first error:
https://stackoverflow.com/questions/62996357/stop-multiline-github-action-on-first-fail
https://github.com/actions/runner-images/issues/6668

dellaert:
I really want to land this PR, I think it is great :-) But I am still very confused.
The CI says linear passes:
```
      Start 1: check_linear_program
  1/1 Test #1: check_linear_program .............   Passed    0.42 sec
```
@ProfFan says there is a regression - but what exactly is the regression if the linear tests pass? I know I asked you to comment on it and request changes yesterday, but looking back at the changes I don't see anything in the PR that comments out a linear test.
@talregev says something about debug asserts I don't understand. The CI I quoted was a debug run, so presumably debug assertion should trigger there. 

I don't think we necessarily *need* to get to the bottom of this before merging, BTW - I think this PR is a *big* net positive, and the CI passes, and I want to thank @talregev for doing this and let him get on with life :-) @ProfFan what do you think?

talregev:
@dellaert The creation of the confusion is coming from my side. When I check the tests, I didn't check them on the ci.
I checked them on my local machine, and compiled it differently as the ci. I compiled it with ninja that add the flag `_DEBUG` automatic.
More information is here:
https://learn.microsoft.com/en-us/visualstudio/debugger/c-cpp-assertions?view=vs-2022
I think it only happen on c++. I am not sure if the same functoriality can be with gcc.
So it crash, then I report it that it crash without check it on the ci. But it actually was running on the ci all along. 
So the test was running all along with my changes. But there may have a problem in your code. As the error msg suggest, you are doing some operation in c++ that expect ordered list/array and it not given. You should investigate it. 
I agree this investigation should not happen in this PR, also not block this PR.  
So for now I uncomment the tests that it will run on the ci. and you can take these changes and continue to build upon it.

ProfFan (approved):
Thank you for the detailed explanation!

varunagrawal:
@talregev I can merge this. Just wanted to make sure you intended to merge this into `develop` and not #1553 as we have been doing previously. I would like to merge #1553 soon as well since you, @mikesheffler and I have spent considerable time figuring out all these kinks.

dellaert:
Merged, many thanks !!!!

PR #1587, opened 2023-07-20
fixes issues with TBB not being found when building downstream packages

nathanhhughes:
Hi! I noticed that #1567 broke downstream packages that build against GTSAM (e.g., Kimera-RPGO) as it switches to linking against TBB interface/alias libraries which aren't declared as a dependency. Regardless, TBB should probably be added via `find_dependency` in GTSAM's cmake config if it's being publicly linked against.

This at least fixes the issue for me when TBB is found, but I'm not entirely sure if:
1) it would be preferable to have a TBB version set in `HandleTBB.cmake` that GTSAM's config could use
2) this works when TBB isn't present on the system

### DIFF
--- a/cmake/Config.cmake.in
+++ b/cmake/Config.cmake.in
@@ -21,6 +21,10 @@ else()
 find_dependency(Boost @BOOST_FIND_MINIMUM_VERSION@ COMPONENTS @BOOST_FIND_MINIMUM_COMPONENTS@)
 endif()
 
+if(@GTSAM_USE_TBB@)
+    find_dependency(TBB 4.4 COMPONENTS tbb tbbmalloc)
+endif()
+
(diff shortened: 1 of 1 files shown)


ProfFan:
Thanks @nathanhhughes ! The package config files probably needs some care as CMake now have evolved so much...

dellaert:
> Thanks @nathanhhughes ! The package config files probably needs some care as CMake now have evolved so much...

@ProfFan you made this comment but then did not approve PR. If you agree this is good (i did not check in detail), please approve and merge?

varunagrawal (approved):
@nathanhhughes you may want to note that the latest versions of TBB have adopted a different versioning scheme that reflects the year of release. Other than that LGTM.

PR #1592, opened 2023-07-25
TableFactor Constructors

varunagrawal:
Add 2 new constructors for the `TableFactor`, one from a `DecisionTreeFactor` and one from `DiscreteConditional`.

These constructors use the new `probabilities()` method added in #1590 and require some rearrangement via the `ComputeLeafOrdering` function.


varunagrawal:
@dellaert I am going to merge this in since it forms a part of getting discrete elimination with pruning working efficiently. The PR will be in our history for future reference. :)

PR #1595, opened 2023-07-27
Add ninja build to all builds

talregev:
Add ninja build to all builds
The debug assertion with the flag `_DEBUG` that explain here:
https://learn.microsoft.com/en-us/visualstudio/debugger/c-cpp-assertions?view=vs-2022
Is not coming from ninja, It coming from vcpkg build I did. 
If you want, you can compile all your builds with ninja.

### DIFF
--- a/.github/scripts/python.sh
+++ b/.github/scripts/python.sh
@@ -9,14 +9,13 @@ set -x -e
 # install TBB with _debug.so files
 function install_tbb()
 {
+  echo install_tbb
   if [ "$(uname)" == "Linux" ]; then
     sudo apt-get -y install libtbb-dev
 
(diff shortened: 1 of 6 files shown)


varunagrawal:
I don't see any reason to have Ninja enabled for all builds since the assertions you've mentioned seem to be Windows specific.

Plus there are a lot of changes here that don't seem directly related to the PR comment.

talregev:
The ninja build is not for the assertion. You can read about ninja build here:
https://ninja-build.org/
The assertion can add in different PR when you want to solve it by adding the flag manually. 
I am closing this PR. 
If you still want it, just comment here.

PR #1597, opened 2023-07-28
Add some new loss functions

NaivelyWritten:
This PR adds some new loss functions that is asymmetric, similar to the L2WithDeadZone already exists.


dellaert (commented):
Add at least one test?

NaivelyWritten:
Add the unit test @dellaert

dellaert (approved):
Awesome!

PR #1598, opened 2023-07-29
Simplify scripts

talregev:
Simplify scripts:
- Minimize the need of change directory (cd) command. It better to have all command from source code directory if possible.
- Evoke cmake build instead of make. It make it more general approach. cmake will invoke make or other build system he was generated for. 
- Change conditional command to be more correctly in the ci. Some of the conditional create error return. If you edit the scripts it can lead that it fail the CI. Also it more readable that it conditional. 
- The script is now independently from the environment variable: `GITHUB_WORKSPACE`. It mean you can also run it locally on your machine and test it. Also work on the CI. 

@dellaert Can you take a look on it?

### DIFF
--- a/.github/scripts/python.sh
+++ b/.github/scripts/python.sh
@@ -9,14 +9,13 @@ set -x -e
 # install TBB with _debug.so files
 function install_tbb()
 {
+  echo install_tbb
   if [ "$(uname)" == "Linux" ]; then
     sudo apt-get -y install libtbb-dev
 
   elif [ "$(uname)" == "Darwin" ]; then
     brew install tbb
-
   fi
-
 }
 
 if [ -z ${PYTHON_VERSION+x} ]; then
@@ -37,19 +36,19 @@ function install_dependencies()
 
   export PATH=$PATH:$($PYTHON -c "import site; print(site.USER_BASE)")/bin
 
-  [ "${GTSAM_WITH_TBB:-OFF}" = "ON" ] && install_tbb
+  if [ "${GTSAM_WITH_TBB:-OFF}" == "ON" ]; then
+    install_tbb
+  fi
 
   $PYTHON -m pip install -r $GITHUB_WORKSPACE/python/requirements.txt
 }
 
 function build()
 {
-  mkdir $GITHUB_WORKSPACE/build
-  cd $GITHUB_WORKSPACE/build
-
   BUILD_PYBIND="ON"
-
-  cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
+  cmake $GITHUB_WORKSPACE \
+      -B build \
+      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
       -DGTSAM_BUILD_TESTS=OFF \
       -DGTSAM_BUILD_UNSTABLE=${GTSAM_BUILD_UNSTABLE:-ON} \
       -DGTSAM_USE_QUATERNIONS=OFF \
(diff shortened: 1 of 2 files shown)


varunagrawal:
I again don't understand the need for these changes. They are not critical or fix any existing bug. Instead we still have to resolve other pending issues.

talregev:
> I don't quite understand the need for these changes. They are not critical or fix any existing bug. Instead we still have to resolve other pending issues.

PRs is not only for critical fixes. My main goal is to improve CI and compilation process. 

> Sure the script looks cleaner, but the existing version is simple and easy to modify for new students and collaborators who may not be familiar with the Actions CI.

If you looking for new students perspective:
- Better to have simple scripts with less lines that it more easy and learn to understand. Easy to learn, understand and modify.
- Removing the `GITHUB_WORKSPACE` it mean they will not have to know about the ci, and can run the scripts locally on their pc. If it work, probably it will work on the CI. current state don't allow that. They need to know about the CI and what it mean `GITHUB_WORKSPACE`.

varunagrawal:
The PR isn't quite improving the CI since it's just replacing one thing with another, so I don't understand the benefit of that. Plus keeping the process explicit with calls to Cmake and then make means that there is no "magic" going on since most people are familiar with this 2 step process.

Using the environment variables should be enforced since it prevents issues due to admin privileges across various images. We've had this before especially with the wrappers and hence we use it. We don't like to make things complicated without a good reason. :)

varunagrawal:
I really appreciate your time and effort into improving GTSAM, and you do a great job. But reviewing also takes up time which we are generally short of due to being undermanned for the most part.

dellaert:
@talregev Please disregard comments earlier in this thread. I will take a look at the changes when I have time, but I'm very busy so it might take a bit. Thanks !!!!

dellaert:
This looks good to me and all CI passes, unless @varunagrawal or @ProfFan have a strong rationale to keep the environment variables? For example, were they introduced for interacting with wrap or some other reason? I know @ProfFan created a set of scripts I use to create the M1 mac wrappers and maybe they were introduced for that purpose?

varunagrawal:
My main concern is that security and sudo settings vary on Linux, Mac and Windows, so having all the code in the GitHub provided location (aka the workspace) can help ameliorate issues related to this.

The Actions CI is super frustrating to debug since we can't SSH into the servers. This is why I prefer keeping the CI as idiomatic as possible.

varunagrawal:
In the CI if we use the environment variable, then we know exactly where the repo is cloned.

The environment variable also ensures that the build directory is set up where user permissions are enabled and not in an admin location like /usr/lib. If we just use a blanket build directory to wherever the repo is cloned, we can't guarantee that all the permissions are enabled.

The Actions team changes the CI image quite frequently without notice and this has been a point of contention for me.

Regarding your point on running the CI locally, the two environments are different (extra environment variables, additional settings on GitHub for security purposes, virtualization settings, etc) so while debugging GTSAM is easy, debugging the CI is a whole other process.

dellaert:
@talregev there seem to be good reasons for the env variable. If this is the main reason for this PR, I'd vote to close it. Or restore the env variable and keep other improvements.

talregev:
> Yes, apparently we added those environment variables on advice of the Github docs, and there is no good reason to remove them. The recommended way to compile/run checks is via cmake and make check. If you want to run the actual CI on a local machine, you will have to define these variables.

I didn't talk about running the CI in the local machine. 
I talk about running the scripts in the local machine. These two different things. 
The way you do it, you make a script that run both on your local machine and the CI, and debug it only on your local machine, and the CI only if needed. It save time and energy. My recommendation is to remove it, and continue to develop upon it. 
If you need the other commands / environment variable to operate these script, my recommendation is to create other scripts that build upon them, that run on your local machine, then it run also the CI with everything is needed. 
I can help to build such a scripts. 

I added the `GITHUB_WORKSPACE` as you requested. 

Please review it again. Thank you.

PR #1599, opened 2023-07-30
Fix/iSAM2 #1101

ProfFan:
It appears that the fixes in #1172 is not merged to develop, we may need to solve the conflicts first though...


dellaert:
Wow - totally forgot about this. Seems CI fails?

ProfFan:
@dellaert

dellaert (approved):
Cool, many thanks !!! I will merge

tanzby:
Dose 4.2 release include this PR? Those changes seems important for iSAM2 @dellaert

PR #1607, opened 2023-08-18
Add a custom loss function for Robust, fix a bug in Asymmetric

ProfFan:
Add a way to use a custom loss function without changing GTSAM code


dellaert (approved):
Wow, awesome!
But please add a unit test in python?

PR #1615, opened 2023-08-30
Fix DSF bug where tracks failing transitivity check were returned

johnwlambert:


gtsam/sfm/DsfTrackGenerator.cpp
```diff
@@ -130,7 +130,7 @@ std::vector<SfmTrack2d> tracksFromPairwiseMatches(
   }
 
   // TODO(johnwlambert): return the Transitivity failure percentage here.
-  return tracks2d;
+  return validTracks;
```
akshay-krishnan, line 133: oh no! good catch!

akshay-krishnan (approved):

akshay-krishnan:
will you be able to run gtsfm with this change before merging this PR?

johnwlambert:
@dellaert can we do a 4.2a10 release now and put this in it?

johnwlambert:
> will you be able to run gtsfm with this change before merging this PR?

I think the newly added unit test that fails without the fix, and passes with the fix, is sufficient validation for now. Once we have rebuilt wheels, will check gtsfm’s performance.

dellaert (approved):
Nice. If you want this in a 4.2 release, could you then please cherry pick the commits in another PR that targets release/4.2? I can then cut a version as requested.

johnwlambert:
> Nice. If you want this in a 4.2 release, could you then please cherry pick the commits in another PR that targets release/4.2? I can then cut a version as requested.

Thanks @dellaert. Btw, it looks like the release/4.2 branch is a little old, e.g. it doesn't include updates to the wrapper that we want to use on the python side in gtsfm https://github.com/borglab/gtsam/commit/5b588f2ea70c159f1891d28df3d2eb3c1c219200
Could we instead make release/4.2 from current develop (is release/4.2 still pending)? (i didn't see it on pypi yet)

Thanks!

dellaert:
No, develop is *very* different from 4.2: all of boost is now optional. You could add any updates to the wrapper in the cherry-pick MR.

johnwlambert:
Got it, thanks. @dellaert would it be possible to do a post-4.2 release from today's develop, e.g. a `4.3a1`? It looks like release/4.2 was cut in January (https://github.com/borglab/gtsam-manylinux-build/pull/36), but the relevant python wrapper changes were in June (https://github.com/borglab/gtsam/pull/1546), and there are a lot of changes in between.

PR #1617, opened 2023-08-31
Hybrid Document Update

varunagrawal:
Updated Hybrid.lyx to clarify the invariant structure of the `GaussianMixture` invariant. You should be able to read the PDF and see the changes, primarily around Eqn 6 and right above Eqn 8.

Also removed my name from the document so we can blame Frank if we find any mistakes in the math. 😝


dellaert (approved):
Awesome !

PR #1618, opened 2023-09-01
Update Python CustomFactor documentation and example

cntaylor:
Simple changes in documentation and CustomFactor example when using Python.  

Types were not correct causing runtime errors when creating your own custom factor.  For example, if you return a float as the previous Python file said you should, it fails.  Also, the Jacobians need to be 2D arrays or you get a runtime failuer.  Hopefully the correct types to use are more clear now.

### DIFF
--- a/python/CustomFactors.md
+++ b/python/CustomFactors.md
@@ -11,7 +11,7 @@ import gtsam
 import numpy as np
 from typing import List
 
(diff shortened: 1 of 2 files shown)


dellaert (approved):
Thanks!!!

PR #1624, opened 2023-09-05
4.2.0 Patch: Clean up GTSAM version setting

varunagrawal:
Update CMake to set the GTSAM version values correctly. Do the same for GTSAM_UNSTABLE.

If `GTSAM_PRERELEASE_VERSION` is specified, e.g. "a9", the version string will be set to "4.2a9", otherwise it will be set to "4.2.0".

Once this is merged, I'll cherry-pick this commit to the `develop` branch.


dellaert (changes requested):
Varun, we need to do this in a gitflow correct way, and I would prefer to wait to fix this along with other bug-fixes in a new 4.2.1 release maybe ina  few weeks. Could you create a new branch release/4.2.1 and target that?

varunagrawal:
> I would prefer to wait to fix this along with other bug-fixes in a new 4.2.1 release maybe ina few weeks.

Sure but this is going to delay the python wheels building. Alternatively, we can merge that PR on gtsam-manylinux-build, and I can put in the fix once 4.2.1 is released.

dellaert:
Why not just do what I suggested which is to start the 4.2.1 branch?
Also, why would it delay building wheels? I built and released all wheels *except* 3.11

Issue #1605, opened 2023-08-14
gtsam doesn't build on i386 or armhf


dkogan:
Hi. I pushed gtsam into Debian, and I see (among other things) that it doesn't
build on armhf and i386. Can one of you please take a look? It'll take yall much
less time to detangle this than it would take me.

Here's the i386 build log: https://buildd.debian.org/status/fetch.php?pkg=gtsam&arch=i386&ver=4.2%7E9%2Bdfsg-4&stamp=1692035308&raw=0

The punchline is that when building `gtsam/linear/tests/testSparseEigen.cpp`
this happens:

```
In file included from ./gtsam/base/FastSet.h:28,
                 from ./gtsam/inference/Key.h:22,
                 from ./gtsam/inference/Ordering.h:23,
                 from ./gtsam/inference/EliminateableFactorGraph.h:26,
                 from ./gtsam/linear/GaussianFactorGraph.h:24,
                 from gtsam/linear/tests/testSparseEigen.cpp:21:
./gtsam/base/Testable.h: In instantiation of 'bool gtsam::assert_equal(const V&, const V&, double) [with V = int]':
gtsam/linear/tests/testSparseEigen.cpp:44:3:   required from here
./gtsam/base/Testable.h:99:26: error: incomplete type 'gtsam::traits<int>' used in nested name specifier
   99 |     if (traits<V>::Equals(actual,expected, tol))
      |         ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
./gtsam/base/Testable.h:102:21: error: incomplete type 'gtsam::traits<int>' used in nested name specifier
  102 |     traits<V>::Print(expected,"expected:\n");
      |     ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
./gtsam/base/Testable.h:103:21: error: incomplete type 'gtsam::traits<int>' used in nested name specifier
  103 |     traits<V>::Print(actual,"actual:\n");
```      

It's reproducible standalone, without cmake like this:

```
< cmake/dllexport.h.in sed \
' s/#cmakedefine/#define/g;
  s/@library_name@/GTSAM/g;
' > gtsam/dllexport.h

< gtsam/config.h.in sed \
' s/#cmakedefine/#define/g;
  s/@GTSAM_VERSION_MAJOR@/4/g;
  s/@GTSAM_VERSION_MINOR@/2/g;
[...]

PR #1625, opened 2023-09-05
relax unit test thresholds to fix 32bit issues

jlblancoc:
This will solve part of the pending issues in #1605  : 

(cc @dkogan )
> * i386 fails one of the first set of tests: it looks like it just barely misses some thresholds, so it's probably not a failure. But since it failed something in the first set, we don't know if any of the second set of tests would have failed

### DIFF
--- a/gtsam/geometry/tests/testSphericalCamera.cpp
+++ b/gtsam/geometry/tests/testSphericalCamera.cpp
@@ -24,6 +24,13 @@
 #include <cmath>
 #include <iostream>
 
+#if defined(__i686__) || defined(__i386__)
+// See issue discussion: https://github.com/borglab/gtsam/issues/1605
+constexpr double TEST_THRESHOLD = 1e-5;
+#else
+constexpr double TEST_THRESHOLD = 1e-7;
+#endif
+
 using namespace std::placeholders;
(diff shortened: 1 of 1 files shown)


varunagrawal:
Would it be wise to set the thresholds based on preprocessor flags indicating the architecture type?

If we have the flag `GTSAM_ARCH_TYPE`, we could check at compile time and set this accordingly. Just a suggestion.

ProfFan (approved):
@jlblancoc Before merging could you add a `constexpr if` check on the current architecture?

jlblancoc:
Take a look at the patch now.

I checked that `i686-linux-gnu-g++` defines both preprocessor symbols. i386 is actually obsolete and only i686 will remain for a few years as a main Debian arch, but I added both just in case.

@varunagrawal The idea of a cmake-level architecture variable that maps into a macro seems good (I think that is what your proposal is about?), but if we are using it just here, maybe just the macros are the minimal solution...

jlblancoc:
One CI task failed, but *I think* it's not related to this PR's changes... I have re-launched it.

varunagrawal:
Awesome! Merged

PR #1626, opened 2023-09-05
Fix GTSAM version setting

varunagrawal:
Update CMake to set the GTSAM version values correctly. Do the same for GTSAM_UNSTABLE.

For the current version 4.3.0, if `GTSAM_PRERELEASE_VERSION` is specified, e.g. "a0", the version string will be set to "4.3a0", otherwise it will be set to "4.3.0".


dellaert:
Cool. How do you know it works?

varunagrawal:
> Cool. How do you know it works?

I built the python wheels locally in a docker container and was able to verify this was correct.

This issue only seems to affect the `GTSAM_VERSION_STRING` variable which is what is used to set the version in the gtsam python package (please refer to `setup.py.in` line 27).
Interestingly, downstream C++ projects (like GTD) use the version set in the `project()` cmake function, so CMake there would show GTSAM as 4.3.0 which is correct.

dellaert (approved):

PR #1628, opened 2023-09-05
Python Wrapper Updates

### LINKED ISSUE #18: surplus requirements?
I was able to build and upload 4.2a0 to pypi. However, I noticed something weird:

image

I wonder why numpy and pyparsing are installed? The latter makes me think these are requirements for *building* the wrapper, but I'm puzzled why they show up as requirements to use gtsam. Likewise, having numpy, while nice, is *not* a requirement for using gtsam I think.

varunagrawal:
1. Update Pybind11 from v2.11.0 to v2.11.1. This should resolve any dependabot alerts as well as bring in new fixes and improvements while being backwards compatible.
2. Introduce new requirements file called `dev_requirements.txt`. This is the requirements file which will be used when building the python wrapper (and includes numpy & pyparsing), while `requirements.txt` will be used for the pip package and specify dependencies (in the current case, only numpy).
3. Update CMake to install python dev dependencies automatically when building the wrapper. This uses the aforementioned `dev_requirements.txt`.
4. Update the `setup.py.in` template to follow new standards and eliminate generated warnings when building and packaging the wheels. The biggest change is due to [PEP 420](https://peps.python.org/pep-0420/) which causes `setuptools` to no longer disambiguate between code directories and data directories.
5. Removed the classifier showing support for Python 2. Python 2 has been dead since January 1st 2020, so now seems like a good time to do this given it has been 3.5 years.


talregev:
You really should not update 3rd party library like that. 
vcpkg is prefer. I already have a PR for that.
https://www.reddit.com/r/cpp/comments/s102ld/comment/hs5e8w2/?utm_source=reddit&utm_medium=web2x&context=3

varunagrawal:
We have some internal reasons for doing it this way, and don't have the bandwidth currently to make a switch.

Moreover, Pybind11 is not a GTSAM dependency so there's no reason to vendor it. It is only used to build the python package which is then uploaded to pip.

dellaert:
Why is target not develop?

dellaert:
Love this, BTW! Will approve when base branch is changed

varunagrawal:
> Why is target not develop?

Ah sorry about that. I was cascading PRs and the CMake fix PR branch got deleted before I could change branches.

ProfFan:
@varunagrawal Could you create a draft PR in gtsam-manylinux-build that target this branch and test if the workflow still pass?

varunagrawal:
> @varunagrawal Could you create a draft PR in gtsam-manylinux-build that target this branch and test if the workflow still pass?

Sure, though there is an issue with the `GTSAM_VERSION_STRING` in the current 4.2.0 release that needs to be resolved first. @dellaert proposed to resolve it in 4.2.1.

varunagrawal:
> Why is target not develop?

This PR targets develop now.

talregev:
This PR is for dev. Can you fix it?

varunagrawal:
> @varunagrawal Could you create a draft PR in gtsam-manylinux-build that target this branch and test if the workflow still pass?

@ProfFan here you go: https://github.com/borglab/gtsam-manylinux-build/pull/42

varunagrawal:
@dellaert @ProfFan this is ready for review whenever you get a chance.

python/CMakeLists.txt
```diff
@@ -1,16 +1,23 @@
+set(PROJECT_PYTHON_SOURCE_DIR ${PROJECT_SOURCE_DIR}/python)
 set(GTSAM_PYTHON_BUILD_DIRECTORY ${PROJECT_BINARY_DIR}/python)
 
 if (NOT GTSAM_BUILD_PYTHON)
     return()
 endif()
 
+# Install development dependencies to build wrapper
```
ProfFan: We really should not do this in CMakeLists, should be in the installation instructions instead

python/README.md
```diff
@@ -16,7 +16,7 @@ For instructions on updating the version of the [wrap library](https://github.co
 - This wrapper needs `pyparsing(>=2.4.2)`, and `numpy(>=1.11.0)`. These can be installed as follows:
 
   ```bash
-  pip install -r <gtsam_folder>/python/requirements.txt
+  pip install -r <gtsam_folder>/python/dev_requirements.txt
```
ProfFan, line 19: 👍

ProfFan (approved):
minus the comments

python/CMakeLists.txt
varunagrawal: I agree, though I was wondering about this since @dellaert prefers things being automated. I'll remove this for now and update the README, and let Frank leave a comment when he gets a chance. We can always add it in a follow-up PR.

PR #1630, opened 2023-09-07
Fix windows tests

talregev:
Fix windows tests

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -150,7 +150,10 @@ jobs:
           # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.dynamics_unstable
           # Compile. Fail with exception
           # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.nonlinear_unstable
-          # Compilation error
+          # Compile. Fail with exception
           # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.slam_unstable
-          # Compilation error
+          # Compile. Fail with exception
           # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.partition
+
+          # Run all tests
+          # cmake --build build -j1 --config ${{ matrix.build_type }} --target check
--- a/gtsam/discrete/AlgebraicDecisionTree.h
+++ b/gtsam/discrete/AlgebraicDecisionTree.h
@@ -36,7 +36,7 @@ namespace gtsam {
    * @ingroup discrete
    */
   template <typename L>
-  class GTSAM_EXPORT AlgebraicDecisionTree : public DecisionTree<L, double> {
+  class AlgebraicDecisionTree : public DecisionTree<L, double> {
     /**
      * @brief Default method used by `labelFormatter` or `valueFormatter` when
      * printing.
--- a/gtsam_unstable/partition/tests/CMakeLists.txt
+++ b/gtsam_unstable/partition/tests/CMakeLists.txt
@@ -1,6 +1,6 @@
 set(ignore_test "testNestedDissection.cpp")
 
-if (NOT GTSAM_USE_BOOST_FEATURES)
+if (NOT GTSAM_USE_BOOST_FEATURES OR MSVC)
 	list(APPEND ignore_test "testFindSeparator.cpp")
 endif()
 
--- a/gtsam/linear/SubgraphBuilder.h
+++ b/gtsam/linear/SubgraphBuilder.h
@@ -182,7 +182,7 @@ GaussianFactorGraph buildFactorSubgraph(const GaussianFactorGraph &gfg,
 
 /** Split the graph into a subgraph and the remaining edges. 
  * Note that the remaining factorgraph has null factors. */
-std::pair<GaussianFactorGraph, GaussianFactorGraph> splitFactorGraph(
+std::pair<GaussianFactorGraph, GaussianFactorGraph> GTSAM_EXPORT splitFactorGraph(
     const GaussianFactorGraph &factorGraph, const Subgraph &subgraph);
 
 }  // namespace gtsam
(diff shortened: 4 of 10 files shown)


talregev:
This PR is fix all the compilation errors from all windows tests.
There is 7 tests are failing from exceptions or correction of the tests that need to fix in different PR.
@dellaert can you take a look?

.github/workflows/build-windows.yml
```diff
@@ -141,16 +141,19 @@ jobs:
           # Run GTSAM_UNSTABLE tests
           cmake --build build -j4 --config ${{ matrix.build_type }} --target check.base_unstable
           # Compile. Fail with exception
-          # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.geometry_unstable
+          # cmake --build build -j1 --config ${{ matrix.build_type }} --target check.geometry_unstable
```
varunagrawal: Any reason you changed this to j1 instead of j4? We would like to the CI to run as fast as possible.

.github/workflows/build-windows.yml
```diff
@@ -141,16 +141,19 @@ jobs:
...
           # Compile. Fail with exception
-          # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.linear_unstable
+          # cmake --build build -j1 --config ${{ matrix.build_type }} --target check.linear_unstable
```
varunagrawal: Additionally, these test commands have not been uncommented. We would like to see all these tests passing.

gtsam/discrete/AlgebraicDecisionTree.h
```diff
@@ -36,7 +36,7 @@ namespace gtsam {
    * @ingroup discrete
    */
   template <typename L>
-  class GTSAM_EXPORT AlgebraicDecisionTree : public DecisionTree<L, double> {
+  class AlgebraicDecisionTree : public DecisionTree<L, double> {
```
varunagrawal, line 39: I'm a bit confused by this change. Don't the gtsam.discrete tests already pass?

varunagrawal (changes requested):
Thanks for the PR @talregev. Left some comments.

.github/workflows/build-windows.yml
talregev: These tests are not passing, that why they on comments. On each group of tests, there is some tests that not passing because of the correctness of the tests. I also add on each group why it on comment. (The same reason for all).

.github/workflows/build-windows.yml
talregev: Yes, some of the tests are not compiling with -j4 due memory allocation or just hang. Sometimes we get a weak vm. (not all vms are equal).

gtsam/discrete/AlgebraicDecisionTree.h
talregev: even if the tests are passing, it shouldn't be added this export, it will be create a linkage error when using `AlgebraicDecisionTree` in different combination.

varunagrawal: I see. Can you please give an example so I can better understand this? That way we can come up with a good solution together.

talregev: I don't have an example for `AlgebraicDecisionTree` class. But you can see other template class here that I remove the export, bring them back and see the linkage error they will bring (they will usually bring errors on different classes that why it hard to find it).
So the thumb rule here: for template class it shouldn't be with export. (Also you can see it work without it). and regular class if use it, it need an export, the same go for functions.
I hope it more clear.
Any other comments overall?

talregev:
@varunagrawal Can you review it?

.github/workflows/build-windows.yml
varunagrawal: Interesting. Could we instead use 2 instead of 1? That way we have some parallel execution.

.github/workflows/build-windows.yml
varunagrawal: If these tests are commented out, there's no point in making changes to them right? We can change these when we uncomment them. :-) Currently this only makes the PR hard to review :-(

gtsam_unstable/partition/tests/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 set(ignore_test "testNestedDissection.cpp")
 
-if (NOT GTSAM_USE_BOOST_FEATURES)
+if (NOT GTSAM_USE_BOOST_FEATURES OR MSVC)
```
varunagrawal, line 3: This doesn't seem right to me. Why are Boost and the use of MSVC correlated?

varunagrawal (commented):
I need to evaluate these changes in more detail on my windows machine. Hope to get to that soon.

gtsam_unstable/partition/tests/CMakeLists.txt
talregev: It correlated to `testFindSeparator.cpp`
As I mention before, the library `metis` that you added manually to gtsam:
https://github.com/borglab/gtsam/tree/develop/gtsam/3rdparty/metis
You add it to Linux only. It cannot compile for windows. So I ignore this test for windows. 
You should not add 3rd library like that, and you should look in solutions like vpckg (the PR I suggested).
We can talk about metis again, on that PR when I succeeded to compile it for windows with vcpkg, and we can fix this test there.
But for now, until vcpkg PR, you cannot compile it with windows. You copy only the metis linux headers, that different from metis windows headers.

talregev:
I will convert it to -j4 as before. it not effect on this PR anyway because it comments. So we can talk about it next PR when we fix correctness windows tests.

.github/workflows/build-windows.yml
talregev: I will convert it back to -j4 because it commented

.github/workflows/build-windows.yml
talregev: I will convert it back to -j4 because it commented

talregev:
@varunagrawal Can you review my PR?

varunagrawal (approved):
Okay looks good except for the metis CMake check, which we can re-evaluate in the vcpkg PR.

talregev:
Thank you!

PR #1640, opened 2023-09-20
Various Fixes and TODOs

varunagrawal:
1. Actually use the `s` argument for `DiscreteKeys::print`.
2. Address TODOs in hybrid machinery.
3. Use `set_difference` in `DecisionTreeFactor` to improve time complexity from `O(M*N)` to `max{O(MlogM), O(NlogN)}` aka significant savings! M and N are the sizes of the two key containers.
4. Improved docstrings in `HybridGaussianISAM`.
5. Enable a test in `testDiscreteMarginals`.
6. Remove unused TBB CMake variable which was a part of the (now-deleted) `FindTBB.cmake` file.
7. Added `Testable` traits for `DiscreteBayesTree` and `HybridBayesTree`.
8. Wrap printing of `Sim2` and `Sim3` and update docstrings.


talregev:
can you also fix windows tests?
My PR is not complete.

.github/workflows/build-linux.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           ubuntu-22.04-clang-14,
         ]
 
-        build_type: [Release]
+        build_type: [Debug, Release]
```
dellaert: wow !!! Don't slip this in, double CI time?

gtsam/basis/Chebyshev2.cpp
```diff
@@ -32,7 +32,7 @@ Weights Chebyshev2::CalculateWeights(size_t N, double x, double a, double b) {
     const double dj =
         x - Point(N, j, a, b);  // only thing that depends on [a,b]
 
-    if (std::abs(dj) < 1e-10) {
+    if (std::abs(dj) < 1e-12) {
```
dellaert: What problem are you solving with this behavior change?

gtsam/basis/Chebyshev2.h
```diff
@@ -51,27 +51,30 @@ class GTSAM_EXPORT Chebyshev2 : public Basis<Chebyshev2> {
...
-    // sin(- M_PI_2 + dtheta*j); also works
-    return cos(-M_PI + dtheta * j);
-  }
-
-  /// Specific Chebyshev point, within [a,b] interval
-  static double Point(size_t N, int j, double a, double b) {
-    assert(j >= 0 && size_t(j) < N);
-    const double dtheta = M_PI / (N - 1);
-    // We add -PI so that we get values ordered from -1 to +1
+    // sin(-M_PI_2 + dtheta*j); also works
```
dellaert: was this not a carefully designed first approximation to ensure correct behavior around those singularities?

dellaert (commented):
Some comments, as some "small" changes require more thought and might have to be proposed in a different PR

.github/workflows/build-linux.yml
varunagrawal: It's a parallel job so it shouldn't add any more time asymptotically.

I was actually surprised when I saw this was taken out. Apparently it was done to speed up CI for some feature you and the folks at Verdant were working on (eliminating Boost etc).

gtsam/basis/Chebyshev2.cpp
varunagrawal: I was trying to debug some odd behavior I was seeing with the Chebyshev differentiation matrix, so I tightened these screws a bit.

I can put this in a separate PR. I discovered an amazing book by Tobin Driscoll (Trefethen's former student and now professor) that shows how to compute the diff matrix twice as fast, and I want to use that instead.

gtsam/basis/Chebyshev2.h
varunagrawal: I don't think so. I looked at this math again when I made the change and the $\pi/(N-1)$ is the main part. We add the $\pi/2$ term so that we go from the [1, -1] interval (as in the book) to the [-1, 1] interval.

Finally when we set a=-1 and b=1, we get back the original equation on the standard interval [-1,1].

.github/workflows/build-linux.yml
dellaert: Hmmm. Please take it out and do in a separate PR, back up your hypothesis with data. Don't slip it in

gtsam/basis/Chebyshev2.cpp
dellaert: We can take it up in a separate PR.

dellaert (approved):
will merge, LGTM!

PR #1650, opened 2023-10-06
Templated `FactorGraph::at` method

varunagrawal:
I added a templated `at` method to the `FactorGraph` base class, so we can switch from doing
```cpp
auto factor = std::dynamic_pointer_cast<HessianFactor>(graph.at(0));
```
to
```cpp
auto factor = graph.at<HessianFactor>(0);
```
which IMHO looks much cleaner and is easier to type out.

I return the shared pointer by value following [this stackoverflow answer](https://stackoverflow.com/a/10643624/1236990).

### DIFF
--- a/gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
+++ b/gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
@@ -42,8 +42,8 @@
 #include <functional>
 #include <iostream>
 #include <iterator>
-#include <vector>
 #include <numeric>
+#include <vector>
 
 #include "Switching.h"
 #include "TinyHybridExample.h"
@@ -613,11 +613,11 @@ TEST(HybridGaussianFactorGraph, assembleGraphTree) {
   // Create expected decision tree with two factor graphs:
 
   // Get mixture factor:
-  auto mixture = std::dynamic_pointer_cast<GaussianMixtureFactor>(fg.at(0));
+  auto mixture = fg.at<GaussianMixtureFactor>(0);
   CHECK(mixture);
 
   // Get prior factor:
-  const auto gf = std::dynamic_pointer_cast<HybridConditional>(fg.at(1));
+  const auto gf = fg.at<HybridConditional>(1);
   CHECK(gf);
   using GF = GaussianFactor::shared_ptr;
   const GF prior = gf->asGaussian();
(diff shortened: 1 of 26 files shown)


dellaert:
Why not to develop?

varunagrawal:
I'm cascading PRs so I can get all these features to work together for my stuff (in contrast to having to switch branches and compile).

Having it target the parent branch makes reviewing easier for you. :)

varunagrawal:
@dellaert this PR has become tricky to review, but the change is exactly what the PR comment says it is.

gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
```diff
@@ -613,11 +613,11 @@ TEST(HybridGaussianFactorGraph, assembleGraphTree) {
   // Create expected decision tree with two factor graphs:
 
   // Get mixture factor:
-  auto mixture = std::dynamic_pointer_cast<GaussianMixtureFactor>(fg.at(0));
+  auto mixture = fg.at<GaussianMixtureFactor>(0);
```
varunagrawal, line 616: This is an example of the usage working for a test.

gtsam/inference/FactorGraph.h
```diff
@@ -310,6 +310,21 @@ class FactorGraph {
...
+  template <typename F>
+  std::shared_ptr<F> at(size_t i) {
+    return std::dynamic_pointer_cast<F>(factors_.at(i));
+  }
+
+  /// Const version of templated `at` method.
+  template <typename F>
+  const std::shared_ptr<F> at(size_t i) const {
+    return std::dynamic_pointer_cast<F>(factors_.at(i));
+  }
```
varunagrawal, line 326: This part is the main contribution of this PR. Every other change that was originally part of this PR was an application of this new method.

dellaert (commented):
So, again a lot of cmake changes in here that I don't fully understand are behavior changes. Can we focus the PR on the templated at method?

varunagrawal:
I'll reopen those other PRs and undo the merge commits. That should resolve this.

varunagrawal:
@dellaert updated

dellaert (approved):
Cool, nice PR now that all clutter is gone :-) Great new functionality, I like it a lot!

PR #1651, opened 2023-10-08
CMake Cleanup

varunagrawal:
Updated the GTSAM cmake so that GTSAM specific flags are not exposed in downstream applications. The benefit is that configuring downstream applications becomes easier and more legible, without all the extra variables involved.

For example, this is the ccmake output of GTDynamics before:

Screen Shot 2023-10-08 at 11 27 17 AM
Screen Shot 2023-10-08 at 11 27 34 AM

<hr/>

and after:
Screen Shot 2023-10-08 at 11 25 53 AM
Screen Shot 2023-10-08 at 11 28 54 AM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,19 +42,19 @@ set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" "${CMAKE_CURRENT_SOURCE_DIR}/cmake"
 include(GtsamMakeConfigFile)
 include(GNUInstallDirs)
 
-# Load build type flags and default to Debug mode
-include(GtsamBuildTypes)
-
-# Use macros for creating tests/timing scripts
-include(GtsamTesting)
-include(GtsamPrinting)
-
 # guard against in-source builds
 if(${GTSAM_SOURCE_DIR} STREQUAL ${GTSAM_BINARY_DIR})
   message(FATAL_ERROR "In-source builds not allowed. Please make a new directory (called a build directory) and run CMake from there. You may need to remove CMakeCache.txt. ")
 endif()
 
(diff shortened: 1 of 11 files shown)


jlblancoc (approved):
So, as far as I can see, this is more about integrating GTSAM in another project (e.g. submodule) so the private ccmake vars are not shown (but marked as private), rather than hiding those private flags downstream, when they should already be absent via the use of "PRIVATE", right?

If that's the idea, I see it correct and well done. From a first read of the PR description it looked like private flags were exposed downstream (!!).

varunagrawal:
> So, as far as I can see, this is more about integrating GTSAM in another project (e.g. submodule) so the private ccmake vars are not shown (but marked as private), rather than hiding those private flags downstream, when they should already be absent via the use of "PRIVATE", right?

Yes, some flags which are considered "private" such as `GTSAM_COMPILE_OPTIONS_PRIVATE` were being exposed in a downstream library (GTDynamics in this case) which didn't seem correct, since why should GTDynamics care about GTSAM's internal compilation flags? This was also causing too many variables to populate which makes configuring the downstream project harder.

Since `set` doesn't have a "PRIVATE" option, I used `mark_as_advanced` to achieve this.

> If that's the idea, I see it correct and well done.

Thank you. :)

> From a first read of the PR description it looked like private flags were exposed downstream (!!).

Yes that is exactly correct. I am assuming flags like `GTSAM_COMPILE_OPTIONS_PRIVATE` are supposed to be private @jlblancoc? And not exposed downstream? I took screenshots showing the before and after so that it was clearer what I was trying to achieve.

dellaert:
Why not to develop?

varunagrawal:
I'm chaining PRs to get the combined features but still make it easy to review. :)

PR #1658, opened 2023-10-19
Re-enable Debug CI for Linux

varunagrawal:
The Debug build was removed in a previous PR for some reason, so I am reintroducing it here to cover all our bases.

Currently, the Linux CI takes [39m 38s](https://github.com/borglab/gtsam/actions/runs/6551953149?pr=1640) to run. I will update this comment with the time taken when the Debug build has completed running to provide evidence that adding the Debug build doesn't add significant more CI time.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -31,7 +31,7 @@ jobs:
           ubuntu-22.04-clang-14,
         ]
 
-        build_type: [Release]
+        build_type: [Debug, Release]
         build_unstable: [ON]
         include:
           - name: ubuntu-20.04-gcc-9


varunagrawal:
Here are the times taken:

<table>
<tr>
 <th> CI Name
 <th> Release Time Taken
 <th> Debug Time Taken
<tr>
 <td> ubuntu-20.04-gcc-9
 <td> 47m 15s
 <td> 50m 56s
<tr>
 <td> ubuntu-20.04-clang-9
 <td> 33m 34s
 <td> 33m 36s
<tr>
 <td> ubuntu-22.04-gcc-11
 <td> 44m 00s
 <td> 44m 49s
<tr>
 <td> ubuntu-22.04-clang-14
 <td> 32m 01s
 <td> 30m 27s
</table>

The maximum time difference is 3 minutes, which is a reasonable gap.

dellaert:
BUt, we run *both* release and debug right? So time doubles? Perhaps I am misunderstanding.

varunagrawal:
Yes we run both but in separate VM instances, thus they run in parallel. This is the reason we have the Job Matrix in the Actions YAML file, so GitHub knows how to run them concurrently.

The only bottleneck is when GitHub/Azure (because Microsoft owns GitHub) can't find a free VM and is waiting on resources, but GitHub seems to have improved that massively in the past year.

dellaert (approved):

PR #1660, opened 2023-10-23
Chebyshev2 Improvements

varunagrawal:
According to Trefethen, given N, we should form a polynomial with N+1 nodes, but we form a polynomial with N nodes (which is mathematically a N-1 degree polynomial).

Hence one of the bigger changes of this PR is to form a polynomial over N+1 nodes. This makes our code correspond directly to the math and makes comparison to Chebfun much easier.

Additionally, I've tightened some bounds checks and improved the overall API.


dellaert:
I wish you had done a design review with me first. I don’t support the change in the meaning of N without extensive discussion…

dellaert (changes requested):
Let’s talk first

varunagrawal:
> I don’t support the change in the meaning of N without extensive discussion…

I've been using Chebfun and the mental gymnastics of relating that to our implementation was getting to me. I believe we were wrong to have polynomials of degree N (instead of N+1) for the specified degree N.

varunagrawal:
@dellaert as discussed offline, I have reverted the change to go back to using order N polynomials (degree N-1) instead of order N+1 (degree N).

gtsam/basis/Chebyshev2.h
```diff
@@ -83,6 +86,11 @@ class GTSAM_EXPORT Chebyshev2 : public Basis<Chebyshev2> {
     return points;
   }
 
+  /// Return a zero initialized Parameter matrix.
+  static Parameters ParameterMatrix(size_t N) {
```
dellaert: I don't see why we can't just say Zero(N) ?

gtsam/basis/Chebyshev2.h
```diff
@@ -94,6 +102,16 @@ class GTSAM_EXPORT Chebyshev2 : public Basis<Chebyshev2> {
...
                                   double b = 1);
 
+  /**
+   * Calculate weights for all x in vector X.
+   * Returns M*N matrix where M is the size of the vector X,
+   * and N is the number of basis functions.
+   *
+   * Overriden for Chebyshev2.
+   */
+  static Matrix WeightMatrix(size_t N, const Vector& X, double a = -1,
```
dellaert: Where is this used? Just being worried about this MxN matrix being confused with the other MxN matrix that is used for VectorEvaluators. But I remember we talked about this as a speedup in some scenarios?

dellaert (commented):
Some questions. I don't see ParameterMatrix as an added value, just more code to maintain?

gtsam/basis/Chebyshev2.h
varunagrawal: Yes this needs to be removed since it was added for the previous change of `N` to `N+1`. Removing.

gtsam/basis/Chebyshev2.h
varunagrawal: This needs to be removed too.

varunagrawal:
@dellaert okay addressed all comments. :)

dellaert (approved):

PR #1663, opened 2023-10-29
Fix apparent non-templated types in expressions

jlblancoc:
Please, check this out... I think it's a typo, isn't it?


varunagrawal (approved):
I'm surprised this wasn't found sooner. Thank you @jlblancoc

PR #1669, opened 2023-11-13
`DiscreteFactor::errorTree` method for all assignments

varunagrawal:
Add a new ~~`error`~~ `errorTree` method for `DiscreteFactor`s which returns a full `AlgebraicDecisionTree` of errors for each assignment.

This is being done in an effort to add a `printErrors` method to `HybridNonlinearFactorGraph`.


dellaert (approved):
LGTM !!!

PR #1678, opened 2023-11-28
Add reprojection with fisheye camera model

kkelchte:
Add the ProjectionFactorPPPCal3Fisheye factor reprojection for the fisheye camera in python interface.


varunagrawal:
@kkelchte can you please add a unit test in the `python/gtsam/tests` directory to verify that the wrapping is indeed successful? It can be very simple, similar to a Cal3_S2 example.

dellaert (approved):
Thanks !!!!

dellaert:
I think this is just adding a type to the wrapper, does not need unit test.

PR #1706, opened 2024-01-07
Run python tests with timeout

talregev:
Run python tests with timeout.
Recently we can see more and more on every PR that python with tbb fail. it hang for 6 hours until it fail. It is waste of time and the limited resource of the CI.
Also we can see which test is hang on. 

Apparently the tests are not hang, and it the github action step that hang. 
https://github.com/borglab/gtsam/actions/runs/7440972762/job/20242575450?pr=1706
And it a known issue:
https://github.com/actions/runner/issues/1326

Maybe memory issue. Try 10 gb swap.


talregev:
It do not solve the problem. I see it hang on my CI.
I can revert to 6gb swap memory.
Do you want to timeout functionality? If not, I will close this PR.

varunagrawal:
We don't need the timeout functionality. Thanks for investigating though.

PR #1707, opened 2024-01-08
ISAM2: Add variables before computing delta

### LINKED ISSUE #301: Bug while using dogleg optimizer
I have created an unit test (`examples/SFMExample_SmartFactor_bug.cpp`) to illustrate the issue and can be found in my [fork](https://github.com/kvmanohar22/gtsam/tree/dogleg_bug)

In the discussion in [issue:367](https://bitbucket.org/gtborg/gtsam/issues/367/isam2-with-smart-factors-null-ptr-segfault), this bug was resolved but seems like it is popping up again.

## Description

This error seems to pop up when using Dogleg optimizer as part of ISAM2. Following is the reason for this error.

Assume `graph_` represents the existing graph (passed previously to `gtsam::ISAM2`) that has smart factors. Suppose a smart factor was already in the previous graph and we update the smart factor with a new measurement from a pose (this pose was not part of previous graph).

Assuming that the following condition holds true ([line](https://github.com/borglab/gtsam/blob/b4eb0c233d683eaf47511ee9b062dc3c822484df/gtsam/nonlinear/ISAM2.cpp#L424))
```cpp
  // Update delta if we need it to check relinearization later
  if (update.relinarizationNeeded(update_count_))
    updateDelta(updateParams.forceFullSolve);
```
the error can be reproduced. Specifically, the error arises from the following [line](https://github.com/borglab/gtsam/blob/b4eb0c233d683eaf47511ee9b062dc3c822484df/gtsam/nonlinear/ISAM2.cpp#L746) which tries to compute the error of the previous graph but without updating `theta_`

I have tried to simulate this in the unit test.

```cpp
nonlinearFactors_.error(theta_),
```

## Steps to reproduce

Just run the new unit test that I have added (in my fork) using

```bash
make SFMExample_SmartFactor_bug.run
```
which results in the following error

```bash
kv@kv:~/slam/gtsam/build$ make SFMExample_SmartFactor_bug.run
[ 33%] Built target metis
[100%] Built target gtsam
[100%] Building CXX object examples/CMakeFiles/SFMExample_SmartFactor_bug.dir/SFMExample_SmartFactor_bug.cpp.o
[100%] Linking CXX executable SFMExample_SmartFactor_bug
[100%] Built target SFMExample_SmartFactor_bug
Initial Estimates:
Values with 3 values:
Value 0: (N5gtsam5Pose3E) R:
[
 	 0.208654,  0.073369, -0.975233;
    	 0.949229, -0.255245,  0.183888;
    	-0.235432, -0.964089, -0.122902
  ]
[29.8, 0.05, 0.1]';

Value 1: (N5gtsam5Pose3E) R:
[
 	-0.523665,  0.232365, -0.819623;
    	 0.818747, -0.128606, -0.559566;
    	-0.235432, -0.964089, -0.122902
  ]
[21.0364, 21.1071, 0.1]';

Value 2: (N5gtsam5Pose3E) R:
[
 	-0.949229,  0.255245, -0.183888;
    	 0.208654,  0.073369, -0.975233;
    	-0.235432, -0.964089, -0.122902
  ]
[-0.05, 29.8, 0.1]';

Results:
Values with 3 values:
Value 0: (N5gtsam5Pose3E) R:
[
 	-0.0253294,  0.0208786,  -0.999461;
    	  0.998264, -0.0526448, -0.0263988;
    	-0.0531676,  -0.998395, -0.0195089
  ]
[30.0013, -0.000248155, 0.00250433]';

Value 1: (N5gtsam5Pose3E) R:
[
 	  -0.709983,   0.0133575,   -0.704092;
    	   0.704144, -0.00108604,   -0.710056;
    	 -0.0102493,    -0.99991, -0.00863457
  ]
[21.2138, 21.2135, 0.00132106]';

Value 2: (N5gtsam5Pose3E) R:
[
 	 -0.999205, -0.0254096,  0.0307042;
    	-0.0320221,  0.0531939,  -0.998071;
    	 0.0237273,  -0.998261, -0.0539653
  ]
[-2.03514, 30.1004, 1.98536]';

terminate called after throwing an instance of 'gtsam::ValuesKeyDoesNotExist'
  what():  Attempting to at the key "3", which does not exist in the Values.
Aborted (core dumped)
examples/CMakeFiles/SFMExample_SmartFactor_bug.run.dir/build.make:57: recipe for target 'examples/CMakeFiles/SFMExample_SmartFactor_bug.run' failed
make[3]: *** [examples/CMakeFiles/SFMExample_SmartFactor_bug.run] Error 134
CMakeFiles/Makefile2:2660: recipe for target 'examples/CMakeFiles/SFMExample_SmartFactor_bug.run.dir/all' failed
make[2]: *** [examples/CMakeFiles/SFMExample_SmartFactor_bug.run.dir/all] Error 2
CMakeFiles/Makefile2:2667: recipe for target 'examples/CMakeFiles/SFMExample_SmartFactor_bug.run.dir/rule' failed
make[1]: *** [examples/CMakeFiles/SFMExample_SmartFactor_bug.run.dir/rule] Error 2
Makefile:884: recipe for target 'SFMExample_SmartFactor_bug.run' failed
make: *** [SFMExample_SmartFactor_bug.run] Error 2

```

## Expected behavior
The error should not have popped up

## Environment
OS: Ubuntu 16.04
Language: C++, custom build

## Additional information
This bug was discussed previously in the following discussion. [link](https://bitbucket.org/gtborg/gtsam/issues/367/isam2-with-smart-factors-null-ptr-segfault), specifically comment by Sean Bowman.

### LINKED ISSUE #1672: `Rot2::fromCosSin` normalizes when header indicates it should not
## Description
In the declaration of `Rot2::fromCosSin`, the header indicates normalization does *not* happen:

https://github.com/borglab/gtsam/blob/1121ece0eb16324c8e1da5247f5abc62a611252e/gtsam/geometry/Rot2.h#L71-L72

But the implementation normalizes the return value:

https://github.com/borglab/gtsam/blob/1121ece0eb16324c8e1da5247f5abc62a611252e/gtsam/geometry/Rot2.cpp#L27-L30

Is this intended? It's possible I'm misunderstanding the use case of `Rot2::fromCosSin` or what "normalization" means here.

## Expected behavior
The returned object from `Rot2::fromCosSin` does not normalize the given `c` and `s`, even if they are denormalized.

## Environment
C++

## Additional information
I'm not familiar enough with the library to understand the full impact of changing this function to strictly `return Rot2(c, s)`.

The introduction of this potential bug goes waaaay back to e2728184b98b3e015de31411c60045cd20e1ba73, in 2010.

varunagrawal:
Fixes #301 

A `gtsam::ValuesKeyDoesNotExist` error is thrown from the function `nonlinearFactors_.error(theta_)`.
The issue is that `nonlinearFactors_.error(theta_)` is computed as an argument to `DoglegOptimizerImpl::Iterate`, where a factor with the new key exists but `theta_` has not been updated with the new key-value pairs yet.

It is a simple quirk that Dogleg requires the updated Values since it needs them to compute the delta update via the dogleg point.

The issue in `ISAM2::update` is that the delta update is computed first to check for relinearization before the new factors and Values are added. The delta update calls the above `nonlinearFactors_.error(theta_)` method with missing keys in `theta_` leading to the exception.

Thus the fix is to add in the new variables `newTheta` with `addVariable(newTheta)` before the call to `updateDelta(updateParams.forceFullSolve);`. AFAIAA this step is to do with fluid relinearization and shouldn't affect the LM path since we call `addVariable(newTheta)` before the actual update step in both the LM and Dogleg cases.

Update: Piggybacking on this PR to also fix #1672.

### DIFF
--- a/tests/testDoglegOptimizer.cpp
+++ b/tests/testDoglegOptimizer.cpp
@@ -184,6 +184,14 @@ TEST(DoglegOptimizer, Constraint) {
 }
 
 /* ************************************************************************* */
+/**
+ * Test created to fix issue in ISAM2 when using the DogLegOptimizer.
+ * Originally reported by kvmanohar22 in issue #301
+ * https://github.com/borglab/gtsam/issues/301
+ *
+ * This test is based on a script provided by kvmanohar22
+ * to help reproduce the issue.
+ */
 TEST(DogLegOptimizer, VariableUpdate) {
   // Make the typename short so it looks much cleaner
   typedef SmartProjectionPoseFactor<Cal3_S2> SmartFactor;
--- a/tests/testDoglegOptimizer.cpp
+++ b/tests/testDoglegOptimizer.cpp
@@ -24,10 +24,9 @@
 #include <gtsam/nonlinear/DoglegOptimizerImpl.h>
 #include <gtsam/nonlinear/NonlinearEquality.h>
 #include <gtsam/slam/BetweenFactor.h>
-#include <gtsam/inference/Symbol.h>
-#include <gtsam/linear/JacobianFactor.h>
-#include <gtsam/linear/GaussianBayesTree.h>
-#include <gtsam/base/numericalDerivative.h>
+#include <gtsam/nonlinear/ISAM2.h>
+#include <gtsam/slam/SmartProjectionPoseFactor.h>
+#include "examples/SFMdata.h"
 
 #include <functional>
 
@@ -36,7 +35,6 @@ using namespace gtsam;
 
 // Convenience for named keys
 using symbol_shorthand::X;
-using symbol_shorthand::L;
 
 /* ************************************************************************* */
 TEST(DoglegOptimizer, ComputeBlend) {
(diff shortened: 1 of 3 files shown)


dellaert (commented):
Touching the internals of iSAM2 needs a bit more explanation. So, two requests:
- Outline the changes you made in iSAM2, and say why they were needed for the Dogleg path, and why the don't disturb the LM path. Just a matter of providing your understanding in the MR for tracking and reviewing purposes.
- if the test was grabbed from an external contributor, you might want to add comments in the code to that effect and also a link to the issue.

varunagrawal:
@dellaert I updated the PR comment to reflect the understanding behind the fix.

About the unit test, @kvmanohar22 was kind enough to create a script illustrating the problem, but I came up with the unit test for it based on my understanding of his script and debugging the issue. His script is instead based on the SFM SmartFactor example you wrote, so I am a bit tangled up about the credit assignment here.

tests/testDoglegOptimizer.cpp
```diff
@@ -185,6 +183,120 @@ TEST(DoglegOptimizer, Constraint) {
 #endif
 }
 
+/* ************************************************************************* */
+TEST(DogLegOptimizer, VariableUpdate) {
```
dellaert, line 195: Please document why this example was created and credit the reporter/contributor as well in helping to craft this unit test.

varunagrawal: Done

dellaert (approved):
Awesome, thanks both!

PR #1709, opened 2024-01-12
Change from biasAccOmegaInt to biasAccOmegaInit

varunagrawal:
Made these changes a few years ago, and wanted to create the PR for whenever it was deemed appropriate to merge.

This PR changes all instances of `biasAccOmegaInt` to `biasAccOmegaInit` which is the correct name for the initial noise on the bias for both the accelerometer and the gyroscope.


ProfFan (commented):
Wanted to do this for years, but I'll let Frank decide :P

ProfFan:
Ping @dellaert

dellaert:
Convince me that this is the correct change. There is also an integration covariance mentioned in the paper. Are we sure this is not that? How is an initialization part of a parameter?

varunagrawal:
`PreintegrationParams` has an `integrationCovariance` which is what is used for integration.

https://github.com/borglab/gtsam/blob/d48b1fc8402e28d5b6af8f59cce8dbe87e66cf0c/gtsam/navigation/CombinedImuFactor.cpp#L149

In that same file, we have `bInitCov` which is assigned from `p().biasAccOmegaInt`.

https://github.com/borglab/gtsam/blob/d48b1fc8402e28d5b6af8f59cce8dbe87e66cf0c/gtsam/navigation/CombinedImuFactor.cpp#L150

`biasAccOmegaInit` should be the initial covariance for the 6 dimensional bias term encapsulating both the acceleration and angular velocity (aka `ConstantBias`) and is used by the `CombinedImuFactor` for preintegrating and propagating the bias covariance.

dellaert:
Ok, you convinced me. But how could we make this change in a way that does not break existing user code?

varunagrawal:
> Ok, you convinced me. But how could we make this change in a way that does not break existing user code?

This has been the big question for me. I wonder if we can put in a deprecation warning like message that lets users know the variable name has changed? If we're cutting 4.3a as a new release, can we make an exception?

dellaert:
Typically, we bracket in ifdefs, so we can do that here as well?

dellaert:
@varunagrawal This parameter is now gone, as you know as you were involved in review, so I'll close this.

Issue #1710, opened 2024-01-12
Cephes requires Cmake 3.12, breaks build on Ubuntu Bionic


berndpfrommer:
The nightly Ubuntu Bionic builds break because the newly-added cephes library requires CMake 3.12 or higher
```
-- Found Eigen version: 3.3.4
CMake Error at gtsam/3rdparty/cephes/CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.12 or higher is required.  You are running version 3.10.2
```
Options:
1) Drop support for Ubuntu Bionic
2) Change required cmake version for libcephes to 3.10. I can build it locally on my machine without encountering errors. It is likely that libcephes does not rely on any features introduced between 3.10 and 3.12

varunagrawal:
Fixed by #1711. Closing, thanks @berndpfrommer

PR #1711, opened 2024-01-12
libcephes only needs 3.10, allowing build on ubuntu 18.04

berndpfrommer:
This PR changes the cmake requirements for libcephes from 3.12 to 3.10. This allows building on Ubuntu 18.04 (bionic), which only has cmake v 3.10.2.
This PR addresses issue #1710

I tested that GTSAM builds correctly on my local machine (which is Ubuntu 22.04, cmake 3.22.1).


varunagrawal (approved):
Thanks @berndpfrommer!

PR #1712, opened 2024-01-16
Update Wrapper

### LINKED ISSUE #1606: KeySet class in python is missing methods
## Feature

The `KeySet` class in python should have more methods available like `__iter__`, `__contains__`, `remove`, etc.

## Motivation

I'd like to be able to do things like:

```
for key in graph.keys():
    if key not in initial_estimate.keys():
        print(f"WARNING: missing key {key} in initial_estimate")
```

varunagrawal:
Update the wrapper with new functionality supporting a new double-underscore (dunder) syntax. This makes wrapping classes that should exhibit some behavior in Python easier.

E.g. there is an ask in #1606 for `KeySet` to allow iteration and other container based operations. By adding `__iter__` and `__contains__` dunder methods to the `gtsam.i` file, we now support those operations.
This means we can now do this:

```python
keys = graph1.keys()
for key in keys:
    if key in graph2.keys():
        print("key is in both graphs")
```

Consequently, fixes #1606.


gtsam/gtsam.i
```diff
@@ -39,6 +39,11 @@ class KeyList {
   void remove(size_t key);
 
   void serialize() const;
+
+  // Specual dunder methods for Python wrapping
```
gchenfc: Typo: "Specual" -> "Special"

gchenfc (approved):
LGTM!  Though I do think in gtwrap the `__contains__` should use `.find()` instead of `std::find` but 👍

gtsam/gtsam.i
varunagrawal: Corrected! I agree with the comment about `.find()` but until we move to C++20, many C++ STL containers won't support a `find` method. :(

PR #1713, opened 2024-01-16
Various Improvements

varunagrawal:
A bunch of improvements to the code, including formatting, more tests and better printing.


dellaert (approved):
LGTM!

dellaert:
You can cancel that last check and merge,  it takes hours. I opened a discussion on killing gcc9 altogether in CI.

varunagrawal:
> You can cancel that last check and merge, it takes hours. I opened a discussion on killing gcc9 altogether in CI.

Looks like the issue is gcc9 + TBB but only with the python wrapper. I think we can just get rid of checking with gcc9+TBB but keep gcc9.

PR #1714, opened 2024-01-17
Remove swap from python ci

talregev:
Remove swap from python ci
It not needed for the ci.
Also may fix the hang problem.

### DIFF
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -145,12 +145,6 @@ jobs:
           echo "GTSAM_WITH_TBB=ON" >> $GITHUB_ENV
           echo "GTSAM Uses TBB"
 
-      - name: Set Swap Space (Linux)
-        if: runner.os == 'Linux'
-        uses: pierotofy/set-swap-space@master
-        with:
-          swap-size-gb: 6
-
       - name: Install System Dependencies (Linux, macOS)
         if: runner.os != 'Windows'
         run: |


varunagrawal:
How do you know the swap space is not needed? Is there any analysis or memory dump showing this?

When we added the swap space, it was because the compilation would take too much memory, causing the VM to be killed. We have empirical evidence of that, however there is no evidence in this PR to the contrary.

talregev:
> How do you know the swap space is not needed? Is there any analysis or memory dump showing this?
> 
> When we added the swap space, it was because the compilation would take too much memory, causing the VM to be killed. We have empirical evidence of that, however there is no evidence in this PR to the contrary.

Because I compile on the ci without it, and the ci don't crash on gcc out of memory.

varunagrawal:
> Because I compile on the ci without it, and the ci don't crash on gcc out of memory.

This is not a very convincing argument. I can make the opposite claim, which means nothing without evidence.

varunagrawal (approved):
I am approving this, we can revert this if it causes any issues.

talregev:
> > Because I compile on the ci without it, and the ci don't crash on gcc out of memory.
> 
> This is not a very convincing argument. I can make the opposite claim, which means nothing without evidence.

This is valid argument, because I show on this PR how I remove the swap and it still running and compiling and running the tests.
If what you say it true, according to your argument it should crash on every time without swap.

What is really happen, when you did your measurements tests linux ci was with low memory for gtsam with tbb compilation.
From that time, the ci was upgraded, now you have stronger machine, and your hack for more memory with swap cause probably to the ci to hang and not needed anymore.

talregev:
> I am approving this, we can revert this if it causes any issues.

Thank you for approving. I agree. If we have issues we can later revet / fix them.

varunagrawal:
> This is valid argument, because I show on this PR how I remove the swap and it still running and compiling and running the tests. If what you say it true, according to your argument it should crash on every time without swap.

Sorry but this is not a valid argument. You may have just gotten lucky and the CI passed for you this time without the root cause being addressed.
 
> What is really happen, when you did your measurements tests linux ci was with low memory for gtsam with tbb compilation. From that time, the ci was upgraded, now you have stronger machine, and your hack for more memory with swap cause probably to the ci to hang and not needed anymore.

This is conjecture. Can you please provide evidence for this?

Going forward I won't be approving PRs that make hand-wavy claims without the necessary discussion and evidence to back it up.

PR #1413, opened 2023-01-24
Add vcpkg ci

talregev:
Add vcpkg ci. Fix #672

I try to minimize the changes as much as possible. 
I delete the `FindTBB.cmake`. I think it old file, and the default of the cmake is working well also on your tests. 
I also add to your workflow, that it will do test only when doing PR to devel and push to devel, as it your most use cases. 
Also I add concurrency section, that it cancel automatic other check when you push to your PR another commit before the other check is not finish. It not doing that when you merge or push to devel, because each change there is important. 
It useful to save resource and time for check your PRs.

Also there some bugs that I encounter that I think needed to fix in other PRs. 
This PR already have many things to discuss. 

This is start of infrastructure how to use vcpkg and can also extended (in other PRs). For example to publish binaries in the PR itself, that it can downloaded and test them externally. 
Maybe other use cases you want to test.
My moto is to compile as much as possible in one test (and hopefully to run the tests themself), to know much quicker to the problem when you doing a PR.

I am not sure why the 2 of your workflow is failing, If it from stuff I did, I will happy to know, and try to fix accordingly. 

gtsam bugs:
~~1. windows gtsam_unstable~~ Solve!
```
gtsam.lib(gtsam.dll) : error LNK2005: "public: __cdecl std::map<unsigned __int64,unsigned __int64,struct std::less<unsigned __int64>,class std::allocator<struct std::pair<unsigned __int64 const ,unsigned __int64> > >::~map<unsigned __int64,unsigned __int64,struct std::less<unsigned __int64>,class std::allocator<struct std::pair<unsigned __int64 const ,unsigned __int64> > >(void)" (??1?$map@_K_KU?$less@_K@std@@V?$allocator@U?$pair@$$CB_K_K@std@@@2@@std@@QEAA@XZ) already defined in LPInitSolver.obj [D:\a\gtsam\gtsam\build\gtsam_unstable\gtsam_unstable.vcxproj]
     Creating library D:/a/gtsam/gtsam/build/lib/Release/gtsam_unstable.lib and object D:/a/gtsam/gtsam/build/lib/Release/gtsam_unstable.exp
```
~~2. Errors on tests:~~ Solve!
```
/home/runner/work/gtsam/gtsam/gtsam/hybrid/tests/testHybridEstimation.cpp:221:8: error: ‘bitset’ is not a member of ‘std’
  221 |   std::bitset<K - 1> seq = x;
```
```
[...]

PR #1719, opened 2024-01-26
Ubuntu 22.04 for python tbb. hopefully will solve the hang problem.

talregev:
Change the os for python + tbb to be ubuntu 22.04. 
It may solve the hang problem.
The hang problem, is a know problem as we can see in this issue:
https://github.com/actions/runner/issues/1326
https://github.com/actions/runner/issues/2454

I am not sure what cause the hang problem. Maybe gtsam code, maybe the machine itself, or maybe python version.

But I want to try change the os. Better then to stay on this state. We can always revert this PR. 
I didn't do much tests (I run the CI on my end) or I don't have any prove, but I think the ubuntu 22.04 machine are stronger. Also maybe the os image github action is using for this os is better. 

Eventually the ci moving forward with the os version, and delete old version and get new one.
This can be tested only on python tbb to see his behavior for long run.

### DIFF
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -29,7 +29,7 @@ jobs:
         name:
           [
             ubuntu-20.04-gcc-9,
-            ubuntu-20.04-gcc-9-tbb,
+            ubuntu-22.04-gcc-9-tbb,
             ubuntu-20.04-clang-9,
             macOS-11-xcode-13.4.1,
             windows-2019-msbuild,
@@ -43,8 +43,8 @@ jobs:
             compiler: gcc
             version: "9"
 
-          - name: ubuntu-20.04-gcc-9-tbb
-            os: ubuntu-20.04
+          - name: ubuntu-22.04-gcc-9-tbb
+            os: ubuntu-22.04
             compiler: gcc
             version: "9"
             flag: tbb


varunagrawal:
Any reason you're not first testing this on your fork's CI to verify correctness?

varunagrawal (approved):

talregev:
> Any reason you're not first testing this on your fork's CI to verify correctness?

I first test it on my ci before I did a PR to here.
As I write on the first msg.


Thank you for your merge!

varunagrawal:
I verified this change in another PR on my local CI and I can confirm that this fixes the CI issue for the Python TBB. Thanks @talregev!

talregev:
@varunagrawal You already revert this commit.

Please consider again the changes here also:
https://github.com/borglab/gtsam/pull/1714

varunagrawal:
Yes, because I wanted to verify the charges one by one. The changes to the swap don't have any effect, so #1714 isn't needed.

I'll be adding the necessary changes in a subsequent PR.

talregev:
> The changes to the swap don't have any effect

That correct, and less code is not needed that it better. That mean that the ci don't need memory to compile tbb as it use to be. 
This is the point of that change. Remove unneeded changes.

varunagrawal:
We need the swap space for the wrapper to compile otherwise the process will sometimes get killed due to being out of memory. We don't add changes unless they are needed. I understand you don't have context for this.

talregev:
> We need the swap space for the wrapper to compile otherwise the process will sometimes get killed due to being out of memory. We don't add changes unless they are needed. I understand you don't have context for this.

Can you share a link to recent ci fail due to out of memory error?

varunagrawal:
This was from a few years ago. Sorry but I don't have the time to devote to this issue anymore.

talregev:
> This was from a few years ago. Sorry but I don't have the time to devote to this issue anymore.

Exactly my point. The issue with out memory is not relevant anymore.

varunagrawal:
> > This was from a few years ago. Sorry but I don't have the time to devote to this issue anymore.
> 
> Exactly my point. The issue with out memory is not relevant anymore.

I don't think you're referring to the same thing. The CI would kill the wrapper compilation due to lack of RAM, and ever since we increased the swap space we haven't had that issue. Until I see an analysis showing that the extra swap is no longer needed, I don't agree with removing it and potentially reintroducing the compilation memory issue.

The OOM issue this PR fixed is specific to the wrapper+TBB. This OOM issue seems to stem from running the tests. so at run time, not compile time.

Also, I think I accidentally deleted your last comment. I was on my phone when responding and might have swiped incorrectly. I have quoted your comment for posterity.

talregev:
> The CI would kill the wrapper compilation due to lack of RAM

This is the exact issue I am talking about. It happen on **compilation** wrapper + python in windows and not happen any more because microsoft upgrade the vms memory. The same apply for linux vms.
You can spot that I was report about this on this PR:
https://github.com/borglab/gtsam/pull/1413#issuecomment-1588546434

talregev:
> Until I see an analysis showing that the extra swap is no longer needed

You already see the CI run for 2-3 weeks (until you decide to remove this commit) with compile wrapper + tbb and not fail on lack of memory.

varunagrawal:
> > The CI would kill the wrapper compilation due to lack of RAM
> 
> This is the exact issue I am talking about. It happen on **compilation** wrapper + python in windows and not happen any more because microsoft upgrade the vms memory. The same apply for linux vms.
> You can spot that I was report about this on this PR:
> https://github.com/borglab/gtsam/pull/1413#issuecomment-1588546434

I'll need to see some proof of this, such as an upgrade notice from GitHub or Microsoft. Your comment is conjecture at this point.

PR #1728, opened 2024-02-29
Close unmatched endif

stevenlovegrove:
Building with arguments:

CMAKE_BUILD_TYPE=Release
GTSAM_ENABLE_BOOST_SERIALIZATION=FALSE
GTSAM_THROW_CHEIRALITY_EXCEPTION=FALSE
GTSAM_BUILD_EXAMPLES_ALWAYS=FALSE
-Wno-dev

Was failing on MacOS with clang with an unmatched ifdef error on head of devel. This PR makes the appropriate simple fix.

### DIFF
--- a/gtsam/base/std_optional_serialization.h
+++ b/gtsam/base/std_optional_serialization.h
@@ -12,8 +12,6 @@
 
 // Defined only if boost serialization is enabled
 #ifdef GTSAM_ENABLE_BOOST_SERIALIZATION
-// Only for old boost
-#if BOOST_VERSION < 108000
 #include <optional>
 #include <boost/config.hpp>
 
@@ -108,5 +106,4 @@ void serialize(Archive& ar, std::optional<T>& t, const unsigned int version) {
 }  // namespace serialization
 }  // namespace boost
 #endif // BOOST_VERSION < 108400
-#endif // BOOST_VERSION < 108000
 #endif // GTSAM_ENABLE_BOOST_SERIALIZATION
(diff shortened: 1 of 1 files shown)


ProfFan (approved):

ProfFan:
Thank you!

gtsam/base/std_optional_serialization.h
```diff
@@ -107,5 +107,6 @@ void serialize(Archive& ar, std::optional<T>& t, const unsigned int version) {
 
 }  // namespace serialization
 }  // namespace boost
-#endif
-#endif
+#endif // BOOST_VERSION < 108400
+#endif // BOOST_VERSION < 108000
```
varunagrawal: @stevenlovegrove can you please kill this endif and its corresponding #ifdef?

varunagrawal (changes requested):

gtsam/base/std_optional_serialization.h
stevenlovegrove: > @stevenlovegrove can you please kill this endif and its corresponding #ifdef?

There are two endif's there, and they are guarding different code blocks. Which do you want to kill? Both? I am not familiar enough with Boost to know how widely used these versions are to know the impact of removing the guards.

varunagrawal: The issue being fixed requires the 108400 block and not the 108000 block, hence we can kill the 108000 block without any impact.

varunagrawal (approved):

PR #1735, opened 2024-03-18
Fixing strength of bias constraints in IMUKittiExampleGPS.cpp

MichalNowicki:
This is a follow-up of a question at gtsam's Google [group](https://groups.google.com/g/gtsam-users/c/cc56DoPbxSA).

`included_imu_measurement_count` variable was not zeroed making each consecutive bias constraint weaker. This PR zeroes it when a new node is placed in the graph. For default parameters, `included_imu_measurement_count` is equal to 100 for all added bias constraints.

This change impacts the final estimate, which I am not sure if was ever compared to ground truth.

Previously, State at 469:
`t:   43.0073055   74.1530101 -0.426735268`

Currently, State at 469:
`t:  44.1523951  74.9626438 0.170519702`

### DIFF
--- a/examples/IMUKittiExampleGPS.cpp
+++ b/examples/IMUKittiExampleGPS.cpp
@@ -241,7 +241,6 @@ int main(int argc, char* argv[]) {
       "-- Starting main loop: inference is performed at each time step, but we "
       "plot trajectory every 10 steps\n");
   size_t j = 0;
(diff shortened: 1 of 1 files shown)


varunagrawal (approved):
LGTM. Thanks for the fix!

varunagrawal:
Looks like the PPA servers are down. I'll re-run the CI later today.

PR #1743, opened 2024-04-03
Some Updates

varunagrawal:
Made some updates and fixes while working on some Contact Estimation examples.

The big change is that we no longer check if a `GaussianFactor` inside a `GaussianMixtureFactor` has 0 keys, since the factor may be a non-null factor with 0 keys (aka a residual).

I also fixed a typo in the `slam/README.md` file.


dellaert (approved):
👍

varunagrawal:
As an FYI, I was going to merge this once we had approvals on the hybrid PRs so that they would be easier to review.

PR #1768, opened 2024-06-25
Update MSVC toolset version

varunagrawal:
Based on the [Windows image](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#microsoft-visual-c) from the Github Actions images repo, the minimum toolset version for MSVC should be 14.40.

This PR updates our CI to follow this.


varunagrawal:
I found the issue. It has to do with numpy 2.0.0. The tests pass perfectly with numpy 1.26.4 (the latest v1 release).
The best way to fix this would be to update pybind11 to 2.13.1. Pybind11 2.12 added support for numpy v2 so that should allow us to be forward compatible.

attn: @dellaert @ProfFan

gchenfc (approved):
This seems reasonable.  If it is relevant, it might be worth mentioning in the INSTALL.md:
https://github.com/borglab/gtsam/blob/a2a7a066464c4f147aad74bd98efccfe0f53baae/INSTALL.md?plain=1#L97
but based on what I understand, this is just a CI thing and not an actual install requirement version change so no change is necessary.

Also, re: pybind upgrade, this also seems reasonable if you want to go for it.  Will that be a separate PR or is it closely related enough to warrant combining in / replacing this PR?

ProfFan:
Pybind update should be done with the script

varunagrawal:
> Pybind update should be done with the script

Yes I'll use the script but it first needs to be updated in the wrap project. Can you please take a look at the PR I made there?

varunagrawal:
Made a temporary fix by restricting the numpy version to be below `v2.0.0`.

PR #1774, opened 2024-06-28
Update namespace for Vector and Matrix

varunagrawal:
As per the new wrapper design to enable better template handling, we now need to add the `gtsam::` namespace to all `Vector` and `Matrix` declarations.
This PR adds those namespaces.


ProfFan:
LGTM, Did you test MATLAB?

varunagrawal:
I did test on Matlab so we're good there.

PR #1780, opened 2024-07-09
modernize containers

### LINKED ISSUE #1779: Modernize GTSAM containers
## Feature

Need some GTSAM containers, noticed the containers we provide are pretty outdated ([latest push to Docker Hub 3 years ago](https://hub.docker.com/r/borglab/ubuntu-gtsam/tags) and based off bionic).

I also noticed some todo's to use the [Compose Spec](https://compose-spec.io/) and noticed that our list of images could be streamlined with the use of tags.

tldr: modernize containers with newer versions, better tooling (compose), use of tags for various configurations, slimmer containers with best practices of writing container files.

## Motivation

Need newer GTSAM images, smaller image sizes, and easier building/pushing to online registries

## Pitch

Replace build.sh scripts with compose.yaml

Use `.env` files to specify various gtsam configurations, such as base ubuntu image, tbb on/off, python on/off, gtsam git tag.
Have a single gtsam image repository (instead of multiple potentially unused intermediary images) with various tags corresponding to build configurations
Have another image for the gtsam-vnc image (of course tagged with a base gtsam image)

Create a simple script to push various tagged images to docker hub in one go.

Create smaller image sizes, by condensing `RUN` commands and using appropriate flags such as [`--no-install-recommends`](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends).

## Alternatives

N/A

## Additional context

For full transparency, I'm writing this issue after I have implemented a resolution for it. See: https://hub.docker.com/r/acxz/gtsam/tags

acxz:
use compose spec, use single image repo with tags for various configs, create a docker hub push script, clean up/slim down container files

resolves #1779

Running this PR on my personal docker hub account: https://hub.docker.com/r/acxz/gtsam/tags


ProfFan (approved):
I 2nd this, thanks Akash!

containers/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+- container files to build images
+- script to push images to a registry
+- instructions to pull images and run containers
+
+## Dependencies
+
+- a container engine
+
+Below we use [podman](https://podman.io), but you should be able to replace any `podman` calls to a Docker-compatible CLI such as [`docker`](https://www.docker.com/) or [`nerdctl`](https://github.com/containerd/nerdctl).
```
dellaert: Could we just use docker to make it easy for folks?

dellaert (approved):
Awesome! But, I have never heard of podman - I'd prefer if the docs were written with docker (and tested to make sure the podman flags were not podman specific).

containers/README.md
acxz: sure, I'll switch `podman` calls out to `docker`.

acxz: Done, removed mentions to alternative container tools.

I figured that if someone cares enough about using a different tool than Docker, they'll put in the effort to install such a tool, figure out the correct commands to build/pull/push images, and the commands to run containers. (Which itself shouldn't be a big deal since tools like podman and nerdctl have a Docker-compatible CLI.)

acxz:
Glad you all like the PR!

Docker does have a huge mindshare as they started the whole container revolution, but I have personally found podman easier to install and more applicable due to rootless daemonless capability (although I imagine Docker may have fixed that security vulnerability by now).

Nonetheless I agree with you and have removed mentions of alternative container tools, but kept general phrasing such as container engine and compose spec implementation for dependencies.

acxz:
I imagine getting the containers on Docker Hub will need to be done by someone with access to the [BORG Lab organization](https://hub.docker.com/u/borglab). The image repositories this PR replaces can/should be removed.

Some repositories might need to change their base images, such as [gtdynamics](https://hub.docker.com/r/borglab/ubuntu-gtdynamics) ([github link](https://github.com/borglab/GTDynamics/blob/master/docker/Dockerfile)). There is also the [manylinux](https://hub.docker.com/r/borglab/gtsam-manylinux) image, that may need to be changed.

I also just noticed a copy of the container related stuff in our [docker-images](https://github.com/borglab/docker-images) repo. Does that repo need to be cleaned up with the submission of this PR? (at the very least remove the duplicate images [`gtsam`](https://github.com/borglab/docker-images/tree/main/gtsam) and [`ubuntu-boost-tbb`](https://github.com/borglab/docker-images/tree/main/ubuntu-boost-tbb) and update corresponding base images.)

acxz:
One more thing regarding future work: I can imagine setting up a github action or other github-fu to build and push a container image to docker hub whenever a new gtsam git release is made. Outside the scope of this PR, but just a thought to prevent stale images. Github might have something that makes this dead simple.

containers/Containerfile
```diff
@@ -0,0 +1,45 @@
...
+# build dependencies
+    build-essential \
+    cmake \
+# download dependencies
+    git \
+    ca-certificates && \
+    rm -rf /var/lib/apt/lists/*
+
+# build flags
+ARG GTSAM_GIT_TAG=4.2.0
```
varunagrawal, line 23: Could we replace `GTSAM_GIT_TAG` with `GTSAM_VERSION_TAG`? Git tag would refer to the commit hash in a general sense.

containers/README.md
```diff
@@ -0,0 +1,127 @@
...
+- a container engine such as [`Docker Engine`](https://docs.docker.com/engine/install/)
+
+## Pull from Docker Hub
+
+Various GTSAM image configurations are available at [`docker.io/borglab/gtsam`](https://hub.docker.com/r/borglab/gtsam). Determine which [tag](https://hub.docker.com/r/borglab/gtsam/tags) you want and pull the image.
+
+Example for pulling an image with GTSAM compiled with TBB and Python support on top of a base Ubuntu 22.04 image.
+
+```bash
+docker pull docker.io/borglab/gtsam:4.2.0-tbb-ON-python-ON_22.04
```
varunagrawal, line 18: Would it be possible to make the naming scheme: `gtsam:4.2.0-tbb-python_22.04`, `gtsam:4.2.0_22.04` etc?

varunagrawal (approved):
Thanks! This looks pretty great. I can update the docker images on Dockerhub once this is approved.

General comments: I am of the opinion that container files should be in their own repository. That way we decouple container update and support from the actual code repository. I do this with much success in https://github.com/varunagrawal/docker-images where I have containers beyond just GTSAM.

varunagrawal:
> I imagine getting the containers on Docker Hub will need to be done by someone with access to the [BORG Lab organization](https://hub.docker.com/u/borglab). The image repositories this PR replaces can/should be removed.

I'll do that after we merge this.

> Some repositories might need to change their base images, such as [gtdynamics](https://hub.docker.com/r/borglab/ubuntu-gtdynamics) ([github link](https://github.com/borglab/GTDynamics/blob/master/docker/Dockerfile?rgh-link-date=2024-07-10T01%3A09%3A49Z)). There is also the [manylinux](https://hub.docker.com/r/borglab/gtsam-manylinux) image, that may need to be changed.

Yeah this is going to be a headache. :)
 
> I also just noticed a copy of the container related stuff in our [docker-images](https://github.com/borglab/docker-images) repo. Does that repo need to be cleaned up with the submission of this PR? (at the very least remove the duplicate images [`gtsam`](https://github.com/borglab/docker-images/tree/main/gtsam?rgh-link-date=2024-07-10T01%3A09%3A49Z) and [`ubuntu-boost-tbb`](https://github.com/borglab/docker-images/tree/main/ubuntu-boost-tbb?rgh-link-date=2024-07-10T01%3A09%3A49Z) and update corresponding base images.)

Should have made my comment here. I strongly believe that the `docker-images` repo should be the container repo. That way we can add Github Actions for easy deployment and manage the connection between that repo and Dockerhub in a clean way, without adding more complexity to the GTSAM repo. @acxz if you have the bandwidth, I'd love to do this so we can update **all** the images we support.

Building containers takes a LOOONG time and that just makes CI on this repo take more time which makes rapid iteration difficult, especially since we don't have a commercial Github license here.
attn: @dellaert @ProfFan

containers/Containerfile
acxz: A git `tag` refers to the output of [`git tag`](https://git-scm.com/book/en/v2/Git-Basics-Tagging), not the commit hash.

Running it on the gtsam repo gives us the following:
```
~\Desktop\acxz\gtsam> git tag                                                   07/11/2024 07:37:32 AM
2.1.0
2.2.0
2.3.0
2.3.1
2.4.0
3.0.0
3.1.0
3.2.0
3.2.1
3.2.2
3.2.3
4.0.0
4.0.0-alpha1
4.0.0-alpha2
4.0.1
4.0.2
4.0.3
4.1.0
4.1.0-1
4.1.1
4.1rc
4.2
4.2.0
4.2a0
4.2a1
4.2a2
4.2a3
4.2a4
4.2a5
4.2a6
4.2a7
4.2a8
4.2a9
wrap-export
```

Specifying the variable as `GTSAM_GIT_TAG` makes it clear what possible values can be picked. If you do want me to change it to `GTSAM_VERSION_TAG`, I'd recommend I change it to `GTSAM_VERSION`, since imo version and tag are redundant descriptors.

containers/README.md
acxz: That would be pretty neat. I could change the name in the `hub_push.sh` scripts when it pushes to the image registry, however, I also specify the image:tag in the `compose.yaml` file. Here, we can do variable substitution, but I'm pretty sure we cannot do additional logic to have empty string instead of "tbb" based on the env variables.

I would like to keep both image names the same between the locally built image `compose.yaml` and the one that we push to docker hub `hub_push.sh`. And with that constraint I do not believe it is possible.

Another solution is to have a separate `compose.yaml` file for when TBB is ON and when python is ON, but then we could have a proliferation of `compose.yaml` files for each build configuration, and we wouldn't have a clean `hub_push.sh` script; it would be more complicated and have various nested if statements. I'd recommend against adding such complexity (even though it is 3 more compose.yaml files and a slightly more complicated bash script).

containers/Containerfile
varunagrawal: I see what you mean. That's definitely confusing because it could mean we're tagging the image version with that, so its a reference.
Plus, even docker images have tags so it's easy to confuse it with that as well (`docker image tag`).

I believe every release tag is associated with a version release so the two should be interchangeable.

acxz:
> Should have made my comment here. I strongly believe that the docker-images repo should be the container repo. That way we can add Github Actions for easy deployment and manage the connection between that repo and Dockerhub in a clean way, without adding more complexity to the GTSAM repo. @acxz if you have the bandwidth, I'd love to do this so we can update all the images we support.

Having the containers as part of a separate repo from the main GTSAM library makes sense to me as well. I'd just suggest a name change to the repo to be `container-images` or `borglab-containers` instead of `docker-images.

I do believe that the action you are proposing be held off until this PR is merged. In my opinion, this PR should be merged first and the newer images be pushed to Docker Hub. After that, if the core GTSAM team decides so, the `docker/`/`containers/` folder here should be created as a pull request to the current docker-images folder. Once that is merged in, the existing `docker`/`containers` folder would be deleted from here and all further borglab container development could happen in that repo.

I don't necessarily want to hold off on getting this work published to Docker Hub because of organizational movement (even though I agree with such movement), especially considering the fact that I may not have the bandwidth to create a proper PR to the `docker-images` repo as there would be the `gtdynamics` and the `manylinux` configurations I would have to worry about, which I am not working on directly.

varunagrawal:
@acxz I like that plan and I agree with it. I think I'm good to merge this PR.

dellaert:
I'm looking at this again now, and I'm now question the wisdom of not having a standalone image like `ubuntu-boost-tbb` was, and which is *still* mentioned on docker hub, despite someone having *removed* the actual image? It would be nice to be able to create a VS code devcontainer on an image that you know has all the pre-requisites to locally build develop. So:
- can we re-instate ubuntu-boost-tbb ?
- how can we split the new Dockerfile to reflect the base image and the"fully-built" images (cake and eat it too)

cntaylor:
Is this helpful?

In [this repo](https://github.com/cntaylor/gtsam-python-docker-vscode/tree/main) I have a multi-stage Docker file that first creates a stage with all of the libraries installed to make GTSAM.  From there it branches to a "doc", "debug", and "release" Docker container, allowing you to locally serve up docs, or run GTSAM in debug or release.  

I think this answers your second question about building a Dockerfile that both builds the base image, but allows you to take advantage of it if it is already there.  Feel free to use it if helpful.

varunagrawal:
The docker-images repo still has the Ubuntu base image. I didn't realize this PR removed that image because I wanted to use it in the CI to cut down on build time.

The good news is that docker-images not only has the images but is connected to dockerhub so the images are automatically built and ready to download. I configured our OSS support from Docker.

varunagrawal:
@cntaylor that's really helpful! Thank you.

PR #1781, opened 2024-07-11
`DiscreteConditional` and `DiscreteBayesNet` improvements

varunagrawal:
1. New constructor for `DiscreteConditional`.
2. `DiscreteConditional::argmax` for arbitrary conditionals, not just conditionals with no parents.
3. Merge `DiscreteLookupTable` and `DiscreteConditional`.
4. Unit tests.

### DIFF
--- a/gtsam/discrete/tests/testDiscreteConditional.cpp
+++ b/gtsam/discrete/tests/testDiscreteConditional.cpp
@@ -293,22 +293,22 @@ TEST(DiscreteConditional, choose) {
 // Check argmax on P(C|D) and P(D)
 TEST(DiscreteConditional, Argmax) {
   DiscreteKey C(2, 2), D(4, 2);
-  DiscreteConditional D_cond(D, "1/3");
-  DiscreteConditional C_given_DE((C | D) = "1/4 1/1");
+  DiscreteConditional D_prior(D, "1/3");
+  DiscreteConditional C_given_D((C | D) = "1/4 1/1");
 
   // Case 1: No parents
-  size_t actual1 = D_cond.argmax();
+  size_t actual1 = D_prior.argmax();
   EXPECT_LONGS_EQUAL(1, actual1);
 
   // Case 2: Given parent values
   DiscreteValues given;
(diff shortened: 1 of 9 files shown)


gtsam/discrete/tests/testDiscreteConditional.cpp
```diff
@@ -289,6 +289,29 @@ TEST(DiscreteConditional, choose) {
   EXPECT(assert_equal(expected3, *actual3, 1e-9));
 }
 
+/* ************************************************************************* */
+// Check argmax on P(C|D) and P(D)
+TEST(DiscreteConditional, Argmax) {
+  DiscreteKey C(2, 2), D(4, 2);
+  DiscreteConditional D_cond(D, "1/3");
```
dellaert: D_prior

gtsam/discrete/tests/testDiscreteConditional.cpp
```diff
@@ -289,6 +289,29 @@ TEST(DiscreteConditional, choose) {
...
+  DiscreteConditional C_given_DE((C | D) = "1/4 1/1");
```
dellaert: C_given_D

gtsam/discrete/tests/testDiscreteConditional.cpp
```diff
@@ -289,6 +289,29 @@ TEST(DiscreteConditional, choose) {
...
+
+  // Case 1: No parents
+  size_t actual1 = D_cond.argmax();
+  EXPECT_LONGS_EQUAL(1, actual1);
```
dellaert: Why is it 1? Document how tie-breaking will be handled in argmax

dellaert (changes requested):
I think the mode function is not correct. The correct way to compute the mode is converting to a factor graph and calling max-product. There will be cases where they give the same result, but not in general. I think I even have a counter-example in the tests somewhere.

dellaert (approved):
Cool, thanks for figuring it out with me.

PR #1782, opened 2024-07-11
Remove warnings in Windows CI

varunagrawal:
For CMake<3.15, when building with MSVC, the compiler warning flags (/W3, /W4) are added by default.

This PR removes the warnings by performing the necessary replacements, which helps with faster compilation on Windows.


ProfFan (approved):

varunagrawal:
This PR isn't working as expected, so closing it.

PR #1789, opened 2024-08-06
add definition of constraints

yetongumich:
First PR in the 4 PRs to implement nonlinear constrained optimization in GTSAM.

1. Introducing factors representing equality and inequality constraints; refactor existing constraint factors to inherit from it.
2. Implementation of existing constrained optimization methods that are compatible with the current gtsam framework, including penalty methods, augmented Lagrangian methods, SQP methods (potentially also interior point methods), with a document explaining the math of the implementations.
3. Constraint manifold optimization that solves problems with equality constraints.
4. Constraint manifold with corners optimization that solves problems with both equality and inequality constraints.
5. (potentially 1 additional PR) That solves an equality-constrained factor graph using nonlinear variable elimination (resulting in an behavior identical to iLQR or DDP).

This PR

- Introduce `NonlinearEqualityConstraint` and `NonlinearInequalityConstraint` inheriting from NoiseModelFactor to represent constraints.
- Refactor the existing classes`NonlinearEquality` to inherit from `NonlinearEqualityConstraint`, and `BoundingConstraint` to inherit from `NonlinearInequalityConstraint`.


dellaert:
CI fails. Talking about CI, try to save on CO2 by not pushing every commit, but only after you think CI will pass.

yetongumich:
> CI fails. Talking about CI, try to save on CO2 by not pushing every commit, but only after you think CI will pass.

It seems pretty weird that only the Windows CI fails. Unfortunately I do not have a Windows environment to test this on. I was trying to tweak with different changes to see which causes the Windows CI to fail.

gtsam/constraint/NonlinearInequalityConstraint.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * @file    NonlinearInequalityConstraint.cpp
+ * @brief   Nonlinear inequality constraints in constrained optimization.
+ * @author  Yetong Zhang, Frank Dellaert
+ * @date    Aug 3, 2024
+ */
+
+#include <gtsam/constraint/NonlinearInequalityConstraint.h>
+
```
ProfFan: Include "
#include <gtsam/inference/FactorGraph-inst.h>"

yetongumich: Added this line, CI still reports the same error in Windows environments.

dellaert:
@yetongatech error is due to a argument type not being recognized:
```
D:\a\gtsam\gtsam\gtsam/inference/FactorGraph-inst.h(69): error C2664: 'double gtsam::NoiseModelFactor::error(const gtsam::Values &) const': cannot convert argument 1 from 'const gtsam::HybridValues' to 'const gtsam::Values &'
```
Try figuring out what type you want and somehow being explicit? I just looked at the error above but maybe you can dig deeper.

gtsam/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ project(gtsam LANGUAGES CXX)
 set (gtsam_subdirs
     base
     basis
+    constraint
```
dellaert: If this directory only contains constraint classes, this is ok. Otherwise (if optimization code will land here) could we rename this to `constrained` ?

gtsam/constraint/InequalityPenaltyFunction.h
```diff
@@ -0,0 +1,140 @@
...
+ public:
+  typedef InequalityPenaltyFunction Base;
+  typedef SoftPlusFunction This;
+  typedef std::shared_ptr<This> shared_ptr;
+
+ protected:
+  double k_;
+
+ public:
+  SoftPlusFunction(const double k = 1) : Base(), k_(k) {}
```
dellaert: doxygen block describing argument

gtsam/constraint/InequalityPenaltyFunction.h
```diff
@@ -0,0 +1,140 @@
...
+  typedef SmoothRampPoly3 This;
+  typedef std::shared_ptr<This> shared_ptr;
+
+ protected:
+  double epsilon_;
+  double a_;
+  double b_;
+
+ public:
+  SmoothRampPoly3(const double epsilon = 1)
```
dellaert: doxygen block describing argument

gtsam/constraint/InequalityPenaltyFunction.h
```diff
@@ -0,0 +1,140 @@
...
+  typedef InequalityPenaltyFunction Base;
+  typedef SmoothRampPoly2 This;
+  typedef std::shared_ptr<This> shared_ptr;
+
+ protected:
+  double epsilon_;
+  double a_;
+
+ public:
+  SmoothRampPoly2(const double epsilon = 1)
```
dellaert: doxygen block describing argument

gtsam/constraint/NonlinearEqualityConstraint.h
```diff
@@ -0,0 +1,180 @@
...
+
+ protected:
+  NoiseModelFactor::shared_ptr factor_;
+
+ public:
+  /**
+   * @brief Constructor.
+   *
+   * @param factor  NoiseModel factor.
+   * @param tolerance   vector representing tolerance in each dimension.
```
dellaert: stale?

gtsam/constraint/NonlinearEqualityConstraint.h
```diff
@@ -0,0 +1,180 @@
...
+    ar& boost::serialization::make_nvp("ExpressionEqualityConstraint",
+                                       boost::serialization::base_object<Base>(*this));
+    ar& BOOST_SERIALIZATION_NVP(expression_);
+    ar& BOOST_SERIALIZATION_NVP(rhs_);
+    ar& BOOST_SERIALIZATION_NVP(dims_);
+  }
+#endif
+};
+
+/** Equality constraint that enforce the cost factor with zero error. */
```
dellaert: Add more guidance, example?

gtsam/constraint/tests/constrainedExample.h
```diff
@@ -0,0 +1,181 @@
...
+inline Double_ operator-(const Double_& x) { return Double_(0.0) - x; }
+
+/// Keys for creating expressions.
+Symbol x1_key('x', 1);
+Symbol x2_key('x', 2);
+Double_ x1(x1_key), x2(x2_key);
+
+}  // namespace constrained_example
+
+/* ************************************************************************* */
```
dellaert: Why are below commented out? If for next PR include in next PR :-)

gtsam/nonlinear/NonlinearEquality.h
```diff
@@ -158,11 +160,20 @@ class NonlinearEquality: public NoiseModelFactorN<VALUE> {
     }
   }
 
+  Vector unwhitenedError(const Values& x, OptionalMatrixVecType H = nullptr) const override {
+    VALUE x1 = x.at<VALUE>(key());
+    if (H) {
+      return evaluateError(x1, &(H->front()));
```
dellaert, line 166: does the base class not have unwhitenedError? Why not? Would that not be easier?

gtsam/slam/BoundingConstraint.h
```diff
@@ -60,14 +59,23 @@ struct BoundingConstraint1: public NoiseModelFactorN<VALUE> {
   virtual double value(const X& x, OptionalMatrixType H =
       OptionalNone) const = 0;
 
-  /** active when constraint *NOT* met */
-  bool active(const Values& c) const override {
```
dellaert, line 64: is active not used? Or is it in base class? Backwards compat...

gtsam/slam/BoundingConstraint.h
```diff
@@ -60,14 +59,23 @@ struct BoundingConstraint1: public NoiseModelFactorN<VALUE> {
...
-    // note: still active at equality to avoid zigzagging
-    double x = value(c.at<X>(this->key()));
-    return (isGreaterThan_) ? x <= threshold_ : x >= threshold_;
+  Vector unwhitenedExpr(const Values& x, OptionalMatrixVecType H = nullptr) const override {
```
dellaert: same, base class unwhitenedError?

gtsam/slam/BoundingConstraint.h
```diff
@@ -133,15 +138,27 @@ struct BoundingConstraint2: public NoiseModelFactorN<VALUE1, VALUE2> {
       OptionalMatrixType H1 = OptionalNone,
       OptionalMatrixType H2 = OptionalNone) const = 0;
 
-  /** active when constraint *NOT* met */
-  bool active(const Values& c) const override {
-    // note: still active at equality to avoid zigzagging
-    double x = value(c.at<X1>(this->key1()), c.at<X2>(this->key2()));
-    return (isGreaterThan_) ? x <= threshold_ : x >= threshold_;
+  Vector unwhitenedExpr(const Values& x, OptionalMatrixVecType H = nullptr) const override {
```
dellaert: same question about active
what is `unwhitenedExpr`? Not documented

gtsam/slam/BoundingConstraint.h
```diff
@@ -133,15 +138,27 @@ struct BoundingConstraint2: public NoiseModelFactorN<VALUE1, VALUE2> {
...
+    } else {
+      double d = value(x1, x2);
+      return Vector1((isGreaterThan_) ? threshold_ - d : d - threshold_);
+    }
   }
 
+  /// TODO: This should be deprecated.
   Vector evaluateError(const X1& x1, const X2& x2,
-      OptionalMatrixType H1, OptionalMatrixType H2) const override {
+      OptionalMatrixType H1 = nullptr, OptionalMatrixType H2 = nullptr) const {
```
dellaert: You used {} elsewhere

dellaert (commented):
Cool ! Some comments/questions.
Still need to pass CI

ProfFan:
I think if you rebase on current `develop` it should fix stuff

yetongumich:
> I think if you rebase on current `develop` it should fix stuff

Seems still not working

dellaert:
@yetongatech you might have to investigate a bit to get past windows issue. None of us have the answers exactly :-)

dellaert:
@yetongumich , checks passed, will you address comments (and resolve them when addressed) ?

gtsam/slam/BoundingConstraint.h
yetongumich: Done

gtsam/CMakeLists.txt
yetongumich: Done

gtsam/constraint/NonlinearEqualityConstraint.h
yetongumich: Done

gtsam/constraint/tests/constrainedExample.h
yetongumich: Removed. Will come in next PR. Done.

gtsam/slam/BoundingConstraint.h
yetongumich: Yes, the previous BoundingConstraint class has this method. it's for backward compatibility.

gtsam/nonlinear/NonlinearEquality.h
yetongumich: The parent classes have unwhitenedError as pure virtual function, because different types of constraint representation have different ways of computing the unwhitenedError. e.g., ExpressionEqualityConstraint computes unwhitenedError by evaluating an expression; ZeroCostConstraint computes unwhitenedError by evaluating the error vector of the underlying factor. NonlinearEquality (refactored from previous code) represents a specific type of constraint with its specific evaluateError function which should be called to compute unwhitenedError.

gtsam/slam/BoundingConstraint.h
yetongumich: For an inequality constraint g(x)<=0, the unwhitenedExpr represents g(x), while unwhitenedError represents  H(g(x)), where H() is the unit ramp function. Therefore, the unwhitenedError can be computed given unwhitenedExpr, which is implemented in the base class, while unwhitenedExpr is pure virtual.

gtsam/slam/BoundingConstraint.h
yetongumich: see above. `unwhitenedExpr` is documented in the base class.

gtsam/constraint/InequalityPenaltyFunction.h
yetongumich: Done

gtsam/constraint/InequalityPenaltyFunction.h
yetongumich: Done

gtsam/constraint/InequalityPenaltyFunction.h
yetongumich: Done

gtsam/constraint/NonlinearEqualityConstraint.h
yetongumich: Done

dellaert (approved):
Wow ! In a a PR, I think you'll add a doc on how to use all this, with examples, right?

dellaert:
> Wow ! In a a PR, I think you'll add a doc on how to use all this, with examples, right?

@yetongatech ping. If so, please feel free to merge.

yetongumich:
> > Wow ! In a a PR, I think you'll add a doc on how to use all this, with examples, right?
> 
> @yetongatech ping. If so, please feel free to merge.

The examples come in next PR that includes constrained optimizers. Merging first.

Issue #1596, opened 2023-07-28
BoundingConstraint1 with double parameter


shounak873:
In my factor graph I have  scalar parameter that is being estimated. This scalar has a lower bound. I wanted to implement this constraint using BoundingConstraint1 struct. I replaced the existing template version `BoundingConstraint.h` with this version below. But I am getting segmentation fault when running with this factor. Is this implementation correct ? Thank you.

```
#pragma once

#include <gtsam/base/Lie.h>
#include <gtsam/nonlinear/NonlinearFactor.h>

namespace gtsam {

/**
 * Unary inequality constraint forcing a scalar to be
 * greater/less than a fixed threshold.  The function
 * will need to have its value function implemented to return
 * a scalar for comparison.
 * @addtogroup SLAM
 */
// template<class VALUE>
struct BoundingConstraint1: public NoiseModelFactor1<double> {
  // typedef VALUE X;
  typedef NoiseModelFactor1<double> Base;
  typedef boost::shared_ptr<BoundingConstraint1> shared_ptr;

  double threshold_;
  bool isGreaterThan_; /// flag for greater/less than

  BoundingConstraint1(Key key, double threshold,
      bool isGreaterThan, double mu = 1000.0) :
        Base(noiseModel::Constrained::All(1, mu), key),
        threshold_(threshold), isGreaterThan_(isGreaterThan) {
  }

  virtual ~BoundingConstraint1() {}

  inline double threshold() const { return threshold_; }
  inline bool isGreaterThan() const { return isGreaterThan_; }

  /**
   * function producing a scalar value to compare to the threshold
   * Must have optional argument for derivative with 1xN matrix, where
   * N = X::dim()
   */
  // virtual double value(const X& x, boost::optional<Matrix&> H =
  //     boost::none) const = 0;

  /** active when constraint *NOT* met */
  bool active(const Values& c) const {
    // note: still active at equality to avoid zigzagging
    double x = c.at<double>(this->key());
    return (isGreaterThan_) ? x <= threshold_ : x >= threshold_;
  }

  Vector evaluateError(const double& x, boost::optional<Matrix&> H =
      boost::none) const {
    // Matrix D;
    double error = x - threshold_;
    Matrix D = Matrix::Zero(1,1);
    D(0,0) = 1.0;
    if (H) {
      if (isGreaterThan_) *H = D;
      else *H = -1.0 * D;
    }

    if (isGreaterThan_)
      return (Vector(1) << error).finished();
    else
[...]

PR #1790, opened 2024-08-07
Adding test and fix for issue #1596

garygra:
Addressing Issue #1596: When a derived Nonlinear Factor is not active, the linearization returns a nullptr (NonlinearFactor.cpp:156), which causes an error when `EliminateSymbolic` is called (SymbolicFactor-inst.h:45) because it does not check if the factor is nullptr.

### DIFF
--- a/tests/testGaussianISAM2.cpp
+++ b/tests/testGaussianISAM2.cpp
@@ -994,6 +994,56 @@ TEST(ISAM2, calculate_nnz)
   EXPECT_LONGS_EQUAL(expected, actual);
 }
 
+class FixActiveFactor : public NoiseModelFactorN<Vector2> {
+  using Base = NoiseModelFactorN<Vector2>;
+  bool is_active_;
+
+public:
+  FixActiveFactor(const gtsam::Key& key, const bool active)
+      : Base(nullptr, key), is_active_(active) {}
+
+  virtual bool active(const gtsam::Values &values) const override {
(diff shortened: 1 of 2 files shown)


ProfFan (approved):
I think this is right, but I'll tag @dellaert for a 2nd check as `symbolic` is key part to all components of GTSAM.

tests/testGaussianISAM2.cpp
```diff
@@ -994,6 +994,56 @@ TEST(ISAM2, calculate_nnz)
...
+  evaluateError(const Vector2& x,
+                Base::OptionalMatrixTypeT<Vector2> H = nullptr) const override {
+    if (H) {
+      *H = Vector2::Identity();
+    }
+    return Vector2::Zero();
+  }
+};
+
+TEST(ActiveFactorTesting, Issue1596) {
```
varunagrawal, line 1019: Just a very small nitpick: for consistency, can you please change this to `TEST(ISAM2, ActiveFactorTesting)`. The commit history will let us know this test was in service for issue 1596. :)

dellaert (approved):

tests/testGaussianISAM2.cpp
dellaert: Let’s not incur another CI for this - maybe Varun you can change it in another PR. I’ll merge.

PR #1795, opened 2024-08-16
test cases for IncrementalFixedLagSmoother

truher:
two cases here:

* SEGV while marginalizing a variable with only a prior.
* IndexError during marginalization that seems distantly caused by the behavior of update()?

### DIFF
--- a/python/gtsam_unstable/tests/test_missing_variable.py
+++ b/python/gtsam_unstable/tests/test_missing_variable.py
@@ -1,0 +1,90 @@
+# pylint: disable=invalid-name,no-name-in-module,no-member,missing-class-docstring,missing-function-docstring
+
+"""
+Demonstrates incorrect pruning of old variables.
+
+It seems like if you call isam.update() with two separate updates for the same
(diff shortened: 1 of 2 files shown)


ProfFan:
Thanks! The `IncrementalFixedLagSmoother` has been known to have problems. There are several workarounds in the issues section, but none of us have the bandwidth for the bugfix right now.

truher:
no worries, i appreciate your quick response.  i can avoid these particular issues, though i'm concerned that there may be more issues lurking in the future.  is there a better choice than `IncrementalFixedLagSmoother` ?  The "batch" version doesn't seem to have implemented the `marginalCovariance` method, which I kinda want.

dellaert:
I appreciate the tests and might take a look. I have been seeing some other segfaults with 3.12.
What version of python are you using?

truher:
thank you!  i know you're very busy.

re: python version, i'm using 3.10.12, on ubuntu 22.04.

truher:
closed so i can have the "develop" branch back.

see https://github.com/borglab/gtsam/pull/1809

PR #1796, opened 2024-08-16
Cherry pick ISAM2 fix to 4.2

ProfFan:


ProfFan:
We may want to tag a 4.2.1 bug fix release after this is merged, as many are still using 4.2

@dellaert

PR #1797, opened 2024-08-16
Fix and add tests for AHRS

andre-michelin:
Fix and add tests to testPreintegratedRotation and testRot3


dellaert (approved):
Awesome, thanks for the extra tests and the Jacobian simplification !!!

ProfFan:
May need `#ifdef`s for the Rot3 = Quaternion case

dellaert:
Interesting- seems like maybe ExpmapDerivative is more approximate in quaternion case…

PR #1801, opened 2024-08-20
Improvements to `GaussianConditional` and `GaussianBayesNet`

varunagrawal:
1. `GaussianBayesNet::logNormalizationConstant` method.
2. Equality operators for both `GaussianBayesNet` and `GaussianConditional`.
3. Print the `logNormalizationConstant` when printing the `GaussianConditional`.
4. Wrap more `GaussianConditional` constructors.

### DIFF
--- a/gtsam/linear/GaussianBayesNet.cpp
+++ b/gtsam/linear/GaussianBayesNet.cpp
@@ -243,5 +243,25 @@ namespace gtsam {
   }
 
   /* ************************************************************************* */
+  double GaussianBayesNet::logNormalizationConstant() const {
+    /*
+    normalization constant = 1.0 / sqrt((2*pi)^n*det(Sigma))
+    logConstant = -0.5 * n*log(2*pi) - 0.5 * log det(Sigma)
+    
+    log det(Sigma)) = -2.0 * logDeterminant()
+    thus, logConstant = -0.5*n*log(2*pi) + logDeterminant()
+
+    BayesNet logConstant = sum(-0.5*n_i*log(2*pi) + logDeterminant_i())
+    = sum(-0.5*n_i*log(2*pi)) + sum(logDeterminant_i())
+    = sum(-0.5*n_i*log(2*pi)) + bn->logDeterminant()
+    */
+    double logNormConst = 0.0;
+    for (const sharedConditional& cg : *this) {
+      logNormConst += cg->logNormalizationConstant();
+    }
+    return logNormConst;
+  }
+
+  /* ************************************************************************* */
 
 } // namespace gtsam
--- a/gtsam/linear/GaussianBayesNet.h
+++ b/gtsam/linear/GaussianBayesNet.h
@@ -82,6 +82,12 @@ namespace gtsam {
     /** Check equality */
     bool equals(const This& bn, double tol = 1e-9) const;
 
+    /// Check exact equality.
+    friend bool operator==(const GaussianBayesNet& lhs,
+                           const GaussianBayesNet& rhs) {
+      return lhs.isEqual(rhs);
+    }
+
     /// print graph
     void print(
         const std::string& s = "",
@@ -227,6 +233,14 @@ namespace gtsam {
      * @param bayesNet The input GaussianBayesNet
      * @return The determinant */
     double logDeterminant() const;
+
+    /**
+     * @brief Get the log of the normalization constant corresponding to the
+     * joint Gaussian density represented by this Bayes net.
+     *
+     * @return double
+     */
+    double logNormalizationConstant() const;
 
     /**
      * Backsubstitute with a different RHS vector than the one stored in this BayesNet.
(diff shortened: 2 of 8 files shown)


gtsam/linear/GaussianBayesNet.cpp
```diff
@@ -243,5 +243,25 @@ namespace gtsam {
   }
 
   /* ************************************************************************* */
+  double GaussianBayesNet::logNormalizationConstant() const {
+    /*
```
dellaert, line 247: Is this copy/paste from gc ?

gtsam/linear/GaussianBayesNet.h
```diff
@@ -82,6 +82,12 @@ namespace gtsam {
     /** Check equality */
     bool equals(const This& bn, double tol = 1e-9) const;
 
+    /// Check exact equality.
+    friend bool operator==(const GaussianBayesNet& lhs,
+                           const GaussianBayesNet& rhs) {
+      return lhs.isEqual(rhs);
```
dellaert, line 88: But this is not exact equality

gtsam/linear/VectorValues.h
```diff
@@ -263,6 +263,11 @@ namespace gtsam {
     /** equals required by Testable for unit testing */
     bool equals(const VectorValues& x, double tol = 1e-9) const;
 
+    /// Check exact equality.
+    friend bool operator==(const VectorValues& lhs, const VectorValues& rhs) {
+      return lhs.equals(rhs);
```
dellaert, line 268: But this is not exact….

gtsam/linear/GaussianBayesNet.h
```diff
@@ -228,6 +234,14 @@ namespace gtsam {
      * @return The determinant */
     double logDeterminant() const;
 
+    /**
+     * @brief Get the log of the normalization constant corresponding to the
+     * joint Gaussian density represented by this Bayes net.
+     *
+     * @return double
+     */
+    double logNormalizationConstant() const;
```
dellaert, line 243: No unit test

dellaert (commented):
Several comments. Also puzzled why not targeting develop. I want to limit the amount of change in the other branch to ease review, and this seems completely orthogonal.

gtsam/linear/GaussianBayesNet.cpp
varunagrawal: Not quite. It collects all the `logNormalizationConstants` from the GCs.

gtsam/linear/GaussianBayesNet.h
varunagrawal: The docstring for the `isEqual` method in `FactorGraph.h` says
```
/// Check exact equality of the factor pointers. Useful for derived ==.
```
You wrote that docstring 3 years ago.

dellaert: Ah, isEqual might be. Please check it. `equals`is not.

varunagrawal:
> Also puzzled why not targeting develop. I want to limit the amount of change in the other branch to ease review, and this seems completely orthogonal.

This was all part of 1 PR initially and I broke it down. Chaining the PRs this way allows me to create a PR on top of both PRs rather than just using the changes in either one.

I don't not target to `develop` on purpose.

gtsam/linear/GaussianBayesNet.h
varunagrawal: Good catch. I'm updating a unit test in `testGaussianBayesNet.cpp` that will test this.

gtsam/linear/GaussianBayesNet.h
varunagrawal: Yup, I edited the other one which uses `equals`. 🙂

dellaert:
> > Also puzzled why not targeting develop. I want to limit the amount of change in the other branch to ease review, and this seems completely orthogonal.
> 
> This was all part of 1 PR initially and I broke it down. Chaining the PRs this way allows me to create a PR on top of both PRs rather than just using the changes in either one.
> 
> I don't not target to `develop` on purpose.

Ok, so the intent is that the base will change to develop?

varunagrawal:
Exactly. Once I merge the ADT PR and delete that branch, GitHub will rebase this PR to develop automatically. :)

gtsam/linear/VectorValues.h
dellaert: Still uses equals, no?

dellaert (approved):
Great. Still one issue - equals instead of isEqual in ==?

gtsam/linear/VectorValues.h
varunagrawal: I edited the docstring and it says `/// Check equality.` Is that not showing for you?

dellaert: operator== implies strict equality. If you want equals, then please rename to `equals` and provide a tolerance parameter.

varunagrawal:
> Great. Still one issue - equals instead of isEqual in ==?

I updated the docstring to say `Check equality` instead of `Check exact equality` immediately after your review comment. I am going to merge this, and if you have another comment, I can add it to the next PR in this series.

dellaert:
Yes, I think you missed my comment:
“operator== implies strict equality. If you want equals, then please rename to equals and provide a tolerance parameter.”

I feel strongly about being consistent across the code base here. If we find other instances of this we should also correct those.

PR #1802, opened 2024-08-20
Working Hybrid

varunagrawal:
1. Updated `GaussianMixtureFactor` to store the Gaussian normalizers as detailed in my thesis.
2. Added 2 unit tests `DifferingMeans` and `DifferingCovariances` which verify the mode selection.
3. Normalization of `logProbabilities` using a scheme detailed in my thesis.
4. Formatting and additional comments.


gtsam/hybrid/GaussianMixture.h
```diff
@@ -72,6 +72,12 @@ class GTSAM_EXPORT GaussianMixture
    */
   GaussianFactorGraphTree asGaussianFactorGraphTree() const;
 
+  /**
+   * @brief Convert a DecisionTree of conditionals into
```
dellaert: Convert a GaussianMixture? Also, maybe mention why this is useful.

BTW the GaussianMixture class-level comment should be updated given our most recent discussions.

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -92,6 +93,35 @@ GaussianFactorGraphTree GaussianMixture::asGaussianFactorGraphTree() const {
   return {conditionals_, wrap};
 }
 
+/*
+*******************************************************************************/
+GaussianBayesNetTree GaussianMixture::add(
```
dellaert: Expecting to see an example/test in testGaussianMixture.cpp

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -316,19 +348,48 @@ AlgebraicDecisionTree<Key> GaussianMixture::logProbability(
...
+  // present in the GaussianMixture
+  KeyVector dKeys = this->discreteKeys_.indices();
+  bool valid_assignment = false;
+  for (auto &&kv : values.discrete()) {
+    if (std::find(dKeys.begin(), dKeys.end(), kv.first) != dKeys.end()) {
+      valid_assignment = true;
+      break;
+    }
+  }
+
```
dellaert: Spelling, but also: what??? I’d expect to throw an Exception on invalid inputs

gtsam/hybrid/GaussianMixtureFactor.cpp
```diff
@@ -28,11 +28,86 @@
...
+ * with the normalizer values.
+ * This is done by storing the normalizer value in
+ * the `b` vector as an additional row.
+ *
+ * @param factors DecisionTree of GaussianFactor shared pointers.
+ * @param varyingNormalizers Flag indicating the normalizers are different for
+ * each component.
+ * @return GaussianMixtureFactor::Factors
+ */
+GaussianMixtureFactor::Factors correct(
```
dellaert: Let’s talk about this? I had not expected this functionality here…

gtsam/hybrid/GaussianMixture.h
varunagrawal: I was going to make a new PR to rename GaussianMixture to GaussianMixtureComponents and update the class docstring accordingly.

I'll update this method docstring though.

gtsam/hybrid/GaussianMixtureFactor.cpp
varunagrawal: This was a neat idea I had after working out the math. Now I can specify factors directly instead of calling `toFactorGraph` on a Bayes net.
Plus `toFactorGraph` is quite restrictive since it doesn't work for nonlinear functions (at least I don't see a clean way of defining nonlinear hybrid factors with how things currently are setup).

It'll come extremely handy when defining nonlinear MixtureFactors as then we can do relinearization during optimization, something that's been missing in the incumbent paradigm.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +196,240 @@ TEST(GaussianMixtureFactor, Error) {
...
+
+/* ************************************************************************* */
+/**
+ * @brief Test components with differing covariances.
+ * The factor graph is
+ *     *-X1-*-X2
+ *          |
+ *          M1
+ */
+TEST(GaussianMixtureFactor, DifferentCovariances) {
```
varunagrawal: @dellaert see this test as an example vs DifferentCovariances2. Specifying the FG is as easy as in continuous case.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +196,240 @@ TEST(GaussianMixtureFactor, Error) {
...
+  auto model1 = noiseModel::Isotropic::Sigma(1, 1e-2);
+  auto prior_noise = noiseModel::Isotropic::Sigma(1, 1e-3);
+
+  auto f0 =
+      std::make_shared<BetweenFactor<double>>(X(1), X(2), between, model0);
+  auto f1 =
+      std::make_shared<BetweenFactor<double>>(X(1), X(2), between, model1);
+  std::vector<NonlinearFactor::shared_ptr> factors{f0, f1};
+
+  // Create via toFactorGraph
```
varunagrawal: This comment is a leftover artifact. Please ignore.

gtsam/hybrid/GaussianMixtureFactor.cpp
dellaert: It does not make sense to me yet. It feels it throws away some of the clarity we achieved in the document. I didn’t even understand how you get the normalization factors here. Let’s talk in person rather than here?

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +196,240 @@ TEST(GaussianMixtureFactor, Error) {
   DiscreteValues discreteValues;
   discreteValues[m1.first] = 1;
   EXPECT_DOUBLES_EQUAL(
-      4.0, mixtureFactor.error({continuousValues, discreteValues}),
-      1e-9);
+      4.0, mixtureFactor.error({continuousValues, discreteValues}), 1e-9);
+}
+
+/* ************************************************************************* */
+// Test components with differing means
```
dellaert: Can you add an extensive comment about this example ? Also, I thought we were going to focus on the HBN -> HFG conversion. Not opposed to showing specifying factors directly is equivalent but I don’t see that happening here.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Same comment. I don’t think this is easy. The HBN -> HFG is easy.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +196,240 @@ TEST(GaussianMixtureFactor, Error) {
...
+  double x1 = 1.0, x2 = 1.0;
+  values.insert(X(1), x1);
+  values.insert(X(2), x2);
+
+  double between = 0.0;
+
+  auto model0 = noiseModel::Isotropic::Sigma(1, 1e2);
+  auto model1 = noiseModel::Isotropic::Sigma(1, 1e-2);
+  auto prior_noise = noiseModel::Isotropic::Sigma(1, 1e-3);
+
```
dellaert: Same comment….

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -197,24 +200,10 @@ std::shared_ptr<GaussianMixtureFactor> GaussianMixture::likelihood(
...
-          // are not all equal.
-          GaussianFactorGraph gfg;
-          gfg.push_back(likelihood_m);
-          Vector c(1);
-          c << std::sqrt(2.0 * Cgm_Kgcm);
-          auto constantFactor = std::make_shared<JacobianFactor>(c);
-          gfg.push_back(constantFactor);
-          return std::make_shared<JacobianFactor>(gfg);
-        }
+        return likelihood_m;
```
dellaert: The likelihood should hide the normalization factor to be correct, I think. It seems maybe (i can't fully follow anymore) that you reverted this in the next PR, but the PR should be correct when it goes into develop. Let's resolve this in a meeting rather than here...

dellaert:
@varunagrawal let me know when the examples are in here and I should look again.

dellaert:
PS, avoid pushing each commit to save on CI cycles :-)

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +197,422 @@ TEST(GaussianMixtureFactor, Error) {
...
+/**
+ * Test a simple Gaussian Mixture Model represented as P(m)P(z|m)
+ * where m is a discrete variable and z is a continuous variable.
+ * m is binary and depending on m, we have 2 different means
+ * and covariances each for the
+ * Gaussian distribution around which we sample z.
+ *
+ * The resulting factor graph should eliminate to a Bayes net
+ * which represents a sigmoid function leaning towards
+ * the tighter covariance Gaussian.
```
dellaert: really?

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +197,422 @@ TEST(GaussianMixtureFactor, Error) {
...
+  // The result should be a sigmoid.
+  // So should be m = 0.5 at z=3.0 - 1.0=2.0
+  VectorValues given;
+  given.insert(z, Vector1(mu1 - mu0));
+
+  HybridGaussianFactorGraph gfg = hbn.toFactorGraph(given);
+  HybridBayesNet::shared_ptr bn = gfg.eliminateSequential();
+
+  HybridBayesNet expected;
+  expected.emplace_back(new DiscreteConditional(m, "0.5/0.5"));
```
dellaert: amazing :-)

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +197,422 @@ TEST(GaussianMixtureFactor, Error) {
...
+
+  auto gm = new GaussianMixture({z}, {}, {m}, {c0, c1});
+  auto mixing = new DiscreteConditional(m, "0.5/0.5");
+
+  HybridBayesNet hbn;
+  hbn.emplace_back(gm);
+  hbn.emplace_back(mixing);
+
+  // The result should be a sigmoid.
+  // So should be m = 0.5 at z=3.0 - 1.0=2.0
```
dellaert: Gravy: derive sigmoid, add a lambda, and compare at 3 different places.

Since $P(z,m) = P(z|m)P(m)$, and $P(m)=50/50$, we have 

$$
P(m|z) \propto P(z|m) \propto exp(-|z-\mu_m|^2/4)
$$

$$
P(0|z) = \frac{exp(-|z-\mu_0|^2/4)}{exp(-|z-\mu_0|^2/4)+exp(-|z-\mu_1|^2/4)}
$$

$$
P(0|z) = \frac{1}{1+exp(|z-\mu_0|^2/4-|z-\mu_1|^2/4)}
$$

image

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +197,422 @@ TEST(GaussianMixtureFactor, Error) {
...
+
+  auto gm = new GaussianMixture({z}, {}, {m}, {c0, c1});
+  auto mixing = new DiscreteConditional(m, "0.5/0.5");
+
+  HybridBayesNet hbn;
+  hbn.emplace_back(gm);
+  hbn.emplace_back(mixing);
+
+  // The result should be a sigmoid leaning towards model1
+  // since it has the tighter covariance.
```
dellaert: Actually *not* a sigmoid:

Since $P(z,m) = P(z|m)P(m)$, and $P(m)=50/50$, we have 

$$
P(m|z) \propto P(z|m) \propto exp(-|z-\mu_m|^2/\sigma_m^2) / \sqrt{2 \pi \sigma_m^2}
$$

Hence:

$$
P(0|z) = \frac{exp(-|z-1|^2/64) / 8}
{exp(-|z-1|^2/64) / 8 + exp(-|z-3|^2/16) / 4}
$$

image

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +197,422 @@ TEST(GaussianMixtureFactor, Error) {
...
+  HybridBayesNet expected;
+  expected.emplace_back(
+      new DiscreteConditional(m, "0.338561851224/0.661438148776"));
+
+  EXPECT(assert_equal(expected, *bn));
+}
+
+/* ************************************************************************* */
+/**
+ * Test a model P(x0)P(z0|x0)p(x1|m1)p(z1|x1)p(m1).
```
dellaert: Puzzled. If this is a motion model, should it not depend on x0 ? i.e., p(x1|m1) -> p(x1|x0, m1) ?

dellaert (commented):
I did the derivations for sigmoid and *non*-sigmoid so you can use them in the unit tests. I am confused about the "motion model" example as I think the motion model should depend on both x0 *and* x1

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: The example that we had on the board leads to this. For the motion model to be dependent on both x0 and x1, we need the measurement z being dependent on both x0 and x1 aka `P(z|x1, x0, m1)` which will give us `L(x1, x0, m1)` as the motion model.

dellaert: That’s right, and I thought that’s what we had. That’s definitely the intent for the state estimation-like example.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -233,6 +233,26 @@ continuousElimination(const HybridGaussianFactorGraph &factors,
...
 
+/* ************************************************************************ */
+/**
+ * @brief Exponential log-probabilities after performing
+ * the necessary normalizations.
+ *
+ * @param logProbabilities DecisionTree of log-probabilities.
+ * @return AlgebraicDecisionTree<Key>
+ */
+static AlgebraicDecisionTree<Key> exponentiateLogProbabilities(
```
dellaert: Consider renaming to probabilitiesFromLogValues and replace “necessary normalization” in comment with: “Exponentiate log-values, not necessarily normalized, normalize, and return as ADT.”

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Still only compared at special case 50/50 point. Could be wrong in other places.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +197,421 @@ TEST(GaussianMixtureFactor, Error) {
...
+
+  // The result should be a bell curve like function
+  // with m1 > m0 close to 3.1333.
+  VectorValues given;
+  given.insert(z, Vector1(3.133));
+  HybridGaussianFactorGraph gfg = hbn.toFactorGraph(given);
+  HybridBayesNet::shared_ptr bn = gfg.eliminateSequential();
+
+  HybridBayesNet expected;
+  expected.emplace_back(
```
dellaert: Please always either indicate something is a regression or alternatively, how you calculated the expected value. In this case, I would do neither but use the closed form function I computed in an earlier comment and use that to check in at least three places, same as above.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Still eagerly waiting for this example to be fixed :-)

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +197,421 @@ TEST(GaussianMixtureFactor, Error) {
...
+    // Since we have a measurement on z2, we get a definite result
+    DiscreteConditional expected(m1, "0.4923083/0.5076917");
+
+    EXPECT(assert_equal(expected, *(bn->at(2)->asDiscrete()), 1e-6));
+  }
+}
+
+/* ************************************************************************* */
+/**
+ * Test a model P(x0)P(z0|x0)p(x1|m1)p(z1|x1)p(m1).
```
dellaert: When fixing, consider using a namespace as a “test fixture” to avoid copy/paste

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: Should I kill this one in favor of `DifferentMeans`? If I change this one, they both will be the same example.

dellaert: I think we had decided on a HBN that has P(x1|x0,m), gets converted to FG, and there would be two versions: one with one measurement on x0, one with two measurements which would inform m.

dellaert: I just looked at differentMeans and I am confused:
I had expected
1. P(x0)P(x1|x1,m)P(m)P(z0|x0)
2. P(x0)P(x1|x1,m)P(m)P(z0|x0)P(z1|x1)
Just like we had on the whiteboard. P(z|x0,x1,m) is something different altogether which resembles (2). Also interesting and should probably give the same result if z==z1-z0, and that would be an interesting test. 

But indeed, if those *three* examples are there, HBN->HFG, then these ones are obsolete.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: Derived this in my thesis doc and updated this accordingly. It is **not** a sigmoid.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: Coolio

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +198,323 @@ TEST(GaussianMixtureFactor, Error) {
...
+
+namespace test_gmm {
+
+/**
+ * Function to compute P(m=1|z). For P(m=0|z), swap `mus and sigmas.
+ * Follows equation 7.108 since it is more generic.
+ */
+double sigmoid(double mu0, double mu1, double sigma0, double sigma1, double z) {
+  double x1 = ((z - mu0) / sigma0), x2 = ((z - mu1) / sigma1);
+  double d = std::sqrt(sigma0 * sigma0) / std::sqrt(sigma1 * sigma1);
```
dellaert: Beautiful! But simplify d:
```
d = sigma0/sigma;
```

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +198,323 @@ TEST(GaussianMixtureFactor, Error) {
...
+ *
+ * The resulting factor graph should eliminate to a Bayes net
+ * which represents a sigmoid function.
+ */
+TEST(GaussianMixtureFactor, GaussianMixtureModel) {
+  using namespace test_gmm;
+
+  double mu0 = 1.0, mu1 = 3.0;
+  double sigma = 2.0;
+  auto model = noiseModel::Isotropic::Sigma(1, sigma);
```
dellaert: Model no longer needed below. Linter should complain?

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +198,323 @@ TEST(GaussianMixtureFactor, Error) {
...
+  hbn.emplace_back(mixing);
+
+  return hbn;
+}
+
+}  // namespace test_gmm
+
+/* ************************************************************************* */
+/**
+ * Test a simple Gaussian Mixture Model represented as P(m)P(z|m)
```
dellaert, line 245: This test is now officially awesome :-)

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +198,323 @@ TEST(GaussianMixtureFactor, Error) {
...
+      4.0, mixtureFactor.error({continuousValues, discreteValues}), 1e-9);
+}
+
+namespace test_gmm {
+
+/**
+ * Function to compute P(m=1|z). For P(m=0|z), swap `mus and sigmas.
+ * Follows equation 7.108 since it is more generic.
+ */
+double sigmoid(double mu0, double mu1, double sigma0, double sigma1, double z) {
```
dellaert: Not necessarily a sigmoid :-)

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Could we copy the pattern from above and make this test also awesome, in addition to the regression?

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +198,323 @@ TEST(GaussianMixtureFactor, Error) {
...
+  // Set a prior P(x0) at x0=0
+  hbn.emplace_back(
+      new GaussianConditional(x0, Vector1(0.0), I_1x1, prior_model));
+
+  // Add measurement P(z0 | x0)
+  auto p_z0 = new GaussianConditional(z0, Vector1(0.0), -I_1x1, x0, I_1x1,
+                                      measurement_model);
+  hbn.emplace_back(p_z0);
+
+  // Add hybrid motion model
```
dellaert, line 408: I think this is still wrong :-/
I think we simply need 
```
P(x1| x0,m) = 
N(x0+mu0,sigma0) if m=0
N(x1+mu1, sigma1) if m=1
```
There is no need for a dummy measurement f01.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +198,323 @@ TEST(GaussianMixtureFactor, Error) {
...
+    HybridBayesNet::shared_ptr bn = gfg.eliminateSequential();
+
+    // Since no measurement on x1, we hedge our bets
+    DiscreteConditional expected(m1, "0.5/0.5");
+    // regression
+    EXPECT(assert_equal(expected, *(bn->at(2)->asDiscrete())));
+  }
+
+  {
+    // Now we add a measurement z1 on x1
```
dellaert, line 473: Nice!!!

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,8 +198,323 @@ TEST(GaussianMixtureFactor, Error) {
...
+    hbn = CreateBayesNet(mu0, mu1, sigma0, sigma1, true);
+
+    given.insert(z1, Vector1(2.2));
+    HybridGaussianFactorGraph gfg = hbn.toFactorGraph(given);
+    HybridBayesNet::shared_ptr bn = gfg.eliminateSequential();
+
+    // Since we have a measurement on z2, we get a definite result
```
dellaert, line 580: Indicate regression. Would be even better if we could calculate the correct result. Probably you could even use your sigmoid function above, after manually eliminating a bit? Not sure.

gtsam/hybrid/tests/testMixtureFactor.cpp
```diff
@@ -115,6 +118,156 @@ TEST(MixtureFactor, Dim) {
   EXPECT_LONGS_EQUAL(1, mixtureFactor.dim());
 }
 
+/* ************************************************************************* */
+// Test components with differing means
```
dellaert: I need more explanation in the comment to understand these tests. Also, since in the end they are just regression tests, it’s unclear they assert correctness of anything :-/ if these examples are supposed to be equivalent to the HBN tests above, and you are using the same values as there to test the result, that is a different story. Unclear now, though.

dellaert (commented):
95% there, awesome progress.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: Wait, why not? For x1 it is and for x0 it is a negative sigmoid.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: So 2 things about this:
1. For `m=1`, did you mean `x0 + mu1`?
2. If `x1` is the frontal variable, then we would have to give some value for `x1` to compute the likelihood which initially came off to me as a chicken and egg problem.

Deeper thought has led me to believe that we provide the initial value of `x1` as a pseudo-measurement $\tilde{x1}$. I am guessing this is what you had in mind as well. Asking because we haven't discussed this and this is purely my thoughts.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Well, only if sigma0==sigma1. I think it also covers the second test below, no? The “gaussian-like” thing. No? I thought that was the whole intent :-)

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Yes, $x0+\mu1$, sorry.
You should not need a measurement at all. $P(x1|x0,x1)$ becomes a factor $\phi(x0,x1,m)$. If our code can’t handle this we learned something :-) let’s chat.

varunagrawal: Yup, the code doesn't handle this, since to convert to a likelihood, we need to provide the frontal variable. The likelihood then removes the frontal variable, which means $P(x1|x0, m1)$ becomes $\phi(x0, m1)$.

dellaert: As we discussed, I think it does. Eager to see it work :-)

dellaert:
seems you have a compile issue. Could you maybe only push once a review is needed? That will eliminate some green-house gases when CI is run for entire matrix.
```
/home/runner/work/gtsam/gtsam/gtsam/hybrid/HybridGaussianFactorGraph.cpp:554:76: error: use of undeclared identifier 'f'
    if (auto gaussianMixture = dynamic_pointer_cast<GaussianMixtureFactor>(f)) {
```

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: Yes, I wrote down the math and derivation and it only is a sigmoid when we set sigma1==sigma0.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: Yup! I did this for the GaussianMixture2 test. Used `prob_m_z` (the new name for the sigmoid function).

gtsam/hybrid/tests/testMixtureFactor.cpp
varunagrawal: I should remove these since they are beyond the scope of this PR.

dellaert:
Assuming the not 50-50 motion model test is the last bit. Let me know by email when I should look again.

varunagrawal:
> Assuming the not 50-50 motion model test is the last bit. Let me know by email when I should look again.

Yup, working on that.

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -86,7 +87,22 @@ GaussianFactorGraphTree GaussianMixture::add(
...
+          this->logConstant_ - gc->logNormalizationConstant();
+      // If there is a difference in the covariances, we need to account for
+      // that since the error is dependent on the mode.
+      if (Cgm_Kgcm > 0.0) {
+        // We add a constant factor which will be used when computing
+        // the probability of the discrete variables.
+        Vector c(1);
+        c << std::sqrt(2.0 * Cgm_Kgcm);
+        auto constantFactor = std::make_shared<JacobianFactor>(c);
+        return GaussianFactorGraph{gc, constantFactor};
```
dellaert, line 103: Ok, nice!

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -316,19 +334,49 @@ AlgebraicDecisionTree<Key> GaussianMixture::logProbability(
...
   auto errorFunc = [&](const GaussianConditional::shared_ptr &conditional) {
-    return conditional->error(continuousValues) +  //
-           logConstant_ - conditional->logNormalizationConstant();
+    // Check if valid pointer
+    if (conditional) {
+      return conditional->error(continuousValues) +  //
+             logConstant_ - conditional->logNormalizationConstant();
+    } else {
+      // If not valid, pointer, it means this conditional was pruned,
+      // so we return maximum error.
```
dellaert: Explain in comment why

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -316,19 +334,49 @@ AlgebraicDecisionTree<Key> GaussianMixture::logProbability(
...
   DecisionTree<Key, double> error_tree(conditionals_, errorFunc);
   return error_tree;
 }
 
 /* *******************************************************************************/
 double GaussianMixture::error(const HybridValues &values) const {
+  // Check if discrete keys in discrete assignment are
+  // present in the GaussianMixture
+  KeyVector dKeys = this->discreteKeys_.indices();
+  bool valid_assignment = false;
```
dellaert: This is a lot of work for checking valid input. What would happen if we did not check?

gtsam/hybrid/GaussianMixture.cpp
```diff
@@ -316,19 +334,49 @@ AlgebraicDecisionTree<Key> GaussianMixture::logProbability(
...
   auto conditional = conditionals_(values.discrete());
-  return conditional->error(values.continuous()) +  //
-         logConstant_ - conditional->logNormalizationConstant();
+  if (conditional) {
+    return conditional->error(values.continuous()) +  //
+           logConstant_ - conditional->logNormalizationConstant();
+  } else {
+    // If not valid, pointer, it means this conditional was pruned,
+    // so we return maximum error.
+    return std::numeric_limits<double>::max();
```
dellaert: Hmmm. These lines of code (both paths) seemed to be copy-pasted. Isolate in method or static function?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -279,21 +311,32 @@ GaussianFactorGraphTree removeEmpty(const GaussianFactorGraphTree &sum) {
 using Result = std::pair<std::shared_ptr<GaussianConditional>,
                          GaussianMixtureFactor::sharedFactor>;
 
-// Integrate the probability mass in the last continuous conditional using
-// the unnormalized probability q(μ;m) = exp(-error(μ;m)) at the mean.
-//   discrete_probability = exp(-error(μ;m)) * sqrt(det(2π Σ_m))
+/**
+ * Compute the probability q(μ;m) = exp(-error(μ;m)) * sqrt(det(2π Σ_m)
```
dellaert: Does this agree with your definition of q in your thesis? I thought we arrived at q == 1…

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -480,14 +523,15 @@ EliminateHybrid(const HybridGaussianFactorGraph &factors,
         std::inserter(continuousSeparator, continuousSeparator.begin()));
 
     // Similarly for the discrete separator.
-    KeySet discreteSeparatorSet;
-    std::set<DiscreteKey> discreteSeparator;
     auto discreteKeySet = factors.discreteKeySet();
+    // Use set-difference to get the difference in keys
```
dellaert: Improve comment. Don’t understand what is happening without reading the code. Why are we doing this? Explain in comment.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -191,13 +198,395 @@ TEST(GaussianMixtureFactor, Error) {
...
+  hbn.emplace_back(p_m1);
+
+  return hbn;
+}
+
+}  // namespace test_two_state_estimation
+
+/* ************************************************************************* */
+/**
+ * Test a model P(x0)P(z0|x0)P(x1|x0,m1)P(z1|x1)P(m1).
```
dellaert: No prior anymore

dellaert (approved):
You did it! Some more detail comments that you can resolve yourself if taken care of. Awesome!

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: This is a different q. I'll change it to `p(m)`.

gtsam/hybrid/GaussianMixture.cpp
varunagrawal: Done

gtsam/hybrid/GaussianMixture.cpp
varunagrawal: Apparently nothing. I may have needed it a while ago but things have evolved and now it's no longer needed. :)

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: I refactored this since we don't need a set difference. All the discrete keys should be the discrete separator when doing `hybridElimination`.

gtsam/hybrid/GaussianMixture.h
```diff
@@ -256,6 +256,10 @@ class GTSAM_EXPORT GaussianMixture
   /// Check whether `given` has values for all frontal keys.
   bool allFrontalsGiven(const VectorValues &given) const;
 
+  /// Helper method to compute the error of a conditional.
+  double conditionalError(const GaussianConditional::shared_ptr &conditional,
```
dellaert, line 260: Static? Fix in next PR…

varunagrawal: It needs the logConstant_ member, so can't be static. I did mark it private though.

gtsam/hybrid/tests/testMixtureFactor.cpp
dellaert: Did you end up removing them?

varunagrawal: Just removed them.

dellaert:
CI passed, want to merge?

PR #1805, opened 2024-08-22
Direct Hybrid Factor Specification

varunagrawal:
Add tests which illustrate how we can specify hybrid factors directly instead of defining the BayesNet and calling `toFactorGraph` on it.

This is useful since it allows us to handle nonlinear functions in the conditionals being converted to factors. This will also help with relinearization, a long standing issue for us in this scheme.


dellaert:
I’d like to chat about this more. While the HBN -> HFG is now very clear, the proposed API here is not. Where do the log normalizers come from? What is their correct value? What do they mean? If the answer is: “think about the conditional and its normalizer” then why not use the “likelihood” approach, or just directly add a Hybrid conditional? Let’s chat in person because I think it requires higher bandwidth than comments back and forth…

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
```diff
@@ -741,6 +741,168 @@ TEST(HybridGaussianFactor, TwoStateModel4) {
...
+  auto f0 =
+      std::make_shared<BetweenFactor<double>>(X(0), X(1), means[0], model0)
+          ->linearize(values);
+  auto f1 =
+      std::make_shared<BetweenFactor<double>>(X(0), X(1), means[1], model1)
+          ->linearize(values);
+
+  // Create HybridGaussianFactor
+  std::vector<GaussianFactorValuePair> factors{
+      {f0, ComputeLogNormalizer(model0)}, {f1, ComputeLogNormalizer(model1)}};
```
dellaert: I'm still very shaky about the semantics of this scalar. I asked a question on the previous PR but I don;t think it was answered. If the scalar simply *adds* to the error, then is that after squaring and halving? If so, I think this needs to be better documented, with an example in the .h file even. There is also uncertainty about the sign: is the normalizer multiplying or dividing? I think the former, but again: be very clear, also in the docs of ComputeLogNormalizer. And, very important, it has to be consistent everywhere.

dellaert:
Please update PR comment?

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
varunagrawal: I sent you the doc with the math written up there. From the derivation, we are adding the pre-halved version. You are right that it should be better documented in the .h file. :)

`ComputeLogNormalizer` returns $\log(2\pi \Sigma^m)$ instead of $\log(\frac{1}{\sqrt{2\pi \Sigma^m}})$, so we should be dividing (or subtracting since it is in log-space). This is actually convenient since we can pull out $-\frac{1}{2}$ in common from both the factor error and the normalizer term. I'm still blown away by how nicely that math worked out.

dellaert: OK, this answer is problematic to me: when specifying the API for hybrid factor graphs, the normalizer is *just an application*. Ideally, the API in the .h file is documented like:
```
In factor graphs the error function typically returns 0.5*|h(x)-z|^2, i.e., the negative log-likelihood for a Gaussian noise model. In hybrid factor graphs we allow *adding* an arbitrary scalar dependent on the discrete assignment. For example, adding a 70/30 mode probability is supported by providing the scalars $-log(.7)$ and $-log(.3)$. Note that adding a common constant will not make any difference in the optimization, so $-log(70)$ and $-log(30)$ work just as well.
```
Please consider adding the above in HNF and the linear version in HGF. And, not that I am advocating here for **adding** these scalars **terms** to the error. When designing the API for the factors, it does not make sense to tell people that a term will be subtracted: we either work in negative log space (we add) or in probability space (we multiply). I see no need to add subtraction into this lexicon.

dellaert: Happy to chat in person BTW. Invite me.

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -45,6 +45,15 @@ using GaussianFactorValuePair = std::pair<GaussianFactor::shared_ptr, double>;
  * where the set of discrete variables indexes to
  * the continuous gaussian distribution.
  *
+ * In factor graphs the error function typically returns 0.5*|h(x)-z|^2, i.e.,
```
dellaert: Here it would be |A*x-b|

varunagrawal: Done

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
varunagrawal: As discussed, I have made all the changes.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -56,7 +56,7 @@ HybridGaussianFactor::Factors augment(
                           const HybridGaussianFactor::sharedFactor &gf) {
     auto jf = std::dynamic_pointer_cast<JacobianFactor>(gf);
     if (!jf) return gf;
-    // If the log_normalizer is 0, do nothing
+    // If the value is 0, do nothing
     if (values(assignment) == 0.0) return gf;
```
dellaert: You are calling the potentially expensive `values(assignment)` twice, so please assign it to a variable.  But why unzip the tree and then call `values()` at all. Can't you just apply on the original tree and get the value for free?

A more problematic issue is that you take the square root. But is this wrong now? Should we not do sqrt(2*value) ? Because the "hidden" value c will be squared and halved: `value == 0.5*c^2`.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
```diff
@@ -741,6 +741,170 @@ TEST(HybridGaussianFactor, TwoStateModel4) {
...
+      std::make_shared<BetweenFactor<double>>(X(0), X(1), means[0], model0)
+          ->linearize(values);
+  auto f1 =
+      std::make_shared<BetweenFactor<double>>(X(0), X(1), means[1], model1)
+          ->linearize(values);
+
+  // Create HybridGaussianFactor
+  // We multiply by -2 since the we want the underlying scalar to be log(|2πΣ|)
+  std::vector<GaussianFactorValuePair> factors{
+      {f0, -2 * model0->logNormalizationConstant()},
```
dellaert: I don't think the factor 2 belongs here.

gtsam/hybrid/tests/testHybridNonlinearFactorGraph.cpp
```diff
@@ -857,21 +857,23 @@ namespace test_relinearization {
...
-  auto prior_noise = noiseModel::Isotropic::Sigma(1, 1e-3);
+  auto prior_noise = noiseModel::Isotropic::Sigma(1, measurement_noise);
 
   auto f0 =
       std::make_shared<BetweenFactor<double>>(X(0), X(1), means[0], model0);
   auto f1 =
       std::make_shared<BetweenFactor<double>>(X(0), X(1), means[1], model1);
 
   // Create HybridNonlinearFactor
+  // We multiply by -2 since the we want the underlying scalar to be log(|2πΣ|)
```
dellaert: I think that's wrong.

gtsam/linear/NoiseModel.cpp
```diff
@@ -238,6 +238,25 @@ void Gaussian::WhitenSystem(Matrix& A1, Matrix& A2, Matrix& A3, Vector& b) const
...
+double Gaussian::logNormalizationConstant() const {
+  // Since noise models are Gaussian, we can get the logDeterminant easily
+  // Sigma = (R'R)^{-1}, det(Sigma) = det((R'R)^{-1}) = det(R'R)^{-1}
+  // log det(Sigma) = -log(det(R'R)) = -2*log(det(R))
+  // Hence, log det(Sigma)) = -2.0 * logDeterminant()
+  // which gives log = -0.5*n*log(2*pi) - 0.5*(-2.0 * logDeterminant())
+  //     = -0.5*n*log(2*pi) + (0.5*2.0 * logDeterminant())
+  //     = -0.5*n*log(2*pi) + logDeterminant()
+  double logDetR =
+      R().diagonal().unaryExpr([](double x) { return log(x); }).sum();
```
dellaert: This is very expensive in case of diagonal, isotropic, unit, because R() creates a *full* matrix. You should override in Diagonal and use invsigmas() there. In Isotropic, there is an even simpler expression, and in Unit even simpler.

gtsam/linear/NoiseModel.h
```diff
@@ -266,7 +266,20 @@ namespace gtsam {
       /// Compute covariance matrix
       virtual Matrix covariance() const;
 
-    private:
+      /**
+       * @brief Helper method to compute the normalization constant
```
dellaert: Not a helper, just a method

gtsam/linear/NoiseModel.h
```diff
@@ -266,7 +266,20 @@ namespace gtsam {
...
+       * for a Gaussian noise model.
+       * k = 1/log(\sqrt(|2πΣ|))
```
dellaert: I thought we defined k as $1/(\sqrt(|2πΣ|)$. This should return $\log(k)$ (and say so).

gtsam/linear/tests/testNoiseModel.cpp
```diff
@@ -807,23 +807,22 @@ TEST(NoiseModel, NonDiagonalGaussian)
   }
 }
 
-TEST(NoiseModel, ComputeLogNormalizer) {
+TEST(NoiseModel, ComputeLogNormalizerConstant) {
```
dellaert: logNormalizationConstant

dellaert: This is a good test for Isotropic, but we also need tests for Gaussian, Diagonal, and Unit :-/

dellaert (commented):
Very close, but still one main issue:
- I think the factor 2 is still in the wrong place for the GaussianHybridFactor

Two possibly significant performance issues:
- unzipping and then undoing the unzip is needlessly expensive.
- only having the Gaussian logNormalization constant might be *very* expensive. 

If you want I can create a PR with the noiseModel changes if you finish the other things.

gtsam/hybrid/HybridGaussianFactor.cpp
varunagrawal: 1. So I need to unzip to normalize the values (so that we don't have negatives, else the sqrt(value) gives a nan). This brings up an interesting point though: should we normalize for the user or should the user pre-normalize? I felt like us doing it makes sense, since the sqrt to hide the variable is something we need to do and the user shouldn't have to worry about those semantics. Thoughts?

2. We can do 2*value, but that is going to have to also change in various other places like `HybridGaussianConditional::likelihood`. I think what you suggest is the better thing, keeping in alignment with my previous point.

gtsam/hybrid/tests/testHybridNonlinearFactorGraph.cpp
varunagrawal: According to the math, this is correct, though it is another way of doing things.
We can hide $\sqrt{\log(|2\pi \Sigma|)}$ so that on squaring and halving, we get $\frac{1}{2}\log(|2\pi \Sigma|) = \log(\sqrt{|2\pi \Sigma|})$.

However, since we've decided to premultiply by 2 right before hiding (inside the `augment` method), then yeah I'll have to pass in $\log(\sqrt{|2\pi \Sigma|})$.

gtsam/linear/NoiseModel.h
varunagrawal: Oops sorry about that.

gtsam/linear/NoiseModel.cpp
varunagrawal: I was going to do this in a follow up PR so less to review, but I'll put it here.

gtsam/linear/tests/testNoiseModel.cpp
varunagrawal: Added tests for all 4 noise models.

gtsam/linear/NoiseModel.cpp
varunagrawal: Done. I defined a new protected class method `logDetR` which isolates and takes advantage of computational efficiencies. This way we can only define `logNormalizationConstant` once and avoid duplication.

dellaert (approved):
Awesome! Liftoff!

PR #1808, opened 2024-08-25
Python port of `LocalizationExample.cpp`

### LINKED ISSUE #1754: LocalizationExample impossible porting to Python
Hello,

I am trying to use the Python version of GTSAM for FGO applied on GNSS data and the first toy example I was interested in implementing was exactly the one from gtsam/examples/LocalizationExample.cpp. I added the measurements as unary factors using the CustomFactor, all the other variables are exactly the same as in LocalizationExample.cpp, and the optimization result I got is this: 

result png

I couldn't get any better results and then discovered that actually LocalizationExample is impossible to port to Python according to gtsam/python/gtsam/examples/README.md. Can comebody give me more details why this is? Is using the pure C++ the only solution for implementing this use case?

Thank you!

varunagrawal:
Following the discussion in #1754, I have written a python version of the `LocalizationExample.cpp` file.
Instead of the Unary factor, I make use of the `PartialPriorFactor` on the Pose2 elements for use with the GPS-like measurements.

Comparing the output of the C++ and Python scripts, I get the same output modulo python formatting rules.

### DIFF
--- a/python/gtsam_unstable/gtsam_unstable.tpl
+++ b/python/gtsam_unstable/gtsam_unstable.tpl
@@ -9,6 +9,7 @@
 
 #include <pybind11/eigen.h>
 #include <pybind11/stl_bind.h>
+#include <pybind11/stl.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/functional.h>
 #include <pybind11/iostream.h>
--- a/gtsam_unstable/gtsam_unstable.i
+++ b/gtsam_unstable/gtsam_unstable.i
@@ -671,6 +671,21 @@ class AHRS {
   //void print(string s) const;
 };
 
+#include <gtsam_unstable/slam/PartialPriorFactor.h>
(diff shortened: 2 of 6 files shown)


python/gtsam_unstable/gtsam_unstable.tpl
```diff
@@ -9,6 +9,7 @@
 
 #include <pybind11/eigen.h>
 #include <pybind11/stl_bind.h>
+#include <pybind11/stl.h>
```
ProfFan, line 12: Is it possible to do this without `stl.h`? If not it's good then.

ProfFan (approved):

python/gtsam_unstable/gtsam_unstable.tpl
varunagrawal: We already have `stl.h` in the `gtsam.tpl` so this is fine I guess?

dellaert:
Nice!

PR #1810, opened 2024-08-25
fix C++ version issue of CppUnitLite

### LINKED ISSUE #1815: GTSAM will not build even with tests switched off under latest boost due to wrong C++ version for CppUnitLite
This is due to the GTSAM feature (C++11/C++17) is not propagated to the bundled CppUnitLite.

Also discovered that CppUnitLite depends on `boost/lexical_cast` regardless of the enable boost setting.

This will be addressed in a later PR.

ProfFan:
Will not build as `boost/lexical_cast` (CppUnitLite's dependency) requires C++11 and newer.

Partially fixes #1815


dellaert:
Seems ok but puzzled: what problem did it solve? All CI worked, right? Might be good to be more detailed in PR description on what problem was encountered and ho this fixes it. Maybe even by creating an issue that this PR resolves?

ProfFan:
#1815 1815

ProfFan:
This is a band-aid fix, to totally fix this issue we need to get rid of the include in `CppUnitLite`.

dellaert (approved):

PR #1811, opened 2024-08-25
Fix warnings and add M1 runner

dellaert:
- fixed some c++ warnings in vendored libraries and one of our tests
- updated checkout actions
- added an M1 action for python to see whether segfaults are reproducible in CI (they are not!)
- made sure make python-install also works in a virtual env (which is used for Mac in CI)
- added an M1 action for regular GTSAM

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -26,6 +26,7 @@ jobs:
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
         name: [
           macos-12-xcode-14.2,
+          macos-14-xcode-15.4,
         ]
 
         build_type: [Debug, Release]
@@ -35,6 +36,11 @@ jobs:
             os: macos-12
             compiler: xcode
             version: "14.2"
+
+          - name: macos-14-xcode-15.4
+            os: macos-14
+            compiler: xcode
+            version: "15.4"
 
     steps:
       - name: Checkout
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -38,7 +38,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Install Dependencies
         run: |
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -110,11 +110,6 @@ jobs:
           echo "CC=clang" >> $GITHUB_ENV
           echo "CXX=clang++" >> $GITHUB_ENV
 
-          python$PYTHON_VERSION -m venv venv
-          source venv/bin/activate
-          echo "PATH=$(pwd)/venv/bin:$PATH" >> $GITHUB_ENV
-          python -m pip install --upgrade pip
-
       - name: Setup msbuild (Windows)
         if: runner.os == 'Windows'
         uses: ilammy/msvc-dev-cmd@v1
@@ -175,7 +170,7 @@ jobs:
 
       - name: Install Python Dependencies
         shell: bash
-        run: python$PYTHON_VERSION -m pip install -r python/dev_requirements.txt
+        run: python$PYTHON_VERSION -m pip install --break-system-packages --user -r python/dev_requirements.txt
 
       - name: Build
         shell: bash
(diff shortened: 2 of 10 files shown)


.github/workflows/build-macos.yml
```diff
@@ -26,6 +26,7 @@ jobs:
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
         name: [
           macos-12-xcode-14.2,
+          macos-14-xcode-15.4,
```
ProfFan, line 29: M1 runner is super expensive, don't know if you want it to be on each PR update...

A manual trigger is probably better, just saying.

python/CMakeLists.txt
```diff
@@ -264,11 +264,13 @@ if(GTSAM_UNSTABLE_BUILD_PYTHON)
 endif()
 
 # Add custom target so we can install with `make python-install`
+# Note below we make sure to install with --user iff not in a virtualenv
 set(GTSAM_PYTHON_INSTALL_TARGET python-install)
 add_custom_target(${GTSAM_PYTHON_INSTALL_TARGET}
-        COMMAND ${PYTHON_EXECUTABLE} -m pip install --user .
+        COMMAND ${PYTHON_EXECUTABLE} -c "import sys, subprocess; cmd = [sys.executable, '-m', 'pip', 'install']; has_venv = hasattr(sys, 'real_prefix') or (hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix); cmd.append('--user' if not has_venv else ''); cmd.append('.'); subprocess.check_call([c for c in cmd if c])"
```
ProfFan, line 269: I don't know if the distro packaging people would want this, tagging @jlblancoc and @acxz 

A default-on option here could be useful

ProfFan (approved):

.github/workflows/build-python.yml
```diff
@@ -162,6 +168,14 @@ jobs:
         run: |
           bash .github/scripts/python.sh -d
 
+      - name: Create virtual on MacOS
```
varunagrawal, line 171: Any specific reason we want to use venvs instead of the installed Python? We could also extend the above `Setup python` step to install the desired python version.

PS I am not sure how python dev libraries play in a venv.

varunagrawal (commented):
A comment on the venv and I second Fan's comment on packaging.

.github/workflows/build-macos.yml
dellaert: Hmmm. Expensive in what sense?

python/CMakeLists.txt
dellaert: Let me know. In a venv --user does not work, and this was achieved after a lot of experimentation.

.github/workflows/build-python.yml
dellaert: Yes, very specific: on M1 the installed python is tagged as not modifiable. If you see the action failures and the many commits you'll see how I had to fight to get this working. There is no "just do this" solution that I could find. The added benefit is that now people will be able to use virtualenvs, whereas before they could not.

dellaert:
@varunagrawal on develop I now get
```
(py310) (base) MacBook-Pro-8:build dellaert$ make -j6 python-install
[  0%] Built target pybind_wrap_gtsam_unstable
[  0%] Built target gtsam_header
[  3%] Built target pybind_wrap_gtsam
[  3%] Built target gtsam_unstable_header
[ 21%] Built target cephes-gtsam
[ 36%] Built target metis-gtsam
[ 81%] Built target gtsam
[ 90%] Built target gtsam_py
[ 96%] Built target gtsam_unstable
[100%] Built target gtsam_unstable_py
gtsam: Failed to import, skipping
cp: stubs/gtsam/: No such file or directory
make[3]: *** [python/CMakeFiles/python-install] Error 1
make[2]: *** [python/CMakeFiles/python-install.dir/all] Error 2
make[1]: *** [python/CMakeFiles/python-install.dir/rule] Error 2
```
and those cmake lines also conflict with this PR. It's weird as your CI passed, but I'm proposing to comment out your cmake lines in this PR and then merging so develop is functional again. We can then test on a number of platforms to make sure all use cases are covered.

varunagrawal:
Sounds good!

.github/workflows/build-python.yml
jlblancoc: > on M1 the installed python is tagged as not modifiable. 

The same goes on with Ubuntu >=24.04, by the way...

Issue #1720, opened 2024-02-07
Python Segfault on MacOS


drpjm:
## Description

I was running through the `robotics` text on performing MAP with multiple sensors and when computing the unnormalized posterior from a DiscreteConditional likelihood, I get a segfault.

This is running in Python 3.11.6, Mac OSX 14.3. Mac OS reports the following;
```

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libgtsam.4.2.0.dylib          	       0x106085dd8 gtsam::DecisionTree<unsigned long long, double>::Choice::choose(unsigned long long const&, unsigned long) const + 96
1   libgtsam.4.2.0.dylib          	       0x10609f210 gtsam::DiscreteConditional::likelihood(gtsam::DiscreteValues const&) const + 220
2   libgtsam.4.2.0.dylib          	       0x10609f6d4 gtsam::DiscreteConditional::likelihood(unsigned long) const + 136
3   gtsam.cpython-311-darwin.so   	       0x106accfa8 0x106980000 + 1363880
4   gtsam.cpython-311-darwin.so   	       0x106996400 0x106980000 + 91136
5   Python                        	       0x1056425b8 cfunction_call + 60
6   Python                        	       0x1055f78f4 _PyObject_MakeTpCall + 128
7   Python                        	       0x1056d5984 _PyEval_EvalFrameDefault + 42108
8   Python                        	       0x1056ca8c4 PyEval_EvalCode + 168
9   Python                        	       0x1057213f0 run_eval_code_obj + 84
10  Python                        	       0x105721354 run_mod + 112
11  Python                        	       0x105721194 pyrun_file + 148
12  Python                        	       0x105720be4 _PyRun_SimpleFileObject + 268
13  Python                        	       0x10572057c _PyRun_AnyFileObject + 216
14  Python                        	       0x10573d164 pymain_run_file_obj + 220
15  Python                        	       0x10573caa4 pymain_run_file + 72
16  Python                        	       0x10573c384 Py_RunMain + 704
17  Python                        	       0x10573d4c0 Py_BytesMain + 40
18  dyld                          	       0x1864f10e0 start + 2360
```

## Steps to reproduce

I am running this in a Python script, not a Jupyter notebook. I have a conductivity sensor based on the DiscreteConditional in the robotics textbook in [Chapter 2.4.4](https://www.roboticsbook.org/S24_sorter_perception.html#likelihood-with-discrete-sensors).

[...]

PR #1812, opened 2024-08-25
Use bundled `pybind11` for Python wrapper

varunagrawal:
Force the use of the bundled `pybind11` directory when compiling the Python wrapper.
This should hopefully fix issues like in #1720.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -30,7 +30,7 @@ if(POLICY CMP0057)
 endif()
 
 # Use bundled pybind11 version
-find_package(pybind11 CONFIG QUIET PATHS "${PROJECT_SOURCE_DIR}/wrap/pybind11")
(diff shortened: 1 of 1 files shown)


dellaert:
Adding Fan as reviewer

dellaert:
Also, CI fails :-(

dellaert:
I'll kill all CI for this push...

dellaert:
Did you try locally before pushing?

varunagrawal:
Ah it's giving an error because we need to also point cmake to the Pybind11 directory so it can generate the pybind11Config.cmake file.

Figuring that out.

varunagrawal:
@dellaert @ProfFan I think this PR should now fix all the segfault woes. Can you please try it out?

dellaert (approved):
Yep, this fixed the segfaults I saw on my system ! Thanks @varunagrawal !!!!

PR #1813, opened 2024-08-25
Generate Python type hints during build

varunagrawal:
This PR leverages stubgen to generate typehint `.pyi` files which then get packaged along with the binaries.
What this means is that we now get **much better** autocomplete in VSCode and other IDEs that read the pyi files.

Below are screenshots of the autocomplete we get as a result:

Screen Shot 2024-08-25 at 4 14 34 PM
Screen Shot 2024-08-25 at 4 14 48 PM

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -265,12 +265,18 @@ endif()
 
 # Add custom target so we can install with `make python-install`
 set(GTSAM_PYTHON_INSTALL_TARGET python-install)
-add_custom_target(${GTSAM_PYTHON_INSTALL_TARGET}
-        COMMAND stubgen -q -p gtsam &&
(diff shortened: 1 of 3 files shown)


dellaert (approved):
Oh, very cool if this works. Did you try it locally? Consider adding a VS code screenshot with then improved hints to the PR comment...

varunagrawal:
Yes I checked it locally first and I have attached screenshots. It's really amazing!

ProfFan:
I remember when I make [`gtsam-stubs`](https://github.com/ProfFan/gtsam-stubs) that I need to do some extra magic to make all types look correct in VSCode. Maybe `stubgen` has improved that one does not need these hacks?

varunagrawal:
Perhaps. It generally has worked well for me and I wrote this functionality last year.

PR #1814, opened 2024-08-26
Fix stubs generation

varunagrawal:
Changed the stubgen command to use the default `out` directory since the custom directory doesn't seem to be working.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -265,24 +265,22 @@ endif()
 # Add custom target so we can install with `make python-install`
 # Note below we make sure to install with --user iff not in a virtualenv
 set(GTSAM_PYTHON_INSTALL_TARGET python-install)
-add_custom_target(${GTSAM_PYTHON_INSTALL_TARGET}
+#TODO(Varun) Maybe move the long command to script?
+# https://stackoverflow.com/questions/49053544/how-do-i-run-a-python-script-every-time-in-a-cmake-build
+if (NOT WIN32) # WIN32=1 is target platform is Windows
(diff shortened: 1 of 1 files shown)


dellaert (approved):
OK, I checked out this branch and that seems to be working.

tianyilim:
Hi, could I know what configuration you use to get the stubs working? For me the type hinting has stopped working.

For instance, I do the following:

```python3
import gtsam

prior_noise = gtsam.gtsam.noiseModel.Diagonal.Sigmas( ...
```

And Pylance reports: `noiseModel" is not a known attribute of module "gtsam"`

I tried also using `gtsam.noiseModel` and importing as `from gtsam import noiseModel` and `import gtsam.gtsam as gtsam` etc. The code still runs fine, it's just the type hinting that's no longer working for me.

ProfFan:
Hi @tianyilim is it the case that only `noiseModel` is not working and everything else works? If that is the case then it's because `pylance` recognizes the import path wrong.

Two questions:
1. If you say `prior_noise = gtsam.noiseModel.xxx()` does it work?
2. does it work if you `pip install gtsam-stubs` and restart vscode? (remember to uninstall it after you test it)

tianyilim:
Hello @ProfFan:

1.  I seem to get import errors for most things:
    image

    VS Code auto suggestions look as follows (I guess straight from the `__init__.py` file)
    image

    And `gtsam.noiseModel` shows up as a class, but with no members:
    image

2. Installing `gtsam-stubs` seems to have made most of the typing errors go away.

ProfFan:
Hi @tianyilim, could you try if #1821 works for you?

PR #1816, opened 2024-08-28
add example CameraResectioning.py

truher:


dellaert:
Please update the README file with @ProfFan 's help before running CI (Fan can approve and run once that's done). Thanks again! In particular, the readme file should be explicit and correct on the matrix F vs C.

truher:
cool, thanks.  actually i'm still poking at this, there seem to be cheirality exceptions for initial estimates that shouldn't produce them.   i'll ping back when i know what's going on.

python/gtsam/examples/CameraResectioning.py
```diff
@@ -0,0 +1,85 @@
...
+    p: Point2,
+    P: Point3,
+) -> NonlinearFactor:
+
+    def error_func(this: CustomFactor, v: Values, H: list[np.ndarray]) -> np.ndarray:
+        pose = v.atPose3(this.keys()[0])
+        camera = PinholeCameraCal3_S2(pose, calib)
+        if H is None:
+            return camera.project(P) - p
+        Dpose = np.zeros((2, 6), order="F")
```
ProfFan, line 26: This is correct, as these are passed as reference to C++. Please check `pybind11`'s documentation on this.

truher: Thanks!  This is what I'd expect given the way pybind seems to work.

The mystery is: what the CustomFactor is doing to allow either "C" or "F" storage for H[0]?

Oh also the cheirality issue I'm seeing doesn't have anything to do with this python code at all, it happens with the CameraResectioning.cpp too.  It seems to be that the optimizer is simply picking negative z's on its way to the solution.  Picking an initial Z of 5, for example, will trigger the exception as the optimizer happily slews around into negative territory.  Turning GTSAM_THROW_CHEIRALITY_EXCEPTION off fixes it; is there a downside to running that way, assuming there are enough points to avoid finding *solutions* on the wrong side of the camera?

ProfFan: It will copy if you give something with wrong storage order, simple and effective :)

The 2nd question is probably not related to Python though. Maybe just catch it in Python would work?

dellaert: > This is correct, as these are passed as reference to C++. Please check `pybind11`'s documentation on this.

Uhm, Fan, my ask to you is: you (as the author of this amazing capability) read the pybind11 docs and make sure the readme file has everything a casual user of GTSAM (with no knowledge of pybind necessary) can use it by following instructions :-). I think the readme currently does not mention the storage order, so this is what I'm asking you to fix (and check the examples) if you're willing.

truher: > The 2nd question is probably not related to Python though. Maybe just catch it in Python would work?

yeah, that's right, it's not a python issue, just something i noticed while working with the camera resectioning example.  you can't just catch it, because you need the optimizer to traverse the 'bad' territory on the way to the correct answer.  in this very simple case, the optimizer walks the "z" axis from the initial value of 5 to the correct value of 2 along this path: [5, -2.5, -2.49, -2.48, -2.35, -1.27, 2.46, 1.9, 1.99, 2.0].

dellaert: Try using LM with a high initial value for lambda

truher: thanks!  that works, i should have thought of it.  would you recommend, in general, the "high lambda" solution over the "don't throw chirality exception" solution?

ProfFan: Oh I mean since you are calling `PinholeCameraCal3_S2` in Python, it is possible that you can catch the exception here in your error function and do special processing.

> readme currently does not mention the storage order

I see. The gist is that whatever thing that got passed to C++ as a **reference to write into** has to be in the "F" format. This is because the buffer will be directly written by C++ code which assumes Eigen (column) layout. If a matrix in C++ got passed to Python for writing, then the ordering matters less, as Pybind11 will do automatic copy when you assign to the matrix with a numpy matrix with row-major layout. There is a small performance penalty though. It's best that "F" matrix is used at all times.

dellaert:
Ok, Fan, will you please “suggest” edits in the review to the readme and
examples in this PR to make sure people know?

On August 29, 2024, GitHub ***@***.***> wrote:
> @ProfFan commented on this pull request.
>

> In python/gtsam/examples/CameraResectioning.py
> <https://github.com/borglab/gtsam/pull/1816#discussion_r1736912583>:
>
> > + + +def resectioning_factor( + model: SharedNoiseModel, + key: int,
> + calib: Cal3_S2, + p: Point2, + P: Point3, +) -> NonlinearFactor: + +
> def error_func(this: CustomFactor, v: Values, H: list[np.ndarray]) ->
> np.ndarray: + pose = v.atPose3(this.keys()[0]) + camera =
> PinholeCameraCal3_S2(pose, calib) + if H is None: + return
> camera.project(P) - p + Dpose = np.zeros((2, 6), order="F")
> Oh I mean since you are calling PinholeCameraCal3_S2 in Python, it is
> possible that you can catch the exception here in your error function
> and do special processing.
>
> > readme currently does not mention the storage order
> >
>
> I see. The gist is that whatever thing that got passed to C++ as a
> reference to write into has to be in the "F" format. This is because
> the buffer will be directly written by C++ code which assumes Eigen
> (column) layout. If a matrix in C++ got passed to Python for writing,
> then the ordering matters less, as Pybind11 will do automatic copy
> when you assign to the matrix with a numpy matrix with row-major
> layout. There is a small performance penalty though. It's best that
> "F" matrix is used at all times.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/borglab/gtsam/pull/1816#discussion_r1736912583>, or
unsubscribe <https://github.com/notifications/unsubscribe-
auth/ACQHGSLKHCD7FANBBXZ7SNDZT5TNDAVCNFSM6AAAAABNIQ7ON2VHI2DSMVQWIX3LMV43YUDVNRWFEZLROVSXG5CSMV3GSZLXHMZDENRZHAYTANRQHE>.
You are receiving this because you commented.Message ID:
***@***.***>

truher:
I really appreciate the attention from you-all.  I've made a couple of python custom factors, and the whole thing seems to work great.

There is one thing I'm not sure how to do, though.  Since my landmarks are fixed, I'd like to pass None as Dpoint in
```
PinholeCamera::project(point, Dpose, Dpoint, Dcal)
```
I've poked around the gtwrap code to try to find an easy to way to add the pybind-recommended "none(true)" suffix to some of the arguments but I didn't find any way to add it to just *some* of them.

According to pybind, if you use std::optional, it should "just work" but OptionalJacobian is not that.

Any advice?  It's not that big a deal, obviously, I could reimplement the projection, but I imagine the need to pass None will come up again.

dellaert:
Try passing zero? Might not work.

ProfFan:
I think the limitation is due to `OptionalJacobian` being implemented as a null pointer (which is unfortunate due to C++'s removal of `std::optional<X&>`). For now I think what we do is add overloads in `*.i` that does not include these optional arguments, which you can just do yourself.

ProfFan:
For the readme, I would suggest adding a section that says:

```
`numpy` matrices that got passed to C++ as a reference to write into has to be in the "F" format. This is because the buffer will be directly written by C++ code which assumes Eigen (column) layout. Failure to maintain this would lead to an error.

If a matrix in C++ got passed to Python for writing, then the ordering matters less, as Pybind11 will do automatic copy when you assign to the (column-major) matrix with a numpy matrix with row-major layout. There is a small performance penalty since you will have an unnecessary copy. It's best that `order="F"` matrix is used at all times with GTSAM code.
```

ProfFan:
@truher Could you update the README file? Thanks :)

dellaert (approved):
I will fix the readme in another PR

Issue #1819, opened 2024-08-30
Ubuntu nightly packaging fails with cmake error, likely due to PR #1818


berndpfrommer:
The Ubuntu nightly builds (on Ubuntu 22.04) fail with the following error:
```
CMake Error at CMakeLists.txt:4 (cmake_policy):
  Policy "CMP0156" is not known to this version of CMake.


CMake Error at CMakeLists.txt:5 (cmake_policy):
  Policy "CMP0167" is not known to this version of CMake.
```
[Here](https://launchpadlibrarian.net/746574457/buildlog_ubuntu-jammy-amd64.gtsam_4.1.0-1ubuntu2~2182.gbp158895_BUILDING.txt.gz) is the log of the failed build.

This is very likely related to PR #1818.

Policy CMP0156 was introduced with CMake 3.29
Policy CMP0167 was introduced with CMake 3.30

I cannot figure out how to check if a policy exists before setting it. The pedestrian fix would probably look something like:
```
if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.29.0") 
  cmake_policy(SET CMP0156 NEW)
endif()

if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.30.0") 
  cmake_policy(SET CMP0167 OLD)
endif()
```

jlblancoc:
Neither Google nor GPT get easily to the right spot, but there's a [better solution](https://cmake.org/cmake/help/latest/command/if.html#existence-checks):

```cmake
if(POLICY CMP0156)
  cmake_policy(SET CMP0156 NEW)
endif()
# etc.
```

dellaert:
Apologies for the break. I can make a PR with whichever solution you prefer.

berndpfrommer:
Thanks!
I think the second fix by @jlblancoc is better.

ProfFan:
This probably can be closed?

jlblancoc:
Yes, it builds now on the PPA and the ROS build farms :+1:

PR #1820, opened 2024-08-30
Check if cmake policy exists

dellaert:
To fix issue #1819

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,16 @@
 cmake_minimum_required(VERSION 3.5)
-cmake_policy(SET CMP0082 NEW) # install from sub-directories immediately
+if (POLICY CMP0082)
+    cmake_policy(SET CMP0082 NEW) # install from sub-directories immediately
+endif()
+if (POLICY CMP0102)
 cmake_policy(SET CMP0102 NEW) # set policy on advanced variables and cmake cache
+endif()
+if (POLICY CMP0156)
 cmake_policy(SET CMP0156 NEW) # new linker strategies
+endif()
+if (POLICY CMP0167)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,8 +1,16 @@
 cmake_minimum_required(VERSION 3.5)
```
dellaert, line 1: Weirded out because how can we have these issues if minimum is 3.5 in the first place :-) ?

jlblancoc: Hi Frank! I'm here because I also saw gtsam [failing](https://build.ros2.org/job/Hdev__gtsam__ubuntu_jammy_amd64/141/console) on ROS build farms :-) 

The point is that, for example CMP0082 is old, but 156, etc. require [cmake >=3.29](https://cmake.org/cmake/help/latest/policy/CMP0156.html) (!) :-) 

It's cool having them, but with these protections keep working on old versions.

jlblancoc (approved):

CMakeLists.txt
berndpfrommer: We are still supporting Ubuntu 20.04, which has cmake 3.16. Since 3.16 > 3.5 this will satisfy cmake_minimum_required(VERSION 3.5), but (see issue #1819) cmake 3.16 does not yet know about the new policies since they were introduced with later versions, and hence it fails.
You would have to bump the version requirement to cmake 3.30 (!!)

berndpfrommer: Dang it, my internet is super flakey so this (from this morning) did not go through. You beat me to it @jlblancoc .

dellaert: Ah, you know, I was being dense: I read 3.5 as > 3.16, interpreting them as floats :-) thanks for merging!

PR #1828, opened 2024-09-08
Add the BATA translation averaging factor with unit tests

akshay-krishnan:
The existing translation averaging factor used the cost function from 1dsfm. 
Here we add the cost function from the [Bilinear Angle Based Translation Averaging](https://bbzh.github.io/document/bata.pdf) paper, which claims this cost be easier to optimize. 

We add a new translation factor for this cost: `BilinearAngleTranslationFactor`, and an option in the translation recovery class to use this factor instead of the default translation factor.

### DIFF
--- a/gtsam/sfm/TranslationFactor.h
+++ b/gtsam/sfm/TranslationFactor.h
@@ -18,6 +18,7 @@
  * @brief Binary factor for a relative translation direction measurement.
  */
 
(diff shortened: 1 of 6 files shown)


dellaert (approved):
Awesome!

dellaert:
I'll let you merge...

PR #1829, opened 2024-09-13
Hybrid tests using importance sampling

dellaert:
Replaced regression values with values taken from an importance sampler. They match!!!!

With some help from [ChatGPT o1-preview](https://x.com/fdellaert/status/1834354794573824483)!

### DIFF
--- a/gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
+++ b/gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
@@ -29,6 +29,7 @@
 #include <gtsam/hybrid/HybridValues.h>
 #include <gtsam/inference/Symbol.h>
 #include <gtsam/linear/GaussianFactorGraph.h>
+#include <gtsam/linear/VectorValues.h>
 #include <gtsam/nonlinear/PriorFactor.h>
 #include <gtsam/slam/BetweenFactor.h>
 
@@ -433,20 +434,20 @@ HybridBayesNet CreateBayesNet(
   return hbn;
 }
 
-/// Create importance sampling network p(x1| x0, m1) p(x0) P(m1),
-/// using Q(x0) = N(z0, sigma_Q) to sample from p(x0)
+/// Create importance sampling network q(x0,x1,m) = p(x1|x0,m1) q(x0) P(m1),
+/// using q(x0) = N(z0, sigma_Q) to sample x0.
 HybridBayesNet CreateProposalNet(
-    const GaussianMixture::shared_ptr& hybridMotionModel, double z0,
+    const GaussianMixture::shared_ptr& hybridMotionModel, const Vector1& z0,
     double sigma_Q) {
   HybridBayesNet hbn;
 
   // Add hybrid motion model
   hbn.push_back(hybridMotionModel);
 
-  // Add proposal Q(x0) for x0
+  // Add proposal q(x0) for x0
   auto measurement_model = noiseModel::Isotropic::Sigma(1, sigma_Q);
   hbn.emplace_shared<GaussianConditional>(
-      GaussianConditional::FromMeanAndStddev(X(0), Vector1(z0), sigma_Q));
+      GaussianConditional::FromMeanAndStddev(X(0), z0, sigma_Q));
 
   // Discrete uniform prior.
   hbn.emplace_shared<DiscreteConditional>(m1, "0.5/0.5");
@@ -466,7 +467,7 @@ void approximateDiscreteMarginal(const HybridBayesNet& hbn,
     HybridValues sample = proposalNet.sample(&rng);
     sample.insert(given);
     double weight = hbn.evaluate(sample) / proposalNet.evaluate(sample);
-    (sample.atDiscrete(m1.first) == 0) ? w0 += weight : w1 += weight;
+    (sample.atDiscrete(M(1)) == 0) ? w0 += weight : w1 += weight;
   }
   double sumWeights = w0 + w1;
   double pm1 = w1 / sumWeights;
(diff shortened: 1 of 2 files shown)


gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -387,50 +389,91 @@ TEST(GaussianMixtureFactor, GaussianMixtureModel2) {
...
+    bool add_second_measurement = false) {
+  HybridBayesNet hbn;
 
-  auto motion = std::make_shared<GaussianMixture>(
-      KeyVector{x1}, KeyVector{x0}, DiscreteKeys{m1}, std::vector{c0, c1});
-  hbn.push_back(motion);
+  // Add measurement model p(z0 | x0)
+  const double measurement_sigma = 3.0;
+  auto measurement_model = noiseModel::Isotropic::Sigma(1, measurement_sigma);
+  hbn.emplace_shared<GaussianConditional>(Z(0), Vector1(0.0), I_1x1, X(0),
```
dellaert: PS, @varunagrawal , signs on identity matrices were problematic. R should be I_1x1. logDeterminant gives NaN otherwise.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -387,50 +389,91 @@ TEST(GaussianMixtureFactor, GaussianMixtureModel2) {
...
-    // Add second measurement
-    auto p_z1 = std::make_shared<GaussianConditional>(
-        z1, Vector1(0.0), -I_1x1, x1, I_1x1, measurement_model);
-    hbn.push_back(p_z1);
+    hbn.emplace_shared<GaussianConditional>(Z(1), Vector1(0.0), I_1x1, X(1),
+                                            -I_1x1, measurement_model);
   }
 
+  // Add hybrid motion model
+  hbn.push_back(hybridMotionModel);
```
dellaert, line 431: Also, motion model needs to be added here otherwise sample (which depends on topological sort order) fails.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -387,103 +390,134 @@ TEST(GaussianMixtureFactor, GaussianMixtureModel2) {
...
+DiscreteKey m1(M(1), 2);
 
-  auto measurement_model = noiseModel::Isotropic::Sigma(1, measurement_sigma);
-  // Add measurement P(z0 | x0)
-  auto p_z0 = std::make_shared<GaussianConditional>(
-      z0, Vector1(0.0), -I_1x1, x0, I_1x1, measurement_model);
-  hbn.push_back(p_z0);
+void addMeasurement(HybridBayesNet& hbn, Key z_key, Key x_key, double sigma) {
+  auto measurement_model = noiseModel::Isotropic::Sigma(1, sigma);
+  hbn.emplace_shared<GaussianConditional>(z_key, Vector1(0.0), I_1x1, x_key,
```
varunagrawal, line 397: The measurement conditional is not the same as the original but it works in this case because `d=0`.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -387,103 +390,134 @@ TEST(GaussianMixtureFactor, GaussianMixtureModel2) {
...
  */
 TEST(GaussianMixtureFactor, TwoStateModel) {
   using namespace test_two_state_estimation;
 
   double mu0 = 1.0, mu1 = 3.0;
-  double sigma = 2.0;
-
-  DiscreteKey m1(M(1), 2);
-  Key z0 = Z(0), z1 = Z(1);
+  double sigma = 0.5;
```
varunagrawal, line 485: Any reason we changed the noise model value here?

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -504,87 +538,110 @@ TEST(GaussianMixtureFactor, TwoStateModel2) {
   using namespace test_two_state_estimation;
 
   double mu0 = 1.0, mu1 = 3.0;
-  double sigma0 = 6.0, sigma1 = 4.0;
-  auto model0 = noiseModel::Isotropic::Sigma(1, sigma0);
-  auto model1 = noiseModel::Isotropic::Sigma(1, sigma1);
-
-  DiscreteKey m1(M(1), 2);
-  Key z0 = Z(0), z1 = Z(1);
+  double sigma0 = 0.5, sigma1 = 2.0;
```
varunagrawal, line 541: Again not sure why these values were changed. I would expect that we would still keep sigma0 > sigma1.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
```diff
@@ -387,103 +390,134 @@ TEST(GaussianMixtureFactor, GaussianMixtureModel2) {
...
+std::pair<double, double> approximateDiscreteMarginal(
+    const HybridBayesNet& hbn,
+    const GaussianMixture::shared_ptr& hybridMotionModel,
+    const VectorValues& given, size_t N = 100000) {
+  /// Create importance sampling network q(x0,x1,m) = p(x1|x0,m1) q(x0) P(m1),
+  /// using q(x0) = N(z0, sigmaQ) to sample x0.
+  HybridBayesNet q;
+  q.push_back(hybridMotionModel);  // Add hybrid motion model
+  q.emplace_shared<GaussianConditional>(GaussianConditional::FromMeanAndStddev(
+      X(0), given.at(Z(0)), /* sigmaQ = */ 3.0));  // Add proposal q(x0) for x0
```
varunagrawal, line 449: Should `sigmaQ` be an argument as well?

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Wanted to make posterior a bit more skewed.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Same.

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Code for the present

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
dellaert: Tell me why, though? I might have forgotten why that was important...

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: Oh gotcha!

gtsam/hybrid/tests/testGaussianMixtureFactor.cpp
varunagrawal: The idea was to model the foot contact, with sigma0 being swing and sigma1 being stance hence a tighter noise model. If this test case passed, then the same principle would work for our contact estimation.

varunagrawal (approved):
Awesome sauce.

PR #1831, opened 2024-09-14
HybridNonlinearFactor Relinearization

varunagrawal:
`HybridNonlinearFactor` and `HybridGaussianFactor` now take a DecisionTree of factor and scalar pairs. This allows us to specify arbitrary values which will be used within the definition of the factors, allowing for relinearization.

### DIFF
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -56,6 +56,15 @@ HybridGaussianConditional::conditionals() const {
 }
 
 /* *******************************************************************************/
+HybridGaussianConditional::HybridGaussianConditional(
+    const KeyVector &continuousFrontals, const KeyVector &continuousParents,
+    const DiscreteKeys &discreteParents,
+    const std::vector<GaussianConditional::shared_ptr> &conditionals)
+    : HybridGaussianConditional(continuousFrontals, continuousParents,
+                                discreteParents,
+                                Conditionals(discreteParents, conditionals)) {}
+
+/* *******************************************************************************/
 // TODO(dellaert): This is copy/paste: HybridGaussianConditional should be
 // derived from HybridGaussianFactor, no?
 GaussianFactorGraphTree HybridGaussianConditional::add(
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -220,22 +220,19 @@ std::shared_ptr<HybridGaussianFactor> HybridGaussianConditional::likelihood(
   const DiscreteKeys discreteParentKeys = discreteKeys();
   const KeyVector continuousParentKeys = continuousParents();
   const HybridGaussianFactor::Factors likelihoods(
-      conditionals_, [&](const GaussianConditional::shared_ptr &conditional) {
-        const auto likelihood_m = conditional->likelihood(given);
+      conditionals_,
+      [&](const GaussianConditional::shared_ptr &conditional)
+          -> std::pair<GaussianFactor::shared_ptr, double> {
+        auto likelihood_m = conditional->likelihood(given);
         const double Cgm_Kgcm =
             logConstant_ - conditional->logNormalizationConstant();
         if (Cgm_Kgcm == 0.0) {
-          return likelihood_m;
+          return {likelihood_m, 0.0};
         } else {
           // Add a constant factor to the likelihood in case the noise models
           // are not all equal.
-          GaussianFactorGraph gfg;
-          gfg.push_back(likelihood_m);
-          Vector c(1);
-          c << std::sqrt(2.0 * Cgm_Kgcm);
-          auto constantFactor = std::make_shared<JacobianFactor>(c);
-          gfg.push_back(constantFactor);
-          return std::make_shared<JacobianFactor>(gfg);
+          double c = std::sqrt(2.0 * Cgm_Kgcm);
+          return {likelihood_m, c};
         }
       });
   return std::make_shared<HybridGaussianFactor>(
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -222,8 +222,8 @@ std::shared_ptr<HybridGaussianFactor> HybridGaussianConditional::likelihood(
   const HybridGaussianFactor::Factors likelihoods(
       conditionals_,
       [&](const GaussianConditional::shared_ptr &conditional)
-          -> std::pair<GaussianFactor::shared_ptr, double> {
-        auto likelihood_m = conditional->likelihood(given);
+          -> GaussianFactorValuePair {
+        const auto likelihood_m = conditional->likelihood(given);
         const double Cgm_Kgcm =
             logConstant_ - conditional->logNormalizationConstant();
         if (Cgm_Kgcm == 0.0) {
@@ -231,8 +231,13 @@ std::shared_ptr<HybridGaussianFactor> HybridGaussianConditional::likelihood(
         } else {
           // Add a constant factor to the likelihood in case the noise models
           // are not all equal.
-          double c = std::sqrt(2.0 * Cgm_Kgcm);
-          return {likelihood_m, c};
+          GaussianFactorGraph gfg;
+          gfg.push_back(likelihood_m);
+          Vector c(1);
+          c << std::sqrt(2.0 * Cgm_Kgcm);
+          auto constantFactor = std::make_shared<JacobianFactor>(c);
+          gfg.push_back(constantFactor);
+          return {std::make_shared<JacobianFactor>(gfg), 0.0};
         }
       });
   return std::make_shared<HybridGaussianFactor>(
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -58,11 +58,11 @@ HybridGaussianConditional::conditionals() const {
 /* *******************************************************************************/
 HybridGaussianConditional::HybridGaussianConditional(
     const KeyVector &continuousFrontals, const KeyVector &continuousParents,
-    const DiscreteKeys &discreteParents,
+    const DiscreteKey &discreteParent,
     const std::vector<GaussianConditional::shared_ptr> &conditionals)
     : HybridGaussianConditional(continuousFrontals, continuousParents,
-                                discreteParents,
-                                Conditionals(discreteParents, conditionals)) {}
+                                DiscreteKeys{discreteParent},
+                                Conditionals({discreteParent}, conditionals)) {}
 
 /* *******************************************************************************/
 // TODO(dellaert): This is copy/paste: HybridGaussianConditional should be
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -219,7 +219,7 @@ std::shared_ptr<HybridGaussianFactor> HybridGaussianConditional::likelihood(
 
   const DiscreteKeys discreteParentKeys = discreteKeys();
   const KeyVector continuousParentKeys = continuousParents();
-  const HybridGaussianFactor::Factors likelihoods(
+  const HybridGaussianFactor::FactorValuePairs likelihoods(
       conditionals_,
       [&](const GaussianConditional::shared_ptr &conditional)
           -> GaussianFactorValuePair {
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -229,15 +229,10 @@ std::shared_ptr<HybridGaussianFactor> HybridGaussianConditional::likelihood(
         if (Cgm_Kgcm == 0.0) {
           return {likelihood_m, 0.0};
         } else {
-          // Add a constant factor to the likelihood in case the noise models
+          // Add a constant to the likelihood in case the noise models
           // are not all equal.
-          GaussianFactorGraph gfg;
-          gfg.push_back(likelihood_m);
-          Vector c(1);
-          c << std::sqrt(2.0 * Cgm_Kgcm);
-          auto constantFactor = std::make_shared<JacobianFactor>(c);
-          gfg.push_back(constantFactor);
-          return {std::make_shared<JacobianFactor>(gfg), 0.0};
+          double c = 2.0 * Cgm_Kgcm;
+          return {likelihood_m, c};
         }
       });
   return std::make_shared<HybridGaussianFactor>(
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -56,24 +56,6 @@ HybridGaussianConditional::conditionals() const {
 }
 
 /* *******************************************************************************/
-HybridGaussianConditional::HybridGaussianConditional(
-    KeyVector &&continuousFrontals, KeyVector &&continuousParents,
-    DiscreteKeys &&discreteParents,
-    std::vector<GaussianConditional::shared_ptr> &&conditionals)
-    : HybridGaussianConditional(continuousFrontals, continuousParents,
-                                discreteParents,
-                                Conditionals(discreteParents, conditionals)) {}
-
-/* *******************************************************************************/
-HybridGaussianConditional::HybridGaussianConditional(
-    const KeyVector &continuousFrontals, const KeyVector &continuousParents,
-    const DiscreteKeys &discreteParents,
-    const std::vector<GaussianConditional::shared_ptr> &conditionals)
-    : HybridGaussianConditional(continuousFrontals, continuousParents,
-                                discreteParents,
-                                Conditionals(discreteParents, conditionals)) {}
-
-/* *******************************************************************************/
 // TODO(dellaert): This is copy/paste: HybridGaussianConditional should be
 // derived from HybridGaussianFactor, no?
 GaussianFactorGraphTree HybridGaussianConditional::add(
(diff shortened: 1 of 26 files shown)


varunagrawal:
@dellaert take a look while I finish adding some tests for `HybridNonlinearFactorGraph` to verify this works correctly. :)

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -235,7 +237,7 @@ std::shared_ptr<HybridGaussianFactor> HybridGaussianConditional::likelihood(
           c << std::sqrt(2.0 * Cgm_Kgcm);
           auto constantFactor = std::make_shared<JacobianFactor>(c);
           gfg.push_back(constantFactor);
-          return std::make_shared<JacobianFactor>(gfg);
+          return {std::make_shared<JacobianFactor>(gfg), 0.0};
```
dellaert: Ok, I don’t get it: if we now have the ability to add a constant, why don’t we? If you want to do the hiding, do it in HGF?

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -80,26 +85,26 @@ class GTSAM_EXPORT HybridGaussianFactor : public HybridFactor {
...
+                       const FactorValuePairs &factors);
 
   /**
    * @brief Construct a new HybridGaussianFactor object using a vector of
    * GaussianFactor shared pointers.
    *
    * @param continuousKeys Vector of keys for continuous factors.
    * @param discreteKeys Vector of discrete keys.
-   * @param factors Vector of gaussian factor shared pointers.
+   * @param factors Vector of gaussian factor shared pointers
```
dellaert, line 100: Can we kill this constructor? Error-prone. Simple change.

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -170,4 +175,16 @@ class GTSAM_EXPORT HybridGaussianFactor : public HybridFactor {
 template <>
 struct traits<HybridGaussianFactor> : public Testable<HybridGaussianFactor> {};
 
+/**
```
dellaert: Move to noiseModel. It might be there already

gtsam/hybrid/tests/Switching.h
```diff
@@ -57,12 +57,15 @@ inline HybridGaussianFactorGraph::shared_ptr makeSwitchingChain(
 
   // keyFunc(1) to keyFunc(n+1)
   for (size_t t = 1; t < n; t++) {
-    hfg.add(HybridGaussianFactor(
-        {keyFunc(t), keyFunc(t + 1)}, {{dKeyFunc(t), 2}},
+    std::vector<GaussianFactorValuePair> components = {
```
dellaert: Avoid making vectors, make trees immediately?

gtsam/hybrid/tests/Switching.h
```diff
@@ -159,9 +162,10 @@ struct Switching {
     for (size_t k = 0; k < K - 1; k++) {
       KeyVector keys = {X(k), X(k + 1)};
       auto motion_models = motionModels(k, between_sigma);
-      std::vector<NonlinearFactor::shared_ptr> components;
+      std::vector<std::pair<NonlinearFactor::shared_ptr, double>> components;
```
dellaert: No typedef?

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -53,9 +53,9 @@ class HybridNonlinearFactor : public HybridFactor {
 
   /**
    * @brief typedef for DecisionTree which has Keys as node labels and
-   * NonlinearFactor as leaf nodes.
+   * pairs of NonlinearFactor & an arbitrary scalar as leaf nodes.
    */
-  using Factors = DecisionTree<Key, sharedFactor>;
+  using Factors = DecisionTree<Key, std::pair<sharedFactor, double>>;
```
dellaert: Have typedef here as well?

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -90,25 +90,27 @@ class HybridNonlinearFactor : public HybridFactor {
...
    * @param normalized Flag indicating if the factor error is already
    * normalized.
    */
   template <typename FACTOR>
-  HybridNonlinearFactor(const KeyVector& keys, const DiscreteKeys& discreteKeys,
-                        const std::vector<std::shared_ptr<FACTOR>>& factors,
-                        bool normalized = false)
+  HybridNonlinearFactor(
+      const KeyVector& keys, const DiscreteKeys& discreteKeys,
+      const std::vector<std::pair<std::shared_ptr<FACTOR>, double>>& factors,
```
dellaert, line 104: Same: don’t have a vector constructor

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -90,25 +90,27 @@ class HybridNonlinearFactor : public HybridFactor {
...
     KeySet factor_keys_set;
-    for (auto&& f : factors) {
+    for (auto&& [f, val] : factors) {
       // Insert all factor continuous keys in the continuous keys set.
       std::copy(f->keys().begin(), f->keys().end(),
                 std::inserter(factor_keys_set, factor_keys_set.end()));
 
       if (auto nf = std::dynamic_pointer_cast<NonlinearFactor>(f)) {
-        nonlinear_factors.push_back(nf);
+        nonlinear_factors.push_back(std::make_pair(nf, val));
```
dellaert: Emplace_back, easy

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -133,9 +135,11 @@ class HybridNonlinearFactor : public HybridFactor {
    */
   AlgebraicDecisionTree<Key> errorTree(const Values& continuousValues) const {
     // functor to convert from sharedFactor to double error value.
-    auto errorFunc = [continuousValues](const sharedFactor& factor) {
-      return factor->error(continuousValues);
-    };
+    auto errorFunc =
+        [continuousValues](const std::pair<sharedFactor, double>& f) {
+          auto [factor, val] = f;
+          return factor->error(continuousValues) + (0.5 * val * val);
```
dellaert: Hmmm - worried here. We should agree on *one* definition for the double. Why are we squaring it here? The error is not squared! Potential bug.

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -387,7 +387,8 @@ TEST(HybridBayesNet, Sampling) {
       std::make_shared<BetweenFactor<double>>(X(0), X(1), 0, noise_model);
   auto one_motion =
       std::make_shared<BetweenFactor<double>>(X(0), X(1), 1, noise_model);
-  std::vector<NonlinearFactor::shared_ptr> factors = {zero_motion, one_motion};
+  std::vector<std::pair<NonlinearFactor::shared_ptr, double>> factors = {
```
dellaert: Use typedef, avoid vectors

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
```diff
@@ -613,13 +614,107 @@ TEST(HybridGaussianFactor, TwoStateModel2) {
   }
 }
 
+/* ************************************************************************* */
+/**
```
dellaert, line 624: Lowercase p for continuous densities

dellaert: Also, did we not already have this test??

dellaert (commented):
Made some comments regarding a vector-based constructor, which I think is problematic, *unless* this is a convenience constructor for a single mode? If there are multiple discrete keys, specifying a vector is always going to lead to bugs.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
varunagrawal: I thought we did but same means + different covariances does not exist.

gtsam/hybrid/HybridGaussianFactor.h
varunagrawal: `DecisionTree` has a vector based constructor and this follows that. Should we leave the correctness to the user? The assumption here is that they understand how the `DecisionTree` class creates the tree though.

gtsam/hybrid/HybridGaussianConditional.cpp
varunagrawal: I was going to do this in a possible next PR, but done now. :)

gtsam/hybrid/tests/testHybridBayesNet.cpp
varunagrawal: Something to note: The DecisionTree API takes a vector, so this constructor just forwards to the DecisionTree constructor.

Also, this allows us to reduce duplication. E.g. with directly specifying DecisionTrees, we have two specify discreteKeys in 2 places:
```cpp
DiscreteKeys discreteKeys{DiscreteKey(M(0), 2)};
  HybridNonlinearFactor::Factors factors(
      discreteKeys, {{zero_motion, 0.0}, {one_motion, 0.0}});
  nfg.emplace_shared<HybridNonlinearFactor>(KeyVector{X(0), X(1)}, discreteKeys,
                                            factors);
```

There is a longstanding TODO to obtain the continuous and discrete keys from the underlying factors and decision tree directly, but until we have that, this redundancy will continue to exist.

gtsam/hybrid/HybridGaussianFactor.h
dellaert: I prefer to kill it. Where is it currently used ?

gtsam/hybrid/HybridNonlinearFactor.h
dellaert: Examine this and respond?

gtsam/hybrid/HybridGaussianFactor.h
varunagrawal: The `DecisionTree` constructor from vectors? I believe in a bunch of places for discrete stuff.

gtsam/hybrid/tests/Switching.h
```diff
@@ -159,9 +162,10 @@ struct Switching {
...
       for (auto &&f : motion_models) {
-        components.push_back(std::dynamic_pointer_cast<NonlinearFactor>(f));
+        components.push_back(
+            {std::dynamic_pointer_cast<NonlinearFactor>(f), 0.0});
```
varunagrawal, line 169: @dellaert here's one reason to keep the vector based constructor. Allows us to define the tree incrementally like this.

gtsam/hybrid/tests/Switching.h
varunagrawal: Done

gtsam/hybrid/HybridNonlinearFactor.h
varunagrawal: Okay, I wrote down the math in detail and you are correct. It should be `0.5*val`. Updated!

gtsam/hybrid/HybridNonlinearFactor.h
varunagrawal: Updated/Removed all vector constructors in the Hybrid API. :)
HybridNonlinearFactor and HybridGaussianFactor only accept a single discrete key, while HybridGaussianConditional doesn't have a vector constructor at all.

gtsam/hybrid/HybridGaussianFactor.h
varunagrawal: Updated/Removed all vector constructors in the Hybrid API. :)
HybridNonlinearFactor and HybridGaussianFactor only accept a single discrete key, while HybridGaussianConditional doesn't have a vector constructor at all.

varunagrawal:
@dellaert based on our discussion, I have updated the API and tests.

gtsam/hybrid/HybridGaussianConditional.h
```diff
@@ -107,25 +107,13 @@ class GTSAM_EXPORT HybridGaussianConditional
                             const Conditionals &conditionals);
 
   /**
-   * @brief Make a Gaussian Mixture from a list of Gaussian conditionals
+   * @brief Make a Gaussian Mixture from a vector of Gaussian conditionals.
+   * The DecisionTree-based constructor is preferred over this one.
```
varunagrawal, line 111: Kept this constructor for the python wrapper, but added this note.

gtsam/hybrid/tests/TinyHybridExample.h
```diff
@@ -43,12 +43,13 @@ inline HybridBayesNet createHybridBayesNet(size_t num_measurements = 1,
...
+    std::vector<GaussianConditional::shared_ptr> conditionals{
+        GaussianConditional::sharedMeanAndStddev(Z(i), I_1x1, X(0), Z_1x1, 0.5),
+        GaussianConditional::sharedMeanAndStddev(Z(i), I_1x1, X(0), Z_1x1, 3)};
     bayesNet.emplace_shared<HybridGaussianConditional>(
         KeyVector{Z(i)}, KeyVector{X(0)}, DiscreteKeys{mode_i},
-        std::vector{GaussianConditional::sharedMeanAndStddev(Z(i), I_1x1, X(0),
-                                                             Z_1x1, 0.5),
-                    GaussianConditional::sharedMeanAndStddev(Z(i), I_1x1, X(0),
-                                                             Z_1x1, 3)});
+        HybridGaussianConditional::Conditionals(modes, conditionals));
```
dellaert: So, this is the case of a single key, so here you can use a vector, as we discussed. You would then specify a single DiscreteKey, not a set.

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -107,9 +107,11 @@ TEST(HybridBayesNet, evaluateHybrid) {
   // Create hybrid Bayes net.
   HybridBayesNet bayesNet;
   bayesNet.push_back(continuousConditional);
+  DiscreteKeys discreteParents{Asia};
   bayesNet.emplace_shared<HybridGaussianConditional>(
-      KeyVector{X(1)}, KeyVector{}, DiscreteKeys{Asia},
-      std::vector{conditional0, conditional1});
+      KeyVector{X(1)}, KeyVector{}, discreteParents,
+      HybridGaussianConditional::Conditionals(
```
dellaert: Same comment

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -168,9 +170,11 @@ TEST(HybridBayesNet, Error) {
              conditional1 = std::make_shared<GaussianConditional>(
                  X(1), Vector1::Constant(2), I_1x1, model1);
 
+  DiscreteKeys discreteParents{Asia};
   auto gm = std::make_shared<HybridGaussianConditional>(
-      KeyVector{X(1)}, KeyVector{}, DiscreteKeys{Asia},
-      std::vector{conditional0, conditional1});
+      KeyVector{X(1)}, KeyVector{}, discreteParents,
+      HybridGaussianConditional::Conditionals(
```
dellaert: Same comment

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -383,14 +387,17 @@ TEST(HybridBayesNet, Sampling) {
...
   auto one_motion =
       std::make_shared<BetweenFactor<double>>(X(0), X(1), 1, noise_model);
-  std::vector<NonlinearFactor::shared_ptr> factors = {zero_motion, one_motion};
-  nfg.emplace_shared<PriorFactor<double>>(X(0), 0.0, noise_model);
-  nfg.emplace_shared<HybridNonlinearFactor>(
-      KeyVector{X(0), X(1)}, DiscreteKeys{DiscreteKey(M(0), 2)}, factors);
+  DiscreteKeys discreteKeys{DiscreteKey(M(0), 2)};
+  HybridNonlinearFactor::Factors factors(
+      discreteKeys, {{zero_motion, 0.0}, {one_motion, 0.0}});
+  nfg.emplace_shared<HybridNonlinearFactor>(KeyVector{X(0), X(1)}, discreteKeys,
```
dellaert: And here as well…

dellaert (commented):
I think this looks good, but I would have the single discrete key constructor as we discussed?

gtsam/hybrid/tests/TinyHybridExample.h
varunagrawal: Ah sorry about that, I had already changed to the DecisionTree constructor before we discussed the single key option. Updated.

gtsam/hybrid/tests/testHybridBayesNet.cpp
varunagrawal: Updated.

gtsam/hybrid/tests/testHybridBayesNet.cpp
varunagrawal: Updated.

gtsam/hybrid/tests/testHybridBayesNet.cpp
varunagrawal: Updated.

gtsam/hybrid/tests/TinyHybridExample.h
varunagrawal: FYI this is for `HybridGaussianConditional` which can have multiple discrete parents. I've kept this for multiple discrete keys mainly for the wrapper.
Should I still go ahead and update this?

varunagrawal: I'm going ahead and doing this in the interest of time. If a use case comes up where we need multiple discrete parents + the vector of conditionals, we can re-evaluate then. :)

dellaert (approved):
LGTM!

PR #1832, opened 2024-09-15
Enum to categorize HybridFactor

varunagrawal:
Replaced the previous boolean based checks for whether a `HybridFactor` is `Discrete`, `Continuous` or `Hybrid` with an enum.
This would prevent a `HybridFactor` from having multiple flags true accidentally.

### DIFF
--- a/gtsam/hybrid/HybridFactor.h
+++ b/gtsam/hybrid/HybridFactor.h
@@ -42,7 +42,7 @@ DiscreteKeys CollectDiscreteKeys(const DiscreteKeys &key1,
                                  const DiscreteKeys &key2);
 
 /// Enum to help with categorizing hybrid factors.
-enum class HybridCategory { Discrete, Continuous, Hybrid };
+enum class HybridCategory { None, Discrete, Continuous, Hybrid };
 
 /**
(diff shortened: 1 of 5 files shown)


gtsam/hybrid/HybridFactor.h
```diff
@@ -41,6 +41,9 @@ KeyVector CollectKeys(const KeyVector &keys1, const KeyVector &keys2);
 DiscreteKeys CollectDiscreteKeys(const DiscreteKeys &key1,
                                  const DiscreteKeys &key2);
 
+/// Enum to help with categorizing hybrid factors.
+enum class HybridCategory { None, Discrete, Continuous, Hybrid };
```
dellaert: I like this a lot, but could we just make it an inner inside the HF class?

varunagrawal: Done. I renamed it to `Category` so it reads nicer as `HybridFactor::Category`.

dellaert (approved):
LGTM, why not target develop? Prefer smaller PRs into develop than having to see this code again in another review…

varunagrawal:
I was trying to avoid major merge conflicts. I guarantee you won't see this code in another review. :)

PR #1834, opened 2024-09-18
Remove HybridNonlinearFactor normalization code

varunagrawal:
Remove @keevindoherty's normalization code since we have updated the scheme to account for this in the conditionals and as extra scalars.


dellaert (approved):
Great

PR #1836, opened 2024-09-20
HybridGaussianConditional inherits from HybridGaussianFactor

varunagrawal:
Analogous to `GaussianConditional` and `GaussianFactor`, make `HybridGaussianConditional` inherit from `HybridGaussianFactor` so we can reduce code duplication.

I also changed `logConstant_` to be defined in the negative logspace, letting us lower bound the values at 0.0 instead of $-\infty$. I'll be updating the `Hybrid.pdf` file to reflect this for synergy.

One of a bunch of small PRs to improve the hybrid API and functionality.

### DIFF
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -51,14 +51,14 @@ HybridGaussianConditional::HybridGaussianConditional(
                  discreteParents, GetFactorValuePairs(conditionals)),
       BaseConditional(continuousFrontals.size()),
       conditionals_(conditionals) {
-  // Calculate logNormalizer_ as the minimum of the log normalizers of the
+  // Calculate logConstant_ as the minimum of the log normalizers of the
   // conditionals, by visiting the decision tree:
-  logNormalizer_ = std::numeric_limits<double>::infinity();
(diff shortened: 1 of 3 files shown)


dellaert (approved):
LG, but this makes me think that maybe *all* log constants should be in negative log space. Either k, or -log k. Never log k.

varunagrawal:
> LG, but this makes me think that maybe _all_ log constants should be in negative log space. Either k, or -log k. Never log k.

Yeah I have been thinking about that as well.

dellaert:
> > LG, but this makes me think that maybe _all_ log constants should be in negative log space. Either k, or -log k. Never log k.
> 
> Yeah I have been thinking about that as well.

I think we should do it. Will you or will I?

varunagrawal:
I can get to it later tonight, if that works?

varunagrawal:
Just to make sure I understand, the change is to update `logNormalizationConstant` so that it returns `-log(k)` where $k = \frac{1}{\sqrt{|2\pi \Sigma|}}$ right?
Currently that is the only method/function that works with `log(k)`.

dellaert:
Yes. But there is one (or more?) in HC class hierarchy and one in noiseModel

PR #1837, opened 2024-09-20
Common `errorTree` method and its use in HybridGaussianFactorGraph

varunagrawal:
Add `errorTree` as a virtual method in `HybridFactor`, and implement it in child classes.
This allows us to make the `HybridGaussianFactorGraph::errorTree` method very clean.

Also added the `HybridNonlinearFactorGraph::error(const HybridValues&)` method and corresponding test.

### DIFF
--- a/gtsam/hybrid/HybridConditional.cpp
+++ b/gtsam/hybrid/HybridConditional.cpp
@@ -130,6 +130,22 @@ double HybridConditional::error(const HybridValues &values) const {
 }
 
 /* ************************************************************************ */
+AlgebraicDecisionTree<Key> HybridConditional::errorTree(
+    const VectorValues &values) const {
+  if (auto gc = asGaussian()) {
+    return AlgebraicDecisionTree<Key>(gc->error(values));
+  }
+  if (auto gm = asHybrid()) {
+    return gm->errorTree(values);
+  }
+  if (auto dc = asDiscrete()) {
+    return AlgebraicDecisionTree<Key>(0.0);
+  }
+  throw std::runtime_error(
+      "HybridConditional::error: conditional type not handled");
+}
+
+/* ************************************************************************ */
 double HybridConditional::logProbability(const HybridValues &values) const {
   if (auto gc = asGaussian()) {
     return gc->logProbability(values.continuous());
--- a/gtsam/hybrid/HybridConditional.h
+++ b/gtsam/hybrid/HybridConditional.h
@@ -179,6 +179,16 @@ class GTSAM_EXPORT HybridConditional
   /// Return the error of the underlying conditional.
   double error(const HybridValues& values) const override;
 
+  /**
+   * @brief Compute error of the HybridConditional as a tree.
+   *
+   * @param continuousValues The continuous VectorValues.
+   * @return AlgebraicDecisionTree<Key> A decision tree with the same keys
+   * as the conditionals involved, and leaf values as the error.
+   */
+  virtual AlgebraicDecisionTree<Key> errorTree(
+      const VectorValues& values) const override;
+
   /// Return the log-probability (or density) of the underlying conditional.
   double logProbability(const HybridValues& values) const override;
 
--- a/gtsam/hybrid/HybridConditional.h
+++ b/gtsam/hybrid/HybridConditional.h
@@ -186,7 +186,7 @@ class GTSAM_EXPORT HybridConditional
    * @return AlgebraicDecisionTree<Key> A decision tree with the same keys
    * as the conditionals involved, and leaf values as the error.
    */
-  virtual AlgebraicDecisionTree<Key> errorTree(
+  AlgebraicDecisionTree<Key> errorTree(
       const VectorValues& values) const override;
 
   /// Return the log-probability (or density) of the underlying conditional.
--- a/gtsam/hybrid/HybridFactor.h
+++ b/gtsam/hybrid/HybridFactor.h
@@ -136,6 +136,10 @@ class GTSAM_EXPORT HybridFactor : public Factor {
   /// Return only the continuous keys for this factor.
   const KeyVector &continuousKeys() const { return continuousKeys_; }
 
+  /// Virtual class to compute tree of linear errors.
+  virtual AlgebraicDecisionTree<Key> errorTree(
+      const VectorValues &values) const = 0;
+
   /// @}
 
  private:
--- a/gtsam/hybrid/HybridGaussianConditional.h
+++ b/gtsam/hybrid/HybridGaussianConditional.h
@@ -175,43 +175,6 @@ class GTSAM_EXPORT HybridGaussianConditional
       const VectorValues &continuousValues) const;
 
   /**
-   * @brief Compute the error of this hybrid Gaussian conditional.
-   *
-   * This requires some care, as different components may have
-   * different normalization constants. Let's consider p(x|y,m), where m is
-   * discrete. We need the error to satisfy the invariant:
-   *
-   *    error(x;y,m) = K - log(probability(x;y,m))
-   *
-   * For all x,y,m. But note that K, the (log) normalization constant defined
-   * in Conditional.h, should not depend on x, y, or m, only on the parameters
-   * of the density. Hence, we delegate to the underlying Gaussian
-   * conditionals, indexed by m, which do satisfy:
-   *
-   *    log(probability_m(x;y)) = K_m - error_m(x;y)
-   *
-   * We resolve by having K == max(K_m) and
-   *
-   *    error(x;y,m) = error_m(x;y) + K - K_m
-   *
-   * which also makes error(x;y,m) >= 0 for all x,y,m.
-   *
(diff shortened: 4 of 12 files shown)


gtsam/hybrid/HybridFactor.h
```diff
@@ -136,6 +136,10 @@ class GTSAM_EXPORT HybridFactor : public Factor {
   /// Return only the continuous keys for this factor.
   const KeyVector &continuousKeys() const { return continuousKeys_; }
 
+  /// Virtual class to compute tree of linear errors.
```
dellaert, line 139: linear?

gtsam/hybrid/HybridGaussianConditional.h
```diff
@@ -109,9 +109,9 @@ class GTSAM_EXPORT HybridGaussianConditional
                             const Conditionals &conditionals);
 
   /**
-   * @brief Make a Hybrid Gaussian Conditional from a vector of Gaussian
-   * conditionals. The DecisionTree-based constructor is preferred over this
-   * one.
+   * @brief Make a Hybrid Gaussian Conditional from
```
dellaert, line 112: For a single discrete parent this constructor is fine: no way to do it wrong.

gtsam/hybrid/HybridGaussianConditional.h
```diff
@@ -208,8 +208,8 @@ class GTSAM_EXPORT HybridGaussianConditional
    * @return AlgebraicDecisionTree<Key> A decision tree on the discrete keys
    * only, with the leaf values as the error for each assignment.
    */
-  AlgebraicDecisionTree<Key> errorTree(
-      const VectorValues &continuousValues) const;
+  virtual AlgebraicDecisionTree<Key> errorTree(
```
dellaert: virtual is not needed when override. Also wondering why this method is overwritten in the first place? Should it not be identical to HGF version? If not, comment on how and why it is different.

gtsam/hybrid/HybridConditional.h
```diff
@@ -179,6 +179,16 @@ class GTSAM_EXPORT HybridConditional
...
   double error(const HybridValues& values) const override;
 
+  /**
+   * @brief Compute error of the HybridConditional as a tree.
+   *
+   * @param continuousValues The continuous VectorValues.
+   * @return AlgebraicDecisionTree<Key> A decision tree with the same keys
+   * as the conditionals involved, and leaf values as the error.
+   */
+  virtual AlgebraicDecisionTree<Key> errorTree(
```
dellaert: no virtual needed (lint error)

gtsam/hybrid/HybridConditional.cpp
```diff
@@ -129,6 +129,22 @@ double HybridConditional::error(const HybridValues &values) const {
       "HybridConditional::error: conditional type not handled");
 }
 
+/* ************************************************************************ */
+AlgebraicDecisionTree<Key> HybridConditional::errorTree(
```
dellaert, line 133: Why not in HybridFactor?

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -148,8 +148,8 @@ class GTSAM_EXPORT HybridGaussianFactor : public HybridFactor {
    * @return AlgebraicDecisionTree<Key> A decision tree with the same keys
    * as the factors involved, and leaf values as the error.
    */
-  AlgebraicDecisionTree<Key> errorTree(
-      const VectorValues &continuousValues) const;
+  virtual AlgebraicDecisionTree<Key> errorTree(
```
dellaert: no virtual

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -74,6 +74,13 @@ class GTSAM_EXPORT HybridNonlinearFactor : public HybridFactor {
   /// Decision tree of Gaussian factors indexed by discrete keys.
   Factors factors_;
 
+  /// HybridFactor method implementation. Should not be used.
```
dellaert, line 77: Hmmm. So then why does it exist in the base class. red flag. Should maybe only exist in Gaussian branch.

dellaert (commented):
Some comments. Not sure I am on board with location of these methods yet.

gtsam/hybrid/HybridNonlinearFactor.h
varunagrawal: This is primarily because of HybridConditional.

The class hierarchy is
- HybridFactor
  - HybridConditional
  - HybridGaussianFactor
      - HybridGaussianConditional

HybridConditional acts as a common container class for GaussianConditional, DiscreteConditional and HybridGaussianConditional. By defining errorTree in HybridFactor, I can just check for HybridFactor. Otherwise, I'll have to check for HybridConditional and HybridGaussianFactor separately which I felt defeated the purpose of this base class method.

dellaert: So, this is a different hierarchy than the non-hybrid case then:
```
class GTSAM_EXPORT GaussianConditional :
    public JacobianFactor,
    public Conditional<JacobianFactor, GaussianConditional>
```

gtsam/hybrid/HybridConditional.cpp
varunagrawal: `HybridFactor` doesn't have the `asGaussian`, `asHybrid` and `asDiscrete` methods.

gtsam/hybrid/HybridNonlinearFactor.h
varunagrawal: Sorry there was a rendering error in my previous comment.
It is the same hierarchy as GaussianConditional since HybridGaussianConditional inherits from HybridGaussianFactor, however HybridConditional complicates things (or at least makes it messier).

gtsam/hybrid/HybridGaussianConditional.h
varunagrawal: Yeah it is not needed. The changes I made when making HybridGaussianConditional inherit from HybridGaussianFactor were sufficient to accommodate this. :)

gtsam/hybrid/HybridFactor.h
varunagrawal: This linear because it only accepts `VectorValues` right? This is also why its definition in `HybridNonlinearFactor` throws a runtime error.

gtsam/hybrid/HybridNonlinearFactor.h
dellaert: What would break if you only have errorTree in HybridGaussianFactor? It would be available in HGC. All done?

varunagrawal: Yes, but then we'd need a separate check for HybridConditional (which is different from HybridGaussianConditional).

I think I'm okay with that. The current scheme just seemed tidier (except for HybridNonlinearFactor which isn't tidy).

dellaert: Why is a check needed? We would never call errorTree on HybridConditional, as it only makes sense for Gaussian variants, right?

varunagrawal: A HybridConditional can have a GaussianConditional or a HybridGaussianConditional as its underlying conditional. 

Remember that HybridConditional is a container based on type erasure and not a base class for HybridGaussianConditional.

dellaert: Right. But then why would we need an errorTree function for HybridConditional. If it can be either, then a method providing VectorValues does not make sense to me. Where would that be used? Is it used???

varunagrawal: It is used in an incremental sense, when we add the conditionals from a HybridBayesNet to the HybridGaussianFactorGraph, the factor graph now has `HybridGaussianFactor` (and potential `HybridGaussianConditional`s) and `HybridConditional`s as well.

The alternative could be to unwrap the conditionals from the `HybridConditional` before adding it to the graph.

I am open to suggestions about this since this was nontrivial for me. I think checking for `HybridGaussianFactor` and `HybridConditional` separately makes sense.

varunagrawal: Oh and another case: if we go from a `HybridBayesNet` to a `HybridGaussianFactorGraph`, then all the factors are added as `HybridConditional`s which are represented as `HybridFactor`s. This causes `DiscreteFactor`s to be considered in the errorTree computations as well. :(

dellaert (approved):
OK, thanks for engaging in a discussion. It feels a bit weird still but this is a consequence from how we did the class hierarch and how defined the hybrid factor graphs.

varunagrawal:
> OK, thanks for engaging in a discussion. It feels a bit weird still but this is a consequence from how we did the class hierarch and how defined the hybrid factor graphs.

Yeah I think we can come up with a better class inheritance hierarchy, which is transparent to the user, later once the API is stable. I have ideas about this!

PR #1839, opened 2024-09-21
Use `k` or `-log(k)` for normalization constant

varunagrawal:
Define the method `negLogConstant` to use the normalization constant $k = \frac{1}{\sqrt{2\pi \Sigma}}$ only in negative log space.


gtsam/linear/GaussianConditional.cpp
```diff
@@ -181,24 +181,24 @@ namespace gtsam {
...
+    // which gives neg-log = 0.5*n*log(2*pi) + 0.5*(-2.0 * logDeterminant())
+    //     = 0.5*n*log(2*pi) - (0.5*2.0 * logDeterminant())
+    //     = 0.5*n*log(2*pi) - logDeterminant()
+    return 0.5 * n * log2pi - logDeterminant();
   }
 
   /* ************************************************************************* */
-  //  density = k exp(-error(x))
-  //  log = log(k) - error(x)
+  //  density = 1/k exp(-error(x))
```
dellaert: No, this is k still

dellaert (commented):
Ok, this is sooo much better actually. HOWEVER, now I do believe the name `logNormalizationConstant` is now very confusing. Because -log is “error space”, we could rename to `errorConstant`. We could also do `negLogConstant`.

`errorConstant` has the advantage that it will read very naturally when creating the hybrid factors.

dellaert:
I assume you will target develop after other PR is merged…

dellaert:
So, `errorConstant` ?

varunagrawal:
I like `errorConstant` but does it make sense for `GaussianConditional` and other such cases? `negLogConstant` seems to fit better there?

gtsam/inference/Conditional.h
```diff
@@ -34,9 +34,10 @@ namespace gtsam {
    * `logProbability` is the main methods that need to be implemented in derived
    * classes. These two methods relate to the `error` method in the factor by:
    *   probability(x) = k exp(-error(x))
-   * where k is a normalization constant making \int probability(x) == 1.0, and
-   *   logProbability(x) = K - error(x)
-   * i.e., K = log(K). The normalization constant K is assumed to *not* depend
+   * where k is a normalization constant making
+   * \int probability(x) = \int k exp(-error(x)) == 1.0, and
+   * logProbability(x) = -(K + error(x))
+   * i.e., K = -log(k). The normalization constant k is assumed to *not* depend
```
varunagrawal, line 40: @dellaert this comment is why I think `negLogConstant` is also suitable. If you think `errorConstant` is better, then I'm good to go.

dellaert:
You’re right, `negLogConstant` might work better.

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -51,14 +51,14 @@ HybridGaussianConditional::HybridGaussianConditional(
...
+  // Calculate logConstant_ as the minimum of the negative-log normalizers of
+  // the conditionals, by visiting the decision tree:
   logConstant_ = std::numeric_limits<double>::infinity();
   conditionals_.visit(
       [this](const GaussianConditional::shared_ptr &conditional) {
         if (conditional) {
-          this->logConstant_ = std::min(
-              this->logConstant_, -conditional->logNormalizationConstant());
+          this->logConstant_ =
+              std::min(this->logConstant_, conditional->negLogConstant());
```
dellaert: Header says max but this is still min. If this is a bug and it should be max, there is a (non-regression) unit test missing somewhere.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -329,9 +329,9 @@ static std::shared_ptr<Factor> createDiscreteFactor(
 
     // Logspace version of:
     // exp(-factor->error(kEmpty)) / conditional->normalizationConstant();
-    // We take negative of the logNormalizationConstant `log(k)`
-    // to get `log(1/k) = log(\sqrt{|2πΣ|})`.
-    return -factor->error(kEmpty) - conditional->logNormalizationConstant();
+    // negLogConstant gives `-log(k)`
+    // which is `-log(k) = log(1/k) = log(\sqrt{|2πΣ|})`.
+    return -factor->error(kEmpty) + conditional->negLogConstant();
```
dellaert: Red flag. Errors are supposed to be in neg-log-space as well. Same comment about potential bug and missing unit test if not caught.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -355,8 +355,9 @@ static std::shared_ptr<Factor> createHybridGaussianFactor(
       auto hf = std::dynamic_pointer_cast<HessianFactor>(factor);
       if (!hf) throw std::runtime_error("Expected HessianFactor!");
       // Add 2.0 term since the constant term will be premultiplied by 0.5
-      // as per the Hessian definition
-      hf->constantTerm() += 2.0 * conditional->logNormalizationConstant();
+      // as per the Hessian definition,
+      // and negative since we want log(k)
```
dellaert, line 359: Hmmm, should this method be be negLogProbabilities as well?

gtsam/hybrid/hybrid.i
```diff
@@ -61,6 +61,7 @@ virtual class HybridConditional {
   size_t nrParents() const;
 
   // Standard interface:
+  double negLogConstant() const;
```
dellaert, line 64: Why do we still have logNormalizationConstant below?

gtsam/inference/Conditional-inst.h
```diff
@@ -59,10 +59,17 @@ double Conditional<FACTOR, DERIVEDCONDITIONAL>::evaluate(
...
-double Conditional<FACTOR, DERIVEDCONDITIONAL>::logNormalizationConstant()
+double Conditional<FACTOR, DERIVEDCONDITIONAL>::negLogConstant()
     const {
   throw std::runtime_error(
-      "Conditional::logNormalizationConstant is not implemented");
+      "Conditional::negLogConstant is not implemented");
+}
+
+/* ************************************************************************* */
+template <class FACTOR, class DERIVEDCONDITIONAL>
```
dellaert: Kill this - new philosophy: we are either in probability space, or in neg-log space. Nothing in between.

gtsam/linear/GaussianBayesNet.cpp
```diff
@@ -246,14 +246,16 @@ namespace gtsam {
   double GaussianBayesNet::logNormalizationConstant() const {
     /*
     normalization constant = 1.0 / sqrt((2*pi)^n*det(Sigma))
-    logConstant = -0.5 * n*log(2*pi) - 0.5 * log det(Sigma)
-    
-    log det(Sigma)) = -2.0 * logDeterminant()
+    logConstant = -log(normalizationConstant)
```
dellaert: this method should be negLogConstant as well.

gtsam/linear/NoiseModel.cpp
```diff
@@ -255,17 +255,21 @@ double Gaussian::logDeterminant() const {
...
   size_t n = dim();
   constexpr double log2pi = 1.8378770664093454835606594728112;
-  // Get 1/log(\sqrt(|2pi Sigma|)) = -0.5*log(|2pi Sigma|)
-  return -0.5 * n * log2pi + logDetR();
+  // Get -log(1/\sqrt(|2pi Sigma|)) = 0.5*log(|2pi Sigma|)
+  return 0.5 * n * log2pi - logDetR();
 }
 
+/* *******************************************************************************/
+double Gaussian::logNormalizationConstant() const {
```
dellaert: Kill

gtsam/linear/linear.i
```diff
@@ -548,6 +548,7 @@ virtual class GaussianConditional : gtsam::JacobianFactor {
   bool equals(const gtsam::GaussianConditional& cg, double tol) const;
   
   // Standard Interface
+  double negLogConstant() const;
```
dellaert, line 551: Kill log version

gtsam/linear/tests/testGaussianConditional.cpp
```diff
@@ -492,9 +492,10 @@ TEST(GaussianConditional, LogNormalizationConstant) {
   VectorValues x;
   x.insert(X(0), Vector3::Zero());
   Matrix3 Sigma = I_3x3 * sigma * sigma;
-  double expectedLogNormalizingConstant = log(1 / sqrt((2 * M_PI * Sigma).determinant()));
+  double expectedLogNormalizationConstant =
```
dellaert: Switch to negLog

dellaert (commented):
Some red flags to check. But biggest comments: let’s be opinionated and kill logNormalizationConstant everywhere. It’s only been in the API a second and not in any release, so why keep it? Simplify!

gtsam/hybrid/HybridGaussianConditional.cpp
varunagrawal: It should be min. I have the math written as well as a unit test somewhere in the codebase. :)

gtsam/hybrid/hybrid.i
varunagrawal: I didn't remove `logNormalizationConstant`, just updated it to call `-negLogConstant`. I'll remove `logNormalizationConstant` completely.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: This is fine, since this is the discrete continuous boundary. We want to negate and exponentiate the error to get the discrete factor values. A few lines below, we have the method `probabilitiesFromLogValues` which does the exponentiation.
Alternatively, `probabilitiesFromLogValues` should probably do the negation too.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: Done

gtsam/hybrid/HybridGaussianConditional.h
```diff
@@ -150,9 +150,15 @@ class GTSAM_EXPORT HybridGaussianConditional
...
-  double logNormalizationConstant() const override { return -logConstant_; }
+  /**
+   * @brief Return log normalization constant in negative log space.
+   *
+   * The log normalization constant is the min of the individual
+   * log-normalization constants.
+   *
+   * @return double
+   */
+  inline double negLogConstant() const override { return logConstant_; }
```
dellaert: Ok, almost done! but this should probably los be renamed to negLogConstant_ no?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -321,23 +321,23 @@ using Result = std::pair<std::shared_ptr<GaussianConditional>,
...
+    return factor->error(kEmpty) - conditional->negLogConstant();
   };
 
-  AlgebraicDecisionTree<Key> logProbabilities(
-      DecisionTree<Key, double>(eliminationResults, logProbability));
+  AlgebraicDecisionTree<Key> negLogProbabilities(
+      DecisionTree<Key, double>(eliminationResults, negLogProbability));
   AlgebraicDecisionTree<Key> probabilities =
-      probabilitiesFromLogValues(logProbabilities);
+      probabilitiesFromLogValues(negLogProbabilities);
```
dellaert: This should probably also be renamed to fromNegLog or something

gtsam/inference/Conditional-inst.h
```diff
@@ -59,16 +59,14 @@ double Conditional<FACTOR, DERIVEDCONDITIONAL>::evaluate(
...
-      "Conditional::logNormalizationConstant is not implemented");
+double Conditional<FACTOR, DERIVEDCONDITIONAL>::negLogConstant() const {
+  throw std::runtime_error("Conditional::negLogConstant is not implemented");
 }
 
 /* ************************************************************************* */
 template <class FACTOR, class DERIVEDCONDITIONAL>
 double Conditional<FACTOR, DERIVEDCONDITIONAL>::normalizationConstant() const {
-  return std::exp(logNormalizationConstant());
+  return std::exp(-negLogConstant());
```
dellaert: Remove this template as well?

dellaert (approved):
A few stragglers still but I’ll trust you to fix and merge after :-)

gtsam/hybrid/HybridGaussianConditional.h
varunagrawal: Yes I agree. Sorry, I need to be more thorough.

gtsam/inference/Conditional-inst.h
varunagrawal: Gotcha. Wasn't sure if we wanted to keep this for easy access / syntactic sugar.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: Done!

PR #1840, opened 2024-09-21
Document and favor easy constructor

dellaert:
- added a new HybridGaussianFactor constructor for the common single key - all zero scalars case and put it first.
- add unit test for all three
- re-ordered constructors and provide more docs
- use easy constructor where possible
- eliminated some copy-paste in tests

### DIFF
--- a/gtsam/hybrid/HybridGaussianFactor.h
+++ b/gtsam/hybrid/HybridGaussianFactor.h
@@ -89,31 +89,48 @@ class GTSAM_EXPORT HybridGaussianFactor : public HybridFactor {
   HybridGaussianFactor() = default;
 
   /**
-   * @brief Construct a new hybrid Gaussian factor.
-   *
-   * @param continuousKeys A vector of keys representing continuous variables.
-   * @param discreteKeys A vector of keys representing discrete variables and
-   * their cardinalities.
-   * @param factors The decision tree of Gaussian factors and arbitrary scalars.
-   */
-  HybridGaussianFactor(const KeyVector &continuousKeys,
-                       const DiscreteKeys &discreteKeys,
-                       const FactorValuePairs &factors);
-
-  /**
-   * @brief Construct a new HybridGaussianFactor object using a vector of
-   * GaussianFactor shared pointers.
+   * @brief Construct a new HybridGaussianFactor on a single discrete key,
+   * providing the factors for each mode m as a vector of factors ϕ_m(x).
+   * The value ϕ(x,m) for the factor is simply ϕ_m(x).
    *
    * @param continuousKeys Vector of keys for continuous factors.
-   * @param discreteKey The discrete key to index each component.
-   * @param factors Vector of gaussian factor shared pointers
-   *  and arbitrary scalars. Same size as the cardinality of discreteKey.
+   * @param discreteKey The discrete key for the "mode", indexing components.
+   * @param factors Vector of gaussian factors, one for each mode.
+   */
+  HybridGaussianFactor(const KeyVector &continuousKeys,
+                       const DiscreteKey &discreteKey,
+                       const std::vector<GaussianFactor::shared_ptr> &factors)
+      : Base(continuousKeys, {discreteKey}), factors_({discreteKey}, factors) {}
+
+  /**
+   * @brief Construct a new HybridGaussianFactor on a single discrete key,
+   * including a scalar error value for each mode m. The factors and scalars are
+   * provided as a vector of pairs (ϕ_m(x), E_m).
+   * The value ϕ(x,m) for the factor is now ϕ_m(x) + E_m.
+   *
+   * @param continuousKeys Vector of keys for continuous factors.
+   * @param discreteKey The discrete key for the "mode", indexing components.
+   * @param factors Vector of gaussian factor-scalar pairs, one per mode.
    */
   HybridGaussianFactor(const KeyVector &continuousKeys,
                        const DiscreteKey &discreteKey,
                        const std::vector<GaussianFactorValuePair> &factors)
       : HybridGaussianFactor(continuousKeys, {discreteKey},
                              FactorValuePairs({discreteKey}, factors)) {}
+
+  /**
+   * @brief Construct a new HybridGaussianFactor on a several discrete keys M,
+   * including a scalar error value for each assignment m. The factors and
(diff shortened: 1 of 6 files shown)


gtsam/hybrid/tests/Switching.h
```diff
@@ -44,33 +47,28 @@ using symbol_shorthand::X;
...
- * @param keyFunc The functional to help specify the continuous key.
- * @param dKeyFunc The functional to help specify the discrete key.
+ * @param x The functional to help specify the continuous key.
+ * @param m The functional to help specify the discrete key.
  * @return HybridGaussianFactorGraph::shared_ptr
  */
 inline HybridGaussianFactorGraph::shared_ptr makeSwitchingChain(
-    size_t n, std::function<Key(int)> keyFunc = X,
-    std::function<Key(int)> dKeyFunc = M) {
+    size_t n, std::function<Key(int)> x = X, std::function<Key(int)> m = M) {
```
varunagrawal, line 55: Should we just kill x and m and use X and M directly?

gtsam/hybrid/tests/Switching.h
```diff
@@ -44,33 +47,28 @@ using symbol_shorthand::X;
...
-                                                  keyFunc(t + 1), I_3x3, Z_3x1),
-                 0.0},
-                {std::make_shared<JacobianFactor>(
-                     keyFunc(t), I_3x3, keyFunc(t + 1), I_3x3, Vector3::Ones()),
-                 0.0}});
-    hfg.add(
-        HybridGaussianFactor({keyFunc(t), keyFunc(t + 1)}, dKeys, components));
+    DiscreteKeys dKeys{{m(t), 2}};
+    std::vector<GaussianFactor::shared_ptr> components;
+    components.emplace_back(
```
varunagrawal, line 64: If you're doing an `emplace_back`, you could just do
```cpp
components.emplace_back<JacobianFactor>(x(t), I_3x3, x(t + 1), I_3x3, Z_3x1);
```

It should prevent an extra allocation from the raw pointer to the vector allocator.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
```diff
@@ -54,32 +54,49 @@ TEST(HybridGaussianFactor, Constructor) {
   CHECK(it == factor.end());
 }
 
+/* ************************************************************************* */
+namespace testA {
```
varunagrawal: I would prefer this namespace being called `test_constructor` or `test_constructor_fixture` so it is obvious what it is being used for.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
```diff
@@ -54,32 +54,49 @@ TEST(HybridGaussianFactor, Constructor) {
...
-  auto f10 = std::make_shared<JacobianFactor>(X(1), A1, X(2), A2, b);
-  auto f11 = std::make_shared<JacobianFactor>(X(1), A1, X(2), A2, b);
   auto f20 = std::make_shared<JacobianFactor>(X(1), A1, X(3), A3, b);
   auto f21 = std::make_shared<JacobianFactor>(X(1), A1, X(3), A3, b);
   auto f22 = std::make_shared<JacobianFactor>(X(1), A1, X(3), A3, b);
-  std::vector<GaussianFactorValuePair> factorsA{{f10, 0.0}, {f11, 0.0}};
-  std::vector<GaussianFactorValuePair> factorsB{
-      {f20, 0.0}, {f21, 0.0}, {f22, 0.0}};
 
   // TODO(Frank): why specify keys at all? And: keys in factor should be *all*
```
varunagrawal, line 95: We should talk about this TODO. I guess the compexity here is getting the discrete keys from the DecisionTree since there is no way to do it currently.

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -89,32 +89,49 @@ class GTSAM_EXPORT HybridGaussianFactor : public HybridFactor {
   HybridGaussianFactor() = default;
 
   /**
-   * @brief Construct a new hybrid Gaussian factor.
+   * @brief Construct a new HybridGaussianFactor on a single discrete key,
```
varunagrawal, line 92: I'm wondering if this should be after the `GaussianFactorValuePair` constructor and specify that the extra error `E_m` is 0.0 in this case?

varunagrawal (approved):
Some comments but otherwise fine. Will not hold up this PR. :)

gtsam/hybrid/HybridGaussianFactor.h
dellaert: My PR is about doing the opposite of that :-) Go from easy to harder, guiding the user. In fact, the more complex constructors are only used once in the entire code (before I added unit tests).

gtsam/hybrid/tests/Switching.h
dellaert: Sure, done

gtsam/hybrid/tests/Switching.h
dellaert: Hmmm, I don't understand this. If yousee how to make that work please do so in a follow-up PR.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
dellaert: Done

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
dellaert: Right, let's chat

gtsam/hybrid/tests/Switching.h
dellaert: Turns out this functionality was actually used in testHGFG

gtsam/hybrid/HybridGaussianFactor.h
varunagrawal: Okay I see that. Awesome.

varunagrawal:
Merging!

gtsam/hybrid/tests/Switching.h
varunagrawal: Okay I was slightly mistaken: the correct syntax would be `components.emplace_back(make_shared<JacobianFactor>(...));` so not achieving a lot.

PR #1842, opened 2024-09-23
Non-loop versions for checking sigma values

varunagrawal:
While updating the `NoiseModel` classes, I noticed these checks which can be massively improved using Eigen's checking mechanisms.
This should also allow for better vectorization of the code.

Moreover, I removed the `goto`s to make the code cleaner.

### DIFF
--- a/gtsam/linear/NoiseModel.cpp
+++ b/gtsam/linear/NoiseModel.cpp
@@ -282,15 +282,10 @@ Diagonal::Diagonal(const Vector& sigmas)
 
 /* ************************************************************************* */
 Diagonal::shared_ptr Diagonal::Variances(const Vector& variances, bool smart) {
-  if (smart) {
-    // check whether all the same entry
-    if ((variances.array() == variances(0)).all()) {
-      return Isotropic::Variance(variances.size(), variances(0), true);
-    } else
-      goto full;
-  }
-full:
-  return shared_ptr(new Diagonal(variances.cwiseSqrt()));
+  // check whether all the same entry
+  return (smart && (variances.array() == variances(0)).all())
+             ? Isotropic::Variance(variances.size(), variances(0), true)
+             : shared_ptr(new Diagonal(variances.cwiseSqrt()));
 }
 
 /* ************************************************************************* */
--- a/gtsam/linear/NoiseModel.cpp
+++ b/gtsam/linear/NoiseModel.cpp
@@ -286,10 +286,10 @@ Diagonal::shared_ptr Diagonal::Variances(const Vector& variances, bool smart) {
     // check whether all the same entry
     if ((variances.array() == variances(0)).all()) {
       return Isotropic::Variance(variances.size(), variances(0), true);
-    } else {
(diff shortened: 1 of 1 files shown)


gtsam/linear/NoiseModel.cpp
```diff
@@ -270,10 +270,7 @@ double Gaussian::logNormalizationConstant() const {
 /* ************************************************************************* */
 // Diagonal
 /* ************************************************************************* */
-Diagonal::Diagonal() :
-    Gaussian(1) // TODO: Frank asks: really sure about this?
-{
-}
+Diagonal::Diagonal() : Gaussian() {}
```
varunagrawal, line 273: The default arguments for the `Gaussian` noise model are
```cpp
Gaussian(size_t dim = 1,
               const std::optional<Matrix>& sqrt_information = {})
```
hence no need to pass any arguments.

gtsam/linear/NoiseModel.cpp
```diff
@@ -287,36 +284,41 @@ Diagonal::Diagonal(const Vector& sigmas)
 Diagonal::shared_ptr Diagonal::Variances(const Vector& variances, bool smart) {
   if (smart) {
     // check whether all the same entry
-    size_t n = variances.size();
-    for (size_t j = 1; j < n; j++)
-      if (variances(j) != variances(0)) goto full;
-    return Isotropic::Variance(n, variances(0), true);
+    if ((variances.array() == variances(0)).all()) {
```
dellaert: goto no longer make sense here (even if does elsewhere). Replace with 
```
return (smart && ( == )) ? Isotropic::Variance(...) : std::make_shared<Diagonal>(...);
```

Also, while at it, use `std::make_shared` wherever we can in this file.

gtsam/linear/NoiseModel.cpp
```diff
@@ -287,36 +284,41 @@ Diagonal::Diagonal(const Vector& sigmas)
...
-    for (size_t j = 1; j < n; j++)
-      if (sigmas(j) != sigmas(0)) goto full;
-    return Isotropic::Sigma(n, sigmas(0), true);
+    if ((sigmas.array() == sigmas(0)).all()) {
+      return Isotropic::Sigma(n, sigmas(0), true);
+    }
   }
-  full: return Diagonal::shared_ptr(new Diagonal(sigmas));
+full:
+  return Diagonal::shared_ptr(new Diagonal(sigmas));
```
dellaert, line 308: make_shared

dellaert (approved):
Approved modulo comments

gtsam/linear/NoiseModel.cpp
varunagrawal: We can't use `make_shared` here since the constructor is protected. :(

gtsam/linear/NoiseModel.cpp
varunagrawal: Same as above. The constructor is not public so we can't use `make_shared`.

PR #1843, opened 2024-09-23
Easy HNF constructors

dellaert:
- I made the HybridNonlinearFactor constructors the same as the earlier PR on HGF.
- I prefer type-checked vectors of shared pointers rather than run-time checks
- that allowed removing templated version and sharing code ebtween two single-key constructors

PR #1844, opened 2024-09-24
Small runtime improvement for hybrid

dellaert:
[Targeting branch from #1843, develop after that's merged]

Using instruments I was profiling testHybridEstimation, and I discovered a small gain (a few %) by making the DecisionTreeFactor::combine a bit more efficient.

However, @varunagrawal, I also noticed that `augment` in HGF is one of the heaviest hitters - probably 10-20%. This is not surprising, as this where the "hiding" happens, which is expensive and not needed. The negLogConstants can just be *added* during elimination,and we probably gain in the QR calls as well. The hiding has no benefit except that we would not have to rewrite a tiny part of the math :-/

### DIFF
--- a/gtsam/discrete/DecisionTree-inl.h
+++ b/gtsam/discrete/DecisionTree-inl.h
@@ -91,7 +91,7 @@ namespace gtsam {
     void dot(std::ostream& os, const LabelFormatter& labelFormatter,
              const ValueFormatter& valueFormatter,
              bool showZero) const override {
-      std::string value = valueFormatter(constant_);
+      const std::string value = valueFormatter(constant_);
       if (showZero || value.compare("0"))
         os << "\"" << this->id() << "\" [label=\"" << value
            << "\", shape=box, rank=sink, height=0.35, fixedsize=true]\n";
@@ -306,7 +306,8 @@ namespace gtsam {
     void dot(std::ostream& os, const LabelFormatter& labelFormatter,
              const ValueFormatter& valueFormatter,
              bool showZero) const override {
-      os << "\"" << this->id() << "\" [shape=circle, label=\"" << label_
+      const std::string label = labelFormatter(label_);
+      os << "\"" << this->id() << "\" [shape=circle, label=\"" << label
           << "\"]\n";
       size_t B = branches_.size();
       for (size_t i = 0; i < B; i++) {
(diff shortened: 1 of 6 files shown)


dellaert:
PS those test write nice pdf files if graphviz is installed. I had to fix something in DecisionTree, though:
image

varunagrawal:
Interesting. I was under the impression that "hiding" would be beneficial since then it's all Matrix-Vector code which Eigen can optimize really well. Guess not...

gtsam/discrete/DecisionTree-inl.h
```diff
@@ -306,7 +306,8 @@ namespace gtsam {
     void dot(std::ostream& os, const LabelFormatter& labelFormatter,
              const ValueFormatter& valueFormatter,
              bool showZero) const override {
-      os << "\"" << this->id() << "\" [shape=circle, label=\"" << label_
+      const std::string label = labelFormatter(label_);
```
varunagrawal, line 309: I've actually contemplated adding very thin wrapper over `AlgebraicDecisionTree<Key>` so we can have sensible defaults.

E.g. `dt.print()` instead of `dt.print("", DefaultKeyFormatter)` for one.

Do you think it's a good idea? I can do it later during the week.

gtsam/discrete/DecisionTreeFactor.cpp
```diff
@@ -179,24 +179,22 @@ namespace gtsam {
...
-    // Get the difference of the frontalKeys and the factor keys using set_difference
-    std::sort(keys_.begin(), keys_.end());
-    std::sort(frontalKeys_.begin(), frontalKeys_.end());
-    std::set_difference(keys_.begin(), keys_.end(), frontalKeys_.begin(),
-                        frontalKeys_.end(), back_inserter(difference));
-
     DiscreteKeys dkeys;
-    for (Key key : difference) {
-      dkeys.push_back(DiscreteKey(key, cardinality(key)));
+    for (auto&& [key, cardinality] : cardinalities_) {
```
varunagrawal, line 194: Okay this is so much better.

varunagrawal (approved):

gtsam/discrete/DecisionTree-inl.h
dellaert: Let's focus on estimation instead...

dellaert: Defaults are already in place...

PR #1847, opened 2024-09-26
Moved mixture model tests to their own file...

dellaert:
...and added a direct FG workflow. Also cleaned up copy/pasta in tests.


varunagrawal (approved):
So glad we figured out the direct FG specification.

PR #1848, opened 2024-09-26
Simpler HybridGaussianFoo constructors

dellaert:
- Remove superfluous  arguments
- Add helper structs and private constructor to ensure proper initialization
- Added input checking on factors and conditional keys but removed it since
  - expensive
  - it was not checked before
 - simplified all call sites -> no code is the best code

### DIFF
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -61,17 +61,22 @@ HybridGaussianConditional::HybridGaussianConditional(
     }
     return {std::dynamic_pointer_cast<GaussianFactor>(c), value};
   };
-  BaseFactor::factors_ = HybridGaussianFactor::Factors(conditionals, func);
+  HybridGaussianFactor::FactorValuePairs pairs(conditionals, func);
 
-  // Initialize base classes
-  KeyVector continuousKeys = frontals;
-  continuousKeys.insert(continuousKeys.end(), parents.begin(), parents.end());
-  BaseFactor::keys_ = continuousKeys;
-  BaseFactor::discreteKeys_ = discreteParents;
+  // Adjust frontals size
   BaseConditional::nrFrontals_ = frontals.size();
 
-  // Assign conditionals
-  conditionals_ = conditionals;  // TODO(frank): a duplicate of factors_ !!!
+  // Initialize HybridFactor
+  HybridFactor::category_ = HybridFactor::Category::Hybrid;
+  HybridFactor::discreteKeys_ = discreteParents;
+  HybridFactor::keys_ = frontals;
+  keys_.insert(keys_.end(), parents.begin(), parents.end());
+
+  // Initialize BaseFactor
+  BaseFactor::factors_ = BaseFactor::augment(pairs);  // TODO(frank): expensive
+
+  // Assign local conditionals. TODO(frank): a duplicate of factors_ !!!
+  conditionals_ = conditionals;
 }
 
 /* *******************************************************************************/
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -52,10 +52,6 @@ HybridGaussianConditional::ConstructorHelper::ConstructorHelper(
     return {std::dynamic_pointer_cast<GaussianFactor>(c), value};
   };
   pairs = HybridGaussianFactor::FactorValuePairs(conditionals, func);
-
-  // Build continuousKeys
-  continuousKeys = frontals;
-  continuousKeys.insert(continuousKeys.end(), parents.begin(), parents.end());
 }
 
 /* *******************************************************************************/
@@ -222,8 +218,8 @@ std::shared_ptr<HybridGaussianFactor> HybridGaussianConditional::likelihood(
           return {likelihood_m, Cgm_Kgcm};
         }
       });
-  return std::make_shared<HybridGaussianFactor>(
-      continuousParentKeys, discreteParentKeys, likelihoods);
+  return std::make_shared<HybridGaussianFactor>(discreteParentKeys,
+                                                likelihoods);
 }
 
 /* ************************************************************************* */
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -37,14 +37,9 @@ HybridGaussianConditional::ConstructorHelper::ConstructorHelper(
       [&](const GaussianConditional::shared_ptr &c) -> GaussianFactorValuePair {
     double value = 0.0;
     if (c) {
-      KeyVector cf(c->frontals().begin(), c->frontals().end());
-      KeyVector cp(c->parents().begin(), c->parents().end());
       if (frontals.empty()) {
-        frontals = cf;
-        parents = cp;
-      } else if (cf != frontals || cp != parents) {
-        throw std::invalid_argument(
-            "All conditionals must have the same frontals and parents");
+        frontals = KeyVector(c->frontals().begin(), c->frontals().end());
+        parents = KeyVector(c->parents().begin(), c->parents().end());
       }
       value = c->negLogConstant();
       negLogConstant = std::min(negLogConstant, value);
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -28,55 +28,64 @@
 #include <gtsam/linear/GaussianFactorGraph.h>
 
 namespace gtsam {
-HybridGaussianFactor::FactorValuePairs GetFactorValuePairs(
-    const HybridGaussianConditional::Conditionals &conditionals) {
-  auto func = [](const GaussianConditional::shared_ptr &conditional)
-      -> GaussianFactorValuePair {
-    double value = 0.0;
-    // Check if conditional is pruned
-    if (conditional) {
-      // Assign log(\sqrt(|2πΣ|)) = -log(1 / sqrt(|2πΣ|))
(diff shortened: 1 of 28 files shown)


dellaert:
Updating to also fix HybridNonlinearFactor

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -28,56 +28,47 @@
...
+
+  auto func =
+      [&](const GaussianConditional::shared_ptr &c) -> GaussianFactorValuePair {
     double value = 0.0;
-    // Check if conditional is pruned
-    if (conditional) {
-      // Assign log(\sqrt(|2πΣ|)) = -log(1 / sqrt(|2πΣ|))
-      value = conditional->negLogConstant();
+    if (c) {
+      if (frontals.empty()) {
```
varunagrawal: Is this needed? Won't a conditional always have a frontal variable?

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -73,22 +66,65 @@ static HybridGaussianFactor::Factors augment(
...
 /* *******************************************************************************/
-HybridGaussianFactor::HybridGaussianFactor(const KeyVector &continuousKeys,
-                                           const DiscreteKeys &discreteKeys,
-                                           const FactorValuePairs &factors)
-    : Base(continuousKeys, discreteKeys), factors_(augment(factors)) {}
+HybridGaussianFactor::ConstructorHelper::ConstructorHelper(
+    const DiscreteKey &discreteKey,
+    const std::vector<GaussianFactor::shared_ptr> &factors)
+    : discreteKeys({discreteKey}) {
+  // Extract continuous keys from the first non-null factor
```
varunagrawal: I'll do a full review in a bit, but I am wondering about the extra compute happening here when we could have just given the keys directly.

Currently I am struggling with the hybrid state estimator being far too slow and having too many hypotheses. Pruning is working but the blow up is still immense, hence my concerns here.

varunagrawal (commented):
Partial review. Will do more later tonight.

gtsam/hybrid/HybridGaussianFactor.cpp
dellaert: Ironically, it's a bit faster.

gtsam/hybrid/HybridGaussianConditional.cpp
dellaert: No use copying frontals if they've already been copied.

dellaert:
> Partial review. Will do more later tonight.

That's fine, I'm moving some more stuff to cpp files now rather than cluttering the headers.

gtsam/hybrid/HybridGaussianFactor.cpp
dellaert: Esp, GaussianConditional, which did *two* tree traversals, now only does one.

dellaert:
> > Partial review. Will do more later tonight.
> 
> That's fine, I'm moving some more stuff to cpp files now rather than cluttering the headers.

OK, done, and cleaned up some missing python test changes.

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -28,56 +28,60 @@
...
+};
+
+/* *******************************************************************************/
+HybridGaussianConditional::HybridGaussianConditional(
+    const DiscreteKeys &discreteParents,
+    const HybridGaussianConditional::Conditionals &conditionals,
+    const ConstructorHelper &helper)
+    : BaseFactor(discreteParents, helper.pairs),
+      BaseConditional(helper.frontals.size()),
+      conditionals_(conditionals),
```
dellaert, line 69: These things are here twice! Once in BaseFactor, and once in conditionals_.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -26,23 +26,16 @@
 #include <gtsam/linear/GaussianFactor.h>
 #include <gtsam/linear/GaussianFactorGraph.h>
 
+#include "gtsam/hybrid/HybridFactor.h"
```
dellaert: Make <>

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -69,18 +69,10 @@ class GTSAM_EXPORT HybridGaussianFactor : public HybridFactor {
   /// typedef for Decision Tree of Gaussian factors.
   using Factors = DecisionTree<Key, sharedFactor>;
 
- private:
+ protected:
```
dellaert: Can remain private I think

gtsam/hybrid/HybridNonlinearFactor.cpp
```diff
@@ -18,55 +18,78 @@
 
 #include <gtsam/hybrid/HybridNonlinearFactor.h>
 
+#include <memory>
+
+#include "gtsam/nonlinear/NonlinearFactor.h"
```
dellaert: make <>

dellaert (commented):
Some things I noticed while looking over PR

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -28,56 +28,60 @@
...
+      double value = 0.0;
+      if (c) {
+        if (frontals.empty()) {
+          frontals = KeyVector(c->frontals().begin(), c->frontals().end());
+          parents = KeyVector(c->parents().begin(), c->parents().end());
+        }
+        value = c->negLogConstant();
+        negLogConstant = std::min(negLogConstant, value);
+      }
+      return {std::dynamic_pointer_cast<GaussianFactor>(c), value};
```
dellaert, line 52: Because we store the value, we should never have to call negLogConstant in this file again!
And that’s why we currently need to keep conditionals :-( at least until the hiding is removed.

dellaert:
@varunagrawal whats up with windows CI?

varunagrawal:
I think the boost download link is invalid. It's affecting multiple PRs.

gtsam/hybrid/HybridGaussianConditional.cpp
varunagrawal: Yeah I thought about that too. There was a reason I couldn't get rid of `conditionals_`, which I think was to do with `negLogConstant`. I'll think about this again so we can remove it.

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -27,57 +27,66 @@
...
+HybridGaussianConditional::HybridGaussianConditional(
+    const DiscreteKeys &discreteParents,
+    const HybridGaussianConditional::Conditionals &conditionals,
+    const ConstructorHelper &helper)
+    : BaseFactor(discreteParents, helper.pairs),
+      BaseConditional(*helper.nrFrontals),
+      conditionals_(conditionals),
+      negLogConstant_(helper.minNegLogConstant) {}
 
 HybridGaussianConditional::HybridGaussianConditional(
```
varunagrawal, line 72: Add `/* *******************************************************************************/` above here?
I can do it later.

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -27,57 +27,66 @@
...
-        if (conditional) {
-          this->negLogConstant_ =
-              std::min(this->negLogConstant_, conditional->negLogConstant());
-        }
-      });
-}
+    : HybridGaussianConditional(discreteParents, conditionals,
+                                ConstructorHelper(conditionals)) {}
+
+HybridGaussianConditional::HybridGaussianConditional(
```
varunagrawal, line 78: Add divider `/* *******************************************************************************/` here as well. I can do it later.

gtsam/hybrid/HybridGaussianConditional.h
```diff
@@ -207,6 +182,23 @@ class GTSAM_EXPORT HybridGaussianConditional
   /// @}
 
  private:
+  /// Helper struct for private constructor.
+  struct ConstructorHelper;
```
varunagrawal, line 186: Do you think it makes sense to put this in its own file (or in HybridGaussianFactor.h`) so we can reuse it? Perhaps not since we record the `minNegLogConstant` but the semantics should be the same as using the value in HybridGaussianFactor.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -73,22 +65,88 @@ static HybridGaussianFactor::Factors augment(
...
+    const DiscreteKey &discreteKey,
+    const std::vector<GaussianFactor::shared_ptr> &factors)
+    : HybridGaussianFactor(ConstructorHelper(discreteKey, factors)) {}
+
+HybridGaussianFactor::HybridGaussianFactor(
+    const DiscreteKey &discreteKey,
+    const std::vector<GaussianFactorValuePair> &factorPairs)
+    : HybridGaussianFactor(ConstructorHelper(discreteKey, factorPairs)) {}
+
+HybridGaussianFactor::HybridGaussianFactor(const DiscreteKeys &discreteKeys,
```
varunagrawal, line 139: All three of the above constructors need the divider 
`/* *******************************************************************************/`

varunagrawal: This is more for me when I add it.

varunagrawal (approved):
LGTM

The whole `ConstructorHelper` code seems a bit too convoluted, and not specifying continuous keys feels a bit strange but that may be my not being used to it.
Good thing we have a slew of unit tests. :)

gtsam/hybrid/HybridGaussianConditional.h
dellaert: Probably not. Hiding will change things anyway.

PR #1853, opened 2024-09-29
More refactoring

dellaert:
- added eliminate as method of HGFG to test in isolation
- moved HBN test to right place
- added three "bougie" constructors that considerably simplify tests and later examples:
```c++
/// Construct from mean `mu_i` and `sigma_i`.
  HybridGaussianConditional(
      const DiscreteKey mode, Key key,  //
      const std::vector<std::pair<Vector, double>> &parameters);

  /// Construct from conditional mean `A1 p1 + b_i` and `sigma_i`.
  HybridGaussianConditional(
      const DiscreteKey mode, Key key,  //
      const Matrix &A, Key parent,
      const std::vector<std::pair<Vector, double>> &parameters);

  /// Construct from conditional mean `A1 p1 + A2 p2 + b_i` and `sigma_i`.
  HybridGaussianConditional(
      const DiscreteKey mode, Key key,  //
      const Matrix &A1, Key parent1, const Matrix &A2, Key parent2,
      const std::vector<std::pair<Vector, double>> &parameters);
 ```

### DIFF
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -51,7 +51,7 @@ struct HybridGaussianConditional::Helper {
     std::vector<GC::shared_ptr> gcs;
     fvs.reserve(p.size());
     gcs.reserve(p.size());
-    for (const auto &[mean, sigma] : p) {
+    for (auto &&[mean, sigma] : p) {
       auto gaussianConditional =
           GC::sharedMeanAndStddev(std::forward<Args>(args)..., mean, sigma);
       double value = gaussianConditional->negLogConstant();
@@ -96,38 +96,34 @@ HybridGaussianConditional::HybridGaussianConditional(
       conditionals_(helper.conditionals),
       negLogConstant_(helper.minNegLogConstant) {}
 
-/* *******************************************************************************/
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKey &mode,
+    const DiscreteKey &discreteParent,
     const std::vector<GaussianConditional::shared_ptr> &conditionals)
-    : HybridGaussianConditional(DiscreteKeys{mode},
-                                Conditionals({mode}, conditionals)) {}
+    : HybridGaussianConditional(DiscreteKeys{discreteParent},
+                                Conditionals({discreteParent}, conditionals)) {}
 
-/* *******************************************************************************/
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKey &mode, Key key,  //
+    const DiscreteKey &discreteParent, Key key,  //
     const std::vector<std::pair<Vector, double>> &parameters)
-    : HybridGaussianConditional(DiscreteKeys{mode},
-                                Helper(mode, parameters, key)) {}
+    : HybridGaussianConditional(DiscreteKeys{discreteParent},
+                                Helper(discreteParent, parameters, key)) {}
 
-/* *******************************************************************************/
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKey &mode, Key key,  //
+    const DiscreteKey &discreteParent, Key key,  //
     const Matrix &A, Key parent,
     const std::vector<std::pair<Vector, double>> &parameters)
-    : HybridGaussianConditional(DiscreteKeys{mode},
-                                Helper(mode, parameters, key, A, parent)) {}
+    : HybridGaussianConditional(
+          DiscreteKeys{discreteParent},
+          Helper(discreteParent, parameters, key, A, parent)) {}
 
-/* *******************************************************************************/
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKey &mode, Key key,  //
+    const DiscreteKey &discreteParent, Key key,  //
     const Matrix &A1, Key parent1, const Matrix &A2, Key parent2,
     const std::vector<std::pair<Vector, double>> &parameters)
     : HybridGaussianConditional(
-          DiscreteKeys{mode},
-          Helper(mode, parameters, key, A1, parent1, A2, parent2)) {}
+          DiscreteKeys{discreteParent},
+          Helper(discreteParent, parameters, key, A1, parent1, A2, parent2)) {}
 
-/* *******************************************************************************/
 HybridGaussianConditional::HybridGaussianConditional(
     const DiscreteKeys &discreteParents,
     const HybridGaussianConditional::Conditionals &conditionals)
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -33,14 +33,14 @@
 
 namespace gtsam {
 /* *******************************************************************************/
-struct HybridGaussianConditional::ConstructorHelper {
+struct HybridGaussianConditional::Helper {
   std::optional<size_t> nrFrontals;
   FactorValuePairs pairs;
   Conditionals conditionals;
   double minNegLogConstant;
 
   /// Construct from tree of GaussianConditionals.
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: gtsam/hybrid/HybridGaussianConditional.h [hunk 1]
```diff
@@ -79,14 +79,57 @@ class GTSAM_EXPORT HybridGaussianConditional
...
    * @brief Construct from one discrete key and vector of conditionals.
    *
-   * @param discreteParent Single discrete parent variable
+   * @param mode Single discrete parent variable
    * @param conditionals Vector of conditionals with the same size as the
    * cardinality of the discrete parent.
    */
   HybridGaussianConditional(
-      const DiscreteKey &discreteParent,
+      const DiscreteKey &mode,
```
varunagrawal: I kind of like `discreteParent` more. It is self-documenting.

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -26,21 +26,49 @@
...
+  template <typename... Args>
+  explicit Helper(const DiscreteKey &mode, const P &p, Args &&...args) {
+    nrFrontals = 1;
+    minNegLogConstant = std::numeric_limits<double>::infinity();
+
+    std::vector<GaussianFactorValuePair> fvs;
+    std::vector<GC::shared_ptr> gcs;
+    fvs.reserve(p.size());
+    gcs.reserve(p.size());
+    for (const auto &[mean, sigma] : p) {
```
varunagrawal: Any reason we're doing `const auto&` and not `auto&&`. The latter is usually preferred for argument forward since it accepts all types of constness.

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -51,37 +79,54 @@ struct HybridGaussianConditional::ConstructorHelper {
...
-    const HybridGaussianConditional::Conditionals &conditionals,
-    const ConstructorHelper &helper)
+    const DiscreteKeys &discreteParents, const Helper &helper)
     : BaseFactor(discreteParents, helper.pairs),
       BaseConditional(*helper.nrFrontals),
-      conditionals_(conditionals),
+      conditionals_(helper.conditionals),
       negLogConstant_(helper.minNegLogConstant) {}
 
-/* *******************************************************************************/
```
varunagrawal, line 72: I am guessing you didn't merge the latest develop with my updates?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -510,20 +509,13 @@ EliminateHybrid(const HybridGaussianFactorGraph &factors,
...
     // Case 3: We are now in the hybrid land!
-    KeySet frontalKeysSet(frontalKeys.begin(), frontalKeys.end());
-
-    // Find all discrete keys.
-    // Since we eliminate all continuous variables first,
-    // the discrete separator will be *all* the discrete keys.
-    std::set<DiscreteKey> discreteSeparator = factors.discreteKeys();
-
-    return hybridElimination(factors, frontalKeys, discreteSeparator);
+    return factors.eliminate(keys);
```
varunagrawal, line 519: Would it be better to call `factors.hybridEliminate`? This way we can also do `factors.discreteEliminate` and `factors.continuousEliminate`.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -436,7 +435,7 @@ hybridElimination(const HybridGaussianFactorGraph &factors,
  */
 std::pair<HybridConditional::shared_ptr, std::shared_ptr<Factor>>  //
 EliminateHybrid(const HybridGaussianFactorGraph &factors,
-                const Ordering &frontalKeys) {
+                const Ordering &keys) {
```
varunagrawal, line 439: This deviates from the naming scheme of `GaussianFactorGraph` and other graphs, where it is `frontalKeys`. Should we strive for consistency?

gtsam/hybrid/tests/testGaussianMixture.cpp
```diff
@@ -99,7 +79,11 @@ TEST(GaussianMixture, GaussianMixtureModel) {
   double mu0 = 1.0, mu1 = 3.0;
   double sigma = 2.0;
 
-  auto hbn = GaussianMixtureModel(mu0, mu1, sigma, sigma);
+  HybridBayesNet hbn;
```
varunagrawal: We should add a comment about what's going on here, since before it was pretty obvious we were defining a `GaussianMixtureModel`.

gtsam/hybrid/tests/testGaussianMixture.cpp
```diff
@@ -133,7 +117,11 @@ TEST(GaussianMixture, GaussianMixtureModel2) {
   double mu0 = 1.0, mu1 = 3.0;
   double sigma0 = 8.0, sigma1 = 4.0;
 
-  auto hbn = GaussianMixtureModel(mu0, mu1, sigma0, sigma1);
+  HybridBayesNet hbn;
```
varunagrawal: Comment about `GaussianMixtureModel`. The function was self-documenting in a way.

varunagrawal (approved):
Approve modulo comments.

I like the new API but it would also be nice to have at least one test using the old API as a tutorial, aka "this is good, this new thing is better".

gtsam/hybrid/HybridGaussianConditional.h
dellaert: Ok, i’’l fix

gtsam/hybrid/HybridGaussianConditional.cpp
dellaert: Sure

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: It’s a HFG, eliminate is hybrid by definition.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Ok, I’ll fix

gtsam/hybrid/tests/testGaussianMixture.cpp
dellaert: Ok

gtsam/hybrid/tests/testGaussianMixture.cpp
dellaert: Ok

dellaert:
I’ll address the comments. But 
- I don’t want the separators between the constructors
- The includes should not depend on transitive inclusion, see Google style guide
Maybe let me edit to avoid conflicts? You have enough to do with experiments :-)

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Actually, no,I went back and looked at GFG and it's always just "keys". So I'll leave this to be consistent. frontal keys is a concept in conditionals, not in elimination. The keys we eliminate will *become* frontal keys.

varunagrawal: Oh okay. I must have been mistaken then.

varunagrawal:
Sure thing wrt the dividers. It's a nitpick anyway :)

varunagrawal:
> - The includes should not depend on transitive inclusion, see Google style guide

Interesting. I like to minimize includes since that helps cut down compilation time. Less header copying.

> Maybe let me edit to avoid conflicts? You have enough to do with experiments :-)

Fair. I was updating the code to examine the ordering, so decided to save a compilation cycle by making the edits.

PR #1856, opened 2024-09-30
Navigation docs

dellaert:
- Generated explainer for attitude factor with o1-preview
- Edited above heavily
- Added README
- Updated IMU Factor reference


gtsam/navigation/AttitudeFactor.h
```diff
@@ -17,59 +17,61 @@
...
- * This factor will give zero error if nRb * bF == nG
+ * @class AttitudeFactor
+ *
+ * @brief Base class for an attitude factor that constrains the rotation between
+ * body and navigation frames.
+ *
+ * This factor enforces that when a known reference direction in the body frame
+ * (e.g., accelerometer axis) is rotated into the navigation frame using the
+ * rotation variable, it should align with a measured direction in the
+ * navigation frame (e.g., gravity vector).
```
andre-michelin, line 35: What is confusing here is the choice of ref / measured: 
- one would expect the direction read from the accelerometer (in body) to be *measured*, and the known gravity direction in nav to be *ref*

gtsam/navigation/AttitudeFactor.md
```diff
@@ -0,0 +1,142 @@
...
+}
+```
+
+#### Explanation:
+- The function computes the rotated reference direction `nRef` and then the error between `nRef` and the measured direction `nZ`.
+- `nZ_.error(nRef)` is a 2D vector-valued error between two directions, defined in [Unit3.h](../geometry/Unit3.h).
+
+### Jacobians
+
+For optimization, the $2 \times 3$ Jacobian of the error function with respect to the rotation parameters is required. The Jacobian is computed using chain rule differentiation, involving the derivative of the rotated vector with respect to the rotation parameters and the derivative of the error with respect to the rotated vector.
```
andre-michelin, line 52: Maybe a mention here of the singularity we found w.r.t. the Jacobian, which also vanishes 180 degrees away from the true direction (because it is calculated in the tangent space?)

andre-michelin (approved):

gtsam/navigation/AttitudeFactor.h
dellaert: OK. This was a big change, but I switched all the mentions of reference and measured. I agree it is much more clear and even the tests had "measurement" in quotes.

gtsam/navigation/AttitudeFactor.md
dellaert: Done

PR #1857, opened 2024-10-01
Fix hybrid bug, double performance

dellaert:
I could not really separate the functional pruning PR from this one. Here's everything in this PR:

## Fixing the bug
- BIG: fix bug in HybridConditional::errorTree and HGFG::errorTree
- made errorTree much faster by calling polymorphic errorTree and using AlgebraicDecisionTree operators

## `discretePosterior`
- renamed both `HBN::evaluate` and `HGFG::probPrime` that take VectorValues to `discretePosterior`: these are very powerful. 

## Simplified API 
- removed `HBN::logProbability`: errorTree is all we need and this one has confusing semantics
- removed `HGC::logProbability` that takes continuous values: very problematic name and only used in one place: again, errorTree is all we need.

## Functional pruning
- pruning is now functional
- got rid of duplicate (unused) prunerFunc
- added a write-access conditional to allow iSAM versions to be still imperative

## Update Oct 1:
- Reworked `HGC::prune` to use `ADT::max`
- Rewrote `HybridBayesNet::prune` completely

Looking at pruned trees for testHybridEstimation is very informative: it seems that the ordering of the keys is working against us in terms of pruning: we have large duplicate trees:

step 2:
image

step 5:
image

...

step 14:
image

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -17,9 +17,12 @@
  */
 
 #include <gtsam/discrete/DiscreteBayesNet.h>
+#include <gtsam/discrete/DiscreteConditional.h>
 #include <gtsam/discrete/DiscreteFactorGraph.h>
 #include <gtsam/hybrid/HybridBayesNet.h>
 #include <gtsam/hybrid/HybridValues.h>
+
+#include <memory>
 
 // In Wrappers we have no access to this so have a default ready
 static std::mt19937_64 kRandomNumberGenerator(42);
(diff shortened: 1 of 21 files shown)


dellaert:
Note: “doubling performance” refers to testHybridEstimation. Could be twice as fast because now the discrete factors are properly taken into account when pruning.

gtsam/hybrid/HybridConditional.cpp
```diff
@@ -100,79 +99,68 @@ bool HybridConditional::equals(const HybridFactor &other, double tol) const {
...
+    return {gc->error(values)};  // NOTE: a "constant" tree
+  } else if (auto gm = asHybrid()) {
     return gm->errorTree(values);
-  }
-  if (auto dc = asDiscrete()) {
-    return AlgebraicDecisionTree<Key>(0.0);
-  }
-  throw std::runtime_error(
-      "HybridConditional::error: conditional type not handled");
+  } else if (auto dc = asDiscrete()) {
```
dellaert, line 133: Bug 1

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -499,22 +505,22 @@ EliminateHybrid(const HybridGaussianFactorGraph &factors,
...
-      error_tree = error_tree + f->errorTree(continuousValues);
-    } else if (auto f = std::dynamic_pointer_cast<DiscreteFactor>(factor)) {
-      // Skip discrete factors
-      continue;
+    if (auto hf = std::dynamic_pointer_cast<HybridFactor>(factor)) {
+      // Add errorTree for hybrid factors, includes HybridGaussianConditionals!
+      result = result + hf->errorTree(continuousValues);
+    } else if (auto df = std::dynamic_pointer_cast<DiscreteFactor>(factor)) {
+      // If discrete, just add its errorTree as well
+      result = result + df->errorTree();
```
dellaert, line 516: Bug 2

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -164,9 +76,10 @@ DecisionTreeFactor HybridBayesNet::pruneDiscreteConditionals(
 }
 
 /* ************************************************************************* */
-HybridBayesNet HybridBayesNet::prune(size_t maxNrLeaves) {
+HybridBayesNet HybridBayesNet::prune(size_t maxNrLeaves) const {
+  HybridBayesNet copy(*this);
   DecisionTreeFactor prunedDiscreteProbs =
-      this->pruneDiscreteConditionals(maxNrLeaves);
+      copy.pruneDiscreteConditionals(maxNrLeaves);
```
varunagrawal: Why a copy if everything is functional?

gtsam/hybrid/HybridBayesNet.h
```diff
@@ -221,29 +217,33 @@ class GTSAM_EXPORT HybridBayesNet : public BayesNet<HybridConditional> {
...
-   * @brief Compute log probability for each discrete assignment,
-   * and return as a tree.
+   * @brief Compute the negative log posterior log P'(M|x) of all assignments up
+   * to a constant, returning the result as an algebraic decision tree.
    *
-   * @param continuousValues Continuous values at which
-   * to compute the log probability.
+   * @note The joint P(X,M) is p(X|M) P(M)
+   * Then the posterior on M given X=x is is P(M|x) = p(x|M) P(M) / p(x).
+   * Ideally we want log P(M|x) = log p(x|M) + log P(M) - log P(x), but
```
varunagrawal: Should be `log p(x)` not `log P(x)`

gtsam/hybrid/tests/testHybridNonlinearFactorGraph.cpp
```diff
@@ -994,16 +994,9 @@ TEST(HybridNonlinearFactorGraph, DifferentCovariances) {
...
 
-  // Check that the error values at the MLE point μ.
-  AlgebraicDecisionTree<Key> errorTree = hbn->errorTree(cv);
-
   DiscreteValues dv0{{M(1), 0}};
   DiscreteValues dv1{{M(1), 1}};
 
-  // regression
-  EXPECT_DOUBLES_EQUAL(9.90348755254, errorTree(dv0), 1e-9);
-  EXPECT_DOUBLES_EQUAL(0.69314718056, errorTree(dv1), 1e-9);
```
varunagrawal, line 1005: I kind of liked these regression tests.

varunagrawal (approved):
Some questions, especially about a copy.

gtsam/hybrid/HybridBayesNet.cpp
dellaert: You're right, it was left-over ! I'll fix

dellaert: Actually, I messed up the discrete part. Did you ever profile the pruning as it is? It's an exponential operation - that alone might be responsible for slowness in some experiments.

I'll fix but we have to chat about this more I think.

gtsam/hybrid/tests/testHybridNonlinearFactorGraph.cpp
dellaert: I will replace them with better, non-regression test based on discreteConditional.

gtsam/hybrid/HybridConditional.cpp
varunagrawal: Oh interesting, I was missing the `DiscreteFactor::errorTree` method. There's actually an improvement here I can make.

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: I haven't. I would like to see your perf-profiling setup and make it standard across projects. :) Seems to be very useful.

dellaert: I'm on a mac. I just build the binary and use Instruments CPU profiler.

dellaert: OK, I fixed it. 

@varunagrawal :
- You probably had pruning tests in your project, but I added some here so I could refactor HGC::prune based on ADT::max. 
- You might want to check out his branch and try whether your experiments still work before we merge?

dellaert:
@varunagrawal please merge this after you established your experiments work?

varunagrawal:
This looks great so merging!

PR #1858, opened 2024-10-01
Implement `errorTree` in `DiscreteFactor`

varunagrawal:
Common base class method to reduce duplication.

### DIFF
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -63,22 +63,6 @@ namespace gtsam {
   }
 
   /* ************************************************************************ */
-  AlgebraicDecisionTree<Key> DecisionTreeFactor::errorTree() const {
-    // Get all possible assignments
-    DiscreteKeys dkeys = discreteKeys();
-    // Reverse to make cartesian product output a more natural ordering.
-    DiscreteKeys rdkeys(dkeys.rbegin(), dkeys.rend());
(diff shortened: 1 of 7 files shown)


dellaert (commented):
Hmmmm. I'm worried about complexity here, *especially* if we have pruned branches. I think you replaced a sometimes errorTree with an always expensive version.

varunagrawal:
> Hmmmm. I'm worried about complexity here, _especially_ if we have pruned branches. I think you replaced a sometimes errorTree with an always expensive version.

Interesting. I copy-pasted the code from `DecisionTreeFactor` to `DiscreteFactor` since `TableFactor` was anyway calling the `DecisionTreeFactor` version.
I didn't see this potential issue because except for `discreteKeys` there is no class specific method being used here. And `discreteKeys` should be the same for both `DiscreteFactor` and `DecisionTreeFactor`.

Maybe the original implementation was already expensive to begin with?

dellaert:
Actually, you are right! We should just call the DecisionTree constructor in the DTF. That is a potential speedup! So, I like your base class implementation for generic, but it should speed up for derived class. Probably also in TableFactor: is toDecisionTreeFactor efficient?

varunagrawal:
> Probably also in TableFactor: is toDecisionTreeFactor efficient?

It should be. It's from @ywkim0606's original PR which you had worked on with him.

dellaert (approved):

PR #1859, opened 2024-10-01
Implement `errorTree` for HybridNonlinearFactorGraph

varunagrawal:
Following the `HybridGaussianFactorGraph::errorTree` method, implement the same for `HybridNonlinearFactorGraph`.
Note that since the input is `gtsam::Values`, we only consider nonlinear and discrete factors, not Gaussian factors.


dellaert (approved):
Nice. But, I am loathe to introduce more regression tests. In this case, the network is small enough that you could just create the correct result in the test?

varunagrawal:
Okay I'll do that.

PR #1862, opened 2024-10-06
Fix Pruning

varunagrawal:
Update the return value of pruned factors/conditionals to return ∞ as the error in negative log space.

This way, when we normalize (by subtracting the minimum value) and take the exponential negative `exp(-e)`, we get `p=0` for pruned assignments, valid probabilities everywhere else.

With this fix, we can now run hybrid legged state estimation for 16 timesteps in 3.8 seconds, down from 13 minutes.


dellaert (approved):
Very cool. Might want to update the PR comment with what you told me in email. Also, a test that fails before but works after would be great, so future PR’s don’t mess this up.

varunagrawal:
Added a test which originally fails, and found another subtle bug which I added a fix for.
Furthermore, improved `testHybridEstimation.cpp` quite a bit.

PR #1863, opened 2024-10-08
Refactored hybrid inference

dellaert:
This is a big PR that gets rind of "hiding" the conditional constants into the Jacobian factors in HGF, in favor of storing a scalar alongside each factor in the HGFG, as is done in HNFG. It also adds *extensive* testing.

## Details

- GaussianFactorGraphTree is now its own class, called HybridGaussianProductFactor
- asGaussianFactorGraphTree is renamed to collectProductFactor, and now only exists in HGF, no longer in HGC
- scalars are initialized with each conditionals *adjusted* negLogConstants, i.e. scalars are zero if they are the same for each mode.
- `augment` is GONE
- added methods discretePosterior which yield much easier "ratio tests".

## Testing

- added many more tests in testHGC, testHBN, and testHGFG
- moved all multi-frontal elimination tests to testHybridBayesTree

## TODO:
- replace all ratio tests
- get rid of HGC::conditionals_ if improves performance

### DIFF
--- a/gtsam/hybrid/HybridGaussianConditional.cpp
+++ b/gtsam/hybrid/HybridGaussianConditional.cpp
@@ -22,8 +22,8 @@
 #include <gtsam/discrete/DiscreteValues.h>
 #include <gtsam/hybrid/HybridGaussianConditional.h>
 #include <gtsam/hybrid/HybridGaussianFactor.h>
-#include <gtsam/hybrid/HybridValues.h>
 #include <gtsam/hybrid/HybridGaussianProductFactor.h>
+#include <gtsam/hybrid/HybridValues.h>
 #include <gtsam/inference/Conditional-inst.h>
 #include <gtsam/linear/GaussianBayesNet.h>
 #include <gtsam/linear/GaussianFactorGraph.h>
@@ -44,7 +44,7 @@ struct HybridGaussianConditional::Helper {
 
   /// Construct from a vector of mean and sigma pairs, plus extra args.
   template <typename... Args>
-  explicit Helper(const DiscreteKey &mode, const P &p, Args &&...args) {
+  explicit Helper(const DiscreteKey& mode, const P& p, Args&&... args) {
     nrFrontals = 1;
     minNegLogConstant = std::numeric_limits<double>::infinity();
 
@@ -52,9 +52,8 @@ struct HybridGaussianConditional::Helper {
     std::vector<GC::shared_ptr> gcs;
     fvs.reserve(p.size());
     gcs.reserve(p.size());
-    for (auto &&[mean, sigma] : p) {
-      auto gaussianConditional =
-          GC::sharedMeanAndStddev(std::forward<Args>(args)..., mean, sigma);
+    for (auto&& [mean, sigma] : p) {
+      auto gaussianConditional = GC::sharedMeanAndStddev(std::forward<Args>(args)..., mean, sigma);
       double value = gaussianConditional->negLogConstant();
       minNegLogConstant = std::min(minNegLogConstant, value);
       fvs.emplace_back(gaussianConditional, value);
@@ -66,10 +65,9 @@ struct HybridGaussianConditional::Helper {
   }
 
   /// Construct from tree of GaussianConditionals.
-  explicit Helper(const Conditionals &conditionals)
-      : conditionals(conditionals),
-        minNegLogConstant(std::numeric_limits<double>::infinity()) {
-    auto func = [this](const GC::shared_ptr &c) -> GaussianFactorValuePair {
+  explicit Helper(const Conditionals& conditionals)
+      : conditionals(conditionals), minNegLogConstant(std::numeric_limits<double>::infinity()) {
+    auto func = [this](const GC::shared_ptr& c) -> GaussianFactorValuePair {
       double value = 0.0;
       if (c) {
         if (!nrFrontals.has_value()) {
@@ -90,56 +88,61 @@ struct HybridGaussianConditional::Helper {
 };
 
 /* *******************************************************************************/
-HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKeys &discreteParents, const Helper &helper)
+HybridGaussianConditional::HybridGaussianConditional(const DiscreteKeys& discreteParents,
+                                                     const Helper& helper)
     : BaseFactor(discreteParents, helper.pairs),
       BaseConditional(*helper.nrFrontals),
       conditionals_(helper.conditionals),
       negLogConstant_(helper.minNegLogConstant) {}
 
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKey &discreteParent,
-    const std::vector<GaussianConditional::shared_ptr> &conditionals)
+    const DiscreteKey& discreteParent,
+    const std::vector<GaussianConditional::shared_ptr>& conditionals)
     : HybridGaussianConditional(DiscreteKeys{discreteParent},
                                 Conditionals({discreteParent}, conditionals)) {}
 
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKey &discreteParent, Key key,  //
-    const std::vector<std::pair<Vector, double>> &parameters)
+    const DiscreteKey& discreteParent,
+    Key key,  //
+    const std::vector<std::pair<Vector, double>>& parameters)
     : HybridGaussianConditional(DiscreteKeys{discreteParent},
                                 Helper(discreteParent, parameters, key)) {}
 
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKey &discreteParent, Key key,  //
-    const Matrix &A, Key parent,
-    const std::vector<std::pair<Vector, double>> &parameters)
-    : HybridGaussianConditional(
-          DiscreteKeys{discreteParent},
-          Helper(discreteParent, parameters, key, A, parent)) {}
+    const DiscreteKey& discreteParent,
+    Key key,  //
+    const Matrix& A,
+    Key parent,
+    const std::vector<std::pair<Vector, double>>& parameters)
+    : HybridGaussianConditional(DiscreteKeys{discreteParent},
+                                Helper(discreteParent, parameters, key, A, parent)) {}
 
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKey &discreteParent, Key key,  //
-    const Matrix &A1, Key parent1, const Matrix &A2, Key parent2,
-    const std::vector<std::pair<Vector, double>> &parameters)
-    : HybridGaussianConditional(
-          DiscreteKeys{discreteParent},
-          Helper(discreteParent, parameters, key, A1, parent1, A2, parent2)) {}
+    const DiscreteKey& discreteParent,
+    Key key,  //
+    const Matrix& A1,
+    Key parent1,
+    const Matrix& A2,
+    Key parent2,
+    const std::vector<std::pair<Vector, double>>& parameters)
+    : HybridGaussianConditional(DiscreteKeys{discreteParent},
+                                Helper(discreteParent, parameters, key, A1, parent1, A2, parent2)) {
+}
 
 HybridGaussianConditional::HybridGaussianConditional(
-    const DiscreteKeys &discreteParents,
-    const HybridGaussianConditional::Conditionals &conditionals)
+    const DiscreteKeys& discreteParents,
+    const HybridGaussianConditional::Conditionals& conditionals)
     : HybridGaussianConditional(discreteParents, Helper(conditionals)) {}
 
 /* *******************************************************************************/
-const HybridGaussianConditional::Conditionals &
-HybridGaussianConditional::conditionals() const {
+const HybridGaussianConditional::Conditionals& HybridGaussianConditional::conditionals() const {
   return conditionals_;
 }
 
 /* *******************************************************************************/
-HybridGaussianProductFactor HybridGaussianConditional::asProductFactor()
-    const {
-  auto wrap = [this](const std::shared_ptr<GaussianConditional> &gc) {
+HybridGaussianProductFactor HybridGaussianConditional::asProductFactor() const {
+  auto wrap = [this](const std::shared_ptr<GaussianConditional>& gc)
+      -> std::pair<GaussianFactorGraph, double> {
     // First check if conditional has not been pruned
     if (gc) {
       const double Cgm_Kgcm = gc->negLogConstant() - this->negLogConstant_;
@@ -151,10 +154,17 @@ HybridGaussianProductFactor HybridGaussianConditional::asProductFactor()
         Vector c(1);
         c << std::sqrt(2.0 * Cgm_Kgcm);
         auto constantFactor = std::make_shared<JacobianFactor>(c);
-        return GaussianFactorGraph{gc, constantFactor};
+        return {GaussianFactorGraph{gc, constantFactor}, Cgm_Kgcm};
+      } else {
+        // The scalar can be zero.
+        // TODO(Frank): after hiding is gone, this should be only case here.
+        return {GaussianFactorGraph{gc}, Cgm_Kgcm};
       }
+    } else {
+      // If the conditional is pruned, return an empty GaussianFactorGraph with zero scalar sum
+      // TODO(Frank): Could we just return an *empty* GaussianFactorGraph?
+      return {GaussianFactorGraph{nullptr}, 0.0};
     }
-    return GaussianFactorGraph{gc};
   };
   return {{conditionals_, wrap}};
 }
@@ -162,7 +172,7 @@ HybridGaussianProductFactor HybridGaussianConditional::asProductFactor()
 /* *******************************************************************************/
 size_t HybridGaussianConditional::nrComponents() const {
   size_t total = 0;
-  conditionals_.visit([&total](const GaussianFactor::shared_ptr &node) {
+  conditionals_.visit([&total](const GaussianFactor::shared_ptr& node) {
     if (node) total += 1;
   });
   return total;
@@ -170,21 +180,19 @@ size_t HybridGaussianConditional::nrComponents() const {
 
 /* *******************************************************************************/
 GaussianConditional::shared_ptr HybridGaussianConditional::choose(
-    const DiscreteValues &discreteValues) const {
-  auto &ptr = conditionals_(discreteValues);
+    const DiscreteValues& discreteValues) const {
+  auto& ptr = conditionals_(discreteValues);
   if (!ptr) return nullptr;
   auto conditional = std::dynamic_pointer_cast<GaussianConditional>(ptr);
   if (conditional)
     return conditional;
   else
-    throw std::logic_error(
-        "A HybridGaussianConditional unexpectedly contained a non-conditional");
+    throw std::logic_error("A HybridGaussianConditional unexpectedly contained a non-conditional");
 }
 
 /* *******************************************************************************/
-bool HybridGaussianConditional::equals(const HybridFactor &lf,
-                                       double tol) const {
-  const This *e = dynamic_cast<const This *>(&lf);
+bool HybridGaussianConditional::equals(const HybridFactor& lf, double tol) const {
+  const This* e = dynamic_cast<const This*>(&lf);
   if (e == nullptr) return false;
 
   // This will return false if either conditionals_ is empty or e->conditionals_
@@ -193,27 +201,26 @@ bool HybridGaussianConditional::equals(const HybridFactor &lf,
 
   // Check the base and the factors:
   return BaseFactor::equals(*e, tol) &&
-         conditionals_.equals(
-             e->conditionals_, [tol](const auto &f1, const auto &f2) {
-               return (!f1 && !f2) || (f1 && f2 && f1->equals(*f2, tol));
-             });
+         conditionals_.equals(e->conditionals_, [tol](const auto& f1, const auto& f2) {
+           return (!f1 && !f2) || (f1 && f2 && f1->equals(*f2, tol));
+         });
 }
 
 /* *******************************************************************************/
-void HybridGaussianConditional::print(const std::string &s,
-                                      const KeyFormatter &formatter) const {
+void HybridGaussianConditional::print(const std::string& s, const KeyFormatter& formatter) const {
   std::cout << (s.empty() ? "" : s + "\n");
   BaseConditional::print("", formatter);
   std::cout << " Discrete Keys = ";
-  for (auto &dk : discreteKeys()) {
+  for (auto& dk : discreteKeys()) {
     std::cout << "(" << formatter(dk.first) << ", " << dk.second << "), ";
   }
   std::cout << std::endl
             << " logNormalizationConstant: " << -negLogConstant() << std::endl
             << std::endl;
   conditionals_.print(
-      "", [&](Key k) { return formatter(k); },
-      [&](const GaussianConditional::shared_ptr &gf) -> std::string {
+      "",
+      [&](Key k) { return formatter(k); },
+      [&](const GaussianConditional::shared_ptr& gf) -> std::string {
         RedirectCout rd;
         if (gf && !gf->empty()) {
           gf->print("", formatter);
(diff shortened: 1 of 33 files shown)


varunagrawal:
Would it be possible to break this into a separate PR for the `HybridGaussianProductFactor`?

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -237,7 +225,7 @@ KeyVector HybridGaussianConditional::continuousParents() const {
     // remove that key from continuousParentKeys:
     continuousParentKeys.erase(std::remove(continuousParentKeys.begin(),
                                            continuousParentKeys.end(), key),
-                               continuousParentKeys.end());
+        continuousParentKeys.end());
```
varunagrawal: Are you using different formatter settings? My formatter is aligning this with the opening parenthesis.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -18,83 +18,52 @@
...
 
-  ConstructorHelper(const DiscreteKey &discreteKey,
-                    const std::vector<GaussianFactor::shared_ptr> &factors)
+  /// Constructor for a single discrete key and a vector of Gaussian factors
+  ConstructorHelper(const DiscreteKey& discreteKey,
+                    const std::vector<GaussianFactor::shared_ptr>& factors)
       : discreteKeys({discreteKey}) {
     // Extract continuous keys from the first non-null factor
-    for (const auto &factor : factors) {
+    for (const auto& factor : factors) {
```
varunagrawal, line 45: This is another formatting thing that differs for me. I am seeing this difference consistently between your code and mine.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -186,61 +157,41 @@ void HybridGaussianFactor::print(const std::string &s,
...
+           }}};
 }
 
 /* *******************************************************************************/
 AlgebraicDecisionTree<Key> HybridGaussianFactor::errorTree(
     const VectorValues &continuousValues) const {
   // functor to convert from sharedFactor to double error value.
-  auto errorFunc = [&continuousValues](const sharedFactor &gf) {
-    return PotentiallyPrunedComponentError(gf, continuousValues);
+  auto errorFunc = [&continuousValues](const auto& pair) {
```
varunagrawal: Please put type information here (aka `sharedFactor` instead of `auto`), else it detracts from a lesson you gave me a year ago.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -153,29 +121,32 @@ bool HybridGaussianFactor::equals(const HybridFactor &lf, double tol) const {
...
+    return match && gtsam::equal(pair1.second, pair2.second, tol);
+  };
+  return Base::equals(*e, tol) && factors_.equals(e->factors_, compareFunc);
 }
 
 /* *******************************************************************************/
 void HybridGaussianFactor::print(const std::string &s,
                                  const KeyFormatter &formatter) const {
   std::cout << (s.empty() ? "" : s + "\n");
-  std::cout << "HybridGaussianFactor" << std::endl;
```
varunagrawal, line 166: Any reason this was removed?

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -153,29 +121,32 @@ bool HybridGaussianFactor::equals(const HybridFactor &lf, double tol) const {
...
   std::cout << "{\n";
   if (factors_.empty()) {
     std::cout << "  empty" << std::endl;
   } else {
     factors_.print(
-        "", [&](Key k) { return formatter(k); },
-        [&](const sharedFactor &gf) -> std::string {
+        "",
+        [&](Key k) { return formatter(k); },
+        [&](const auto& pair) -> std::string {
```
varunagrawal: Please leverage type information. It also helps with autocomplete and code analysis, beyond making things easier to read.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -123,22 +95,18 @@ struct HybridGaussianFactor::ConstructorHelper {
...
     const std::vector<GaussianFactor::shared_ptr> &factors)
     : HybridGaussianFactor(ConstructorHelper(discreteKey, factors)) {}
 
-/* *******************************************************************************/
 HybridGaussianFactor::HybridGaussianFactor(
     const DiscreteKey &discreteKey,
     const std::vector<GaussianFactorValuePair> &factorPairs)
     : HybridGaussianFactor(ConstructorHelper(discreteKey, factorPairs)) {}
 
-/* *******************************************************************************/
```
varunagrawal, line 141: These could have been in a separate PR. Makes reviewing hard...

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -186,61 +157,41 @@ void HybridGaussianFactor::print(const std::string &s,
...
 }
 
 /* *******************************************************************************/
-double HybridGaussianFactor::error(const HybridValues &values) const {
+double HybridGaussianFactor::error(const HybridValues& values) const {
   // Directly index to get the component, no need to build the whole tree.
-  const sharedFactor gf = factors_(values.discrete());
-  return PotentiallyPrunedComponentError(gf, values.continuous());
+  const auto pair = factors_(values.discrete());
+  return pair.first ? pair.first->error(values.continuous()) + pair.second
```
varunagrawal: We can keep `PotentiallyPrunedComponentError` as an inline function and use it in both `error` and `errorTree`.

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -57,7 +58,7 @@ using GaussianFactorValuePair = std::pair<GaussianFactor::shared_ptr, double>;
  * @ingroup hybrid
  */
 class GTSAM_EXPORT HybridGaussianFactor : public HybridFactor {
- public:
+public:
```
varunagrawal: What's going on here? The single space for the access specifiers is a common formatting spec throughout the codebase.

gtsam/hybrid/HybridGaussianFactor.h
```diff
@@ -148,8 +138,8 @@ class GTSAM_EXPORT HybridGaussianFactor : public HybridFactor {
    * @return AlgebraicDecisionTree<Key> A decision tree with the same keys
    * as the factors involved, and leaf values as the error.
    */
-  AlgebraicDecisionTree<Key> errorTree(
-      const VectorValues &continuousValues) const override;
+  AlgebraicDecisionTree<Key>
+  errorTree(const VectorValues &continuousValues) const override;
```
varunagrawal: Okay now I am convinced something is up with your formatter. I am being a bit strong on this, because it breaks consistency and I can't just auto-format files anymore (instead of having to be painstakingly meticulous about code chunks).

gtsam/hybrid/HybridGaussianProductFactor.cpp
```diff
@@ -0,0 +1,90 @@
...
+  Base::print(s, formatter, printer);
+  if (!keys.empty()) {
+    std::stringstream ss;
+    ss << s << " Keys:";
+    for (auto&& key : keys) ss << " " << formatter(key);
+    std::cout << ss.str() << "." << std::endl;
+  }
+}
+
+HybridGaussianProductFactor HybridGaussianProductFactor::removeEmpty() const {
```
varunagrawal, line 96: I can add the dividers later. I have found them useful when collapsing code.

gtsam/hybrid/HybridGaussianProductFactor.h
```diff
@@ -0,0 +1,129 @@
...
+  void print(const std::string& s = "",
+             const KeyFormatter& formatter = DefaultKeyFormatter) const;
+
+  /**
+   * @brief Check if this HybridGaussianProductFactor is equal to another
+   * @param other The other HybridGaussianProductFactor to compare with
+   * @param tol Tolerance for floating point comparisons
+   * @return true if equal, false otherwise
+   */
+  bool equals(const HybridGaussianProductFactor& other,
```
varunagrawal, line 99: The implementation can also go in the cpp file.

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -189,6 +193,20 @@ TEST(HybridBayesNet, Tiny) {
   auto fg = bayesNet.toFactorGraph({{Z(0), Vector1(5.0)}});
   EXPECT_LONGS_EQUAL(3, fg.size());
 
+  // Create the product factor for eliminating x0:
+  HybridGaussianFactorGraph factors_x0;
+  factors_x0.push_back(fg.at(0));
+  factors_x0.push_back(fg.at(1));
+  auto productFactor = factors_x0.collectProductFactor();
+
+  // Check that scalars are 0 and 1.79
```
varunagrawal: Maybe add `(regression)`?

gtsam/hybrid/tests/testHybridBayesTree.cpp
```diff
@@ -28,9 +29,327 @@
...
+/* ************************************************************************* */
+namespace two {
+std::vector<GaussianFactor::shared_ptr> components(Key key) {
+  return {std::make_shared<JacobianFactor>(key, I_3x3, Z_3x1),
+          std::make_shared<JacobianFactor>(key, I_3x3, Vector3::Ones())};
+}
+}  // namespace two
+
+/* ************************************************************************* */
+TEST(HybridGaussianFactorGraph, eliminateFullMultifrontalSimple) {
```
varunagrawal: Prefer to name tests in CamelCase, since the test name is set as `HybridGaussianFactorGrapheliminateFullMultifrontalSimple`. With CamelCase, it becomes `HybridGaussianFactorGraphEliminateFullMultifrontalSimple`.

gtsam/hybrid/tests/testHybridBayesTree.cpp
```diff
@@ -28,9 +29,327 @@
...
+
+  hfg.add(DecisionTreeFactor(m1, {2, 8}));
+  // TODO(Varun) Adding extra discrete variable not connected to continuous
+  // variable throws segfault
+  //  hfg.add(DecisionTreeFactor({m1, m2, "1 2 3 4"));
+
+  HybridBayesTree::shared_ptr result = hfg.eliminateMultifrontal();
+
+  // The bayes tree should have 3 cliques
+  EXPECT_LONGS_EQUAL(3, result->size());
```
varunagrawal, line 84: I don't understand this code movement. All of it primarily deals with the 	`HybridGaussianFactorGraph` and hence should be in `testHybridGaussianFactorGraph.cpp`.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
```diff
@@ -82,44 +82,29 @@ TEST(HybridGaussianFactor, ConstructorVariants) {
...
-  EXPECT(actual.at(0) == f11);
-  EXPECT(actual.at(1) == f22);
+  EXPECT_LONGS_EQUAL(3, hybridFactorB.keys().size());
+  EXPECT_LONGS_EQUAL(2, hybridFactorB.continuousKeys().size());
+  EXPECT_LONGS_EQUAL(1, hybridFactorB.discreteKeys().size());
 }
 
 /* ************************************************************************* */
-TEST(HybridGaussianFactor, Printing) {
+TEST_DISABLED(HybridGaussianFactor, Printing) {
```
varunagrawal: This should be an easy fix?

gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
```diff
@@ -187,295 +168,233 @@ TEST(HybridGaussianFactorGraph, eliminateFullSequentialSimple) {
...
-    std::vector<Key> ordC;
-    std::transform(naturalC.begin(), naturalC.end(), std::back_inserter(ordC),
-                   [](int x) { return M(x); });
-
-    // std::copy(ordC.begin(), ordC.end(), std::back_inserter(ordering));
-    const auto [ndC, lvls] = makeBinaryOrdering(ordC);
-    std::copy(ndC.begin(), ndC.end(), std::back_inserter(ordering));
+// Test API for the smallest switching network.
+// None of these are regression tests.
+TEST(HybridBayesNet, Switching) {
```
varunagrawal, line 171: Test group name change to `HybridGaussianFactorGraph`.

gtsam/hybrid/tests/testHybridGaussianProductFactor.cpp
```diff
@@ -0,0 +1,196 @@
...
+  product += hybridFactorA;
+
+  // Let's check that this worked:
+  Assignment<Key> mode;
+  mode[m1.first] = 0;
+  auto actual = product(mode);
+  EXPECT(actual.first.at(0) == f10);
+  EXPECT_DOUBLES_EQUAL(10, actual.second, 1e-9);
+
+  // TODO(Frank): when killed hiding, f11 should also be there
```
varunagrawal, line 148: Should this be done before this PR lands? Looks to be important.

gtsam/hybrid/tests/testHybridGaussianProductFactor.cpp
```diff
@@ -0,0 +1,196 @@
...
+  product += hybridFactorB;
+
+  // Let's check that this worked:
+  auto actual00 = product({{M(1), 0}, {M(2), 0}});
+  EXPECT(actual00.first.at(0) == f10);
+  EXPECT(actual00.first.at(1) == f20);
+  EXPECT_DOUBLES_EQUAL(10 + 20, actual00.second, 1e-9);
+
+  auto actual12 = product({{M(1), 1}, {M(2), 2}});
+  // TODO(Frank): when killed hiding, these should also equal:
```
varunagrawal, line 169: Same as above. Since this is a new class and makes a *significant* change, I would prefer not to let TODOs linger.

gtsam/hybrid/tests/testHybridNonlinearFactorGraph.cpp
```diff
@@ -529,7 +529,7 @@ TEST(HybridNonlinearFactorGraph, Full_Elimination) {
 /****************************************************************************
  * Test printing
  */
-TEST(HybridNonlinearFactorGraph, Printing) {
+TEST_DISABLED(HybridNonlinearFactorGraph, Printing) {
```
varunagrawal: I'll fix this in a follow up PR. Hopefully this explains why I sometimes take longer to fix things, because I have to also get these kinds of tests passing. :)

gtsam/hybrid/tests/testHybridGaussianFactorGraph.cpp
varunagrawal: Or I would put this in the `testHybridBayesNet.cpp` file instead.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -238,20 +235,14 @@ discreteElimination(const HybridGaussianFactorGraph &factors,
     } else if (auto gmf = dynamic_pointer_cast<HybridGaussianFactor>(f)) {
       // Case where we have a HybridGaussianFactor with no continuous keys.
       // In this case, compute discrete probabilities.
-      auto logProbability =
-          [&](const GaussianFactor::shared_ptr &factor) -> double {
-        // If the factor is null, it is has been pruned hence return ∞
-        // so that the exp(-∞)=0.
-        if (!factor) return std::numeric_limits<double>::infinity();
-        return factor->error(VectorValues());
+      auto potential = [&](const auto& pair) -> double {
```
varunagrawal: Typing is not the bottleneck, right?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -238,20 +235,14 @@ discreteElimination(const HybridGaussianFactorGraph &factors,
...
+        auto [factor, scalar] = pair;
+        // If factor is null, it has been pruned, hence return potential zero
+        if (!factor) return 0.0;
+        return exp(-scalar - factor->error(kEmpty));
```
varunagrawal: This is a major bug!!!
when we don't normalize the values, if we have large factor errors (e.g. even 80), then `exp(-80) ~= 0.0` and everything collapses. This was one of the major issues I fixed last December that got things working.

Unfortunately this is not revealed in the the GTSAM tests because the sigmas are too simple but show up in the hybrid estimation code (same reason why the pruning PR didn't show any issues in GTSAM but caused experiments to fail).

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -272,75 +263,57 @@ discreteElimination(const HybridGaussianFactorGraph &factors,
...
+    }
   };
 
-  AlgebraicDecisionTree<Key> negLogProbabilities(
-      DecisionTree<Key, double>(eliminationResults, negLogProbability));
-  AlgebraicDecisionTree<Key> probabilities =
-      probabilitiesFromNegativeLogValues(negLogProbabilities);
-
-  return std::make_shared<DecisionTreeFactor>(discreteSeparator, probabilities);
+  DecisionTree<Key, double> potentials(eliminationResults, potential);
```
varunagrawal: We'll have the same bug here.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -272,75 +263,57 @@ discreteElimination(const HybridGaussianFactorGraph &factors,
...
-      // Add 2.0 term since the constant term will be premultiplied by 0.5
-      // as per the Hessian definition,
-      // and negative since we want log(k)
-      hf->constantTerm() += -2.0 * conditional->negLogConstant();
+  auto correct = [&](const auto &pair) -> GaussianFactorValuePair {
+    const auto &[conditional, factor] = pair.first;
+    const double scalar = pair.second;
+    if (conditional && factor) {
+      const double negLogK = conditional->negLogConstant();
+      return {factor, scalar - negLogK};
```
varunagrawal, line 326: Is this mathematically the same? There is no 2.0 term.

varunagrawal:
Here are my `clang-format` settings:

```json
"clang-format.executable": "/usr/local/bin/clang-format",
"clang-format.language.c.style": "Google",
"clang-format.language.cpp.style": "Google",
"clang-format.fallbackStyle": "Google",
```

gtsam/hybrid/HybridGaussianFactor.cpp
varunagrawal: This is also easy to update (use `GaussianFactorValuePair`). Doing this in a follow up PR.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: Confirming that I am running into a similar issue as before.

varunagrawal:
~~I re-added the normalization and tests that initially passed are now failing. :(~~
Nvm I was missing the scalar value.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: This is valid since the 2.0 factor is from the Hessian definition.

gtsam/hybrid/HybridGaussianFactor.cpp
varunagrawal: This has been added via the `printFactors` method.

dellaert:
> Would it be possible to break this into a separate PR for the `HybridGaussianProductFactor`?

It's possible to split the PR - but not with this version of HybridGaussianProductFactor

gtsam/hybrid/HybridGaussianConditional.cpp
dellaert: There is a .clang-format file. I don't fully understand either.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Ok, I'll fix this

dellaert: I will fix it *only* here, when the exp is taken. I'd like the original errors preserved in the factors, not shifted before it's needed. I can explain in person if you vehemently object.

dellaert:
> Here are my `clang-format` settings:
> 
> ```json
> "clang-format.executable": "/usr/local/bin/clang-format",
> "clang-format.language.c.style": "Google",
> "clang-format.language.cpp.style": "Google",
> "clang-format.fallbackStyle": "Google",
> ```

This file is *checked in* in:
```
BasedOnStyle: Google

BinPackArguments: false
BinPackParameters: false
ColumnLimit: 100
DerivePointerAlignment: false
IncludeBlocks: Preserve
PointerAlignment: Left
```

Maybe we can just change that file to not have any modifications on top of Google style.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: OK, for this case and case below, I added a function 
```c++
/// Take negative log-values, shift them so that the minimum value is 0, and
/// then exponentiate to create a DecisionTreeFactor (not normalized yet!).
static DecisionTreeFactor::shared_ptr DiscreteFactorFromErrors(
    const DiscreteKeys &discreteKeys,
    const AlgebraicDecisionTree<Key> &errors) {
  double min_log = errors.min();
  AlgebraicDecisionTree<Key> potentials = DecisionTree<Key, double>(
      errors, [&min_log](const double x) { return exp(-(x - min_log)); });
  return std::make_shared<DecisionTreeFactor>(discreteKeys, potentials);
}
```
Please note I realized it is *not* necessary to normalize to probabilities: it is still just a *factor* that we need, not a *conditional*. That was an error on my part before - although it only cost some computation, nothing else.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Resolved with same function.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: I believe these are the only two places.

Also I should explain: the shifting is what we need, not necessarily the normalization by the sum, so I agree with this assessment.

I was going to make a PR once I figured out the Windows error, so I'll just rebase your changes on to my branch and run the experiments again.

dellaert:
Pushed new changes. I have no idea on how to fix windows. Because `istream` is involved it could be serialization?

gtsam/hybrid/HybridGaussianConditional.cpp
varunagrawal: Are you using VSCode? You may have to set the settings for the VS Code version which can be done in the settings.
I use the [Clang-Format extension](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format).

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: Also I would like to hear the explanation for my own curiosity. I have a suspicion we may need to normalize somewhere.

varunagrawal:
> Pushed new changes. I have no idea on how to fix windows. Because `istream` is involved it could be serialization?

I figured out a hack to fix the Windows CI. It is in #1866.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Actually, it might still need to happen in the discretePosterior methods. It seems what we need is a constructor in DecisionTreeFactor. factors are not normalized, so for discretePosterior it should be a constructor in DiscreteConditional (which is normalized).

varunagrawal: I can do it and piggyback it off of #1867

varunagrawal: Apparently we can't create a DiscreteConditional since we don't have the cardinality information of the Keys from an AlgebraicDecisionTree. Thoughts?

varunagrawal: Also `discretePosterior` already normalizes due to the `p / p.sum()`.

dellaert:
@varunagrawal Can we now merge this? Do your experiments work?

varunagrawal:
Yes, but I wanted to fix all the pending TODOs. I'll do them in the other PR.

varunagrawal (approved):
LGTM

PR #1864, opened 2024-10-08
Fix machine precision bug in DoglegOptimizerImpl::ComputeBlend

### LINKED ISSUE #1861: Edge Case Bug in iSAM2 when using DogLeg Optimization
## Description

There is an edge case in the implementation of `DoglegOptimizerImpl::ComputeBlend` that can cause the incorrect solution to be returned due to machine-precision level error in the intermediate values in this function. 

Compute blend solves a quadratic equation to determine the proportion  ($\alpha$) used in linear interpolation of the Gauss-Newton Step and the Gradient Step. This proportion is restricted to be in the range $[0,1]$ (inclusive), however, there is a special case that can cause the incorrect solution to the quadratic to be returned and in turn an invalid interpolation to be preformed. Specifically when the Trust Region is equal to the magnitude of the Gauss Newton Step ($\Delta = \delta_{x_n}$)  

The origin of the issue is in the logic that decides which solution of the quadratic to return:
https://github.com/borglab/gtsam/blob/23f4282b08394dfc510106b98965ab0af044a999/gtsam/nonlinear/DoglegOptimizerImpl.cpp#L71-L77 

Due to machine precision it is possible for the following to occur `tau1 = 1.000000000000000002`  where `tau1` is off of 1 by a very very small amount and therefore the expression in line 71 above evaluates to false causing `tau2` to be returned. This is not caught since GTSAM compiles in NDEBUG mode by default so the assert expression is never run. Additionally, this can cause the DogLeg Optimizer to take a optimization step in the completely wrong direction and significantly impact the performance of iSAM2.

This issue can also vary by machine which I outline below in the "Environment" section.

## Steps  to reproduce

The following minimal example reproduces this error under certain environments (see environment section below).
```
#include <gtsam/base/Matrix.h>
#include <gtsam/nonlinear/DoglegOptimizerImpl.h>

using namespace gtsam;

int main() {
  std::map<Key, size_t> dims;
  dims[0] = 3;
  VectorValues n(Vector3(0.3233546123, -0.2133456123, 0.3664345632), dims);
  VectorValues d(Vector3(0.0023456342, -0.04535687, 0.087345661212), dims);
  DoglegOptimizerImpl::ComputeBlend(n.norm(), d, n, true);
}
```
And is built with the following simple CMake
```
project(imesa_experiments CXX C)
cmake_minimum_required(VERSION 2.8.3)
set(CMAKE_CXX_STANDARD 17)

# Dependency - GTSAM
find_package(GTSAM REQUIRED)
include_directories(${GTSAM_INCLUDE_DIR})

add_executable(test "test.cpp")
target_link_libraries(test gtsam)
```

Which produces the output: `In blend region with fraction -1.31316 of Newton's method point`

## Expected behavior

For this edge case I would expect the correct solution of `1.0` to be returned.

## Environment

For all evaluation I used the following in-common:
GTSAM Version - 4.2.0 (Default build arguments)
C++ Version - C++17

However, I do see different behavior based on the OS version that I use namely when compiling and running on Ubuntu 20.04 (default compiler = gcc9) I see the issue, however when running on Ubuntu 22.04  (default compiler = gcc11) I do not see this issue.

My best guess is that the true issue is in the change of compiler that is used by default on the two different OS versions but I have not thought of a satisfactory explanation for "WHY?" a compiler change would cause this behavior. I think regardless though this method should be made robust to floating-point precision errors.

## Additional information

I am more than happy to do the leg work to add tests + fix the bug, but I wanted to post an issue first to get input on how to fix given that it can be a platform + environment dependent isssue.

As another side note: This function is also not resilient to a trust region less than the gradient step size. However, this is a more pathological issue since you should never need to compute the blend for that case.

-Thanks! -Dan

DanMcGann:
Fixes #1861

This commit fixes a bug that could cause the incorrect solution to be returned from `DoglegOptimizerImpl::ComputeBlend` that is documented in Issue #1861.

The fix incorporates machine precision epsilon using `std::numeric_limits` into the bounds checks that determine which solution is returned from the quadratic documented in `doc/trustregion.pdf`. This should ensure that the correct solution is returned regardless of machine prevision errors.

This PR also adds a test for the edgecases that are know to cause the bug on some machines.

### DIFF
--- a/gtsam/nonlinear/DoglegOptimizerImpl.cpp
+++ b/gtsam/nonlinear/DoglegOptimizerImpl.cpp
@@ -67,12 +67,14 @@ VectorValues DoglegOptimizerImpl::ComputeBlend(double delta, const VectorValues&
   double tau1 = (-b + sqrt_b_m4ac) / (2.*a);
   double tau2 = (-b - sqrt_b_m4ac) / (2.*a);
 
+  // Determine correct solution accounting for machine precision
   double tau;
-  if(0.0 <= tau1 && tau1 <= 1.0) {
-    assert(!(0.0 <= tau2 && tau2 <= 1.0));
+  const double eps = std::numeric_limits<double>::epsilon();
+  if(-eps <= tau1 && tau1 <= 1.0 + eps) {
+    assert(!(-eps <= tau2 && tau2 <= 1.0 + eps));
     tau = tau1;
   } else {
-    assert(0.0 <= tau2 && tau2 <= 1.0);
+    assert(-eps <= tau2 && tau2 <= 1.0 + eps);
     tau = tau2;
   }
 
--- a/tests/testDoglegOptimizer.cpp
+++ b/tests/testDoglegOptimizer.cpp
@@ -73,6 +73,24 @@ TEST(DoglegOptimizer, ComputeBlend) {
 }
 
 /* ************************************************************************* */
+TEST(DoglegOptimizer, ComputeBlendEdgeCases) {
+  // Test Derived from Issue #1861
+  // Evaluate ComputeBlend Behavior for edge cases where the trust region
+  // is equal in size to that of the newton step or the gradient step.
+
+  // Simulated Newton (n) and Gradient Descent (u) step vectors w/ ||n|| > ||u||
+  VectorValues::Dims dims;
+  dims[0] = 3;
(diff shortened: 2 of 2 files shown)


varunagrawal (approved):
LGTM. Thanks Dan!

PR #1865, opened 2024-10-08
Updates to `No Hiding` PR

varunagrawal:
This PR makes some formatting and type information changes. It doesn't change the code semantics in any way.

I also fixed the issue with the Windows CI.


dellaert:
It’s a bit weird: there is a clang-format file that we should be using, but cursor and vs code format differently still. I had already tried minimizing formatting changes manually, but I’ll approve this to reduce PR even more. Thanks!

dellaert (approved):

varunagrawal:
I still haven't figured out why the Windows CI is failing but I know it's because of the `HybridGaussianProductFactor` files. Going to continue investigating.

PR #1867, opened 2024-10-09
Improve the `DiscreteFactorFromErrors`

varunagrawal:
Added a new constructor to `AlgebraicDecisionTree` so we can reduce some boilerplate code.

Improved the docstring for `DiscreteFactorFromErrors`.

### DIFF
--- a/gtsam/discrete/AlgebraicDecisionTree.h
+++ b/gtsam/discrete/AlgebraicDecisionTree.h
@@ -234,12 +234,9 @@ namespace gtsam {
      * @brief Helper method to perform normalization such that all leaves in the
      * tree sum to 1
      *
-     * @param sum
      * @return AlgebraicDecisionTree
      */
-    AlgebraicDecisionTree normalize(double sum) const {
-      return this->apply([&sum](const double& x) { return x / sum; });
-    }
+    AlgebraicDecisionTree normalize() const { return (*this) / this->sum(); }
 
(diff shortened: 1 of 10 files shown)


gtsam/discrete/AlgebraicDecisionTree.h
```diff
@@ -182,6 +182,21 @@ namespace gtsam {
...
+     *
+     * @tparam X The type of the leaf of the original DecisionTree
+     * @tparam Func Type signature of functional `f`.
+     * @param other The original DecisionTree from which the
+     * AlgbraicDecisionTree is constructed.
+     * @param f Functional used to operate on
+     * the leaves of the input DecisionTree.
+     */
+    template <typename X, typename Func>
+    AlgebraicDecisionTree(const DecisionTree<L, X>& other, Func f)
```
dellaert, line 197: An ADT stores doubles, so ideally Func should be enforced to return a double. We can merge as is but any signature that does not will yield strange compilation errors.

dellaert (approved):
Small nit on Func, but we can merge as is.

dellaert:
Please merge after merging no_hiding so this targets develop

gtsam/discrete/AlgebraicDecisionTree.h
varunagrawal: Hmm I am not sure I know how to enforce that. :( I'll check later.

PR #1868, opened 2024-10-10
Record cardinalities in DecisionTree

varunagrawal:
This PR updates `DecisionTree` to store the cardinalities of the nodes so we can recover them easily later.

Note that I had to settle for strange naming conventions to avoid collisions when a class inherits from both `DiscreteFactor` and `DecisionTree`.


dellaert (changes requested):
Hey Varun, for a variety of reasons I do not want us to do this. Let's talk.

varunagrawal:
> Hey Varun, for a variety of reasons I do not want us to do this. Let's talk.

Sure thing.

dellaert:
Closing this for now.

PR #1870, opened 2024-10-11
precompiled_header.h: remove chartTesting.h

### LINKED ISSUE #1788: Cannot build with msvc and C++20
## Description

I'm trying to build GTSAM library using conan 1.62.0 using C++20. Compilation fails on `gtsam/base/chartTesting.h` file.
There is a function `testDefaultChart` used in `CHECK_CHART_CONCEPT` macro. 

To be more specific, compilation fails on `gtsam::DefaultChart<T>`. But what is weird for me that none of above types (function, macro, `gtsam::DefaultChart<T>` type) is not used anywhere.  The error was exactly the same as in [this comment](https://github.com/borglab/gtsam/pull/1502#issuecomment-1501315692).

When I remove this file, everything compiles. I think that this file is leftover after some cleaning and it should be removed in future

## Steps to reproduce

1. Build GTSAM with msvc, with C++20, build tools v143.

## Expected behavior

Successful build

## Environment
Gtsam (I tried 4.0.2, 4.0.3, 4.2)
Windows
MSVC; v143
C++20
conan 1.62.0
cmake 3.27.1

valgur:
The [chartTesting.h](https://github.com/borglab/gtsam/blob/develop/gtsam/base/chartTesting.h) header is no longer used anywhere in the codebase, but it causes a compilation error on MSVC when building with C++20 or newer:
```
C:\Users\martin\.conan2\p\b\gtsam83ef710337281\b\src\gtsam\base\chartTesting.h(42,3): error C3878: syntax error: unexpected token ')' following 'expression_statement' [C:\Users\martin\.conan2\p\b\gtsam83ef710337281\b\build\gtsam 
\gtsam.vcxproj]
  (compiling source file '../../src/gtsam/precompiled_header.cpp')
```

This PR removes it from the list of precompiled headers as a fix. Should the chartTesting.h header be removed altogether, perhaps?

Fixes #1788.

### DIFF
--- a/gtsam/precompiled_header.h
+++ b/gtsam/precompiled_header.h
@@ -23,7 +23,6 @@
 // numericalDerivative.h : includes things in linear, nonlinear :-(
 // testLie.h: includes numericalDerivative
 #include <gtsam/base/Lie.h>
-#include <gtsam/base/chartTesting.h>
 #include <gtsam/base/cholesky.h>
 #include <gtsam/base/concepts.h>
 #include <gtsam/base/ConcurrentMap.h>


ProfFan (approved):
Thank you!

PR #1875, opened 2024-10-15
Faster DecisionTree

dellaert:
Several improvements yield maybe 5% speedup in hybrid (I thought it'd be more, mileage may vary):
- Make several methods static
- Add a convertFrom that does not do "Shannon expansion" as labels are not changed
- Make `push_back` take an r-value and use std::move where used


varunagrawal (approved):
Cool!

PR #1876, opened 2024-10-15
Improve Linear Conjugate Gradient Method

varunagrawal:
1. Improved `PCGSolverParameters` to not expose raw shared pointer.
2. Improved access specifiers.
3. Moved `NonlinearConjuagateGradient` test from `tests/` to `gtsam/nonlinear/tests`.
4. Improved docstrings.


gtsam/linear/ConjugateGradientSolver.h
```diff
@@ -24,20 +24,22 @@
...
+ * Parameters for the Conjugate Gradient method
  */
 class GTSAM_EXPORT ConjugateGradientParameters : public IterativeOptimizationParameters {
 
-public:
+ public:
   typedef IterativeOptimizationParameters Base;
   typedef std::shared_ptr<ConjugateGradientParameters> shared_ptr;
 
+ protected:
```
dellaert: Hmmm. If anything, I’d make this a struct and remove the underscore so all fields can just be set and read easily. And deprecate the getters/setters, esp since we can now support this in python.

gtsam/linear/PCGSolver.h
```diff
@@ -31,29 +31,33 @@ class VectorValues;
 struct PreconditionerParameters;
 
 /**
- * Parameters for PCG
+ * Parameters for Preconditioned Conjugate Gradient solver.
  */
 struct GTSAM_EXPORT PCGSolverParameters: public ConjugateGradientParameters {
-public:
+ public:
```
dellaert: Same. These are *structs*, make everything public, remove getters/setters.

gtsam/linear/PCGSolver.h
```diff
@@ -87,16 +91,16 @@ class GTSAM_EXPORT PCGSolver: public IterativeSolver {
...
-
   const GaussianFactorGraph &gfg_;
   const Preconditioner &preconditioner_;
-  const KeyInfo &keyInfo_;
-  const std::map<Key, Vector> &lambda_;
+  KeyInfo keyInfo_;
+  std::map<Key, Vector> lambda_;
+
+ public:
+  GaussianFactorGraphSystem(const GaussianFactorGraph &gfg,
```
dellaert, line 89: This change is good.

gtsam/nonlinear/NonlinearConjugateGradientOptimizer.h
```diff
@@ -40,10 +38,10 @@ class GTSAM_EXPORT NonlinearConjugateGradientOptimizer : public NonlinearOptimiz
     System(const NonlinearFactorGraph &graph) :
         graph_(graph) {
     }
-    double error(const State &state) const;
-    Gradient gradient(const State &state) const;
-    State advance(const State &current, const double alpha,
-        const Gradient &g) const;
+    double error(const Values &state) const;
```
dellaert: I disagree with this change. Other systems have different “State” and I would like to keep the naming uniform.

gtsam/nonlinear/NonlinearConjugateGradientOptimizer.h
```diff
@@ -40,10 +38,10 @@ class GTSAM_EXPORT NonlinearConjugateGradientOptimizer : public NonlinearOptimiz
...
+    VectorValues gradient(const Values &state) const;
```
dellaert: Same comment for Gradient

tests/testPreconditioner.cpp
```diff
@@ -60,13 +60,15 @@ TEST( PCGsolver, verySimpleLinearSystem) {
...
-  pcg->preconditioner_ = std::make_shared<gtsam::DummyPreconditionerParameters>();
+  pcg->setPreconditionerParams(
+      std::make_shared<gtsam::DummyPreconditionerParameters>());
   VectorValues deltaPCGDummy = PCGSolver(*pcg).optimize(simpleGFG);
   EXPECT(assert_equal(exactSolution, deltaPCGDummy, 1e-7));
   //deltaPCGDummy.print("PCG Dummy");
 
   // With Block-Jacobi preconditioner
-  pcg->preconditioner_ = std::make_shared<gtsam::BlockJacobiPreconditionerParameters>();
+  pcg->setPreconditionerParams(
```
dellaert: As I said, direct access to parameter *structs* is simpler, less clutter

dellaert (changes requested):
I actually disagree with most changes :-( I’d go the opposite way with parameters: struct all the way, reduce clutter of setters and getters. They might be needed in wrappers which is the only reason they should exist.

gtsam/nonlinear/NonlinearConjugateGradientOptimizer.h
varunagrawal: Okay, but I propose moving the `System` class outside of `NonlinearConjugateGradientOptimizer` so they are decoupled. I am not sure how you would update `System` in this case.

dellaert:
I don’t know why you need to change anything. What’s the driver here?
I’ll try to track down some code that uses a different “system” and add
it to the unit tests but before we have that in place I don’t see a
reason to make any changes.

On October 16, 2024, GitHub ***@***.***> wrote:
> @varunagrawal commented on this pull request.
>

> In gtsam/nonlinear/NonlinearConjugateGradientOptimizer.h
> <https://github.com/borglab/gtsam/pull/1876#discussion_r1802310616>:
>
> > @@ -40,10 +38,10 @@ class GTSAM_EXPORT
> NonlinearConjugateGradientOptimizer : public NonlinearOptimiz
> System(const NonlinearFactorGraph &graph) : graph_(graph) { } - double
> error(const State &state) const; - Gradient gradient(const State
> &state) const; - State advance(const State &current, const double
> alpha, - const Gradient &g) const; + double error(const Values &state)
> const;
> Okay, but I propose moving the System class outside of
> NonlinearConjugateGradientOptimizer so they are decoupled. I am not
> sure how you would update System in this case.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/borglab/gtsam/pull/1876#discussion_r1802310616>, or
unsubscribe <https://github.com/notifications/unsubscribe-
auth/ACQHGSJYI3OJDGIYRHHMDWDZ3XP4PAVCNFSM6AAAAABP72P4GWVHI2DSMVQWIX3LMV43YUDVNRWFEZLROVSXG5CSMV3GSZLXHMZDGNZRGEZTENJVHE>.
You are receiving this because you commented.Message ID:
***@***.***>

varunagrawal:
So my primary idea was to add better unit test coverage. That way we could verify things like convergence rate etc, and so I wanted to simplify code so we could have comprehensive unit tests.

gtsam/nonlinear/NonlinearConjugateGradientOptimizer.cpp
```diff
@@ -16,11 +16,11 @@
  * @date   Jun 11, 2012
  */
 
-#include <gtsam/nonlinear/NonlinearConjugateGradientOptimizer.h>
-#include <gtsam/nonlinear/internal/NonlinearOptimizerState.h>
-#include <gtsam/nonlinear/Values.h>
 #include <gtsam/linear/GaussianFactorGraph.h>
 #include <gtsam/linear/VectorValues.h>
+#include <gtsam/nonlinear/NonlinearConjugateGradientOptimizer.h>
```
varunagrawal, line 21: This whole file is just formatting changes.

gtsam/nonlinear/NonlinearConjugateGradientOptimizer.h
```diff
@@ -24,89 +24,80 @@
 namespace gtsam {
 
 /**  An implementation of the nonlinear CG method using the template below */
-class GTSAM_EXPORT NonlinearConjugateGradientOptimizer : public NonlinearOptimizer {
-
+class GTSAM_EXPORT NonlinearConjugateGradientOptimizer
```
varunagrawal, line 27: This whole file is also simply formatting changes.

dellaert (approved):
Cool

PR #1877, opened 2024-10-15
Refactor `NonlinearConjugateGradientOptimizer`

varunagrawal:
1. ~~Removed templates.~~
2. ~~Removed the `System` class so we only need a factor graph.~~
3. Formatting


dellaert (changes requested):
This needed a design review before making a drastic change like this. There are reasons why System exists. Please talk to me first.

varunagrawal:
Gotcha. It just seemed strange that System was inside the optimizer rather than being passed as an argument. It made it seem like extensibility was not easy in that particular case.
Let's chat about this.

varunagrawal:
Merging this to parent PR since this is just formatting changes.

PR #1878, opened 2024-10-15
Additional Nonlinear CG Direction Methods

varunagrawal:
Added other direction methods for CG beyond Polak-Ribiere.

### DIFF
--- a/gtsam/nonlinear/NonlinearConjugateGradientOptimizer.cpp
+++ b/gtsam/nonlinear/NonlinearConjugateGradientOptimizer.cpp
@@ -28,46 +28,6 @@ namespace gtsam {
 
 typedef internal::NonlinearOptimizerState State;
 
-/* ************************************************************************* */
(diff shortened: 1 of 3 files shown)


dellaert:
I’ll look at this after other PRs have been addressed.

dellaert:
Could you also template these in the same vein as Polak-Ribiere? The idea -as I remember it-  is that CG does not need to know whether the underlying "system" is actually a dense, sparse, graph, or other system that has the right operators. I do regret that we did not add unit tests in that vein.

varunagrawal:
> Could you also template these in the same vein as Polak-Ribiere? The idea -as I remember it- is that CG does not need to know whether the underlying "system" is actually a dense, sparse, graph, or other system that has the right operators. I do regret that we did not add unit tests in that vein.

Ah good catch. I'll do that right now.

varunagrawal:
I still need to add some tests. Will do that later today.

dellaert (approved):
Cool!

PR #1881, opened 2024-10-22
Significant speedup

dellaert:
# Speedups across the board
## In DT:
- speed up DT conversion if labels stay 

## In HGC
- remove double storage: just one tree in HGF base class
- special constructor straight from this base class storage

## In HGFG
- prune empty graphs on the fly
- new `Result` struct to eliminate double compute of negLogK
- use new HGC constructor

## Extra
- DT method `split` : unused for now but could speed up if split could build only two trees instead of three; in that case, we could try building ResultTree as two trees instead, and one of them could be used to create the HGCs directly

### DIFF
--- a/gtsam/discrete/DecisionTree.h
+++ b/gtsam/discrete/DecisionTree.h
@@ -156,10 +156,10 @@ namespace gtsam {
     template <typename It, typename ValueIt>
     static NodePtr build(It begin, It end, ValueIt beginY, ValueIt endY);
 
-    /** Internal helper function to create from
-     * keys, cardinalities, and Y values.
-     * Calls `build` which builds thetree bottom-up,
-     * before we prune in a top-down fashion.
+    /**
+     * Internal helper function to create a tree from keys, cardinalities, and Y
+     * values. Calls `build` which builds the tree bottom-up, before we prune in
+     * a top-down fashion.
      */
     template <typename It, typename ValueIt>
     static NodePtr create(It begin, It end, ValueIt beginY, ValueIt endY);
@@ -239,7 +239,7 @@ namespace gtsam {
     DecisionTree(const DecisionTree<L, X>& other, Func Y_of_X);
 
     /**
-     * @brief Convert from a different value type X to value type Y, also transate
+     * @brief Convert from a different value type X to value type Y, also translate
      * labels via map from type M to L.
      *
      * @tparam M Previous label type.
@@ -405,6 +405,18 @@ namespace gtsam {
     std::string dot(const LabelFormatter& labelFormatter,
                     const ValueFormatter& valueFormatter,
                     bool showZero = true) const;
+
+    /**
+     * @brief Convert into two trees with value types A and B.
+     *
+     * @tparam A First new value type.
+     * @tparam B Second new value type.
+     * @param AB_of_Y Functor to convert from type X to std::pair<A, B>.
+     * @return A pair of DecisionTrees with value types A and B respectively.
+     */
+    template <typename A, typename B>
+    std::pair<DecisionTree<L, A>, DecisionTree<L, B>> split(
+        std::function<std::pair<A, B>(const Y&)> AB_of_Y);
 
     /// @name Advanced Interface
     /// @{
--- a/gtsam/discrete/DecisionTree.h
+++ b/gtsam/discrete/DecisionTree.h
@@ -416,7 +416,7 @@ namespace gtsam {
      */
     template <typename A, typename B>
     std::pair<DecisionTree<L, A>, DecisionTree<L, B>> split(
-        std::function<std::pair<A, B>(const Y&)> AB_of_Y);
+        std::function<std::pair<A, B>(const Y&)> AB_of_Y) const;
 
     /// @name Advanced Interface
     /// @{
--- a/gtsam/discrete/DecisionTree.h
+++ b/gtsam/discrete/DecisionTree.h
@@ -85,7 +85,7 @@ namespace gtsam {
 
     /** ------------------------ Node base class --------------------------- */
     struct Node {
-      using Ptr = std::shared_ptr<const Node>;
+      using Ptr = std::shared_ptr<Node>;
 
 #ifdef DT_DEBUG_MEMORY
       static int nrNodes;
@@ -227,6 +227,15 @@ namespace gtsam {
     /** Create DecisionTree from two others */
     DecisionTree(const L& label, const DecisionTree& f0,
                  const DecisionTree& f1);
+
+    /**
+     * @brief Move constructor for DecisionTree. Very efficient as does not
+     * allocate anything, just changes in-place. But `other` is consumed.
+     *
+     * @param op The unary operation to apply to the moved DecisionTree.
+     * @param other The DecisionTree to move from, will be empty afterwards.
+     */
+    DecisionTree(const Unary& op, DecisionTree&& other) noexcept;
 
     /**
      * @brief Convert from a different value type.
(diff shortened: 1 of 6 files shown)


dellaert:
@varunagrawal Hoping to get a quick review so I can cut an alpha version of GTSAM soon.
Before I can do that, I'd also need to ascertain whether memory leak is in GTSAM proper or in your application code. Any information on that?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -363,34 +367,31 @@ HybridGaussianFactorGraph::eliminate(const Ordering &keys) const {
   // any difference in noise models used.
   HybridGaussianProductFactor productFactor = collectProductFactor();
 
-  // Convert factor graphs with a nullptr to an empty factor graph.
-  // This is done after assembly since it is non-trivial to keep track of which
-  // FG has a nullptr as we're looping over the factors.
-  auto prunedProductFactor = productFactor.removeEmpty();
+  auto isNull = [](const GaussianFactor::shared_ptr& ptr) { return !ptr; };
```
varunagrawal: We should keep the docstring. I remember that this was a point of confusion when we came back to this code after a while.

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -288,28 +310,39 @@ HybridGaussianConditional::shared_ptr HybridGaussianConditional::prune(
...
 /* *******************************************************************************/
-double HybridGaussianConditional::evaluate(const HybridValues &values) const {
-  auto conditional = conditionals_(values.discrete());
-  return conditional->evaluate(values.continuous());
+double HybridGaussianConditional::evaluate(const HybridValues& values) const {
+  auto [factor, _] = factors()(values.discrete());
+  if (auto conditional = std::dynamic_pointer_cast<GaussianConditional>(factor))
+    return conditional->evaluate(values.continuous());
+  else
+    throw std::logic_error(
```
varunagrawal: I'm wondering if there is a better way to reduce the duplication of this and logProbability? Probably by passing in a functional?
E.g.
```cpp
double checkAndRun(const HybridValues& values, std::function<double(const GaussianConditional::shared_ptr& conditional, const HybridValues& values)>& functional) {
  auto [factor, _] = factors()(values.discrete());
  if (auto conditional = std::dynamic_pointer_cast<GaussianConditional>(factor))
    return functional(conditional, values);
  else
    throw std::logic_error(
        "A HybridGaussianConditional unexpectedly contained a non-conditional");
}
```

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -288,28 +310,39 @@ HybridGaussianConditional::shared_ptr HybridGaussianConditional::prune(
...
+      return {nullptr, std::numeric_limits<double>::infinity()};
+    else
+      return pair;
   };
 
-  auto pruned_conditionals = conditionals_.apply(pruner);
-  return std::make_shared<HybridGaussianConditional>(discreteKeys(),
-                                                     pruned_conditionals);
+  FactorValuePairs prunedConditionals = factors().apply(pruner);
+  return std::shared_ptr<HybridGaussianConditional>(
```
varunagrawal: Why not `make_shared` over here? The constructor is being called in both cases.

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -288,28 +310,39 @@ HybridGaussianConditional::shared_ptr HybridGaussianConditional::prune(
...
   // Check the max value for every combination of our keys.
   // If the max value is 0.0, we can prune the corresponding conditional.
-  auto pruner = [&](const Assignment<Key> &choices,
-                    const GaussianConditional::shared_ptr &conditional)
-      -> GaussianConditional::shared_ptr {
-    return (max->evaluate(choices) == 0.0) ? nullptr : conditional;
+  auto pruner =
+      [&](const Assignment<Key> &choices,
+          const GaussianFactorValuePair &pair) -> GaussianFactorValuePair {
+    if (max->evaluate(choices) == 0.0)
```
varunagrawal, line 323: This could still be a ternary condition

gtsam/hybrid/HybridGaussianConditional.cpp
```diff
@@ -254,12 +270,18 @@ std::shared_ptr<HybridGaussianFactor> HybridGaussianConditional::likelihood(
...
-        const auto likelihood_m = conditional->likelihood(given);
-        const double Cgm_Kgcm = conditional->negLogConstant() - negLogConstant_;
-        return {likelihood_m, Cgm_Kgcm};
+      factors(),
+      [&](const GaussianFactorValuePair &pair) -> GaussianFactorValuePair {
+        if (auto conditional =
+                std::dynamic_pointer_cast<GaussianConditional>(pair.first)) {
+          const auto likelihood_m = conditional->likelihood(given);
+          // scalar is already correct.
+          assert(pair.second ==
```
varunagrawal: Maybe we can remove this assert and save some performance?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -57,10 +57,20 @@ template class EliminateableFactorGraph<HybridGaussianFactorGraph>;
 
 using std::dynamic_pointer_cast;
 using OrphanWrapper = BayesTreeOrphanWrapper<HybridBayesTree::Clique>;
-using Result =
-    std::pair<std::shared_ptr<GaussianConditional>, GaussianFactor::shared_ptr>;
-using ResultValuePair = std::pair<Result, double>;
-using ResultTree = DecisionTree<Key, ResultValuePair>;
+
+/// Result from elimination.
+struct Result {
```
varunagrawal, line 61: Neat!

gtsam/discrete/DecisionTree.h
```diff
@@ -406,6 +415,18 @@ namespace gtsam {
...
+    /**
+     * @brief Convert into two trees with value types A and B.
+     *
+     * @tparam A First new value type.
+     * @tparam B Second new value type.
+     * @param AB_of_Y Functor to convert from type X to std::pair<A, B>.
+     * @return A pair of DecisionTrees with value types A and B respectively.
+     */
+    template <typename A, typename B>
+    std::pair<DecisionTree<L, A>, DecisionTree<L, B>> split(
```
varunagrawal, line 427: Not sure where this is used. Could probably have been another PR.

varunagrawal (approved):
Couple of comments but LGTM.
Removing the conditionals duplication was on my TODO list (planned to do it after figuring out memory issues) so this is great!

gtsam/discrete/DecisionTree.h
varunagrawal: Nevermind, read the last comment of the PR description again. Could still have been another PR 😅

gtsam/hybrid/HybridGaussianConditional.cpp
dellaert: Done

gtsam/hybrid/HybridGaussianConditional.cpp
dellaert: I tried but {} is issue

gtsam/hybrid/HybridGaussianConditional.cpp
dellaert: Done

gtsam/hybrid/HybridGaussianConditional.cpp
dellaert: For another day...

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Good point. Done.

gtsam/hybrid/HybridGaussianConditional.cpp
varunagrawal: I already worked on it so pushed the changes.

PR #1883, opened 2024-10-23
Some Hybrid iSAM cleanup

dellaert:
- some naming conventions and spelling
- Made a more robust interface to Switching
- changed HybridGaussianISAM tests to reuse update graphs
- kept HybridNonlinearISAM tests the same but modernized them a bit
- in both cases, kept some regression but made tests have real teeth
- Added chain factors in testHybridEstimation

PS, ISAM test in latter produces BayesTrees like this, which seems right:

image

### DIFF
--- a/gtsam/hybrid/HybridNonlinearISAM.cpp
+++ b/gtsam/hybrid/HybridNonlinearISAM.cpp
@@ -39,8 +39,8 @@ void HybridNonlinearISAM::update(const HybridNonlinearFactorGraph& newFactors,
   if (newFactors.size() > 0) {
     // Reorder and relinearize every reorderInterval updates
     if (reorderInterval_ > 0 && ++reorderCounter_ >= reorderInterval_) {
-      // TODO(Varun) Relinearization doesn't take into account pruning
-      reorder_relinearize();
+      // TODO(Varun) Re-linearization doesn't take into account pruning
+      reorderRelinearize();
       reorderCounter_ = 0;
     }
 
@@ -60,7 +60,7 @@ void HybridNonlinearISAM::update(const HybridNonlinearFactorGraph& newFactors,
 }
 
 /* ************************************************************************* */
-void HybridNonlinearISAM::reorder_relinearize() {
+void HybridNonlinearISAM::reorderRelinearize() {
   if (factors_.size() > 0) {
     // Obtain the new linearization point
     const Values newLinPoint = estimate();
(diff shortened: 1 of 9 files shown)


gtsam/hybrid/tests/Switching.h
```diff
@@ -136,7 +145,7 @@ struct Switching {
    */
   Switching(size_t K, double between_sigma = 1.0, double prior_sigma = 0.1,
             std::vector<double> measurements = {},
-            std::string discrete_transition_prob = "1/2 3/2")
+            std::string transitionProbabilityTable = "1/2 3/2")
```
varunagrawal, line 148: I thought variables were snake case and functions were camel case.

gtsam/hybrid/tests/testHybridGaussianISAM.cpp
```diff
@@ -227,7 +180,7 @@ TEST(HybridGaussianElimination, Approx_inference) {
   }
 
   // Now we calculate the actual factors using full elimination
-  const auto [unprunedHybridBayesTree, unprunedRemainingGraph] =
+  const auto [unPrunedHybridBayesTree, unPrunedRemainingGraph] =
```
varunagrawal, line 183: Nitpick: Isn't unpruned one word?

Source: https://www.oed.com/dictionary/unpruned_adj1?tl=true

varunagrawal (approved):
2 comments but LGTM

gtsam/hybrid/tests/Switching.h
dellaert: in google style, yes, but historically not in GTSAM.

gtsam/hybrid/tests/testHybridGaussianISAM.cpp
dellaert: eh. My code spellchecker complained. Will leave as is as otherwise another CI :-/

varunagrawal: I can update this in a follow up. :)

gtsam/hybrid/HybridNonlinearISAM.cpp
```diff
@@ -39,8 +39,8 @@ void HybridNonlinearISAM::update(const HybridNonlinearFactorGraph& newFactors,
   if (newFactors.size() > 0) {
     // Reorder and relinearize every reorderInterval updates
     if (reorderInterval_ > 0 && ++reorderCounter_ >= reorderInterval_) {
-      // TODO(Varun) Relinearization doesn't take into account pruning
-      reorder_relinearize();
+      // TODO(Varun) Re-linearization doesn't take into account pruning
+      reorderRelinearize();
```
varunagrawal, line 43: PS this name came from `NonlinearISAM` (the continuous only version).

varunagrawal: It's `NonlinearISAM::reorder_relinearize` there which is why we kept the same name.

PR #1884, opened 2024-10-24
Two fundamental matrix classes

dellaert:
## Overview

Two F matrix classes for GTSFM:
- a general class based on [Sweeney15] F = USV with S=diag(1,s,0)
- a "simple" class based on E and two unknown focal lengths (and known principal points)
- a TripleF class that can do epipolar transfer

The idea is that the latter can be used to *immediately* optimize the [Sweeney15] objective *and* optimize for focal lengths at the same time.

### Tests

Many unit tests, but key test is
```c++
  // Check that transfer works
  EXPECT(assert_equal<Point2>(p[0], triplet.transfer0(p[1], p[2]), 1e-9));
  EXPECT(assert_equal<Point2>(p[1], triplet.transfer1(p[0], p[2]), 1e-9));
  EXPECT(assert_equal<Point2>(p[2], triplet.transfer2(p[0], p[1]), 1e-9));
```

### Next PR

"Someone" should ~write a TransferFactor and~ (TransferFactor implemented using numerical derivatives [here](https://github.com/borglab/gtsam/pull/1885)) implement the derivatives for the transfer. The latter might not be super-trivial, as there is a cross product and normalization from homogeneous coordinates. It's possible that this is easier with expressions. Maybe the next PR should just try to create derivatives for calculating the epipolar lines, i.e., the 3*7 derivative of

```c++
leftK().transpose().inverse() * E_.matrix() * rightK().inverse() * p
```

This might not be as daunting as it seems, as E_.matrix() is just R*[t], and the calibrations are actually rather simple and are just functions of fa_ and fb_, respectively.

## Reference

[Sweeney15] https://sites.cs.ucsb.edu/~holl/pubs/Sweeney-2015-ICCV.pdf

### DIFF
--- a/gtsam/geometry/FundamentalMatrix.cpp
+++ b/gtsam/geometry/FundamentalMatrix.cpp
@@ -10,21 +10,20 @@
 namespace gtsam {
 
 //*************************************************************************
-Point2 FundamentalMatrix::transfer(const Matrix3& F01, const Point2& p1,
-                                   const Matrix3& F02, const Point2& p2) {
-  // Create lines in camera 0 from projections of the other two cameras
-  Vector3 line1 = F01 * Vector3(p1.x(), p1.y(), 1);
-  Vector3 line2 = F02 * Vector3(p2.x(), p2.y(), 1);
+Point2 Transfer(const Matrix3& Fca, const Point2& pa,  //
+                const Matrix3& Fcb, const Point2& pb) {
+  // Create lines in camera a from projections of the other two cameras
+  Vector3 line_a = Fca * Vector3(pa.x(), pa.y(), 1);
+  Vector3 line_b = Fcb * Vector3(pb.x(), pb.y(), 1);
 
   // Cross the lines to find the intersection point
-  Vector3 intersectionPoint = line1.cross(line2);
+  Vector3 intersectionPoint = line_a.cross(line_b);
 
   // Normalize the intersection point
   intersectionPoint /= intersectionPoint(2);
 
   return intersectionPoint.head<2>();  // Return the 2D point
 }
-
 //*************************************************************************
 GeneralFundamentalMatrix::GeneralFundamentalMatrix(const Matrix3& F) {
   // Perform SVD
--- a/gtsam/geometry/FundamentalMatrix.cpp
+++ b/gtsam/geometry/FundamentalMatrix.cpp
@@ -25,7 +25,7 @@ Point2 Transfer(const Matrix3& Fca, const Point2& pa,  //
   return intersectionPoint.head<2>();  // Return the 2D point
 }
 //*************************************************************************
-GeneralFundamentalMatrix::GeneralFundamentalMatrix(const Matrix3& F) {
+FundamentalMatrix::FundamentalMatrix(const Matrix3& F) {
   // Perform SVD
   Eigen::JacobiSVD<Matrix3> svd(F, Eigen::ComputeFullU | Eigen::ComputeFullV);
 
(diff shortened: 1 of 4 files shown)


gtsam/geometry/FundamentalMatrix.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+  // Cross the lines to find the intersection point
+  Vector3 intersectionPoint = line_a.cross(line_b);
+
+  // Normalize the intersection point
+  intersectionPoint /= intersectionPoint(2);
+
+  return intersectionPoint.head<2>();  // Return the 2D point
+}
+//*************************************************************************
```
akshay-krishnan, line 27: nit: newline before?

gtsam/geometry/FundamentalMatrix.h
```diff
@@ -0,0 +1,183 @@
...
+  /// @}
+};
+
+/**
+ * @brief Transfer projections from cameras a and b to camera c
+ *
+ * Take two fundamental matrices Fca and Fcb, and two points pa and pb, and
+ * returns the 2D point in view (c) where the epipolar lines intersect.
+ */
+GTSAM_EXPORT Point2 Transfer(const Matrix3& Fca, const Point2& pa,
```
akshay-krishnan, line 150: Might be useful to have a more verbose name here. Maybe `TransferPointsToThirdView`  ?

gtsam/geometry/FundamentalMatrix.h
```diff
@@ -0,0 +1,183 @@
...
+ public:
+  /// Default constructor
+  SimpleFundamentalMatrix()
+      : E_(), fa_(1.0), fb_(1.0), ca_(0.0, 0.0), cb_(0.0, 0.0) {}
+
+  /// Construct from essential matrix and focal lengths
+  SimpleFundamentalMatrix(const EssentialMatrix& E,  //
+                          double fa, double fb,
+                          const Point2& ca = Point2(0.0, 0.0),
+                          const Point2& cb = Point2(0.0, 0.0))
```
akshay-krishnan, line 109: so these are not in pixel space, but with respect to image center?

gtsam/geometry/FundamentalMatrix.h
dellaert: Good point, was FundamentalMatrix::Transfer before. Propose to rename `EpipolarTransfer`

gtsam/geometry/FundamentalMatrix.h
dellaert: ca and cb are the pixel centers. If you do not provide ca and cb, then your are promising to provide coordinates with respect to the pixel center, yes.

gtsam/geometry/FundamentalMatrix.cpp
akshay-krishnan: somehow I still don't see a newline here, wonder why the formatter does not fix this.

akshay-krishnan (approved):

gtsam/geometry/FundamentalMatrix.h
travisdriver: I think it would be better to have ca and cb be required inputs so users don't unknowingly make this promise

gtsam/geometry/FundamentalMatrix.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+FundamentalMatrix FundamentalMatrix::retract(const Vector& delta) const {
+  Rot3 newU = U_.retract(delta.head<3>());
+  double newS = s_ + delta(3);  // Update scalar
+  Rot3 newV = V_.retract(delta.tail<3>());
+  return FundamentalMatrix(newU, newS, newV);
+}
+
+//*************************************************************************
+Matrix3 SimpleFundamentalMatrix::leftK() const {
```
travisdriver, line 101: nit: maybe this should be `Ka()` (`Kb()`) since that's how variables from the left (right) view are referred to everywhere else in the code? E.g., `fa`, `fb`, `Fca`, etc. Don't feel super strongly about this though.

travisdriver (approved):

gtsam/geometry/FundamentalMatrix.cpp
dellaert: It's there in my code. I'll make sure it's in last PR

gtsam/geometry/FundamentalMatrix.h
dellaert: OK, will do

gtsam/geometry/FundamentalMatrix.cpp
dellaert: I like it. Also make them private

dellaert:
I'll incorporate remaining comments in follow-up PRs. Thanks!

PR #1888, opened 2024-10-28
F wrapper and bugfix

dellaert:
- ~changed internal representation of F to fix a bug when flipping U or V~
- now flip columns of U and V which always allows for $U*diag(1,s,0)*V^T$
- added wrapper definitions
- added python unit test

### DIFF
--- a/gtsam/geometry/FundamentalMatrix.cpp
+++ b/gtsam/geometry/FundamentalMatrix.cpp
@@ -26,6 +26,11 @@ Point2 EpipolarTransfer(const Matrix3& Fca, const Point2& pa,  //
 }
 
 //*************************************************************************
+FundamentalMatrix::FundamentalMatrix(const Matrix& U, double s,
+                                     const Matrix& V) {
+  initialize(U, s, V);
+}
+
 FundamentalMatrix::FundamentalMatrix(const Matrix3& F) {
   // Perform SVD
   Eigen::JacobiSVD<Matrix3> svd(F, Eigen::ComputeFullU | Eigen::ComputeFullV);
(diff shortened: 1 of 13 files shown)


dellaert:
@akshay-krishnan , @travisdriver  please look at FundamentalMatrix.h first, then rest will make sense. This PR should also make the last PR easier to review.

gtsam/geometry/FundamentalMatrix.h
```diff
@@ -15,34 +15,36 @@ namespace gtsam {
...
-   *
-   * @param U Left rotation matrix
-   * @param s Scalar parameter for the fundamental matrix
-   * @param V Right rotation matrix
+   * Initializes the FundamentalMatrix From the SVD representation
+   * U*diag(1,s,0)*V^T. It will internally convert to using SO(3).
    */
-  FundamentalMatrix(const Rot3& U, double s, const Rot3& V)
-      : U_(U), s_(s), V_(V) {}
+  FundamentalMatrix(const Matrix& U, double s, const Matrix& V);
```
dellaert: Should be Matrix3

gtsam/geometry/FundamentalMatrix.h
```diff
@@ -54,22 +56,33 @@ class GTSAM_EXPORT FundamentalMatrix {
    */
   FundamentalMatrix(const Matrix3& F);
 
+  /**
+   * @brief Construct from essential matrix and calibration matrices
+   *
+   * Initializes the FundamentalMatrix from the given essential matrix E
+   * and calibration matrices Ka and Kb.
```
dellaert: Provide equation

gtsam/geometry/FundamentalMatrix.cpp
```diff
@@ -47,28 +52,44 @@ FundamentalMatrix::FundamentalMatrix(const Matrix3& F) {
...
+  s_ = s;
+  sign_ = 1.0;
 
-  // Check if V is a reflection
-  if (V.determinant() < 0) {
-    V = -V;
-    s_ = -s_;  // Change sign of scalar if U is a reflection
+  // Check if U is a reflection and its determinant is close to -1 or 1
+  double detU = U.determinant();
+  if (std::abs(std::abs(detU) - 1.0) > 1e-9) {
```
dellaert: This is always true so I’ll kill it

gtsam/geometry/FundamentalMatrix.h
dellaert: Done

gtsam/geometry/FundamentalMatrix.cpp
dellaert: Done

akshay-krishnan (approved):
I did not look at the tests, rest LGTM!

PR #1889, opened 2024-10-29
Allow parent project to override options

spacey-sooty:
I needed to depend on gtsam without boost, but when using fetch content I couldn't override the options relating to boost usage to disable it

I fixed this by setting polity CMP0077

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,11 @@
 cmake_minimum_required(VERSION 3.0)
 
+# allow parent project to override options
+if (POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif(POLICY CMP0077)
+
+
 # Set the version number for the library
 set (GTSAM_VERSION_MAJOR 4)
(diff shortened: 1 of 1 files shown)


dellaert:
Please provide a description. What prompted this and how did you solve it?

spacey-sooty:
> Please provide a description. What prompted this and how did you solve it?

Done

CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 cmake_minimum_required(VERSION 3.0)
 
+# allow parent project to override options
```
varunagrawal, line 3: Nitpick: Would your project be the parent project or the _dependent_ project? Other than potentially updating this, your PR comment explains the need well. :)

spacey-sooty: It would be the dependent project but in this case is acting as the parent as it has to build gtsam (eg by fetchcontent)

varunagrawal (approved):
Awesome LGTM

PR #1891, opened 2024-10-29
Fix warning: ‘beta’ may be used uninitialized in this function

jlblancoc:
Simple change. Trying to make ROS builds quieter...


varunagrawal (approved):
Thank you @jlblancoc

PR #1894, opened 2024-11-04
Lots of Hybrid Improvements

varunagrawal:
1. `HybridGaussianFactorGraph::discreteFactors()` to retrieve all the discrete factors. More of a convenience method.
2. `prune` method for `HybridNonlinearFactor`.
3. Prune in `reorderRelinearize` to improve nonlinear efficiency.
4. Improve Switching tests.


gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -580,4 +580,15 @@ GaussianFactorGraph HybridGaussianFactorGraph::choose(
   return gfg;
 }
 
+/* ************************************************************************ */
+DiscreteFactorGraph HybridGaussianFactorGraph::discreteFactors() const {
+  DiscreteFactorGraph dfg;
+  for (auto &&f : factors_) {
+    auto discreteFactor = std::dynamic_pointer_cast<DiscreteFactor>(f);
+    assert(discreteFactor);
```
dellaert: I don’t understand how this can work. Not all factors are discrete, right?

dellaert (commented):
Just one question

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: Yup, which is why we have the assert, but I guess the `if(auto discreteFactor = ...)` check is better.

dellaert (approved):
OK, that makes more sense :-)

PR #1895, opened 2024-11-04
Update MacOS Images for CI

varunagrawal:
Github Actions has [deprecated macos-12](https://github.com/actions/runner-images), so upgrading to macos-13.

Additionally, adding support for Mac ARM64 processors in the CI.


dellaert (commented):
Thanks! But why omit 13 now in the python build? Could you just switch 12 to 13?

varunagrawal:
> Thanks! But why omit 13 now in the python build? Could you just switch 12 to 13?

Done.

dellaert (approved):

PR #1897, opened 2024-11-06
Add missing #if for clang

jlblancoc:
One of the remaining warnings in the ROS build farms for gtsam dev builds.


ProfFan (approved):
Thanks!

PR #1898, opened 2024-11-06
Reduce memory consumption during discrete pruning

varunagrawal:
When running pruning, especially when we have a large number of discrete variables, the code would try to collect all the values in a `std::vector`. In the case we had >20 discrete variables with cardinality 2 each, this would lead to vectors of size greater than a million, thus taking up a large amount of memory.
Once the allocated heap memory exceeded a limit, the OS would automatically kill the process.

To address this, I implemented a simple `MinHeap` class and used it to maintain at most `maxNrAssignments` values in the heap, popping smaller values as necessary.

This successfully addresses the memory issue, and on profiling with gperftools, the heap profile **drops from 25 GB to less than 1 MB**. Consequently, I have been able to run the hybrid estimator to **over 120 timesteps, up from a limit of 28**.

The estimator is still struggling once we go beyond 150 due to the large number of discrete variables at that point. This would seem like a good candidate for some sort of marginalization scheme, since the initial 50 discrete variables shouldn't be affected much.

### DIFF
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -363,13 +363,15 @@ namespace gtsam {
     /// Push value onto the heap
     void push(double x) {
       v_.push_back(x);
-      std::make_heap(v_.begin(), v_.end(), std::greater<double>{});
+      std::push_heap(v_.begin(), v_.end(), std::greater<double>{});
     }
 
     /// Push value `x`, `n` number of times.
     void push(double x, size_t n) {
-      v_.insert(v_.end(), n, x);
-      std::make_heap(v_.begin(), v_.end(), std::greater<double>{});
+      for (size_t i = 0; i < n; ++i) {
+        v_.push_back(x);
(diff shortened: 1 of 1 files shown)


dellaert (commented):
Obvious question: why not use a built-in heap data structure from the standard library? Is there not one that satisfies our needs?

varunagrawal:
> Obvious question: why not use a built-in heap data structure from the standard library? Is there not one that satisfies our needs?

I tried looking for one as my first plan of attack, but some recommend using the [std::priority_queue](https://en.cppreference.com/w/cpp/container/priority_queue), while others recommend using the [heap functions around a container](https://en.cppreference.com/w/cpp/algorithm/make_heap).
Since I only ever keep the top N values, I found the later to be efficient and useful, especially since the `std::priority_queue` also uses a `std::vector` as the container under the hood.

Additionally, I was able to customize the API to use the std::vector with heap functions (allowing for convenient printing, for example) but mimic a simple API similar to the priority_queue. Win-win.

dellaert (approved):
Great!

PR #1899, opened 2024-11-07
Bump min required cmake version to 3.9

jlblancoc:
In the top-level cmake file. It was 3.5 and recent cmake versions complain about it becoming deprecate.
Also, the wrap directory already required 3.9 anyhow.

PS: This is only an issue to users of Ubuntu 16.04 or older...


varunagrawal (approved):
I'm a fan of following deprecation warnings. Hopefully we don't get too many complaints from users on older systems.

PR #1911, opened 2024-12-01
Change the deprecated C++ enum math to constexprs

ProfFan:
anonymous enum math operations have been deprecated for many years now


dellaert:
How about changing all then?

ProfFan:
@dellaert I changed all of them

dellaert (approved):
Awesome!

PR #1912, opened 2024-12-02
Updates based on profiling

varunagrawal:
- Fixed `gperftools` linking.
- Added a new `computeThreshold` method to isolate computing the pruning threshold.
- Added a flag `pruned` to `HybridGaussianConditional` to avoid pruning an already pruned conditional.
- Extra: Exposed `Rot3::expmap` in the wrapper.

The `prune` flag helps in longer sequences, but is not the major bottleneck. That would be `EliminateDiscrete`.
I am wondering if there is a way to not prune all the way down on the `HybridBayesTree`? Seems like extra computation for no reason.


dellaert (approved):
Cool.
About question at the end: let’s be data-driven: first need to address quadratic growth, then we can assess linear growth if significant constants.

varunagrawal:
> Cool.
> About question at the end: let’s be data-driven: first need to address quadratic growth, then we can assess linear growth if significant constants.

I emailed you about the quadratic growth. I've managed to isolate the problem, but the solution needs some pondering over. :)

PR #1913, opened 2024-12-02
add error(HybridValues) to NonlinearFactor

truher:
fixes the python build


varunagrawal (approved):
Nice catch, though interesting that clang didn't complain about this.

PR #1918, opened 2024-12-06
Document disabling tests and examples in INSTALL.md

spacey-sooty:
When you have gtsam as a dependency in CMake these would get built. This doesn't make sense and increases build times, allowing them to be disabled fixes this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,10 +112,17 @@ add_subdirectory(CppUnitLite)
 add_subdirectory(gtsam)
 
 # Build Tests
-add_subdirectory(tests)
+option(BUILD_TESTS "Builds unit tests" ON)
+if (BUILD_TESTS)
+    add_subdirectory(tests)
+endif()
+
 
 # Build examples
-add_subdirectory(examples)
+option(BUILD_EXAMPLES "Builds examples" ON)
+if (BUILD_EXAMPLES)
+    add_subdirectory(examples)
+endif()
 
 # Build timing
 add_subdirectory(timing)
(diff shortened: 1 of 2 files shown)


dellaert:
I think there was already a flag for building examples or not. Also, that flag shows the naming convention for compile time flags :-)

varunagrawal:
Hi @spacey-sooty. You can set the following CMake variables to OFF to disable building tests and examples.
```cmake
GTSAM_BUILD_EXAMPLES_ALWAYS=OFF
GTSAM_BUILD_TESTS=OFF
```

I agree that building examples takes a long time, so I have them disabled myself when compiling.

spacey-sooty:
> Hi @spacey-sooty. You can set the following CMake variables to OFF to disable building tests and examples.
> 
> ```cmake
> 
> GTSAM_BUILD_EXAMPLES_ALWAYS=OFF
> 
> GTSAM_BUILD_TESTS=OFF
> 
> ```
> 
> 
> 
> I agree that building examples takes a long time, so I have them disabled myself when compiling.

I couldn't find this at all when searching through the documentation. Mentioning in there would be good.

varunagrawal:
Yeah I agree. We have a lot of work to do wrt the docs. :)

varunagrawal:
@spacey-sooty would you be willing to update this PR to instead make it a docs update with the flags above? I can approve and merge it then.

spacey-sooty:
Where would I find the docs in the repo?

varunagrawal:
I would recommend adding it to the [INSTALL.md](https://github.com/borglab/gtsam/blob/develop/INSTALL.md) file since it has other notes about CMake flags.

spacey-sooty:
@varunagrawal I've pushed a new commit documenting it, my IDE has also caught some unused whitespace and fixed that but I can revert that if needed

varunagrawal (approved):

varunagrawal:
Thank you so much!

PR #1921, opened 2024-12-08
fix CHECK_EQUAL

truher:
Previously, CHECK_EQUAL returned upon success, skipping the rest of the test.  :-)

### DIFF
--- a/CppUnitLite/Test.h
+++ b/CppUnitLite/Test.h
@@ -129,7 +129,7 @@ protected:
     result_.addFailure(Failure(name_, __FILE__, __LINE__, #expected, #actual)); }
 
 #define CHECK_EQUAL(expected,actual)\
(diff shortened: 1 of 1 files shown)


varunagrawal (approved):
Brilliant catch!!

varunagrawal:
@dellaert tagging for visibility. Can confirm this was an issue that this PR fixes.

dellaert:
Wait a minute :-) this is *expected* behavior: CHECK exits immediately, EXPECT continues, no?

truher:
oh yeah, that's true.  previously CHECK_EQUAL would return *upon success* which is never right.  my version isn't right either.

ha ha, apparently Michael Feathers noticed the same thing:

http://wiki.c2.com/?CppUnitLite

truher:
see https://github.com/borglab/gtsam/pull/1923

PR #1922, opened 2024-12-08
Ring struct in a separate file

varunagrawal:
Moved the `Ring` struct to a separate file so we can use a common implementation across the board.

This will help in the refactor to reduce code duplication and allow common base class methods.


dellaert (approved):
Much better

PR #1923, opened 2024-12-09
actually fix CHECK_EQUAL

truher:
return on failure as discussed here https://github.com/borglab/gtsam/pull/1921

correct fix is described here by the original author: https://wiki.c2.com/?CppUnitLite


varunagrawal (approved):

varunagrawal:
@truher can you please add the link to the CppUnitLite webpage where you saw the correct fix in the PR description?

PR #1924, opened 2024-12-09
Kalman filter update

dellaert:
no pun intended :-)

- Adds better docs
- Uses LLT when full covariance matrices are provided.
- only use JacobianFactors as default factorization is QR anyway
- Some re-formatting from very old style.


gtsam/linear/KalmanFilter.cpp
```diff
@@ -87,72 +102,77 @@ void KalmanFilter::print(const string& s) const {
...
-  Matrix G = Ht * M * H;
-  Vector g = Ht * M * z;
-  double f = dot(z, M * z);
-  return fuse(p, std::make_shared<HessianFactor>(k, G, g, f));
+
+  // Perform Cholesky decomposition of Q
+  FACTORIZE_Q_INTO_L(Q)
+
+  // Pre-multiply H and z with R, respectively
+  const Matrix RtH = L.solve(H);  // = R * H
```
dellaert: should be RH

gtsam/linear/KalmanFilter.h
```diff
@@ -32,120 +32,186 @@ namespace gtsam {
...
 
-/* ************************************************************************* */
+  /**
+   * Return the dimensionality of the state.
+   * @return Dimensionality of the state.
+   */
+  size_t dim() const { return n_; }
+};
 
+}  // namespace gtsam
```
dellaert: add newline

gtsam/linear/KalmanFilter.cpp
```diff
@@ -87,72 +102,77 @@ void KalmanFilter::print(const string& s) const {
...
+  const DenseIndex n = dim();
+  assert(F.cols() == n);
   assert(F.rows() == n);
   assert(B.rows() == n);
   assert(B.cols() == u.size());
   assert(Q.rows() == n);
   assert(Q.cols() == n);
 #endif
 
-  // The factor related to the motion model is defined as
```
varunagrawal, line 115: Maybe keep this comment? It's pedagogical.

gtsam/linear/KalmanFilter.cpp
```diff
@@ -63,20 +71,27 @@ KalmanFilter::fuse(const State& p, GaussianFactor::shared_ptr newFactor) const {
...
-KalmanFilter::State KalmanFilter::init(const Vector& x, const Matrix& P0) const {
-
+KalmanFilter::State KalmanFilter::init(const Vector& x0,
+                                       const Matrix& P0) const {
   // Create a factor graph f(x0), eliminate it into P(x0)
   GaussianFactorGraph factorGraph;
-  factorGraph.emplace_shared<HessianFactor>(0, x, P0); // 0.5*(x-x0)'*inv(Sigma)*(x-x0)
+
+  // Perform Cholesky decomposition of P0
+  FACTORIZE_Q_INTO_L(P0)
```
varunagrawal: In all honesty, I am not a big fan of this. It's not functional and the variable `L` seems to magically appear, making the code less readable.

I did some investigating and perhaps we can instead define the following function
```cpp
Matrix factorizeQIntoR(const Matrix& Q) {
  Eigen::LLT<Matrix> llt(Q);
  auto L = llt.matrixL();
  Matrix I = Matrix::Identity(Q.rows(), Q.cols());
  return L.solve(I);
}
```

and you can use it as
```cpp
const Matrix R = factorizeQIntoR(P0);
const Vector b = R * x0;  // = R*x0
```
which makes it self-documenting as a bonus.

varunagrawal: One more thing: you can do `R.triangularView<Eigen::Lower>() * x0` to improve performance since it takes advantage of the fact that R is lower triangular.

I can't seem to find a way to return an `Eigen::TriangularView<const Matrix, Eigen::Lower>` in a way that works,  which I guess is why you went the macro route in the first place.

dellaert: Yeah, I tried a function first. Does your snippet compile?

dellaert: R is upper triangular. L.solve(x) is same as multiplying with R, read comment more closely :-)

dellaert: I can’t use your function as I need to solve multiple times with same L. Storing R is less efficient.
Is the macro a dealbreaker? It works, it’s faster, it avoids hard Eigen type issues.

varunagrawal: Then I found something interesting. I printed `L.solve(I)` to verify things when coming up with the above, and `R` is lower triangular.

Regarding the macro, I am not opposed to it, I was hoping for a cleaner, less magicky solution. :)

varunagrawal (approved):

gtsam/linear/KalmanFilter.cpp
dellaert: Oh, my, this was a rabbithole, but indeed. $R^TR = L^{-T}L^{-1}$ does *not* imply $R = L^{-1}$. I updated the comment and found an elegant way to not have a macro:

```c++
// In the code below we often get a covariance matrix Q, and we need to create a
// JacobianFactor with cost 0.5 * |Ax - b|^T Q^{-1} |Ax - b|. Factorizing Q as
//   Q = L L^T
// and hence
//   Q^{-1} = L^{-T} L^{-1} = M^T M, with M = L^{-1}
// We then have
//   0.5 * |Ax - b|^T Q^{-1} |Ax - b|
// = 0.5 * |Ax - b|^T M^T M |Ax - b|
// = 0.5 * |MAx - Mb|^T |MAx - Mb|
// The functor below efficiently multiplies with M by calling L.solve().

struct InverseL : public Eigen::LLT<Matrix> {
  InverseL(const Matrix& Q) : Eigen::LLT<Matrix>(Q) {}
  Matrix operator*(const Matrix& A) const { return matrixL().solve(A); }
};
```

gtsam/linear/KalmanFilter.cpp
dellaert: It's still there in predict2

PR #1919, opened 2024-12-07
Discrete Elimination Refactor

varunagrawal:
Refactor discrete elimination so it is agnostic to the underlying derived type of the `DiscreteFactor`.

PR #1925, opened 2024-12-09
Make `DiscreteFactor::operator()` a common base method

varunagrawal:
1. Make `operator()` a common method amongst discrete factors which just calls `evaluate(values)`. This reduces duplication.
2. Update `DiscreteValues` to `Assignment<Key>` (the base class) since this is more generic. Otherwise `operator()(DiscreteValues)` doesn't work in all cases.

I am pulling out useful updates from #1919 and making smaller PRs. :)


dellaert (approved):
I like it! I’m a little bit worried about whether the type change will impact roboticsbook.org, but I think not. I’ll deal with it if so.

varunagrawal:
I tried making it as general as possible so hopefully you have no impact. 🤞🏾

PR #1927, opened 2024-12-10
Various Discrete Improvements

varunagrawal:
1. Use the `DiscreteFactorGraph::product` method since it has a pointer check in it.
2. Put common code in `ProductAndNormalize`.
3. Return `DiscreteFactor` as the $\tau$ factor after elimination, instead of `DecisionTreeFactor`.
4. Reduce some code duplication in `DiscreteConditional` aka less code is better code.
5. Formatting.

### DIFF
--- a/gtsam/discrete/DiscreteFactorGraph.cpp
+++ b/gtsam/discrete/DiscreteFactorGraph.cpp
@@ -14,6 +14,7 @@
  *  @date Feb 14, 2011
  *  @author Duy-Nguyen Ta
  *  @author Frank Dellaert
+ *  @author Varun Agrawal
  */
 
 #include <gtsam/discrete/DiscreteBayesTree.h>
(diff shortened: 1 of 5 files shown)


gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -110,15 +111,18 @@ namespace gtsam {
...
+   * @param factors The factors to multiply as a DiscreteFactorGraph.
+   * @return DecisionTreeFactor
+   */
+  static DecisionTreeFactor ProductAndNormalize(
+      const DiscreteFactorGraph& factors) {
     // PRODUCT: multiply all factors
     gttic(product);
-    DecisionTreeFactor product;
-    for (auto&& factor : factors) product = (*factor) * product;
+    DecisionTreeFactor product = factors.product();
```
varunagrawal, line 125: This is the biggest fix in this PR since now we also check for nullptrs.

dellaert (approved):

PR #1928, opened 2024-12-11
TableFactor Fixes

varunagrawal:
Found some issues in the `TableFactor` while debugging an incorrect assignment on conversion from `DiscreteConditional` to `TableFactor`. The fixes are listed below.

1. Fix the `equals` method so it also checks the keys are the same. Updated a unit test to show this.
2. Updated printing so the vertical bars are neatly aligned.
3. Check if the size of the provided table matches the maximum cardinality of the keys. Else we get a weird result with collisions like below:
```cpp
TableFactor:
 f[ (0,2), (1,3), ]
(0, 0)(1, 0) | 0.166667   | 0  // Repeat
(0, 0)(1, 1) | 0.277778   | 1
(0, 0)(1, 2) | 0.3        | 2  // Repeat
(0, 0)(1, 0) | 0.333333   | 3  // Repeat
(0, 1)(1, 1) | 0.333333   | 4  // Repeat
(0, 1)(1, 2) | 0.333333   | 5
(0, 1)(1, 0) | 0.5        | 6
(0, 1)(1, 1) | 0.388889   | 7  // Repeat
(0, 0)(1, 2) | 0.366667   | 8  // Repeat
number of nnzs: 9
```

I am also working on improving the conversion from `DecisionTreeFactor` to `TableFactor` in a way that should be faster and more memory-efficient than the current scheme. I'll probably push that to the next PR.

### DIFF
--- a/gtsam/discrete/tests/testTableFactor.cpp
+++ b/gtsam/discrete/tests/testTableFactor.cpp
@@ -144,7 +144,7 @@ TEST(TableFactor, constructors) {
   EXPECT(assert_equal(expected_f5, f5, 1e-6));
 
   TableFactor f5_with_wrong_keys(V & O, expected_values);
-  EXPECT(!assert_equal(f5_with_wrong_keys, f5, 1e-9));
+  EXPECT(assert_inequal(f5_with_wrong_keys, f5, 1e-9));
 }
 
 /* ************************************************************************* */
--- a/gtsam/discrete/tests/testTableFactor.cpp
+++ b/gtsam/discrete/tests/testTableFactor.cpp
@@ -134,14 +134,17 @@ TEST(TableFactor, constructors) {
   EXPECT(assert_equal(expected, f4));
 
   // Test for 9=3x3 values.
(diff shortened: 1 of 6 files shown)


gtsam/discrete/tests/testTableFactor.cpp
```diff
@@ -134,14 +134,17 @@ TEST(TableFactor, constructors) {
   EXPECT(assert_equal(expected, f4));
 
   // Test for 9=3x3 values.
-  DiscreteKey V(0, 3), W(1, 3);
+  DiscreteKey V(0, 3), W(1, 3), O(100, 3);
   DiscreteConditional conditional5(V | W = "1/2/3 5/6/7 9/10/11");
   TableFactor f5(conditional5);
-  // GTSAM_PRINT(f5);
-  TableFactor expected_f5(
-      X & Y,
```
varunagrawal, line 142: `X&Y` should give a table 3x3, but we provide 9 values and the code didn't complain. :( This is one of those silent bugs.

dellaert (approved):
Stronger for it!

PR #1929, opened 2024-12-13
Fix conversion from `DecisionTree` to `TableFactor`

varunagrawal:
I recently discovered a bug in the way we are converting a `DecisionTree` (and its derivatives such as `DecisionTreeFactor`) into a `TableFactor`, where the leaf value ordering is incorrect.
I have added a test that showcases this error, called `Conversion`.

## Issue Description

The issue stems from the fact that the DecisionTree is structured with the highest label first, so a DecisionTreeFactor with keys `(x0, x1, x2)` will have a leaf visiting scheme where x0 is the fast moving index and x2 is the slowest moving one.
This results in a different ordering of the resulting vector of leaf values than what the TableFactor expects (where x0 is the slowest moving index and x2 is the fastest), and thus incorrect sparse table construction.

Currently I can't seem to figure out what the correct approach is to "invert" this sequencing. For example, for a tree with leaves `0.5 0.4 0.2 0.5 0.6 0.8` (from the `TableFactor::constructors` test), the output of `DecisionTreeFactor::probabilities` is `0.5, 0.5, 0.4, 0.6, 0.2, 0.8` which gives the ordering as `[0, 4, 2, 6, 1, 5, 3, 7]`.

I spent all day yesterday trying to figure out the mapping between the ordering but there is always some test case that fails. If there is a formula/algorithm for this conversion of orderings, please let me know so I can use that instead of the current approach.

## Proposed Fix

To address the above issue in a general fashion, I updated `ComputeLeafOrdering` to visit the leaves of the tree and generate the `SparseVector` index from the corresponding `Assignment` of discrete keys. This allows us to directly construct the sparse table with the correct indexing and structure, circumventing 2 issues:

1. The need to perform the mapping from highest label first to lowest label first (or any other scheme) ordering.
2. `DecisionTree::probabilities` constructs a **dense** vector which can cause exponential blowup in memory. The proposed approach in this fix prevents the need to ever construct this vector. Of course this means that performance is predicated on the fact that the DecisionTree is "sparse" to begin with.

The major drawback though is that based on a benchmarking test I wrote, this scheme is slower than the current approach, but premature optimization is the root of all evil so we will cross that bridge when we come to it.

### DIFF
--- a/gtsam/discrete/TableFactor.cpp
+++ b/gtsam/discrete/TableFactor.cpp
@@ -62,40 +62,55 @@ TableFactor::TableFactor(const DiscreteKeys& dkeys,
     : TableFactor(dkeys, DecisionTreeFactor(dkeys, dtree)) {}
 
 /**
- * @brief Compute the correct ordering of the leaves in the decision tree.
+ * @brief Compute the indexing of the leaves in the decision tree based on the
+ * assignment and add the (index, leaf) pair to a SparseVector.
+ *
+ * We visit each leaf in the tree, and using the cardinalities of the keys,
+ * compute the correct index to add the leaf to a SparseVector which
+ *  is then used to create the TableFactor.
  *
  * @param dt The DecisionTree
  * @return Eigen::SparseVector<double>
  */
-static Eigen::SparseVector<double> ComputeLeafOrdering(
+static Eigen::SparseVector<double> ComputeSparseTable(
     const DiscreteKeys& dkeys, const DecisionTreeFactor& dt) {
   // SparseVector needs to know the maximum possible index,
   // so we compute the product of cardinalities.
-  size_t prod_cardinality = 1;
+  size_t cardinalityProduct = 1;
   for (auto&& [_, c] : dt.cardinalities()) {
-    prod_cardinality *= c;
+    cardinalityProduct *= c;
   }
-  Eigen::SparseVector<double> sparse_table(prod_cardinality);
+  Eigen::SparseVector<double> sparseTable(cardinalityProduct);
   size_t nrValues = 0;
   dt.visit([&nrValues](double x) {
     if (x > 0) nrValues += 1;
   });
-  sparse_table.reserve(nrValues);
+  sparseTable.reserve(nrValues);
 
   std::set<Key> allKeys(dt.keys().begin(), dt.keys().end());
 
+  /**
+   * @brief Functor which is called by the DecisionTree for each leaf.
+   * For each leaf value, we use the corresponding assignment to compute a
+   * corresponding index into a SparseVector. We then populate sparseTable with
+   * the value at the computed index.
+   *
+   * Takes advantage of the sparsity of the DecisionTree to be efficient. When
+   * merged branches are encountered, we enumerate over the missing keys.
+   *
+   */
   auto op = [&](const Assignment<Key>& assignment, double p) {
     if (p > 0) {
       // Get all the keys involved in this assignment
-      std::set<Key> assignment_keys;
+      std::set<Key> assignmentKeys;
       for (auto&& [k, _] : assignment) {
-        assignment_keys.insert(k);
+        assignmentKeys.insert(k);
       }
 
       // Find the keys missing in the assignment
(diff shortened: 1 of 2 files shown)


gtsam/discrete/TableFactor.cpp
```diff
@@ -64,27 +64,69 @@ TableFactor::TableFactor(const DiscreteKeys& dkeys,
 /**
  * @brief Compute the correct ordering of the leaves in the decision tree.
  *
- * This is done by first taking all the values which have modulo 0 value with
- * the cardinality of the innermost key `n`, and we go up to modulo n.
- *
  * @param dt The DecisionTree
- * @return std::vector<double>
+ * @return Eigen::SparseVector<double>
```
dellaert, line 73: This function says it computes an ordering but then returns a sparse vector. Maybe add more documentation and possibly even change the name if it no longer reflects the code.

gtsam/discrete/TableFactor.cpp
```diff
@@ -64,27 +64,69 @@ TableFactor::TableFactor(const DiscreteKeys& dkeys,
...
-                                        const DecisionTreeFactor& dt) {
-  std::vector<double> probs = dt.probabilities();
-  std::vector<double> ordered;
+static Eigen::SparseVector<double> ComputeLeafOrdering(
+    const DiscreteKeys& dkeys, const DecisionTreeFactor& dt) {
+  // SparseVector needs to know the maximum possible index,
+  // so we compute the product of cardinalities.
+  size_t prod_cardinality = 1;
+  for (auto&& [_, c] : dt.cardinalities()) {
+    prod_cardinality *= c;
```
dellaert: Two style comments:
- we do not use Google style. This can be debated, but for now, following the rest of GTSAM, should be prodCardinality
- Abbreviations make the code less readable, so consider productCardinality, or even better here: cardinalityProduct

gtsam/discrete/TableFactor.cpp
```diff
@@ -64,27 +64,69 @@ TableFactor::TableFactor(const DiscreteKeys& dkeys,
...
+  Eigen::SparseVector<double> sparse_table(prod_cardinality);
+  size_t nrValues = 0;
+  dt.visit([&nrValues](double x) {
+    if (x > 0) nrValues += 1;
+  });
+  sparse_table.reserve(nrValues);
+
+  std::set<Key> allKeys(dt.keys().begin(), dt.keys().end());
+
+  auto op = [&](const Assignment<Key>& assignment, double p) {
```
dellaert, line 102: Add a block comment of what will happen here

gtsam/discrete/TableFactor.cpp
```diff
@@ -64,27 +64,69 @@ TableFactor::TableFactor(const DiscreteKeys& dkeys,
...
+          prev_cardinality *= dt.cardinality(it->first);
+        }
+        sparse_table.coeffRef(idx) = p;
       }
     }
-  }
-  return ordered;
+  };
+
+  dt.visitWith(op);
```
dellaert, line 144: Add a comment

dellaert (approved):
Cool. I’ll approve but would like to see the comments addressed. I know the style comment might rankle. Fee free to ignore IF this file is consistent in using underscores. But typically with new code I’d adhere to GTSAM.

varunagrawal:
I'll update the style. My main goal was to get this out and get comments on whether there was a better way to handle the ordering.

Another approach I had was to reverse the order of keys to the TableFactor constructor but that causes all sorts of equality checks to fail, both for keys (since they are stored as a KeyVector, I am assuming the ordering matters mathematically though operationally I don't see why) and the SparseVector.

dellaert:
> I'll update the style. My main goal was to get this out and get comments on whether there was a better way to handle the ordering.
> 
> Another approach I had was to reverse the order of keys to the TableFactor constructor but that causes all sorts of equality checks to fail, both for keys (since they are stored as a KeyVector, I am assuming the ordering matters mathematically though operationally I don't see why) and the SparseVector.

I think that’s probably safe. And your fix also prevents exponential blow-up :-)

PR #1930, opened 2024-12-14
Endow NavState with group operations

dellaert:
This is a smaller PR than #1613 that adds $SE_2(3)$ group operations to NavState. Be careful though: we use R,t,v order.
A lot of this code is written by @varunagrawal and possibly @ProfFan.

This PR:
- Takes all unit-tested changes from #1613 except transposes and todos
- Adds `ExpmapTranslation` to `Pose3`, whose derivative with respect to angular velocity is exactly Q
- Uses that in both `Pose3::Expmap` and `NavState::Expmap` 

`ExpmapTranslation` avoids code duplication and clearly exposes the structure of the exponential map Jacobian. `NavState::Expmap` is also faster because we don't create the 3x3 W matrices *unless* the derivative is asked for.

### DIFF
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -26,8 +26,6 @@
 
 namespace gtsam {
 
-using std::vector;
-
 /** instantiate concept checks */
 GTSAM_CONCEPT_POSE_INST(Pose3)
 
@@ -167,21 +165,23 @@ Pose3 Pose3::interpolateRt(const Pose3& T, double t) const {
 /* ************************************************************************* */
 /** Modified from Murray94book version (which assumes w and v normalized?) */
 Pose3 Pose3::Expmap(const Vector6& xi, OptionalJacobian<6, 6> Hxi) {
-  if (Hxi) *Hxi = ExpmapDerivative(xi);
+  // Get angular velocity omega and translational velocity v from twist xi
+  const Vector3 w = xi.head<3>(), v = xi.tail<3>();
 
-  // get angular velocity omega and translational velocity v from twist xi
-  Vector3 omega(xi(0), xi(1), xi(2)), v(xi(3), xi(4), xi(5));
+  // Compute rotation using Expmap
+  Rot3 R = Rot3::Expmap(w);
 
-  Rot3 R = Rot3::Expmap(omega);
-  double theta2 = omega.dot(omega);
-  if (theta2 > std::numeric_limits<double>::epsilon()) {
-    Vector3 t_parallel = omega * omega.dot(v);  // translation parallel to axis
-    Vector3 omega_cross_v = omega.cross(v);     // points towards axis
-    Vector3 t = (omega_cross_v - R * omega_cross_v + t_parallel) / theta2;
-    return Pose3(R, t);
-  } else {
-    return Pose3(R, v);
+  // Compute translation and optionally its Jacobian in w
+  Matrix3 Q;
+  const Vector3 t = ExpmapTranslation(w, v, Hxi ? &Q : nullptr, R);
+
+  if (Hxi) {
+    const Matrix3 Jw = Rot3::ExpmapDerivative(w);
+    *Hxi << Jw, Z_3x3,
+             Q, Jw;
   }
+
+  return Pose3(R, t);
 }
 
 /* ************************************************************************* */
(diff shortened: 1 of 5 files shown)


gtsam/geometry/Pose3.cpp
```diff
@@ -240,55 +240,65 @@ Vector6 Pose3::ChartAtOrigin::Local(const Pose3& pose, ChartJacobian Hpose) {
...
-    const double s = sin(phi), c = cos(phi);
-    const double phi2 = phi * phi, phi3 = phi2 * phi, phi4 = phi3 * phi,
-                 phi5 = phi4 * phi;
-    // Invert the sign of odd-order terms to have the right Jacobian
-    Q = -0.5 * V + (phi - s) / phi3 * (W * V + V * W - WVW) +
-        (1 - phi2 / 2 - c) / phi4 * (W * W * V + V * W * W - 3 * WVW) -
-        0.5 * ((1 - phi2 / 2 - c) / phi4 - 3 * (phi - s - phi3 / 6.) / phi5) *
-            (WVW * W + W * WVW);
+
+  // TODO(Frank): this threshold is *different*. Why?
```
ProfFan, line 288: The following is probably not correct, refer to #780. Have no bandwidth to look into this more carefully.

gtsam/navigation/NavState.cpp
```diff
@@ -103,9 +106,166 @@ bool NavState::equals(const NavState& other, double tol) const {
...
+//------------------------------------------------------------------------------
+Vector9 NavState::ChartAtOrigin::Local(const NavState& state,
+                                       ChartJacobian Hstate) {
+  return Logmap(state, Hstate);
+}
+
 //------------------------------------------------------------------------------
 NavState NavState::retract(const Vector9& xi, //
     OptionalJacobian<9, 9> H1, OptionalJacobian<9, 9> H2) const {
+  // return LieGroup<NavState, 9>::retract(xi, H1, H2);
```
ProfFan: TODO comment?

gtsam/navigation/tests/testNavState.cpp
```diff
@@ -271,6 +424,262 @@ TEST(NavState, Stream)
...
+}
+
+//******************************************************************************
+TEST(NavState, ChartDerivatives) {
+  NavState id;
+  if (ROT3_DEFAULT_COORDINATES_MODE == Rot3::EXPMAP) {
+    CHECK_CHART_DERIVATIVES(id, id);
+      //  CHECK_CHART_DERIVATIVES(id,T2);
+      //  CHECK_CHART_DERIVATIVES(T2,id);
+      //  CHECK_CHART_DERIVATIVES(T2,T3);
```
ProfFan: TODO comments?

gtsam/navigation/tests/testNavState.cpp
```diff
@@ -271,6 +424,262 @@ TEST(NavState, Stream)
...
+/* ************************************************************************* */
+TEST(NavState, retract_localCoordinates2) {
+  NavState t1 = T;
+  NavState t2 = T3;
+  NavState origin;
+  Vector d12 = t1.localCoordinates(t2);
+  EXPECT(assert_equal(t2, t1.retract(d12)));
+  Vector d21 = t2.localCoordinates(t1);
+  EXPECT(assert_equal(t1, t2.retract(d21)));
+  // EXPECT(assert_equal(d12, -d21));
```
ProfFan: Check

gtsam/navigation/tests/testNavState.cpp
```diff
@@ -271,6 +424,262 @@ TEST(NavState, Stream)
...
+}
+
+/* ************************************************************************* */
+TEST(NavState, Expmap_b) {
+  NavState p1(Rot3(), Point3(-100, 0, 0), Point3(100, 0, 0));
+  NavState p2 = p1.retract(
+      (Vector(9) << 0.0, 0.0, 0.1, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0).finished());
+  NavState expected(Rot3::Rodrigues(0.0, 0.0, 0.1), Point3(-100.0, 0.0, 0.0),
+                    Point3(100.0, 0.0, 0.0));
+  EXPECT(assert_equal(expected, p2, 1e-2));
```
ProfFan: Is `1e-2` expected?

ProfFan (approved):
Approving sans some comments. I verified most equations against my notes but I haven't verified formally the `adjoint`'s Jacobians, no bandwidth for that.

I am still very concerned about the use of `epsilon` around the repository. I fixed (while leaving some side effects unsolved) Rot3 in #780, and I am pretty sure there could be more. But this is not super related to this PR, so we can always check later.

gtsam/geometry/Pose3.cpp
dellaert: You're right! And epsilon hides it: we never use it, actually :-) So, I removed epsilon, use same threshold in both cases, and corrected as
```
v + 0.5 * w.cross(v)
```

gtsam/navigation/tests/testNavState.cpp
dellaert: No, passes w default

gtsam/navigation/tests/testNavState.cpp
dellaert: Not expected for retractions

PR #1932, opened 2024-12-15
Left and Right SO(3) Jacobians

dellaert:
Attempt at getting a simpler expression for $Q$, as t in $SE(3)$ and $SE_2(3)$ can be calculated just as JL*v, where JL is the left Jacobian of $SO(3)$. To do this, I
- simplified SO(3) functor
- added leftJacobian
- added applyLeftJacobian
- added doubleCross and verified its derivatives
- add `pose3::ExpmapFunctor` to compute Q block.

But, I asw of yet don't understand (anymore) the expmap derivative and why it does not work out by using ApplyLeftJacobian derivatives. Still, this PR adds a lot of goodies and simplifies code for right and left jacobian opertations in SO3.

### DIFF
--- a/gtsam/geometry/Point3.h
+++ b/gtsam/geometry/Point3.h
@@ -61,9 +61,9 @@ GTSAM_EXPORT Point3 cross(const Point3& p, const Point3& q,
                           OptionalJacobian<3, 3> H_q = {});
 
 /// double cross product @return p x (p x q)
-Point3 doubleCross(const Point3& p, const Point3& q,
-                   OptionalJacobian<3, 3> H1 = {},
-                   OptionalJacobian<3, 3> H2 = {});
+GTSAM_EXPORT Point3 doubleCross(const Point3& p, const Point3& q,
+                                OptionalJacobian<3, 3> H1 = {},
+                                OptionalJacobian<3, 3> H2 = {});
 
 /// dot product
 GTSAM_EXPORT double dot(const Point3& p, const Point3& q,
--- a/gtsam/geometry/Point3.h
+++ b/gtsam/geometry/Point3.h
@@ -55,10 +55,15 @@ GTSAM_EXPORT double norm3(const Point3& p, OptionalJacobian<1, 3> H = {});
 /// normalize, with optional Jacobian
 GTSAM_EXPORT Point3 normalize(const Point3& p, OptionalJacobian<3, 3> H = {});
 
-/// cross product @return this x q
+/// cross product @return p x q
 GTSAM_EXPORT Point3 cross(const Point3& p, const Point3& q,
                           OptionalJacobian<3, 3> H_p = {},
                           OptionalJacobian<3, 3> H_q = {});
+
+/// double cross product @return p x (p x q)
+Point3 doubleCross(const Point3& p, const Point3& q,
+                   OptionalJacobian<3, 3> H1 = {},
+                   OptionalJacobian<3, 3> H2 = {});
 
 /// dot product
 GTSAM_EXPORT double dot(const Point3& p, const Point3& q,
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -168,14 +168,14 @@ Pose3 Pose3::Expmap(const Vector6& xi, OptionalJacobian<6, 6> Hxi) {
   const Vector3 w = xi.head<3>(), v = xi.tail<3>();
 
   // Compute rotation using Expmap
-  Rot3 R = Rot3::Expmap(w);
+  Matrix3 Jw;
+  Rot3 R = Rot3::Expmap(w, Hxi ? &Jw : nullptr);
 
   // Compute translation and optionally its Jacobian in w
   Matrix3 Q;
   const Vector3 t = ExpmapTranslation(w, v, Hxi ? &Q : nullptr, R);
 
   if (Hxi) {
-    const Matrix3 Jw = Rot3::ExpmapDerivative(w);
     *Hxi << Jw, Z_3x3,
              Q, Jw;
   }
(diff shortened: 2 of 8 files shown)


gtsam/geometry/SO3.cpp
```diff
@@ -41,19 +43,23 @@ GTSAM_EXPORT Matrix99 Dcompose(const SO3& Q) {
...
   if (H) *H = Dcompose(R);
   return MR;
 }
 
 void ExpmapFunctor::init(bool nearZeroApprox) {
+  WW = W * W;
   nearZero =
       nearZeroApprox || (theta2 <= std::numeric_limits<double>::epsilon());
   if (!nearZero) {
     sin_theta = std::sin(theta);
```
dellaert: These 3 might not even be needed anymore anywhere else

gtsam/geometry/SO3.cpp
```diff
@@ -62,64 +68,84 @@ ExpmapFunctor::ExpmapFunctor(const Vector3& omega, bool nearZeroApprox)
...
-  }
 }
 
 SO3 ExpmapFunctor::expmap() const {
   if (nearZero)
-    return SO3(I_3x3 + W);
+    return SO3(I_3x3 + W + 0.5 * WW);
   else
-    return SO3(I_3x3 + sin_theta * K + one_minus_cos * KK);
+    return SO3(I_3x3 + A * W + B * WW);
```
dellaert: Ethan Eade, much nicer now

gtsam/geometry/Point3.h
```diff
@@ -55,11 +55,16 @@ GTSAM_EXPORT double norm3(const Point3& p, OptionalJacobian<1, 3> H = {});
...
 GTSAM_EXPORT Point3 normalize(const Point3& p, OptionalJacobian<3, 3> H = {});
 
-/// cross product @return this x q
+/// cross product @return p x q
 GTSAM_EXPORT Point3 cross(const Point3& p, const Point3& q,
                           OptionalJacobian<3, 3> H_p = {},
                           OptionalJacobian<3, 3> H_q = {});
 
+/// double cross product @return p x (p x q)
+Point3 doubleCross(const Point3& p, const Point3& q,
```
dellaert: Needs EXPORT

gtsam/geometry/SO3.cpp
```diff
@@ -62,64 +68,84 @@ ExpmapFunctor::ExpmapFunctor(const Vector3& omega, bool nearZeroApprox)
...
-  }
 }
 
 ExpmapFunctor::ExpmapFunctor(const Vector3& axis, double angle,
                              bool nearZeroApprox)
     : theta2(angle * angle), theta(angle) {
   const double ax = axis.x(), ay = axis.y(), az = axis.z();
-  K << 0.0, -az, +ay, +az, 0.0, -ax, -ay, +ax, 0.0;
-  W = K * angle;
+  W << 0.0, -az, +ay, +az, 0.0, -ax, -ay, +ax, 0.0;
```
dellaert: Why not skewSymmetric?

gtsam/geometry/Pose3.cpp
```diff
@@ -243,29 +243,24 @@ namespace pose3 {
...
  public:
   ExpmapFunctor(const Vector3& omega, bool nearZeroApprox = false)
-      : so3::DexpFunctor(omega, nearZeroApprox) {}
+      : so3::DexpFunctor(omega, nearZeroApprox) {
+        F = nearZero ? - one_twenty_fourth : (B - 0.5) / theta2;
+      }
 
   // Compute the bottom-left 3x3 block of the SE(3) Expmap derivative
   // TODO(Frank): t = applyLeftJacobian(v), it would be nice to understand
   // how to compute mess below from its derivatives in w and v.
```
ProfFan: Still todo?

gtsam/geometry/SO3.h
```diff
@@ -155,30 +156,57 @@ class GTSAM_EXPORT ExpmapFunctor {
 
 /// Functor that implements Exponential map *and* its derivatives
 class DexpFunctor : public ExpmapFunctor {
```
ProfFan: GTSAM_EXPORT?

ProfFan (approved):
I verified most derivatives but not the simplification with E and F

```
LINK : warning LNK4075: ignoring '/INCREMENTAL' due to '/FORCE' specification
   Creating library bin\check_geometry_program.lib and object bin\check_geometry_program.exp
testSO3.cpp.obj : error LNK2019: unresolved external symbol "public: class Eigen::Matrix<double,3,1,0,3,1> __cdecl gtsam::so3::DexpFunctor::cross(class Eigen::Matrix<double,3,1,0,3,1> const &,class gtsam::OptionalJacobian<3,3>)const " (?cross@DexpFunctor@so3@gtsam@@QEBA?AV?$Matrix@N$02$00$0A@$02$00@Eigen@@AEBV45@V?$OptionalJacobian@$02$02@3@@Z) referenced in function "public: virtual void __cdecl SO3CrossTest::run(class TestResult &)" (?run@SO3CrossTest@@UEAAXAEAVTestResult@@@Z)
testSO3.cpp.obj : error LNK2019: unresolved external symbol "public: class Eigen::Matrix<double,3,1,0,3,1> __cdecl gtsam::so3::DexpFunctor::doubleCross(class Eigen::Matrix<double,3,1,0,3,1> const &,class gtsam::OptionalJacobian<3,3>)const " (?doubleCross@DexpFunctor@so3@gtsam@@QEBA?AV?$Matrix@N$02$00$0A@$02$00@Eigen@@AEBV45@V?$OptionalJacobian@$02$02@3@@Z) referenced in function "public: virtual void __cdecl SO3DoubleCrossTest::run(class TestResult &)" (?run@SO3DoubleCrossTest@@UEAAXAEAVTestResult@@@Z)
```

gtsam/geometry/Pose3.cpp
dellaert: Yes, because I cannot just use the applyLeftJacobian Jacobians. Which I think should be able to. Note Pose3 constructor Jacobian is 6x3 in R and 6x3 in t, with latter $[Z_3x3, R.transpose()]$, so my reading was Q = R.transpose()*H1, where H1 is apply jacobian for w.

gtsam/geometry/SO3.h
dellaert: Dang, yes.

dellaert:
@ProfFan any idea why I still need to branch for applyDex, if I also fix rightJacobian to just have the one path?

gtsam/geometry/Pose3.cpp
ProfFan: I think it's likely related to the A,B,C etc not a constant but a function of w. So you need to cache the raw double cross derivatives not the result of `applyLeftJacobian`.

dellaert:
> @ProfFan any idea why I still need to branch for applyDex, if I also fix rightJacobian to just have the one path?

This was resolved. We now have no more branching:
```c++
  Matrix3 rightJacobian() const { return I_3x3 - B * W + C * WW; }
  Matrix3 leftJacobian()  const { return I_3x3 + B * W + C * WW; }
```

gtsam/geometry/Pose3.cpp
dellaert: Yeah, but crossB and doubleCrossC explicitly account for that, e.g., D is related to derivative of B, and E is related to derivative of C. And they are (fairly) extensively unit-tested, and so is applyLeftJacobian.

dellaert: Note:
image
And C is the only second order "constant". In the derivative of applyL, we have:
- $C$ * doubleCrossJacobian = $C*(w v I + v w^T - 2 w v^T )$
and in Q we have:
- $C * (W * V + V * W - W V W)$

It's weird that there even is a W V W here, as that is third-order.

dellaert: @ProfFan I solved it! The applyLeftJacobian yields two jacobians, $H_1$ and $H_2$, with $H_2$ being leftJacobian $J_L$. But the corresponding block is the right Jacobian $J_R$. Hence, the action X of R^T in the mapping must be
$J_R = X J_L$
hence
$X = J_R J_L^{-1}$
I verified that $Q = X H_1$, will make a PR to that effect. Massively simpler.

PR #1933, opened 2024-12-15
Faster `TableFactor::toDecisionTreeFactor`

varunagrawal:
I remembered that `DecisionTree` prefers the keys being ordered highest to lowest for faster speed. This PR updates the vector passed to the DecisionTree constructor such that the ordering matches the key ordering which is highest to lowest.

This update results in `toDecisionTreeFactor` going from taking 2.95 seconds to 0.81 seconds for a tree with 17 keys, which means `2^17 = 131,072` values.
It's still not great, since 17 can be hit pretty quickly, but it is a step forward.

Perhaps avoiding conversions altogether would be the way forward? Maybe subclass `DiscreteConditional` from the `TableFactor`?

### DIFF
--- a/gtsam/discrete/TableFactor.cpp
+++ b/gtsam/discrete/TableFactor.cpp
@@ -262,16 +262,21 @@ DecisionTreeFactor TableFactor::toDecisionTreeFactor() const {
     // k will be in reverse key order already
     uint64_t k;
     ss >> k;
-    std::cout << "ss: " << ss.str() << ", k=" << k << std::endl;
+    std::cout << "ss: " << ss.str() << ", k=" << k
+              << ", v=" << sparse_table_.coeff(i) << std::endl;
     pair_table.push_back(std::make_pair(k, sparse_table_.coeff(i)));
   }
 
-  // Sort based on key so we get values in reverse key order.
+  // Sort based on key assignment so we get values in reverse key order.
   std::sort(
       pair_table.begin(), pair_table.end(),
       [](const std::pair<uint64_t, double>& a,
          const std::pair<uint64_t, double>& b) { return a.first <= b.first; });
 
+  std::cout << "Sorted pair_table:" << std::endl;
+  for (auto&& [k, v] : pair_table) {
+    std::cout << "k=" << k << ", v=" << v << std::endl;
+  }
   // Create the table vector
   std::vector<double> table;
   std::for_each(pair_table.begin(), pair_table.end(),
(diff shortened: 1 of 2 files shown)


varunagrawal:
@dellaert the strict equality check fixed the last CI.

gtsam/discrete/TableFactor.cpp
```diff
@@ -252,12 +251,35 @@ DecisionTreeFactor TableFactor::operator*(const DecisionTreeFactor& f) const {
 /* ************************************************************************ */
 DecisionTreeFactor TableFactor::toDecisionTreeFactor() const {
   DiscreteKeys dkeys = discreteKeys();
-  std::vector<double> table;
+
+  DiscreteKeys rdkeys(dkeys.rbegin(), dkeys.rend());
```
dellaert, line 260: Comment on what will happen in this block

gtsam/discrete/TableFactor.cpp
```diff
@@ -252,12 +251,35 @@ DecisionTreeFactor TableFactor::operator*(const DecisionTreeFactor& f) const {
...
+    }
+    // k will be in reverse key order already
+    uint64_t k;
+    ss >> k;
+    pair_table.push_back(std::make_pair(k, sparse_table_.coeff(i)));
   }
-  // NOTE(Varun): This constructor is really expensive!!
-  DecisionTreeFactor f(dkeys, table);
+
+  // Sort based on key assignment so we get values in reverse key order.
```
dellaert: sort what?

gtsam/discrete/TableFactor.cpp
```diff
@@ -252,12 +251,35 @@ DecisionTreeFactor TableFactor::operator*(const DecisionTreeFactor& f) const {
...
+  std::sort(
+      pair_table.begin(), pair_table.end(),
+      [](const std::pair<uint64_t, double>& a,
+         const std::pair<uint64_t, double>& b) { return a.first < b.first; });
+
+  // Create the table vector
```
dellaert: by...

dellaert (approved):
LGTM, just more comments in code, as not easy stuff. Also, comment on why in code. No one will come back to look at the PR comment.

dellaert:
I’ll let you merge

varunagrawal:
I'm working on the timing branch. I'll merge this after I have everything set up so I can see the difference between current develop and after merging this branch. Please don't merge until then.

PR #1934, opened 2024-12-16
projection and SFM for 2d poses

truher:
These are the "robot on the floor" versions of Projection factors, patterned after the similarly-named pose3 versions.  These are intended for localization in high school robotics, where the landmarks are AprilTag markers in known locations.

PlanarProjectionFactor1 uses constant Pose3 landmarks, a constant camera Pose3 offset, and a constant camera Cal3DS2 calibration.  The measurement is camera pixels.  The free variable is a Pose2 representing the position of the robot on the floor, and the expectation is that this factor will be used for real-time smoothing.

PlanarProjectionFactor2 is a conventional SFM factor with variable pose and landmarks.

PlanarProjectionFactor3 uses variable robot Pose2, constant Pose3 landmarks and camera pixel measurement, as in PlanarProjectionFactor1, but instead of using constant camera Pose3 offset and Cal3DS2 calibration, these are also variables.  This factor is primarily intended for calibration of camera offset and distortion, and it will really only work if you use a pretty good prior, and give it lots of samples.

Reminder for FRC folk, the camera bore is +z.  Team100 have a habit of measuring camera offset as if the camera were looking at +x, so that will need to be transformed by  the client, or we need to switch to +z.


gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,154 @@
...
+     * @class PlanarProjectionFactor
+     * @brief Camera projection for robot on the floor.
+     */
+    class PlanarProjectionFactor : public NoiseModelFactorN<Pose2> {
+        typedef NoiseModelFactorN<Pose2> Base;
+        static const Pose3 CAM_COORD;
+
+    protected:
+
+        Point3 landmark_; // landmark
```
dellaert: Before I do a more thorough review, quick sanity check: things that are stored in the factory are typically constant parameters, but here you store a bunch of things. Could you clarify in the PR comment what this factor optimizes for, and what the measurement function is? Be very clear on what is a constant and what is a variable.

dellaert: Also, optimizing over calibration parameters is brave - and makes it rather different from other basic SFM factors

truher: OK, I updated the top comment.  You're right that there are a bunch of constants for the projection factor, but really only one more than ProjectionFactor itself: the constant landmark.

truher: > Also, optimizing over calibration parameters is brave - and makes it rather different from other basic SFM factors

re: brave, oh, ha ha, it is?  :-)  it does seem to work in simulation.  I haven't tried it on a real robot yet.

re: difference, I think maybe this factor should have a different name anyway, since there's definitely no "S" being extracted from "M," given the constant landmarks.  What would a better name be?

dellaert: > > Also, optimizing over calibration parameters is brave - and makes it rather different from other basic SFM factors
> 
> re: brave, oh, ha ha, it is? :-) it does seem to work in simulation. I haven't tried it on a real robot yet.

It can definitely work. To avoid convergence issues I’d put at least a prior on your calibration parameters.

> re: difference, I think maybe this factor should have a different name anyway, since there's definitely no "S" being extracted from "M," given the constant landmarks. What would a better name be?

You’re right. We sometimes have more than one related factor in one file. I can see several versions here, it should almost be the same effort

- PlanarProjectionFactor1: 1 unknown, the pose
- PlanarProjectionFactor2: conventional: 2 unknowns, the pose and the landmarks. You could typedef this to PlanarSfmFactor.
- PlanarProjectionFactor3: 3 unknowns, including calibration

These can all share a common measurement function: h(pose, point, cal, H1, H2, H3), defined in a PlatProjectionFactorBase

I’d put all of these in one file.

truher:
ok, i consolidated into the single file with a few factors in it, as you suggest.  wdyt?

re: calibration convergence, oh yeah, in my experiments it only works with a suitable prior.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+                        (*H2)(0, 1) = DposeOffsetFwd(0, 4); // du/dy
+                        (*H2)(1, 1) = DposeOffsetFwd(1, 4); // dv/dy
+                        (*H2)(0, 2) = DposeOffsetFwd(0, 2); // du/dyaw
+                        (*H2)(1, 2) = DposeOffsetFwd(1, 2); // dv/dyaw
+                    }
+                    return result;
+                } else {
+                    return camera.project(landmark, {}, {}, {});
+                }
+            } catch (CheiralityException& e) {
```
dellaert, line 87: We have a typedef `GTSAM_THROW_CHEIRALITY_EXCEPTION`. Please ifdef your catch based on that.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+        PlanarProjectionFactorBase() {}
+
+        /**
+         * @param measured corresponding point in the camera frame
+         * @param poseKey index of the robot pose in the z=0 plane
+         */
+        PlanarProjectionFactorBase(const Point2& measured) : measured_(measured) {}
+
+        /**
+         * @param landmark point3
```
dellaert: Please say what the function does :-) Also, suggest to call it `predict` rather than (frowned upon) single letter.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+         * @param offset oriented parallel to pose2 zero i.e. +x
+         * @param calib
+         * @param H1 landmark jacobian
+         * @param H2 pose jacobian
+         * @param H3 offset jacobian
+         * @param H4 calib jacobian
+         */
+        Point2 h(
+            const Point3& landmark,
+            const Pose2& pose,
```
dellaert: I suggest you call this `wTb`, as in, the pose of the vehicle body in the world. A convention that keeps your sanity:
```
aP = aTc * cP;
```
Same for rotations, aRb, transforms b coords to a coords

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+            const Pose3& offset,
```
dellaert: bTc ? camera in body frame

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+
+        /**
+         * @param measured corresponding point in the camera frame
+         * @param poseKey index of the robot pose in the z=0 plane
+         */
+        PlanarProjectionFactorBase(const Point2& measured) : measured_(measured) {}
+
+        /**
+         * @param landmark point3
+         * @param pose
```
dellaert: param wTb planar pose of vehicle body frame in world frame

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+            const Cal3DS2& calib,
+            OptionalMatrixType H1, // 2x3 (x, y, z)
+            OptionalMatrixType H2, // 2x3 (x, y, theta)
+            OptionalMatrixType H3, // 2x6 (rx, ry, rz, x, y, theta)
+            OptionalMatrixType H4  // 2x9
+        ) const {
+            try {
+                // this is x-forward z-up
+                gtsam::Matrix H0; // 6x6
+                Pose3 offset_pose = Pose3(pose).compose(offset, H0);
```
dellaert: I suggest
```c++
// Get pose of camera in world
const Pose3 wTc = Pose3(pose).compose(offset, H2 ? &H0 : nullptr);
```
The ternary prevents compose from doing work if H2 is not requested.

Also, I am not convinced the Pose3 -> Pose2 conversion has an identity Jacobian. Maybe add a quick test to make sure that you know it's 6x3 Jacobian does not have an R^T in it, like Pose3::Create.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+                // this is z-forward y-down
+                gtsam::Matrix H00; // 6x6
+                Pose3 camera_pose = offset_pose.compose(CAM_COORD, H00);
```
dellaert: Oh, I thought we already had wTc. Hmmm, what is CAM_COORD? Where is it defined?

Because the two offsets combine, I actually think the factor would be more general/useful if you just had one bTc that combines offset and extrinsics. Maybe I'm not seeing something.

Style nits:
- use something like kCameraExtrinsics rather than python-style uppercase.
- use camelCase for variables, myVariable.
- But, for poses, use aTc naming convention.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+                PinholeCamera<Cal3DS2> camera = PinholeCamera<Cal3DS2>(camera_pose, calib);
+                if (H2 || H3) {
+                    // Dpose is for pose3, 2x6 (R,t)
+                    gtsam::Matrix Dpose;
+                    Point2 result = camera.project(landmark, Dpose, H1, H4);
```
dellaert: I'd probably do
```c++                    
const Point2 predicted = camera.project(landmark, Hpose ? H_wTc : nullptr, Hlandmark, Hcalib);
```

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+         * @param H2 pose jacobian
+         * @param H3 offset jacobian
+         * @param H4 calib jacobian
+         */
+        Point2 h(
+            const Point3& landmark,
+            const Pose2& pose,
+            const Pose3& offset,
+            const Cal3DS2& calib,
+            OptionalMatrixType H1, // 2x3 (x, y, z)
```
dellaert: Because there are 4, maybe name them Hlandmark, Hpose, Hoffset, Hcalib. Provide nullptr defaults?

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+         * @param model stddev of the measurements, ~one pixel?
+         * @param poseKey index of the robot pose2 in the z=0 plane
+         */
+        PlanarProjectionFactor1(
+            const Point3& landmark,
+            const Point2& measured,
+            const Pose3& offset,
+            const Cal3DS2& calib,
+            const SharedNoiseModel& model,
+            Key poseKey)
```
dellaert: I'd put poseKey first, and default model to nullptr

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+
+        /// @return a deep copy of this factor
+        gtsam::NonlinearFactor::shared_ptr clone() const override {
+            return std::static_pointer_cast<gtsam::NonlinearFactor>(
+                gtsam::NonlinearFactor::shared_ptr(new PlanarProjectionFactor1(*this)));
+        }
+
+
+        /**
+         * @param landmark point3 in the world
```
dellaert, line 128: Say what it is :-)
brief constructor with known landmark, offset, and calibration.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+        /**
+         * @param measured corresponding point in the camera frame
+         * @param offset constant camera offset from pose
+         * @param calib constant camera calibration
+         * @param model stddev of the measurements, ~one pixel?
+         * @param poseKey index of the robot pose2 in the z=0 plane
+         * @param landmarkKey index of the landmark point3
+         */
+        PlanarProjectionFactor2(
+            const Point2& measured,
```
dellaert, line 199: now i would put poseKey then landmarKey, model -> nullptr default

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+         * @param measured corresponding point2 in the camera frame
+         * @param model stddev of the measurements, ~one pixel?
+         * @param poseKey index of the robot pose2 in the z=0 plane
+         * @param offsetKey index of camera offset from pose
+         * @param calibKey index of camera calibration         */
+        PlanarProjectionFactor3(
+            const Point3& landmark,
+            const Point2& measured,
+            const SharedNoiseModel& model,
+            Key poseKey,
```
dellaert, line 262: again, keys first is more natural here.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,318 @@
...
+    struct traits<PlanarProjectionFactor1> :
+        public Testable<PlanarProjectionFactor1> {};
+
+    /**
+     * @class PlanarProjectionFactor2
+     * @brief Two unknowns: the pose and the landmark.
+     * Camera offset and calibration are constant.
+     * This is similar to GeneralSFMFactor, used for SLAM.
+    */
+    class PlanarProjectionFactor2 : public PlanarProjectionFactorBase, public NoiseModelFactorN<Point3, Pose2> {
```
dellaert: I'd reverse and do Pose Point. I think that's what we do everywhere?

dellaert (commented):
Looks pretty awesome! I made a bunch of suggestions. OK If I look at it again before approving?

dellaert:
PS, the 6x3 Jacobian of converting from Pose2 to Pose3 is
```
[[0. 0. 0.]
 [0. 0. 0.]
 [0. 0. 1.]
 [1. 0. 0.]
 [0. 1. 0.]
 [0. 0. 0.]]
 ```
You can use this :-) Maybe even add it as OptionalJacobian<6,3> to a new “FromPose2” constructor.

File: gtsam/slam/PlanarProjectionFactor.h (hunk 2 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h:87]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 4 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 6 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 5 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 3 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 7 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 9 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: You mean instead of if (HwTb || HbTc) ?

File: gtsam/slam/PlanarProjectionFactor.h (hunk 10 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 11 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: i reordered the arguments but I don't understand the default.  Isn't a non-null model required here?  What would it mean to provide null?

File: gtsam/slam/PlanarProjectionFactor.h (hunk 12 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h:128]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 14 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h:262]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 13 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h:199]: done

File: gtsam/slam/PlanarProjectionFactor.h (hunk 15 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: done

**Comment by truher** [comment]
> PS, the 6x3 Jacobian of converting from Pose2 to Pose3 is
> 
> ```
> [[0. 0. 0.]
>  [0. 0. 0.]
>  [0. 0. 1.]
>  [1. 0. 0.]
>  [0. 1. 0.]
>  [0. 0. 0.]]
> ```
> 
> You can use this :-) Maybe even add it as OptionalJacobian<6,3> to a new “FromPose2” constructor.

oh, ha ha, yes, this is what i was doing by picking out the values in the previous version.  of course this is how it should work.

File: gtsam/slam/PlanarProjectionFactor.h (hunk 8 above)
Comment (truher) [reply gtsam/slam/PlanarProjectionFactor.h]: The "CAM_COORD" transforms the supplied "bTc" offset so that bTc zero is pointing at +x instead of +z, because that's where FRC code thinks the pose3 zero is, so the caller will want to supply an offset like that.

https://github.com/wpilibsuite/allwpilib/blob/80c391e18278fcb91189a664bb2b2ddaad7eb7a9/wpimath/src/main/java/edu/wpi/first/math/geometry/Pose3d.java#L40

Am I understanding correctly that the camera zero is aiming at +z?  Maybe there's another way handle this transform?

**Comment by truher** [comment]
OK, I made the changes I could, and I'd appreciate your guidance for the rest.  I'm grateful for your kind attention to this project!

File: gtsam/slam/PlanarProjectionFactor.h (hunk 11 above)
Comment (dellaert) [reply gtsam/slam/PlanarProjectionFactor.h]: It’s actually equivalent to identity covariance. If you are using 1 pixel isotopic noise, null pointer will be a bit faster. No idea how much - you could time it :-)

File: gtsam/geometry/Pose3.h [hunk 16]
```diff
@@ -78,6 +78,8 @@ class GTSAM_EXPORT Pose3: public LieGroup<Pose3, 6> {
                       OptionalJacobian<6, 3> HR = {},
                       OptionalJacobian<6, 3> Ht = {});
```
dellaert, line 80: Comment

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+         * @param HwTb jacobian
+         * @param HbTc jacobian
+         * @param Hcalib jacobian
+         */
+        Point2 predict(
+            const Point3& landmark,
+            const Pose2& wTb,
+            const Pose3& bTc,
+            const Cal3DS2& calib,
+            OptionalMatrixType Hlandmark = nullptr, // 2x3 (x, y, z)
```
dellaert: Try to use OptionalJacobian<2,3> Hlandmark = {} here

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+                    if (HwTb) 
+                        *HwTb =  DposeOffset * H0 * Hp;
+                    return result;
+                } else {
+                    return camera.project(landmark, {}, {}, {});
+                }
+#ifndef GTSAM_THROW_CHEIRALITY_EXCEPTION
+            } catch (CheiralityException& e) {
+                std::cout << "****** CHIRALITY EXCEPTION ******\n";
+                if (Hlandmark) *Hlandmark = Matrix::Zero(2, 3);
```
dellaert: Then you can use Hlandmark->setZero() here.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+            const Cal3DS2& calib,
+            OptionalMatrixType Hlandmark = nullptr, // 2x3 (x, y, z)
+            OptionalMatrixType HwTb = nullptr, // 2x3 (x, y, theta)
+            OptionalMatrixType HbTc = nullptr, // 2x6 (rx, ry, rz, x, y, theta)
+            OptionalMatrixType Hcalib = nullptr // 2x9
+        ) const {
+#ifndef GTSAM_THROW_CHEIRALITY_EXCEPTION
+            try {
+#endif
+                gtsam::Matrix Hp; // 6x3
```
dellaert: This is an (expensive) malloc. Use Matrix63, fixed size, on the stack. Try to use everywhere you can.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+                gtsam::Matrix H0; // 6x6
```
dellaert: Same, and BTW, you can lose all `gtsam::as inside the namespace

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+                // this is x-forward z-up
+                Pose3 wTc = Pose3::FromPose2(wTb, Hp).compose(bTc, HwTb ? &H0 : nullptr);
```
dellaert: Actually, 
- Thanks for adding the Pose3 named constructor. Let’s keep it!
- Now I realize though, it’s better to make a static const Matrix63 in this file, as it’s always the same :-)

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+                    gtsam::Matrix Dpose;
+                    Point2 result = camera.project(landmark, Dpose, Hlandmark, Hcalib);
+                    gtsam::Matrix DposeOffset = Dpose * H00; // 2x6
+                    if (HbTc)
+                        *HbTc = DposeOffset; // with Eigen this is a deep copy (!)
+                    if (HwTb) 
+                        *HwTb =  DposeOffset * H0 * Hp;
+                    return result;
+                } else {
+                    return camera.project(landmark, {}, {}, {});
```
dellaert, line 84: No need for nil, is default.

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+#ifndef GTSAM_THROW_CHEIRALITY_EXCEPTION
+            try {
+#endif
+                gtsam::Matrix Hp; // 6x3
+                gtsam::Matrix H0; // 6x6
+                // this is x-forward z-up
+                Pose3 wTc = Pose3::FromPose2(wTb, Hp).compose(bTc, HwTb ? &H0 : nullptr);
+                // this is z-forward y-down
+                gtsam::Matrix H00; // 6x6
+                Pose3 camera_pose = wTc.compose(CAM_COORD, H00);
```
dellaert: I like how you’re thinking about this: in your application you have a particular configuration and the “calibration” is just an offset from that. However:
- you’re incurring another compose and Jacobian computation here
- You’re adding another matrix multiply in the chain
- You can easily initialize the calibration unknown to include CAM_COORD outside the factor

This will make the factor more general, and save tons of compute :-)

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+         * @param calib constant camera calibration
+         * @param model stddev of the measurements, ~one pixel?
+         */
+        PlanarProjectionFactor1(
+            Key poseKey,
+            const Point3& landmark,
+            const Point2& measured,
+            const Pose3& bTc,
+            const Cal3DS2& calib,
+            const SharedNoiseModel& model)
```
dellaert: = nullptr. Trust me :-)

gtsam/slam/PlanarProjectionFactor.h
```diff
@@ -0,0 +1,320 @@
...
+            : PlanarProjectionFactorBase(measured),
+            NoiseModelFactorN(model, poseKey),
+            landmark_(landmark),
+            bTc_(bTc),
+            calib_(calib) {
+            assert(2 == model->dim());
```
dellaert: Kill assert as will trip on nullptr. Is checked deeper in the code.

gtsam/slam/tests/testPlanarProjectionFactor.cpp
```diff
@@ -0,0 +1,370 @@
+/**
+ * @file testPlanarProjectionFactor.cpp
+ * @date Dec 3, 2024
+ * @author joel@truher.org
+ * @brief unit tests for PlanarProjectionFactor
+ */
+
+#include <random>
```
dellaert, line 8: Randomness in test can be problematic. Ok but definitely seed with, say, 42.

gtsam/slam/tests/testPlanarProjectionFactor.cpp
```diff
@@ -0,0 +1,370 @@
...
+    Matrix23 H1Expected = (Matrix23() << //
+        -360, 280, 640, //
+        -360, 80, 440).finished();
+    CHECK(assert_equal(H1Expected, H1Actual, 1e-6));
+}
+
+TEST(PlanarProjectionFactor1, jacobian) {
+    // test many jacobians with many randoms
+
+    std::default_random_engine g;
```
dellaert: Consider rng or engine

gtsam/slam/tests/testPlanarProjectionFactor.cpp
```diff
@@ -0,0 +1,370 @@
...
+    std::vector<Matrix> actualHs(3);
+
+    gtsam::Vector actual = factor.unwhitenedError(values, actualHs);
+    CHECK(assert_equal(Vector2(0, 0), actual));
+
+    const Matrix& H1Actual = actualHs.at(0);
+    const Matrix& H2Actual = actualHs.at(1);
+    const Matrix& H3Actual = actualHs.at(2);
+
+    CHECK_EQUAL(2, H1Actual.rows());
```
dellaert: These checks will be moot if you use OptionalJacobian<r,c>

gtsam/slam/tests/testPlanarProjectionFactor.cpp
```diff
@@ -0,0 +1,370 @@
...
+    CHECK(assert_equal(H3Expected, H3Actual, 1e-6));
+}
+
+TEST(PlanarProjectionFactor3, error3) {
+    Point3 landmark(1, 1, 1);
+    Point2 measured(0, 0);
+    Pose3 offset;
+    Cal3DS2 calib(200, 200, 0, 200, 200, -0.2, 0.1);
+
+    SharedNoiseModel model = noiseModel::Diagonal::Sigmas(Vector2(1, 1));
```
dellaert, line 259: That’s the same as nullptr :-)

gtsam/slam/tests/testPlanarProjectionFactor.cpp
```diff
@@ -0,0 +1,370 @@
...
+            X(0), C(0), K(0), landmark, measured, model);
+
+        Pose2 pose(s(g), s(g), s(g));
+
+        // actual H
+        Matrix H1, H2, H3;
+        factor.evaluateError(pose, offset, calib, H1, H2, H3);
+
+        Matrix expectedH1 = numericalDerivative31<Vector, Pose2, Pose3, Cal3DS2>(
+            [&factor](const Pose2& p, const Pose3& o, const Cal3DS2& c) {
```
dellaert, line 317: Copy/pasta :-) assign to auto. maybe you’re doing this because auto might not work. If you type with std::function<> it will work though.

dellaert (commented):
Getting tight!
Not approving yet because I do feel strongly about CAM_COORD. Happy to hear arguments for it, but I think that’s really an application-dependent thing, and will be more expensive to boot…

gtsam/slam/PlanarProjectionFactor.h
dellaert: Oops, I actually never saw this question. I think what you're saying in code now is:
```c++
camera_pose = FromPose2(wTb) * bTc * CAM_COORD.
```

and you are optimizing over wTb and also, when calibrating, over bTc. Say, in the latter, you start the optimization with bTc_0 and it optimizes to bTc_1. You then use bTc_1 during operation.

OK, well, you can also do this:
```c++
camera_pose = FromPose2(wTb) * new_bTc
```
As a user, you can initialize to new_bTc_0 = bTc_0 * CAM_COORD, and this will then optimize to new_bTc_1 = bTc_1 * CAM_COORD. Then use new_bTc_1 during operation. 

In other words: unless I'm missing something, not something the factor need to be aware of :-)

gtsam/slam/PlanarProjectionFactor.h
dellaert: You're right, that might already be handled. The key is to examine your logic to make sure no Jacobians are computed unless needed.

truher:
mostly done, sorry for the delay, I'm spending a bit of time to make sure I understand how the client should transform the marginal covariances.

dellaert:
> mostly done, sorry for the delay, I'm spending a bit of time to make sure I understand how the client should transform the marginal covariances.

I don’t quite know what you’re trying to do, but if you’re trying to see how covariance transform under compose, you can use the Jacobian $H$ of compose, then $H \Sigma H^T$

truher:
yeah, that's exactly what i'm trying to see.  thanks!

gtsam/slam/PlanarProjectionFactor.h
truher: That's what I thought too, but clang complained that without the explicit nils, the overload was ambiguous.

gtsam/slam/tests/testPlanarProjectionFactor.cpp
truher: yeah, i fiddled with a few options before settling on this one.

truher:
Happy holidays!  I addressed most of your feedback, in particular removing the whole "cam coord" idea (now the camera orientation is a client concern), and adding a test that illustrates transforming the camera covariance matrix to the body frame.  WDYT?

dellaert (approved):
LGTM. Merry Christmas :-)

The windows CI does not pass, so maybe fix up a few small things for final CI:
- windows needs even looser than 1e-6 tolerances on these tests. I’m wondering about why, but I’m ok with it.
- Some files are missing final newlines now
- Ideally, start the name of the tests with a capital letter, so synthesized test classes have PascalCase.

Many thanks for this very cool addition! It is a common use case and will make for better reconstructions when robots are traveling on a plane!

dellaert:
Merged! Thanks again! FRC FTW :-)

Issue #1233, opened 2022-07-04
Regression in Logmap (v4.2a7)


lpea:
Hi all,

As discussed in https://groups.google.com/g/gtsam-users/c/7dEKY53G4Jk:

> After an upgrade from version 4.0.3 to 4.2a7, we've noticed a strange behavior in our unit tests which we were able to track down to back and forth conversions between AngleAxis and Rotation representations for rotations.

According to @ProfFan, this is a side-effect of #887. 

## Description

There is a precision error in Logmap for rotation angles close to PI. Furthermore, the conversion error exhibits a strong discontinuity.

## Steps to reproduce

1. Generate random test data in the form of 3D vectors with a norm close to PI.
2. Apply Logmap / Expmap to each vector.
3. Compare the vectors before / after this transformation − using the same metric as [Eigen::DenseBase::isApprox()](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ae8443357b808cd393be1b51974213f9c).

C++ code:

```c++
#include <gtsam/geometry/Rot3.h>
#include <fstream>
#include <iomanip>
#include <random>

template <typename Matrix>
double err(const Eigen::PlainObjectBase<Matrix>& v, const Eigen::PlainObjectBase<Matrix>& w)
{
  // Distance used in Eigen::DenseBase::isApprox(), see
  // https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ae8443357b808cd393be1b51974213f9c
  return (v - w).norm() / std::min(v.norm(), w.norm());
}

int main()
{
  std::srand(0);
  std::mt19937 gen(0);
  std::uniform_real_distribution<> dis(M_PI - 0.1, M_PI);
  
  std::ofstream ofs("val.txt");
  ofs << std::setprecision(10) << "norm(v),err\n";
  
  for (auto i = 0; i < 20000; ++i) {
    // Generate a 3D vector with a norm close to PI (in [PI - 0.1, PI))
    const gtsam::Vector3 v = dis(gen) * gtsam::Vector3::Random().normalized();
    const auto rot = gtsam::Rot3::Expmap(v);
    const auto v_calc = gtsam::Rot3::Logmap(rot);
    ofs << v.norm() << "," << err(v, v_calc) << "\n";
  }
}
```

Python script to generate the graphs:

```python
import numpy as np
import matplotlib.pyplot as plt

FONTSIZE = 18
plt.rc('font', size=FONTSIZE)  # controls default text size
[...]

PR #1938, opened 2024-12-16
Better SE(3) and SE2(3) Jacobians

dellaert:
Implements ExpmapTranslation with SO3::applyLeftJacobian and simplifies the Expmap jacobians in a way that also provides insight in how they relate to the applyLeftJacobian Jacobians. The key piece of code is this:
```c++
  Matrix3 H;
  Vector t = local.applyLeftJacobian(v, H); // = J_l * v !

  // We return Jacobians for use in Expmap, so we multiply with X, that
  // translates from left to right for our right expmap convention:
  Matrix3 X = local.rightJacobian() * local.leftJacobianInverse();
  Q = X * H;
```
This does three things:
- replaces very complicated expression $-0.5 * V + C * (W * V + V * W - WVW) + F * (WW * V + V * WW - 3 * WVW) - 0.5 * E * (WVW * W + W * WVW)$ from before
- explains why the bottom-right block of the Expmap Jacobian was $J_r$, not $J_l$
- functor in Pose3 is now no longer needed.

Smaller things:
- added applyLeftJacobianInverse and tested
- ExpmapTranslation no longer needs R
- Changed sign of D and E "constants" and renamed them to E and F.
- New D "constant" for use in inverses

### DIFF
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -22,6 +22,7 @@
 #include <cmath>
 #include <iostream>
 #include <string>
+#include "gtsam/geometry/Rot3.h"
 
 namespace gtsam {
 
@@ -241,25 +242,18 @@ Vector6 Pose3::ChartAtOrigin::Local(const Pose3& pose, ChartJacobian Hpose) {
 /* ************************************************************************* */
 namespace pose3 {
 struct GTSAM_EXPORT ExpmapFunctor : public so3::DexpFunctor {
-  // Constant used in computeQ
-  double F;  // (B - 0.5) / theta2 or -1/24 for theta->0
-
   ExpmapFunctor(const Vector3& omega, bool nearZeroApprox = false)
       : so3::DexpFunctor(omega, nearZeroApprox) {
-    F = nearZero ? _one_twenty_fourth : (B - 0.5) / theta2;
   }
 
-  // Compute the bottom-left 3x3 block of the SE(3) Expmap derivative
-  // TODO(Frank): t = applyLeftJacobian(v), it would be nice to understand
-  // how to compute mess below from applyLeftJacobian derivatives in w and v.
+  // Compute the bottom-left 3x3 block of the SE(3) Expmap derivative.
   Matrix3 computeQ(const Vector3& v) const {
-    const Matrix3 V = skewSymmetric(v);
-    const Matrix3 WVW = W * V * W;
-    return -0.5 * V + C * (W * V + V * W - WVW) +
-           F * (WW * V + V * WW - 3 * WVW) - 0.5 * E * (WVW * W + W * WVW);
+    // X translate from left to right for our right expmap convention:
+    Matrix X = rightJacobian() * leftJacobianInverse();
+    Matrix3 H;
+    applyLeftJacobian(v, H);
+    return X * H;
   }
-
-  static constexpr double _one_twenty_fourth = - 1.0 / 24.0;
 };
 }  // namespace pose3
 
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -164,22 +164,22 @@ Pose3 Pose3::interpolateRt(const Pose3& T, double t) const {
 }
 
 /* ************************************************************************* */
-/** Modified from Murray94book version (which assumes w and v normalized?) */
 Pose3 Pose3::Expmap(const Vector6& xi, OptionalJacobian<6, 6> Hxi) {
   // Get angular velocity omega and translational velocity v from twist xi
   const Vector3 w = xi.head<3>(), v = xi.tail<3>();
 
   // Compute rotation using Expmap
-  Matrix3 Jw;
-  Rot3 R = Rot3::Expmap(w, Hxi ? &Jw : nullptr);
+  Matrix3 Jr;
+  Rot3 R = Rot3::Expmap(w, Hxi ? &Jr : nullptr);
 
-  // Compute translation and optionally its Jacobian in w
+  // Compute translation and optionally its Jacobian Q in w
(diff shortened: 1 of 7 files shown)


gtsam/geometry/SO3.h
```diff
@@ -157,10 +157,16 @@ struct GTSAM_EXPORT ExpmapFunctor {
...
+  // Ethan's C constant used in Jacobians
+  double C;  // (1 - A) / theta^2 or 1/6 for theta->0
+
+  // Constant used in inverse Jacobians
+  double D;  // (1 - A/2B) / theta2 or 1/12 for theta->0
+
   // Constants used in cross and doubleCross
-  double D;  // (A - 2.0 * B) / theta2 or -1/12 for theta->0
-  double E;  // (B - 3.0 * C) / theta2 or -1/60 for theta->0
+  double E;  // (A - 2.0 * B) / theta2 or -1/12 for theta->0
```
dellaert: 1/12 and 1/60 now, without negatives

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -900,17 +900,6 @@ TEST(Pose3, ExpmapDerivative4) {
   }
 }
 
-TEST( Pose3, ExpmapDerivativeQr) {
```
dellaert, line 903: I should actually look into detail why this test failed. It might be invoking nearZero for not small enough values.

gtsam/geometry/tests/testSO3.cpp
```diff
@@ -19,6 +19,7 @@
 #include <gtsam/base/Testable.h>
 #include <gtsam/base/testLie.h>
 #include <gtsam/geometry/SO3.h>
+#include <iostream>
```
varunagrawal: Kill if not used?

gtsam/geometry/tests/testPose3.cpp
varunagrawal, line 903: Yeah I'm surprised we are removing a test.

gtsam/navigation/NavState.cpp
```diff
@@ -114,24 +114,23 @@ NavState NavState::inverse() const {
 
 //------------------------------------------------------------------------------
 NavState NavState::Expmap(const Vector9& xi, OptionalJacobian<9, 9> Hxi) {
-  // Get angular velocity w, translational velocity v, and acceleration a
-  Vector3 w = xi.head<3>();
-  Vector3 rho = xi.segment<3>(3);
-  Vector3 nu = xi.tail<3>();
+  // Get angular velocity w and components rho and nu from xi
```
varunagrawal: $\rho$ should be translation and $\nu$ should be velocity and that should be mentioned in this comment IMHO.

varunagrawal (approved):

varunagrawal:
Tentative approval but that unit test being killed is bothering me.

gtsam/geometry/Pose3.cpp
```diff
@@ -23,6 +23,8 @@
 #include <iostream>
 #include <string>
 
+#include "gtsam/geometry/Rot3.h"
```
ProfFan:
```suggestion
#include <gtsam/geometry/Rot3.h>
```

gtsam/geometry/Pose3.cpp
```diff
@@ -158,26 +160,26 @@ bool Pose3::equals(const Pose3& pose, double tol) const {
 /* ************************************************************************* */
 Pose3 Pose3::interpolateRt(const Pose3& T, double t) const {
   return Pose3(interpolate<Rot3>(R_, T.R_, t),
-                interpolate<Point3>(t_, T.t_, t));
+               interpolate<Point3>(t_, T.t_, t));
 }
 
 /* ************************************************************************* */
-/** Modified from Murray94book version (which assumes w and v normalized?) */
```
ProfFan, line 165: Citation, maybe also change the 'lyx' files?

gtsam/geometry/tests/testPose3.cpp
```diff
@@ -900,17 +900,6 @@ TEST(Pose3, ExpmapDerivative4) {
...
-  Vector6 w = Vector6::Random();
```
ProfFan, line 904: Change this to a non random value instead?

ProfFan (changes requested):
I think it's better that we stop and think about the `nearZero` throughly. As I mentioned in #1233, we have to be very careful in these approximations.

for example, the expression for D here has a 1000x error when theta is small.

ProfFan:
[Lie Derivatives.pdf](https://github.com/user-attachments/files/18157142/Lie.Derivatives.pdf)

output-01
output-02
output-03
output-04
output-05

ProfFan:
Particularly D and E are problematic.

image

dellaert:
@ProfFan I agree we have to look into it - and that issue is amazing - but I'd rather have a separate PR to deal with the nonZero thresholds. I will add the more thorough tests in this PR, but would like to merge after that.

I will add a citation to Ethan Eade's document as I am using his A,B,C constants now.

ProfFan (approved):
Cool cool

dellaert:
PS, why did you say "The cutoff should be 1 × 10−3 to use the Taylor expansion" for A? And, I don't understand what you are plotting, actually...

gtsam/navigation/NavState.cpp
dellaert: Done

gtsam/geometry/tests/testPose3.cpp
dellaert: no longer relevant, added more systematic tests

gtsam/geometry/Pose3.cpp
dellaert: Referenced Ethan Eade. Lyx: later

PR #1939, opened 2024-12-17
Add `serialize()` to various preintegration classes

### LINKED ISSUE #1931: issue with copying gtsam objects in python
Hello, there is a problem with copying items in Python 3.11 and GTSAM API. 
Some objects can be copied and deepcopied successfully, some others - no. 

For example, gtsam.NonlinearFactorGraph() instance can be successfully copied.

whereas copying gtsam.PreintegratedCombinedMeasurements() instance  leads to errors in deepcopy module of Python.

as well as gtsam.PreintegratedImuMeasurements . **BUT** here is more interesting thing. If i create gtsam.NonlinearFactorGraph() instance, the gtsam.PreintegratedImuMeasurements() can be copied:

and If i remove the creation of FactorGraph, the pim can not be copied anymore.

the same with  gtsam.PreintegrationParams: if gtsam.NonlinearFactorGraph() has been called before, params can be copied, otherwise - No. 

Also if one adds this IMU factor in NonlinearFactorGraph, the copying error occurs and it is not possible to copy the graph anymore. Meanwhile if one adds, for example, gtsam.PriorFactorPose3() to the graph, it can be copied successfully.

This issue is strange and maybe somehow connected with calling gtsam.NonlinearFactorGraph() before. However, for gtsam.PreintegratedCombinedMeasurements() instance the pre-calling of the graph still does not  allow to copy as "pickle error" occurs

this is clean python 3.11 and gtsam  4.2 installed with pip.

varunagrawal:
Fixes #1931

### DIFF
--- a/gtsam/navigation/AHRSFactor.h
+++ b/gtsam/navigation/AHRSFactor.h
@@ -139,9 +139,6 @@ class GTSAM_EXPORT AHRSFactor: public NoiseModelFactorN<Rot3, Rot3, Vector3> {
 
   PreintegratedAhrsMeasurements _PIM_;
 
-  /** Default constructor - only use for serialization */
-  AHRSFactor() {}
-
 public:
 
   // Provide access to the Matrix& version of evaluateError:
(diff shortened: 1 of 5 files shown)


varunagrawal:
~~On further analysis, this might need the wrapper to be updated for deep copy support: https://pybind11.readthedocs.io/en/stable/advanced/classes.html#deepcopy-support~~

The issue is to do with the ImuFactor and the PIM needing at least one measurement integrated.

ProfFan (approved):

PR #1943, opened 2024-12-23
Various Improvements

varunagrawal:
1. ~~Suppress `maybe-uninitialized` warnings from Eigen for Release build. We only see this warning in Eigen, and since this is a vendored version, the assumption is that the original developers have already accounted for this to not be a problem.~~
2. Fix some docstrings.
3. Use `resetIntegration` to initialize the `ImuFactor` and `CombinedImuFactor` classes.
4. Wrap `ImuFactor2`.
5. ~~Added a new macro `GTSAM_COUT` which is for classes/types that don't have a `.print()` defined or have the `<<` operator overloaded. An example usage is below:~~

```cpp
Rot3 wRb = Rot3();
GTSAM_COUT(wRb);
```
which prints
```shell
wRB: 1, 0, 0  // Note that it also prints the variable name similar to GTSAM_PRINT.
0, 1, 0
0, 0, 1
```

Interestingly, there is no `CombinedImuFactor2` which works on `NavState`s. This should be a TODO for us given the recent importance of NavState.

### DIFF
--- a/gtsam/3rdparty/Eigen/Eigen/src/Core/util/DisableStupidWarnings.h
+++ b/gtsam/3rdparty/Eigen/Eigen/src/Core/util/DisableStupidWarnings.h
@@ -66,6 +66,7 @@
   #endif
   #if __GNUC__>=6
     #pragma GCC diagnostic ignored "-Wignored-attributes"
+    #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
   #endif
   #if __GNUC__==7
     // See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89325
--- a/gtsam/3rdparty/Eigen/Eigen/src/Core/util/DisableStupidWarnings.h
+++ b/gtsam/3rdparty/Eigen/Eigen/src/Core/util/DisableStupidWarnings.h
@@ -66,7 +66,6 @@
   #endif
   #if __GNUC__>=6
     #pragma GCC diagnostic ignored "-Wignored-attributes"
-    #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
   #endif
   #if __GNUC__==7
     // See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89325
--- a/gtsam/base/Testable.h
+++ b/gtsam/base/Testable.h
@@ -41,7 +41,6 @@
 #include <string>
 
 #define GTSAM_PRINT(x)((x).print(#x))
-#define GTSAM_COUT(x) (std::cout << #x ": " << x << std::endl)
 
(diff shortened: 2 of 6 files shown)


varunagrawal:
@dellaert reminder

gtsam/3rdparty/Eigen/Eigen/src/Core/util/DisableStupidWarnings.h
```diff
@@ -66,6 +66,7 @@
   #endif
   #if __GNUC__>=6
     #pragma GCC diagnostic ignored "-Wignored-attributes"
+    #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```
dellaert: That feels very dangerous. Needs explanation in PR comment why this is the correct thing here

varunagrawal: Updated the PR description.

gtsam/base/Testable.h
```diff
@@ -41,6 +41,7 @@
 #include <string>
 
 #define GTSAM_PRINT(x)((x).print(#x))
+#define GTSAM_COUT(x) (std::cout << #x ": " << x << std::endl)
```
dellaert: But print already assumes cout

varunagrawal: This is more for types that don't support `.print()`. I can use this to print doubles, strings etc.

dellaert: Then just use std::cout, please.

varunagrawal: Wouldn't it be convenient to do `GTSAM_COUT(x)` instead of `std::cout << "x: " << x << std::endl;`?

dellaert: Not a GTSAM thing

gtsam/3rdparty/Eigen/Eigen/src/Core/util/DisableStupidWarnings.h
Gold856: Should be unnecessary now that Eigen headers are labeled as system headers, which causes all warnings from Eigen to be suppressed.

varunagrawal: Yeah thanks for that! We had contemplated doing it before but I forget why we didn't.

gtsam/base/Testable.h
varunagrawal: It's not a GTSAM thing, but more of a helper thing to make debugging and analysis easier. Oh well.

dellaert (approved):
LGTM

PR #1944, opened 2024-12-23
Improve build time when updating Boost flags

varunagrawal:
I came across an [interesting article](https://www.kdab.com/setting-defines-with-cmake/) that gives recommendations on how to set preprocessor flags such that it improves build time.

The key idea is to get rid of `add_definitions` calls in CMake since it passes the preprocessor flag to **all** cpp files, and instead use a definition within the `config.h.in` file. This would then only require re-compiling files that include `gtsam/config.h` and not the complete project in the event the flags were updated.

Since we already have the `config.h.in`, this was an easy update.

I recommend reading the article, since it also discusses the benefits of using `#cmakedefine01` over `#cmakedefine`.


dellaert:
LGTM but maybe get review from @jlblancoc ? ROS interactions?

jlblancoc:
Hi, 
I totally agree with this change, @varunagrawal !! 

My only comment is: wouldn't the ``#ifdef XX`` need to be ``#if XXX`` if using ``#cmakedefine01`` ? (didn't test myself)

varunagrawal:
@jlblancoc great catch! I'll update it.

jlblancoc (approved):
Now, it LGTM!

PR #1946, opened 2024-12-24
Add BUILD_SHARED_LIBS as a configurable option

Gold856:
`GTSAM_FORCE_SHARED_LIB` defaulting to `ON` is unintuitive. This means that building static libraries requires setting `GTSAM_FORCE_SHARED_LIB` to false and `GTSAM_FORCE_STATIC_LIB` to true. It would be significantly more natural to set `BUILD_SHARED_LIBS` once, and only reach into the `FORCE` options if needed. By adding `BUILD_SHARED_LIBS` as a configurable option defaulting to `ON` and setting `GTSAM_FORCE_SHARED_LIB` to `OFF`, backwards compatibility is maintained, as shared libraries are still the default and the `FORCE` options work as intended, but now only `BUILD_SHARED_LIBS` needs to be set to select a build type.


varunagrawal:
1. Either `GTSAM_FORCE_SHARED_LIB` or `GTSAM_FORCE_STATIC_LIB` should be enabled. We can't have both OFF by default currently.
2. Another option is to get rid of these Cmake flags altogether and use only `BUILD_SHARED_LIBS`, with it being set to ON by default. This is tricky since it may break backwards compatibility.
3. The third option is to add a check that `GTSAM_FORCE_SHARED_LIB` is the same as `BUILD_SHARED_LIBS` and error out if it is not.

My understanding is that we have custom flags since it gives us more control over the subprojects. I may be wrong and if we can deem these custom flags unnecessary, that may be a good thing.

Gold856:
What if we add `BUILD_SHARED_LIBS` as an option with the default value being `ON`, with `GTSAM_FORCE_SHARED_LIB` defaulting to `OFF`? [CMake docs](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) suggest this, and it's exactly the same behavior (shared libraries by default), with the option of forcing it to shared or static somewhere else if desired.

Gold856:
I added `BUILD_SHARED_LIBS` as an option and also added an error if both `GTSAM_FORCE_SHARED_LIB` and `GTSAM_FORCE_STATIC_LIB` were true. I think this is the best way to provide control over library type, since it allows `BUILD_SHARED_LIBS` to be used while allowing the `FORCE` options to override it for advanced use cases. I'm still open to other suggestions though, so let me know!

varunagrawal:
I'm a bit confused. Doesn't adding `BUILD_SHARED_LIBS` as an option now do the same thing as just using `GTSAM_FORCE_SHARED_LIB`?
I was reading the CMake docs and it says that this option has to be provided by the project and is not defined by default. SInce GTSAM doesn't provide `BUILD_SHARED_LIBS` as an option, we should probably point users to using `GTSAM_FORCE_SHARED_LIB` as an equivalent variable.

jlblancoc:
I think @Gold856 's point is that CMake's standard is BUILD_SHARED_LIBS, and most experienced users should be aware of it, while the others are GTSAM's custom cmake variables. 

This PR is now IMO good to go as a workaround to keep backwards compatibility... but, thinking of a next release, it might probably be just easier to provide BUILD_SHARED_LIBS as in this PR, and remove all the logic around static/dynamic libraries and let CMake to just follow BUILD_SHARED_LIBS. Docs should then be updated, but it would make gtsam to behave as a "standard" CMake package.

Gold856:
Yeah, that was the idea. I use `BUILD_SHARED_LIBS` all the time to define library type, and a bunch of projects that I work on use it as well, so it's a little inconvenient that gtsam ignores it completely. And it would be nice one day to have the custom variables removed (maybe not now though.)

dellaert (approved):
I like this if @jlblancoc likes this :-)

Gold856:
All checks should pass now. Apparently building metis as a shared library was broken on Windows due to symbols not being exported, which I fixed by turning on `WINDOWS_EXPORT_ALL_SYMBOLS` for METIS.

PR #1947, opened 2024-12-25
Discrete Improvements

varunagrawal:
1. Undid changes to `TableFactor::toDecisionTreeFactor` since it is giving an incorrect result for a hybrid elimination case with 30+ discrete variables.
2. Renamed `normalization` to `denominator`.
3. Pass in discrete conditional directly instead of re-converting it.


varunagrawal:
@dellaert reminder

dellaert (approved):
Approve modulo discussion on “normalizer”.

PR #1948, opened 2024-12-27
Hybrid Timing

varunagrawal:
Add machinery for timing and generating CSVs of runtimes for various hybrid functions and classes.


varunagrawal:
TBB based environments are failing because of a race condition.

gtsam/base/timing.cpp
```diff
@@ -106,6 +106,56 @@ void TimingOutline::print(const std::string& outline) const {
 #endif
 }
 
+/* ************************************************************************* */
+void TimingOutline::print_csv_header(bool addLineBreak) const {
```
dellaert: naming convention: printCvsHeader

gtsam/base/timing.cpp
```diff
@@ -106,6 +106,56 @@ void TimingOutline::print(const std::string& outline) const {
...
+#ifdef GTSAM_USE_BOOST_FEATURES
```
dellaert, line 112: Why can this only work with boost?

gtsam/base/timing.cpp
```diff
@@ -106,6 +106,56 @@ void TimingOutline::print(const std::string& outline) const {
...
+  }
+  if (addLineBreak) {
+    std::cout << std::endl;
+  }
+  std::cout.flush();
+#endif
+}
+
+/* ************************************************************************* */
+void TimingOutline::print_csv(bool addLineBreak) const {
```
dellaert: Same two comments here

gtsam/base/timing.h
```diff
@@ -268,6 +291,14 @@ inline void tictoc_finishedIteration_() {
 inline void tictoc_print_() {
   ::gtsam::internal::gTimingRoot->print(); }
 
+// print timing in CSV format
+inline void tictoc_print_csv_(bool displayHeader = false) {
```
dellaert: naming

gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -121,15 +121,25 @@ namespace gtsam {
   static DecisionTreeFactor ProductAndNormalize(
       const DiscreteFactorGraph& factors) {
     // PRODUCT: multiply all factors
-    gttic(product);
+#if GTSAM_HYBRID_TIMING
```
dellaert: That's a bit weird. So if timing is on, but hybrid is off, we will get all timing *but* hybrid? gttic and gttoc are designed to be unobtrusive and resolve to empty if the timing flag is off.

gtsam/hybrid/HybridGaussianISAM.cpp
```diff
@@ -104,7 +104,13 @@ void HybridGaussianISAM::updateInternal(
       elimination_ordering, function, std::cref(index));
 
   if (maxNrLeaves) {
+#if GTSAM_HYBRID_TIMING
```
dellaert, line 107: same

gtsam/discrete/DiscreteFactorGraph.cpp
varunagrawal: I was thinking about it the other way: if timing is on, we get far too much information which isn't helpful. This way we can filter out hybrid timing information.

gtsam/base/timing.cpp
varunagrawal: I used the same form as `TimingOutline::print`. I didn't think too deeply about this since I assumed `print` had it for a reason.

dellaert (approved):

dellaert (commented):
All of a sudden this PR has non timing-related changes.

varunagrawal:
> All of a sudden this PR has non timing-related changes.

I think that's okay. This timing PR has helped us work on improving efficiency so it is still under the same umbrella.

dellaert:
I'm worried it's a merging branches issue: I see product defined here, and serialization. I don;t think you intended this in this PR? I'd rather this be pure timing.

dellaert:
Oh, I think I remember what you meant to do: everything goes into timing and then we merge all changes at the same time?

varunagrawal:
> Oh, I think I remember what you meant to do: everything goes into timing and then we merge all changes at the same time?

Yes that was what I was hoping to do, in case we needed to rollback or fix some things. Once we're happy with the timing of hybrid inference overall, this can be merged. :)

varunagrawal:
Also you shouldn't have to worry about branch merging when it comes to me. My git-fu is strong. :)

PR #1950, opened 2024-12-29
Fix deprecated copies and redundant moves

mcm001:
While working on https://github.com/wpilibsuite/allwpilib/pull/7606 I had to disable deprecated copy and redundant move warnings. I'm not super great with this aspect of C++, but I've added default constructors where they were previously implicitly declared.﻿

### DIFF
--- a/gtsam/base/ConcurrentMap.h
+++ b/gtsam/base/ConcurrentMap.h
@@ -85,6 +85,8 @@ public:
   /** Copy constructor from the base map class */
   ConcurrentMap(const Base& x) : Base(x) {}
 
+  ConcurrentMap& operator=(const ConcurrentMap& other) = default;
+
   /** Handy 'exists' function */
   bool exists(const KEY& e) const { return this->count(e); }
 
--- a/gtsam/base/FastList.h
+++ b/gtsam/base/FastList.h
@@ -62,6 +62,8 @@ public:
   /// Construct from c++11 initializer list:
   FastList(std::initializer_list<VALUE> l) : Base(l) {}
 
+  FastList& operator=(const FastList& other) = default;
+
 #ifdef GTSAM_ALLOCATOR_BOOSTPOOL
   /** Copy constructor from a standard STL container */
   FastList(const std::list<VALUE>& x) {
--- a/gtsam/base/FastMap.h
+++ b/gtsam/base/FastMap.h
@@ -54,6 +54,8 @@ public:
   /** Copy constructor from another FastMap */
   FastMap(const FastMap<KEY,VALUE>& x) : Base(x) {}
 
+  FastMap& operator=(const FastMap<KEY,VALUE>& x) = default;
+
   /** Copy constructor from the base map class */
   FastMap(const Base& x) : Base(x) {}
 
--- a/gtsam/base/FastSet.h
+++ b/gtsam/base/FastSet.h
@@ -80,6 +80,8 @@ public:
   Base(x) {
   }
 
+  FastSet& operator=(const FastSet& other) = default;
+
 #ifdef GTSAM_ALLOCATOR_BOOSTPOOL
   /** Copy constructor from a standard STL container */
   FastSet(const std::set<VALUE>& x) {
--- a/gtsam/base/GenericValue.h
+++ b/gtsam/base/GenericValue.h
@@ -56,9 +56,10 @@ public:
   GenericValue(){}
 
   /// Construct from value
-  GenericValue(const T& value) :
-      value_(value) {
-  }
+  GenericValue(const T& value) : Value(),
+      value_(value) {}
+
+  GenericValue(const GenericValue& other) = default;
 
   /// Return a constant value
   const T& value() const {
@@ -112,7 +113,7 @@ public:
      * Clone this value (normal clone on the heap, delete with 'delete' operator)
      */
     std::shared_ptr<Value> clone() const override {
-		return std::allocate_shared<GenericValue>(Eigen::aligned_allocator<GenericValue>(), *this);
+      return std::allocate_shared<GenericValue>(Eigen::aligned_allocator<GenericValue>(), *this);
     }
 
     /// Generic Value interface version of retract
--- a/gtsam/base/Value.h
+++ b/gtsam/base/Value.h
@@ -38,6 +38,9 @@ namespace gtsam {
    */
   class GTSAM_EXPORT Value {
   public:
+    // todo - not sure if valid
+    Value() = default;
+    Value(const Value& other) = default;
 
     /** Clone this value in a special memory pool, must be deleted with Value::deallocate_, *not* with the 'delete' operator. */
     virtual Value* clone_() const = 0;
--- a/gtsam/discrete/SignatureParser.cpp
+++ b/gtsam/discrete/SignatureParser.cpp
@@ -38,7 +38,7 @@ std::optional<Row> static ParseConditional(const std::string& token) {
   } catch (...) {
     return std::nullopt;
   }
-  return std::move(row);
+  return row;
 }
 
 std::optional<Table> static ParseConditionalTable(
@@ -62,7 +62,7 @@ std::optional<Table> static ParseConditionalTable(
       }
     }
   }
-  return std::move(table);
+  return table;
 }
 
 std::vector<std::string> static Tokenize(const std::string& str) {
--- a/gtsam/geometry/Pose2.h
+++ b/gtsam/geometry/Pose2.h
@@ -60,7 +60,10 @@ public:
   }
 
   /** copy constructor */
-  Pose2(const Pose2& pose) : r_(pose.r_), t_(pose.t_) {}
+  Pose2(const Pose2& pose) = default;
+  //  : r_(pose.r_), t_(pose.t_) {}
+
+  Pose2& operator=(const Pose2& other) = default;
 
   /**
    * construct from (x,y,theta)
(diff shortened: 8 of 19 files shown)


varunagrawal:
Doesn't C++ automatically add these if they're not defined? I may be wrong though.

mcm001:
Yes, but if you have deprecated copy warnings enabled they'll reported warnings. I'm trying to find a primary source, but per gcc's [bugzilla](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58407),

"The implicit definition of a copy constructor as defaulted is deprecated if the class has a user-declared copy assignment operator or a user-declared destructor. The implicit definition of a copy assignment operator as defaulted is deprecated if the class has a user-declared copy constructor or a user-declared destructor"

TLDR: compile with -Wdeprecated-copy and observe compiler output

mcm001:
Although this might be worth a larger discussion. What warnings does GTSAM currently compile with? Is there any interest in enabling more compiler warnings/enabling Werror?

varunagrawal:
Okay I'm going to need your help understanding this better. I currently don't get these warnings so is it a result of a newer version of GCC?

In general we ignore warnings since most of the time they don't reflect a real problem. I'm curious if suppressing these warnings are sufficient in your use case?

My guiding philosophy here is less code is good code.

dellaert:
> 
> In general we ignore warnings since most of the time they don't reflect a real problem. I'm curious if suppressing these warnings are sufficient in your use case?

Disagree. If we can -Wall so much the better!

> My guiding philosophy here is less code is good code.

Totally agree here, but I have a suspicion that this extra code might actually benefit performance even for other compilers? any idea, @mcm001 ?

mcm001:
Hey! At least in [some Godbolt testing](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAGwAHKSuADJ4DJgAcj4ARpjEEtIADqgKhE4MHt6%2BAcGp6Y4C4ZExLPGJUraY9kUMQgRMxATZPn5B1bWZDU0EJdFxCUm2jc2tuR0Ko30RA%2BVDUgCUtqhexMjsHOYAzBHI3lgA1Cbbbl6OtIQAnifYJhoAgjt7B5jHp2wsJDfbd4/3DymxC8DkOABEmI03AYFApjgB2KyPQ4ow4RAiHVQnJEPVGHdCrWL0Q4/HEAvEQqEwhQQdGYgD6pHxhOJV3pi2OyLxqJAmJMiNU9P5YKZN0RbOFCKs8LBAPJqMpTGhTFhEDQDCm4MhSup5n8h1QBAQCUWcrMuyo4OwygAStg3A8ACrYMH0twAeWUAE15SiAPR%2Bw7ILxTVAsIOoZJXCMagjAhwkQ4AWkOwOJqEtAFZDgA3JwGOq%2Bw4Bw5UYhhw4IAgEZIKEABxgAOmQyWSxEwVASjA2zbDfoA7n6W8k/QYjF4mMBMH605h6Rn6Ub25gi/zpbL/uaakoiyWwR2mF5aI4jIcjW80FGY0CQURiIc2IY4f2L4ZDlM6LRd4H4odMKpkjwZBCCZWJznfC5aEOegMXPCMWEA%2Bh7ywGhIjRBCvy5FETjBfEDyPAhsTNbZXDwKg5RlIj/keKZ0BAEAFAQJpMHQelkjjE43EVZVYVuB8mAAa0wB4jQiYAIA5Nci0PIgzwQMT3lw2j6JYQS50Y5j0E47jdV%2BCAuCZMxG22U1tjJTcLUOO0wTkKIwQeKJHXpABZd0ADU/lxVES1EuEp0iYhIUwOEk0iJpaCuFM4IUVg3lU5B5LQtAsFIb9DjAjE8AUBgwA4DFIg2WEmhuLDU0wAg1gYCC6JAL4c0wCBRKMUyyS3WglE5LyUXbCriCqprgCop4twYfByP%2BSjqIeOlVIiCSpSLZSGKY9tWPY4htO1HiFD4gbFP4oSRMS8SWrlUqZNQQ4mH2nSVR2xFpEOTNG38YUhrxC70purbdURMwmS4F63rM6T9tiIa3seDhlloThM14PwOC0UhUE4NxrGsd9VnWN4dh4UgCE0aHlgEkAzHhRsNDMLhAkzeEAE5zUken4XJ/7YY4SQEaJlHOF4OsNAJonljgWAYEQFAw0QhJyEoNAELoIZgC4Mx/poY8EjrCBYh52IImKzh8b15hiCud1Ym0TAHEN3h5bYQR3QYCKeawMDgDcMR2pt0gsFUoxxCR3h8HbBw8HqutA5CVQrfOTZkfRGoecuWJAtNjwsB5uM8BYb36uIWI0kwfc/eAS4jGFvgDGABQ3LwTB%2B3dZJGG9/hBBEMR2CqVv5CUNQed0Ayx2MDHLH0PBYjrSBlkjOoI6Td1tmTWicNMSxrDMDRkwAdU97esDbK2gvQJNL2jJMwWs2z7Mclz3LubYwS4ZMwTBa07QdZ1XQ9b0cPx1A8%2BIHgLAk8JKdCtnUFwo1xh%2BAMmEWYZQKh6AKBkAQ0CkFpBQQwfoCCFhgNDgIHoYxPBtD0HYcB3RpjYMGJUEYvQ0EGSmL0Kh8xKjLAUNjDYEgYZw25pHVGHBMSBH8EmfwkhDjAGQMgQ4KtGxmEOBAXAhBEx40WLwQmgdFjLGNEwLAiRQGk38EZQxZh6aZg0IELg9N/Csy4NsfQnAuakBzmY0giNkb8P5iAQW6itAi3FhAJAqwaznFlmqKWitiBRFipwVQQiRFiIkVI2xsjeAsSUYAuiBlu7t3EF3WQigVDqEjgPUg/ZArJBttwjg8NXE834e6c4yRwIZkEcI0R4jJHSKMnIiAHgFZIWOOaLgqihYaK0ZgHRQxQEc0cc4wWbjeAeNsF40ZviSYgG2PTF6khJCBHMfCSQ2wuD%2BHptIDm2xeHuL5qs6GpBRZQH8UgeW0syAUDCf0pWUg1Z0AIJrSgOtI7GwNtwXgQLTbm0ttbEFpA7aMAII7Z2kdXZeHdp7CO%2BNfaGFLnHIOeAQ6OHDjzf8MdfnewThzZGydU5XHTjigmgCc7QrzgXJQxcsVl1ABoyuk4a51wbk3RG%2BNsmiFydIbuBS%2B7FLJvoLFKAR42GTiA6e7FMhzwXkvAgWkH6rysJYDe29d5Ji3vvEOR8T6RjPhfF0V8HJOVch5HCm9z6v1tPaJ0Lo3Seh9A/dxACgHBXgGwmo5DnAQFcPQkIo1mGIIMsguoEa42ZGjbgsh%2BD6jTAjamuohCZilGoaQjNxDcgMMofA/NwyVhrE4RWjmNSFm8wEbEtpCTOlSFkfIxRd5BlHJGT44mpBSbbG2MZIdo6x1jvsZzXgczal8OuQLG5fj4ABJAEEppBBQnPIiVEtgMS4ntMSdIyQKSQj4DvEAvQwqO7DHFb3IpyNdD/TKUwCpIKql1rqZwBpwSMQtKbfEjpSTj09L6S87tZhe3C3GZMvRVTZkgBcfWpZC6%2B2aIHRs4d46sOjsnRc2dVyOBqKg5Osw06JAaHmZ%2Bwji7lh53SM4SQQA%3D), I don't see a meaningful difference in the actual binary output of this contrived test program with GCC 12.2 on x86 or arm64. I think this just comes down to if y'all are in favor of fixing warnings even if the code output is (probably) correct. I think that fixing rule-of-3/5/0-related warnings is a good idea, and deleting redundant moves also lets us write less code ;)

dellaert (approved):
I’m in favor of -Wall so approving. We should probably add a gcc12 CI, maybe instead of gcc11. Not sure whether gcc9 is already EOL?

mcm001:
I've also got a tiny patchset that lets me build with -Wall -Wpedantic on GCC 12, should be ready here in a moment. Happy to add that in a follow-on PR, or add to this one

mcm001:
And from https://gcc.gnu.org/gcc-9/ and https://gcc.gnu.org/gcc-10/ , it appears GCC 9/10 are EOL yeah

dellaert:
Awesome ! Maybe do it in a follow-up PR where you also switch 11 to 12 in CI?
I know some folks were using gcc9 on some embedded platforms but maybe we could switch that to 10…

mcm001:
Sounds good! Will add in a follow-on

varunagrawal:
Wanted to clarify my comment on suppressing warnings.

A lot of warnings right now are from 3rd party libraries (due to lack of updates from them) or from invalid type mismatches (e.g. size_t vs Eigen::Index). I prefer suppressing these because they are either frivolous or we don't own that code.
And this is from someone who has fixed a lot of these warnings over the years.

My personal issue with warnings is that they clog error messages and make identifying the true issue difficult.

As such, we display warnings only in Debug mode where more information may be relevant.

I am up for using -Wall, but there should be an equal effort to fix those warnings and not ignore them, otherwise what is the point? 😊

PR #1953, opened 2024-12-31
TableFactor and TableDistribution

varunagrawal:
Use the `TableFactor` for discrete elimination and use a new `TableDistribution` which maintains an underlying representation of the distribution using the same `SparseVector` mechanism.


gtsam/discrete/DiscreteTableConditional.cpp
```diff
@@ -0,0 +1,203 @@
...
+  keys_.insert(keys_.end(), orderedKeys.begin(), orderedKeys.end());
+}
+
+/* ************************************************************************** */
+DiscreteTableConditional::DiscreteTableConditional(const Signature& signature)
+    : BaseConditional(1, DecisionTreeFactor()),
+      table_(TableFactor(signature.discreteKeys(), signature.cpt())) {}
+
+/* ************************************************************************** */
+DiscreteTableConditional DiscreteTableConditional::operator*(
```
dellaert: Why is this method implemented? I thought we were going for prune only?

gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -64,10 +65,17 @@ namespace gtsam {
   }
 
   /* ************************************************************************ */
-  DecisionTreeFactor DiscreteFactorGraph::product() const {
-    DecisionTreeFactor result;
+  TableFactor DiscreteFactorGraph::product() const {
```
dellaert: why is this a TableFactor now? I don't think we want to switch *all* discrete inference to TableFactors

gtsam/discrete/DiscreteConditional.h
```diff
@@ -214,6 +214,15 @@ class GTSAM_EXPORT DiscreteConditional
...
   size_t argmax(const DiscreteValues& parentsValues = DiscreteValues()) const;
 
+  /**
+   * @brief Create new conditional by maximizing over all
+   * values with the same separator.
+   *
+   * @param keys The keys to sum over.
+   * @return DiscreteConditional::shared_ptr
+   */
+  virtual DiscreteConditional::shared_ptr max(const Ordering& keys) const;
```
dellaert: What does this mean? What is the math in terms of probability distributions?

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -52,15 +52,22 @@ HybridBayesNet HybridBayesNet::prune(size_t maxNrLeaves) const {
...
   }
 
   // Prune the joint. NOTE: again, possibly quite expensive.
-  const DecisionTreeFactor pruned = joint.prune(maxNrLeaves);
+  const DiscreteConditional::shared_ptr pruned = joint.prune(maxNrLeaves);
 
   // Create a the result starting with the pruned joint.
   HybridBayesNet result;
-  result.emplace_shared<DiscreteConditional>(pruned.size(), pruned);
+  result.push_back(std::move(pruned));
```
dellaert: Not sure whether moving a shared ptr is worth it

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -126,7 +133,14 @@ HybridValues HybridBayesNet::optimize() const {
 
   for (auto &&conditional : *this) {
     if (conditional->isDiscrete()) {
-      discrete_fg.push_back(conditional->asDiscrete());
+      if (auto dtc = conditional->asDiscrete<TableDistribution>()) {
```
dellaert, line 133: did asDiscrete always have a template argument?

gtsam/hybrid/HybridBayesTree.cpp
```diff
@@ -52,8 +72,10 @@ HybridValues HybridBayesTree::optimize() const {
 
   //  The root should be discrete only, we compute the MPE
   if (root_conditional->isDiscrete()) {
-    discrete_fg.push_back(root_conditional->asDiscrete());
-    mpe = discrete_fg.optimize();
+    auto discrete = std::dynamic_pointer_cast<TableDistribution>(
+        root_conditional->asDiscrete());
```
dellaert: would this then be asDiscrete<TableDistribution> ?

gtsam/hybrid/HybridBayesTree.cpp
```diff
@@ -41,6 +42,25 @@ bool HybridBayesTree::equals(const This& other, double tol) const {
...
 }
 
+/* ************************************************************************* */
+DiscreteValues HybridBayesTree::discreteMaxProduct(
+    const DiscreteFactorGraph& dfg) const {
+  TableFactor product = TableProduct(dfg);
+
+  uint64_t maxIdx = 0;
+  double maxValue = 0.0;
+  Eigen::SparseVector<double> sparseTable = product.sparseTable();
```
dellaert: breaks encapsulation. Should this be inside TableDistribution?

gtsam/hybrid/HybridGaussianConditional.h
```diff
@@ -235,7 +236,7 @@ class GTSAM_EXPORT HybridGaussianConditional
    * @return Shared pointer to possibly a pruned HybridGaussianConditional
    */
   HybridGaussianConditional::shared_ptr prune(
-      const DecisionTreeFactor &discreteProbs) const;
+      const DiscreteConditional::shared_ptr &discreteProbs) const;
```
dellaert: Why shared pointer? Just const DiscreteConditional&

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -255,21 +255,19 @@ static TableFactor::shared_ptr DiscreteFactorFromErrors(
...
 
-/**
- * @brief Multiply all the `factors` using the machinery of the TableFactor.
- *
- * @param factors The factors to multiply as a DiscreteFactorGraph.
- * @return TableFactor
- */
-static TableFactor TableProduct(const DiscreteFactorGraph &factors) {
+/* ************************************************************************ */
+TableFactor TableProduct(const DiscreteFactorGraph &factors) {
```
dellaert: Make a DiscreteFactorGraph method?

gtsam/hybrid/HybridBayesTree.cpp
```diff
@@ -179,16 +201,18 @@ VectorValues HybridBayesTree::optimize(const DiscreteValues& assignment) const {
 
 /* ************************************************************************* */
 void HybridBayesTree::prune(const size_t maxNrLeaves) {
-  auto discreteProbs = this->roots_.at(0)->conditional()->asDiscrete();
+  auto discreteProbs =
```
dellaert: I really dislike "SetData", it's very unsafe. Can we not replace these 6 lines with 
```
this->roots_.at(0)->conditional()->prune()
```
where prune is imperative and works for both DiscreteConditional and TableDistribution?

gtsam/discrete/DiscreteConditional.cpp
```diff
@@ -478,6 +478,25 @@ double DiscreteConditional::evaluate(const HybridValues& x) const {
...
+
+/* ************************************************************************* */
+DiscreteConditional::shared_ptr DiscreteConditional::max(
+    const Ordering& keys) const {
+  auto m = *BaseFactor::max(keys);
+  return std::make_shared<DiscreteConditional>(m.discreteKeys().size(), m);
+}
+
+/* ************************************************************************* */
+DiscreteConditional::shared_ptr DiscreteConditional::prune(
```
dellaert: remove SetData and make this function imperative (and overwrite it in tableDistribution)

gtsam/discrete/TableDistribution.h
```diff
@@ -0,0 +1,201 @@
...
+  using Values = DiscreteValues;  ///< backwards compatibility
+
+  /// @name Standard Constructors
+  /// @{
+
+  /// Default constructor needed for serialization.
+  TableDistribution() {}
+
+  /// Construct from factor, taking the first `nFrontals` keys as frontals.
+  TableDistribution(size_t nFrontals, const TableFactor& f);
```
dellaert: Not right: we always take all keys as frontals

gtsam/discrete/TableDistribution.h
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+   * Construct from DiscreteKeys and SparseVector, taking the first
+   * `nFrontals` keys as frontals, in the order given.
+   */
+  TableDistribution(size_t nFrontals, const DiscreteKeys& keys,
```
dellaert: Same.

gtsam/discrete/TableDistribution.h
```diff
@@ -0,0 +1,201 @@
...
+  explicit TableDistribution(const Signature& signature);
+
+  /**
+   * Construct from key, parents, and a Signature::Table specifying the
+   * conditional probability table (CPT) in 00 01 10 11 order. For
+   * three-valued, it would be 00 01 02 10 11 12 20 21 22, etc....
+   *
+   * Example: TableDistribution P(D, {B,E}, table);
+   */
+  TableDistribution(const DiscreteKey& key, const DiscreteKeys& parents,
```
dellaert: Never parents as is Distribution

gtsam/discrete/TableDistribution.h
```diff
@@ -0,0 +1,201 @@
...
+      : TableDistribution(Signature(key, parents, table)) {}
+
+  /**
+   * Construct from key, parents, and a vector<double> specifying the
+   * conditional probability table (CPT) in 00 01 10 11 order. For
+   * three-valued, it would be 00 01 02 10 11 12 20 21 22, etc....
+   *
+   * Example: TableDistribution P(D, {B,E}, table);
+   */
+  TableDistribution(const DiscreteKey& key, const DiscreteKeys& parents,
```
dellaert: same

gtsam/discrete/TableDistribution.h
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Construct from key, parents, and a string specifying the conditional
+   * probability table (CPT) in 00 01 10 11 order. For three-valued, it would
+   * be 00 01 02 10 11 12 20 21 22, etc....
+   *
+   * The string is parsed into a Signature::Table.
+   *
+   * Example: TableDistribution P(D, {B,E}, "9/1 2/8 3/7 1/9");
+   */
+  TableDistribution(const DiscreteKey& key, const DiscreteKeys& parents,
```
dellaert: same

gtsam/discrete/TableDistribution.h
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+   * Construct from DiscreteKeys and SparseVector, taking the first
+   * `nFrontals` keys as frontals, in the order given.
+   */
+  TableDistribution(size_t nFrontals, const DiscreteKeys& keys,
+                    const Eigen::SparseVector<double>& potentials);
+
+  /** Construct from signature */
+  explicit TableDistribution(const Signature& signature);
```
dellaert: Signatures allow for parents, so not relevant?

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: This will anyway go away once we make the `prune` method imperative.

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: No, I added it to make it easier to cast to `TableDistribution`. It has a default template parameter of `DiscreteConditional`.

gtsam/hybrid/HybridBayesTree.cpp
varunagrawal: Nice catch.

gtsam/hybrid/HybridBayesTree.cpp
varunagrawal: I think that should be inside `TableFactor` as an `argmax` method.

gtsam/hybrid/HybridGaussianConditional.h
varunagrawal: I am very used to dynamic dispatch being a pointer thing, but the reference also works!

varunagrawal: Question about this: once we make the `prune` method imperative, we will have a `DiscreteConditional::shared_ptr`. Doesn't it make sense to just pass in the pointer to the `HybridGaussianConditional::shared_ptr`? A reference doesn't hurt since there is no copy, but I figured just give what we get.

gtsam/hybrid/HybridBayesTree.cpp
varunagrawal: Done!

gtsam/discrete/DiscreteConditional.h
varunagrawal: This does $\max_A P(A|B)$

varunagrawal: I copied the docstring from `TableFactor::max`, so it should be updated everywhere if we decide to do so.

dellaert: And what does that yield? Is not P(A), nor P(B), so why would it be a DiscreteConditional? Where do you use this?

dellaert: max on a factor makes sense, max on a conditional does not.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: I thought we weren't going to touch `DiscreteFactorGraph`?

gtsam/discrete/DiscreteConditional.h
varunagrawal: We use it when computing the pruned discrete probabilities. From this discussion, I should probably return a `DiscreteFactor` and not a `DiscreteConditional`.

gtsam/hybrid/HybridBayesTree.cpp
```diff
@@ -41,6 +42,16 @@ bool HybridBayesTree::equals(const This& other, double tol) const {
   return Base::equals(other, tol);
 }
 
+/* ************************************************************************* */
+DiscreteValues HybridBayesTree::discreteMaxProduct(
+    const DiscreteFactorGraph& dfg) const {
+  TableFactor product = TableProduct(dfg);
+
+  uint64_t maxIdx = TableDistribution(product).argmax();
```
dellaert: This leaks privileged information outside of TableDistribution. I think the uint64 argmax should be internal, and like argmax in other factors/conditionals, an argmax should return a DiscreteValues.

gtsam/discrete/TableDistribution.h
```diff
@@ -0,0 +1,184 @@
...
+ public:
+  // typedefs needed to play nice with gtsam
+  typedef TableDistribution This;            ///< Typedef to this class
+  typedef std::shared_ptr<This> shared_ptr;  ///< shared_ptr to this class
+  typedef DiscreteConditional
+      BaseConditional;  ///< Typedef to our conditional base class
+
+  using Values = DiscreteValues;  ///< backwards compatibility
+
+  /// @name Standard Constructors
```
dellaert, line 54: Why so many constructors? Are we using them all or is this a copy/paste artifact?

gtsam/discrete/TableDistribution.h
```diff
@@ -0,0 +1,184 @@
...
+   * @return DiscreteFactor::shared_ptr
+   */
+  virtual DiscreteFactor::shared_ptr max(const Ordering& keys) const override;
+
+  /**
+   * @brief Return assignment that maximizes value.
+   *
+   * @return maximizing assignment for the variables.
+   */
+  uint64_t argmax() const;
```
dellaert: Should return DiscreteValues

gtsam/discrete/DiscreteConditional.cpp
```diff
@@ -478,6 +478,19 @@ double DiscreteConditional::evaluate(const HybridValues& x) const {
   return this->evaluate(x.discrete());
 }
 
+/* ************************************************************************* */
+DiscreteFactor::shared_ptr DiscreteConditional::max(
+    const Ordering& keys) const {
+  return BaseFactor::max(keys);
```
dellaert, line 492: If this just returns BaseFactor::max, why does this method exist?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Sure.

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -52,15 +52,22 @@ HybridBayesNet HybridBayesNet::prune(size_t maxNrLeaves) const {
...
-  // Prune the joint. NOTE: again, possibly quite expensive.
-  const DecisionTreeFactor pruned = joint.prune(maxNrLeaves);
+  // Prune the joint. NOTE: imperative and, again, possibly quite expensive.
+  DiscreteConditional pruned(joint);
+  pruned.prune(maxNrLeaves);
 
   // Create a the result starting with the pruned joint.
   HybridBayesNet result;
-  result.emplace_shared<DiscreteConditional>(pruned.size(), pruned);
+  result.push_back(std::make_shared<DiscreteConditional>(pruned));
```
dellaert: Probably better to first emplace then prune to avoid a copy
```
result.emplace_shared<DiscreteConditional>(joint);
result->back()->prune(maxNrLeaves);
```

gtsam/hybrid/tests/testGaussianMixture.cpp
```diff
@@ -79,8 +80,8 @@ TEST(GaussianMixture, GaussianMixtureModel) {
   double midway = mu1 - mu0;
   auto eliminationResult =
       gmm.toFactorGraph({{Z(0), Vector1(midway)}}).eliminateSequential();
-  auto pMid = *eliminationResult->at(0)->asDiscrete();
-  EXPECT(assert_equal(DiscreteConditional(m, "60/40"), pMid));
+  auto pMid = eliminationResult->at(0)->asDiscrete<TableDistribution>();
+  EXPECT(assert_equal(TableDistribution(m, "60 40"), *pMid));
```
dellaert, line 84: I'm still worried that we now changed the entire hybrid inference pipeline instead of  just the use case we wanted to optimize. In particular, we now silently impose the 32 variable in the last clique constraint on everyone. This is why I was wondering about a more "bespoke" flag as to when to use the TableDistribution. Let;s brainstorm?

gtsam/discrete/DiscreteConditional.cpp
varunagrawal: So we can override it within `TableDistribution`. The original signature for `max` is inherited from DecisionTreeFactor as
```cpp
DecisionTreeFactor::shared_ptr max(const Ordering& keys) const;
```
which returns a DecisionTreeFactor. It won't work if we need to return a TableFactor, and changing the return type in DecisionTreeFactor is not backwards compatible (needs an explicit cast).

gtsam/hybrid/HybridBayesTree.cpp
varunagrawal: I agree with the comment but FYI except for `DiscreteLookupDAG` no argmax returns a DiscreteValues. That was why I originally did it this way.

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: So I will have to do `DiscreteConditional pruned = *result.back()->asDiscrete()` because we need `pruned` for pruning the `HybridGaussianConditional`s.

I guess that didn't help with the extra copy :(

dellaert: No, I think results stores shared pointers, so that will work fine.

gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -124,24 +124,17 @@ namespace gtsam {
    * @param factors The factors to multiply as a DiscreteFactorGraph.
    * @return DiscreteFactor::shared_ptr
    */
-  static DiscreteFactor::shared_ptr DiscreteProduct(
-      const DiscreteFactorGraph& factors) {
+  DiscreteFactor::shared_ptr DiscreteFactorGraph::scaledProduct() const {
```
dellaert, line 122: Doxygen needs to be in .h file. Maybe move it there and merge with your more terse comment.

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -450,11 +450,18 @@ TEST(HybridBayesNet, UpdateDiscreteConditionals) {
 
   DiscreteConditional joint;
   for (auto&& conditional : posterior->discreteMarginal()) {
-    joint = joint * (*conditional);
+    // The last discrete conditional may be a TableDistribution
```
dellaert: I think this code should still work as is.

gtsam/hybrid/HybridGaussianFactorGraph.h
```diff
@@ -20,6 +20,7 @@
 
 #include <gtsam/discrete/DiscreteFactorGraph.h>
 #include <gtsam/discrete/DiscreteKey.h>
+#include <gtsam/discrete/TableDistribution.h>
```
dellaert: Why? Move to cpp to remove header bloat

gtsam/hybrid/HybridBayesTree.cpp
```diff
@@ -41,6 +42,22 @@ bool HybridBayesTree::equals(const This& other, double tol) const {
   return Base::equals(other, tol);
 }
 
+/* ************************************************************************* */
+DiscreteValues HybridBayesTree::discreteMaxProduct(
```
dellaert, line 46: This should be in DiscreteFactorGraph?

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -52,15 +52,24 @@ HybridBayesNet HybridBayesNet::prune(size_t maxNrLeaves) const {
   // Multiply into one big conditional. NOTE: possibly quite expensive.
   DiscreteConditional joint;
   for (auto &&conditional : marginal) {
-    joint = joint * (*conditional);
+    // The last discrete conditional may be a TableDistribution
```
dellaert: Why does code no longer work as is?

dellaert (commented):
Still have some concerns, and several smaller comments. MAybe we should chat about the joint calculation in particular. I did not expect that to break.

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: Must have been from before we refactored discrete multiplication.

dellaert: Think about how it can still work this way, please? Seems it should be possible. The new code is too complex

varunagrawal: So this is not working as is because we enabled multiplication for `DiscreteFactor::shared_ptr` so the result is a `DiscreteFactor::shared_ptr` which has no concept of frontal variables.

We can either convert this to `DiscreteFactor`s and then convert to a `DiscreteConditional` at the end, or use a `TableDistribution` as the type for `joint`.

dellaert: Just seeing this comment. This will work with the fix we discussed, right? Noticed the comments is older than that...

varunagrawal:
NOTE: I had to override `sample` in `TableDistribution` since setting the ADT to null revealed a bug in sampling.

gtsam/discrete/AlgebraicDecisionTree.h
```diff
@@ -57,6 +57,9 @@ namespace gtsam {
 
     AlgebraicDecisionTree(double leaf = 1.0) : Base(leaf) {}
 
+    /// Constructor which accepts root pointer
+    AlgebraicDecisionTree(const typename Base::NodePtr root) : Base(root) {}
```
varunagrawal, line 61: I added this new constructor so I could set the ADT root node to null explicitly.

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: I am starting to wonder if this has been a distraction. The previous `HybridBayesNet::TableProduct` method worked great for us.

gtsam/hybrid/HybridBayesTree.cpp
varunagrawal: This is special behavior leveraging the `TableFactor`. I don't think we should mess with `DiscreteFactorGraph` wrt this.

dellaert (approved):
LGTM!

dellaert:
I’ll let you merge

PR #1955, opened 2024-12-31
Customize discrete elimination in Hybrid

varunagrawal:
PR 2/3: Customize the elimination of the discrete factors into a conditional on all the discrete variables since they form a fully-connected clique.


gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -255,6 +255,48 @@ static TableFactor::shared_ptr DiscreteFactorFromErrors(
...
 }
 
+/**
+ * @brief Multiply all the `factors` and normalize the
+ * product to prevent underflow.
+ *
+ * @param factors The factors to multiply as a DiscreteFactorGraph.
+ * @return TableFactor
+ */
+static TableFactor ProductAndNormalize(const DiscreteFactorGraph &factors) {
```
dellaert: Why normalize? Not needed in this PR, and can be done in derived class constructor in next PR?

varunagrawal: Wouldn't it be better to pre-normalize? That way we won't have to duplicate normalization in various constructors (if the need arises).

varunagrawal: I am normalizing since this is what we do in `EliminateDiscrete` as well.

gtsam/discrete/DiscreteConditional.cpp
```diff
@@ -47,6 +47,15 @@ DiscreteConditional::DiscreteConditional(const size_t nrFrontals,
                                          const DecisionTreeFactor& f)
     : BaseFactor(f / (*f.sum(nrFrontals))), BaseConditional(nrFrontals) {}
 
+/* ************************************************************************** */
```
dellaert: Fix TableFactor in separate PR and remove this constructor

gtsam/discrete/TableFactor.cpp
```diff
@@ -252,6 +252,15 @@ DecisionTreeFactor TableFactor::operator*(const DecisionTreeFactor& f) const {
 DecisionTreeFactor TableFactor::toDecisionTreeFactor() const {
   DiscreteKeys dkeys = discreteKeys();
 
+  // If no keys, then return empty DecisionTreeFactor
```
dellaert: Fix in separate PR, with unit test you already have

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Please rename after our discussion as this might give the false impression (later relied upon) that a factor is normalized.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -255,6 +255,48 @@ static TableFactor::shared_ptr DiscreteFactorFromErrors(
   return std::make_shared<TableFactor>(discreteKeys, potentials);
 }
 
+/**
+ * @brief Multiply all the `factors` and normalize the
```
dellaert: That is an implementation detail. It is not "normalized".

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -255,6 +255,48 @@ static TableFactor::shared_ptr DiscreteFactorFromErrors(
...
+        product = TableFactor(product * (*dtf));
+      }
+    }
+  }
+#if GTSAM_HYBRID_TIMING
+  gttoc_(DiscreteProduct);
+#endif
+
+  // Max over all the potentials by pretending all keys are frontal:
+  auto normalizer = product.max(product.size());
```
dellaert: call this denominator

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -299,13 +339,40 @@ discreteElimination(const HybridGaussianFactorGraph &factors,
...
+#endif
+    auto conditional = std::make_shared<DiscreteConditional>(
+        frontalKeys.size(), product.toDecisionTreeFactor());
 #if GTSAM_HYBRID_TIMING
-  gttoc_(EliminateDiscrete);
+    gttoc_(EliminateDiscreteFormDiscreteConditional);
 #endif
 
-  return {std::make_shared<HybridConditional>(result.first), result.second};
+    TableFactor::shared_ptr sum = product.sum(frontalKeys);
```
varunagrawal, line 369: I still need to compute `sum` as the $\tau$ factor to return.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -299,13 +339,40 @@ discreteElimination(const HybridGaussianFactorGraph &factors,
 #if GTSAM_HYBRID_TIMING
   gttic_(EliminateDiscrete);
 #endif
-  // NOTE: This does sum-product. For max-product, use EliminateForMPE.
-  auto result = EliminateDiscrete(dfg, frontalKeys);
+  // Check if separator is empty
```
dellaert: Maybe isolate for-loop above in a separate function

dellaert: Seems very complicated. Does `frontalKeys().size() == dfg.keyVector().size()` not work?

gtsam/hybrid/HybridGaussianFactorGraph.cpp
dellaert: Yikes, then we compute the sum twice. Maybe have an output argument in TableDistribution constructor?

gtsam/discrete/DiscreteConditional.cpp
varunagrawal: Added to #1960

gtsam/discrete/TableFactor.cpp
varunagrawal: Added in #1960

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: Is it possible that the number of keys are the same but the actual keys are different?
E.g. `{x0, x1, x2}` and `{x0, x1, y0}`?

I've updated to use the size check as suggested, but checking for explicit equality seemed wiser.

dellaert: No, not possible. Think about it for a bit :-)

varunagrawal: Yup I thought so, was curious.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
varunagrawal: That will have to be in the follow up PR that introduces the TableDistribution. I can add a factory constructor that returns the sum.

gtsam/hybrid/HybridGaussianFactorGraph.cpp
```diff
@@ -255,15 +255,55 @@ static TableFactor::shared_ptr DiscreteFactorFromErrors(
   return std::make_shared<TableFactor>(discreteKeys, potentials);
 }
 
+/**
+ * @brief Multiply all the `factors` using the machinery of the TableFactor.
+ *
+ * @param factors The factors to multiply as a DiscreteFactorGraph.
+ * @return TableFactor
+ */
+static TableFactor TableProduct(const DiscreteFactorGraph &factors) {
```
dellaert, line 264: Can be in another PR, but maybe this should be a method of DiscreteFactorGraph

dellaert (approved):
Nice !

PR #1959, opened 2025-01-02
Enable more warnings and remove unused variables

mcm001:
Captures follow-on work from https://github.com/borglab/gtsam/pull/1950, including:
- Removes lingering stray unused variables
- Enables `-Wall -Werror -Wpedantic -Wextra $<$<COMPILE_LANGUAGE:CXX>:-Wextra -Wno-unused-parameter>` on non-MSVC compilers, and `/W3 /WX` on MSVC
- Bumps Linux CI to ubuntu-22.04-gcc-12 from ubuntu-22.04-gcc-11

Open work:
- [x] ~Build without warnings on Linux + GCC 11 locally~
- [x] ~Test on Windows~
- [x] ~Test with Clang and AppleClang~
- [x] ~**[Upgrade CI from GCC 11 to 12](https://github.com/borglab/gtsam/pull/1950#issuecomment-2568122774)**~
- [ ] Potentially bump gcc9 to 10

### DIFF
--- a/gtsam/base/TestableAssertions.h
+++ b/gtsam/base/TestableAssertions.h
@@ -32,7 +32,8 @@ namespace gtsam {
 /**
  * Equals testing for basic types
  */
-inline bool assert_equal(const Key& expected, const Key& actual, double tol = 0.0) {
+inline bool assert_equal(const Key& expected, const Key& actual) {
+  // TODO - why isn't tol used?
   if(expected != actual) {
     std::cout << "Not equal:\nexpected: " << expected << "\nactual: " << actual << std::endl;
     return false;
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -27,7 +27,7 @@ jobs:
         name: [
           ubuntu-20.04-gcc-9,
           ubuntu-20.04-clang-9,
-          ubuntu-22.04-gcc-11,
+          ubuntu-22.04-gcc-12,
           ubuntu-22.04-clang-14,
         ]
 
@@ -44,7 +44,7 @@ jobs:
             compiler: clang
             version: "9"
 
-          - name: ubuntu-22.04-gcc-11
+          - name: ubuntu-22.04-gcc-12
             os: ubuntu-22.04
             compiler: gcc
             version: "11"
--- a/cmake/GtsamBuildTypes.cmake
+++ b/cmake/GtsamBuildTypes.cmake
@@ -111,7 +111,7 @@ if(MSVC)
   set(CMAKE_3_15 $<VERSION_LESS:${CMAKE_VERSION},3.15>)
   set(CMAKE_3_25 $<VERSION_LESS:${CMAKE_VERSION},3.25>)
   # Common to all configurations, next for each configuration:
-  set(GTSAM_COMPILE_OPTIONS_PRIVATE_COMMON          /W3 /GR /EHsc /MP  CACHE STRING "(User editable) Private compiler flags for all configurations.")
+  set(GTSAM_COMPILE_OPTIONS_PRIVATE_COMMON          /W3 /WX /GR /EHsc /MP  CACHE STRING "(User editable) Private compiler flags for all configurations.")
   set(GTSAM_COMPILE_OPTIONS_PRIVATE_DEBUG           $<${CMAKE_3_15}:/MDd> $<${CMAKE_3_25}:/Zi> /Ob0 /Od /RTC1  CACHE STRING "(User editable) Private compiler flags for Debug configuration.")
   set(GTSAM_COMPILE_OPTIONS_PRIVATE_RELWITHDEBINFO  $<${CMAKE_3_15}:/MD> /O2  $<${CMAKE_3_25}:/Zi>  CACHE STRING "(User editable) Private compiler flags for RelWithDebInfo configuration.")
   set(GTSAM_COMPILE_OPTIONS_PRIVATE_RELEASE         $<${CMAKE_3_15}:/MD> /O2  CACHE STRING "(User editable) Private compiler flags for Release configuration.")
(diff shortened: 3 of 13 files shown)


mcm001:
A lot of the [new windows warnings](https://github.com/borglab/gtsam/actions/runs/12586541885/job/35080588408?pr=1959) don't seem super helpful. Will probably disable C4365. But I did find that `tol` is [unused in TestableAssertions](https://github.com/borglab/gtsam/blob/c2a124226700e050b857b921ba96893bf4fdd5ac/gtsam/base/TestableAssertions.h#L35), which seems like a bug?

gtsam/base/TestableAssertions.h
```diff
@@ -33,6 +33,7 @@ namespace gtsam {
  * Equals testing for basic types
  */
 inline bool assert_equal(const Key& expected, const Key& actual, double tol = 0.0) {
+  // TODO - why isn't tol used?
```
mcm001, line 36: Is tol intended to be unused?

dellaert: Key is a big int so assert_equal with tolerance does not make sense. I forget if there is a reason we added it but at this point I'm thinking it should not exist : in unit test we should use LONG_EQUALS.

mcm001: Just deleting tol doesn't break anything - removing

dellaert:
Getting there! When this needs review request one from me and take out of draft status.

mcm001:
I think with Windows tests passing the only open work is if we want to bump CI from gcc 9 to 10? IIRC you mentioned some embedded users are still stuck on 9?

gtsam/CMakeLists.txt
```diff
@@ -175,6 +175,8 @@ target_include_directories(gtsam SYSTEM BEFORE PUBLIC
   # CCOLAMD
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/CCOLAMD/Include>
   $<INSTALL_INTERFACE:include/gtsam/3rdparty/CCOLAMD>
+  # Eigen
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/Eigen>
```
dellaert: Explain?

gtsam/CMakeLists.txt
```diff
@@ -194,6 +196,14 @@ if(WIN32) # library to help with demangling variable names on Windows
   target_link_libraries(gtsam PRIVATE Dbghelp)
 endif()
 
+# Add warnings
+if(MSVC)
+  target_compile_options(gtsam PRIVATE /W3 /WX)
+else()
+  # No-unused-paramter isn't super helpful, as some of our functions are totally ifdefed out in some build configurations
+  target_compile_options(gtsam PRIVATE -Wall -Werror -Wpedantic -Wextra $<$<COMPILE_LANGUAGE:CXX>:-Wextra -Wno-unused-parameter>)
```
dellaert: Wow, I’m surprised this passes given that indeed some of our third libraries incur warnings.

dellaert (commented):
Left a few comments. I think we can punt on the GCC version issue. What I’m wondering though is why we don’t get warnings in third parties like Eigen and metis . Every time I compiled those I do see warnings. We are not supposed to change their source, however, under their licenses.

gtsam_unstable/CMakeLists.txt
```diff
@@ -113,6 +113,14 @@ if(WIN32) # Add 'lib' prefix to static library to avoid filename collision with
 	endif()
 endif()
 
+# Add warnings
+if(MSVC)
+  target_compile_options(gtsam PRIVATE /W3 /WX)
```
Gold856: Should be `gtsam_unstable`.

Gold856 (changes requested):

Gold856:
I noticed that `-Wall` was already applied to gtsam's libraries via `gtsam_apply_build_flags` in cmake/GtsamBuildTypes.cmake, but this directly applies the flags by using `target_compile_options`. Is there a preference for how these flags are applied?

mcm001:
Oh interesting, it does look like these flags should already be present. Let me double check the generated compile_commands.json on develop to confirm - either way, these seem like they ought to live there instead.

dellaert:
Please merge in #1964 and adapt if necessary? Then maybe request review from @Gold856 again, and then we can merge :-)

gtsam/CMakeLists.txt
mcm001: The TLDR is that by marking Eigen as a system include, our compiler will ignore warnings generated by files inside this directory. Without this we do indeed see warnings generated by Eigen, which would be elevated to errors by Werror.

mcm001: From the [horses mouth](https://cmake.org/cmake/help/latest/command/include_directories.html)

> If the SYSTEM option is given, the compiler will be told the directories are meant as system include directories on some platforms. Signaling this setting might achieve effects such as the compiler skipping warnings, or these fixed-install system files not being considered in dependency calculations - see compiler docs

Gold856: #1964 did this already so it's a little bit cleaner.

Gold856 (changes requested):

gtsam/CMakeLists.txt
mcm001: Oh great; I'll remove it entirely then

gtsam_unstable/CMakeLists.txt
mcm001: Removed and moved into GtsamBuildTypes

gtsam/CMakeLists.txt
mcm001: Looks like @Gold856 added Eigen as a system include in #1964 (see https://github.com/Gold856/gtsam/blob/bf4f6f5554c3b190539da5631e6246f146ef8fc6/cmake/HandleEigen.cmake#L45) already

dellaert (approved):
Awesome!

dellaert:
Ready to merge?

mcm001:
Good to go if @Gold856 is -- thanks!

Gold856:
CMake build looks good to me. Ship it!

PR #1960, opened 2025-01-02
Table Factor Fixes

varunagrawal:
1. Fixed a bug where a `DecisionTreeFactor` with reverse ordering of keys would cause incorrect table construction.
2. Fixed the case where getting a DecisionTreeFactor from an empty or singular TableFactor would throw a segfault.
3. Use faster sets and vectors.


dellaert (approved):
Cool! I guess this does not yet fix the "divide" bug?

varunagrawal:
> Cool! I guess this does not yet fix the "divide" bug?

The divide bug is in `DecisionTreeFactor` (at least from what I've seen). I'll check in TableFactor as well.

PR #1961, opened 2025-01-03
TableFactor Serialization

varunagrawal:
Add support for serializing `TableFactor`.

The key change here is adding a custom serialize method for `Eigen::SparseVector`. I had tried using a generic `Eigen::SparseMatrix` serialization method, but for some reason it is not getting picked up as a type match.


dellaert (approved):
LGTM!

PR #1962, opened 2025-01-03
Fix DecisionTreeFactor division

varunagrawal:
Added a unit test that ~~shows the number of keys is incorrect on division, and subsequently fixed it~~ improves the understanding of discrete factor division.

~~Also added a `Signature` based constructor and updated the tests to use it.~~

### DIFF
--- a/gtsam/discrete/tests/testDecisionTreeFactor.cpp
+++ b/gtsam/discrete/tests/testDecisionTreeFactor.cpp
@@ -30,6 +30,12 @@
 using namespace std;
 using namespace gtsam;
 
+/** Convert Signature into CPT */
+DecisionTreeFactor create(const Signature& signature) {
+  DecisionTreeFactor p(signature.discreteKeys(), signature.cpt());
+  return p;
+}
+
 /* ************************************************************************* */
 TEST(DecisionTreeFactor, ConstructorsMatch) {
   // Declare two keys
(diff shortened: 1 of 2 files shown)


dellaert:
Hmmm, a signature-based constructor only makes sense to me for conditionals (or distributions). I think this mixes up the semantics

gtsam/discrete/tests/testDecisionTreeFactor.cpp
```diff
@@ -217,27 +226,21 @@ void maybeSaveDotFile(const DecisionTreeFactor& f, const string& filename) {
...
   gttic_(asiaCPTs);
-  DecisionTreeFactor pA = create(A % "99/1");
-  DecisionTreeFactor pS = create(S % "50/50");
-  DecisionTreeFactor pT = create(T | A = "99/1 95/5");
-  DecisionTreeFactor pL = create(L | S = "99/1 90/10");
-  DecisionTreeFactor pB = create(B | S = "70/30 40/60");
-  DecisionTreeFactor pE = create((E | T, L) = "F T T T");
-  DecisionTreeFactor pX = create(X | E = "95/5 2/98");
-  DecisionTreeFactor pD = create((D | E, B) = "9/1 2/8 3/7 1/9");
+  DecisionTreeFactor pA(A % "99/1");
```
dellaert: All good if these are DiscreteConditionals, which already have a signature constructor

varunagrawal: Okay I'll revert these changes.

dellaert:
I see why you did this, when looking at the tests, but there is a reason I left it in the tests: it is because signatures are a conditional thing. I think a better way is to create conditionals in the tests, which are derived from DTF anyway, so you can use them as factors in the tests

varunagrawal:
> I see why you did this, when looking at the tests, but there is a reason I left it in the tests: it is because signatures are a conditional thing. I think a better way is to create conditionals in the tests, which are derived from DTF anyway, so you can use them as factors in the tests

It doesn't really matter to be honest. I thought it was better to have Signature constructor since the `create` method was doing that indirectly, but I didn't have the full picture.

dellaert:
CI fails now, maybe `create` was not re-instated?
Actually, I think my suggested change, creating conditionals to create the factors made sense, no?

varunagrawal:
From the diff, it looks like `create` was added back. I'll look into this some more.

varunagrawal:
On more thinking about this, I believe the mathematical structure should include all keys since we would have functional division $(\frac{f}{g})(x, y) = f(x, y) / g(x)$.

I'll update the unit test to showcase this and update the docstring.

dellaert (approved):
Cool!

PR #1963, opened 2025-01-05
DiscreteFactor multiply method

varunagrawal:
Add a multiply method for `DiscreteFactor` which can be used to handle derived classes easily.

One caveat I encountered is that I have to also make `sum` and `max` as abstract base class methods, and finally add new constructors for `DiscreteLookupTable` and `DiscreteConditional` which take `DiscreteFactor` as an input. I can do this in another PR for cleanliness sake.

### DIFF
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -18,9 +18,10 @@
  */
 
 #include <gtsam/base/FastSet.h>
-#include <gtsam/hybrid/HybridValues.h>
 #include <gtsam/discrete/DecisionTreeFactor.h>
 #include <gtsam/discrete/DiscreteConditional.h>
+#include <gtsam/discrete/TableFactor.h>
+#include <gtsam/hybrid/HybridValues.h>
 
 #include <utility>
 
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -70,6 +70,9 @@ namespace gtsam {
       result = std::make_shared<TableFactor>((*tf) * TableFactor(*this));
     } else if (auto dtf = std::dynamic_pointer_cast<DecisionTreeFactor>(f)) {
       result = std::make_shared<DecisionTreeFactor>(this->operator*(*dtf));
+    } else {
+      // Simulate double dispatch in C++
+      result = std::make_shared<DecisionTreeFactor>(f->operator*(*this));
     }
     return result;
   }
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -68,11 +68,20 @@ namespace gtsam {
       const DiscreteFactor::shared_ptr& f) const {
     DiscreteFactor::shared_ptr result;
     if (auto tf = std::dynamic_pointer_cast<TableFactor>(f)) {
+      // If f is a TableFactor, we convert `this` to a TableFactor since this
+      // conversion is cheaper than converting `f` to a DecisionTreeFactor. We
+      // then return a TableFactor.
       result = std::make_shared<TableFactor>((*tf) * TableFactor(*this));
+
     } else if (auto dtf = std::dynamic_pointer_cast<DecisionTreeFactor>(f)) {
+      // If `f` is a DecisionTreeFactor, simply call operator*.
       result = std::make_shared<DecisionTreeFactor>(this->operator*(*dtf));
+
     } else {
       // Simulate double dispatch in C++
+      // Useful for other classes which inherit from DiscreteFactor and have
+      // only `operator*(DecisionTreeFactor)` defined. Thus, other classes don't
+      // need to be updated.
       result = std::make_shared<DecisionTreeFactor>(f->operator*(*this));
     }
     return result;
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -63,6 +63,18 @@ namespace gtsam {
   }
 
   /* ************************************************************************ */
+  DiscreteFactor::shared_ptr DecisionTreeFactor::multiply(
+      const DiscreteFactor::shared_ptr& f) const override {
+    DiscreteFactor::shared_ptr result;
+    if (auto tf = std::dynamic_pointer_cast<TableFactor>(f)) {
+      result = std::make_shared<TableFactor>((*tf) * TableFactor(*this));
+    } else if (auto dtf = std::dynamic_pointer_cast<DecisionTreeFactor>(f)) {
+      result = std::make_shared<DecisionTreeFactor>(this->operator*(*dtf));
+    }
+    return result;
+  }
+
+  /* ************************************************************************ */
   double DecisionTreeFactor::safe_div(const double& a, const double& b) {
     // The use for safe_div is when we divide the product factor by the sum
     // factor. If the product or sum is zero, we accord zero probability to the
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -64,7 +64,7 @@ namespace gtsam {
 
   /* ************************************************************************ */
   DiscreteFactor::shared_ptr DecisionTreeFactor::multiply(
-      const DiscreteFactor::shared_ptr& f) const override {
+      const DiscreteFactor::shared_ptr& f) const {
     DiscreteFactor::shared_ptr result;
     if (auto tf = std::dynamic_pointer_cast<TableFactor>(f)) {
       result = std::make_shared<TableFactor>((*tf) * TableFactor(*this));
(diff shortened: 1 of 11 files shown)


gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -65,11 +65,11 @@ namespace gtsam {
...
   DecisionTreeFactor DiscreteFactorGraph::product() const {
-    DecisionTreeFactor result;
-    for (const sharedFactor& factor : *this) {
-      if (factor) result = (*factor) * result;
+    DiscreteFactor::shared_ptr result = *this->begin();
+    for (auto it = this->begin() + 1; it != this->end(); ++it) {
+      if (*it) result = result->multiply(*it);
     }
-    return result;
+    return result->toDecisionTreeFactor();
```
varunagrawal, line 79: Unfortunately, I had to call the potentially expensive `toDecisionTreeFactor` method here to avoid more changes.

gtsam/discrete/DiscreteFactor.h
```diff
@@ -129,6 +129,16 @@ class GTSAM_EXPORT DiscreteFactor : public Factor {
...
   virtual DecisionTreeFactor operator*(const DecisionTreeFactor&) const = 0;
 
+  /**
+   * @brief Multiply in a DiscreteFactor and return the result as
+   * DiscreteFactor, both via shared pointers.
+   *
+   * @param df DiscreteFactor shared_ptr
+   * @return DiscreteFactor::shared_ptr
+   */
+  virtual DiscreteFactor::shared_ptr multiply(
```
varunagrawal, line 139: I feel like this can be `DiscreteFactor::shared_ptr operator*(const DiscreteFactor::shared_ptr&)` instead of `multiply`.

dellaert: No, operator* does not normally do shared pointers

gtsam/discrete/DecisionTreeFactor.h
```diff
@@ -22,6 +22,7 @@
 #include <gtsam/discrete/DiscreteFactor.h>
 #include <gtsam/discrete/DiscreteKey.h>
 #include <gtsam/discrete/Ring.h>
+#include <gtsam/discrete/TableFactor.h>
```
dellaert: Why? Move to cpp if possible

gtsam_unstable/discrete/AllDiff.h
```diff
@@ -53,6 +53,13 @@ class GTSAM_UNSTABLE_EXPORT AllDiff : public Constraint {
   /// Multiply into a decisiontree
   DecisionTreeFactor operator*(const DecisionTreeFactor& f) const override;
 
+  /// Multiply factors, DiscreteFactor::shared_ptr edition
+  DiscreteFactor::shared_ptr multiply(
```
dellaert: These all seem to be the same. Can it not be done in the Constraint base class?

gtsam/discrete/TableFactor.cpp
```diff
@@ -254,6 +254,22 @@ DecisionTreeFactor TableFactor::operator*(const DecisionTreeFactor& f) const {
   return toDecisionTreeFactor() * f;
 }
 
+/* ************************************************************************ */
+DiscreteFactor::shared_ptr TableFactor::multiply(
+    const DiscreteFactor::shared_ptr& f) const {
+  DiscreteFactor::shared_ptr result;
+  if (auto tf = std::dynamic_pointer_cast<TableFactor>(f)) {
+    result = std::make_shared<TableFactor>(this->operator*(*tf));
```
dellaert, line 263: I would think operator* with call multiply and not the other way around

dellaert (commented):
Quick Look, more later

gtsam/discrete/DiscreteFactor.h
varunagrawal: Gotcha

gtsam/discrete/TableFactor.cpp
varunagrawal: But then operator* would have to accept a shared pointer and your previous comment contradicts this.

dellaert: You're right.

gtsam/discrete/DecisionTreeFactor.h
```diff
@@ -147,6 +148,10 @@ namespace gtsam {
     /// Calculate error for DiscreteValues `x`, is -log(probability).
     double error(const DiscreteValues& values) const override;
 
+    /// Multiply factors, DiscreteFactor::shared_ptr edition
```
dellaert: Please document behavior here, esp. regarding return type and expected complexity

gtsam/discrete/DecisionTreeFactor.h
dellaert: Not resolved yet I think

gtsam/discrete/DecisionTreeFactor.cpp
```diff
@@ -62,6 +62,21 @@ namespace gtsam {
...
+  /* ************************************************************************ */
+  DiscreteFactor::shared_ptr DecisionTreeFactor::multiply(
+      const DiscreteFactor::shared_ptr& f) const {
+    DiscreteFactor::shared_ptr result;
+    if (auto tf = std::dynamic_pointer_cast<TableFactor>(f)) {
+      result = std::make_shared<TableFactor>((*tf) * TableFactor(*this));
+    } else if (auto dtf = std::dynamic_pointer_cast<DecisionTreeFactor>(f)) {
+      result = std::make_shared<DecisionTreeFactor>(this->operator*(*dtf));
+    } else {
+      // Simulate double dispatch in C++
```
dellaert, line 81: I don't know what that means

gtsam/discrete/DecisionTreeFactor.cpp
```diff
@@ -62,6 +62,21 @@ namespace gtsam {
     return error(values.discrete());
   }
 
+  /* ************************************************************************ */
+  DiscreteFactor::shared_ptr DecisionTreeFactor::multiply(
+      const DiscreteFactor::shared_ptr& f) const {
+    DiscreteFactor::shared_ptr result;
```
dellaert, line 69: add comments below

gtsam/discrete/DiscreteFactorGraph.cpp
dellaert: Will be done in next PR then, right?

gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -65,11 +65,18 @@ namespace gtsam {
 
   /* ************************************************************************ */
   DecisionTreeFactor DiscreteFactorGraph::product() const {
-    DecisionTreeFactor result;
-    for (const sharedFactor& factor : *this) {
-      if (factor) result = (*factor) * result;
+    DiscreteFactor::shared_ptr result;
+    for (auto it = this->begin(); it != this->end(); ++it) {
```
dellaert, line 69: Why is it not just
```
for (const sharedFactor& factor : *this) 
      if (factor) result = factor-> multiply(result);
```

gtsam/discrete/TableFactor.h
```diff
@@ -178,6 +178,10 @@ class GTSAM_EXPORT TableFactor : public DiscreteFactor {
   /// multiply with DecisionTreeFactor
   DecisionTreeFactor operator*(const DecisionTreeFactor& f) const override;
 
+  /// Multiply factors, DiscreteFactor::shared_ptr edition
```
dellaert: Same. Provide meaningful comment about return type and complexity implicafions of passing one version vs other

gtsam/discrete/TableFactor.cpp
```diff
@@ -254,6 +254,22 @@ DecisionTreeFactor TableFactor::operator*(const DecisionTreeFactor& f) const {
   return toDecisionTreeFactor() * f;
 }
 
+/* ************************************************************************ */
+DiscreteFactor::shared_ptr TableFactor::multiply(
+    const DiscreteFactor::shared_ptr& f) const {
+  DiscreteFactor::shared_ptr result;
```
dellaert, line 260: Add some comments

dellaert (approved):
Approve **modulo comments**

gtsam/discrete/DecisionTreeFactor.cpp
varunagrawal: Double dispatch is a programming technique, a special case of multiple dispatch, to handle proper method calls over multiple types with low redundancy. Using it, we can avoid defining anything in other classes on the chance we add a new class in the future. That new class has to only define `operator*(DecisionTreeFactor)` and whatever other specialization.

Source: https://en.wikipedia.org/wiki/Double_dispatch

gtsam/discrete/DiscreteFactorGraph.cpp
varunagrawal: Because `result` is initially null. So `factor*result` will give null every time.

gtsam/discrete/DiscreteFactorGraph.cpp
varunagrawal: Yup. That's #1919

gtsam/discrete/DiscreteFactorGraph.cpp
varunagrawal: Reminder that `result` is now a shared_ptr not an object, so it initially points to nothing and has no data associated with it.

PR #1964, opened 2025-01-07
Clean up CMake build

Gold856:
This increases the use of more standard CMake variables for controlling things like PIC, output directories, and MSVC flags, and also removes redundancies like empty prefixes on Windows. This also sets all CMake policies up to 3.29 to NEW, which allows certain MSVC flags to be controlled by variables instead, and removes /W3 from the default set of flags, which reduces the number of warnings of compile flags being overriden to about 2.


dellaert:
Oooh, nice! I am not a cmake guru, but @jlblancoc is :-) and he might also have visibility on effect on ROS (if any). I requested a review from him to look over this. The CI will also tell us a lot.

dellaert:
PS, @Gold856, not a deal-breaker, but I *am* more comfortable with contributors that have their name and affiliation listed. We can then also properly acknowledge you in a next release.

cmake/HandleEigen.cmake
```diff
@@ -46,7 +42,7 @@ else()
 
     add_library(gtsam_eigen3 INTERFACE)
 
-    target_include_directories(gtsam_eigen3 INTERFACE
+    target_include_directories(gtsam_eigen3 SYSTEM INTERFACE
```
jlblancoc, line 45: @dellaert @varunagrawal FYI: This "SYSTEM" prefix here allows one to raise the level of warnings in "our" code, while ignoring errors in other's "system" headers. I think it's a good thing to have.

jlblancoc (approved):
All changes look awesome to me! 

@Gold856 is the new CMake guru in town ;-)

jlblancoc:
@dellaert ROS builds shouldn't be affected (but I'll keep an eye on the build farm). 
All CI passed here, so you can merge it.

PR #1971, opened 2025-01-11
Add cassert include to Chebyshev2.cpp

mcm001:
With gcc 11.4.0-1ubuntu1~22.04, I'm seeing this include missing from Chebyshev2.cpp. I haven't yet dug into why this isn't something we see in our CI on borglab/gtsam


ProfFan (approved):

mcm001:
Thanks!

PR #1974, opened 2025-01-12
Fix examples using discrete factor graph product

varunagrawal:
I didn't realize these were broken until I compiled examples for my own hybrid example.
Simple fixes.


dellaert (approved):

dellaert:
Could you make at least one of the special cases build the examples? I think a previous PR disabled examples in the interest of faster CI, but we should have at least one...

varunagrawal:
> Could you make at least one of the special cases build the examples? I think a previous PR disabled examples in the interest of faster CI, but we should have at least one...

Updated. There is only one case that builds examples now.

PR #1975, opened 2025-01-13
Adding lever-arm to GPS factors

mnissov:
I noticed the GPS factors had no option for a non-zero lever arm, the feature became necessary for me so I figured I would make this PR.

Of course similar behavior is possible with `setBodyPSensor` but this comes at the cost of additional linearization error I believe ([ref](https://groups.google.com/g/gtsam-users/c/nG_B_-5VDxA/m/Ob3r82J5AQAJ)). As a result the performance is actually different.

### DIFF
--- a/gtsam/navigation/GPSFactor.cpp
+++ b/gtsam/navigation/GPSFactor.cpp
@@ -64,6 +64,37 @@ pair<Pose3, Vector3> GPSFactor::EstimateState(double t1, const Point3& NED1,
 
   return make_pair(nTb, nV);
 }
+
+//***************************************************************************
+void GPSFactorArm::print(const string& s, const KeyFormatter& keyFormatter) const {
+  cout << s << "GPSFactorArm on " << keyFormatter(key()) << "\n";
+  cout << "  GPS measurement: " << nT_.transpose() << "\n";
+  cout << "  Lever arm: " << bL_.transpose() << "\n";
+  noiseModel_->print("  noise model: ");
+}
+
+//***************************************************************************
+bool GPSFactorArm::equals(const NonlinearFactor& expected, double tol) const {
+  const This* e = dynamic_cast<const This*>(&expected);
+  return e != nullptr && Base::equals(*e, tol) &&
+         traits<Point3>::Equals(nT_, e->nT_, tol) &&
+         traits<Point3>::Equals(bL_, e->bL_, tol);
+}
+
+//***************************************************************************
+Vector GPSFactorArm::evaluateError(const Pose3& p,
+    OptionalMatrixType H) const {
+  const Matrix3 nRb = p.rotation().matrix();
+  if (H) {
+    H->resize(3, 6);
+
+    H->block<3, 3>(0, 0) = -nRb * skewSymmetric(bL_);
+    H->block<3, 3>(0, 3) = nRb;
+  }
+
+  return p.translation() + nRb * bL_ - nT_;
+}
+
 //***************************************************************************
 void GPSFactor2::print(const string& s, const KeyFormatter& keyFormatter) const {
   cout << s << "GPSFactor2 on " << keyFormatter(key()) << "\n";
@@ -85,5 +116,34 @@ Vector GPSFactor2::evaluateError(const NavState& p,
 }
 
 //***************************************************************************
+void GPSFactor2Arm::print(const string& s, const KeyFormatter& keyFormatter) const {
+  cout << s << "GPSFactorArm on " << keyFormatter(key()) << "\n";
+  cout << "  GPS measurement: " << nT_.transpose() << "\n";
+  cout << "  Lever arm: " << bL_.transpose() << "\n";
+  noiseModel_->print("  noise model: ");
+}
+
+//***************************************************************************
+bool GPSFactor2Arm::equals(const NonlinearFactor& expected, double tol) const {
+  const This* e = dynamic_cast<const This*>(&expected);
+  return e != nullptr && Base::equals(*e, tol) &&
+         traits<Point3>::Equals(nT_, e->nT_, tol) &&
+         traits<Point3>::Equals(bL_, e->bL_, tol);
+}
+
+//***************************************************************************
+Vector GPSFactor2Arm::evaluateError(const NavState& p,
+    OptionalMatrixType H) const {
+  const Matrix3 nRb = p.attitude().matrix();
+  if (H) {
+    H->resize(3, 9);
+
+    H->block<3, 3>(0, 0) = -nRb * skewSymmetric(bL_);
+    H->block<3, 3>(0, 3) = nRb;
+    H->block<3, 3>(0, 6).setZero();
+  }
+
+  return p.position() + nRb * bL_ - nT_;
+}
 
 }/// namespace gtsam
--- a/gtsam/navigation/GPSFactor.cpp
+++ b/gtsam/navigation/GPSFactor.cpp
@@ -26,20 +26,31 @@ namespace gtsam {
 void GPSFactor::print(const string& s, const KeyFormatter& keyFormatter) const {
   cout << (s.empty() ? "" : s + " ") << "GPSFactor on " << keyFormatter(key())
        << "\n";
-  cout << "  GPS measurement: " << nT_ << "\n";
+  cout << "  GPS measurement: " << nT_.transpose() << "\n";
+  cout << "  Lever arm: " << B_t_BG_.transpose() << "\n";
   noiseModel_->print("  noise model: ");
 }
 
 //***************************************************************************
 bool GPSFactor::equals(const NonlinearFactor& expected, double tol) const {
   const This* e = dynamic_cast<const This*>(&expected);
-  return e != nullptr && Base::equals(*e, tol) && traits<Point3>::Equals(nT_, e->nT_, tol);
+  return e != nullptr && Base::equals(*e, tol) &&
+         traits<Point3>::Equals(nT_, e->nT_, tol) &&
+         traits<Point3>::Equals(B_t_BG_, e->B_t_BG_, tol);
 }
 
 //***************************************************************************
 Vector GPSFactor::evaluateError(const Pose3& p,
     OptionalMatrixType H) const {
-  return p.translation(H) -nT_;
+  const Matrix3 rot = p.rotation().matrix();
+  if (H) {
+    H->resize(3, 6);
+
+    H->block<3, 3>(0, 0) = -rot * skewSymmetric(B_t_BG_);
+    H->block<3, 3>(0, 3) = rot;
+  }
+
+  return p.translation() - (nT_ - rot * B_t_BG_);
 }
 
 //***************************************************************************
@@ -67,7 +78,8 @@ pair<Pose3, Vector3> GPSFactor::EstimateState(double t1, const Point3& NED1,
 //***************************************************************************
 void GPSFactor2::print(const string& s, const KeyFormatter& keyFormatter) const {
   cout << s << "GPSFactor2 on " << keyFormatter(key()) << "\n";
-  cout << "  GPS measurement: " << nT_.transpose() << endl;
+  cout << "  GPS measurement: " << nT_.transpose() << "\n";
+  cout << "  Lever arm: " << B_t_BG_.transpose() << "\n";
   noiseModel_->print("  noise model: ");
 }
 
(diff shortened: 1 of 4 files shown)


mnissov:
I see the [macos-14-xcode-15.4 Release Python 3](https://github.com/borglab/gtsam/actions/runs/12751148016/job/35544623027#logs) check failed. This I have very little experience with, but in hindsight I see the serialization and interfacing were not appropriately updated.

Perhaps I should ask the following questions
* Is this of interest for GTSAM?
* Would you prefer the extrinsic added as an estimated state?
* Do you have any tips for resolving the aforementioned failed check, or how I can test this locally. Because running `make check` succeeds across the board.

dellaert:
I’ll take a look and then answer your questions. If it can be done easily with another state then probably that is preferred.

gtsam/navigation/GPSFactor.cpp
```diff
@@ -67,21 +78,32 @@ pair<Pose3, Vector3> GPSFactor::EstimateState(double t1, const Point3& NED1,
...
-         traits<Point3>::Equals(nT_, e->nT_, tol);
+         traits<Point3>::Equals(nT_, e->nT_, tol) &&
+         traits<Point3>::Equals(B_t_BG_, e->B_t_BG_, tol);
 }
 
 //***************************************************************************
 Vector GPSFactor2::evaluateError(const NavState& p,
     OptionalMatrixType H) const {
-  return p.position(H) -nT_;
+  const Matrix3 rot = p.attitude().matrix();
```
dellaert: Please use `aRb` naming convention, where R transforms b coordinates into a coordinates.

gtsam/navigation/GPSFactor.h
```diff
@@ -39,6 +39,7 @@ class GTSAM_EXPORT GPSFactor: public NoiseModelFactorN<Pose3> {
   typedef NoiseModelFactorN<Pose3> Base;
 
   Point3 nT_; ///< Position measurement in cartesian coordinates
+  Point3 B_t_BG_; ///< Lever arm between GPS and BODY frame
```
dellaert: The naming convention ‘nT_’ says that T is in the navigation frame. I think if you want to include a lever arm in this factor you need to include more documentation about the frames involved.

gtsam/navigation/tests/testGPSFactor.cpp
```diff
@@ -61,10 +64,12 @@ TEST( GPSFactor, Constructor ) {
   // Factor
   Key key(1);
   SharedNoiseModel model = noiseModel::Isotropic::Sigma(3, 0.25);
-  GPSFactor factor(key, Point3(E, N, U), model);
+  GPSFactor factor(key, Point3(E, N, U), model, leverArm);
```
dellaert: I would not expect you to change an existing test, but rather add an additional test with a non-zero lever arm.

gtsam/navigation/GPSFactor.cpp
```diff
@@ -67,21 +78,32 @@ pair<Pose3, Vector3> GPSFactor::EstimateState(double t1, const Point3& NED1,
...
+  if (H) {
```
dellaert, line 170: I did not see a numerical derivative test on this Jacobian (right?) so in my book all bets are off :-)

gtsam/navigation/GPSFactor.cpp
```diff
@@ -67,21 +78,32 @@ pair<Pose3, Vector3> GPSFactor::EstimateState(double t1, const Point3& NED1,
...
+    H->resize(3, 9);
+
+    H->block<3, 3>(0, 0) = -rot * skewSymmetric(B_t_BG_);
+    H->block<3, 3>(0, 3) = rot;
+    H->block<3, 3>(0, 6).setZero();
+  }
+
+  return p.position() - (nT_ - rot * B_t_BG_);
```
dellaert: This formula is a bit counter-intuitive to me. I would expect to see $h(x)-z$, where $h(x)$ is the measurement prediction and $z$ the measurement, in this case nT_. If I rewrite your equation that way I get
```
h(x) = p.position() + rot * B_t_BG_
```
From which I deduce that `rot==nRb` which matches my definition of NavState, and `B_t_BG_==bL_`, a lever arm in the body frame. So, I think the documentation/notation I’m looking for is 
```
bL_ is a lever arm in the body frame, denoting the 3D position of the GPS antenna in the body frame. Because the actual location of the antenna depends on both position and attitude, providing a lever arm makes components of the attitude observable in turns.
```

dellaert (commented):
I think this is great, and I think it should be added to a factor. But I propose to add a GpsFactor2 or GpsFactorArm in the same file, because the Jacobians are more expensive. Adding documentation of the nature I suggested in both the old and the new factor would make it easier to see the difference. 

PS we will soon start adding notebooks to document factors in the source tree itself. Consider authoring a small colab that demonstrates both factors that we could include :-) That’s not needed for this PR to merge, though.

gtsam/navigation/GPSFactor.cpp
mnissov: The code I modified in `testGPSFactor.cpp` includes a numerical derivative calculation and comparison, so I think it's covered by this, right?

Of course following the other comment the test needs to be changed as well anyway, so I'll still write the new one :).

gtsam/navigation/GPSFactor.cpp
mnissov: Lever arm is definitely in body frame, will adapt this to GTSAM format.

mnissov:
I think I have addressed all your points in the review with new factors `GPSFactorArm` and `GPSFactor2Arm` mirroring their namesakes but with an argument for lever arm.

I had some follow-up questions
* Do you still want the lever-arm coded in as a state instead of as a constant? It would be pretty simple to do so from my side.
*  Regarding notebooks, is there an example of one I can look at? I assume it wouldn't be so hard for me to make one for this.
* I've left out `EstimateState` when making `GPSFactorArm` as there is now some inconsistency about the `nTb` actually being `nTg` in a way. Of course this is correctable for the pose, less so for the velocity (missing angular rate I think).

gtsam/navigation/GPSFactor.h
```diff
@@ -24,7 +24,10 @@
 namespace gtsam {
 
 /**
- * Prior on position in a Cartesian frame.
+ * Prior on position in a Cartesian frame, assuming position prior is in body
```
dellaert: Hmm, no, I think gpsIn is given in the navigation frame N. Without lever arm, it is a measurement of the origin of the body frame, in the navigation frame. Confusing, I know :-)

gtsam/navigation/GPSFactor.h
```diff
@@ -112,7 +115,73 @@ class GTSAM_EXPORT GPSFactor: public NoiseModelFactorN<Pose3> {
...
+ * and accounts for position measurement vs state discrepancies while turning.
+ * @ingroup navigation
+ */
+class GTSAM_EXPORT GPSFactorArm: public NoiseModelFactorN<Pose3> {
+
+private:
+
+  typedef NoiseModelFactorN<Pose3> Base;
+
+  Point3 nT_;  ///< Position measurement in cartesian coordinates
```
dellaert: Position measurement in cartesian coordinates (navigation frame)

gtsam/navigation/GPSFactor.h
```diff
@@ -112,7 +115,73 @@ class GTSAM_EXPORT GPSFactor: public NoiseModelFactorN<Pose3> {
...
+
+  /// Typedef to this class
+  typedef GPSFactorArm This;
+
+  /// default constructor - only use for serialization
+  GPSFactorArm():nT_(0, 0, 0), bL_(0, 0, 0) {}
+
+  ~GPSFactorArm() override {}
+
+  /// Constructor from a measurement in a Cartesian frame.
```
dellaert: Document all parameters now I think, usie docygen /** */

gtsam/navigation/GPSFactor.h
```diff
@@ -112,7 +115,73 @@ class GTSAM_EXPORT GPSFactor: public NoiseModelFactorN<Pose3> {
...
+  void print(const std::string& s = "", const KeyFormatter& keyFormatter =
+                                            DefaultKeyFormatter) const override;
+
+  /// equals
+  bool equals(const NonlinearFactor& expected, double tol = 1e-9) const override;
+
+  /// vector of errors
+  Vector evaluateError(const Pose3& p, OptionalMatrixType H) const override;
+
+  inline const Point3 & measurementIn() const {
```
dellaert, line 177: /// Return the measurement, in the navigation frame

gtsam/navigation/GPSFactor.h
```diff
@@ -112,7 +115,73 @@ class GTSAM_EXPORT GPSFactor: public NoiseModelFactorN<Pose3> {
...
+    return nT_;
+  }
+
+  inline const Point3 & leverArm() const {
```
dellaert, line 182: /// Return the lever arm, a position in the body frame

gtsam/navigation/GPSFactor.h
```diff
@@ -180,4 +249,68 @@ class GTSAM_EXPORT GPSFactor2: public NoiseModelFactorN<NavState> {
...
+
+  /// Typedef to this class
+  typedef GPSFactor2Arm This;
+
+  /// default constructor - only use for serialization
+  GPSFactor2Arm():nT_(0, 0, 0), bL_(0, 0, 0) {}
+
+  ~GPSFactor2Arm() override {}
+
+  /// Constructor from a measurement in a Cartesian frame.
```
dellaert: Same, doxygen all parameters :-)

gtsam/navigation/GPSFactor.cpp
```diff
@@ -64,6 +64,37 @@ pair<Pose3, Vector3> GPSFactor::EstimateState(double t1, const Point3& NED1,
...
+    OptionalMatrixType H) const {
+  const Matrix3 nRb = p.rotation().matrix();
+  if (H) {
+    H->resize(3, 6);
+
+    H->block<3, 3>(0, 0) = -nRb * skewSymmetric(bL_);
+    H->block<3, 3>(0, 3) = nRb;
+  }
+
+  return p.translation() + nRb * bL_ - nT_;
```
dellaert, line 95: Awesome!

dellaert (approved):
This looks great ! You could probably save a bit of copy/pasta by inheriting from the non-arm factors, but that's entirely optional.

dellaert:
> I had some follow-up questions
> * Do you still want the lever-arm coded in as a state instead of as a constant? It would be pretty simple to do so from my side.

I would say that is a different factor, one that would incur a lot of computation, actually: this new state variable would connect all GPS factors. It would be useful in a calibration setting, but not for everyday use, if you already have a calibrated lever arm.

> * Regarding notebooks, is there an example of one I can look at? I assume it wouldn't be so hard for me to make one for this.

Sure: https://www.roboticsbook.org/S64_driving_perception.html, open in colab :-)

> * I've left out `EstimateState` when making `GPSFactorArm` as there is now some inconsistency about the `nTb` actually being `nTg` in a way. Of course this is correctable for the pose, less so for the velocity (missing angular rate I think).

I don't quite understand this :-)

gtsam/navigation/GPSFactor.h
mnissov: Ah yeah that was silly, think I was thinking of with respect to body frame.

mnissov:
> This looks great ! You could probably save a bit of copy/pasta by inheriting from the non-arm factors, but that's entirely optional.

Thanks! I had considered inheritance, but I had some trouble with base classes and indirect bases. Plus I think inheritance only saves copy-pasting `clone` and `measurementIn`, the rest have, albeit, minor changes. So I didn't pursue this much further.

> I don't quite understand this :-)

I'll try to explain it differently, suppose aPab is the position of {b} with respect to {a}, expressed in coordinate frame {a} (for coordinate frames {a} and {b}). If there is a lever arm, the measurement is really wPwg instead of wPwb for world frame {w}, GPS antenna {g}, and body frame {b}.

Therefore `EstimateState`, if there is a nonzero lever arm, calculates not velocity wVwb but instead wVwg. So if the lever arm is large, while rotating there can be a significant difference. In addition, rotating about {b} means that the position of {g} will change, but not the position of {b}. The position difference can be corrected I think, but the velocity not without knowing the angular rate as well.

This is what I was trying to say. I hope this makes a little more sense.

dellaert:
> > This looks great ! You could probably save a bit of copy/pasta by inheriting from the non-arm factors, but that's entirely optional.
> 
> Thanks! I had considered inheritance, but I had some trouble with base classes and indirect bases. Plus I think inheritance only saves copy-pasting `clone` and `measurementIn`, the rest have, albeit, minor changes. So I didn't pursue this much further.

Agreed !

> > I don't quite understand this :-)
> This is what I was trying to say. I hope this makes a little more sense.

Yep, I think so :-)

So, this is ready to merge after CI succeeds, yes? I propose we do a notebook in a follow-up PR, if you're game. In fact, we're soon going to add a mechanism to GTSAM to check these in and they will be automatically served. And, we can featsure them as blog posts on gtsam.org, as well.

mnissov:
Sorry about that, I was a bit slow but I added a factor with the lever arm as a state as well. Ended up calling it `GPSFactorArmCalib`, which is a bit verbose but I'm struggling to find something else :).

> So, this is ready to merge after CI succeeds, yes?

Otherwise yes, should be all good from my side.

> I propose we do a notebook in a follow-up PR, if you're game. In fact, we're soon going to add a mechanism to GTSAM to check these in and they will be automatically served. And, we can featsure them as blog posts on gtsam.org, as well.

Yeah I am down for that. The model for this is a numerical example and a bit of description/story-telling?

dellaert:
Thanks @mnissov !!! Will be quite useful, makes orientation observable if you have two receivers, as well :-)

PR #1979, opened 2025-01-15
Faster SE(3) and SE_2(3) exponential maps

dellaert:
`DexpFunctor` was instantiated several times:
  - for computing R via `SO3::Expmap`
  - for `ExmapTranslation`, *twice* so in `NavState`
 
This PR 
- gets rid of `ExpmapTranslation` and inlines it in the 4 places it was used;
- in the `Pose3` and `Navstate` `Expmaps`, we also now use the *one* instantiated `DexpFunctor` to compute the rotation;
- according to Barfoot Eq 8.83, X = J_r * inv(J_l) = R^T, so substitute that wherever we used X before: another big saving!
- in addition, I realized why R^T: it's the jacobian of Pose3::Create !

This PR saves a bunch of compute (many sines and cosines, and computation of skew W and W*W) and is more "transparent" about what is going on. In addition, the X=R^T saves even more on top of that.

### DIFF
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -189,7 +189,7 @@ Pose3 Pose3::Expmap(const Vector6& xi, OptionalJacobian<6, 6> Hxi) {
 
   // Compute rotation using Expmap
 #ifdef GTSAM_USE_QUATERNIONS
-  const Rot3 R = traits<gtsam::Quaternion>::Expmap(v);
+  const Rot3 R = traits<gtsam::Quaternion>::Expmap(w);
 #else
   const Rot3 R(local.expmap());
 #endif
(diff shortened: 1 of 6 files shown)


dellaert:
@timbarfoot I was inspired to look at this code more critically because of you :-)

gtsam/geometry/Pose3.cpp
```diff
@@ -183,16 +183,35 @@ Pose3 Pose3::Expmap(const Vector6& xi, OptionalJacobian<6, 6> Hxi) {
   // Get angular velocity omega and translational velocity v from twist xi
   const Vector3 w = xi.head<3>(), v = xi.tail<3>();
 
+  // Instantiate functor for Dexp-related operations:
+  const bool nearZero = (w.dot(w) <= 1e-5);
```
ProfFan, line 187: `bool nearZero = (w.dot(w) <= nearZeroThreshold);` down below

ProfFan: Is this intended?

ProfFan (changes requested):

gtsam/geometry/Pose3.cpp
dellaert: Yes. This has not changed.

dellaert: This PR is threshold neutral. It's not configurable nor consistent, but this PR does not change that in any way and doe not intend to fix it.

ProfFan (approved):

PR #1985, opened 2025-01-21
Marginalize Discrete Modes

varunagrawal:
Add a flag to `HybridBayesNet::prune` which enables dead mode removal.

Here, we check if a certain discrete mode is only a single value. If so, we remove it from the pruned joint discrete conditional, as well as convert all `HybridGaussianConditional`s which only depend on the dead modes to a `GaussianConditional`.


gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -66,6 +68,30 @@ HybridBayesNet HybridBayesNet::prune(size_t maxNrLeaves) const {
...
+      int index = -1;
+      auto threshold = (probabilities.array() > 0.99);
+      // If atleast 1 value is non-zero, then we can find the index
+      // Else if all are zero, index would be set to 0 which is incorrect
+      if (!threshold.isZero()) {
+        threshold.maxCoeff(&index);
+      }
+
+      if (index >= 0) {
+        deadModesValues.insert(std::make_pair(dkey.first, index));
```
dellaert: emplace?

gtsam/hybrid/HybridConditional.cpp
```diff
@@ -169,4 +169,25 @@ double HybridConditional::evaluate(const HybridValues &values) const {
   return std::exp(logProbability(values));
 }
 
+/* ************************************************************************ */
+void HybridConditional::removeDiscreteModes(const DiscreteValues &given) {
```
dellaert: Could be moved to DiscreteConditional?

gtsam/hybrid/tests/testHybridBayesNet.cpp
```diff
@@ -390,17 +398,82 @@ TEST(HybridBayesNet, Pruning) {
...
+
+  // Check that discrete joint only has M0 and not (M0, M1)
+  // since M0 is removed
+  KeyVector actual_keys = prunedBayesNet.at(0)->asDiscrete()->keys();
+  EXPECT(KeyVector{M(0)} == actual_keys);
+
+  // Check that hybrid conditionals that only depend on M1 are no longer hybrid
+  EXPECT(prunedBayesNet.at(0)->isDiscrete());
+  EXPECT(prunedBayesNet.at(1)->isHybrid());
+  // Only P(X2 | X1, M1) depends on M1, so it is Gaussian
```
dellaert: update comments?

dellaert (approved):
Amazing. Just some small comments

gtsam/hybrid/HybridConditional.cpp
varunagrawal: Moved it and changed the for loop to a `std::for_each` to be more idiomatic.

PR #1986, opened 2025-01-21
Improvements to Hybrid Inference

varunagrawal:
Made some improvements over the weekend that helped with debugging and writing examples.

1. Separate `mpe()` method for discrete max-product.
2. `DiscreteValues::insert` to add only a single key-value pair.
3. Format `DiscreteMarginals.h`.


dellaert (approved):
LGTM!

PR #1977, opened 2025-01-14
Fix/1452

JokerJohn:
Add a minimal  test case of  `IncrementalFixedLagSmoother`  in `gtsam/gtsam_unstable/examples` folder using real-world data to showcase how setting `findUnusedFactorSlots = true` addresses the issue #1452 in GTSAM.

PR #1987, opened 2025-01-22
fix the link error of IncrementalFixedLagsmootherExample

JokerJohn:
The newly added **IncrementalFixedLagsmootherExample.cpp** from [issue 1452](https://github.com/borglab/gtsam/pull/1977) cannot be placed in the `example` directory since it requires link to the **gtsam_unstable** module, which is not enabled by default. Therefore, I moved it to `gtsam_unstable/example`.

image


dellaert (approved):

dellaert:
Thanks !!!

PR #1989, opened 2025-01-23
City10000 dataset example

varunagrawal:
Example using hybrid estimation for the City10000 dataset, which is also used in MH-ISAM2 by Hsiao and Kaess.

Result below for 900 timesteps. Green dashes are ground truth, red is from ISAM2 (continuous only) with the ambiguous measurements, and blue is using the HybridNonlinearISAM optimizer (not ISAM2 based) again with ambiguous measurements.
hybrid_city10000


dellaert (approved):
Cool

Issue #1994, opened 2025-01-23
Build fails on WSL Ubuntu 24


juanlm-verus:
Following up from #1981 

Running WSL2

```

Distributor ID: Ubuntu
Description:    Ubuntu 24.04.1 LTS
Release:        24.04
Codename:       noble

```

Following instructions as described in the INSTALL.md file:

```

$ mkdir build
$ cd build
$ cmake ..
$ make check # (optional, runs unit tests)
$ make install

```

I get the following:

```

juan@XLS-LT-482PCC:/mnt/c/Users/juan.lopera/gtsam$ mkdir build && cd build
juan@XLS-LT-482PCC:/mnt/c/Users/juan.lopera/gtsam/build$ cmake ..
-- The CXX compiler identification is GNU 13.3.0
-- The C compiler identification is GNU 13.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- GTSAM_POSE3_EXPMAP=ON, enabling GTSAM_ROT3_EXPMAP as well
-- Performing Test COMPILER_HAS_WSUGGEST_OVERRIDE
-- Performing Test COMPILER_HAS_WSUGGEST_OVERRIDE - Success
-- Performing Test COMPILER_HAS_WMISSING_OVERRIDE
-- Performing Test COMPILER_HAS_WMISSING_OVERRIDE - Failed
-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.83.0/BoostConfig.cmake (found suitable version "1.83.0", minimum required is "1.65") found components: serialization system filesystem thread program_options date_time timer chrono regex 
-- Found Eigen version: 3.4.0
CMake Deprecation Warning at gtsam/3rdparty/metis/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- Looking for execinfo.h
-- Looking for execinfo.h - found
-- Looking for getline
-- Looking for getline - found
[...]

Issue #1967, opened 2025-01-10
ROS build farm are broken for develop branch


jlblancoc:
## Description

See: https://build.ros2.org/job/Jdev__gtsam__ubuntu_noble_amd64/136/

## Why?

Reason: the use of embedded Eigen version by default leads to not using the `SYSTEM` flag for those includes, so any warning there is treated as an error. 

## Solutions 
To modify CMake scripts so..
1) Use Eigen from the system version by default, 
2) Like (1), but only when it's detected we are building for the ROS build farm (kida hack, but have been there...)
3) Add the SYSTEM flag to embedded version of Eigen (if not done already, right? @Gold856  )

What do you think, @dellaert ?

I would like (1), but I know you like backwards compatible behavior, so the second choice would probably be (3), and if it's already done but is not working for some cmake limitation (cannot use SYSTEM for non-system directories? never checked that..), then use (2).

dellaert:
You are correctly modeling my view: (3) if possible, (2) if not :-)

jlblancoc:
SYSTEM was already there: 
https://github.com/borglab/gtsam/blob/develop/cmake/HandleEigen.cmake#L45

so it will have to be (2)...

Gold856:
What if we instead use a pragma in the problematic file to suppress that specific warning? That's probably a much cleaner solution and much more fine grained than using a pragma in an Eigen header and having it affect all downstream consumers.
Alternatively, if this is a big enough problem, we could put the pragma in DisableStupidWarnings.h like originally planned.

jlblancoc:
Mmm... Do you mean the embedded Eigen3's `DisableStupidWarnings.h` file? If you agree with this, it would be another workaround. 
It would imply remembering about this when/if someday the embedded version is upgraded. 

Do you agree with this change, Frank @dellaert ?

Gold856:
Yes, that's the header I was referencing. Although I still think using pragmas in the affected gtsam files is a better method, mostly because of the granularity of control we get and the fact that it means we don't need to mess with Eigen. Also, it's likely that system Eigen packages would also be affected, so this solution would ensure that the warnings don't trigger on system Eigen packages as well.

jlblancoc:
I agree with you, @Gold856 . 

If you have time for a small PR , please tag me so I can check it. If not I'll add it to my list...
Log with the error: https://build.ros2.org/job/Jdev__gtsam__ubuntu_noble_amd64/136/console

[...]

PR #1995, opened 2025-01-23
fix: add GTSAM_USE_BOOST_FEATURES to definitions

ArkadiuszNiemiec:
This fixes the issue #1994 and some issues mentioned in #1967 and  #1981.

The change defines the proper identifier if `GTSAM_USE_BOOST_FEATURES` option is set in CMakeLists.txt.
It's used in:
https://github.com/borglab/gtsam/blob/7b56d6689ce685de9c5d88c0fd35c9e5338844a1/gtsam/base/concepts.h#L11-L24
...and without it the compiler throws `BOOST_CONCEPT_USAGE` redefinition error.

### DIFF
--- a/gtsam/base/concepts.h
+++ b/gtsam/base/concepts.h
@@ -8,6 +8,8 @@
 
 #pragma once
 
+#include <gtsam/config.h>
+
 #if GTSAM_USE_BOOST_FEATURES
 #include <boost/concept_check.hpp>
 #include <boost/concept/assert.hpp>


dellaert:
Thanks, but I might not fully understand then: I checked and we don't do this for GTSAM_ENABLE_BOOST_SERIALIZATION, but we can still branch on that in code. Why is GTSAM_USE_BOOST_FEATURES different?

ArkadiuszNiemiec:
> Thanks, but I might not fully understand then: I checked and we don't do this for GTSAM_ENABLE_BOOST_SERIALIZATION, but we can still branch on that in code. Why is GTSAM_USE_BOOST_FEATURES different?

Oh, right. The `GTSAM_ENABLE_BOOST_SERIALIZATION` didn't generate a build error for me, but it should be added as well. Fixed.

dellaert:
Well, I guess my question was the other way :-) Why can we branch on GTSAM_ENABLE_BOOST_SERIALIZATION currently? And could we use the same mechanism to branch on GTSAM_USE_BOOST_FEATURES? It seems the `add_definitions` was not needed for GTSAM_ENABLE_BOOST_SERIALIZATION?

ArkadiuszNiemiec:
AFAIK both identifiers/macros were always not defined in the code. My linter confirms that but to prove it I added this to one of the examples:
image 

...checked out my first commit (with just `GTSAM_USE_BOOST_FEATURES` fixed), made sure the flag is ON and built:
image

The output:
image


...and you are right. For some reason both are ON 😕 So why it's not working during compile time?

dellaert:
right :-)
No idea ! Sleuthing required!
But if we find a]out why we can avoid the “add_definitions” step.
Note: it requires `#if` and not `#ifdef`. Maybe there is a straggling `#ifdef` ?

dellaert:
Note, your test *is* compile time. All macro pre-processing is gone at runtime.

ArkadiuszNiemiec:
OK, got it. GTSAM is using https://github.com/borglab/gtsam/blob/develop/gtsam/config.h.in
It creates a config.h that needs to be included if one want to use the CMake variables. To include was missing from that one specific file. Thank you @dellaert, great questions!

dellaert (approved):
Ah ! Nice, LGTM :-) I will kick off CI and we can merge.

varunagrawal:
FYI `add_definitions` is an inefficient approach. Please see #1944 for details.

Also sorry I didn't see this before, was busy with other stuff. Else we could have fixed this sooner. :)

PR #1996, opened 2025-01-24
Better `HybridSmoother`

varunagrawal:
Updated the `HybridSmoother` to
1. Only use conditionals connected to the variables which are connected to the new factors being added.
2. Improve the `getOrdering` method to give orderings that **don't include all the variables**.
3. Removed extra loops and made the whole things a bit more efficient.


dellaert (approved):
LGTM. Some weirdness in .h file, doxygen of get ordering …

varunagrawal:
Fixed the doxygen issue. That was my mistake, trying to regenerate the docstring automatically.

CI will pass once we merge in #1999.

dellaert:
Merge at will

PR #1997, opened 2025-01-24
Clean up cartesian product

dellaert:
- and address comments from #1991

PR #1998, opened 2025-01-24
Fix/refactor cartesian product

dellaert:
Replaces #1997


varunagrawal (approved):
Am I missing something? Why not update 1997?

PR #1999, opened 2025-01-25
Fix Dead Mode Removal

varunagrawal:
Currently, I am encountering an issue where if we remove a dead mode and then we add a factor which has the mode key in it, the elimination algorithm errors out in the construction of the elimination tree. This is particularly true if we add a discrete chain, where we have conditionals like $P(m_{k+1} | m_k)$ and $m_k$ was the dead mode removed.

The simplest fix is to add the mode marginals to the `HybridBayesNet` when we remove the mode from the joint so that any future factors that get added referencing those modes don't complain during elimination.

I also added a check for the joint discrete conditional being empty (aka no keys) and we no longer add it. A new unit test was added to confirm this works, but failed before.

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.h
+++ b/gtsam/hybrid/HybridBayesNet.h
@@ -217,15 +217,15 @@ class GTSAM_EXPORT HybridBayesNet : public BayesNet<HybridConditional> {
    * @brief Prune the Bayes Net such that we have at most maxNrLeaves leaves.
    *
    * @param maxNrLeaves Continuous values at which to compute the error.
-   * @param removeDeadModes Flag to enable removal of modes which only have a
-   * single possible assignment.
    * @param deadModeThreshold The threshold to check the mode marginals against.
    * If greater than this threshold, the mode gets assigned that value and is
    * considered "dead" for hybrid elimination.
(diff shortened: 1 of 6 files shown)


gtsam/hybrid/HybridBayesNet.h
```diff
@@ -219,9 +219,13 @@ class GTSAM_EXPORT HybridBayesNet : public BayesNet<HybridConditional> {
    * @param maxNrLeaves Continuous values at which to compute the error.
    * @param removeDeadModes Flag to enable removal of modes which only have a
    * single possible assignment.
+   * @param deadModeThreshold The threshold to check the mode marginals against.
+   * If greater than this threshold, the mode gets assigned that value and is
+   * considered "dead" for hybrid elimination.
    * @return A pruned HybridBayesNet
    */
-  HybridBayesNet prune(size_t maxNrLeaves, bool removeDeadModes = false) const;
+  HybridBayesNet prune(size_t maxNrLeaves, bool removeDeadModes = false,
```
dellaert: You can combine both arguments with a std optionial

dellaert (approved):
Just one comment

PR #2000, opened 2025-01-25
Update installation and build documentation for Windows & Python on Windows

p-zach:
There are some workarounds involved (e.g. commenting out lines in generated files) but these steps work reliably to build on Windows 10 and build the Python wrapper.


varunagrawal (approved):
LGTM! Thank you

ProfFan:
Hmm, we should probably fix the Python wrapper include issue itself if that is the cause...

PR #1978, opened 2025-01-14
Fix GCC 13 warnings

Gold856:
Fixes #1967 and fixes #1981.

PR #2002, opened 2025-01-25
Silence a warning about a memcpy overread for gcc 13

jmackay2:
This is the final fix to get gtsam building again on Ubuntu 24.04. I'm not super thrilled just ignoring this warning. This could probably use a revisit on ways to fix the issue. 

Here is the error this is fixing:
```
    inlined from ‘void Eigen::internal::call_assignment(Dst&, const Src&) [with Dst = Eigen::Block<Eigen::Matrix<double, -1, -1>, -1, -1, false>; Src = Eigen::Matrix<double, -1, -1>]’ at gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/AssignEvaluator.h:836:18,
    inlined from ‘Derived& Eigen::MatrixBase<Derived>::operator=(const Eigen::DenseBase<OtherDerived>&) [with OtherDerived = Eigen::Matrix<double, -1, -1>; Derived = Eigen::Block<Eigen::Matrix<double, -1, -1>, -1, -1, false>]’ at gtsam/gtsam/3rdparty/Eigen/Eigen/src/Core/Assign.h:66:28,
    inlined from ‘void gtsam::JacobianFactor::fillTerms(const TERMS&, const gtsam::Vector&, const gtsam::SharedDiagonal&) [with TERMS = std::map<long unsigned int, Eigen::Matrix<double, 1, 1, 0, 1, 1> >]’ at gtsam/gtsam/linear/JacobianFactor-inl.h:94:14:
/usr/lib/gcc/x86_64-linux-gnu/13/include/emmintrin.h:169:19: error: ‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ reading 16 or more bytes from a region of size 8 [-Werror=stringop-overread]
  169 |   *(__m128d *)__P = __A;
      |   ~~~~~~~~~~~~~~~~^~~~~
```

This seems to be pointing to this assignment: https://github.com/borglab/gtsam/blob/a744cfcdd8e278dd65ab3f70208cffbd262e1533/gtsam/linear/JacobianFactor-inl.h#L94

This fix is based on the same solution used in #1978 . 

I am able to get a successful Ubuntu 24.04 build after this fix.

### DIFF
--- a/gtsam/linear/JacobianFactor-inl.h
+++ b/gtsam/linear/JacobianFactor-inl.h
@@ -20,6 +20,10 @@
 
 #include <gtsam/linear/linearExceptions.h>
 
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic warning "-Wstringop-overread"
+#endif
+
 namespace gtsam {
 
   /* ************************************************************************* */
--- a/gtsam/linear/JacobianFactor-inl.h
+++ b/gtsam/linear/JacobianFactor-inl.h
@@ -20,7 +20,7 @@
 
 #include <gtsam/linear/linearExceptions.h>
 
-#if defined(__GNUC__) && !defined(__clang__)
+#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 13
 #pragma GCC diagnostic warning "-Wstringop-overread"
 #endif
 


dellaert:
CI issue?

ProfFan:
Need to guard against older GCC:
error: unknown option after ‘#pragma GCC diagnostic’ kind [-Werror=pragmas]

dellaert:
Ping @jmackay2

dellaert (approved):
Thanks!!!!

PR #2003, opened 2025-01-25
move constructors

dellaert:
I added move constructors for GaussianConditional, JacobianFactor, and VerticalBlockMatrix, hoping to see a performance difference, but if anything it made it worse. Below are timings for timeBatch and timeSFMBAL, where 
- `Before` is without any move constructors
- `After` is without VerticalBlockMatrix move constructors (just GaussianConditional, JacobianFactor)
- `After2` is with VerticalBlockMatrix move constructors

image

`After` is 2% *more* expensive, and `After2` restores that to "about the same".

Any ideas, e.g., @ProfFan ?

### DIFF
--- a/gtsam/linear/HessianFactor.cpp
+++ b/gtsam/linear/HessianFactor.cpp
@@ -470,7 +470,7 @@ std::shared_ptr<GaussianConditional> HessianFactor::eliminateCholesky(const Orde
 
     // TODO(frank): pre-allocate GaussianConditional and write into it
     const VerticalBlockMatrix Ab = info_.split(nFrontals);
-    conditional = std::make_shared<GaussianConditional>(keys_, nFrontals, Ab);
+    conditional = std::make_shared<GaussianConditional>(keys_, nFrontals, std::move(Ab));
 
     // Erase the eliminated keys in this factor
     keys_.erase(begin(), begin() + nFrontals);
--- a/gtsam/linear/HessianFactor.cpp
+++ b/gtsam/linear/HessianFactor.cpp
@@ -245,7 +245,7 @@ HessianFactor::HessianFactor(const GaussianFactorGraph& factors,
 
   // Form A' * A
   gttic(update);
-  info_.setZero();
+  info_.setAllZero();
   for(const auto& factor: factors)
     if (factor)
       factor->updateHessian(keys_, &info_);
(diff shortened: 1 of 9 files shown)


gtsam/linear/HessianFactor.cpp
```diff
@@ -470,7 +470,7 @@ std::shared_ptr<GaussianConditional> HessianFactor::eliminateCholesky(const Orde
 
     // TODO(frank): pre-allocate GaussianConditional and write into it
     const VerticalBlockMatrix Ab = info_.split(nFrontals);
-    conditional = std::make_shared<GaussianConditional>(keys_, nFrontals, Ab);
+    conditional = std::make_shared<GaussianConditional>(keys_, nFrontals, std::move(Ab));
```
dellaert, line 471: This is called *many* times. I had hoped this would make a performance difference.

ProfFan:
image

ProfFan:
these two timing scripts are probably too small for any perf benefits to show up

Currently a lot of time is spent in
```
    Scatter scatter(factors, keys);
    jointFactor = std::make_shared<HessianFactor>(factors, scatter);
```

dellaert:
Hmmm, weird. All right, I’ll use instruments to profile.

dellaert:
> these two timing scripts are probably too small for any perf benefits to show up
> 
> Currently a lot of time is spent in
> 
> ```
>     Scatter scatter(factors, keys);
>     jointFactor = std::make_shared<HessianFactor>(factors, scatter);
> ```

Yeah. I found the same. But I think it’s legit computation.

dellaert:
@ProfFan will you review? I think the changes in this PR are sensible, even if they only yielded a few % points. I could gather some more profiling with the experiments @varunagrawal is doing now, which are very CPU intensive.

ProfFan (approved):

dellaert:
This made a noticeable difference in a large experiment, 5-10% speedup.

gtsam/linear/HessianFactor.cpp
versatran01: `Ab` is `const` and moving it has no effect.

dellaert: Wow, ok. Let me see whether that does anything.

PR #2004, opened 2025-01-26
Hybrid Fixes

varunagrawal:
Check for null factors and allow for continuous only `HybridBayesTree`.

Made these fixes during debugging the City10000 example.


dellaert (approved):
LGTM!

PR #2005, opened 2025-01-27
Search for K-best solutions

dellaert:
A new class that uses branch-and-bound search to find the K-best solutions quickly. Should be much more efficient than forming the product and enumerating all assignments. Works for both DiscreteBayesNets and DiscreteBayesTrees.

## Testing:

> time gtsam/discrete/tests/testDiscreteSearch

Number of expansions: 35  <<  Asia **Bayes net**
Number of expansions: 33  <<  Asia **Bayes tree**
There were no test failures

real    0m0.027s
user    0m0.006s
sys     0m0.010s

### DIFF
--- a/gtsam/discrete/DiscreteSearch.cpp
+++ b/gtsam/discrete/DiscreteSearch.cpp
@@ -171,48 +171,48 @@ DiscreteSearch::DiscreteSearch(const DiscreteBayesTree& bayesTree) {
   costToGo_ = computeCostToGo(conditionals_);
 };
 
-using SearchNodeQueue = std::priority_queue<SearchNode, std::vector<SearchNode>,
-                                            SearchNode::Compare>;
-
-std::vector<Solution> DiscreteSearch::run(size_t K) const {
-  SearchNodeQueue expansions;
-  expansions.push(SearchNode::Root(conditionals_.size(),
-                                   costToGo_.empty() ? 0.0 : costToGo_.back()));
-
-  Solutions solutions(K);
-
-  auto expandNextNode = [&] {
+struct SearchNodeQueue
+    : public std::priority_queue<SearchNode, std::vector<SearchNode>,
+                                 SearchNode::Compare> {
+  void expandNextNode(
+      const std::vector<DiscreteConditional::shared_ptr>& conditionals,
+      const std::vector<double>& costToGo, Solutions* solutions) {
     // Pop the partial assignment with the smallest bound
-    SearchNode current = expansions.top();
-    expansions.pop();
+    SearchNode current = top();
+    pop();
 
-    // If we already have K solutions, prune if we cannot beat the worst
-    // one.
-    if (solutions.prune(current.bound)) {
+    // If we already have K solutions, prune if we cannot beat the worst one.
+    if (solutions->prune(current.bound)) {
       return;
     }
 
     // Check if we have a complete assignment
     if (current.isComplete()) {
-      solutions.maybeAdd(current.error, current.assignment);
+      solutions->maybeAdd(current.error, current.assignment);
       return;
     }
 
     // Expand on the next factor
-    const auto& conditional = conditionals_[current.nextConditional];
+    const auto& conditional = conditionals[current.nextConditional];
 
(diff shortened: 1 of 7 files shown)


dellaert:
@varunagrawal review this version? Substantial refactor coming in next PR

gtsam/discrete/DiscreteSearch.cpp
```diff
@@ -0,0 +1,246 @@
...
+  }
+};
+
+struct CompareSolution {
+  bool operator()(const Solution& a, const Solution& b) const {
+    return a.error < b.error;
+  }
+};
+
+// Define the Solutions class
```
varunagrawal, line 98: Maybe more information on why and where this is used so one doesn't have to go hunting through the source code?

gtsam/discrete/DiscreteSearch.cpp
```diff
@@ -0,0 +1,246 @@
...
+    if (full && error >= pq_.top().error) return false;
+    if (full) pq_.pop();
+    pq_.emplace(error, assignment);
+    return true;
+  }
+
+  /// Check if we have any solutions
+  bool empty() const { return pq_.empty(); }
+
+  // Method to print all solutions
```
varunagrawal, line 120: Mix of regular and doxygen-like comments. Prefer to be doxygen.

gtsam/discrete/DiscreteSearch.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+  /// Check if (partial) solution with given bound can be pruned. If we have
+  /// room, we never prune. Otherwise, prune if lower bound on error is worse
+  /// than our current worst error.
+  bool prune(double bound) const {
+    if (pq_.size() < maxSize_) return false;
+    return bound >= pq_.top().error;
+  }
+
+  // Method to extract solutions in ascending order of error
```
varunagrawal, line 139: Same as before.

gtsam/discrete/DiscreteValues.cpp
```diff
@@ -26,12 +26,24 @@ using std::stringstream;
...
 
+/* ************************************************************************ */
+static void stream(std::ostream& os, const DiscreteValues& x,
+                   const KeyFormatter& keyFormatter) {
+  for (const auto& kv : x)
+    os << "(" << keyFormatter(kv.first) << ", " << kv.second << ")";
+}
+
+/* ************************************************************************ */
+std::ostream& operator<<(std::ostream& os, const DiscreteValues& x) {
```
varunagrawal, line 37: I like this. I did this for a bunch of classes in `geometry`, really helps with flexibility in using both `.print()` and `<<`.

We could make this a potential GSoC project.

gtsam/discrete/tests/AsiaExample.h
```diff
@@ -0,0 +1,61 @@
...
+DiscreteBayesNet createAsiaExample() {
+  DiscreteBayesNet asia;
+  asia.add((Dyspnea | Either, Bronchitis) = "9/1 2/8 3/7 1/9");
+  asia.add(XRay | Either = "95/5 2/98");
+  asia.add(Bronchitis | Smoking = "70/30 40/60");
+  for (const auto& factor : createFragment()) asia.push_back(factor);
+  return asia;
+}
+}  // namespace asia_example
+}  // namespace gtsam
```
varunagrawal, line 61: New line?

varunagrawal (approved):
Some docstring improvements to be had, but pretty straightforward.

dellaert:
OK, I will address all these comments in #2006

gtsam/discrete/DiscreteSearch.cpp
dellaert: One remark: doxygen is a language to document APIs, and in GTSAM is restricted to headers only - at least that's how I always saw them. Some structs here were migrated from header to cpp file, and hence have doxygen remnants. I will convert all to concise // and /* */ style.

varunagrawal: Got it. I was a bit confused because there's both types of docstrings in this struct.

PR #2006, opened 2025-01-28
Search for k-best in factor graph

dellaert:
This PR builds upon #2005, now enabling constructing a DiscreteSearch instance from a factor graph. This avoids having to do costly inference to transform into a Bayes net or Bayes tree. There are two versions:
- one that groups factors using an elimination tree, one per variable in the given ordering
- one that additionally groups the factors according to cliques, multi-frontal style

There is still a bit of compute that happens in the constructor: factors get multiplied together for each "search slot". So, this does the "product" of "sum-product", but not the sum, not the division to get a conditional.

The choice between e-tree or j-tree is documented:
```c++
  /**
   * Construct from a DiscreteFactorGraph.
   *
   * Internally creates either an elimination tree or a junction tree. The
   * latter incurs more up-front computation but the search itself might be
   * faster. Then again, for the elimination tree, the heuristic will be more
   * fine-grained (more slots).
   *
   * @param factorGraph The factor graph to search over.
   * @param ordering The ordering used to create etree (and maybe jtree).
   * @param buildJunctionTree Whether to build a junction tree or not.
   */
  static DiscreteSearch FromFactorGraph(const DiscreteFactorGraph& factorGraph,
                                        const Ordering& ordering,
                                        bool buildJunctionTree = false);
```

## Testing:
All unit tests pass, and all variants give the same answer for Asia, for both MPE and 4-best solutions:
```c++
  auto fromETree =
      DiscreteSearch::FromFactorGraph(asia::factorGraph, asia::ordering);
  auto fromJunctionTree =
      DiscreteSearch::FromFactorGraph(asia::factorGraph, asia::ordering, true);
  const DiscreteSearch fromBayesNet(asia::bayesNet);
  const DiscreteSearch fromBayesTree(asia::bayesTree);

  for (auto& search :
       {fromETree, fromJunctionTree, fromBayesNet, fromBayesTree}) {
    // Ask for the MPE
    auto mpe = search.run();

    EXPECT_LONGS_EQUAL(1, mpe.size());
    // Regression test: check the MPE solution
    EXPECT_DOUBLES_EQUAL(1.236627, std::fabs(mpe[0].error), 1e-5);

    // Check it is equal to MPE via inference
    EXPECT(assert_equal(asia::mpe, mpe[0].assignment));

    // Ask for top 4 solutions
    auto solutions = search.run(4);

    EXPECT_LONGS_EQUAL(4, solutions.size());
    // Regression test: check the first and last solution
    EXPECT_DOUBLES_EQUAL(1.236627, std::fabs(solutions[0].error), 1e-5);
    EXPECT_DOUBLES_EQUAL(2.201708, std::fabs(solutions[3].error), 1e-5);
  }
```


gtsam/discrete/DiscreteJunctionTree.cpp
```diff
@@ -16,19 +16,43 @@
...
+template class EliminatableClusterTree<DiscreteBayesTree, DiscreteFactorGraph>;
+template class JunctionTree<DiscreteBayesTree, DiscreteFactorGraph>;
+
+/* ************************************************************************* */
+DiscreteJunctionTree::DiscreteJunctionTree(
+    const DiscreteEliminationTree& eliminationTree)
+    : Base(eliminationTree) {}
+/* ************************************************************************* */
+namespace {
+struct PrintForestVisitorPre {
```
varunagrawal: If this is only used in `DiscreteJunctionTree::print`, maybe move it inside that method to achieve the same encapsulation as putting it inside a namespace?

gtsam/discrete/DiscreteSearch.cpp
```diff
@@ -150,97 +134,154 @@ class Solutions {
   }
 };
 
+/// @brief Get the factor associated with a node, possibly product of factors.
+template <typename NodeType>
+static auto getFactor(const NodeType& node) {
```
varunagrawal: Not a fan of the `auto` return type. Makes it harder to read and debug later on.

gtsam/discrete/DiscreteSearch.h
```diff
@@ -28,9 +32,32 @@ namespace gtsam {
  */
 class GTSAM_EXPORT DiscreteSearch {
  public:
-  /**
-   * @brief A solution to a discrete search problem.
-   */
+  /// We structure the search as a set of slots, each with a factor and
```
varunagrawal: Nit: we should be using `/** */` for multiline docstrings.

varunagrawal (approved):
LGTM modulo some minor comments.

gtsam/discrete/DiscreteJunctionTree.cpp
dellaert: done, with a lambda

PR #2007, opened 2025-01-28
fix hybrid uninitialized warning

jmackay2:
There is an error when building on Ubuntu 24.04:
```
gtsam/examples/Hybrid_City10000.cpp:207:16: error: ‘key_t’ may be used uninitialized [-Werror=maybe-uninitialized]
  207 |   write_results(results, (key_t + 1), "HybridISAM_city10000.txt");
```

This solves the error by initializing `key_t` to remove the warning. This was tested by performing a successful build.

### DIFF
--- a/examples/Hybrid_City10000.cpp
+++ b/examples/Hybrid_City10000.cpp
@@ -120,7 +120,7 @@ int main(int argc, char* argv[]) {
 
   SmootherUpdate(smoother, graph, init_values, maxNrHypotheses, &results);
 
-  size_t key_s, key_t;
+  size_t key_s, key_t{0};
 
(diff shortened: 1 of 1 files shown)


varunagrawal:
So strange, I don't get this issue? What version of GCC are you using? It would behoove us to add a CI check with that version if it is stricter.

varunagrawal (approved):
Thank you!

jmackay2:
> So strange, I don't get this issue? What version of GCC are you using? It would behoove us to add a CI check with that version if it is stricter.

I'm on Ubuntu 24.04 with gcc 13.3.0.

PR #2010, opened 2025-01-28
Updates to City10000 Example

varunagrawal:
Improved script for inference on the City10000 dataset.

1. Added loop closure factors.
2. Timing information
3. Results for 3000 timesteps
4. Scale discrete factors so we don't get all 0s.
5. Various other bug fixes.

### DIFF
--- a/examples/Hybrid_City10000.cpp
+++ b/examples/Hybrid_City10000.cpp
@@ -141,7 +141,7 @@ int main(int argc, char* argv[]) {
 
   std::list<double> time_list;
 
-  HybridSmoother smoother;
+  HybridSmoother smoother(0.99);
 
   HybridNonlinearFactorGraph graph;
 
--- a/examples/Hybrid_City10000.cpp
+++ b/examples/Hybrid_City10000.cpp
@@ -44,7 +44,7 @@ using symbol_shorthand::M;
 using symbol_shorthand::X;
 
 // Testing params
-const size_t max_loop_count = 3000;  // 2000;  // 200 //2000 //8000
+const size_t max_loop_count = 2000;  // 2000;  // 200 //2000 //8000
 
 noiseModel::Diagonal::shared_ptr prior_noise_model =
     noiseModel::Diagonal::Sigmas(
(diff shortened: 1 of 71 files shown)


gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -222,6 +222,8 @@ namespace gtsam {
     // sum out frontals, this is the factor on the separator
     gttic(sum);
     DiscreteFactor::shared_ptr sum = product->sum(frontalKeys);
+    // Normalize/scale to prevent underflow.
```
dellaert: Ok, this is now wrong again, right?

varunagrawal: Yes, give me a minute.

examples/Hybrid_City10000.cpp
```diff
@@ -75,15 +76,54 @@ void write_results(const Values& results, size_t num_poses,
...
+ * @return HybridNonlinearFactor
+ */
+HybridNonlinearFactor HybridLoopClosureFactor(size_t loop_counter, size_t key_s,
+                                              size_t key_t,
+                                              const Pose2& measurement) {
+  DiscreteKey l(L(loop_counter), 2);
+
+  auto f0 = std::make_shared<BetweenFactor<Pose2>>(
+      X(key_s), X(key_t), measurement,
+      noiseModel::Diagonal::Sigmas(Vector3::Ones() * 10));
```
varunagrawal: @dellaert this was the change. I updated it from 100 to 10.

dellaert:
I have substantially refactored the script, which is already in develop. It would be nice to remove the copy/ pasta between both examples and have a city10k header file that has a class that can read all the things and create constraints, etc..

alternatively, and this might be much easier, create a second method in the script class currently in develop that does iSam and remove the copy/paste one. I really like that idea actually, it’s probably super-easy.

varunagrawal:
@dellaert I went with option 1 since it was cleaner and for new folks they know exactly which script to look at.

dellaert (approved):
LGTM!

PR #2011, opened 2025-01-28
kill expNormalize in favor of scale

dellaert:
Small PR with some orthogonal changes made in #2010 
Adding as separate PR to check correctness in CI

### DIFF
--- a/gtsam/discrete/DiscreteFactorGraph.cpp
+++ b/gtsam/discrete/DiscreteFactorGraph.cpp
@@ -215,6 +215,7 @@ namespace gtsam {
   EliminateDiscrete(const DiscreteFactorGraph& factors,
                     const Ordering& frontalKeys) {
     gttic(product);
+    // `product` is scaled later to prevent underflow.
     DiscreteFactor::shared_ptr product = factors.product();
     gttoc(product);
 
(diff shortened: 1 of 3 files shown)


varunagrawal (approved):

varunagrawal:
Yeah I considered making this a separate PR but then the City10000 script wouldn't work.

dellaert:
Uhm, unfortunately, this made many unit tests fail. Could you take a look?

varunagrawal:
Yup, I'll finish some writing and then jump into this after dinner.

gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -222,6 +219,8 @@ namespace gtsam {
     // sum out frontals, this is the factor on the separator
     gttic(sum);
     DiscreteFactor::shared_ptr sum = product->sum(frontalKeys);
+    // Normalize/scale to prevent underflow.
+    sum = sum->scale();
```
dellaert: @varunagrawal I think this is the only new thing and probably wrong?

dellaert:
OK, @varunagrawal , I think I fixed, we'll let CI say so.

gtsam/discrete/DiscreteFactorGraph.cpp
```diff
@@ -217,12 +214,21 @@ namespace gtsam {
...
     // sum out frontals, this is the factor on the separator
     gttic(sum);
     DiscreteFactor::shared_ptr sum = product->sum(frontalKeys);
     gttoc(sum);
+    
+    // Normalize/scale to prevent underflow.
+    gttic(scale);
+    DiscreteFactor::shared_ptr denominator = sum->max(sum->size());
+    product = product->operator/(denominator);
+    sum = sum->operator/(denominator);
```
varunagrawal, line 230: I don't get it. Why does this work?
We went from doing `product/max(product)` to `product/max(sum)`, and in both cases `sum->scale()`. The denominators will cancel later, but  haven't we now changed what was computed?

varunagrawal:
@dellaert do you want to merge this directly?

dellaert:
Done.

PR #2013, opened 2025-01-30
Fix pruning

dellaert:
- moved pruning and dead mode removal to DiscreteBayesNet
- renamed threshold to `marginalThreshold`
- two new DiscreteValues support methods
- made docs more consistent there
- added a `restrict` method in HybridConditional
- used it inside prune
- no  longer adding marginals
- add discrete factors to end in smoother

Hybrid and discrete tests pass

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.cpp
+++ b/gtsam/hybrid/HybridBayesNet.cpp
@@ -19,7 +19,6 @@
 #include <gtsam/discrete/DiscreteBayesNet.h>
 #include <gtsam/discrete/DiscreteConditional.h>
 #include <gtsam/discrete/DiscreteFactorGraph.h>
-#include <gtsam/discrete/DiscreteMarginals.h>
 #include <gtsam/discrete/TableDistribution.h>
 #include <gtsam/hybrid/HybridBayesNet.h>
 #include <gtsam/hybrid/HybridValues.h>
@@ -43,10 +42,6 @@ bool HybridBayesNet::equals(const This &bn, double tol) const {
 }
 
 /* ************************************************************************* */
-// The implementation is: build the entire joint into one factor and then prune.
-// TODO(Frank): This can be quite expensive *unless* the factors have already
-// been pruned before. Another, possibly faster approach is branch and bound
-// search to find the K-best leaves and then create a single pruned conditional.
 HybridBayesNet HybridBayesNet::prune(
     size_t maxNrLeaves, const std::optional<double> &deadModeThreshold) const {
 #if GTSAM_HYBRID_TIMING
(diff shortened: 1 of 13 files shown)


gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -43,115 +42,60 @@ bool HybridBayesNet::equals(const This &bn, double tol) const {
...
+    DiscreteValues *fixedValues) const {
+#if GTSAM_HYBRID_TIMING
+  gttic_(HybridPruning);
+#endif
   // Collect all the discrete conditionals. Could be small if already pruned.
   const DiscreteBayesNet marginal = discreteMarginal();
 
+  // Prune discrete Bayes net
+  DiscreteValues fixed;
+  auto prunedBN = marginal.prune(maxNrLeaves, marginalThreshold, &fixed);
```
varunagrawal, line 56: This is just a `DiscreteBayesNet` so no need for auto. Makes the code easier to read and helps autocomplete.

Also potentially rename to `prunedDBN` to signify it is discrete.

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -43,115 +42,60 @@ bool HybridBayesNet::equals(const This &bn, double tol) const {
...
+
   // Multiply into one big conditional. NOTE: possibly quite expensive.
-  DiscreteConditional joint;
-  for (auto &&conditional : marginal) {
-    joint = joint * (*conditional);
+  DiscreteConditional pruned;
+  for (auto &&conditional : prunedBN) pruned = pruned * (*conditional);
```
varunagrawal, line 60: This is not needed since by construction `prunedBN` only has 1 or 0 conditionals.
It should be:
```cpp
DiscreteConditional pruned;
if (prunedDBN.size() > 0) pruned = prunedDBN.at(0);
```

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -43,115 +42,60 @@ bool HybridBayesNet::equals(const This &bn, double tol) const {
 }
 
 /* ************************************************************************* */
-// The implementation is: build the entire joint into one factor and then prune.
-// TODO(Frank): This can be quite expensive *unless* the factors have already
-// been pruned before. Another, possibly faster approach is branch and bound
-// search to find the K-best leaves and then create a single pruned conditional.
 HybridBayesNet HybridBayesNet::prune(
-    size_t maxNrLeaves, const std::optional<double> &deadModeThreshold) const {
+    size_t maxNrLeaves, const std::optional<double> &marginalThreshold,
```
varunagrawal, line 46: I am not a fan of `marginalThreshold` since it doesn't indicate which marginals we're talking about (discrete? continuous?), and it seems to indicate we already have marginals which we wish to threshold on, not compute marginals.
Better name please?

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -43,115 +42,60 @@ bool HybridBayesNet::equals(const This &bn, double tol) const {
...
+
   // Multiply into one big conditional. NOTE: possibly quite expensive.
-  DiscreteConditional joint;
-  for (auto &&conditional : marginal) {
-    joint = joint * (*conditional);
+  DiscreteConditional pruned;
+  for (auto &&conditional : prunedBN) pruned = pruned * (*conditional);
+
+  // Set the fixed values if requested.
+  if (marginalThreshold && fixedValues) {
```
varunagrawal, line 63: Any reason we don't just pass in `fixedValues` to `marginal.prune` like we do for computing `OptionalJacobian`? This check is done within `DiscreteBayesNet::prune` so we're already set.

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -43,115 +42,60 @@ bool HybridBayesNet::equals(const This &bn, double tol) const {
...
-          result.push_back(
-              prunedHybridGaussianConditional->choose(deadModesValues));
-        } else {
-          // Add as-is
-          result.push_back(prunedHybridGaussianConditional);
-        }
-      } else {
-        // Type-erase and add to the pruned Bayes Net fragment.
-        result.push_back(prunedHybridGaussianConditional);
+      if (!prunedHybridGaussianConditional) {
```
varunagrawal, line 81: Do we want to throw this exception?
All other operations already check for `nullptr` or empty so we can still get the same result. This exception would be an annoyance in a perfectly formed problem.

Rather we should let it error out, and add checks where missing. I found a few myself last week.

gtsam/hybrid/HybridBayesNet.cpp
```diff
@@ -43,115 +42,60 @@ bool HybridBayesNet::equals(const This &bn, double tol) const {
...
-          std::dynamic_pointer_cast<DiscreteConditional>(marginals(key)));
-    }
+    // No-op if not a HybridGaussianConditional.
+    if (marginalThreshold) conditional = conditional->restrict(fixed);
 
-  } else {
-    result.emplace_shared<DiscreteConditional>(pruned);
-  }
-
-  /* To prune, we visitWith every leaf in the HybridGaussianConditional.
```
varunagrawal, line 112: Maybe keep this comment for future reference?

dellaert:
@varunagrawal ?

gtsam/hybrid/HybridConditional.cpp
```diff
@@ -169,4 +169,41 @@ double HybridConditional::evaluate(const HybridValues &values) const {
...
+HybridConditional::shared_ptr HybridConditional::restrict(
+    const DiscreteValues &discreteValues) const {
+  if (auto gc = asGaussian()) {
+    return std::make_shared<HybridConditional>(gc);
+  } else if (auto dc = asDiscrete()) {
+    return std::make_shared<HybridConditional>(dc);
+  };
+
+  auto hgc = asHybrid();
+  if (!hgc)
```
varunagrawal, line 182: Not needed as `asHybrid` will error out during the conversion.

varunagrawal (approved):
LGTM with some comments to improve things.

gtsam/hybrid/HybridBayesNet.cpp
dellaert: Not in general, we could opt to use K-best

gtsam/hybrid/HybridBayesNet.cpp
dellaert: It could be zero and I need result here

gtsam/hybrid/HybridConditional.cpp
dellaert: No it is a nullptr I think

gtsam/hybrid/HybridBayesNet.cpp
dellaert: I like this. We literally threshold the marginals

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: I was hoping to cross that bridge when we get there.

gtsam/hybrid/HybridBayesNet.cpp
varunagrawal: 0 is still a value right? Isn't that why we opted to make the threshold a `std::optional`?

PR #2017, opened 2025-02-02
Re-linearization

dellaert:
- extensive implementation of restrict and re-linearization
- some are not implemented, which is ok for now.

### DIFF
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -536,5 +536,11 @@ namespace gtsam {
     return DecisionTreeFactor(this->discreteKeys(), thresholded);
   }
 
+/* ************************************************************************ */
+DiscreteFactor::shared_ptr DecisionTreeFactor::restrict(
+    const DiscreteValues& assignment) const {
+  throw std::runtime_error("DecisionTreeFactor::restrict not implemented");
+}
+
   /* ************************************************************************ */
 }  // namespace gtsam
--- a/gtsam/hybrid/HybridConditional.h
+++ b/gtsam/hybrid/HybridConditional.h
@@ -153,7 +153,8 @@ class GTSAM_EXPORT HybridConditional
    * @return HybridGaussianConditional::shared_ptr otherwise
    */
   HybridGaussianConditional::shared_ptr asHybrid() const {
-    return std::dynamic_pointer_cast<HybridGaussianConditional>(inner_);
+    if (!isHybrid()) return nullptr;
+    return std::static_pointer_cast<HybridGaussianConditional>(inner_);
   }
 
   /**
@@ -162,7 +163,8 @@ class GTSAM_EXPORT HybridConditional
    * @return GaussianConditional::shared_ptr otherwise
    */
   GaussianConditional::shared_ptr asGaussian() const {
-    return std::dynamic_pointer_cast<GaussianConditional>(inner_);
+    if (!isContinuous()) return nullptr;
+    return std::static_pointer_cast<GaussianConditional>(inner_);
   }
 
   /**
@@ -172,7 +174,8 @@ class GTSAM_EXPORT HybridConditional
    */
   template <typename T = DiscreteConditional>
   typename T::shared_ptr asDiscrete() const {
-    return std::dynamic_pointer_cast<T>(inner_);
+    if (!isDiscrete()) return nullptr;
+    return std::static_pointer_cast<T>(inner_);
   }
 
   /// Get the type-erased pointer to the inner type
@@ -221,7 +224,8 @@ class GTSAM_EXPORT HybridConditional
    * which is just a GaussianConditional. If this conditional is *not* a hybrid
    * conditional, just return that.
    */
-  shared_ptr restrict(const DiscreteValues& discreteValues) const;
+  std::shared_ptr<Factor> restrict(
+      const DiscreteValues& assignment) const override;
 
   /// @}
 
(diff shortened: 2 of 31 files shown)


dellaert:
Yay, @varunagrawal , tests finally pass. Had to fix some stragglers.

gtsam/hybrid/HybridConditional.h
```diff
@@ -153,7 +153,8 @@ class GTSAM_EXPORT HybridConditional
    * @return HybridGaussianConditional::shared_ptr otherwise
    */
   HybridGaussianConditional::shared_ptr asHybrid() const {
-    return std::dynamic_pointer_cast<HybridGaussianConditional>(inner_);
+    if (!isHybrid()) return nullptr;
+    return std::static_pointer_cast<HybridGaussianConditional>(inner_);
```
varunagrawal, line 157: Wow we've been doing it wrong for a while now.

varunagrawal: Actually do we want this? If we call `asHybrid` on a non-hybrid factor, it should error out. I recommend updating the `(!isHybrid())` check to throw an exception perhaps?

gtsam/hybrid/HybridConditional.h
```diff
@@ -162,7 +163,8 @@ class GTSAM_EXPORT HybridConditional
    * @return GaussianConditional::shared_ptr otherwise
    */
   GaussianConditional::shared_ptr asGaussian() const {
-    return std::dynamic_pointer_cast<GaussianConditional>(inner_);
+    if (!isContinuous()) return nullptr;
```
varunagrawal, line 166: Preferably throw exception instead of returning a nullptr since this can be a silent error/failure.

gtsam/hybrid/HybridConditional.h
```diff
@@ -172,7 +174,8 @@ class GTSAM_EXPORT HybridConditional
    */
   template <typename T = DiscreteConditional>
   typename T::shared_ptr asDiscrete() const {
-    return std::dynamic_pointer_cast<T>(inner_);
+    if (!isDiscrete()) return nullptr;
```
varunagrawal, line 177: Preferably throw exception instead of returning a nullptr since this can be a silent error/failure.

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -165,15 +169,16 @@ class GTSAM_EXPORT HybridNonlinearFactor : public HybridFactor {
   bool equals(const HybridFactor& other, double tol = 1e-9) const override;
 
   /// @}
+  /// @name Standard API
+  /// @{
```
varunagrawal, line 173: Thank you!

gtsam/hybrid/HybridNonlinearFactorGraph.cpp
```diff
@@ -221,5 +221,30 @@ AlgebraicDecisionTree<Key> HybridNonlinearFactorGraph::discretePosterior(
...
+    if (!f) {
+      continue;
+    }
+    // Check if it is a hybrid factor
+    if (auto hf = dynamic_pointer_cast<HybridFactor>(f)) {
+      result.push_back(hf->restrict(discreteValues));
+    } else if (auto df = dynamic_pointer_cast<DiscreteFactor>(f)) {
+      result.push_back(df->restrict(discreteValues));
+    } else {
+      result.push_back(f);  // Everything else is just added as is
```
varunagrawal, line 242: Add a TODO here. Potentially lots of performance left on the table here, particularly in `HybridNonlinearISAM` where we keep track of the nonlinear factors and relinearize.

gtsam/discrete/DecisionTreeFactor.cpp
```diff
@@ -536,5 +536,11 @@ namespace gtsam {
     return DecisionTreeFactor(this->discreteKeys(), thresholded);
   }
 
+/* ************************************************************************ */
+DiscreteFactor::shared_ptr DecisionTreeFactor::restrict(
+    const DiscreteValues& assignment) const {
+  throw std::runtime_error("DecisionTreeFactor::restrict not implemented");
```
varunagrawal, line 542: This should just be `Base::restrict` and convert to a factor, right?

varunagrawal (approved):
LGTM. I can address the comments in my own PR or a follow up.

varunagrawal:
Merging so I can merge into my own PR.

gtsam/hybrid/HybridConditional.h
dellaert: No, @varunagrawal , this is used to find out whether it is of this type. I don’t think we should mess with that semantics

gtsam/discrete/DecisionTreeFactor.cpp
dellaert: Hmm, is already a factor, so it should restrict using the fact that it is a decision tree, and then return a restricted decision tree. Base::restrict can’t do that.

gtsam/hybrid/HybridNonlinearFactorGraph.cpp
dellaert: Maybe. These are shared pointers so not costly here

gtsam/hybrid/HybridConditional.h
dellaert: No, disagree. I did not change the semantics here, just the performance. I like these semantics.

gtsam/hybrid/HybridConditional.h
dellaert: Nope

PR #2019, opened 2025-02-09
Python script for hybrid inference on City10000 dataset

varunagrawal:
Wrapped a bunch of methods and recreated the script `Hybrid_City10000.cpp` as `HybridCity10000.py`.

I get a similar final error:
`Final error: 0.6931471805599453`
and a similar total time:
`Total time: 21.018189191818237 seconds`

Please see the results of this new script below.

hybrid_city10000_py

Update: I added a python script for plotting the results. It takes an arbitrary number of estimate files (.txt) but needs a corresponding number of labels and colors associated.


examples/Hybrid_City10000.cpp
```diff
@@ -127,7 +127,7 @@ class Experiment {
     auto bayesNet = linearized->eliminateSequential();
     HybridValues delta = bayesNet->optimize();
     initial_ = initial_.retract(delta.continuous());
-    smoother_.reInitialize(std::move(*bayesNet));
+    smoother_.reInitialize(*bayesNet);
```
dellaert: Why? Seems appropriate here

gtsam/hybrid/HybridSmoother.h
```diff
@@ -49,8 +49,8 @@ class GTSAM_EXPORT HybridSmoother {
   /**
    * Re-initialize the smoother from a new hybrid Bayes Net.
    */
-  void reInitialize(HybridBayesNet&& hybridBayesNet) {
-    hybridBayesNet_ = std::move(hybridBayesNet);
+  void reInitialize(HybridBayesNet& hybridBayesNet) {
```
dellaert: Just add another non && constructor rather than deleting this one?

dellaert (approved):
Cool. I’d undo the std::move changes though. Just adding. Constructor should low you to wrap.

PR #2020, opened 2025-02-10
IMU factor with Gravity Estimation

nkhedekar:
As discussed here (https://groups.google.com/g/gtsam-users/c/BMlDTdKvjRg) the intended use of the IMU factor is with a constant gravity vector (that may be pointing either up or down based on the user's convention) in the navigation frame and an initial attitude that should be estimated. Inaccuracy in this initial attitude will have a large detrimental effect with larger translations in the XY plane. This problem becomes even more evident when creating a large monolithic map as in LiDAR-Inertial Odometry. Estimating the initial orientation is additionally challenging if the motion of the robot does not involve rotations (until much further into the trajectory) making the problem unobservable.

A possible solution to this was presented in Lupton and Sukkarieh's paper [1](Section V). Particularly, they proposed fixing the initial body frame to be the navigation frame and estimated the direction of gravity in it. While [2] theoretically improved on [1], this solution from [1] seems to not have been implemented. 

This PR adds a new IMUFactorWithGravity so that this solution can be implemented. The IMUFactorWithGravity is a 6-way factor that is essentially a modification of the IMUFactor to connect it to a variable for the gravity while trying to not duplicate too much code. 

Note: This first version implements directly a `Vector3` for the gravity. Since the gravity magnitude is known to a high degree of accuracy around the globe, its probably better to have the gravity as a `Unit3` with the magnitude being specified by norm of the vector in the preintegration params (to maintain compatibility)? 

#### References
[1] T. Lupton and S. Sukkarieh, "Visual-Inertial-Aided Navigation for High-Dynamic Motion in Built Environments Without Initial Conditions," in IEEE Transactions on Robotics, vol. 28, no. 1, pp. 61-76, Feb. 2012, doi: 10.1109/TRO.2011.2170332.
[2] Christian Forster, Luca Carlone, Frank Dellaert, Davide Scaramuzza, "IMU Preintegration on Manifold for Efficient Visual-Inertial Maximum-a-Posteriori Estimation" in Robotics: Science and Systems XI


nkhedekar:
[5f18305](https://github.com/borglab/gtsam/pull/2020/commits/5f183050431f681560099ff949540439cd611dd3) sets the gravity up as a `Unit3` for the direction and utilizes the `p().n_gravity.norm()` for the magnitude.

dellaert:
Hi @nkhedekar 

Thanks for the PR - we always welcome contributions. But if I'm honest - I don't fully understand the reason for this.

Our IMU factor is based on Lupton & Sukkarieh. We are not fixing any initial orientation of the body frame: that is part of the optimization process. That was also my answer in that discussion thread.

The assumption we make is that the gravity is constant and aligned with either Z-up or Z-down in the navigation frame. I don't really see a good reason to *not* assume that: an earth-fixed tangent frame is what most of our users assume. In particular, making the original body frame the navigation frame is suboptimal as it would require - as you implement here - tying the gravity direction to every IMU factor. I see the cost, but not really the benefit :-) 

Can you explain why optimizing over the whole trajectory (which will recover gravity as well, by changing the entire trajectory to align) is not the solution? If done incrementally, because every *non-accelerating* IMUY measurements points to gravity, the correct orientation should be stable very quickly.

Best 
FD

dellaert:
@nkhedekar maybe you want to chat with me about it separately? Send me an email at dellaert@gmail.com and we can reserve some time.

nkhedekar:
> @nkhedekar maybe you want to chat with me about it separately? Send me an email at [dellaert@gmail.com](mailto:dellaert@gmail.com) and we can reserve some time.

Sure! I have sent an email :)

dellaert:
@nkhedekar nice chatting! Let me know via email when I should look at this again.

nkhedekar:
Hi @dellaert , 
After significant delays (my bad), this PR is ready for another look. The branch is rewritten from scratch on develop to to add maintainable clear functionality of optimizable gravity within the different variants of IMU factors.

The default path of the IMU factors remains unchanged, so if one is operating in a gravity-aligned nav frame nothing changes. The new factors introduced here are for the cases where this assumption doesn't hold:
1. nav frames that are *not* gravity-aligned by construction, e.g. LiDAR-odometry map frames anchored to the first scan (several recent LIO systems (FAST-LIO2, Wildcat, D-LIOM, and Nemiroff/Chen/Lopez [3] (who do it on S² inside iSAM2) estimate gravity for this reason)
2. the initialization-free operation of Lupton & Sukkarieh [1, Sec. V], where the graph is anchored on the initial pose and gravity becomes the estimated quantity

The core change is now just an optional 9×3 Jacobian: gravity enters preintegration only in the `correctPIM` correction (as shown in [1]), so `NavState::correctPIM` and `PreintegrationBase::predict/computeError/computeErrorAndJacobians` gained gravity-vector overloads with `∂ξ/∂g = [0; ½Δt²Rᵀ; ΔtRᵀ]`, and every existing signature delegates to them unchanged.

Both parametrizations are provided, as thin chain-rule adapters over one Jacobian:

- `ImuFactorWithGravityDirection` - `Unit3` direction × fixed magnitude (explicit constructor parameter, defaulting to `‖params.n_gravity‖`), the preferred form on Earth. This reuses the scale×direction decomposition already in the codebase via a new shared `Unit3::scaled()`, which `MagFactor3` now also uses.
- `ImuFactorWithGravityVector` - free `Point3`, faithful to [1], paired with a new general-purpose `VectorNormFactor<3>` as the magnitude pseudo-observation (added once per gravity variable; also useful for `MagFactor2` field magnitudes).
- `CombinedImuFactorWithGravity{Direction,Vector}` — the same for the combined factor (bias rows have zero gravity Jacobian).

Naming somewhat follows SymForce, which added the equivalent `ImuWithGravityFactor` / `ImuWithGravityDirectionFactor` pair (https://symforce.org/api-cpp/file/imu__factor_8h.html) since this PR was opened.

Beyond the factors, this PR also includes analytic-vs-numerical Jacobian tests for both parametrizations × both preintegration backends, tilted-gravity recovery tests for both modes, serialization round-trips, Merge with gravity-key/magnitude checks, wrapper entries with a Python smoke test, a "Gravity as an Optimized Variable" section in `doc/ImuFactor.lyx`/`.pdf` (citing [1] and [3]), executed notebooks (`ImuFactorWithGravity.ipynb`, `VectorNormFactor.ipynb`) following the per-class notebook conventions, cross-links from the ImuFactor/CombinedImuFactor/MagFactor notebooks, and `python/gtsam/examples/ImuFactorWithGravityExample.py`. Full `make check` passes (352 suites). 

[1] T. Lupton and S. Sukkarieh, "Visual-Inertial-Aided Navigation for High-Dynamic Motion in Built Environments Without Initial Conditions," in IEEE Transactions on Robotics, vol. 28, no. 1, pp. 61-76, Feb. 2012, doi: 10.1109/TRO.2011.2170332.

[2] Christian Forster, Luca Carlone, Frank Dellaert, Davide Scaramuzza, "IMU Preintegration on Manifold for Efficient Visual-Inertial Maximum-a-Posteriori Estimation" in Robotics: Science and Systems XI

[3] R. Nemiroff, K. Chen and B. T. Lopez, "Joint On-Manifold Gravity and Accelerometer Intrinsics Estimation for Inertially Aligned Mapping," 2023 IEEE/RSJ International Conference on Intelligent Robots and Systems (IROS), Detroit, MI, USA, 2023, pp. 1388-1394, doi: 10.1109/IROS55552.2023.10342424.

dellaert:
I apologize for the delay in reviewing. 

Some basic things first, will read more carefully soon:

Please rebase on develop and be careful to merge conflicts carefully in favor of develop, where appropriate, e.g., in PreintegrationBase.cpp: current develop changed Jacobian assignments from Eigen's comma initializer << to ordinary assignment. Please make sure to adopt that style in the tests.

wrapper structure diverges from repository guidance. The templated gravity factors are exposed as separate handwritten direction/vector wrapper classes. We require exact parameter-name agreement and prefer the normal .i template mechanism, so  please 
- express these bindings through wrapper templates where supported, letting the generator create the concrete names
- make all wrapper parameter names match the headers exactly. 
If the generator cannot represent the two-template-parameter pattern, document that limitation and minimize the manual declarations.

gtsam/navigation/MagFactor.h
```diff
@@ -209,16 +209,16 @@ class MagFactor3: public NoiseModelFactorN<double, Unit3, Point3> {
       const Point3& bias, OptionalMatrixType H1,
       OptionalMatrixType H2, OptionalMatrixType H3) const override {
     // measured bM = nRb� * nM + b, where b is unknown bias
-    Unit3 rotated = bRn_.rotate(direction, OptionalNone, H2);
-    Point3 hx = scale * rotated.point3() + bias;
+    Matrix22 D_rotated_direction;
```
dellaert: what's happening here? Maybe should be separate PR?

nkhedekar: Sure, will move it to a separate PR :+1:

nkhedekar:
Thanks for the review! I think the new commits address all your points, and the branch is now rebased onto develop. Here's a summary of the changes:

- Rebased onto today's develop (ffd471d5ca40a10a3a06e9ce83478817f1e507b0) and resolved conflicts in develop's favor.
- The four handwritten classes are gone. Both factor
families are now expressed through the normal template mechanism, following the
`RangeFactor` pattern in `sam.i`:

  ```
  template <PIM, GRAVITY>
  virtual class ImuFactorWithGravityT : gtsam::NonlinearFactor { ... };
  typedef gtsam::ImuFactorWithGravityT<gtsam::PreintegratedImuMeasurements,
                                      gtsam::Unit3> ImuFactorWithGravityDirection;
  typedef gtsam::ImuFactorWithGravityT<gtsam::PreintegratedImuMeasurements,
                                      gtsam::Point3> ImuFactorWithGravityVector;
  ```

  (and the same for `CombinedImuFactorWithGravityT`), so the generator creates the concrete names — the two-template-parameter pattern is representable, no manual declarations needed. All wrapper parameter names now match the headers exactly. Python-visible names are unchanged, so the tests/example/notebooks needed no edits.
- Tests pass, the generated bindings expose all four factor names with header-matching argument names, and `python/gtsam/tests/test_ImuFactorWithGravity.py` (5 tests) and `ImuFactorWithGravityExample.py` both pass against the freshly built wrapper.
- One note per your point (c): `VectorNormFactor3` remains a single handwritten entry because the generator's template mechanism only accepts type parameters, and `VectorNormFactor` is templated on `int N`; the entry now carries a comment documenting exactly that, and it wraps the C++ alias `VectorNormFactor3 = VectorNormFactor<3>` so the name exists in C++.

Scope note: there is deliberately no `ImuFactor2` (NavState-keyed) gravity variant in this PR — the `PreintegrationBase` gravity overloads already contain everything it needs, so I plan to add it in a separate follow-up PR to keep this one's scope contained.

One consequence of the shared template: the Point3-typedef'd
classes expose the `gravityMagnitude` constructor overload too. Rather than silently ignoring it, the constructor now throws `std::invalid_argument`(→ `ValueError` in Python) if a magnitude is passed to the Point3 parametrization, with a message pointing at `VectorNormFactor<3>` as the way to constrain the magnitude of a free gravity vector. Covered by new C++ and Python tests; `Merge` forwards the magnitude only for the Unit3 parametrization.

gtsam/nonlinear/nonlinear.i
```diff
@@ -879,6 +879,22 @@ virtual class ConcentratedGaussian : gtsam::ExtendedPriorFactor<T> {
   This operator*(const This& other) const;
 };
 
+#include <gtsam/nonlinear/VectorNormFactor.h>
+// Wrapped by hand: the wrapper template mechanism cannot instantiate the
```
dellaert: I think it can :-)

gtsam/navigation/ImuFactor.h
```diff
@@ -436,6 +436,209 @@ using ImuFactor2 = ImuFactor2T<>;
 template <class PIM>
 GTSAM_EXPORT std::ostream& operator<<(std::ostream& os, const ImuFactor2T<PIM>& f);
 
+/**
+ * ImuFactorWithGravityT is a 6-ways factor: in addition to the previous and
```
dellaert: I'd prefer adding a new header ImuFactorWithGravity.h. Same for Combined.

dellaert (commented):
Some more comments this morning. Still reading detailed code but I'm on it today :-) Hope you are ok with piecemeal review....

gtsam/nonlinear/nonlinear.i
nkhedekar: Yep, it can :sweat_smile: Changed

gtsam/navigation/ImuFactor.h
nkhedekar: Moved it :+1:

dellaert (approved):
Amazing! Will run CI...

dellaert:
Excellent contribution! Many thanks @nkhedekar !!!

nkhedekar:
Thank you @dellaert ! :smile:

dellaert:
@nkhedekar are you interested in doing a blog post on how it enables a downstream app?

nkhedekar:
Of course! I'll submit a PR in gtsam.org late next week :smile:

PR #2021, opened 2025-02-11
Set up MyST site deploy

p-zach:
Set up MyST Python documentation site for `borglab.github.io/gtsam`. Pose2.ipynb is my work based on Pose3.ipynb by @dellaert. I will continue to make PRs for both of these and notebooks for more gtsam classes. MyST site is bare bones for now and still has "Made with MyST" in the top left, will require a .png logo to replace that (cannot just replace with text). 

**Before this change is merged,** the following steps must be taken to set up GH pages in the gtsam repo:

1. Navigate to your GitHub Pages settings
2. Enable GitHub Pages
3. Use GitHub Actions as the source

After that, the site should redeploy automatically after all changes to .md or .ipynb files in the develop branch (docs are built from the develop branch, not main)


p-zach:
Currently the site looks like this:

image

myst.yml
```diff
@@ -0,0 +1,29 @@
...
+  id: 7a62a86d-a893-4ab1-9473-b1a957f78902
+  title: GTSAM Docs
+  description: GTSAM is a library of C++ classes that implement smoothing and mapping (SAM) in robotics and vision, using factor graphs and Bayes networks as the underlying computing paradigm rather than sparse matrices.
+  github: https://github.com/borglab/gtsam
+  toc:
+    - file: README.md
+    - file: INSTALL.md
+    - title: Python Reference
+      children: 
+        - pattern: './doc/PythonReference/*'
```
dellaert: So, I think we want these notebooks- if possible, alongside the c++ source code in a `doc` folder (next to `tests`). Even though the example in code in these is python, I see that as incidental - the content (aside from the demo code) should be language agnostic: it is a about the math.

myst.yml
```diff
@@ -0,0 +1,29 @@
...
+version: 1
+project:
+  id: 7a62a86d-a893-4ab1-9473-b1a957f78902
+  title: GTSAM Docs
+  description: GTSAM is a library of C++ classes that implement smoothing and mapping (SAM) in robotics and vision, using factor graphs and Bayes networks as the underlying computing paradigm rather than sparse matrices.
+  github: https://github.com/borglab/gtsam
+  toc:
+    - file: README.md
+    - file: INSTALL.md
+    - title: Python Reference
```
dellaert: This would be then become “User Guide”. I think we need to have a top-level and then specifically (and in a specific order) list the sub-modules: base, inference,geometry, slam, etc…

myst.yml
```diff
@@ -0,0 +1,29 @@
+# See docs at: https://mystmd.org/guide/frontmatter
+version: 1
+project:
+  id: 7a62a86d-a893-4ab1-9473-b1a957f78902
+  title: GTSAM Docs
+  description: GTSAM is a library of C++ classes that implement smoothing and mapping (SAM) in robotics and vision, using factor graphs and Bayes networks as the underlying computing paradigm rather than sparse matrices.
+  github: https://github.com/borglab/gtsam
+  toc:
```
dellaert, line 8: There are already various markdown files in the docs folder that we might also want to include.

dellaert (commented):
Very cool - some comments.
Let’s talk in person about which URL we want this at.
And whether there is a way to avoid. A full CI if we make changes to only notebooks and md files…

p-zach:
Full CI can be avoided by making a few changes to the `.github/workflows/build-*.yml`s. Something like here: https://stackoverflow.com/questions/62968897/is-it-possible-to-not-run-github-action-for-readme-updates

Location of docs changed in commit to follow. User guide folder structure is easily implementable through the myst.yml table of contents. I'll move the other docs around to more uniform places in a later commit or after we get the site up.

p-zach:
Site now has this layout: 
image

dellaert:
Ok, maybe try to add the “path ignore” changes to the workflows immediately here? Seems silly to run unit tests on 34 different configurations if you’re only changing a markdown or ipynb.

of course, you’d want to run the new myst workflow…

dellaert:
PS why borglab.github.io/gtsam? Can we make it appear at gtsam.org/docs instead?

p-zach:
Made that change.

`<user/org>.github.io/<repo_name>` is the default host location for a GH pages site. Someone with maintainer privileges on this repo can change the host location to `gtsam.org/docs` by editing the GH pages settings; the deploy.yml will also need to be modified slightly in that case (`BASE_URL: /${{ github.event.repository.name }}` to `BASE_URL: /docs` I think).

Edit: the gtsam.org DNS records will also need to be modified to accept the GH pages host, I believe.

dellaert:
Dang - maybe we’ll just go with the default URL and then redirect from gtsam org

dellaert (approved):
Awesomesauce!

akshay-krishnan:
> Site now has this layout: image

Looks great! Could we change the "Made with MyST" on the top left corner to something else?

p-zach:
> Could we change the "Made with MyST" on the top left corner to something else?

Yes. It's an image, not text, so it has to be replaced by another image, but we'll replace it when (or before) we migrate from the borglab.github.io url to gtsam.org. Still a WIP.

PR #2023, opened 2025-02-13
Add overviews for existing docs folders on MyST site

p-zach:
Add overview files on the site for User guide and Geometry as requested @dellaert. Also, as expected, this PR does not run checks now.
Edit: Actually, develop still "requires" a check so it shows up despite not running anything. It doesn't actually waste any compute, so if having "Some checks haven't completed yet" show up on a docs PR isn't undesirable, no changes needed. Otherwise, general settings or specific branch protection settings for develop will need to be changed to remove the required check.

### DIFF
--- a/gtsam/geometry/geometry.md
+++ b/gtsam/geometry/geometry.md
@@ -1,0 +1,3 @@
+# Geometry
+
+The geometry folder contains classes that handle fundamental geometric representations and transformations commonly used in applications of GTSAM.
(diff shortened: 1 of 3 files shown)


dellaert (approved):
Nice!

dellaert:
Nice. *but* I can't merge because of this required status :-)

dellaert:
@p-zach I merged (bypassing rule). I did not see a way to make a required rule specific  to docs, but if you find one let me know.

PR #2024, opened 2025-02-16
Edit required check so that it doesn't hang when only non-code files modified

p-zach:
Fixes issue encountered in #2023 where the check required by the repository rule (Linux CI) does not run when only Markdown files are edited, but it's still "required," so it blocks the PR from being merged because GitHub just waits for the check to run, but it never does.

Works by running a preliminary job to see if relevant files were modified, and only runs the main job if they were. 

The other checks can keep the simpler `paths-ignore` syntax because they are not required under a ruleset.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -1,11 +1,9 @@
 name: Linux CI
 
-on:
-  pull_request:
-    paths-ignore:
-      - '**.md'
-      - '**.ipynb' 
-      - 'myst.yml'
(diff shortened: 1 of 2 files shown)


p-zach:
Whoops, I got the required check wrong. I thought it was Linux CI; it's Python CI. I will fix that.

p-zach:
Now it's right.

Alternatively, the required check could just be turned off (or just require a PR before merging instead of a PR + check), and this PR can be rejected. @dellaert up to you

dellaert (approved):
Cool, I’ll merge and we’ll see…

PR #2028, opened 2025-02-20
Fix small inefficiency in QR path

dellaert:
EliminateQR was doing too much work by setting below-diagonal to zero, as inplaceQR already does that.
Also got rid of zeroBelowDiagonal, as Eigen has it, and was used in only a single place (in inplaceQR). Now uses Eigen.


varunagrawal (approved):
LGTM

PR #2029, opened 2025-02-20
Add `cibuildwheel` workflow

yambati03:
### What is this PR?
This PR introduces a new pipeline that uses `cibuildwheel` to build platform-dependent python wheels. This is the PyPA recommended way for building compliant Python wheels across platforms.

### Why modify `has_ext_modules()` in the `distclass`?
`has_ext_modules()` is called by `setuptools` to indentify if the built wheel is a pure-Python wheel or a platform-dependent wheel. However, `has_ext_modules()` only returns True if the `ext_modules` parameter is defined with the _source_ C/C++ files to build. However, we pre-build our C++ library _prior_ to building the wheel. Thus, we need to force `has_ext_modules()` to return True by overwriting it.

### CMake target `python-stubs` results in _no such file_

Previously, the `python-stubs` target invoked the command `pybind11-stubgen`. However, in Github Actions, it seems like the correct command is `pybind11_stubgen`, causing the below error. This PR changes this in `CMakeLists.txt`, which results in a working build.

**Note:** This workflow is currently only set up to target `Linux x86_64` architectures. However, since it is using `cibuildwheel` it can easily be extended to other architectures (such as Linux ARM). See the below image from the `cibuildwheel` docs.
image

Link to TestPyPi page: https://test.pypi.org/project/gtsam-develop/. Currently targets Linux x86_64 (`python>=3.10`) and Linux aarch64 (`python>=3.10`). MacOS and Windows support will be added in a subsequent PR.

cc @dellaert @akshay-krishnan @truher @ProfFan as interested parties :)


CMakeLists.txt
```diff
@@ -10,10 +10,16 @@ set (GTSAM_VERSION_PATCH 0)
 set (GTSAM_PRERELEASE_VERSION "a0")
 math (EXPR GTSAM_VERSION_NUMERIC "10000 * ${GTSAM_VERSION_MAJOR} + 100 * ${GTSAM_VERSION_MINOR} + ${GTSAM_VERSION_PATCH}")
 
-if ("${GTSAM_PRERELEASE_VERSION}" STREQUAL "")
+if (DEFINED ENV{NIGHTLY})
```
dellaert: I think we need to maybe re-brand to gtsam-develop, as in our slack discussion.

CMakeLists.txt
```diff
@@ -10,10 +10,16 @@ set (GTSAM_VERSION_PATCH 0)
...
+    string(TIMESTAMP NOW "%Y.%m.%d.%H.%M")
```
dellaert: And make this the PR merge commit hash

build_tools/wheels/build_wheels.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+set -x
+
+python -m pip install cibuildwheel
+python -m cibuildwheel build/python --output-dir wheelhouse
```
dellaert: Add missing newline

build_tools/wheels/cibw_before_all.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
dellaert, line 1: Add some file docs (to all files) ? What does this file do?

dellaert (commented):
This is incredible! I left some comments, but main request is to rebrand to gtsam-develop and have merge commit hashes rather than dates. The base version of course should be in there as well. We are “pre-4.3”, so I guess GTSAM-develop.4.3.HASHHHH or something.

maybe find some docs on using commit hashes to see if it is standard practice somewhere, so in the docs (add a README file?) we can refer to that.

ProfFan:
Ah, finally something from PyPa that does all the stuff we had to do in `gtsam-manylinux-build`! Please be sure to generate the `.pyi` type annotation files (refer to `python-stubs` target in CMake) as well and put that inside the wheels.

ProfFan:
Also please check the built macOS wheels locally if they work. `delocate` sometimes would leave some `dylib`s linking against system dependencies, not sure if they are good now.

yambati03:
@dellaert It looks like having the Git commit hash in the version string may not be possible. See [the PyPA docs](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers. ).

It seems that version specifiers must comply with the format `[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`. So, something like `4.3a0` complies with this. If we wanted to add the commit hash to `gtsam-develop` builds, it would have to look like `4.3a0+xxxxxxx`. This is because an arbitrary string, such as the commit hash, _does not_ comply with the above format. Thus, it can only be included as a "local identifier" (which must be delineated with a +). However, local identifiers are explicitly disallowed on PyPI. See [this PR reply](https://github.com/pypi/legacy/issues/731#issuecomment-345461596), and the quote from PyPA below.

> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server, but MAY be used to identify private builds created directly from the project source.

What do you think about the versioning scheme `{major}.{minor}{prerelease}.dev{datetime}` for use on PyPI instead?

dellaert:
> @dellaert It looks like having the Git commit hash in the version string may not be possible. See [the PyPA docs](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers.).
> 
> It seems that version specifiers must comply with the format `[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`. So, something like `4.3a0` complies with this. If we wanted to add the commit hash to `gtsam-develop` builds, it would have to look like `4.3a0+xxxxxxx`. This is because an arbitrary string, such as the commit hash, _does not_ comply with the above format. Thus, it can only be included as a "local identifier" (which must be delineated with a +). However, local identifiers are explicitly disallowed on PyPI. See [this PR reply](https://github.com/pypi/legacy/issues/731#issuecomment-345461596), and the quote from PyPA below.
> 
> > Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server, but MAY be used to identify private builds created directly from the project source.
> 
> What do you think about the versioning scheme `{major}.{minor}{prerelease}.dev{datetime}` for use on PyPI instead?

That works. Maybe make sure then that you always use 2 digits for the month. E.g.,
`25.02.21`

dellaert:
Cool! What’s the status on this PR?
note it does not have to be perfect before we merge - we can add platforms like ARM/Mac/Windows in subsequent PRs.
I do think we need to support all current Python versions starting with 3.10.

yambati03:
Gotcha! At the moment, the Linux x86_64 (for all python versions) and ARM targets work fine. I was trying to add support for MacOS, but that can be added in a subsequent PR. I was also trying to address @ProfFan's comment about building `python-stubs` and including the resulting `.pyi` files in the wheel. Do you think this is something that needs to be done in this PR?

varunagrawal:
I say leave the python-stubs for later. That feature has caused a lot of headaches in the past and required some hacky Cmake (which I am not a big fan of), so just getting a nightly build going is a massive step forward.

varunagrawal:
Also for later consideration: I believe we should have a separate repo for this nightly build which can be run every day at the same time via [a cron job](https://www.theanshuman.dev/articles/free-cron-jobs-with-github-actions-31d6) and reduce build times for development PRs.

Moreover, if in the future something changes and this CI doesn't work anymore (because Actions changes something), we won't require people to wait on us to fix it.

python/CMakeLists.txt
```diff
@@ -284,9 +284,9 @@ endif()
 add_custom_target(
         python-stubs
         COMMAND
-        ${CMAKE_COMMAND} -E env
-        "PYTHONPATH=${GTSAM_PYTHON_BUILD_DIRECTORY}/$ENV{PYTHONPATH}"
-        pybind11-stubgen -o . --enum-class-locations \"KernelFunctionType|NoiseFormat:gtsam.gtsam\" --enum-class-locations \"OrderingType:gtsam.gtsam.Ordering\" --numpy-array-use-type-var --ignore-all-errors gtsam
+          ${CMAKE_COMMAND} -E env
+          "PYTHONPATH=${GTSAM_PYTHON_BUILD_DIRECTORY}/$ENV{PYTHONPATH}"
+        ${PYTHON_EXECUTABLE} -m pybind11_stubgen -o . --enum-class-locations \"KernelFunctionType|NoiseFormat:gtsam.gtsam\" --enum-class-locations \"OrderingType:gtsam.gtsam.Ordering\" --numpy-array-use-type-var --ignore-all-errors gtsam
```
dellaert, line 289: Comment on what you found, and mention in PR comments that stubgen already works, and double-check with Varun.

dellaert (commented):
Please add *many* comments for someone who reads this in 3 years

dellaert (approved):
Nice !!!

.github/workflows/build-cibw.yml
```diff
@@ -0,0 +1,175 @@
...
+      - name: Set Develop Flag
+        run: |
+          echo "DEVELOP=1" >> $GITHUB_ENV
+          echo "TIMESTAMP=${{ needs.get_system_time.outputs.timestamp }}" >> $GITHUB_ENV
+
+      - name: Install Dependencies
+        run: |
+          python3 -m pip install -r python/dev_requirements.txt
+          if [ "$RUNNER_OS" == "Linux" ]; then
+            sudo apt-get install -y wget libicu-dev python3-pip python3-setuptools libboost-all-dev ninja-build
```
varunagrawal, line 116: This is fishy. You're installing `pip` after running `pip install`?
Probably should be a step before `Install Dependencies`.

varunagrawal:
I believe you're getting the error message when using `pybind11-stubgen` because the VM which Actions uses doesn't set up the paths correctly. This is normal for security reasons (can't have arbitrary executables being run), but essentially `bash` doesn't know where to find the `pybind11-stubgen` executable (which should be under `<path to python installation>/<version>/bin`).

The fix is valid, since importing the module automatically calls `main()` on `pybind11-stubgen` but the reason provided for it is not quite correct. `pybind11-stubgen` (with the hyphen) is indeed the correct command to call, however @yambati03's fix generalizes its application to all kinds of scenarios.

dellaert:
@yambati03 some of Varun’s comments came after I merged. Please address them in the macOS PR?

PR #2030, opened 2025-02-20
Add Rot3 documentation

p-zach:


dellaert (approved):
We discussed some small improvements in the meeting, assuming you took notes and will apply :-)

PR #2032, opened 2025-02-21
Cleanup eigen svd clang warning. Update svd options usage.

jmackay2:
This removes the GCC diagnostic handling for svd by checking the svd info. This was made available in Eigen 3.4, which this repository uses. 

This also updates a few of the JacobiSVD to use the templated options. This is the preferred method instead of passing it in as an argument, which will be deprecated in the future.


dellaert:
Cool! but still some Ci issue :-)

dellaert:
BTW, this might be an issue: 
- what version of Eigen do we bundle? Is it 3.4??
- what “system Eigen” do we support? I think we should look at earliest supported a Ubuntu and see what it comes with.
So, this PR might introduce incompatibilities.

PR #2033, opened 2025-02-25
fix maybe unitialized warning

jmackay2:
This fixes the maybe uninitialized warning by checking the output of svd.

### DIFF
--- a/gtsam/geometry/FundamentalMatrix.cpp
+++ b/gtsam/geometry/FundamentalMatrix.cpp
@@ -9,10 +9,6 @@
 #include <gtsam/geometry/FundamentalMatrix.h>
 #include <gtsam/geometry/Point2.h>
 
-#if defined(__GNUC__) && !defined(__clang__)
-#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#endif
-
 namespace gtsam {
 
 //*************************************************************************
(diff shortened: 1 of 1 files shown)


ProfFan (approved):

ProfFan:
Hmm, actually I wonder if this would actually fix the problem? Is GCC that smart?

jmackay2:
> Hmm, actually I wonder if this would actually fix the problem? Is GCC that smart?

I was able to test on two different computers with Ubuntu 24.04. I get the warning without the GCC preamble. With the code addition, it builds successfully with no warning.

berndpfrommer:
This commit seems to break the build of our nightly packages:
```
/<<PKGBUILDDIR>>/gtsam/geometry/FundamentalMatrix.cpp: In constructor ‘gtsam::FundamentalMatrix::FundamentalMatrix(const Matrix3&)’:
/<<PKGBUILDDIR>>/gtsam/geometry/FundamentalMatrix.cpp:39:11: error: ‘class Eigen::JacobiSVD<Eigen::Matrix<double, 3, 3>, 2>’ has no member named ‘info’
   39 |   if (svd.info() != Eigen::ComputationInfo::Success) {
      |           ^~~~
```

berndpfrommer:
The broken build is on an arm64 device running Ubuntu 20.04 (focal fossey). Not sure why I have not seen broken build messages for amd64, but that could still come.
It looks like the Eigen library does not provide the info() method for this case.  The version of Eigen that is running on Ubuntu 20.04 is 3.3.7-2.
It turns out the info() method was added to JacobiSVD "only" [3 years ago](https://gitlab.com/libeigen/eigen/-/commit/5bbc9cea93ef29cee2b8ffb2084d4ebca32600ba).

jmackay2:
> The broken build is on an arm64 device running Ubuntu 20.04 (focal fossey). Not sure why I have not seen broken build messages for amd64, but that could still come. It looks like the Eigen library does not provide the info() method for this case. The version of Eigen that is running on Ubuntu 20.04 is 3.3.7-2. It turns out the info() method was added to JacobiSVD "only" [3 years ago](https://gitlab.com/libeigen/eigen/-/commit/5bbc9cea93ef29cee2b8ffb2084d4ebca32600ba).

Yes, this can happen if you chose not to use the Eigen version that comes with gtsam. I'll put in a PR to fix this issue soon.

PR #2039, opened 2025-03-04
Bugfix Bearing/Range Pose3toPose3 Jacobians

DanMcGann:
Previously the Jacobians for bearing between two Pose3 objects were computed incorrectly as it neglected the Jacobian of a pose's translation wrt its tangent vector. 

The Jacobian for range between two Pose3 objects was computed correctly but in a non-canonical way. (Canonical = using OptionalJacobians for all chain rule computation) 

This commit fixes the bearing Jacobian, and provides a canonical implementation for both Pose3-Pose3 bearing and range Jacobians.

This was discovered while doing testing for multi-robot cases (where robots directly observe each other via bearing-range measurements) and had a significant and detrimental affect on results even causing GaussNewton solvers to completely diverge.

Also based on git this has been hiding for 8 years? So hopefully has not been causing issues for anyone!

Finally, I will note that the deleted comment in the test is just plain wrong since jacobians are of course compute WRT the tangent elements.

### DIFF
--- a/gtsam/geometry/Pose3.cpp
+++ b/gtsam/geometry/Pose3.cpp
@@ -442,9 +442,11 @@ double Pose3::range(const Point3& point, OptionalJacobian<1, 6> Hself,
 /* ************************************************************************* */
 double Pose3::range(const Pose3& pose, OptionalJacobian<1, 6> Hself,
                     OptionalJacobian<1, 6> Hpose) const {
+  Matrix36 D_point_pose;
   Matrix13 D_local_point;
-  double r = range(pose.translation(), Hself, Hpose ? &D_local_point : 0);
-  if (Hpose) *Hpose << Matrix13::Zero(), D_local_point * pose.rotation().matrix();
+  Point3 point = pose.translation(Hpose ? &D_point_pose : 0);
+  double r = range(point, Hself, Hpose ? &D_local_point : 0);
+  if (Hpose) *Hpose = D_local_point * D_point_pose;
   return r;
 }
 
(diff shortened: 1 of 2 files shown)


varunagrawal:
I don't know if the comment is wrong per se, or perhaps I am missing something. The Jacobians are computed wrt the tangent elements of the second pose (hence `numericalDerivative22`), but since the second pose is a point, it is $\in \mathbb{R}^3$ (which is its own tangent elements). There is no rotation aspect, which makes the test incomplete IMO but the comment explicitly states that.

I would really like to see an additional test that does a full $SE(3)$ to $SE(3)$ jacobian computation.

DanMcGann:
You are completely right! 
I actually missed the fact that the test was actually testing the `(Point3)` signature function. 
It has now been update to test the `(Pose3)` signature function and full jacobians.

Sorry for the miss, TBH I did not run the test locally since it takes a very long time to build the tests on my local machine.
BUT I did test the new version of the test locally. So hopefully everything is squared away.

varunagrawal (approved):
LGTM. Thank you @DanMcGann!

varunagrawal:
> You are completely right! I actually missed the fact that the test was actually testing the `(Point3)` signature function. It has now been update to test the `(Pose3)` signature function and full jacobians.
> 
> Sorry for the miss, TBH I did not run the test locally since it takes a very long time to build the tests on my local machine. BUT I did test the new version of the test locally. So hopefully everything is squared away.

No worries. Thank you for being prompt on this and the nice fix. :)

ProfFan:
> Also based on git this has been hiding for 8 years?

Thank you for the fix lol, I guess it's just peeps really only use range factors (<me)

PR #2040, opened 2025-03-05
Fix Pruning Bug

varunagrawal:
There is a bug in the `computeThreshold` method where it returns the smallest value of the discrete probabilities even if the max number of assignments allowed is greater than the number of discrete probabilities. This was leading to incorrect dead mode removal.

I added a test that fails without the fix and passes with.

This gives us slightly different results.

### Previous 220 steps
city10000_results_220

### New 220 steps
city10000_results_220_new

The final results are a little bit better!

### Previous 6500 steps
city10000_results_6500

### New 6500 steps
city10000_results_6500_new

### DIFF
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -501,6 +501,10 @@ namespace gtsam {
     };
     this->visitWith(op);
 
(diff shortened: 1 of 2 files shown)


dellaert (approved):
LGTM!

PR #2043, opened 2025-03-07
Deprecate `GTSAM_SLOW_BUT_CORRECT_EXPMAP`

varunagrawal:
Set `GTSAM_SLOW_BUT_CORRECT_EXPMAP` to ON by default and print deprecation warning.

This is based on findings by @dellaert and I on the accuracy of the exponential map of `Pose2` for hybrid state estimation with the City10000 dataset.

With the flag off, we get
city10000_results_110
where the trajectory is elongated.

With the flag on, we get a better trajectory.
city10000_results_110_2

Similar results can be seen for longer trajectories (e.g. 220 timesteps).

#### Flag OFF
city10000_results_220_original

#### Flag ON
city10000_results_220


varunagrawal:
Some of the failing tests are because of quite different results (e.g. $\theta = -\pi$ instead of $\pi$) so going to spend time later tonight debugging this some more.

cmake/HandleGeneralOptions.cmake
```diff
@@ -47,7 +47,9 @@ option(GTSAM_ALLOW_DEPRECATED_SINCE_V43     "Allow use of methods/functions depr
 option(GTSAM_SUPPORT_NESTED_DISSECTION      "Support Metis-based nested dissection" ON)
 option(GTSAM_TANGENT_PREINTEGRATION         "Use new ImuFactor with integration on tangent space" ON)
 option(GTSAM_SLOW_BUT_CORRECT_BETWEENFACTOR "Use the slower but correct version of BetweenFactor" OFF)
-option(GTSAM_SLOW_BUT_CORRECT_EXPMAP        "Use slower but correct expmap for Pose2"  OFF)
+option(GTSAM_SLOW_BUT_CORRECT_EXPMAP        "Use slower but correct expmap for Pose2"  ON)
+
+message(WARNING "GTSAM_SLOW_BUT_CORRECT_EXPMAP has been deprecated and will be removed in a future release.")
```
dellaert: Should this message be displayed only when set to OFF?

varunagrawal:
@dellaert the TSAM factor tests are failing with wildly different Jacobians (e.g. values of 7 in expected and 3 in actual).

It might be something with how the factor Jacobians are being computed. Any thoughts since I am not super familiar with this at the moment?

dellaert:
I'll take a look, please review Hat&Vee PR?

dellaert:
I fixed all tests (on my MacBook) but had to comment out a test in GnC, @lucacarlone

varunagrawal:
Okay nice, I had similar fixes locally but wanted to double check with you.

dellaert (changes requested):
I think we need to add some computer time benchmarks before merging this

cmake/HandleGeneralOptions.cmake
varunagrawal: Good point.

dellaert:
The performance concern raised here has since been addressed by #2661, which introduced the relevant microbenchmark and `w10000` end-to-end benchmark for the correct Lie-group Jacobian path. That PR is now merged into `develop`. This PR builds on that work by changing the `Pose2` chart default; local validation shows no end-to-end regression, so we do not duplicate the benchmark infrastructure here.

dellaert (approved):
Speed difference is enormous! (See #2661)

PR #2044, opened 2025-03-07
Move scripts for building wheels to CI directory

varunagrawal:
Since we use CMake for building GTSAM, most users will have a build directory, and this may cause confusion with the newly added `build_tools` directory.

This PR moves the `cibuildwheel` scripts to `.github/scripts/python_wheels` so that everything is nicely packaged under a single directory tree (which is `.github`).

@yambati03 you may have had some good design choices when creating `build_tools`, which I would really like to understand. 🙂


yambati03 (approved):
This looks good to me! I originally had them in a top-level `build_tools` directory since it seemed like the common practice across other large open source projects such as `scikit-learn` and `numpy`. However, I can see this being confusing and moving them to the `.github/scripts` folder makes it clear that these are associated with the Github Actions workflow. Thanks!

varunagrawal:
Got it! That actually makes a lot of sense if we were a python oriented codebase. Thanks for the information. :)

PR #2045, opened 2025-03-08
Hat and Vee

dellaert:
- made MatrixLieGroup concept that enforces Hat and Vee static functions
- deprecated wedge
- added a bunch more unit tests
- discovered Similarity2 Expmap/Logmap are incorrect - will make issue

### DIFF
--- a/gtsam/base/Lie.h
+++ b/gtsam/base/Lie.h
@@ -171,21 +171,22 @@ namespace internal {
 /// Assumes existence of: identity, dimension, localCoordinates, retract,
 /// and additionally Logmap, Expmap, compose, between, and inverse
 template<class Class>
-struct LieGroupTraits: GetDimensionImpl<Class, Class::dimension> {
-  typedef lie_group_tag structure_category;
+struct LieGroupTraits: public GetDimensionImpl<Class, Class::dimension> {
+  using structure_category = lie_group_tag;
 
   /// @name Group
   /// @{
-  typedef multiplicative_group_tag group_flavor;
+  using group_flavor = multiplicative_group_tag;
   static Class Identity() { return Class::Identity();}
   /// @}
 
   /// @name Manifold
   /// @{
-  typedef Class ManifoldType;
+  using ManifoldType = Class;
   inline constexpr static auto dimension = Class::dimension;
-  typedef Eigen::Matrix<double, dimension, 1> TangentVector;
-  typedef OptionalJacobian<dimension, dimension> ChartJacobian;
+  using LieAlgebra = Class::LieAlgebra;
+  using TangentVector = Eigen::Matrix<double, dimension, 1>;
+  using ChartJacobian = OptionalJacobian<dimension, dimension>;
 
   static TangentVector Local(const Class& origin, const Class& other,
       ChartJacobian Horigin = {}, ChartJacobian Hother = {}) {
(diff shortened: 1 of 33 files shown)


dellaert:
PS @varunagrawal I had to disable a TableFactor test as it was taking a loooong time.

gtsam/base/Lie.h
```diff
@@ -225,10 +225,28 @@ struct LieGroupTraits: GetDimensionImpl<Class, Class::dimension> {
   /// @}
 };
 
+
 /// Both LieGroupTraits and Testable
 template<class Class> struct LieGroup: LieGroupTraits<Class>, Testable<Class> {};
 
-} // \ namepsace internal
+/// Adds LieAlgebra, Hat, and Vie to LieGroupTraits
```
ProfFan, line 232: Vee

gtsam/base/Lie.h
```diff
@@ -301,22 +338,21 @@ T BCH(const T& X, const T& Y) {
...
+ * class T needs a constructor from Matrix.
  * @param x exponential coordinates, vector of size n
  * @ return a T
  */
 template <class T>
-T expm(const Vector& x, int K=7) {
-  Matrix xhat = wedge<T>(x);
-  return T(expm(xhat,K));
+T expm(const Vector& x, int K = 7) {
+  auto xhat = T::Hat(x);
```
ProfFan, line 354: `auto` with Eigen types is probably fine for this usage, but it looks scary

varunagrawal, line 354: I prefer statically defined types for code. I can update this in another PR.

gtsam/base/VectorSpace.h
```diff
@@ -51,6 +51,8 @@ struct VectorSpaceImpl {
   /// @name Lie Group
   /// @{
 
+  typedef Eigen::Matrix<double, N, 1> LieAlgebra;
```
varunagrawal, line 54: `using` instead of `typedef` to be consistent?

gtsam/base/VectorSpace.h
```diff
@@ -411,6 +420,8 @@ struct DynamicTraits {
 
   /// @name Lie Group
   /// @{
+  using LieAlgebra = Dynamic;
+    
   static TangentVector Logmap(const Dynamic& m, ChartJacobian H = {}) {
```
varunagrawal, line 425: Maybe replace `Dynamic` with `LieAlgebra`?

gtsam/geometry/Quaternion.h
```diff
@@ -138,6 +138,16 @@ struct traits<QUATERNION_TYPE> {
     return omega;
   }
 
+  using LieAlgebra = Matrix3;
```
varunagrawal, line 141: Do we need this if it isn't being used here?

gtsam/geometry/Similarity3.h
```diff
@@ -164,16 +166,18 @@ class GTSAM_EXPORT Similarity3 : public LieGroup<Similarity3, 7> {
...
+
   /**
-   * wedge for Similarity3:
+   * Hat for Similarity3:
    * @param xi 7-dim twist (w,u,lambda) where
    * @return 4*4 element of Lie algebra that can be exponentiated
-   * TODO(frank): rename to Hat, make part of traits
    */
-  static Matrix4 wedge(const Vector7& xi);
+  static Matrix4 Hat(const Vector7& xi);
```
varunagrawal, line 177: No Vee for Sim3?

varunagrawal (approved):
Awesome. Couple of minor comments.

gtsam/base/Lie.h
dellaert: It’s already fixed. In this case bTW Hat is defined differently for every T so auto makes sense, except it will need to be cast anyway to a Matrix for expm

gtsam/base/VectorSpace.h
dellaert: Good catch. I am not gonna launch another entire CI for this one change but maybe it can be cleaned up for all of group/manifold/lie/VectorSpace in a “using” PR.

gtsam/base/VectorSpace.h
dellaert: No, that does not make sense I think.

gtsam/base/VectorSpace.h
varunagrawal: You can merge this and I can update it in my new PR.

PR #2047, opened 2025-03-09
Use `GTSAM_CONCEPT_USAGE` in lieu of `BOOST_CONCEPT_USAGE`

varunagrawal:
In an effort to have less Boost influence on the code base, this PR updates `BOOST_CONCEPT_USAGE` with `GTSAM_CONCEPT_USAGE`.

### DIFF
--- a/gtsam/base/Group.h
+++ b/gtsam/base/Group.h
@@ -45,7 +45,7 @@ public:
   typedef typename traits<G>::group_flavor flavor_tag;
   //typedef typename traits<G>::identity::value_type identity_value_type;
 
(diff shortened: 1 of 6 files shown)


dellaert:
This still uses the boost thing when boost features are enabled, right? It should, as that is actually more sophisticated and we’d want that safety blanket until c++ provides these things.

varunagrawal:
Yes it does! I aliased the macro in concepts.h and ran tests locally before pushing. :)

dellaert (approved):

PR #2051, opened 2025-03-10
Fix Sim(2) Expmap and Logmap

### LINKED ISSUE #2046: Similarity2 Expmap/Logmap are incorrect
## Expmap and LogMap in Similarity2 are incorrect

They don't actually implement the exponential map and its inverse.

## Steps to reproduce

enable this test:

// Checks correct exponential map (Expmap) with brute force matrix exponential
TEST_DISABLED(Similarity2, BruteForceExpmap) {
  const Vector4 xi(0.1, 0.2, 0.3, 0.4);
  EXPECT(assert_equal(Similarity2::Expmap(xi), expm<Similarity2>(xi), 1e-4));
}

## Expected behavior

Should match matrix exponential, as in Similarity3

## Additional information

Similarity3 should probably be modernized as well to use the new SO3 functor implementations, as in Pose3. Then we can work by analogy.

varunagrawal:
Using the reference [here](https://ethaneade.com/lie_groups.pdf), I fixed the Expmap and Logmap implementations for `Similarity2`.

Fixes #2046.

### DIFF
--- a/gtsam/geometry/Similarity2.cpp
+++ b/gtsam/geometry/Similarity2.cpp
@@ -194,6 +194,33 @@ Similarity2 Similarity2::Align(const Pose2Pairs& abPosePairs) {
   return internal::AlignGivenR(abPointPairs, aRb_estimate);
 }
 
+Matrix2 Similarity2::GetV(double theta, double lambda) {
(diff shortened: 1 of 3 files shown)


dellaert (approved):
Awesomesauce!

Issue #2050, opened 2025-03-10
Apt package build fails on Ubuntu 20.04, seems related to "Hat"


berndpfrommer:
Last night's build failed with the below error (and many more following).

I have trouble figuring out what the root cause of it is. It has something to do with Eigen, something to do with "Hat". Most likely the Eigen version (3.3.7.2) of Ubuntu 20.04 does not support some feature that is used with some of the most recent commits. Unfortunately there have been a number of commits the last few days  and I'm not sure which one is the culprit. 
[Here](https://launchpadlibrarian.net/781133363/buildlog_ubuntu-focal-amd64.gtsam_4.1.0-1ubuntu2~2780.gbpfb9999_BUILDING.txt.gz) a link to the full build log.

```
[...]

PR #2053, opened 2025-03-11
Fix Pose3 Wedge operator

ProfFan:
#2050 

Interesting this is not caught by CI, did CI not test the option?


dellaert (approved):
Hmm, CI should have caught this.

PR #2054, opened 2025-03-11
Implement Equivariant Filter Bias example in Python

jenniferoum:
An implementation of Alessandro Fornasier's ABC-EqF in GTSAM.

- EqF.py is a module containing general function and class definitions for the equivariant filter
- EqF.ipynb is a notebook containing an example implementation of the EqF code using sample data
- EqFdata.csv is sample measurement data provided by Fornasier in his ABC-EqF


dellaert:
Awesome. Before I review, do you want to use the new “Hat” operators I added yesterday? (instead of wedge)

dellaert:
https://github.com/borglab/gtsam/blob/33d2262e07578e51c5b991d960c9763508afff04/gtsam/geometry/geometry.i#L381

dellaert:
Also, csv file should go in examples/Data. You can then call `FindExampleDataFile` (or something like that, declared in datasets.h) - which will work across platforms.

jenniferoum:
Yeah, I will work on adding in the new hat and vee operators.

python/gtsam/examples/EqF.py
```diff
@@ -0,0 +1,887 @@
...
+    :return: numpy array representing a block diagonal matrix composed of n-times the blocks A
+    """
+
+    res = None
+    for _ in range(n):
+        res = blockDiag(res, A)
+    return res
+
+
+def Rot3LeftJacobian(arr: np.ndarray) -> np.ndarray:
```
dellaert: duplicate?

python/gtsam/examples/EqF.py
```diff
@@ -0,0 +1,887 @@
...
+import math
+from dataclasses import dataclass
+import progressbar
+import pandas as pd
+import matplotlib.pyplot as plt
+from typing import List
+
+coordinate = "EXPONENTIAL"
+
+def Rot3LeftJacobian(arr: np.ndarray) -> np.ndarray:
```
dellaert: Maybe add as staticmethod in class G

python/gtsam/examples/EqF.py
```diff
@@ -0,0 +1,887 @@
...
+
+        # Append to data_list
+        d = Data(State(R, b, S), n_cal, Input(w, Sigma_wb), meas, n_meas, t, dt)
+        
+        data_list.append(d)
+
+    return data_list
+
+
+def sim(filter_args, data):
```
dellaert: I propose you leave this code in the notebook (and the corresponding plotting imports)

python/gtsam/examples/EqF.py
```diff
@@ -0,0 +1,887 @@
...
+        """Return a identity state  with n calibration states
+
+        :param n: number of elements in list B associated with calibration states
+        :return: The identity element of the State
+        """
+
+        return State(Rot3.Identity(),
+                     np.zeros(3),
+                     [Rot3.Identity() for _ in range(n)])
+class Input:
```
dellaert, line 105: Use a formatter like black to format code.

python/gtsam/examples/EqF.py
```diff
@@ -0,0 +1,887 @@
...
+        self.Sigma = Sigma
+        self.cal_idx = i
+
+@dataclass
+class Data:
+    """Define ground-truth, input and output data"""
+
+    # Ground-truth state
+    xi: State
+    n_cal: int
```
dellaert, line 545: Seems redundant. You can get it from the state

python/gtsam/examples/EqF.py
```diff
@@ -0,0 +1,887 @@
...
+
+    # Input measurements
+    u: Input
+
+    # Output measurements as a list of Measurement
+    y: list
+    n_meas: int
```
dellaert, line 552: also redundant: len(y)

dellaert (approved):
This is awesome! A few minor comments. I'll approve but will wait to merge until they have been addressed. (or tell me why you think it's ok to leave as is)

python/gtsam/examples/EqF.py
jenniferoum: I prefer to leave n_meas and n_cal alone since calling len a bunch of times isn't necessary when the value is never going to change.

PR #2056, opened 2025-03-12
Hybrid Updates

varunagrawal:
Some improvements to the hybrid code to make for easier debugging.

1. Print the full factors within `HybridNonlinearFactor`.
2. Make the `involved` function functional, rather than use a capture since `involvedKeys` is later updated.

PS @dellaert can we also get those other PRs merged soon?


dellaert (approved):

dellaert:
I approved 2 PRs. I prefer if authors with write permission merge their own PRs (and delete the branch).
The exmap PR needs some benchmark numbers in the PR before we can merge. I’ll do that when I get to it.

Issue #2058, opened 2025-03-15
Unable to build GTSAM without Boost after latest update


Smithangshu:
## Description

With the latest version whenever I am trying to build the library without `BOOST` in `Windows` with `CMake`, it fails. But in previous version. I could build it without any issue.

I am getting following error

`3>D:\Gits\gtsam\gtsam\base\Lie.h(316,1): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
3>(compiling source file '../../../gtsam/gtsam/precompiled_header.cpp')
3>D:\Gits\gtsam\gtsam\base\Lie.h(316,1):`

## Steps to reproduce

1. Just disable all the Boost related flags in CMake GUI.
2. Try to compile the build with VS 2022.

## Expected behavior

It should work without any issue.

## Environment
Windows 11.
VS 2022.
CMake GUI.

ProfFan:
Could be #2047

dellaert:
@Smithangshu could you try with #2047 undone?

Smithangshu:
@ProfFan yes, it is the cause.

Smithangshu:
@dellaert if I comment out this section from Lie.h. And build it, then it works like charm.

![Image](https://github.com/user-attachments/assets/c8302302-af16-4214-b734-777c2b42e88c)

dellaert:
Ah, that’s my bad! It should now use the new macro. It was my Hat/Vee that probably landed concurrently with @varunagrawal ‘s PR

ProfFan:
Should be solved with #2060

PR #2060, opened 2025-03-16
Fix incorrect pybind version in `wrap`

### LINKED ISSUE #2059: Python extension to GTSAM seems broken on develop
Extensions to base GTSAM with Pybind11 seems broken. I don't have time to debug for now.

ProfFan:
`pybind11` master (v3) was inadvertently pulled instead of the release version (v2). This PR reverts that change.

Should resolve #2059 where 3rd party extensions to GTSAM failed to work as the ABI changed between major pybind versions.

Also piggybacking the fix for #2058 as it is preventing CI execution


varunagrawal (approved):

dellaert:
Thanks both!

PR #2062, opened 2025-03-17
Nonlinear Hybrid Smoother

varunagrawal:
Updated the `HybridSmoother` to accept nonlinear hybrid factor graphs as well as perform relinearization.

To confirm, I re-ran the `HybridCity10000.py` script for 6500 iterations and got the following, which is the same as the previous version of the script.

city10000_results_6500


dellaert (approved):
Cool!

dellaert:
Merge when convenient

PR #2064, opened 2025-03-20
Fix the build for example Hybrid City 10000

jmackay2:
# Issue:
When building, I get the following error:
```
gtsam/examples/Hybrid_City10000.cpp:113:35: error: cannot convert ‘size_t’ {aka ‘long unsigned int’} to ‘const gtsam::Values&’
  113 |     smoother_.update(*linearized, maxNrHypotheses);
      |                                   ^~~~~~~~~~~~~~~
      |                                   |
      |                                   size_t {aka long unsigned int}

```

Instead of using the size_t, this changes it to use the Values called `initial_`.


varunagrawal (approved):
This actually needs some more work but that should be my responsibility. Thanks!

PR #2065, opened 2025-03-20
CI using Docker

### LINKED ISSUE #2037: Use Docker for CI
## Feature

Have a custom docker image uploaded to the [borglab dockerhub account](https://hub.docker.com/r/borglab/gtsam) which has all the dependencies pre-installed.
We can then use these images in the CI and simplify+speedup the overall process.

## Motivation

Currently, the CI pulls various dependencies and installs them. Using a docker image which already has these dependencies installed can provide a nice speedup of the CI.

## Pitch

This is not very difficult to do but requires some time and familiarity with respect to Docker, Github Actions, and how these two work together.

The impact of this feature would be quite large, since it would lead to better development turnaround time.

suveergarg:
This is a starting step that does CI for Linux with pulling docker images.

I tested building the docker images and running the builds and test locally and it worked. However since I do not have permissions to push to the borglab docker account, I need some help in pushing images here to docker hub with the `build_and_push.sh` script that has been committed.  I can then test the CI steps.


suveergarg:
This MR is ready for a review. It mainly switches the Linux based workflow to docker.

You need to setup a `DOCKER_USERNAME` and `DOCKER_TOKEN` secret in the gtsam project for these changes to work properly. You also need to create a repository called `gtsam-ci` on your dockerhub.

Following are successful runs from my fork:
1. Special Cases : https://github.com/suveergarg/gtsam/actions/runs/17747292139
2. Linux : https://github.com/suveergarg/gtsam/actions/runs/17748317518 (Speed up of 2x)


I cannot change the workflows for Windows and mac-os with docker because:

1. macos does not play nice with docker. Most images are community built and it is hard to control the xcode version etc installed in these images. Using Github actions macOS runners is probably the best practice here.
2. I do not have a windows machine - so cannot build a windows container : https://stackoverflow.com/questions/42158596/can-windows-containers-be-hosted-on-linux

dellaert:
cool. Running CI, then I'll take a more detailed look.

suveergarg:
Remember to setup a DOCKER_USERNAME and DOCKER_TOKEN secret in the gtsam project for these changes to work properly. You also need to create a repository called gtsam-ci on your dockerhub.

dellaert:
@varunagrawal promised to look into this over the weekend.

varunagrawal:
First impressions: AMAZING STUFF!!

Thanks @suveergarg. As promised I'll take a more detailed look by this weekend. :)

suveergarg:
@varunagrawal, did you get a chance to look at this? Let me know if you need help with any questions.

varunagrawal:
Hi @suveergarg. I am so sorry, I had conference planning to take care of this past week. I'll start the review now.

.github/workflows/build-linux.yml
```diff
@@ -74,50 +68,28 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Install Dependencies
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
```
varunagrawal: Can we use `login-action@latest` here so we are always using the latest version in case there are improvements or changes?

.github/workflows/build-linux.yml
```diff
@@ -74,50 +68,28 @@ jobs:
...
-          elif [ "${{ matrix.compiler }}" = "clang" ]; then
-            export LDFLAGS="-fuse-ld=mold"
-          else
-            export LDFLAGS=""
-          fi
-          export CMAKE_EXE_LINKER_FLAGS="$LDFLAGS"
-          export CMAKE_SHARED_LINKER_FLAGS="$LDFLAGS"
-          bash .github/scripts/unix.sh -t
+        run: docker run -v $GITHUB_WORKSPACE:/gtsam gargsuveer/gtsam-ci:${{ matrix.os }}-${{ matrix.compiler }}-${{ matrix.version }} \
+          -e GTSAM_USE_BOOST_FEATURES=${{ env.GTSAM_USE_BOOST_FEATURES || 'OFF' }} \
```
varunagrawal: Can we put this in a file that is then loaded into the docker container? `docker run` has the arg `--env-file` which has the description: `Read in a file of environment variables`. This would make this cleaner and we can put the env file somewhere within the `.github` directory.

.github/workflows/build-special.yml
```diff
@@ -182,9 +166,17 @@ jobs:
...
       - name: Build & Test
-        env:
-          LDFLAGS: "-fuse-ld=mold"
         run: |
-          export CMAKE_EXE_LINKER_FLAGS="$LDFLAGS"
-          export CMAKE_SHARED_LINKER_FLAGS="$LDFLAGS"
-          bash .github/scripts/unix.sh -t
+          docker run -v $GITHUB_WORKSPACE:/gtsam gargsuveer/gtsam-ci:${{ matrix.os }}-${{ matrix.compiler }}-${{ matrix.version }} \
+            -e CTEST_OUTPUT_ON_FAILURE=ON \
+            -e CTEST_PARALLEL_LEVEL=2 \
```
varunagrawal: Same env file comments here.

containers/gtsam-ci/build_and_push.sh
```diff
@@ -0,0 +1,42 @@
...
+DOCKER_REPOSITORY="gtsam-ci"
+
+# Define arrays for Ubuntu versions and compilers
+declare -a build_configs=( 
+  "ubuntu 22.04 gcc 9"
+  "ubuntu 22.04 clang 11"
+  "ubuntu 22.04 clang 14"
+  "ubuntu 24.04 gcc 14"
+  "ubuntu 24.04 clang 16"
+)
```
varunagrawal: This is cool! I didn't know you could define configs this way and yield a list!

containers/gtsam-ci/build_and_push.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
varunagrawal: I am wondering if it makes sense to put this script in the https://github.com/borglab/docker-images repository since it only needs to run every few times when we update the images? Thoughts?

My big picture idea about this is that the containers here don't depend on GTSAM, rather GTSAM depends on them in the CI, and hence it should be an "external" dependency.

containers/gtsam-ci/Dockerfile.ubuntu-clang
```diff
@@ -0,0 +1,51 @@
+ARG UBUNTU_VERSION=22.04
+FROM docker.io/ubuntu:${UBUNTU_VERSION}
+
+ARG COMPILER_VERSION
+
+# Install dependencies
+RUN sed -i 's|http://archive.ubuntu.com/ubuntu/|http://mirror.math.princeton.edu/pub/ubuntu/|g' /etc/apt/sources.list
+RUN apt-get -y update && apt-get -y install software-properties-common
+RUN add-apt-repository ppa:ubuntu-toolchain-r/test -y && apt-get -y update
+RUN apt-get -y install cmake \
```
varunagrawal: Is it possible to format this file? Having indentation would make it easier to read and update.

containers/gtsam-ci/Dockerfile.ubuntu-clang
```diff
@@ -0,0 +1,51 @@
...
+RUN gpg --keyserver keyserver.ubuntu.com --recv-key ${LLVM_KEY} || gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-key ${LLVM_KEY}
+RUN gpg -a --export ${LLVM_KEY} | apt-key add -
+# LLVM (clang) 9/14 is not in Bionic's repositories so we add the official LLVM repository.
+RUN add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main"
+# LLVM (clang) 9/14 is not in 22.04 (jammy)'s repositories so we add the official LLVM repository.
+RUN add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy main"
+
+RUN apt-get update
+RUN apt-get install -y clang-${COMPILER_VERSION} \
+g++-multilib \
```
varunagrawal: Do we need `g++` if we're using clang? Isn't this part of GCC? I may be misremembering why we put it here.

varunagrawal (commented):
This is very, very nice!!
I have a few comments about formatting and structuring, but everything looks great.

One question I have is whether we should put this here or in the [docker images](https://github.com/borglab/docker-images) repository since then we can have DockerHub build the images directly by pulling in the Dockerfile from there.
The images would get updated when we update that repo.

The reason for doing it this way is that these containers don't depend on GTSAM, but GTSAM depends on them, so the containers being in an "external" repo would make sense.

Once we finalize where to put this code and make the requested changes, I am more than happy to approve this PR. :)

.github/workflows/build-linux.yml
suveergarg: Seems like github actions does not have a `@latest` tag. docker/login-action@v3 and `actions/checkout@v4` are the latest at the moment.

.github/workflows/build-linux.yml
suveergarg: Other than 3 env variables that are constant, all the others are conditional based on the type of run from the matrix.  Do you have ideas of how to make that work with a --env-file?

containers/gtsam-ci/build_and_push.sh
suveergarg: I think that makes sense - I will make a PR to that repository.

suveergarg: https://github.com/borglab/docker-images/pull/2

containers/gtsam-ci/Dockerfile.ubuntu-clang
suveergarg: Great catch! I was not sure why it was needed either. Tested after removing  this line in the Dockerfile and things seem to work as expected.

containers/gtsam-ci/Dockerfile.ubuntu-clang
suveergarg: addressed

.github/workflows/build-linux.yml
Gold856: It's best to keep things pinned to a specific version anyways, because you don't want your CI breaking all of a sudden because upstream made a breaking change. It's slightly better for security as well (although ideally you use full-length commit hashes for that instead.)

varunagrawal: I don't think that applies in this case though. Since the action is only logging into DockerHub, there is no dependency involved for us.

However CI will break if v3 is taken offline in the future. In that case having @latest would ensure continued operation, since it doesn't matter how the action logs into DockerHub, just that it does.

Gold856: The dependency is on the action itself. It's a piece of code you depend on, like any other, that can break, or be compromised. Unpinning the version can cause breakage sometime in the future if they decide to make breaking changes. It ends up being the same result; something will break, and you probably will spend a few minutes figuring out why and fixing it up. But pinning is better practice.

varunagrawal: I see. Can you please help me understand this better?

In my mind, if the action introduces a breaking change in the future, the end result of logging into DockerHub is still going to happen as part of the contract between the CI and the action, which is what we want. Can you please explain more how will that affect us?
Or am I misunderstanding something?

suveergarg: I am not sure `@latest` is available for `actions/checkout` and `actions/login-action`. From what I read online - we are forced to use a version for these actions. I did try changing to @latest and CI start to fail with `latest tag not available`. Maybe I am missing something - Do you even see the option of putting the @latest tag?

.github/workflows/build-linux.yml
suveergarg: @varunagrawal do you think we should create multiple small `.env` files for the different cases and append them before passing it to `docker run`?

.github/workflows/build-linux.yml
varunagrawal: Oh I was just responding to @Gold856's comments. I saw your original note and was satisfied with that. :)

varunagrawal:
I have some things I need to finish by tomorrow afternoon, but after that I'll look at the docker-images PR and approve that.

Then I can verify that DockerHub is building those images, and we can come back here to migrate the CI.

Sound like a plan?

Also, @suveergarg can I count on you to record the before and after CI times? It could be simple screenshots or a table showing the amount of time CI took. We should see a major difference between the two thanks to the docker containers.

.github/workflows/build-linux.yml
varunagrawal: Let me think about that. You're work here is really sophisticated which is awesome, but most of us here in the lab are not as talented as you so I want to keep things simple as well.

.github/workflows/build-linux.yml
Gold856: For breaking changes, say that Docker changed how login credentials work (not unprecedented in the ecosystem; as an example, GitHub has removed password authentication for the API, and now you have to use a token instead). They release a new version of the login action that drops support for the old login system, but because you always asked for the latest version, your CI suddenly breaks, and you have to go fix it immediately. Or perhaps they changed the inputs around, such that your inputs no longer work. There's more to the contract than just what the actions does. Pinning to a specific version prevents CI from suddenly breaking.

The story is similar if you want to look at it from a security perspective, since if the repo is compromised, they can push up a new version and then dump your secrets or similar when your CI runs.

varunagrawal: But if DockerHub changes the way you login (e.g. use a token as you said) wouldn't the pinned version of the action fail anyway because it uses the old format? We would have to go change the action in that case as well right?

Regarding security, isn't it conventional wisdom that a newer version would include security patches and fixes? It's rarely the case that newer versions are more compromised (e.g. Dependabot never asks us to downgrade a dependency). Have you had different experiences that you can please elaborate on?

Gold856: It could fail, but sake of argument, let's say they keep the old version working for a certain period of time to allow for migration, and just release a new version that people can gradually migrate to that will continue to work after the deprecation period is over. That's generally how these kinds of migrations work. So the old version would keep working for a certain period of time, to allow people to migrate to the new version, until the old version eventually stops working and only the new version works.

For security, yes, typically newer versions include fixes, but the software world has changed radically over the years, and we're seeing many, many different supply chain attacks across many different ecosystems. PyPI, npm, and many other package repositories for various programming languages have had malicious packages uploaded to them, and even popular, well-known packages like Nx from the JS ecosystem [was compromised recently](https://nx.dev/blog/s1ngularity-postmortem). Even [some actions on GitHub have had this happen](https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066). Is it common? No. But it happens, and it happens often enough to be a concern. Ideally everyone uses full-length commit hashes (but that's out of scope for this PR).

If you pin to a tag (or ideally, a commit hash), things won't break out from under you, and you significantly lessen the small, but very real and impactful, chance of getting compromised by a compromised action. Simple as that.

varunagrawal: Won't it break even after the migration period ends? Usually those are for developers and since we're not using the docker API, I don't think we'll get a notice of that. Plus, as an academic lab, we don't have dedicated members who look after these things. :(

I see your point, but I guess we're coming from different worlds since everything you mentioned is what a company with a dedicated DevOps team (which I'm assuming you have) would be responsible for.

dellaert:
Is this a big deal? Let’s just pin?

.github/workflows/build-linux.yml
dellaert: Let’s just pin - I don’t see a downside.

varunagrawal: Yup, I already confirmed that with Suveer.

Gold856: I _am_ the DevOps person :) Hence my concern when you suggested unpinning the actions version. And yeah, I get it, not everyone has people dedicated to this. I'm just trying to make sure someone doesn't have a bad day because CI exploded :) Thanks for hearing me out!

varunagrawal: Ohhh that's awesome and I really appreciate your advice and information on this matter. I learned a lot from you. :)

varunagrawal: @Gold856 I went through the articles you shared. Super cool!! I'm going to start pinning my actions from now on thanks to your advice. :)

suveergarg:
@varunagrawal  - sounds good to me. The before and after runtimes seem to be variable - but it seems there is atleast a tiny bit of speed up. I took the before screenshots from the runs on this repo and compared it to my fork. Maybe you have custom runners that run these tests faster. So the best before and after comparison would be if we could run CI on this PR with the borglab account. See below:

suveergarg:
Before:

before_linux
before_special

suveergarg:
After:

after_linux
after_special_cases

dellaert:
@suveergarg please merge in develop as there have been several merged PRs now. Also, I'm unsure as to where we stand now on whether this is ready to merge - or still needs dockerhub changes.

suveergarg:
@dellaert I have merged develop and made the required changes so that my PR is still valid. Here is the paired PR [https://github.com/borglab/docker-images/pull/3/files] on the `docker-images` repo that keeps things consistent. I think at the moment, things are ready to go if we are ok with using the `build_and_push.sh` script in `docker-images/gtsam-ci` for uploading the images to docker hub. If we want to use the automated system within docker hub to trigger uploads of docker images, we probably need a few more changes - which is worth a more deeper discussion.

suveergarg:
@varunagrawal : why are only a max of 4 processors used for building GTSAM in CI? I think if this is set to nproc - 2 or similar, there might be a boost in speed.
 
 I see that in .github/scripts/unix.sh
 
  ```
if [ "$(uname)" == "Linux" ]; then
    if (($(nproc) > 2)); then
      cmake --build build -j4
    else
      cmake --build build -j2
    fi
  elif [ "$(uname)" == "Darwin" ]; then
    cmake --build build -j$(sysctl -n hw.physicalcpu)
  fi
  ```

varunagrawal:
That's because when we use a greater number, it results in more RAM consumption and GitHub kills the runner when it reaches a point.

We've found that -j4 is a nice sweetspot.

.github/workflows/build-linux.yml
```diff
@@ -83,56 +77,30 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Install Dependencies
+      - name: Pull Docker Image
```
varunagrawal: I don't think we need to pull the container. As per the [Actions docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/run-jobs-in-a-container) we can specify the container directly and Actions will run the CI inside the container image.

That being said, we may need to update the container images to not run a script as the ENTRYPOINT.

.github/workflows/build-special.yml
```diff
@@ -91,28 +84,8 @@ jobs:
...
-            gpg --keyserver keyserver.ubuntu.com --recv-key $LLVM_KEY || gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-key $LLVM_KEY
-            gpg -a --export $LLVM_KEY | sudo apt-key add -
-            sudo add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy main"
-          fi
-
-          sudo apt-get -y install cmake build-essential pkg-config libpython3-dev python3-numpy libicu-dev ninja-build
-          sudo apt-get install -y clang-${{ matrix.version }} g++-multilib mold
-          echo "CC=clang-${{ matrix.version }}" >> $GITHUB_ENV
-          echo "CXX=clang++-${{ matrix.version }}" >> $GITHUB_ENV
+      - name: Pull Docker Image
```
varunagrawal: Same deal here.

.github/workflows/build-special.yml
```diff
@@ -182,9 +153,16 @@ jobs:
...
+            -e GTSAM_BUILD_UNSTABLE=${{ env.GTSAM_BUILD_UNSTABLE || 'ON' }} \
+            -e GTSAM_ALLOW_DEPRECATED_SINCE_V43=${{ env.GTSAM_ALLOW_DEPRECATED_SINCE_V43 || 'OFF' }} \
+            -e GTSAM_USE_QUATERNIONS=${{ env.GTSAM_USE_QUATERNIONS || 'OFF' }} \
+            -e GTSAM_WITH_TBB=${{ env.GTSAM_WITH_TBB || 'OFF' }} \
+            -e GTSAM_POSE3_EXPMAP=${{ env.GTSAM_POSE3_EXPMAP || 'ON' }} \
+            -e GTSAM_ROT3_EXPMAP=${{ env.GTSAM_ROT3_EXPMAP || 'ON' }} \
+            -e GTSAM_BUILD_EXAMPLES_ALWAYS=${{ env.GTSAM_BUILD_EXAMPLES_ALWAYS || 'OFF' }} \
+            -e GTSAM_USE_SYSTEM_EIGEN=${{ env.GTSAM_USE_SYSTEM_EIGEN || 'OFF' }} \
+            -e GTSAM_USE_SYSTEM_METIS=${{ env.GTSAM_USE_SYSTEM_METIS || 'OFF' }} \
+            borglab/gtsam-ci:${{ matrix.os }}-${{ matrix.compiler }}-${{ matrix.version }}
```
varunagrawal: Please add a newline at the end.

dellaert:
> @dellaert I have merged develop and made the required changes so that my PR is still valid. Here is the paired PR [https://github.com/[borglab/docker-images/pull/3](https://github.com/borglab/docker-images/pull/3)/files] on the `docker-images` repo that keeps things consistent. I think at the moment, things are ready to go if we are ok with using the `build_and_push.sh` script in `docker-images/gtsam-ci` for uploading the images to docker hub. If we want to use the automated system within docker hub to trigger uploads of docker images, we probably need a few more changes - which is worth a more deeper discussion.

OK, @varunagrawal and I meet on Thursdays so maybe we can reserve some time to chat about this. We can then also merge this PR in the meeting if it's ready to go :-)

Many thanks for this !!!!!

.github/workflows/build-special.yml
suveergarg: One thing to note here is that we will not be able to create swap space from inside the container. 
If it is ok to run these without the additional swap space, I can make the switch.

varunagrawal: Hmm that's a good point. I think it should be fine. If we have issues, we can update the container to add swap space there.

varunagrawal:
@dellaert one more benefit of this approach is that we can now debug the CI locally by using the docker container as the common environment. This is the biggest win IMO.

varunagrawal:
Also we can do the same with a Windows container which will remove the need for having a Windows machine. I know this has been a major pain point for lots of people with Windows and the Windows CI. :)

varunagrawal:
FYI

https://learn.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-base-images#choosing-a-base-image

And

https://github.com/dockur/windows

Dockur also has a similar image for MacOS.

varunagrawal:
@suveergarg merge in the latest develop. I've added extra swap space for the vcpkg ubuntu runner so it runs to completion.

dellaert:
I think windows is under control. Let's focus on landing this. Talk Thursday.

suveergarg:
@varunagrawal  Actions now run inside the container. Entrypoint is overriden by default, so no additional work was needed there. It definitely cleans up the CI files a lot. This PR should be near completion now. CI jobs complete successfully on my fork.

Re : Window and MacOS containerization  - Those docker  images still need a host kernel that matches the container OS. So windows containers run on windows and MacOS containers run on mac. If you guys have a spare system (with Windows and Mac) that you can add me as a user with a remote desktop or something - I could crack this.

varunagrawal:
> Re : Window and MacOS containerization - Those docker images still need a host kernel that matches the container OS. So windows containers run on windows and MacOS containers run on mac. If you guys have a spare system (with Windows and Mac) that you can add me as a user with a remote desktop or something - I could crack this.

Regarding this, it seems like this can be accomplished by some kind of virtualization. It's an interesting problem to think about. :)

varunagrawal (approved):
Wow! Isn't this PR so beautiful now? Very clean and we have enabled debugging of the Linux CI locally. Awesome work @suveergarg.

We've also figured out the process of getting the docker containers running in the CI, so if we can get Windows and MacOS supported as well, that would be amazing.

varunagrawal:
Regarding containers with virtualization, check these out: https://hub.docker.com/u/dockurr

It uses Qemu KVM as the virtualization technology so we would have to run the containers on Linux runners.

suveergarg:
@varunagrawal let me see if I can throw together something for windows and macos as well.

dellaert:
Mac I’m ok, I’d like to steer clear of windows - we have an enthusiastic
core of people that just reduced build time by half and I’d like to not
mess with that rn.

suveergarg:
@dellaert  sounds good. From what I have been reading about these "virtualized" docker images running macos with a KVM/Qemu layer, it is almost better to run CI for them natively on the host OS than dockerizing these workflows due to the overhead and size of these images. The other thing is that unlike ubuntu docker containers, these macos docker containers do not support `RUN` steps using standard macos commands like `xcode`,` brew` etc.
The mac-os containers are more setup for people who want to run macos on their ubuntu systems - not as much for the flexibility and speed that CI would benefit from. FYI @varunagrawal 

I am happy to continue contributing in other ways. Let me know if either of you have ideas. I have a robotics background so could be useful in a lot of places.

PR #2067, opened 2025-03-21
Python bindings: Enable use of Similarity2/Similarity3 in nonlinear optimization

mhalber:
When building python bindings from the develop branch, the installed bindings (v4.3a0) do contain `Similarity2` and `Similarity3` objects, as well as `BetweenFactorSimilarity3`. However one cannot use these in optimization (`NonlinearFactorGraph`/'Values'), since addPriorSimilarity3, insert and at methods are not exposed.

This is a simple PR extends python bindings to support nonlinear optimization in python using `Similarity2` and `Similarity3` constraints.

To test the extended python functionality I've added two tests to `test_Sim3.py` that are adopted from `gtsam/geometry/tests/testSimilarity3.cpp`

Please not that I'm not super familiar with the gtsam codebase. While things seem to be working for my use case, if there are any changes required to merge this to the develop branch I'm very happy to hear your feedback and address any issues. 

Thanks!

### DIFF
--- a/gtsam/geometry/Similarity2.h
+++ b/gtsam/geometry/Similarity2.h
@@ -200,6 +200,19 @@ class GTSAM_EXPORT Similarity2 : public LieGroup<Similarity2, 4> {
   /// Dimensionality of tangent space = 4 DOF
   inline size_t dim() const { return 4; }
 
+ private:
+
+  #if GTSAM_ENABLE_BOOST_SERIALIZATION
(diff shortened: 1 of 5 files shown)


varunagrawal (approved):
LGTM, thanks!

dellaert:
Awesome contribution, @mhalber , many thanks !!!!

PR #2068, opened 2025-03-22
Clean up Ordering + headers

dellaert:
...and add test with disconnected graph.

### DIFF
--- a/examples/SFMExample_bal_COLAMD_METIS.cpp
+++ b/examples/SFMExample_bal_COLAMD_METIS.cpp
@@ -23,7 +23,6 @@
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
 #include <gtsam/nonlinear/LevenbergMarquardtOptimizer.h>
 #include <gtsam/inference/Symbol.h>
-#include <gtsam/inference/Ordering.h>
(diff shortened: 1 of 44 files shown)


dellaert:
@akshay-krishnan the test I added passes. Can you share your failing test?

akshay-krishnan:
This was my test, I did not EXPECT anything here, but was hoping to not crash when constructing ordering. 

```

#ifdef GTSAM_SUPPORT_NESTED_DISSECTION
TEST(Ordering, metisBrokenGraph) {

  SymbolicFactorGraph symbolicGraph;

  symbolicGraph.push_factor(0);
  symbolicGraph.push_factor(0, 1);
  symbolicGraph.push_factor(1, 2);
  symbolicGraph.push_factor(3);

  Ordering metis = Ordering::Metis(symbolicGraph);
  const vector<int> xadjExpected{0, 1, 3, 4}, adjExpected{1, 0, 2, 1};
}
#endif
```
It did not fail, but printed the message here: https://github.com/borglab/gtsam/blob/fae071d6037833797f02cac1222163d60002bed6/gtsam/inference/Ordering.cpp#L242

It is similar to the test you have added. Did you make any changes to the implementation? I dont see any.

PR #2077, opened 2025-04-01
Left invariant EKF on SE(2) without factors using odometry/gps measurements.

scottiyio:
A Left Invariant IEKF on SE(2) using odometry and GPS measurements.


dellaert:
@scottiyio , I think this should probably be moved to tests/testIEKF.cpp, so you can compare with the reference. In that same file you could do the SO(3) example. Then the next step would be to add a template in `nonlinear/InvariantEKF.h`, and document it in `nonlinear/doc/InvariantEKF.ipynb`, with code. That in turn would need the wrapper to instantiate the template for Pose2 and Rot3 (at least). I'd also add NavState so we can start playing with ${SE}_2(3)$. PS @mkielo3 knows how to do the wrapper bit - and it will *also* create the MATLAB wrapper :-)

PR #2078, opened 2025-04-02
Discrete-Continuous Smoothing and Mapping

varunagrawal:
An implementation of DCSAM (Doherty22ral) using our Hybrid Factor Graph framework.

I've added a new factor `DiscreteBoundaryFactor` which acts as the `DCDiscreteFactor` from DCSAM, and added as many tests as possible from the DCSAM repo.


keevindoherty:
Hey! This is so cool!!

varunagrawal:
I want to add an analog to the SemanticBearingRangeFactor but will do that once I have finished some pending task items.

dellaert:
Hey @keevindoherty , you are of course the DC-SAM guru. Would you have time to review the PR with your expert eye? :-)

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+#include <gtsam/inference/Symbol.h>
+#include <gtsam/nonlinear/ISAM2.h>
+#include <gtsam/nonlinear/Marginals.h>
+#include <gtsam/nonlinear/NonlinearFactorGraph.h>
+
+#include <map>
+#include <set>
+#include <tuple>
+#include <utility>
+#include <vector>
```
keevindoherty, line 37: Does GTSAM follow the Google style guide? It suggests [putting the C++ stdlib includes first here](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+
+namespace gtsam {
+
+class GTSAM_EXPORT DCSAM {
```
keevindoherty, line 49: It could be useful to add a brief description of DC-SAM here ("class that implements the hybrid alternating minimization solver from the RA-L 2022 paper ...." or whatever you like), as in the [Shonan averaging](https://github.com/borglab/gtsam/blob/develop/gtsam/sfm/ShonanAveraging.h#L107-L121) code? It might be helpful to have a reference to either the RA-L or the arXiv so a user can find the mathematical details?

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+   * Given the latest discrete values (discreteVals),
+   * a set of new factors (newFactors),
+   * and an initial guess for any new keys (initialGuess),
+   * this function marks any affected keys as such, and calls
+   * `isam_.update` with the new factors and initial guess.
+   * See implementation for more detail.
+   */
+  void updateContinuousInfo(const DiscreteValues &discreteVals,
+                            const NonlinearFactorGraph &newFactors,
+                            const Values &initialGuess);
```
keevindoherty: I realize I am reviewing my own code here, so I'm sorry that I left this in such a mess. I believe:
- The `update{Continuous, Discrete}Info` functions should be `private` or `protected` functions not exposed as part of the normal `public` API, since they're really implementation details of DC-SAM
- They probably should do something other than what they do now.
- They should _definitely_ be renamed.

`updateContinuous` is really very different from `updateDiscrete`. `updateContinuous` really a version of `solveDiscrete` but (a) for continuous states and (b) it modifies the internal state of DCSAM. I think it could be fine to have `updateContinuous` and `updateDiscrete` public (or protected?) member functions, which would allow child classes to modify update schedules, for example, but their behavior should be closer to what `iSAM2` does in the continuous-only case (and have a similar API, taking new factors and new values).

The functions `update{Continuous, Discrete}Info` are just really poorly named (sorry again!). It would be better for these functions to isolate the cache update behavior (for each hybrid factor copy in the continuous graph, update the discrete state cache, and vice versa).

I think the way to slice this would be to start by inlining all of this in the `update` function, since that is the only place any of these functions are actually called and then try to figure out what reusable chunks can be pulled out, and what sort of encapsulation makes sense.

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+   */
+  HybridValues calculateEstimate() const;
+
+  /**
+   * Used to obtain the marginals from the solver.
+   *
+   * NOTE: not obviously correct (see DCSAM.cpp implementation) at the moment.
+   * Should perhaps retrieve the marginals for the factor graph obtained as
+   * isam_.getFactorsUnsafe() and `dfg_` rather than taking as a parameter?
+   * I think this was originally intended to mimic the gtsam `Marginals` class.
```
keevindoherty, line 142: I agree with all of this 🙂

gtsam/hybrid/DCSAM.cpp
```diff
@@ -0,0 +1,167 @@
...
+HybridValues DCSAM::calculateEstimate() const {
+  // NOTE: if we have these cached from solves, we could presumably just return
+  // the cached values.
+  Values continuousVals = isam_.calculateEstimate();
+  DiscreteValues discreteVals = dfg_.optimize();
+  HybridValues values(VectorValues(), discreteVals, continuousVals);
+  return values;
+}
+
+//TODO(Varun) Create HybridMarginals
```
keevindoherty: This sounds cool! I'd be curious how that would work.

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+   * `isam_.update` with the new factors and initial guess.
+   * See implementation for more detail.
+   */
+  void updateContinuousInfo(const DiscreteValues &discreteVals,
+                            const NonlinearFactorGraph &newFactors,
+                            const Values &initialGuess);
+
+  /**
+   * Solve for discrete variables given continuous variables. Internally, calls
+   * `dfg_.optimize()`
```
keevindoherty: This is not strictly accurate, I think, because the solve actually depends as well on the current state of the discrete variables.

For example, in the case of a max-mixture of hybrid factors (or an EM-factor of hybrid factors) the mode you're in (or mode probabilities) can depend on both the continuous assignment _and_ the discrete assignment.

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+   *
+   * @param graph
+   * @param continuousEst
+   * @param dfg
+   */
+  // DCMarginals getMarginals(const NonlinearFactorGraph &graph,
+  //                          const Values &continuousEst,
+  //                          const DiscreteFactorGraph &dfg);
+
+  DiscreteFactorGraph getDiscreteFactorGraph() const { return dfg_; }
```
keevindoherty: Probably should return a `const DiscreteFactorGraph&` ?

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+
+  NonlinearFactorGraph getNonlinearFactorGraph() const {
```
keevindoherty: Probably should be `const NonlinearFactorGraph&` ?

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+   *
+   * @param graph - a HybridNonlinearFactorGraph containing all factors to add.
+   * @param initialGuessContinuous - an initial guess for any new
+   * continuous keys that appear in the updated factors
+   * (or if one wants to force override previously obtained continuous values).
+   * @param initialGuessDiscrete - Initial guess for discrete variables.
+   */
+  void update(const HybridNonlinearFactorGraph &graph,
+              const Values &initialGuessContinuous = Values(),
+              const DiscreteValues &initialGuessDiscrete = DiscreteValues());
```
keevindoherty: I wonder if the new hybrid machinery could be used to simplify this? Is there some kind of `HybridValues` (or equivalent?)

gtsam/hybrid/DiscreteBoundaryFactor.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+namespace gtsam {
+
+/* ************************************************************************ */
+DecisionTreeFactor DiscreteFactorFromErrors(
+    const DiscreteKeys& discreteKeys,
+    const AlgebraicDecisionTree<Key>& errors) {
+  double min_log = errors.min();
+  AlgebraicDecisionTree<Key> potentials(
+      errors, [&min_log](const double x) { return exp(-(x - min_log)); });
```
keevindoherty, line 28: Niiiice

gtsam/hybrid/DiscreteBoundaryFactor.h
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * A discrete probabilistic factor which computes its values from the
+ * continuous factors connected to it and a set of values.
+ *
+ * Helps in computing the discrete probabilities at the boundary of
+ * discrete and continuous factors.
+ *
+ * @ingroup discrete
+ */
+class GTSAM_EXPORT DiscreteBoundaryFactor : public DecisionTreeFactor {
```
keevindoherty, line 46: Really cool

gtsam/hybrid/DiscreteBoundaryFactor.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+/* ************************************************************************ */
+DecisionTreeFactor DiscreteBoundaryFactor::ComputeDiscreteBoundary(
+    const DiscreteKeys& dkeys,
+    const HybridNonlinearFactor::FactorValuePairs& factors,
+    const gtsam::Values& values) {
+  auto calculateError = [&](const NonlinearFactorValuePair& pair) -> double {
+    if (pair.first) {
+      auto gaussianNoiseModel = std::dynamic_pointer_cast<noiseModel::Gaussian>(
+          pair.first->noiseModel());
```
keevindoherty: I vaguely remember doing some weird things to handle different noise model cases. Will this "just work" even if you were using a Cauchy noise model or something?

gtsam/hybrid/DiscreteBoundaryFactor.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ *  @file DiscreteBoundaryFactor.cpp
+ *  @date Mach, 2025
+ *  @author Varun Agrawal
+ */
+
+#include <gtsam/hybrid/DiscreteBoundaryFactor.h>
```
keevindoherty, line 18: Again, not sure if you guys use the Google style guide, but if so, consider [including what you use](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use) rather than relying on transitive includes here.

gtsam/hybrid/DCSAM.cpp
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * @file    DCSAM.cpp
+ * @brief   Discrete-Continuous Smoothing and Mapping for Factored Models
+ * @author  Kevin Doherty
+ * @author  Varun Agrawal
+ * @date    March 2025
+ */
+
+#include <gtsam/hybrid/DCSAM.h>
+#include <gtsam/hybrid/DiscreteBoundaryFactor.h>
```
keevindoherty, line 21: Again, not sure if you guys use the Google style guide, but if so, consider [including what you use](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use) rather than relying on transitive includes here.

gtsam/hybrid/DiscreteBoundaryFactor.h
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * @file DiscreteBoundaryFactor.h
+ * @date March, 2025
+ * @author Varun Agrawal
+ */
+
+#pragma once
+
+#include <gtsam/discrete/DecisionTreeFactor.h>
+#include <gtsam/hybrid/HybridNonlinearFactor.h>
```
keevindoherty, line 21: Again, not sure if you guys use the Google style guide, but if so, consider [including what you use](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use) rather than relying on transitive includes here.

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+      // - negLogK is log normalization constant from the noise model
+      return pair.second + pair.first->error(values);
+    } else {
+      // If the factor has been pruned, return infinite error
+      return std::numeric_limits<double>::infinity();
+    }
+  };
+  AlgebraicDecisionTree<Key> errors(dcMixture.factors(), calculateError);
+  DecisionTreeFactor dtf =
+      DiscreteFactorFromErrors(dcMixture.discreteKeys(), errors);
```
keevindoherty: I think this is recycling some of the code used in `DiscreteBoundaryFactor`, I wonder if you could just use that directly instead? Or maybe I'm misunderstanding something here.

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+ * Test simple DiscretePriorFactor. We construct a DiscreteFactor factor
+ * graph with a single binary variable d1 and a single custom
+ * DiscretePriorFactor p(d1) specifying the distribution:
+ * p(d1 = 0) = 0.1, p(d1 = 1) = 0.9.
+ *
+ * After constructing the factor graph, we validate that the marginals are
+ * correct (i.e. they match the input p(d1)) and the most probable estimate
+ * (computed by maximizing p(d1) over possible values of d1) is 1.
+ */
+TEST(DCSAM, DiscretePriorFactor) {
```
keevindoherty: Strictly speaking this isn't a `DCSAM` test, it is more like an example of using the GTSAM discrete factor graph machinery (which was useful to me at the time it was originally written, because we [in John's lab] were still learning how this stuff worked). It probably could be removed if there are other discrete examples / tests (or moved to be co-located with those if you guys think it is useful).

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+ * Construct a single factor (for a single variable) consisting of a
+ * discrete-conditional mixture.
+ * Here we have a "null hypothesis" consisting of
+ * a Gaussian with large variance and an "alternative hypothesis"
+ * consisting of a Gaussian with smaller variance.
+ * After initializing the continuous variable far away from
+ * the ground-truth solution (x1 = 0), the discrete hypothesis selector
+ * will initially choose the null hypothesis.
+ */
+TEST(DCSAM, DiscreteMixture) {
```
keevindoherty: Again, technically not a `DCSAM` test and the motivation was really more for my own understanding. You can keep it or toss it, but you certainly have my blessing to remove it if you want 😂

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+ * far away from the ground-truth solution (x1 = 0), the discrete hypothesis
+ * selector will initially choose the null hypothesis.
+ *
+ * After a step of continuous optimization, the continuous solution will move
+ to
+ * x1 = 0 (since the problem is linear, we have convergence in one step).
+ * Finally, updating the discrete value will show that the correct discrete
+ * hypothesis will shift to the "alternative hypothesis."
+ */
+TEST(DCSAM, ContinuousMixture) {
```
keevindoherty: Same comment as above re: "not actually a `DCSAM` test"

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+const double meas_sigma = 1.0;
+
+auto prior_noise = noiseModel::Isotropic::Sigma(3, prior_sigma);
+auto meas_noise = noiseModel::Isotropic::Sigma(3, meas_sigma);
+
+Pose2 pose0(0, 0, 0);
+Pose2 dx(1, 0, 0.78539816);
+Pose2 noise(0.01, 0.01, 0.01);
+
+}  // namespace slam_fixture
```
keevindoherty, line 119: So much cleaner 🙂

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+
+/**
+ * This is a basic (qualitative) octagonal pose graph SLAM test
```
keevindoherty:
```suggestion
 * This is a basic octagonal pose graph SLAM test
```
It's not qualitative anymore! (Thanks for making this _actually test something_!)

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+  expectedContinuous.insert<Pose2>(X(4), Pose2(0.918069, 2.41298, -3.10159));
+  expectedContinuous.insert<Pose2>(X(5), Pose2(-0.0805657, 2.35886, -2.30619));
+  expectedContinuous.insert<Pose2>(X(6), Pose2(-0.740961, 1.60781, -1.5108));
+  expectedContinuous.insert<Pose2>(X(7), Pose2(-0.66688, 0.61046, -0.715398));
+
+  EXPECT(assert_equal(expectedContinuous, dcvals.nonlinear(), 1e-5));
+}
+
+/**
+ * This is a basic qualitative octagonal pose graph SLAM test
```
keevindoherty, line 180:
```suggestion
 * This is a basic octagonal pose graph SLAM test
```
(as above)

It's worth mentioning that both of these tests are really just making sure the continuous solver in DCSAM does what we expect. They're not suuuper valuable, but it is nice to know that running the continuous solver inside DC-SAM isn't fundamentally breaking things, I guess.

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+  expectedContinuous.insert<Pose2>(X(5), Pose2(-0.0805657, 2.35886, -2.30619));
+  expectedContinuous.insert<Pose2>(X(6), Pose2(-0.740961, 1.60781, -1.5108));
+  expectedContinuous.insert<Pose2>(X(7), Pose2(-0.66688, 0.61046, -0.715398));
+
+  EXPECT(assert_equal(expectedContinuous, dcvals.nonlinear(), 1e-5));
+}
+
+/**
+ * This test is a sanity check that DCSAM works correctly for solely
+ discrete problems
```
keevindoherty, line 251: Same as for the continuous case - maybe not super useful? But could be good to make sure nothing is fundamentally broken...

gtsam/hybrid/tests/testDCSAM.cpp
```diff
@@ -0,0 +1,685 @@
...
+
+  // Get the most probable estimate
+  const size_t mpeD = dcvals.discrete().at(dk.first);
+
+  // Ensure that the prediction is correct
+  EXPECT_LONGS_EQUAL(1, mpeD);
+}
+
+/**
+ * This is a basic qualitative octagonal pose graph SLAM test with two
```
keevindoherty:
```suggestion
 * This is a basic octagonal pose graph SLAM test with two
```
(as above, you made it a real test!!)

gtsam/hybrid/DCSAM.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+namespace gtsam {
+
+DCSAM::DCSAM() {
+  // Setup isam
+  ISAM2Params isam_params;
+  isam_params.relinearizeThreshold = 0.01;
+  isam_params.relinearizeSkip = 1;
+  isam_params.setOptimizationParams(ISAM2DoglegParams());
+  isam_ = ISAM2(isam_params);
```
keevindoherty: These should probably just be the true ISAM2 default params?

gtsam/hybrid/DCSAM.h
```diff
@@ -0,0 +1,202 @@
...
+#include <vector>
+
+namespace gtsam {
+
+class GTSAM_EXPORT DCSAM {
+ private:
+  // Global factor graph and iSAM2 instance
+  NonlinearFactorGraph nfg_;
+  DiscreteFactorGraph dfg_;
+  ISAM2 isam_;
```
keevindoherty, line 61: I always wanted to be able to configure what solver you used here. In principle you should be able to use any continuous solver you want (and any discrete solver, really). Would be a cool feature in the future.

keevindoherty (commented):
This is really really cool to see! It's awesome that you guys have been pushing forward on all of this, and I hope it can help increase the community's awareness of (and interest in) these hybrid estimation problems and solvers.

I tried to give a somewhat comprehensive review - All my complaints are about code that I wrote 😂 I tried to give some ideas I've had for cool things to do and opportunities for refactoring.

The `DiscreteBoundaryFactor` seems really nice!

gtsam/hybrid/DCSAM.h
varunagrawal: We use the clang formatter with Google as the style, but auto-formatting seems to prefer this order. :/

gtsam/hybrid/DCSAM.h
varunagrawal: Yes there is!!

gtsam/hybrid/DiscreteBoundaryFactor.cpp
varunagrawal: Unfortunately, I don't think so. We haven't really considered robust noise models in our scheme yet, but that's something we hope to accomplish soon!

varunagrawal: I believe all we need is a way to define `logDetR` for RobustNoiseModel and that should allow us to use `noiseModel::Base` as the parent class.

gtsam/hybrid/tests/testDCSAM.cpp
varunagrawal: No this is exactly the same. I used this as a test before making the factor so I should update it.

gtsam/hybrid/DCSAM.h
dellaert: It's not entirely Google, unfortunately.
Google style is: close (gtsam) to far (std) inclusion. But that has largely fallen by wayside :-(

dellaert:
@varunagrawal "deleted the branch develop [1 hour ago](https://github.com/borglab/gtsam/pull/2078#event-17173825487)" ??

varunagrawal:
> @varunagrawal "deleted the branch develop [1 hour ago](https://github.com/borglab/gtsam/pull/2078#event-17173825487)" ??

That's a red herring.

gtsam/hybrid/DCSAM.h
varunagrawal: That's interesting because the implementation doesn't seem to use the current discrete variables.

gtsam/hybrid/tests/testDCSAM.cpp
varunagrawal: I like these tests for the completeness. They might also help others in their understanding. 🙂

gtsam/hybrid/DCSAM.h
varunagrawal: True! I'll add a todo

varunagrawal:
@keevindoherty based on your incredible review comments, I did a full overhaul of the class so now it should be smaller, have better named methods, and be cleaner to read. :)

dellaert:
I have not yet done a detailed review but I read through the unit tests. Comments there are:
- move non-DCSAM tests to other test files. If the idea is to re-use a fixture, put it in a .h file, and explain to user where they can find non-DCSAM tests.
- better explain the other tests. If these are from a paper, say so. E.g, what is a semantic factor? We should be able to get a quick read on what the hybrid problem is without leaving the file.

varunagrawal:
@dellaert I've addressed you comments.

This is in the docstring, but I am adding at additionally here: a semantic landmark is one with unknown correspondences.

dellaert (approved):
Awesome !!!

keevindoherty:
Woohoo!!! This is so exciting!

Thanks @varunagrawal for all your hard work on this!

PR #2082, opened 2025-04-05
AHRS docs and wrap

dellaert:
- Add PreintegratedRotation to wrapper
- Add navigation to docs site
- Add docs for PreintegratedRotation and AHRSFactor


gtsam/navigation/doc/PreintegratedRotation.ipynb
p-zach, line 1: It's probably just the Github view, but the figure at the bottom doesn't appear. If it appears in vscode then I imagine it will work in MyST.

p-zach (approved):
All looks good to me

gtsam/navigation/doc/PreintegratedRotation.ipynb
dellaert: It should work with myst. Github can't render plotly.

PR #2083, opened 2025-04-06
Move IncrementalFixedLagSmoother from unstable to stable

p-zach:
and its dependency, BayesTreeMarginalizationHelper. References in gtsam_unstable.i copied to nonlinear.i.


gtsam/nonlinear/nonlinear.i
```diff
@@ -731,6 +731,20 @@ virtual class BatchFixedLagSmoother : gtsam::FixedLagSmoother {
   VALUE calculateEstimate(size_t key) const;
 };
 
+#include <gtsam_unstable/nonlinear/IncrementalFixedLagSmoother.h>
```
dellaert: not unstable any more :-) Probably why CI fails

dellaert (approved):
Cool, thanks !!!! Will merge after CI runs (needs an edit of .i file)

dellaert:
Hmmm, apparently we can't have the wrapper in both gtsam and gtsam_unstable :-/ CI fails again.

PR #2085, opened 2025-04-09
Fix the FixedLagSmootherExample and move it out of gtsam_unstable

jmackay2:
Fix the include location since the `IncrementalFixedLagSmoother.h` moved out of `gtsam_unstable`. Otherwise this fails by throwing the warning in the header of the old location. Also move the Example out of `gtsam_unstable`.


varunagrawal (approved):
Thank you!

PR #2086, opened 2025-04-10
Release/4.2 python ctor wrapper stereo

zygfrydw:
This MR exposes `body_P_sensor` parameter in Python wrapper for `GenericStereoFactor3D`.

I am working on a stereo vision application for a robot where the camera frame differs from the body link. I cannot prototype in Python because the `body_P_sensor` parameter for `GenericStereoFactor3D` constructor is not exposed.

After the MR:
```
graph.push_back(GenericStereoFactor3D(
    StereoPoint2(uL, uR, v), noise_model, 1, 3, K, body_to_camera)
```

### DIFF
--- a/gtsam/slam/slam.i
+++ b/gtsam/slam/slam.i
@@ -167,6 +167,18 @@ virtual class GenericStereoFactor : gtsam::NoiseModelFactor {
   GenericStereoFactor(const gtsam::StereoPoint2& measured,
                       const gtsam::noiseModel::Base* noiseModel, size_t poseKey,
                       size_t landmarkKey, const gtsam::Cal3_S2Stereo* K);
+  GenericStereoFactor(const gtsam::StereoPoint2& measured,
+                      const gtsam::noiseModel::Base* noiseModel,
+                      size_t poseKey, size_t landmarkKey,
+                      const gtsam::Cal3_S2Stereo* K,
+                      const POSE& body_P_sensor);
+  GenericStereoFactor(const gtsam::StereoPoint2& measured,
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Awesome! After CI succeeds, let’s merge (whoever gets to it first).

dellaert:
Oopsies. Build fails on Linux. Something with FindBoost. Ideas anyone?

zygfrydw:
The "Install Boost" CI step fails due to `gzip: stdin: not in gzip format` after wget.

dellaert:
Oh, I see, you’re targeting the release branch. That requires a bit of care, actually. We are gearing up to release 4.3, so could you please target develop? I will close this PR as I think if you’re starting from 4.2 the PR will be problematic. Better to check out develop, patch slam.i, and PR to develop.

zygfrydw:
I have investigated the boost problem and the CI failed because the archive used for the CI does not exist any more.
The command `wget https://boostorg.jfrog.io/artifactory/main/release/1.67.0/source/boost_1_67_0.tar.gz` returns a HTML with an error instead of `tar.gz`.

I will port this MR to the `develop` branch for community, but for the project I am working on, we are using GTSAM 4.2. Due to a limited time schedule, I don't think we can switch to 4.3 in the foreseeable future, so we will continue with the fork.

zygfrydw:
I checked the `develop` branch, and the changes I propose are already there [slam.i](https://github.com/borglab/gtsam/blob/develop/gtsam/slam/slam.i#L204).

dellaert:
That sounds like a plan :-)

PR #2093, opened 2025-04-16
Add support for building MacOS wheels in `cibuildwheel`

yambati03:
This PR adds support for in `cibuildwheel` for generating Python wheels for MacOS 10.15+. 

<ins>**How was this tested?**</ins>
The wheel generated from a test run of the updated `cibuildwheel` was `pip installed`-ed on a laptop running MacOS 14.6. We then ran the following example https://github.com/borglab/gtsam/blob/develop/python/gtsam/examples/ImuFactorExample.py, which functioned as expected.

We also note that all builds ran successfully in a forked repository, and wheel files were generated for each OS + Python version pair.
image


dellaert (approved):
Very cool!!! Could you add to the PR comment how you know it works? I will then approve and merge.

dellaert:
Oops I already approved. So, merge at will but *after* you added (for tracking sake) how you now it works and why you are sure it will work in CI :-)

yambati03:
I `pip install`-ed the wheel generated in CI on a friends computer (since I don't have a Mac) and ran `import gtsam` and wrote a quick factor graph, which worked fine. But I also want to run the full suite of tests as well. Was hoping that you and/or @ktk-sdk would help me out with that at tomorrow's meeting. I've got a wheel ready to use :)

PR #2097, opened 2025-04-17
Update navbar

p-zach:
**Current navbar:**
image
**Updated navbar:**
image
"Getting started" removed because it was redundant; the links are in the sidebar.


p-zach:
I can change "GTSAM.org" to be different text that points to gtsam.org, but I couldn't think of a more appropriate title.

dellaert (approved):
Yay !

PR #2098, opened 2025-04-17
Ordering doc update

p-zach:
Additional content about COLAMD vs. METIS and better visuals. Also made Symbol more clear,


dellaert (approved):
Yeah!

PR #2100, opened 2025-04-17
Docs minor updates

p-zach:
Minor fixes in Ordering.ipynb and changed all pip installs to `%pip install --quiet gtsam-develop`


dellaert (approved):
Nice!

PR #2102, opened 2025-04-18
Add `cibuildwheel` targets for MacOS Silicon

yambati03:
This PR adds targets in the `cibuildwheel` workflow for MacOS ARM64. This will support M-series Macs.


dellaert (approved):
Yay!

PR #2106, opened 2025-04-22
Re-visiting Logmap

### LINKED ISSUE #1233: Regression in Logmap (v4.2a7)

dellaert:
In [issue 1233](https://github.com/borglab/gtsam/issues/1233) the behavior of Logmap was examined in detail, and a regression from GTSAM 4.0.3 to 4.2a7 was noted. In SO3.ipynb those graphs are reproduced so the issue can be further examined.

image

@ProfFan I'm at a loss right now how to simultaneously resolve #1233 and #887.


ProfFan:
I thought about this for a while, and I think the best possible solutions are:
a) ignore the issue
b) figure out a smooth map between the two (interpolation)

Also could look into what Sophus is doing, need a bit of time for that

dellaert:
codex: This revisits `SO3`/`Pose3` Logmap behavior and tests in response to a documented regression between GTSAM releases. Labeled bug because the work targets restoration of previously correct numerical behavior.

DLuminary:
The 2 + 2*R_aa in the near-π branch should be 1 + 2*R_aa - tr; the two differ by exactly 1 + tr = 4cos²(θ/2), so the branch is only exact at θ=π — that's the 1e-4 plateau. Fixing it means the branch point can move to tr > 0, and the threshold stops mattering at all: max error stays at 3.4e-16 for any switch between 174° and 60°. #887 stays green — 0/165 stress failures, test_axisangle → 3.1396582476.

Re: what Sophus does — it's Eigen::Quaternion(R) followed by an atan-based quaternion log, [citing Hertzberg et al. 2011](https://github.com/strasdat/Sophus/blob/main/sophus/so3.hpp#L285-L332). The quaternion extraction is the largest-pivot method, and GTSAM already ships it: [gtsam/3rdparty/Eigen/.../Quaternion.h#L816-L853](https://github.com/borglab/gtsam/blob/develop/gtsam/3rdparty/Eigen/Eigen/src/Geometry/Quaternion.h#L816-L853), which uses mat(i,i) - mat(j,j) - mat(k,k) + 1 — i.e. the correct pivot. This PR is algorithmically the same, with the quaternion left un-normalized since atan2 is scale-invariant. PR: #2723

dellaert:
Superseded by #2723

PR #2111, opened 2025-04-24
New docs: SLAM

p-zach:
New docs for the SLAM directory.

Current issues:
* FrobeniusFactor seems to need FrobeniusPrior. Wrap it?
* KarcherMeanFactor doesn't know how to use JacobianFactor
* LAGO: invalid map
* OrientedPlane not wrapped
* PlanarProjectionFactor: weird type error
* PoseRotationPrior: not sure why import isn't working
* PoseTranslationPrior: same as RotationPrior
* ReferenceFrameFactor: not wrapped
* RotateFactor: not wrapped
* SmartFactorParams: just needs a couple functions wrapped
* SmartProjectionFactor: not wrapped
* SmartProjectionPoseFactor: not wrapped
* SmartProjectionRigFactor: not wrapped
* StereoFactor: Cal3S2_Stereo.backproject not wrapped


dellaert:
Sure, wrap FrobeniusPrior and everything else :-)

p-zach:
Decided in meeting that I will resolve KarcherMean and PoseRotationPrior issues; @dellaert will look into LAGO and PlanarProjectionFactor issues. I will also fix any typographical errors and wrap all the unwrapped functions.

dellaert:
Sorry @p-zach I accidentally fixed KarcherMeanFactor. Hope you have not already done it also.

p-zach:
Fixed most of the notebooks' problems.
# Remaining issues
- OrientedPlane3Factor.ipynb
  - OrientedPlane3 does not act like a Value, so it can't be added to Values, etc. 
- ReferenceFrameFactor.ipynb
  - Templating issues. I think I should be able to fix these tomorrow.
- SmartProjectionFactor.ipynb, SmartProjectionRigFactor.ipynb
  - These have the same problem, which stems from an issue with wrap (<https://github.com/borglab/wrap/issues/175>). In short, their `add` functions aren't wrapped properly because some of the parameters are based on typedefs within the double-templated class (double-templated meaning like `SmartProjectionFactor<PinholeCamera<Cal3_S2>>`). In this case, it's PinholeCamera's Measurement/MeasurementVector. I have not yet found a solution to this.

dellaert:
# Remaining issues
> * OrientedPlane3Factor.ipynb

Fixed by adding to values.i

> * SmartProjectionFactor.ipynb

Fixed by using typedefs in SimpleCamera.h

> SmartProjectionRigFactor.ipynb

Wrapping works but type mismatch in notebook is still an issue

dellaert (approved):
Amazing effort ! Smart factors was harder than anticipated.
I think there are no more issues, so approving now. Let me know if not the case @p-zach .

p-zach:
@dellaert Great! I was wrestling with RigFactor too, glad you figured it out. All good to merge after CI

PR #2113, opened 2025-04-25
Left invariant EKF header file, with an example on SE(2) and NavState

scottiyio:

PR #2115, opened 2025-04-26
Invariant EKF

dellaert:
A Left Invariant IEKF on SE(2).
Changes by Frank: template methods rather than a base class
Replaces #2113 

This introduced the following class structure:
```mermaid
graph TD; 
ManifoldEKF --> LieGroupEKF --> InvariantEKF
````

### CODE REVIEW

**Comment by dellaert** [comment]
Ci on windows fails because of Unicode characters. I’ll fix together with any other comments.

File: examples/LIEKF_NavstateExample.cpp [hunk 1]
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * @file LIEKF_NavstateExample.cpp
+ * @brief LIEKF on NavState (SE_2(3)) with IMU (predict) and GPS (update)
+ * @date April 25, 2025
+ * @authors Scott Baker, Matt Kielo, Frank Dellaert
+ */
+
+#include <gtsam/base/Matrix.h>
+#include <gtsam/base/OptionalJacobian.h>
```
scottiyio, line 20: is this needed since OptionalJacobian.h is included in LIEKF.h?

examples/LIEKF_NavstateExample.cpp
```diff
@@ -0,0 +1,96 @@
...
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file LIEKF_NavstateExample.cpp
+ * @brief LIEKF on NavState (SE_2(3)) with IMU (predict) and GPS (update)
+ * @date April 25, 2025
+ * @authors Scott Baker, Matt Kielo, Frank Dellaert
+ */
+
+#include <gtsam/base/Matrix.h>
```
scottiyio, line 19: are line 19 and 20 needed includes, since they are already included in LIEKF.h?

examples/LIEKF_Rot3Example.cpp
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * @file LIEKF_Rot3Example.cpp
+ * @brief Left‐Invariant EKF on SO(3) with state‐dependent pitch/roll control
+ * and a single magnetometer update.
+ * @date April 25, 2025
+ * @authors Scott Baker, Matt Kielo, Frank Dellaert
+ */
+
+#include <gtsam/base/Matrix.h>
+#include <gtsam/base/OptionalJacobian.h>
```
scottiyio, line 21: same includes - required or not?

examples/LIEKF_SE2_SimpleGPSExample.cpp
```diff
@@ -0,0 +1,118 @@
...
+  // // Initialize the filter's state, covariance, and time interval values.
+  Pose2 X0(0.0, 0.0, 0.0);
+  Matrix3 P0 = Matrix3::Identity() * 0.1;
+
+  // Create the filter with the initial state, covariance, and measurement
+  // function.
+  LIEKF<Pose2> ekf(X0, P0);
+
+  // Define the process covariance and measurement covariance matrices Q and R.
+  Matrix3 Q = (Vector3(0.05, 0.05, 0.001)).asDiagonal();
```
scottiyio, line 60: Is there any better way to initialize matrices like this when we want different sigma^2 on the diagonal? Or is this just the best way if they are different valurd?

examples/LIEKF_SE2_SimpleGPSExample.cpp
```diff
@@ -0,0 +1,118 @@
...
+  Matrix2 R = (Vector2(0.01, 0.01)).asDiagonal();
```
scottiyio, line 61: change to Matrix2::Identity() * 0.01

examples/LIEKF_NavstateExample.cpp
scottiyio: Line (19) and (20) are already included in LIEKF.h

examples/LIEKF_Rot3Example.cpp
scottiyio: Line (20) and (21) are already included in LIEKF.h

examples/LIEKF_SE2_SimpleGPSExample.cpp
scottiyio: May want to find a better way to create covariance matrices when the sigma^2 are different on the diagonal

gtsam/navigation/LIEKF.h
```diff
@@ -0,0 +1,185 @@
...
+ * functions, along with measurement updates using tangent-space corrections.
+ *
+ * @date    April 24, 2025
+ * @authors Scott Baker, Matt Kielo, Frank Dellaert
+ */
+
+#pragma once
+
+#include <gtsam/base/Matrix.h>
+#include <gtsam/base/OptionalJacobian.h>
```
scottiyio: Same includes are not needed, in LIEKF.h already

scottiyio (approved):
Looks great! I saw no errors, just some minor includes that may not be needed.

examples/LIEKF_NavstateExample.cpp
dellaert: I try to follow https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

examples/LIEKF_Rot3Example.cpp
dellaert: Same: IWYU

dellaert:
@scottiyio I made the following changes:
- I added new unit-tests for state-dependent dynamics, with new method `predictMean` 
- renamed to InvariantEKF
- reverted this to draft PR because the connection with invariant filtering is not clear enough yet

I would like to look at Potokar’s code how the dynamics and measurements there are specified in general.

dellaert:
@scottiyio Gemini and I created the ManifoldEKF -> GroupEKF -> InvariantEKF hierarchy (plus tests)
Take a look.
GroupEKF or LieEKF or LieGroupEKF?

examples/IEKF_NavstateExample.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+using namespace std;
+using namespace gtsam;
+
+/**
+ * @brief Left-invariant dynamics for NavState.
+ * @param imu  6×1 vector [a; ω]: linear acceleration and angular velocity.
+ * @return     9×1 tangent: [ω; 0₃; a].
+ */
+Vector9 dynamics(const Vector6& imu) {
```
dellaert, line 34: @scottiyio I think this example still needs some unit-testing. In particular, show that
- $X \cdot U = X \cdot \exp(\delta t \xi^\wedge)$ the same as the nonlinear dynamics in Potokar
- Show that $Ad_{U^{-1}}$ matches the LIEKF Jacobian in Potokar.

gtsam/navigation/GroupEKF.h
```diff
@@ -0,0 +1,196 @@
...
+    using TangentVector = typename Base::TangentVector; ///< Tangent vector type for the group G
+    using MatrixN = typename Base::MatrixN; ///< Square matrix of size n for covariance and Jacobians
+    using Jacobian = Eigen::Matrix<double, n, n>; ///< Jacobian matrix type specific to the group G
+
+    /// Constructor: initialize with state and covariance
+    GroupEKF(const G& X0, const MatrixN& P0) : Base(X0, P0) {
+      static_assert(IsLieGroup<G>::value, "Template parameter G must be a GTSAM Lie Group");
+    }
+
+    /**
```
dellaert: I am thinking of *moving* these to InvariantEKF and comment that, if the dynamics does not depend on the state, they should use an InvariantEKF

scottiyio: I know that InvariantEKF inherits from this.

I wish we could call it something like "StateDependentEKF" because you should use GroupEKF on state dependency, but considering InvariantEKF inherits from here as a special case I don't know if that makes sense.

Or maybe, it inherits that the "state dependency" is zero. That structure, to me, makes it more clear thaty you should use Invariant if you have no state dependency but Group if there is.

examples/IEKF_NavstateExample.cpp
scottiyio: I'll write some test cases here

gtsam/navigation/GroupEKF.h
dellaert: I opted for
```mermaid
graph TD; 
ManifoldEKF --> LieGroupEKF --> InvariantEKF
````

PR #2120, opened 2025-05-01
New docs: hybrid

p-zach:
New docs for hybrid folder. There were a lot of classes in here that had many, many functions that were not yet wrapped. Frank also exposed `AlgebraicDecisionTreeKey` for methods like errorTree().

### Finished Notebooks
- HybridSmoother
- HybridValues
- HybridBayesTree
- HybridFactor
- HybridBayesNet
- HybridConditional
- HybridEliminationTree
- HybridFactorGraph
- HybridGaussianConditional
- HybridGaussianFactor
- HybridGaussianFactorGraph
- HybridGaussianISAM
- HybridGaussianProductFactor
- HybridJunctionTree
- HybridNonlinearFactor
- HybridNonlinearFactorGraph
- HybridNonlinearISAM


dellaert (approved):
Thanks @p-zach !!! I fixed all remaining wrapper issues

dellaert:
@varunagrawal FYI

.github/workflows/build-special.yml
```diff
@@ -197,11 +197,25 @@ jobs:
           echo "GTSAM_BUILD_UNSTABLE=OFF" >> $GITHUB_ENV
           echo "GTSAM 'unstable' will not be built."
 
-      - name: Set Swap Space
+      - name: Create swap (Linux only)
         if: runner.os == 'Linux'
-        uses: pierotofy/set-swap-space@master
-        with:
-          swap-size-gb: 8
+        shell: bash
```
varunagrawal: Can you please document what's going on here? It's much more complicated than what we had before and is not quite obvious.

gtsam/hybrid/HybridConditional.cpp
```diff
@@ -112,25 +112,25 @@ bool HybridConditional::equals(const HybridFactor &other, double tol) const {
 }
 
 /* ************************************************************************ */
-double HybridConditional::error(const HybridValues &values) const {
+double HybridConditional::error(const HybridValues &hybridValues) const {
```
varunagrawal, line 115: This change is a bit superficial since `HybridConditional::error` only ever accepts `HybridValues`.

gtsam/hybrid/HybridFactor.h
```diff
@@ -162,4 +162,7 @@ class GTSAM_EXPORT HybridFactor : public Factor {
 template <>
 struct traits<HybridFactor> : public Testable<HybridFactor> {};
 
+// For wrapper:
```
varunagrawal, line 165: The reason I haven't wrapped a lot of stuff in hybrid is primarily because of this class, and also because I want to add a new class called `KeyDecisionTree` which specializes the template `AlgebraicDecisionTree`.

This would have many benefits since we wouldn't have to always pass in `keyFormatter` arguments, and wrapping would be a cinch, but it is a big PR which is lower priority than some others (such as DCSAM).

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -193,16 +194,44 @@ AlgebraicDecisionTree<Key> HybridGaussianFactor::errorTree(
...
-    const DiscreteValues& assignment) const {
-  throw std::runtime_error("HybridGaussianFactor::restrict not implemented");
+  const DiscreteValues& assignment) const {
+  FactorValuePairs restrictedTree = this->factors_;  // Start with the original tree
+
+  const DiscreteKeys& currentFactorDiscreteKeys = this->discreteKeys();
+  DiscreteKeys newFactorDiscreteKeys;  // For the new, restricted factor
+
+  // Iterate over the discrete keys of the current factor
+  for (const DiscreteKey& factor_dk_pair : currentFactorDiscreteKeys) {
```
varunagrawal: Nitpick: The iterate can simply be called `discreteKey`.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -193,16 +194,44 @@ AlgebraicDecisionTree<Key> HybridGaussianFactor::errorTree(
...
+    const Key& key = factor_dk_pair.first;
+
+    // Check if this key is specified in the assignment
+    auto assignment_it = assignment.find(key);
```
varunagrawal: This can be combined with the below line
```cpp
if (assignment.find(key) != assignment.end()) {
```

and since `assignment` is a std::map, you can just do `assignment.at(key)` to get the assigned index. You shouldn't need the iterator `assignment_it`.

gtsam/hybrid/HybridGaussianFactor.cpp
```diff
@@ -193,16 +194,44 @@ AlgebraicDecisionTree<Key> HybridGaussianFactor::errorTree(
...
+
+    if (assignment_it != assignment.end()) {
+      // Key is in assignment: restrict the tree by choosing the branch
+      size_t assigned_value = assignment_it->second;
+      restrictedTree = restrictedTree.choose(key, assigned_value);
```
varunagrawal: Any reason you do this and not just call `restrictedTree.restrict(assignment)`?
You can then use the loop only for finding the assigned keys, which can possibly be done with a set intersection so it is faster than looping.

gtsam/hybrid/HybridNonlinearFactor.cpp
```diff
@@ -138,6 +138,7 @@ void HybridNonlinearFactor::print(const std::string& s,
     auto [factor, val] = v;
     if (factor) {
       RedirectCout rd;
+      std::cout << "(" << val << ") ";
```
varunagrawal: I would prefer this to be `std::cout << "(val=" << val << ") ";`.

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -72,14 +72,16 @@ class GTSAM_EXPORT HybridNonlinearFactor : public HybridFactor {
   /// Decision tree of nonlinear factors indexed by discrete keys.
   FactorValuePairs factors_;
 
-  /// HybridFactor method implementation. Should not be used.
+  /// HybridFactor method implementation. Should not be used for this class.
```
varunagrawal: Let's be consistent and use `/** ... */` for multiline docstrings?

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -122,84 +124,104 @@ class GTSAM_EXPORT HybridNonlinearFactor : public HybridFactor {
...
+
   /**
    * @brief Compute error of the HybridNonlinearFactor as a tree.
    *
-   * @param continuousValues The continuous values for which to compute the
-   * error.
-   * @return AlgebraicDecisionTree<Key> A decision tree with the same keys
-   * as the factor, and leaf values as the error.
+   * @param continuousValues The continuous gtsam::Values for which to compute the error.
+   * @return AlgebraicDecisionTree<Key> A decision tree with the same discrete
```
varunagrawal, line 142: Any reason you killed the blank line? It usually helps with readability.

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -122,84 +124,104 @@ class GTSAM_EXPORT HybridNonlinearFactor : public HybridFactor {
...
   AlgebraicDecisionTree<Key> errorTree(const Values& continuousValues) const;
 
   /**
    * @brief Compute error of factor given both continuous and discrete values.
    *
-   * @param continuousValues The continuous Values.
-   * @param discreteValues The discrete Values.
-   * @return double The error of this factor.
+   * @param continuousValues The continuous gtsam::Values.
+   * @param assignment The discrete Values (assignment for discrete keys).
```
varunagrawal: This should probably be `The assignment for the discrete keys.`

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -122,84 +124,104 @@ class GTSAM_EXPORT HybridNonlinearFactor : public HybridFactor {
...
-                                       const DiscreteValues& assignment) const;
+    const DiscreteValues& assignment) const;
 
   /// Linearize all the continuous factors to get a HybridGaussianFactor.
   std::shared_ptr<HybridGaussianFactor> linearize(
-      const Values& continuousValues) const;
+    const Values& continuousValues) const;
 
   /// Prune this factor based on the discrete probabilities.
+  /// @param discreteProbs A DecisionTreeFactor representing P(M) or P(M|...).
```
varunagrawal: Multiline comment please.

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -122,84 +124,104 @@ class GTSAM_EXPORT HybridNonlinearFactor : public HybridFactor {
...
+  /// Entries with probability 0 (or very small) in discreteProbs will lead to
+  /// pruning of corresponding branches in this factor.
   HybridNonlinearFactor::shared_ptr prune(
       const DecisionTreeFactor& discreteProbs) const;
 
   /// Restrict the factor to the given discrete values.
+  /// If all discrete keys in this factor are assigned, the result will be
```
varunagrawal: Multiline comment please.

gtsam/hybrid/HybridNonlinearFactor.h
```diff
@@ -122,84 +124,104 @@ class GTSAM_EXPORT HybridNonlinearFactor : public HybridFactor {
...
+  /// @}
 
  private:
   /// Helper struct to assist private constructor below.
   struct ConstructorHelper;
 
   // Private constructor using ConstructorHelper above.
   HybridNonlinearFactor(const ConstructorHelper& helper);
+
+#if GTSAM_ENABLE_BOOST_SERIALIZATION
```
varunagrawal, line 235: Is this tested? If not, please add the test or remove it.

gtsam/hybrid/hybrid.i
```diff
@@ -81,25 +114,50 @@ class HybridGaussianFactor : gtsam::HybridFactor {
...
+      const gtsam::Matrix& A, size_t parent,
+      const std::vector<std::pair<gtsam::Vector, double>>& parameters);
+  HybridGaussianConditional(
+      const gtsam::DiscreteKey& discreteParent, size_t key,  //
+      const gtsam::Matrix& A1, size_t parent1, const gtsam::Matrix& A2,
+      size_t parent2,
+      const std::vector<std::pair<gtsam::Vector, double>>& parameters);
+
+  // Standard API
+  gtsam::GaussianConditional::shared_ptr choose(
```
varunagrawal: `shared_ptr` has specific wrapper syntax: `gtsam::GaussianConditional*`.

Also did you test the matlab wrapper?

gtsam/hybrid/hybrid.i
```diff
@@ -81,25 +114,50 @@ class HybridGaussianFactor : gtsam::HybridFactor {
...
+//       const gtsam::DiscreteValues &discreteValues) const;
+  size_t nrComponents() const;
+  gtsam::KeyVector continuousParents() const;
+  double negLogConstant() const;
 
   gtsam::HybridGaussianFactor* likelihood(
       const gtsam::VectorValues& frontals) const;
   double logProbability(const gtsam::HybridValues& values) const;
   double evaluate(const gtsam::HybridValues& values) const;
+//   double operator()(const gtsam::HybridValues &values) const;
```
varunagrawal, line 226: This should just work right?

gtsam/hybrid/tests/testHybridBayesTree.cpp
```diff
@@ -495,6 +496,50 @@ TEST(HybridBayesTree, Choose) {
   EXPECT(assert_equal(expected_gbt, gbt));
 }
 
+/* ************************************************************************* */
+TEST(HybridBayesTree, EliminateMultifrontal) {
```
varunagrawal: This test doesn't seem to be verifying correctness. Seems more like a test for dot printing.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
```diff
@@ -191,9 +192,30 @@ TEST(HybridGaussianFactor, Error) {
...
+  DiscreteValues assignment1;
+  assignment1[m1.first] = 0;
+
+  auto f = hfg.restrict(assignment1);
+  auto restricted = std::dynamic_pointer_cast<HybridGaussianFactor>(f);
+  CHECK(restricted != nullptr);
+
+  // Check discrete keys now empty
+  DiscreteKeys expected_dk1;
+  GTSAM_PRINT(restricted->discreteKeys());
```
varunagrawal: Please kill prints.

gtsam/inference/BayesTree.h
```diff
@@ -249,7 +249,7 @@ namespace gtsam {
 
     /** private helper method for saving the Tree to a text file in GraphViz format */
     void dot(std::ostream &s, sharedClique clique, const KeyFormatter& keyFormatter,
-             int parentnum = 0) const;
+             size_t parentnum = 0) const;
```
varunagrawal, line 252: This may be problematic on different compilers due to the platform dependent `size_t` definitions.

varunagrawal (commented):
Okay this PR is a very nice effort, but I am not comfortable with the wrapping for 2 reasons:
1. Lack of tests: whatever I wrapped, I extensively unit tested, and my own work allowed for integration testing.
2. Verification with the Matlab wrapper. The matlab wrapper is more finicky than the python wrapper since it cares a whole lot more about return types and other smaller details. I don't see if this was checked (CI doesn't do it).

Moreover, I want to land other hybrid related PRs first (such as DCSAM) and include a new class called `Key[Algebraic]DecisionTree` which specializes `AlgebraicDecisionTree<Key>` to make its use much easier and cleaner. It's a low priority item since it simplifies the API but is a lot of effort for a single developer (aka me).

I am not selecting `Request changes` (which would be the right way to go), but I DO NOT recommend merging this yet.

varunagrawal (changes requested):
Please verify the matlab wrapper compiles and update the restrict method.

varunagrawal:
Couple of things:
1. I don't think my review comments were addressed?
2. The Matlab wrapper is indeed not compiling. Apparently it's been this PR along with 5 others that have caused it to break.

varunagrawal (approved):
Fixed everything so this should be good to go. :)

varunagrawal:
@dellaert can you take one more look at this to make sure it matches your expectations? There were some merge conflicts which I fixed, so wanted to get your sign off on those.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
```diff
@@ -55,7 +56,7 @@ TEST(HybridGaussianFactor, Constructor) {
 }
 
 /* ************************************************************************* */
-namespace test_hgn_constructor {
+namespace fixture {
```
dellaert: Dangerous, as namespaces are global. Better to keep it specific.

gtsam/hybrid/hybrid.i
dellaert: Try?

dellaert (approved):
Great! Two minor comments.

gtsam/hybrid/hybrid.i
varunagrawal: I tried this but I am getting an error from Pybind11. This is due to the fact that there are two `operator()` methods (i.e. it is overloaded), and the binding code generated is 
```cpp
.def("__call__", &gtsam::HybridGaussianConditional::operator());
```
which cannot disambiguate between the two `operator()` overloads in `HybridGaussianConditional` since there is no specification of the argument types. There is a way to fix this but it will need more wrapper updates, so commenting this out until then.

gtsam/hybrid/tests/testHybridGaussianFactor.cpp
varunagrawal: Reverted

PR #2126, opened 2025-05-07
Smart factors docs

dellaert:
- added SmartFactors.md, which gives detailed math and implementation details for smart factors
- RegularHessianFactor: Format, document, optimize slightly - Gemini 2.5


dellaert:
Merging without review as functionality-neutral

PR #2128, opened 2025-05-07
Pull in newest GeographicLib (2.5)

dellaert:
Just swapped in newest source distribution.

@jlblancoc and @JaiXJM-BB , their cmake changed a bit, so don;t know how to fix include install dir as in: 
https://github.com/borglab/gtsam/commit/903e7de15f2e24f1807c63b02c408ca97c3aee9f
Comment with your opinion?


dellaert:
So, I tried something, with help from gemini. Seems to work.

PS, an alternative is to fully *remove* GeographicLib. It's vendored in (and adds complexity) for basically three tests and was only used in a side project, so we could take the stance that - if you need it, install it.

dellaert:
Actually, windows still does not work. Unless I get feedback on why to keep it, I will go with the “remove” option tomorrow. Will save 11MB in cloning, as well.

jlblancoc:
I can't see the error in Windows CI, only cancelled pipelines (?). 

But anyway, if it's not needed by any core functionality, it's probably easier to "let it go" as vendored in; as years go by, I get more and more convinced that we should try to keep libraries as simple as possible... :thinking:

dellaert:
Thanks @jlblancoc - I think I will do just that. It will save 11MB as well.

PR #2134, opened 2025-05-14
Wrap DiscreteMarginals

varunagrawal:
I also separated `DiscreteMarginals` into a .h and .cpp file to help with compilation since the implementation details won't be copied everywhere.

Additionally I added a print method.


dellaert (approved):
LGTM but CI fails

PR #2135, opened 2025-05-15
Hybrid Improvements

varunagrawal:
1. Add a factor with the Dead Mode Removal marginal threshold when a discrete variable which was removed is added back again. This helps to get back information which improves the performance in later steps by allowing for Dead Mode Removal to operate better.
2. Update `TableFactor` to only record probabilities above 1e-11. This helps keep the TableFactor smaller and improve sparsity.
3. Added `DiscreteBayesNet::joint` as a new method since it is a common operation.

Here is an illustrative example:
Say we have a HybridBayesNet with the conditionals
$BN = P(X(0) | M(0))P(M(0))$
where `X(0)` is a continuous variable and `M(0)` is a discrete variable. Assuming `P(M(0))` was `0.1, 0.9`, then DMR (with threshold 0.85) would remove this conditional and assign `M(0)=1`.

In the next update step, we add the factor $\phi(X(1), M(0))$, and the discrete measurement $\phi(M(0))$. Note that we have reintroduced `M(0)` again and there is new information.
From our previous step, `M(0) = 1`, but the new factors might make it such that `M(0)=0` is the better assignment.
The way I handle that is to reintroduce `P(M(0)) = {0.15, 0.85}` so we get the graph:
$FG = \phi(X(1), M(0)) \phi(M(0)) P(M(0))$.
This way, when we eliminate the information from the previous and current steps are fused giving us a better estimate of `M(0)`.

### DIFF
--- a/gtsam/discrete/DiscreteBayesNet.cpp
+++ b/gtsam/discrete/DiscreteBayesNet.cpp
@@ -71,7 +71,7 @@ DiscreteValues DiscreteBayesNet::sample(DiscreteValues result) const {
 
 /* ************************************************************************* */
 // The implementation is: build the entire joint into one factor and then prune.
-// NOTE: This can be quite expensive *unless* the factors have already
+// NOTE(Frank): This can be quite expensive *unless* the factors have already
 // been pruned before. Another, possibly faster approach is branch and bound
 // search to find the K-best leaves and then create a single pruned conditional.
 DiscreteBayesNet DiscreteBayesNet::prune(
--- a/gtsam/discrete/DiscreteBayesNet.cpp
+++ b/gtsam/discrete/DiscreteBayesNet.cpp
@@ -71,16 +71,14 @@ DiscreteValues DiscreteBayesNet::sample(DiscreteValues result) const {
 
 /* ************************************************************************* */
 // The implementation is: build the entire joint into one factor and then prune.
-// TODO(Frank): This can be quite expensive *unless* the factors have already
+// NOTE: This can be quite expensive *unless* the factors have already
 // been pruned before. Another, possibly faster approach is branch and bound
 // search to find the K-best leaves and then create a single pruned conditional.
 DiscreteBayesNet DiscreteBayesNet::prune(
     size_t maxNrLeaves, const std::optional<double>& marginalThreshold,
     DiscreteValues* fixedValues) const {
   // Multiply into one big conditional. NOTE: possibly quite expensive.
-  DiscreteConditional joint;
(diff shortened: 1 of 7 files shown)


gtsam/discrete/DiscreteBayesNet.cpp
```diff
@@ -71,16 +71,14 @@ DiscreteValues DiscreteBayesNet::sample(DiscreteValues result) const {
 
 /* ************************************************************************* */
 // The implementation is: build the entire joint into one factor and then prune.
-// TODO(Frank): This can be quite expensive *unless* the factors have already
+// NOTE: This can be quite expensive *unless* the factors have already
```
dellaert: Please add my name back. For any notes/to dos we should have a name.

gtsam/hybrid/HybridSmoother.cpp
```diff
@@ -86,13 +86,28 @@ Ordering HybridSmoother::maybeComputeOrdering(
...
+    const HybridGaussianFactorGraph &graph,
     const HybridGaussianFactorGraph &newFactors) {
-  for (Key key : newFactors.discreteKeySet()) {
+  // Initialize graph
+  HybridGaussianFactorGraph updatedGraph(graph);
+
+  for (DiscreteKey dkey : newFactors.discreteKeys()) {
+    Key key = dkey.first;
     if (fixedValues_.find(key) != fixedValues_.end()) {
+      // Add corresponding discrete factor to reintroduce the information
```
dellaert, line 98: I don’t understand this.

dellaert (commented):
The CI fails. That’s probably easy to fix, but a bigger problem is that I don’t really understand what you’re saying about “re-introducing information”. The whole point of dead mode removal is getting rid of information, so I think this needs more explanation. Maybe you can give an example in PR comment that is also implemented in a unit test.

gtsam/hybrid/HybridSmoother.cpp
varunagrawal: I updated the PR comment. I can add a unit test later if that's okay? I'll put it in as a TODO.

dellaert:
OK but then I guess you have to keep track of which variables were removed? That can be a pretty long list. And where do you keep this list? Or maybe I’m still not understanding.

varunagrawal:
We keep track of those variables in a DiscreteValues object called fixedValues_. This is why this PR was pretty straightforward once I figured out the issue. :)

dellaert (approved):
Ok

PR #2136, opened 2025-05-15
Discrete Sampling

varunagrawal:
This PR wraps `mt19937_64` and exposes the `rng` parameter in sampling.
Updated `DiscreteBayesNet` to accept an rng, and added tests to show it working.

PR #2137, opened 2025-05-16
Update `sample` methods

varunagrawal:
Using lessons learned in #2136, I've updated the sampling methods to use default parameters for the PRNG. This not only reduces code duplication, but is also fully backwards-compatible.

I use `nullptr` as the default parameter since that is idiomatic in C++.

### DIFF
--- a/gtsam/hybrid/HybridBayesNet.h
+++ b/gtsam/hybrid/HybridBayesNet.h
@@ -181,10 +181,11 @@ class GTSAM_EXPORT HybridBayesNet : public BayesNet<HybridConditional> {
    *   auto sample = bn.sample(given, &rng);
    *
    * @param given Values of missing variables.
-   * @param rng The pseudo-random number generator.
+   * @param rng The optional pseudo-random number generator.
    * @return HybridValues
    */
-  HybridValues sample(const HybridValues &given, std::mt19937_64 *rng) const;
+  HybridValues sample(const HybridValues &given,
+                      std::mt19937_64 *rng = nullptr) const;
 
   /**
    * @brief Sample using ancestral sampling.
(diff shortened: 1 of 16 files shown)


gtsam/hybrid/HybridBayesNet.h
```diff
@@ -181,10 +181,11 @@ class GTSAM_EXPORT HybridBayesNet : public BayesNet<HybridConditional> {
    *   auto sample = bn.sample(given, &rng);
    *
    * @param given Values of missing variables.
-   * @param rng The pseudo-random number generator.
+   * @param rng The optional pseudo-random number generator.
    * @return HybridValues
    */
-  HybridValues sample(const HybridValues &given, std::mt19937_64 *rng) const;
+  HybridValues sample(const HybridValues &given,
```
dellaert, line 187: What happens when nullptr? Seems different from yesterday's PR where we pass in a kSomething

gtsam/linear/GaussianConditional.cpp
```diff
@@ -347,6 +347,10 @@ namespace gtsam {
 
     VectorValues solution = solve(parentsValues);
     Key key = firstFrontalKey();
+
+    // Check if rng is nullptr, then assign default
+    rng = (rng == nullptr) ? &kRandomNumberGenerator : rng;
```
dellaert, line 350: Is this a local or a global default rng?

dellaert (commented):
An improvement, but maybe inconsistent now between various files? I think it should all be nullptr default or all something else, no? And we should clarify if the default will then be a file-specific local rng or a global rng.

gtsam/linear/GaussianConditional.cpp
varunagrawal: It's not a global default. There is one definition in `GaussianConditional.cpp` and one in `DiscreteConditional.h`. I think we should make it a global default so code like `ShonanAveraging` can also leverage it (they all use the same seed of 42).

gtsam/hybrid/HybridBayesNet.h
varunagrawal: Passing in a static variable address as a default argument is
1. Not idiomatic C++ which recommends using a `nullptr` as the default for a raw pointer.
2. Did not play well with the python wrapper due to memory allocation.

Whenever a `sample` method gets a nullptr for the `rng`, it keeps passing it along until we actually use the rng (e.g. in `Sample::sample`). At that point, I have a simple check:
```cpp
rng = (rng == nullptr) ? &kRandomNumberGenerator : rng;
```

This makes the code really clean since we don't have to worry about the rng pointer until we actually need it.

dellaert: Ok. And yesterday’s changes are now also defaulting to null ptr?

varunagrawal: Yes, I needed to do that to get the CI passing.

dellaert (approved):

dellaert:
Rather than renaming, (i don’t like the new name so much) you can put each one in a cpp file in an anonymous namespace

varunagrawal:
I was hoping to follow your suggestion, perhaps in another PR.

PR #2150, opened 2025-05-27
Updated wrapper for HybridValues

dellaert:
- Added `insertNonlinear`
- Updated wrapper for HybridValues and added docs.
- Links in Markdown currently don't work yet, should work after #2120 lands.


dellaert:
Merging after CI passed.

Issue #2146, opened 2025-05-26
Some issues in hybrid


dellaert:
## Description

Two small issues I encountered while helping to document hybrid:
- choose in a HybrdBayesTree yields Gaussian cliques that are empty, presumably from discrete only cliques
- prune in a HybridBayesNet with max_leaves == 1 should convert to entirely non-hybrid, I think

## Steps to reproduce

1. Try display in HybridBayesTree.ipynb 
2. Try removing marginal threshold in HybridBayesNet.ipynb

## Some other issues:
- HybridValues.ipynb: is `retract` working as expected in the last cell?
- Best way to print DiscreteValues? `PrintDiscreteValues` doesn't seem to print anything in Jupyter, and print(DiscreteValues) doesn't have useful output.

## Additional information

Notebooks are still in unmerged docs-hybrid branch at the moment

varunagrawal:
#2170 should have resolved part 1. For part 2, I think it should still be hybrid since 1 leaf does not mean we have made a choice on the discrete assignment. Another case is when more than one leaf in the DiscreteConditional has the same value so they get merged together. It is now 1 leaf but not a specific assignment.

`PrintDiscreteValues` works in Jupyter, but it prints out of order due to [different buffers for C++ and Python](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html).

varunagrawal:
I've fixed `PrintDiscreteValues` in #2120.

dellaert:
Part 2 still seems an issue: does one leaf not correspond to one assignment? I’m probably missing a subtlety here.

varunagrawal:
No. Think of a decision tree with one discrete key and probabilities {0.5, 0.5}. Due to merging, it will have 1 leaf, but that doesn't mean it has one assignment.

dellaert:
Take your word for it.

PR #2152, opened 2025-05-28
Wrap HybridBayesNet and HybridConditional methods

dellaert:
Second step in more careful wrapper changes that were part of #2120 and thoroughly unit test hybrid.
This adds HybridBayesNet and HybridCondtitional notebooks, and changed to wrapper needed to make this happen.

@varunagrawal take a close look at TODOs in test_HybridBayesNet.py and also see issue #2146. I am not convinced the behavior for prune is how we want it - so we should meet and discuss. If something needs fixing we can fix it in this PR or a follow-up PR.

This does make a choice to expose AlgebraicDecisionTreeKey and adds AssignmentKey in hybrid.h.


dellaert:
@varunagrawal added a prune test in DiscreteBN

varunagrawal (approved):
LGTM

varunagrawal:
Merging so I can look into the TODOs more.

PR #2153, opened 2025-05-28
Fix paths in notebooks

dellaert:
Noticed that colab button was not working.
Added missing `gtsam` to colab button source for 3 notebooks.

Tested: opened one of them in colab, tested it.


ktk-sdk (approved):
Amazing beautiful fantastic change!

PR #2155, opened 2025-05-31
Fixed SLAM Example link in Pose2 Doc

CodeXTL:
Noticed that the link to the SLAM example in the Pose2 documentation was not working.
Fixed the path to direct to `Pose2SLAMExample.ipynb`.

Test: opened link in new tab, no more file not found error.


dellaert (approved):
Thanks!!

Issue #2145, opened 2025-05-23
Build fails with `GTSAM_TANGENT_PREINTEGRATION` off


p-zach:
## Description

When building with `GTSAM_TANGENT_PREINTEGRATION` off, the build fails because `preintegrated()` is not a member of `PreintegratedImuMeasurements`, which the compiler discovers upon creating `navigation.cpp` from `navigation.i`. It is not a member because it doesn't exist in `ManifoldPreintegration`, which `PreintegratedImuMeasurements` inherits from instead of `TangentPreintegration` (where `preintegrated()` does exist) when `GTSAM_TANGENT_PREINTEGRATION` is off.

## Steps to reproduce

1. Set up GTSAM to build Python wrapper
2. Turn off `GTSAM_TANGENT_PREINTEGRATION`
3. Try to build

## Expected behavior

The build succeeds.

## Environment

- Windows 10
- Python 3.11
- GTSAM 4.3a0

## Additional information

My current solution is commenting out the `preintegrated()` line:
https://github.com/borglab/gtsam/blob/ce5f93f4a0c3c59bd1ff0604f64fd41c279a41ce/gtsam/navigation/navigation.i#L168
Permanent solutions would be either providing a real or dummy implementation in `ManifoldPreintegration` or officially removing `preintegrated()` from `navigation.i`.

dellaert:
I think, with the plethora of new integration methods coming, we should actually use a template argument to decide between integration methods. I'll look into this.

PR #2157, opened 2025-06-01
Add template for IMU backend selection to move away from compiler flag

p-zach:
## Description

This PR refactors `ImuFactor`, `ImuFactor2`, `CombinedImuFactor`, and their associated `Preintegrated*Measurements` classes to use a template-based approach for selecting the underlying preintegration method (Manifold or Tangent). This change allows  users to choose a preintegration backend without requiring GTSAM to be recompiled with different flags, while maintaining full backward compatibility.

## Motivation & Background

Previously, the choice between `ManifoldPreintegration` and `TangentPreintegration` was controlled by the `GTSAM_TANGENT_PREINTEGRATION` preprocessor flag at GTSAM compile time.

This refactor addresses this limitation by:
1.  Introducing templated versions of the IMU factor and preintegrated measurements classes
2.  Parameterizing these templates by the preintegration backend type via the preintegrated measurements class
3.  Maintaining backward compatibility by providing `typedef`s for the original class names (e.g., `ImuFactor`) which default to the backend selected by the `GTSAM_TANGENT_PREINTEGRATION` flag (via `DefaultPreintegrationBackend`).

## Key Changes

*   **Templated Classes:**
    *   `PreintegratedImuMeasurements` -> `PreintegratedImuMeasurementsT<PreintegrationBackend>`
    *   `PreintegratedCombinedMeasurements` -> `PreintegratedCombinedMeasurementsT<PreintegrationBackend>`
    *   `ImuFactor` -> `ImuFactorT<PIMType_ = PreintegratedImuMeasurements>`
    *   `ImuFactor2` -> `ImuFactor2T<PIMType_ = PreintegratedImuMeasurements>`
    *   `CombinedImuFactor` -> `CombinedImuFactorT<PIMType_ = PreintegratedCombinedMeasurements>`
*   **Backward Compatibility Typedefs:**
    *   `typedef PreintegratedImuMeasurementsT<DefaultPreintegrationBackend> PreintegratedImuMeasurements;` (and similar for Combined)
    *   `typedef ImuFactorT<> ImuFactor;` (and similar for ImuFactor2, CombinedImuFactor), which use the default `PIMType_` that in turn uses `DefaultPreintegrationBackend`.
*   **SFINAE for Backend-Specific Methods:** Methods like `mergeWith` and static `Merge` functions, previously guarded by `#ifdef GTSAM_TANGENT_PREINTEGRATION`, are now enabled using SFINAE (substitution failure is not an error, a C++ paradigm) only for `TangentPreintegration`-based instantiations. These are also now defined in the header files instead of the .cpp files.
*   **Explicit Instantiations:** Explicit template instantiations for both `ManifoldPreintegration` and `TangentPreintegration` backends are provided in the `.cpp` files. This ensures that:
    *   Implementations can reside in `.cpp` files.
    *   Commonly used instantiations are pre-compiled and exported from the GTSAM library.
    *   Users linking against GTSAM can use these common instantiations without needing the full template method definitions.
*   **Robustness for Dependent Names:** `this->` is used consistently when accessing members of templated base classes (`PreintegrationBackend`) to ensure correct name lookup in dependent contexts, particularly for portability across compilers.

## How to Use

**1. Backward Compatible Usage (no code change needed):**
Existing code using `gtsam::ImuFactor`, `gtsam::PreintegratedImuMeasurements`, etc., will continue to work as before. The preintegration backend will be determined by how GTSAM was compiled (i.e., the `GTSAM_TANGENT_PREINTEGRATION` flag).

```cpp
// Old code - still works, uses GTSAM's compiled default backend
gtsam::PreintegratedImuMeasurements pim(params, bias);
// ... integrate measurements ...
gtsam::ImuFactor factor(keyPose1, keyVel1, keyPose2, keyVel2, keyBias, pim);
```

**2. Explicit Backend Selection (new):**
Users can now explicitly choose the preintegration backend at their own compile time, regardless of GTSAM's compilation flags.

**Example: Explicitly using `ManifoldPreintegration`**
```cpp
#include <gtsam/navigation/ImuFactor.h>
#include <gtsam/navigation/ManifoldPreintegration.h>

// Define the PIM type using ManifoldPreintegration as the backend
typedef gtsam::PreintegratedImuMeasurementsT<gtsam::ManifoldPreintegration> MyManifoldPim;

// Create an ImuFactor that uses this PIM type
typedef gtsam::ImuFactorT<MyManifoldPim> MyManifoldImuFactor;

// --- Usage ---
std::shared_ptr<gtsam::PreintegrationParams> params = /* ... initialize ... */;
gtsam::imuBias::ConstantBias bias; // Or your estimated bias

MyManifoldPim pim(params, bias);
// ... pim.integrateMeasurement(...) ...

gtsam::Key keyPose1(1), keyVel1(2), keyPose2(3), keyVel2(4), keyBias(5);
MyManifoldImuFactor factor(keyPose1, keyVel1, keyPose2, keyVel2, keyBias, pim);
```

**Example: Explicitly using `TangentPreintegration`**
```cpp
#include <gtsam/navigation/ImuFactor.h>
#include <gtsam/navigation/TangentPreintegration.h>

// Define the PIM type using TangentPreintegration as the backend
typedef gtsam::PreintegratedImuMeasurementsT<gtsam::TangentPreintegration> MyTangentPim;

// Create an ImuFactor that uses this PIM type
typedef gtsam::ImuFactorT<MyTangentPim> MyTangentImuFactor;

// --- Usage (similar to above) ---
std::shared_ptr<gtsam::PreintegrationParams> params = /* ... initialize ... */;
gtsam::imuBias::ConstantBias bias;

MyTangentPim pim(params, bias);
// ... pim.integrateMeasurement(...) ...

gtsam::Key keyPose1(1), keyVel1(2), keyPose2(3), keyVel2(4), keyBias(5);
MyTangentImuFactor factor(keyPose1, keyVel1, keyPose2, keyVel2, keyBias, pim);

// If using TangentPreintegration, Merge methods are available:
// MyTangentPim merged_pim = MyTangentImuFactor::Merge(pim1, pim2);
```

## Note

Code changes generated by Gemini; I went over them with a fine-toothed comb and tested. PR description also partially generated by Gemini for the sake of full coverage of changes.

Please suggest changes, as this is a pretty comprehensive overhaul @varunagrawal @ProfFan @dellaert

### DIFF
--- a/gtsam/navigation/CombinedImuFactor.cpp
+++ b/gtsam/navigation/CombinedImuFactor.cpp
@@ -21,6 +21,8 @@
  **/
 
 #include <gtsam/navigation/CombinedImuFactor.h>
+#include <gtsam/navigation/ManifoldPreintegration.h>
+#include <gtsam/navigation/TangentPreintegration.h>
 #if GTSAM_ENABLE_BOOST_SERIALIZATION
 #include <boost/serialization/export.hpp>
 #endif
@@ -58,33 +60,37 @@ bool PreintegrationCombinedParams::equals(const PreintegratedRotationParams& oth
 }
 
 //------------------------------------------------------------------------------
-// Inner class PreintegratedCombinedMeasurements
+// Inner class PreintegratedCombinedMeasurementsT
 //------------------------------------------------------------------------------
-void PreintegratedCombinedMeasurements::print(const string& s) const {
-  PreintegrationType::print(s);
+template <class PreintegrationBackend>
+void PreintegratedCombinedMeasurementsT<PreintegrationBackend>::print(const string& s) const {
+  PreintegrationBackend::print(s);
   cout << "  preintMeasCov [ " << preintMeasCov_ << " ]" << endl;
 }
 
 //------------------------------------------------------------------------------
-bool PreintegratedCombinedMeasurements::equals(
-    const PreintegratedCombinedMeasurements& other, double tol) const {
-  return PreintegrationType::equals(other, tol)
+template <class PreintegrationBackend>
+bool PreintegratedCombinedMeasurementsT<PreintegrationBackend>::equals(
+    const PreintegratedCombinedMeasurementsT<PreintegrationBackend>& other, double tol) const {
+  return PreintegrationBackend::equals(other, tol)
       && equal_with_abs_tol(preintMeasCov_, other.preintMeasCov_, tol);
 }
 
 //------------------------------------------------------------------------------
-void PreintegratedCombinedMeasurements::resetIntegration() {
+template <class PreintegrationBackend>
+void PreintegratedCombinedMeasurementsT<PreintegrationBackend>::resetIntegration() {
   // Base class method to reset the preintegrated measurements
-  PreintegrationType::resetIntegration();
+  PreintegrationBackend::resetIntegration();
   preintMeasCov_.setZero();
 }
 
 //------------------------------------------------------------------------------
-void PreintegratedCombinedMeasurements::resetIntegration(
+template <class PreintegrationBackend>
+void PreintegratedCombinedMeasurementsT<PreintegrationBackend>::resetIntegration(
     const gtsam::Matrix6& Q_init) {
   // Base class method to reset the preintegrated measurements
-  PreintegrationType::resetIntegration();
-  p().biasAccOmegaInt = Q_init;
+  PreintegrationBackend::resetIntegration();
+  this->p().biasAccOmegaInt = Q_init;
   preintMeasCov_.setZero();
 }
 
@@ -103,7 +109,8 @@ void PreintegratedCombinedMeasurements::resetIntegration(
 #define D_g_g(H) (H)->block<3,3>(12,12)
 
 //------------------------------------------------------------------------------
-void PreintegratedCombinedMeasurements::integrateMeasurement(
+template <class PreintegrationBackend>
+void PreintegratedCombinedMeasurementsT<PreintegrationBackend>::integrateMeasurement(
     const Vector3& measuredAcc, const Vector3& measuredOmega, double dt) {
   if (dt <= 0) {
     throw std::runtime_error(
@@ -113,7 +120,7 @@ void PreintegratedCombinedMeasurements::integrateMeasurement(
   // Update preintegrated measurements.
   Matrix9 A; // Jacobian wrt preintegrated measurements without bias (df/dx)
   Matrix93 B, C;  // Jacobian of state wrpt accel bias and omega bias respectively.
-  PreintegrationType::update(measuredAcc, measuredOmega, dt, &A, &B, &C);
+  PreintegrationBackend::update(measuredAcc, measuredOmega, dt, &A, &B, &C);
 
   // Update preintegrated measurements covariance: as in [2] we consider a first
   // order propagation that can be seen as a prediction phase in an EKF
@@ -144,10 +151,10 @@ void PreintegratedCombinedMeasurements::integrateMeasurement(
 
   // propagate uncertainty
   // TODO(frank): use noiseModel routine so we can have arbitrary noise models.
-  const Matrix3& aCov = p().accelerometerCovariance;
-  const Matrix3& wCov = p().gyroscopeCovariance;
-  const Matrix3& iCov = p().integrationCovariance;
-  const Matrix6& bInitCov = p().biasAccOmegaInt;
+  const Matrix3& aCov = this->p().accelerometerCovariance;
+  const Matrix3& wCov = this->p().gyroscopeCovariance;
+  const Matrix3& iCov = this->p().integrationCovariance;
+  const Matrix6& bInitCov = this->p().biasAccOmegaInt;
 
   // first order uncertainty propagation
   // Optimized matrix mult: (1/dt) * G * measurementCovariance * G.transpose()
@@ -174,8 +181,8 @@ void PreintegratedCombinedMeasurements::integrateMeasurement(
       (vel_H_acc * (aCov / dt) * vel_H_acc.transpose())  //
       + (vel_H_biasAccInit * bInitCov11 * vel_H_biasAccInit.transpose());
 
-  D_a_a(&G_measCov_Gt) = dt * p().biasAccCovariance;
-  D_g_g(&G_measCov_Gt) = dt * p().biasOmegaCovariance;
+  D_a_a(&G_measCov_Gt) = dt * this->p().biasAccCovariance;
+  D_g_g(&G_measCov_Gt) = dt * this->p().biasOmegaCovariance;
 
   // OFF BLOCK DIAGONAL TERMS
   D_R_t(&G_measCov_Gt) =
@@ -197,23 +204,10 @@ void PreintegratedCombinedMeasurements::integrateMeasurement(
 }
 
 //------------------------------------------------------------------------------
-// CombinedImuFactor methods
+// CombinedImuFactorT methods
 //------------------------------------------------------------------------------
-CombinedImuFactor::CombinedImuFactor(Key pose_i, Key vel_i, Key pose_j,
-    Key vel_j, Key bias_i, Key bias_j,
-    const PreintegratedCombinedMeasurements& pim) :
-    Base(noiseModel::Gaussian::Covariance(pim.preintMeasCov_), pose_i, vel_i,
-        pose_j, vel_j, bias_i, bias_j), _PIM_(pim) {
-}
-
-//------------------------------------------------------------------------------
-gtsam::NonlinearFactor::shared_ptr CombinedImuFactor::clone() const {
-  return std::static_pointer_cast<gtsam::NonlinearFactor>(
-      gtsam::NonlinearFactor::shared_ptr(new This(*this)));
-}
-
-//------------------------------------------------------------------------------
-void CombinedImuFactor::print(const string& s,
+template <class PIMType_>
+void CombinedImuFactorT<PIMType_>::print(const string& s,
     const KeyFormatter& keyFormatter) const {
   cout << (s.empty() ? s : s + "\n") << "CombinedImuFactor("
        << keyFormatter(this->key<1>()) << "," << keyFormatter(this->key<2>()) << ","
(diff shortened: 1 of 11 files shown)


p-zach:
Related to https://github.com/borglab/gtsam/issues/2145

> I think, with the plethora of new integration methods coming, we should actually use a template argument to decide between integration methods.

p-zach:
Will look into solution for Linux & Mac failures. I guess this shows Windows is superior...

dellaert:
Excellent!
Quick Q: why would I have to typedef the IMU factor? Does the type of PIM matter? I thought they had a common base class.

gtsam/navigation/ImuFactor.h
```diff
@@ -65,10 +68,11 @@ typedef ManifoldPreintegration PreintegrationType;
  *
  * @ingroup navigation
  */
-class GTSAM_EXPORT PreintegratedImuMeasurements: public PreintegrationType {
-
-  friend class ImuFactor;
-  friend class ImuFactor2;
+template <class PreintegrationBackend>
+class GTSAM_EXPORT PreintegratedImuMeasurementsT: public PreintegrationBackend {
```
dellaert: Is it not simpler to just do
```
template <class PreintegrationType=tangentPreintegration>
class PreintegratedImuMeasurements : public PreintegrationType {
```
Or does that not work?

gtsam/navigation/CombinedImuFactor.cpp
```diff
@@ -296,10 +292,40 @@ Vector CombinedImuFactor::evaluateError(const Pose3& pose_i,
...
+// Instantiate operator<<
+template GTSAM_EXPORT std::ostream& operator<<<PreintegratedCombinedMeasurementsT<ManifoldPreintegration>>(
+    std::ostream& os, const CombinedImuFactorT<PreintegratedCombinedMeasurementsT<ManifoldPreintegration>>& f);
+template GTSAM_EXPORT std::ostream& operator<<<PreintegratedCombinedMeasurementsT<TangentPreintegration>>(
+    std::ostream& os, const CombinedImuFactorT<PreintegratedCombinedMeasurementsT<TangentPreintegration>>& f);
+
+
 }  // namespace gtsam
+
+// Undefine macros to prevent scope leakage
```
dellaert: Don’t understand these, and if needed they’re for a different PR I think

varunagrawal:
@p-zach I'll look at this on Thursday if that's okay? I need to finish up some stuff before then.

gtsam/navigation/ImuFactor.h
p-zach: That would also work. The format `...T` is a pattern used when refactoring a class from concrete to templated, since changing how the concrete type works might cause issues elsewhere in the code in certain scenarios. If you prefer this version I can make that work

p-zach:
> why would I have to typedef the IMU factor? Does the type of PIM matter? I thought they had a common base class.

They do, and it wouldn't need to be templated but for the `Merge` functions. Since we can't control the declaration of those functions by compiler flag anymore, it has to be with SFINAE, so it needs a template.

gtsam/navigation/ImuFactor.h
dellaert: No, I understand now.

dellaert:
> > why would I have to typedef the IMU factor? Does the type of PIM matter? I thought they had a common base class.
> 
> They do, and it wouldn't need to be templated but for the `Merge` functions. Since we can't control the declaration of those functions by compiler flag anymore, it has to be with SFINAE, so it needs a template.

Ah, so, if we did not use SFINAE, but just threw “not implemented”, it would work without templating IMUFactor? I’d prefer that, for sure. It would keep this PR - and disruption- small.

dellaert:
@varunagrawal , @ProfFan i’ll work with porter on this.

p-zach:
> Ah, so, if we did not use SFINAE, but just threw “not implemented”, it would work without templating IMUFactor?

Well, that would work, but then the user would lose the ability to choose the PIM backend of IMUFactor--it would be determined by whatever the default is, based on the compiler flag. So that would defeat the purpose. (I forgot about that in my above response, my bad)

dellaert:
> > Ah, so, if we did not use SFINAE, but just threw “not implemented”, it would work without templating IMUFactor?
> 
> Well, that would work, but then the user would lose the ability to choose the PIM backend of IMUFactor--it would be determined by whatever the default is, based on the compiler flag. So that would defeat the purpose. (I forgot about that in my above response, my bad)

Unless the IMUFactor takes a base class ref, right? Then it can accept any PIM derived from the base PIM class

p-zach:
Hmm. Dependency injection might work, but I have a feeling it might cause other problems. It's not immediately clear to me how to make that backwards compatible, and for e.g. the static Merge method I think there would be awkward issues. I will take a look at that approach and get back to you.

varunagrawal:
Cool

p-zach:
Fixed those build errors, just difference in compiler strictness for MSVC vs. GCC/Clang.

Dependency injection would work, but it wouldn't be as easy as just passing in a `PreintegrationBase` and being done with it. `ImuFactor` uses `preintMeasCov()` which is only within `Preintegrated...Measurements`. So we'd need to make a new interface or something for `ImuFactor` to have access to everything it needs. We would lose the compile-time safety for the `Merge` methods, which is fine, but something to consider.

p-zach:
@dellaert I put together a PR showing the effect of de-templating IMU factor #2166. C++ interfaces are finicky--I think it is not what we want.

gtsam/navigation/ImuFactor.cpp
```diff
@@ -30,28 +32,32 @@ namespace gtsam {
 using namespace std;
 
 //------------------------------------------------------------------------------
-// Inner class PreintegratedImuMeasurements
+// Inner class PreintegratedImuMeasurementsT
 //------------------------------------------------------------------------------
-void PreintegratedImuMeasurements::print(const string& s) const {
-  PreintegrationType::print(s);
+template <class PreintegrationBackend>
```
dellaert: Prefer keeping the name `PreintegrationType`

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -77,45 +77,44 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurements
    */
   Eigen::Matrix<double, 15, 15> preintMeasCov_;
 
-  friend class CombinedImuFactor;
+  template <class PIMType_> friend class CombinedImuFactorT;
```
dellaert: Could we just use `PIM` instead of PIMType_ ?

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -300,17 +302,24 @@ class GTSAM_EXPORT CombinedImuFactor
  public:
   GTSAM_MAKE_ALIGNED_OPERATOR_NEW
 };
-// class CombinedImuFactor
+// class CombinedImuFactorT
+
+// For backward compatibility:
+typedef CombinedImuFactorT<> CombinedImuFactor;
```
dellaert: For these default typedefs, let’s use `using`

gtsam/navigation/ImuFactor.cpp
```diff
@@ -254,6 +186,29 @@ Vector ImuFactor2::evaluateError(const NavState& state_i,
 }
 
 //------------------------------------------------------------------------------
+// Explicit instantiations
```
dellaert, line 189: Excellent!

gtsam/navigation/ImuFactor.h
```diff
@@ -27,12 +27,15 @@
...
 #include <gtsam/base/debug.h>
 
+#include <type_traits> // For std::is_same, std::enable_if
+ 
 namespace gtsam {
 
+// Determine default preintegration backend
 #ifdef GTSAM_TANGENT_PREINTEGRATION
-typedef TangentPreintegration PreintegrationType;
+typedef TangentPreintegration DefaultPreintegrationBackend;
```
dellaert: DefaultPreintegrationType

dellaert (approved):
Awesome. Some name change and style requests, but otherwise a fantastic PR. I will merge after those edits have been made.

Thanks also for exploring the other option - you were right that this is better.

gtsam/navigation/CombinedImuFactor.h
p-zach: Yes. Do you also want that for inside the factors? Then we would have `PIM _PIM_`.

p-zach: On second look, this makes sense

p-zach:
Wow, I was really quite puzzled at this not passing checks when I hadn't really changed anything... just realized we have new ones! However it does raise the problem that I hadn't thought of -- are we ok with breaking code if users were relying on `PreintegrationType` for the sake of renaming it to the more clear/intentional `DefaultPreintegrationType`? If not, we can keep `PreintegrationType` as representing the default and change the name that PIMs are templated on (e.g. back to `PreintegrationBackend`, or something else). Let me know your thoughts @dellaert and I'll make the change and we can get this merged. I will also add the relevant info to the docs in this PR.

dellaert:
> Wow, I was really quite puzzled at this not passing checks when I hadn't really changed anything... just realized we have new ones! However it does raise the problem that I hadn't thought of -- are we ok with breaking code if users were relying on `PreintegrationType` for the sake of renaming it to the more clear/intentional `DefaultPreintegrationType`? If not, we can keep `PreintegrationType` as representing the default and change the name that PIMs are templated on (e.g. back to `PreintegrationBackend`, or something else). Let me know your thoughts @dellaert and I'll make the change and we can get this merged. I will also add the relevant info to the docs in this PR.

I think that the current naming is more clear, and this CI failure is an atypical/rare case and easily fixed. So, I’m ok with this backwards compatibility issue.

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -206,31 +208,28 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurements
...
  public:
  private:
-  typedef CombinedImuFactor This;
+  typedef CombinedImuFactorT<PIMType_> This;
   typedef NoiseModelFactorN<Pose3, Vector3, Pose3, Vector3,
                             imuBias::ConstantBias, imuBias::ConstantBias>
       Base;
 
-  PreintegratedCombinedMeasurements _PIM_;
+  PIMType_ _PIM_;
```
dellaert: I would rename to 
```
PIM pim_;
```

gtsam/navigation/ImuFactor.h
```diff
@@ -167,30 +184,28 @@ class GTSAM_EXPORT PreintegratedImuMeasurements: public PreintegrationType {
...
     imuBias::ConstantBias> {
 private:
 
-  typedef ImuFactor This;
+  typedef ImuFactorT<PIMType_> This;
   typedef NoiseModelFactorN<Pose3, Vector3, Pose3, Vector3,
       imuBias::ConstantBias> Base;
 
-  PreintegratedImuMeasurements _PIM_;
+  PIMType_ _PIM_;
```
dellaert: Same

dellaert:
Ready to merge?

p-zach:
Yes. Wanted to make sure all the tests passed since there was a weird segfault last time.

dellaert:
Thanks @p-zach !!!

PR #2160, opened 2025-06-04
Created jupyter notebook for the LQR example.

zhangzdd:
As part of the OSPO internship, I created the notebook for a LQR example based on the legacy website post (https://gtsam.org/2019/11/07/lqr-control.html).


dellaert:
I think we discussed some changes in our meeting?

zhangzdd:
I have made several refinements to the previous version, including:
Added much richer mathematical explanation text to the code, providing more details on LQR problem itself and how GTSAM views the problem. Added link to colab runtime. Added visualization of the factor graph and Bayes net (obtained from different elimination ordering). Added a title.

dellaert:
I still think the ordering should be reversed and match the blog post, to indicate the DAG encodes policies P(u|x)

dellaert:
The ordering in cell 20 is still wrong :-/

dellaert (commented):
I know you fixed the ordering in our meeting but maybe you did not commit/push yet?

dellaert (approved):

Issue #2089, opened 2025-04-12
Build fails on 32-bit architectures


dkogan:
Hello.

The Debian release of gtsam failed to rebuild on the 32-bit architectures: https://buildd.debian.org/status/package.php?p=gtsam

Click on "Build-Attempted" to get the build log. It looks like things build, but the tests fail.

Let me know if you need more info. Thanks.

ProfFan:
Seems to be a `typeinfo` related issue. Can't currently determine whether it's in Pybind11 or GTSAM itself.

Could you try current `develop` on the same build environment? Many thanks!

dkogan:
Hi. It's quite a bit of work to try to rebuild on `develop` because there are a lot of patches that were needed for the packaging work, and they will need to be updated to apply to `develop` successfully: https://salsa.debian.org/science-team/gtsam/-/tree/master/debian/patches?ref_type=heads

Getting as many of these as possible merged upstream would be good. Maybe yall already did that, and that's why the patches don't apply anymore.

Is trying with the latest `develop` branch required for you to work on this? I'm fully aware that you don't want to work on anything that's not your bleeding edge, but the new Debian release will be out in a few months, and 4.2 is what we'll ship. Let me know if you are able to debug 4.2. I can also run experiments myself, if you can suggest specific things.

Thanks

ProfFan:
Hi @dkogan, I was very busy (defended my thesis today) but now I have more time on this. I can run some more tests in the following week. By the way, what I mean is not building the debian package, but just building the `develop` version on the same environment as the build machine, including the python wrapper, and test if that works :)

dkogan:
Congrats on the defense! I tried to just build the develop branch on i386 without the debian stuff, and it's not entirely clear what that's doing. Any able to debug this, or can you suggest some things to me about how to debug? Thanks

ProfFan:
I'll try spin up an ARMv7 box to see what is wrong with it

ProfFan:
Just run the test suite on i386/ARM (I think ARM should have better tooling support nowadays) with `develop` (`make python-tests`), and see if that works. Also check if `make check` has any failures. If only Python is failing, most likely it's due to `typeinfo` comparisons as from the log in your Debian build it seems only type conversions are failing.

dkogan:
Hi. I just did this at this commit:

```
commit 42cc0ac922460d506a7bbf3a37a4a0b1e103b79e (HEAD -> develop, origin/develop, origin/HEAD)
[...]

PR #2162, opened 2025-06-05
Fix misuses of `size_t` for `gtsam::Key`

ProfFan:
Should fix issue #2089


dellaert (approved):
Thanks!!!

varunagrawal:
So this PR breaks the Matlab wrapper since during the type check of arguments `is(varargin{1}, 'gtsam.Key')`. It probably would have been better to replace `size_t` with `std::uint64_t` (which is what gtsam::Key is aliased to), to avoid any major changes to the wrapper library.

PR #2164, opened 2025-06-06
Add vcpkg ci

### LINKED ISSUE #672: Using Vcpkg/conan for dependencies and third party libraries
## Feature

use vcpkg/conan instead of system packaging for dependencies and third party libraries

## Motivation

System Packaging have the problem that it can't be updated till next release update of the system, or some custom repo with needed library. On the other hand we can use conan or vcpkg to build these library which is updated regularly, and if certain library build configuration is not suitable, you can edit it and publish the edited  version to be used by other to build your library and its working cross-platform (linux-windows-macos) so one script fit all -which will help in CI-.
## Pitch

I have used vcpkg and it worked for example

```powershell
./vcpkg.exe install tbb metis boost eigen3 suitesparse[metis] ceres --triplet x64-windows
```
but I had to change some linking and includes to work. So if we can use one of these package managers it will remove the need for almost all Find*.cmake files and it will make build easier on windows, linux and macos

talregev:
Add vcpkg ci
Fix #672
continue this PR #1413

### DIFF
--- a/.github/workflows/vcpkg.yml
+++ b/.github/workflows/vcpkg.yml
@@ -1,0 +1,163 @@
+name: vcpkg
+on:
+  pull_request:
+
+# Every time you make a push to your PR, it cancel immediately the previous checks,
+# and start a new one. The other runner will be available more quickly to your PR.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:      
+          - os: windows-latest
+            triplet: x64-windows-release
+            build_type: Release
+            test_target: RUN_TESTS
+            binary_cache: C:\Users\runneradmin\AppData\Local\vcpkg\archives
+            vcpkg_info: C:/vcpkg/installed/vcpkg/info/*
+            python: python
+            # check_constraint_program fail on windows. should fix it for windows.
+            # Remove this excluded test when you fix it for windows.
+            ctest_extra_flags: -E check_constraint_program
+          - os: ubuntu-latest
+            triplet: x64-linux
+            build_type: Release
+            test_target: test
+            binary_cache: /home/runner/.cache/vcpkg/archives
+            vcpkg_info: /usr/local/share/vcpkg/installed/vcpkg/info/*
+            cxxflags: -DCMAKE_CXX_FLAGS="-Wno-error=nonnull -Wno-error=maybe-uninitialized"
+            python: python3
+          - os: macos-latest
+            triplet: arm64-osx
+            build_type: Release
+            test_target: test
+            binary_cache: /Users/runner/.cache/vcpkg/archives
+            vcpkg_info: /usr/local/share/vcpkg/installed/vcpkg/info/*
+            python: python3
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Restore cache dependencies
+        uses: actions/cache/restore@v3
+        with:
+          path: ${{ matrix.binary_cache }}
+          key: ${{ matrix.os }}
+          restore-keys: ${{ matrix.os }}
+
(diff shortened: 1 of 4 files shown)


dellaert:
Adding Dr. Fan to review this

.github/workflows/vcpkg.yml
```diff
@@ -0,0 +1,160 @@
...
+          $VCPKG_INSTALLATION_ROOT/installed/${{ matrix.triplet }}/tools/python3/${{ matrix.python }} -m ensurepip --upgrade
+          $VCPKG_INSTALLATION_ROOT/installed/${{ matrix.triplet }}/tools/python3/${{ matrix.python }} -m pip install -r python/dev_requirements.txt
+
+      - name: cmake config
+        if: success()
+        shell: bash
+        run: |
+          export CL=-openmp:experimental
+
+          # This is due an error in linux:
```
ProfFan, line 126: Which line does this comment relate to? L129?

.github/workflows/vcpkg.yml
```diff
@@ -0,0 +1,160 @@
...
+      matrix:
+        include:      
+          - os: windows-latest
+            triplet: x64-windows-release
+            build_type: Release
+            test_target: RUN_TESTS
+            binary_cache: C:\Users\runneradmin\AppData\Local\vcpkg\archives
+            vcpkg_info: C:/vcpkg/installed/vcpkg/info/*
+            python: python
+            ctest_extra_flags: -E check_constraint_program
```
ProfFan, line 28: Add a TODO before this line saying why this one has to be disabled?

wrap/pybind11/CMakeLists.txt
```diff
@@ -100,7 +100,11 @@ if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
 else()
   set(PYBIND11_MASTER_PROJECT OFF)
   set(pybind11_system SYSTEM)
-  set(_pybind11_findpython_default OFF)
+  if(CMAKE_VERSION VERSION_LESS "3.18")
```
ProfFan: We vendor pybind11 but we don't make changes directly on it. Could you explain why this change is required? If absolutely required this should be submitted upstream or as a VCPKG patch, and we add an option in `wrap` to use system or vcpkg-provided pybind11

ProfFan (changes requested):
Thank you! Mostly looks good to me, just need a few changes :)

.github/workflows/vcpkg.yml
talregev: It related to line L144. for ${{ matrix.cxxflags }}
The flags on linux convert it from warning as error, to warning. I found 2 warning as error like that.
On your regular linux build you supress all the warning, so you don't aware of it.

.github/workflows/vcpkg.yml
talregev: this test is failed on windows. On the regular windows build, the test is not running, so you don't aware it fail on windows. 
I will add a TODO.

wrap/pybind11/CMakeLists.txt
talregev: I agree, I don't suppose to change the wrapper. I notice that after your comment.
I did it to support vcpkg.
I already install pybind via vcpkg, and ideal gtsam should find vcpkg pybind instead this wrapper. 
For now I will disable python tests in vcpkg, and revert this change. 
We can explore python test with vcpkg in other PR.

talregev (commented):
review comment

wrap/pybind11/CMakeLists.txt
ProfFan: Sounds good, Just remove changes in `wrap` and comment out the Python VCPKG tests,  and add a TODO. I'll find some time to update `wrap` to allow external Pybind11.

.github/workflows/vcpkg.yml
talregev: done.

wrap/pybind11/CMakeLists.txt
talregev: done.

ProfFan (approved):
Thank you!

PR #2166, opened 2025-06-07
ImuFactor de-templated

p-zach:
Draft PR showing what the code would look like if ImuFactor switched back to just depending on the passed-in PIM instead of being templated. The result is not great.
- You can't just have a reference to an interface in C++, you need to have a pointer to it (`std::shared_ptr<const PreintegratedImuMeasurementsInterface> _PIM_;` instead of `PreintegratedImuMeasurements _PIM_;`). This creates a whole host of cascading issues that required me to make many more changes than I would have expected to implement this.
- Because of that, it's **no longer backwards compatible.** I had to make changes to some of ImuFactor's function signatures, because C++ has severe limitations on how an interface can be passed around. (`const PreintegratedImuMeasurements& preintegratedMeasurements()` -> `const std::shared_ptr<const PreintegratedImuMeasurementsInterface> preintegratedMeasurements()`). This means some code samples no longer compile and the Python interface has to be changed. And any other external code depending on ImuFactor would stop working.
- Due to the above, you need ugly casts to get the underlying PIM from the interface held by an ImuFactor: `*dynamic_cast<const PreintegratedImuMeasurementsT<TangentPreintegration>*>(factor->preintegratedMeasurements().get())`

This code doesn't compile as-is; there are still one or two issues. I also haven't made the conversion for CombinedImuFactor, that's still templated. But I thought I would put it up so we can make a decision whether to move forward with this... given that the motivation for this method was to keep "this PR - and disruption- small", I don't think it's what we want.

### DIFF
--- a/gtsam/navigation/CombinedImuFactor.cpp
+++ b/gtsam/navigation/CombinedImuFactor.cpp
@@ -21,6 +21,8 @@
  **/
 
 #include <gtsam/navigation/CombinedImuFactor.h>
+#include <gtsam/navigation/ManifoldPreintegration.h>
+#include <gtsam/navigation/TangentPreintegration.h>
 #if GTSAM_ENABLE_BOOST_SERIALIZATION
 #include <boost/serialization/export.hpp>
 #endif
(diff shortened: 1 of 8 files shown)


dellaert:
This is indeed not what we want. Thanks for investigating.

Issue #2161, opened 2025-06-04
4.3a0 Build issue with error: variable ‘m_runtime’ set but not used


PeterQFR:
## Description

Building tag v4.3a0 with the following changes to turn off Boost Features
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index cabde7653..ce461ce57 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,8 +68,8 @@ include(GtsamPrinting)
 
 ############### Decide on BOOST ######################################
 # Enable or disable serialization with GTSAM_ENABLE_BOOST_SERIALIZATION
-option(GTSAM_ENABLE_BOOST_SERIALIZATION "Enable Boost serialization" ON)
-option(GTSAM_USE_BOOST_FEATURES "Enable Features that use Boost" ON)
+option(GTSAM_ENABLE_BOOST_SERIALIZATION "Enable Boost serialization" OFF)
+option(GTSAM_USE_BOOST_FEATURES "Enable Features that use Boost" OFF)
 
 if(GTSAM_ENABLE_BOOST_SERIALIZATION OR GTSAM_USE_BOOST_FEATURES)
 include(cmake/HandleBoost.cmake)
```
in a 20.04 dockerfile and I get:
 
```

gtsam/gtsam/navigation/ManifoldEKF.h:162:11: error: variable ‘m_runtime’ set but not used [-Werror=unused-but-set-variable]
  162 |       int m_runtime;
      |           ^~~~~~~~~
cc1plus: all warnings being treated as errors
```

Not sure of the issue here. I'm building inside a ros2 galactic workspace with colcon, and I've not seen this error before.

PeterQFR:
I think this stems from a compiler issue with the use of constexpr and an if/else statement on line 

https://github.com/borglab/gtsam/blob/0fba7eb2d9fe78e52596ef823f8df041bb969a8d/gtsam/navigation/ManifoldEKF.h#L175

where the variable m_runtime is set but never used in the else part of the statement. 
by placing a simple use of the variable ie a dummy printf, the error goes away

e.g.
on line 182:
```
printf("Mruntime %d", m_runtime);
```
obviously this is a hack to test the issue, but not sure how to handle this appropriately, maybe move the declaration of m_runtime inside the only part of the if/else statement where it is used ie move line 162 to before line 164. Because on line 175, it is never used so I think that code is redundant and not required.

ProfFan:
[...]

PR #2168, opened 2025-06-11
Fix for Issue #2161 in borglab/gtsam

DanielChaseButterfield:
- [Fix for #2161] Moves definition of  `m_runtime` into the first if statement on line `163`, and removes assignment to `m_runtime` in else statement.


dellaert (approved):
Perfect, thanks!!!

PR #2171, opened 2025-06-14
ROS package.xml: do not depend on boost

jlblancoc:
GTSAM could be built now without boost, so let's not propagate the dependency downstream.

LMK if I'm wrong, but boost should be completely optional by now, right? 

If you could also please merge this into the release 4.3 branch, I will create a new ROS release from that to see how it goes... :crossed_fingers:

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -23,7 +23,6 @@
 
   <buildtool_depend>cmake</buildtool_depend>
 
-  <depend>boost</depend>
   <depend>eigen</depend>
   <depend>tbb</depend>
 


ProfFan:
Serialization is still on boost

ProfFan:
We might really start to consider using Cereal for the serialization (optionally). But we probably still want binary compatibility with older serialized files using boost::serialization...

jlblancoc:
Hmm... but so the current state is, if we try to build w/o Boost: does not build? or builds but it "only" lacks serialization? 
Because it would be great to have a ros package that does not introduce the many boost packages as build dependencies (think e.g. on minimal docker images).

ProfFan:
It totally builds and works, just there's no serialization

jlblancoc:
Great! Then, if you all agree with the ROS package of GTSAM not having serialization (with the gain of being lighter), this PR could be merged. Mentioned on ROS discourse [here](https://discourse.ros.org/t/on-new-gtsam-version-without-boost/44307/3) too.

dellaert:
I have no objection to that.

dellaert (approved):

dellaert:
@ProfFan if you agree please merge.
One doubt, @jlblancoc , the default flags seem to build with boost:
```
############### Decide on BOOST ######################################
# Enable or disable serialization with GTSAM_ENABLE_BOOST_SERIALIZATION
option(GTSAM_ENABLE_BOOST_SERIALIZATION "Enable Boost serialization" ON)
option(GTSAM_USE_BOOST_FEATURES "Enable Features that use Boost" ON)

if(GTSAM_ENABLE_BOOST_SERIALIZATION OR GTSAM_USE_BOOST_FEATURES)
include(cmake/HandleBoost.cmake)
endif()
######################################################################
```

So some process on ros build will need to turn them off.

jlblancoc:
Thanks for the pointer Frank! Yes, we'll probably need to turn that off, via cmake cli flags, no need to touch gtsam upstream. 
So, I'm merging this one :+1:

PR #2172, opened 2025-06-17
Updated `geometry.md` and Minor Fixes

CodeXTL:
Updated `geometry.md` to be more elaborate like the other markdown pages. Also made minor fixes to the header files of the calibration models.

Primary Update(s):
- Added more content to `geometry.md`
- Added a lean UML diagram showing hierarchical structure of all the calibration classes, with focus on specific parameters and dimension size
- Removed redundant `inline` keywords in all calibration model header files since modern compilers ignore this keyword in favor of their own inline heuristics.

Test(s):
- `geometry.md` looks fine on a local deployment.
- Ran `make check.geometry` with 100% pass


dellaert:
Nice! Three comments:
1) It would be nice to mention that the "groups" are not just groups but "manifolds and groups". They are defined in the same file but *could* be defined separately. See Pose3.ipynb for explanation
2) It would be nice to mention that a compile flag (mention which one) switches Rot3 between SO3 and Quaternion representations
3) Cal3 is not a manifold (just a container), so I was surprised it *has* a dimension. It should not I think. Could you try removing and do `make check` to confirm all unit tests still pass?

CodeXTL:
@dellaert Relating to comment 3. Simply removing `dimension` doesn't seem to work. I could just removed it from `Cal3` and make `Dim()` return a literal integer 5, which does work and pass all the geometry tests. Is this fine? Or should I look for some other solution?

dellaert:
Dim() and dimension should be defined for all *manifolds*. Cal3 is not a manifold, just a base class. If one of the derived classes (that is a manifold) fails when you remove that, you have to add it back in there.

what I am trying to avoid is that Cal3 says it’s “dimension” is 5. That’s a lie :-) it might have 5 fields, but the base classes define how many of them can be updated using a retraction.

please see Manifold.h for what a Manifold needs to have.

dellaert:
As discussed in the meeting, we’ll merge this PR. The dimension thing could be explored in a different PR.

dellaert (approved):

PR #2174, opened 2025-06-17
Fix constructor preintMeasCov reset

### LINKED ISSUE #2173: PreintegratedCombinedMeasurements possibly erroneous constructor
## Description

The PreintegratedCombinedMeasurementsT constructor looks like this:
```c++
PreintegratedCombinedMeasurementsT(
      const std::shared_ptr<Params>& p,
      const imuBias::ConstantBias& biasHat = imuBias::ConstantBias(),
      const Eigen::Matrix<double, 15, 15>& preintMeasCov =
          Eigen::Matrix<double, 15, 15>::Zero())
      : PreintegrationType(p, biasHat), preintMeasCov_(preintMeasCov) {
    this->resetIntegration();
  }
```
and resetIntegration looks like this:
```c++
template <class PreintegrationType>
void PreintegratedCombinedMeasurementsT<PreintegrationType>::resetIntegration() {
  // Base class method to reset the preintegrated measurements
  PreintegrationType::resetIntegration();
  preintMeasCov_.setZero();
}
```
Therefore, preintMeasCov_ is immediately reset to zero after being set from the parameter.

I think that the constructor should instead be calling `this->PreintegrationType::resetIntegration();`, as in this existing constructor:
```c++
PreintegratedCombinedMeasurementsT(
      const PreintegrationType& base,
      const Eigen::Matrix<double, 15, 15>& preintMeasCov)
      : PreintegrationType(base), preintMeasCov_(preintMeasCov) {
    this->PreintegrationType::resetIntegration();
  }
```

p-zach:
Resolves #2173

### DIFF
--- a/gtsam/navigation/CombinedImuFactor.h
+++ b/gtsam/navigation/CombinedImuFactor.h
@@ -97,7 +97,7 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurementsT : public PreintegrationTyp
       const Eigen::Matrix<double, 15, 15>& preintMeasCov =
           Eigen::Matrix<double, 15, 15>::Zero())
       : PreintegrationType(p, biasHat), preintMeasCov_(preintMeasCov) {
-    this->resetIntegration();
+    this->PreintegrationType::resetIntegration();
   }
 
   /**


dellaert (approved):
Thanks. definitely a bug.

PR #2178, opened 2025-06-25
Removed `dimension` field from `Cal3`

CodeXTL:
Primary Change(s):
- Removed `dimension` field and `Dim()` from `Cal3.h`
- Made `dim()` a pure virtual function in `Cal3.h`. This made `Cal3` an abstract class. This was done rather than removing the function entirely because derived classes override this function.
- Added an implementation of `dim()` with override keyword so that `Cal3_S2` becomes a concrete class. Not doing so made it abstract, which caused some errors in the unit tests

Check(s) Done:
- Executed `make check`, 100% tests passed

Note(s):
- This is a small PR as I have decided to divide my work based on relevance in order to prevent a backlog of PRs.

### DIFF
--- a/gtsam/geometry/Cal3.h
+++ b/gtsam/geometry/Cal3.h
@@ -73,7 +73,6 @@ class GTSAM_EXPORT Cal3 {
   double u0_ = 0.0f, v0_ = 0.0f;  ///< principal point
 
  public:
-  constexpr static auto dimension = 5;
   ///< shared pointer to calibration object
   using shared_ptr = std::shared_ptr<Cal3>;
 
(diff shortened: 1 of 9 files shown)


CodeXTL:
I believe I have realized what's causing the Python CI checks to fail; fixing it right now

dellaert:
Thanks! Two comments:
- I think dim() should also be removed from Cal3. What test fails if you remove it?
- then you don’t have to remove the Cal3 constructors, either.

CodeXTL:
@dellaert Responses for comments:
- All the derived classes override `dim()` from `Cal3` at the moment. I could remove `dim()` from `Cal3` if that's preferred, and the only change I would need to do is remove all the override keyword from the derived classes' definition of `dim()`
- Is there a reason why we want to keep the wrappers for the constructors given that `Cal3` is an abstract base class? Would there ever exist a reason for a user to create an instance of `Cal3`?

dellaert:
So:
- yes, please remove dim()
- Then it is no more an abstract base class :-) so keep constructors

CodeXTL:
@dellaert Got it! I've removed `dim()` and restored the wrappers for the `Cal3` constructors. I did the following checks afterwards:
- `make check`: 100% test pass
- `make python-test`: 100% pass

dellaert (approved):

PR #2180, opened 2025-06-27
Updates to Similarity2 and Similarity3

dellaert:
- added group and manifold operations to wrapper
- tested them
- added docs


dellaert:
@ShaneHolmes10 one this lands this should make it easy

PR #2184, opened 2025-07-02
Add Gal3 wrapper and...

dellaert:
...enable and test several other Matrix Lie groups with FrobeniusFactor classes

### DIFF
--- a/gtsam/geometry/Gal3.cpp
+++ b/gtsam/geometry/Gal3.cpp
@@ -41,9 +41,6 @@ namespace gtsam {
 namespace { // Anonymous namespace for internal linkage
   constexpr double kSmallAngleThreshold = 1e-10;
 
-  // The type of the Lie algebra (matrix representation)
-  using LieAlgebra = Matrix5;
-
   // Helper functions for accessing tangent vector components
   Eigen::Block<Vector10, 3, 1> rho(Vector10& v) { return v.block<3, 1>(0, 0); }
   Eigen::Block<Vector10, 3, 1> nu(Vector10& v) { return v.block<3, 1>(3, 0); }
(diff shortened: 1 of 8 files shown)


gtsam/geometry/Gal3.cpp
```diff
@@ -55,6 +52,45 @@ namespace { // Anonymous namespace for internal linkage
...
+          0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+          0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+          // --- M(i,4) --- rows 20-24
+          1, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+          0, 1, 0, 0, 0, 0, 0, 0, 0, 0,
+          0, 0, 1, 0, 0, 0, 0, 0, 0, 0,
+          0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+          0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+          ).finished();
+
```
p-zach, line 93: It might be good to have a source/short explanation of why this matrix is arranged the way it is; to me this just looks like a magic box of 1s and 0s. But maybe that's just because I am uneducated in Gal3

p-zach (approved):
Looks great

gtsam/geometry/Gal3.cpp
dellaert: Thanks for the comment. It does have this:
```cpp
  // The 25x10 matrix of vectorized generators for SGal(3).
  // The columns correspond to perturbations in [rho, nu, theta, t_tan].
  // The tangent vector is [rho(3), nu(3), theta(3), t_tan(1)].
  // P_gal3 = [vec(G_rho), vec(G_nu), vec(G_theta), vec(G_t)]
```
```cpp
        // The Jacobian is given by the formula H = (I_5 ⊗ T) * P_gal3
        // where P_gal3 is the matrix of vectorized generators.
        // This can be implemented efficiently with block-wise multiplication.
```
which I think is good enough for now. But I'm thinking of moving all of this to Lie.h so it can work for any Matrix Lie group, so I'll try to give more explanation then.

dellaert: PS a generator is Hat(e_i) where e_i is all zeros except the ith element

PR #2186, opened 2025-07-03
Ported StereoVOExample

CodeXTL:
Ported the C++ StereoVOExample to a Python Notebook. The new notebook example is functionally the same as the C++ example with some additional explanation spread throughout. 

Check(s):
- Checked the results of the C++ example and the newly ported notebook example, and they match
- Running `make check` passes all tests.

Note: this was originally planned to be part of a larger PR about stereo calibration models and stereo cameras, but I figured that it will be easier for review if I split this particular change into it's own PR. A PR with user guides for both `Cal3_S2Stereo` and `StereoCamera` will come in the near future.


dellaert (approved):
Awesome !!!

PR #2190, opened 2025-07-05
add analytical gal3 expmapderivative

mkielo3:
Replaced the Gal3::ExpmapDerivative with an analytical derivative. Long term need to make this more concise. All prior tests pass (including reconciliation to numerical derivatives)

[100%] Linking CXX executable testGal3
[100%] Built target testGal3
There were no test failures
[100%] Built target testGal3.run

### DIFF
--- a/gtsam/geometry/Gal3.cpp
+++ b/gtsam/geometry/Gal3.cpp
@@ -437,15 +437,105 @@ Vector10 Gal3::adjoint(const Vector10& xi, const Vector10& y, OptionalJacobian<1
 }
 
 //------------------------------------------------------------------------------
+static Matrix3 computeQMatrix(const so3::DexpFunctor& dexp, const Vector3& vec) {
+    Matrix3 vec_hat = skewSymmetric(vec);
+    Vector4 coeffs;
+    if (dexp.nearZero) {
+        coeffs << 0.5, 1.0 / 6.0, 0.0, 0.0;
+    } else {
+        coeffs << 0.5,
+            dexp.C,
+            (1.0 - 2.0 * dexp.B) / (2.0 * dexp.theta2),
+            (2.0 * dexp.theta - 3.0 * std::sin(dexp.theta) + dexp.theta * std::cos(dexp.theta)) / (2.0 * dexp.theta2 * dexp.theta2 * dexp.theta);
+    }
+    
+    Matrix3 T1 = vec_hat;
+    Matrix3 T2 = dexp.W * vec_hat + vec_hat * dexp.W + dexp.W * vec_hat * dexp.W;
+    Matrix3 T3 = dexp.WW * vec_hat + vec_hat * dexp.WW - 3.0 * (dexp.W * vec_hat * dexp.W);
+    Matrix3 T4 = dexp.WW * vec_hat * dexp.W + dexp.W * vec_hat * dexp.WW;
+    
+    return coeffs(0) * T1 + coeffs(1) * T2 + coeffs(2) * T3 + coeffs(3) * T4;
+}
+
+static Matrix3 computeGal3LeftJacobianU1(const gtsam::so3::DexpFunctor& dexp) {
+	// move into dexpfunctor?
+  if (dexp.nearZero) {
+    return 0.5 * Matrix3::Identity() + (1.0 / 3.0) * dexp.W;
+  } else {
+    return 0.5 * Matrix3::Identity() + (dexp.B - dexp.C) * dexp.W +
+           ((0.5 - dexp.A + dexp.B) / dexp.theta2) * dexp.WW;
+  }
+}
+
+static Matrix3 computeGamma1ThetaNuThetaTau(const so3::DexpFunctor& dexp, const Vector3& nu_vec) {
+    Matrix3 nu_hat = skewSymmetric(nu_vec);
+    Vector7 coeffs;
+    if (dexp.nearZero) {
+        coeffs << 1.0 / 6.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0;
+    } else {
+        const double tn3 = dexp.theta * dexp.theta2;
+        const double tn4 = dexp.theta2 * dexp.theta2;
+        const double tn5 = dexp.theta * tn4;
+        const double tn6 = dexp.theta2 * tn4;
+        const double s = std::sin(dexp.theta);
+        const double c = std::cos(dexp.theta);
+        
+        coeffs << 1.0 / 6.0,
+            dexp.E,  
+            (1.0 - 2.0 * dexp.B) / (2.0 * dexp.theta2), 
+            (tn3 + 6.0 * dexp.theta + 6.0 * dexp.theta * c - 12.0 * s) / (6.0 * tn5),
+            (12.0 * s - 12.0 * dexp.theta * c - 3.0 * dexp.theta2 * s - tn3) / (6.0 * tn5),
+            (tn3 - 6.0 * dexp.theta + 6.0 * s) / (6.0 * tn5),
+            (4.0 + dexp.theta2 + dexp.theta2 * c - 4.0 * dexp.theta * s - 4.0 * c) / (4.0 * tn6);
+    }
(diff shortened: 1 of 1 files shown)


gtsam/geometry/Gal3.cpp
```diff
@@ -437,15 +437,105 @@ Vector10 Gal3::adjoint(const Vector10& xi, const Vector10& y, OptionalJacobian<1
...
+    Vector3 rho_vec = minus_xi.segment<3>(0);
+    Vector3 nu_vec = minus_xi.segment<3>(3);
+    Vector3 theta_vec = minus_xi.segment<3>(6);
+    double t_val = minus_xi(9);
+
+    so3::DexpFunctor dexp(theta_vec);
+
+    Matrix10 Jl_minus_xi = Matrix10::Zero();
+
+    // Note: Because dexp is constructed on -theta, dexp.leftJacobian() returns Jr(theta)
```
dellaert, line 517: But, that makes it rather hard to compare with Navstate. Can we just use xi, not -xi?

gtsam/geometry/Gal3.cpp
```diff
@@ -437,15 +437,105 @@ Vector10 Gal3::adjoint(const Vector10& xi, const Vector10& y, OptionalJacobian<1
...
-    // Related to the left Jacobian in Equations 31-36, Pages 10-11
-    // NOTE: Using numerical approximation instead of implementing the analytical
-    // expression for the Jacobian. Future work to replace this
-    // with analytical derivative.
-    if (xi.norm() < kSmallAngleThreshold) return Matrix10::Identity();
-    std::function<Gal3(const Vector10&)> fn =
-        [](const Vector10& v) { return Gal3::Expmap(v); };
-    return numericalDerivative11<Gal3, Vector10>(fn, xi, 1e-5);
+    Vector minus_xi = -xi;
+    Vector3 rho_vec = minus_xi.segment<3>(0);
```
dellaert, line 508: Please use helper functions defined on line 44 etc...

gtsam/geometry/Gal3.cpp
```diff
@@ -437,15 +437,105 @@ Vector10 Gal3::adjoint(const Vector10& xi, const Vector10& y, OptionalJacobian<1
...
+    Vector3 nu_vec = minus_xi.segment<3>(3);
+    Vector3 theta_vec = minus_xi.segment<3>(6);
+    double t_val = minus_xi(9);
+
+    so3::DexpFunctor dexp(theta_vec);
+
+    Matrix10 Jl_minus_xi = Matrix10::Zero();
+
+    // Note: Because dexp is constructed on -theta, dexp.leftJacobian() returns Jr(theta)
+    Matrix3 right_jacobian = dexp.leftJacobian();
```
dellaert, line 518: After you use xi, just call this Jr so we can compare with navState?
In fact, take a look at NavState::expmap and use exactly the same names if you can
```cpp
  Vector3 w = xi.head<3>(), rho = xi.segment<3>(3), nu = xi.tail<3>();

  // Instantiate functor for Dexp-related operations:
  const so3::DexpFunctor local(w);

  // Compute rotation using Expmap
#ifdef GTSAM_USE_QUATERNIONS
  const Rot3 R = traits<gtsam::Quaternion>::Expmap(w);
#else
  const Rot3 R(local.expmap());
#endif

  // Compute translation and velocity. See Pose3::Expmap
  Matrix3 H_t_w, H_v_w;
  const Vector3 t = local.applyLeftJacobian(rho, Hxi ? &H_t_w : nullptr);
  const Vector3 v = local.applyLeftJacobian(nu, Hxi ? &H_v_w : nullptr);

  if (Hxi) {
    const Matrix3 Jr = local.rightJacobian();

```

dellaert: Please make absolutely sure also that `rho` is for position and `nu` is for velocity - as it is in NavState. Rename the helper functions if needed.

gtsam/geometry/Gal3.cpp
```diff
@@ -437,15 +437,105 @@ Vector10 Gal3::adjoint(const Vector10& xi, const Vector10& y, OptionalJacobian<1
...
 Matrix10 Gal3::ExpmapDerivative(const Vector10& xi) {
-    // Related to the left Jacobian in Equations 31-36, Pages 10-11
-    // NOTE: Using numerical approximation instead of implementing the analytical
-    // expression for the Jacobian. Future work to replace this
-    // with analytical derivative.
-    if (xi.norm() < kSmallAngleThreshold) return Matrix10::Identity();
-    std::function<Gal3(const Vector10&)> fn =
-        [](const Vector10& v) { return Gal3::Expmap(v); };
-    return numericalDerivative11<Gal3, Vector10>(fn, xi, 1e-5);
+    Vector minus_xi = -xi;
```
dellaert, line 507: Please do the calculation of the derivate *inside* `ExpMap`, as in `NavState`. That will actually save a lot of compute, and will make the connection with Pose3 and NavState more clear as well.

gtsam/geometry/Gal3.cpp
```diff
@@ -437,15 +437,105 @@ Vector10 Gal3::adjoint(const Vector10& xi, const Vector10& y, OptionalJacobian<1
...
+    Matrix10 Jl_minus_xi = Matrix10::Zero();
+
+    // Note: Because dexp is constructed on -theta, dexp.leftJacobian() returns Jr(theta)
+    Matrix3 right_jacobian = dexp.leftJacobian();
+    Jl_minus_xi.block<3,3>(0,0) = right_jacobian;   // ∂ρ/∂ρ block
+    Jl_minus_xi.block<3,3>(3,3) = right_jacobian;   // ∂ν/∂ν block
+    Jl_minus_xi.block<3,3>(6,6) = right_jacobian;   // ∂ϕ/∂ϕ block
+
+    Jl_minus_xi.block<3,3>(0,3) = -computeGal3LeftJacobianU1(dexp) * t_val;
+    Jl_minus_xi.block<3,3>(3,6) = computeQMatrix(dexp, nu_vec);
```
dellaert, line 524: I am 99% convinced that this the computation here is
```cpp
  Matrix3 H_t_w, H_v_w;
  const Vector3 v = local.applyLeftJacobian(nu, Hxi ? &H_v_w : nullptr);
```
where v is the velocity we obtain and H_v_w is your Q matrix. In other words, 
$v = J_l \nu$
and `H_v_w` is the derivative with respect to $\omega$. Provided you do *not* use $-\xi$.

gtsam/geometry/Gal3.cpp
```diff
@@ -437,15 +437,105 @@ Vector10 Gal3::adjoint(const Vector10& xi, const Vector10& y, OptionalJacobian<1
...
+            dexp.E,  
+            (1.0 - 2.0 * dexp.B) / (2.0 * dexp.theta2), 
+            (tn3 + 6.0 * dexp.theta + 6.0 * dexp.theta * c - 12.0 * s) / (6.0 * tn5),
+            (12.0 * s - 12.0 * dexp.theta * c - 3.0 * dexp.theta2 * s - tn3) / (6.0 * tn5),
+            (tn3 - 6.0 * dexp.theta + 6.0 * s) / (6.0 * tn5),
+            (4.0 + dexp.theta2 + dexp.theta2 * c - 4.0 * dexp.theta * s - 4.0 * c) / (4.0 * tn6);
+    }
+    
+    Matrix3 T0 = nu_hat;
+    Matrix3 T1 = dexp.W * nu_hat;
```
dellaert, line 493: These are just cross-products, so I think T2=-T1, and T4=-T3=-T5. Hence, you can probably combine coeffs 1 and 2, and coeffs 3,4, and 5.

dellaert (commented):
*very* cool, but I do have lots of comments :-)

gtsam/geometry/Gal3.cpp
dellaert: OOps I might be wrong about this !

dellaert:
This is now done in #2213

PR #2191, opened 2025-07-05
Fix timing if no boost

dellaert:
Using Chrono...

Timing FrobeniusBetweenFactor::evaluateError with derivatives for 10000000 iterations.

Rot2:   41.4726 ns
SO3:    60.0901 ns
Rot3:   65.0493 ns
Pose2:  97.0823 ns
SO4:    729.742 ns
Pose3:  414.104 ns
Gal3:   635.637 ns

### DIFF
--- a/gtsam/base/timing.cpp
+++ b/gtsam/base/timing.cpp
@@ -12,8 +12,8 @@
 /**
  * @file    timing.cpp
  * @brief   Timing utilities
- * @author  Richard Roberts, Michael Kaess
- * @date     Oct 5, 2010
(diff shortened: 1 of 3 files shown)


dellaert:
Dang, segfault in hybrid/tests *only* on Windows. @varunagrawal do you have a windows machine you can try this on?

dellaert:
Update: segfault seems to be in PR #2192

varunagrawal:
I do have a Windows machine. Let me look at this PR in more detail though.

dellaert:
I think this PR is fine, just needs a quick nod. I think the segfault is in custom_vecs - suspect a memory overrun.

PR #2193, opened 2025-07-06
Test both PIM variants

dellaert:
After @p-zach made the PIM type selectable at run-time, it made sense to make the tests run for all PIM types, especially as we might soon have new integration types. This PR validates that manifold and tangent integration versions behave identically, modulo test coverage.

### DIFF
--- a/gtsam/navigation/tests/imuFactorTesting.h
+++ b/gtsam/navigation/tests/imuFactorTesting.h
@@ -76,3 +76,25 @@ struct SomeMeasurements : vector<ImuMeasurement> {
 };
 
 }  // namespace testing
+namespace {
+// Macro to test ImuFactor with both Manifold and Tangent preintegration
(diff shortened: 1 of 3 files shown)


dellaert:
OK, @varunagrawal, now I am getting a segfault on windows, in debug, in a completely unrelated PR. Again in the hybrid tests. This makes me wonder whether there is a memory bug there - as it always occurs in hybrid, and is rather random. Maybe if you are able to ascertain in which specific test the segfault is that (a) could lead to tracking down the issue, but (b) we could skip that test for now in CI to allow other PRs to land.

p-zach (approved):
LGTM. For what it's worth, the same segfault occurred on the templated IMU PR. Rerunning the tests passed, but obviously is just a bandaid

dellaert:
Thanks for that info, @p-zach ! I tried adding some address sanitizing on the windows CLI, we'll see.

PR #2196, opened 2025-07-09
iLQR example notebook

zhangzdd:
The iLQR algorithm is implemented with CustomFactor in GTSAM. Results are compared against those from classical iLQR recursions (they almost perfectly match). Added mathematical expositions of both the iLQR algorithm and how it is handled in GTSAM


dellaert:
Could you finalize this week?

dellaert (approved):

PR #2197, opened 2025-07-10
EquivariantFilter and GroupAction

jenniferoum:
## Overview

This PR templates the Equivariant Filter on Group and Manifold structures, reorganizing the ABC (Attitude-Bias-Calibration) system into a more modular and generic architecture. The changes separate geometry-specific definitions from filter-specific functions and provide a reusable framework for equivariant filtering.

## Key Changes:

- Introduces generic EquivariantFilter class templated on manifold state and symmetry group
- Moves ABC geometry definitions to `gtsam_unstable/geometry/ABC.h`
- Adds GroupAction base class for defining group actions on manifolds
- Refactors ManifoldEKF to expose reusable Kalman gain and Joseph update methods
 - Renamed ABC_EQF_Demo to `AbcEquivariantFilterExample.cpp` 

## Testing
- unit tests in testEquivariantFilter and testABC
- Example runs and gives same results as Fornasier:
```bash
[100%] Built target AbcEquivariantFilterExample
ABC-EqF: Attitude-Bias-Calibration Equivariant Filter Demo
==============================================================
Loading data from /Users/dellaert/git/github/examples/Data/EqFdata.csv............... Done!
Loaded 12001 data points
Processing 12001 data points with EqF...
Progress: ........... Done!

=== Filter Performance Summary ===
Processing time: 2.33972 seconds
Processed measurements: 24002 (valid: 4923)

-- Average Errors --
Attitude: 1.13084°
Bias: 0.00754989
Calibration: 0.957467°

-- Final Errors --
Attitude: 3.66202°
Bias: 0.00206526
Calibration: 0.364382°

-- Final State vs Ground Truth --
Attitude (RPY) - Estimate:  -13.8045 -0.642714   86.7444° | Truth: -17.4067 -1.04149  86.2702°
Bias - Estimate: 0.00468984 0.00994432 0.00194401 | Truth: 0.00399696 0.00799933 0.00189684
Calibration (RPY) - Estimate: 24.8167 5.05822  29.674° | Truth: 25  5 30°

EqF demonstration completed successfully.
[100%] Built target AbcEquivariantFilterExample.run
```


gtsam_unstable/geometry/ABC.h
```diff
@@ -244,16 +214,261 @@ struct G {
...
+    return xi.localCoordinates(transformed);
+  };
+
+  Vector zeros = Vector::Zero(6 + 3 * N);
+  Matrix differential = numericalDifferential(coordsAction, zeros);
+  return differential;
+}
+
+template <size_t N>
+struct ABCGeometry {
```
dellaert: I think this is specifying too much: you are already specifying a group in the template, and via traits G already knows everything about it's group-nature. For example, `traits<G>::Compose` is the composition. See Lie.h and MatrixLieGroup.h for additional properties like Hat/Vee/vec.

So, action:
- remove any group things from geometry
- use traits in the template instead

gtsam_unstable/geometry/ABC.h
```diff
@@ -244,16 +214,261 @@ struct G {
...
+}
+
+/**
+ * @brief Calculates the Jacobian matrix using central difference approximation
+ * @param f Vector function f
+ * @param x The point at which Jacobian is evaluated
+ * @return Matrix containing numerical partial derivatives of f at x
+ * Uses Vector's size() and Zero(), Matrix's Zero() and col() methods
+ */
+Matrix numericalDifferential(std::function<Vector(const Vector&)> f,
```
dellaert: This is available in GTSAM in numericalDerivative.h. Not only for Vectors, but for general manifolds/groups.

examples/ABC.h
```diff
@@ -244,16 +214,261 @@ struct G {
...
+      if (idx >= static_cast<int>(N)) {
+        throw std::out_of_range("Calibration index out of range");
+      }
+      return X_hat.B[idx].matrix();
+    } else {
+      return X_hat.A.matrix();
+    }
+  }
+
+  static constexpr int DOF = 6 + 3 * N;
```
dellaert: This can be gotten from G or M?

examples/EquivariantFilter.h
```diff
@@ -0,0 +1,195 @@
...
+  Matrix Dphi0;           // Differential of phi at origin
+  Matrix InnovationLift;  // Innovation lift matrix
+
+ public:
+  /**
+   * Initialize EqF
+   * @param Sigma Initial covariance
+   * @param m Number of sensors
+   */
+  EqF(const Matrix& Sigma, int m);
```
dellaert: What if we have an initial state?

examples/EquivariantFilter.h
```diff
@@ -0,0 +1,195 @@
...
+EqF<G, M, Geometry>::EqF(const Matrix& Sigma, int m)
+    : X_hat(Geometry::identityGroup()),
+      Sigma(Sigma),
+      xi_0(Geometry::identityState()) {
+  if (Sigma.rows() != DOF || Sigma.cols() != DOF) {
+    throw std::invalid_argument(
+        "Initial covariance dimensions must match the degrees of freedom");
+  }
+
+  // Check positive semi-definite
```
dellaert: Get rid of semi-definite check: assume.

dellaert (commented):
Nice progress! Some initial comments.

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,351 @@
...
+  Matrix Dphi0;           // Differential of phi at origin
+  Matrix InnovationLift;  // Innovation lift matrix
+
+ public:
+  /**
+   * Initialize EqF
+   * @param Sigma Initial covariance
+   * @param m Number of sensors
+   */
+  EqF(const G& X0, const M& x0, const Matrix& Sigma, int m);
```
dellaert: X0 and x0 are related by x0 = \phi(X0, ref), and if it's an action, we can always get X0 from  this. So I think: we need to give 2 M thingies: x0, and ref. X0 is derived.

Because the state is defined on the manifold, the user will want to specify the initial state on the manifold. And, more often than not, they will initialize the filter *at* the reference element.

EqF(const M& xi_circle, const M& xi_0, const Matrix& Sigma, int m);
EqF(const M& xi_circle, const Matrix& Sigma, int m);

examples/EquivariantFilter.h
```diff
@@ -0,0 +1,351 @@
...
+  M state_est = stateEstimate();            // General manifold
+  Vector L = Geometry::lift(state_est, u);  // Lifted vector in tangent space
+
+  Matrix Phi = Geometry::stateTransitionMatrix(u, dt, X_hat);
+  Matrix Bt = Geometry::inputMatrixBt(X_hat);
+
+  Matrix Q = Geometry::processNoise(u);  // Replaces blockDiag(...)
+
+  X_hat = traits<G>::Compose(X_hat, traits<G>::Expmap(L * dt));
+  Sigma = Phi * Sigma * Phi.transpose() + Bt * Q * Bt.transpose() * dt;
```
dellaert: This is already done in the ManifoldEKF, specialized to groups in the LieGroupEKF, which is the main reason to derive form that class.

examples/EquivariantFilter.h
```diff
@@ -0,0 +1,351 @@
...
+    return;  // Skip this measurement
+  }
+
+  Matrix Ct = Geometry::measurementMatrixC(y.d, y.cal_idx);
+
+  Vector3 action_result = outputAction(X_hat.inv(), y.y, y.cal_idx);
+  Vector3 delta_vec = Rot3::Hat(y.d.unitVector()) * action_result;
+  Matrix Dt = Geometry::outputMatrixDt(y.cal_idx, X_hat);
+
+  Matrix S = Ct * Sigma * Ct.transpose() + Dt * y.Sigma * Dt.transpose();
```
dellaert: You'll also find this somewhere in the existing hierarchy.

AlessandroFornasier:
Nice to see this one in gtsam 👍🏻

dellaert:
Thans @AlessandroFornasier !We're going to substantially rewrite and base on LieGroupEKF (or derived), we'll let you know :-)

gtsam_unstable/geometry/ABC.h
```diff
@@ -164,46 +164,48 @@ class State {
...
+struct Group {
   Rot3 A;                 /// First SO(3) element
   Matrix3 a;              /// so(3) element (skew-symmetric matrix)
   std::array<Rot3, N> B;  /// List of SO(3) elements for calibration
   static constexpr int dimension = 6 + 3 * N;
   using TangentVector = Eigen::Matrix<double, dimension, 1>;
-
-  /// Initialize the symmetry group G
-  G(const Rot3& A = Rot3::Identity(), const Matrix3& a = Matrix3::Zero(),
+  static constexpr int numSensors = N;
```
dellaert: Should also be small `n` if we are to match the paper.

gtsam_unstable/geometry/ABC.h
```diff
@@ -214,23 +216,60 @@ struct G {
...
-    return gtsam::traits<G>::Compose(*this, gtsam::traits<G>::Expmap(v));
+  Group retract(const TangentVector& v,
+            OptionalJacobian<dimension, dimension> H = {},
+            OptionalJacobian<dimension, dimension> Hv = {}) const {
+    return gtsam::traits<Group>::Compose(*this, gtsam::traits<Group>::Expmap(v));
+  }
+
+  // Adjoint matrix of this group element (for SE(3) or similar)
+  Eigen::Matrix<double, dimension, dimension> AdjointMap() const {
+    // TODO: implement properly for your group structure.
```
dellaert: It would be nice to implement for your group, or at least figure out the math, as in the invariant filter the A matrix is just $Ad_{U^{-1}}$. Might that be the case as well here, with U the “lifted” dynamics?

gtsam_unstable/geometry/ABC.h
```diff
@@ -261,7 +300,7 @@ State<N> operator*(const G<N>& X, const State<N>& xi) {
  * the input equivariance
  */
 template <size_t N>
-Input velocityAction(const G<N>& X, const Input& u) {
+Input velocityAction(const Group<N>& X, const Input& u) {
```
dellaert: The Input type is mixing two things I believe: a mathematical object, $(\omega,0)$ where the zeros are that “virtual” input. And then a data structure that is used to store noise parameters and what not. I think the code will be a lot more clear if you separate those concerns, rename `Input` to `InputData`, and then redefine Input as simply
```
Using Input = Vector6
```
Or even just use Vector6.

gtsam_unstable/geometry/ABC.h
```diff
@@ -321,16 +360,16 @@ Matrix numericalDifferential(std::function<Vector(const Vector&)> f,
  */
 template <size_t N>
 Matrix stateActionDiff(const State<N>& xi) {
-  return gtsam::numericalDerivative11<Vector, G<N>>(
-      [&xi](const G<N>& g) { return xi.localCoordinates(g * xi); },
-      gtsam::traits<G<N>>::Identity());
+  return gtsam::numericalDerivative11<Vector, Group<N>>(
```
dellaert: If this is only used in tests move it there.

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,363 @@
...
+ * Estimation with Online Calibration" by Fornasier et al. Authors: Darshan
+ * Rajasekaran & Jennifer Oum
+ */
+
+#ifndef EQUIVARIANTFILTER_H
+#define EQUIVARIANTFILTER_H
+#pragma once
+#include <gtsam/base/Matrix.h>
+#include <gtsam/base/Vector.h>
+#include <gtsam/geometry/Rot3.h>
```
dellaert, line 18: These includes seem out of place when M and G are generic.

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,363 @@
...
+  void predict(const typename Geometry::Input& u, double dt);
+
+  /**
+   * Update the filter state with a measurement
+   * @param y Direction measurement
+   */
+  void update(const typename Geometry::Measurement& y);
+
+  static constexpr int DOF = gtsam::traits<G>::dimension;
+  static constexpr int n_cal = Geometry::n_cal;
```
dellaert: Also out of place in a generic filter.

gtsam_unstable/geometry/ABC.h
```diff
@@ -0,0 +1,603 @@
...
+    Matrix Phi2 = repBlock(Phi22, N);
+    return blockDiag(Phi1, Phi2);
+  }
+  /**
+   * Computes linearized continuous time state matrix
+   * @param u Angular velocity
+   * @return Linearized state matrix
+   * Uses Matrix zero and Identity functions
+   */
+  static Matrix stateMatrixA(const GType& X_hat, const Input& u) {
```
dellaert: Is this actually used in the filter?

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,363 @@
...
+  M state_est = stateEstimate();            // your manifold version
+  Vector L = Geometry::lift(state_est, u);  // tangent vector
+
+
+  Matrix Phi = Geometry::stateTransitionMatrix(u, dt, this->X_);
+  Matrix Bt  = Geometry::inputMatrixBt(this->X_);
+  Matrix Q   = Geometry::processNoise(u);
+
+
+  this->X_ = traits<G>::Compose(this->X_, traits<G>::Expmap(L * dt));
```
dellaert: This makes me think $\Phi = Ad_{U^{-1}}$ with $U=L dt$. If so we can then kill it from Geometry - and I bet this holds for any equivariant filter that updates the state this way. Look at the InvariantEKF.

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,363 @@
...
+  Vector3 action_result = outputAction(this->X_.inv(), y.y, y.cal_idx);
+  Vector3 delta_vec = Rot3::Hat(y.d.unitVector()) * action_result;
+  Matrix Dt = Geometry::outputMatrixDt(y.cal_idx, this->X_);
+
+  Matrix S = Ct * this->P_ * Ct.transpose() + Dt * y.Sigma * Dt.transpose();
+  Matrix K = this->P_ * Ct.transpose() * S.inverse();
+  Vector Delta = InnovationLift * K * delta_vec;
+  this->X_ = traits<G>::Compose(traits<G>::Expmap(Delta), this->X_);
+  this->P_ = (Matrix::Identity(DOF, DOF) - K * Ct) * this->P_;
+}
```
dellaert: Try to see whether you can do this update by constructing the parameters of `update` in the base class. It will be informative.

gtsam_unstable/geometry/ABC.h
```diff
@@ -0,0 +1,603 @@
...
+  return result;
+}
+
+//========================================================================
+// Core Data Types
+//========================================================================
+
+/// Input struct for the Biased Attitude System
+
+struct Input {
```
dellaert: I would split this file up: ABC.h is *pure* geometry. All the other data structures (Input, etc) should be moved to the example - there should be zero “infra” code in this geometry code.

dellaert (commented):
Good meeting. Trying to give some more feedback here. The goal is whether we can get rid of Geometry altogether.

dellaert:
@AlessandroFornasier want to take a new look? I think we successfully replicated your paper in an elegant framework now.

gtsam/navigation/tests/testEquivariantFilter.cpp
```diff
@@ -0,0 +1,464 @@
...
+    return Q.unrotate(eta, H_Q, H_eta);
+  }
+};
+
+//---------------------------------------------------------------------------
+// Dynamics
+//------------------------------------------- --------------------------------
+
+// Compute the dynamics at a given point on the manifold
+Vector2 dynamics(const Vector3& omega, const M& eta) {
```
dellaert, line 61: Needed for filter or only for tests??

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,308 @@
...
+   *
+   * @tparam Innovation Functor computing ν(ξ̂) and its Jacobian.
+   * @param innovation Innovation functor.
+   * @param R Measurement noise covariance.
+   */
+  template <typename Innovation>
+  void update(const Innovation& innovation, const Eigen::MatrixXd& R) {
+    const M xi_hat = this->state();
+    auto nu = innovation(xi_hat);
+    using VectorZ = decltype(nu);
```
dellaert: kill vectorZ, move innovation call down

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,308 @@
...
+   * @brief Update with measurement, using provided measurement Jacobian C.
+   *
+   * @tparam Innovation Functor computing ν(ξ̂) without needing the action.
+   * @param innovation Innovation functor.
+   * @param R Measurement noise covariance.
+   * @param C Measurement Jacobian (DimZ x DimM).
+   */
+  template <typename Innovation>
+  void update(const Innovation& innovation, const Eigen::MatrixXd& R,
+              const Eigen::MatrixXd& C) {
```
dellaert: only dynamic allowed?

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,308 @@
...
+   * Concept requirements:
+   * - `Innovation` must be callable as `innovation(xi_hat, H)` where H is an
+   *   OptionalJacobian (fixed-size) or MatrixXd* (dynamic-size).
+   *
+   * @tparam Innovation Functor computing ν(ξ̂) and its Jacobian.
+   * @param innovation Innovation functor.
+   * @param R Measurement noise covariance.
+   */
+  template <typename Innovation>
+  void update(const Innovation& innovation, const Eigen::MatrixXd& R) {
```
dellaert: Consider unifying with update API in ManifoldEKF. Also: document update steps in the other filters better.

dellaert (commented):
With Scott

gtsam/navigation/ManifoldEKF.h
```diff
@@ -142,6 +144,20 @@ class ManifoldEKF {
     P_ = F * P_ * F.transpose() + Q;
   }
 
+  /// Kalman gain K = P H^T S^-1.
+  template <typename HMatrix, typename RMatrix>
+  auto KalmanGain(const HMatrix& H, const RMatrix& R) const {
+    auto S = H * P_ * H.transpose() + R;  // Innovation covariance
```
scottiyio, line 134: i like this, makes the ekf clearer

gtsam/navigation/tests/testEquivariantFilter.cpp
```diff
@@ -0,0 +1,464 @@
...
+  // TODO(Frank): suspect, should depend on eta? Or make sure it works for
+  // chosen reference direction.
+  return Matrix23::Identity();
+}
+
+struct Innovation {
+  Innovation(const Vector3& z_measured, double c_m)
+      : z_measured_(z_measured), c_m_(c_m) {}
+
+  /// Innovation: ẑ - z_measured, with ẑ = c_m * η̂.
```
scottiyio: This concerns me. I have always seen "z - zhat" or "measured - predicted". why is it switched here? is there something in eqf that causes this, or some convention?

2) This innovation takes in a vector - what about other residuals that are multiplicative (e.g. rotation matrices or a quaternion)

scottiyio (approved):
Just want to check on  the notation of z - zhat

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,308 @@
...
+   *   that accepts the current group estimate X and returns the mapped input
+   *   (no other methods are required by the filter).
+   *
+   * @tparam Lift Functor for the lift Λ(ξ, u).
+   * @tparam InputOrbit Functor for the input orbit ψ_u.
+   * @param psi_u Input Orbit instance.
+   * @return MatrixM The calculated error dynamics matrix A.
+   */
+  template <typename Lift, typename InputOrbit>
+  MatrixM computeErrorDynamicsMatrix(const InputOrbit& psi_u) const {
```
AlessandroFornasier, line 131: Mentioning only for future work.

An EqF can still be designed even if a compatible action psi is not found. Current version seems to not support this case. Could be a nice addition in future.

In such case you would have something like the following:
```
//...
typename Symmetry::Orbit act_on_x(this->X_);
typename Symmetry::Orbit coords_act(act_on_x(traits<G>::Identity()));

coords_act(g_.inverse(), D_phi);

Lift lift_u_origin(u);
lift_u_origin(act_on_ref_(g_), &D_lift);

return D_phi * D_lift
```

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,308 @@
...
+    TangentG Lambda = lift_u(xi_est);  // Pure lift
+
+    g_ = traits<G>::Compose(g_, traits<G>::Expmap(Lambda * dt));
+    M xi_next = act_on_ref_(g_);
+
+    // 2. Covariance Propagation on Manifold
+    MatrixM Phi = transitionMatrix<K>(A, dt);
+
+    // Qc is manifold continuous-time covariance: Q_M = Qc * dt
+    CovarianceM Q_manifold = Qc * dt;
```
AlessandroFornasier, line 220: Do I understand correctly that `Qc` provided here is `inputMatrixB * Qc * inputMatrixB.transpose()`. I think it is not fully clear from the docstring (same applies to `predict`).

I also wonder whether it is less error prone to have those function be provided with `Qc` and have the B matrix be either provided (in `predictWithJacobian`) or computed internally (in `predict`)

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,308 @@
...
+
+    if constexpr (DimZ == Eigen::Dynamic) {
+      Eigen::Matrix<double, Eigen::Dynamic, DimM> H(nu.rows(), DimM);
+      innovation(xi_hat, &H);
+      updateInternal(nu, H, R);
+    } else {
+      Eigen::Matrix<double, DimZ, DimM> H;
+      VectorZ nu_with = innovation(xi_hat, &H);
+      updateInternal(nu_with, H, R);
+    }
```
AlessandroFornasier: Similar to the comment above about the `inputMatrixB`. is `R` provided here `outputMatrixD * R * outputMatrixD.transpose()`?

gtsam/navigation/EquivariantFilter.h
```diff
@@ -0,0 +1,308 @@
...
+ protected:
+  /**
+   * @brief Internal update implementation.
+   *
+   * Uses standard EKF equations:
+   * K = P * H^T * (H * P * H^T + R)^-1
+   * P = (I - K * H) * P * (I - K * H)^T + K * R * K^T
+   */
+  template <typename VectorZ, typename MatrixZM, typename MatrixZ>
+  void updateInternal(const VectorZ& innovation, const MatrixZM& H,
```
AlessandroFornasier: Mentioning only for future work.

A reset step (also called curvature correction) could be implemented in the update and enabled under a flag eventually. See page 38-39 https://arxiv.org/pdf/2407.14297

examples/AbcEquivariantFilterExample.cpp
```diff
@@ -252,25 +267,40 @@ void processDataWithEqF(EqFilter& filter, const std::vector<Data>& data_list,
...
       }
 
       try {
-        filter.update(y);
+        Innovation innovation(measurement.y, measurement.d,
+                              measurement.cal_idx);
+        // Use Explicit Matrices API
+        Matrix C =
+            abc::measurementMatrixC<n>(measurement.d, measurement.cal_idx);
+        filter.update(innovation, measurement.R, C);
```
AlessandroFornasier: Isn't the application of the `outputMatrixD` missing here?

AlessandroFornasier (commented):
Generally LGTM. Nice job! I am very happy to see this within GTSAM!

gtsam_unstable/geometry/ABC.h
```diff
@@ -0,0 +1,555 @@
...
+      const Rot3& Shati = xi_hat.S[index_];
+      const Rot3 Bi = S0i.inverse() * A * Shati;
+      transformed_y = Bi.rotate(y_.unitVector());
+    }
+
+    const Matrix3 wedge_d = Rot3::Hat(d_.unitVector());
+    const Vector3 nu = wedge_d * transformed_y;
+
+    if (H) {
+      *H = measurementMatrixC<N>(d_, index_);
```
AlessandroFornasier, line 452: Just a question out of curiosity. Any plan to extend this to support `measurementMatrixCStar` ? Reference to equations 3.24-3.25 page 34 https://arxiv.org/pdf/2407.14297

gtsam/navigation/tests/testEquivariantFilter.cpp
AlessandroFornasier: Hi!

Don't know the reason for such implementation but from a theoretical point of view there is no convention forcing the innovation to be `z - z_hat` or `z_hat - z` in the EqF. In fact `(z - z_hat) = (-1) * (z_hat - z)` so the only difference would be the sign of the Jacobian.

To answer point 2. The EqF does not assume measurements to be vectors but measurements can be elements of an output space (manifold) N (can be a vectorspace or anything fancy you can imagine). Since N is a differentiable manifold you can write local coordinates, therefore, the innovation is required to be a vector written in those local coordinates about a fixed measurement `z_ref`. In general you can always write the innovation as `innovation = delta(z) - delta(z_hat)`. `delta()` represents local coordinates. 

For example, imagine measuring a direction (an element of S2). In this case you cannot write `innovation = z - z_hat` but you can define a reference measurement `z_ref` and define local coordinates `delta(z) = z_ref.cross(z)`. Then the innovation would just be `innovation = delta(z) - delta(z_hat)`.

This way is general and works also for vector measurements. In this case `delta(z) = z - z_ref`. Then `innovation = delta(z) - delta(z_hat)` becomes `innovation = z - z_ref - z_hat + z_ref = z - z_hat`.

What I just said is always true (with or without output equivariance). However, for completeness sake, I want to mention that if the output is equivariant the innovation can also be computed exploiting the output action.

More details page 34-35 https://arxiv.org/pdf/2407.14297

I hope this clarifies it. Feel free to ask if you have any more question about EqFs.

AlessandroFornasier:
One more thing, maybe `EKF-variants.md` should be extended?

dellaert:
Thanks a bunch, @scottiyio and @AlessandroFornasier ! I will keep some of these as future work (non equivariant input, and equivariant output), but will address the rest. In a meeting with Stephan I also simplified the abc::Group.

gtsam/navigation/EquivariantFilter.h
dellaert: Yeah. This agrees with other EKF variants, where we chose not to do the explicit `B` modeling. I see that as a more sophisticated modeling of Q as a tuning knob. Agreed it's more principled but choose uniformity in API for now.

gtsam/navigation/tests/testEquivariantFilter.cpp
dellaert: Agree just a sign change, but moot as I will align this API with ManifoldEKF, which also takes any Manifold type as Measurement. C* is for a follow-up PR.

dellaert (approved):
Thanks everyone! A team effort, and great reviews! I will merge now!

PR #2199, opened 2025-07-22
Remove initial cov from CombinedImuFactor

dellaert:
...with a notebook showing equivalence of this new version.

I will update the lyx file soon, as well, and include the pdf in the PR comment. Bottom line: the initial mechanism of providing an initial covariance to the CombinedImuFactor was misguided: this should be done by adding a prior to the first bias; the factor graph machinery will take care of the rest.

[gtsam_gtsam_navigation_doc_Combined_vs_ImuFactor.ipynb at feature_nav_docs · borglab_gtsam.pdf](https://github.com/user-attachments/files/21392668/gtsam_gtsam_navigation_doc_Combined_vs_ImuFactor.ipynb.at.feature_nav_docs.borglab_gtsam.pdf)

[ImuFactor.pdf](https://github.com/user-attachments/files/21392675/ImuFactor.pdf)


gtsam/navigation/CombinedImuFactor.cpp
```diff
@@ -39,24 +39,17 @@ using namespace std;
...
-  cout << "biasAccOmegaInt:\n[\n" << biasAccOmegaInt << "\n]"
-       << endl;
+  cout << "biasAccCovariance:\n[\n" << biasAccCovariance << "\n]" << endl;
+  cout << "biasOmegaCovariance:\n[\n" << biasOmegaCovariance << "\n]" << endl;
 }
 
 //------------------------------------------------------------------------------
 bool PreintegrationCombinedParams::equals(const PreintegratedRotationParams& other,
-                                  double tol) const {
+  double tol) const {
```
varunagrawal: This indentation seems weird. Are you sure your formatting options are set up as with previous GTSAM code? It may be because of different formatting options with another code base that's now being applied here.

gtsam/navigation/CombinedImuFactor.cpp
```diff
@@ -39,24 +39,17 @@ using namespace std;
...
   auto e = dynamic_cast<const PreintegrationCombinedParams*>(&other);
   return e != nullptr && PreintegrationParams::equals(other, tol) &&
-         equal_with_abs_tol(biasAccCovariance, e->biasAccCovariance,
-                            tol) &&
-         equal_with_abs_tol(biasOmegaCovariance, e->biasOmegaCovariance,
-                            tol) &&
-         equal_with_abs_tol(biasAccOmegaInt, e->biasAccOmegaInt, tol);
+    equal_with_abs_tol(biasAccCovariance, e->biasAccCovariance, tol) &&
+    equal_with_abs_tol(biasOmegaCovariance, e->biasOmegaCovariance,tol);
```
varunagrawal: Yeah this formatting is very suspect.

varunagrawal (commented):
I have a feeling the formatter is off so I recommend looking into that, but also CI is failing?

gtsam/navigation/CombinedImuFactor.cpp
dellaert: Thanks. Yes, I want to get to the bottom as to why my formatting seems to be different from yours, but will not worry about it in this PR.

gtsam/navigation/CombinedImuFactor.cpp
dellaert: same

gtsam/navigation/CombinedImuFactor.h
```diff
@@ -171,6 +164,15 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurementsT : public PreintegrationTyp
 
   /// @}
 
+#ifdef GTSAM_ALLOW_DEPRECATED_SINCE_V43
+/// @deprecated: biasAccOmegaInt is no longer used. Use a prior on first bias instead.
+  void resetIntegration(const gtsam::Matrix6& Q_init) {
```
varunagrawal, line 169: We should have a new version that is something like:
```cpp
void resetIntegration() {
    PreintegrationType::resetIntegration();
    preintMeasCov_.setZero();
}
```

dellaert: It is there.

gtsam/navigation/CombinedImuFactor.cpp
dellaert: Fixed

gtsam/navigation/CombinedImuFactor.cpp
dellaert: fixed

varunagrawal (approved):
Awesome!

lucacarlone (approved):

dellaert:
Thanks all! Merging now.

PR #2200, opened 2025-07-23
Added User Guides for `Cal3_S2Stereo` and `StereoCamera`

CodeXTL:
This PR adds user guides for `Cal3_S2Stereo` and `StereoCamera` along with some minor related changes. Details listed below.

Update(s):
- Added user guides for `Cal3_S2Stereo` and `StereoCamera`
- Minor correction in StereoVOExample notebook and `Cal3_S2` user guide
- Fixed Python wrapper for `Cal3_S2Stereo`
- Added `calibrate(Vector3 p)` to wrapper for `Cal3_S2` (wrapper for this function was missing prior to this)
- Removed unnecessary `inline` keyword for static function in `CalibratedCamera.h` 
- Replaced the old `typedef` with its modern replacement `using` in `CalibratedCamera.h`, `Point3.h`, `StereoCamera.h`

### DIFF
--- a/gtsam/geometry/CalibratedCamera.h
+++ b/gtsam/geometry/CalibratedCamera.h
@@ -222,7 +222,7 @@ public:
    * exponential map parameterization
    * @return a pair of [start, end] indices into the tangent space vector
    */
-  static std::pair<size_t, size_t> translationInterval() {
+  static std::pair<size_t, size_t> TranslationInterval() {
     return {3, 5};
   }
 
--- a/gtsam/geometry/CalibratedCamera.h
+++ b/gtsam/geometry/CalibratedCamera.h
@@ -56,15 +56,15 @@ class GTSAM_EXPORT PinholeBase {
 public:
 
   /** Pose Concept requirements */
-  typedef Rot3 Rotation;
-  typedef Point3 Translation;
+  using Rotation = Rot3;
+  using Translation = Point3;
 
   /**
    *  Some classes template on either PinholeCamera or StereoCamera,
    *  and this typedef informs those classes what "project" returns.
    */
-  typedef Point2 Measurement;
-  typedef Point2Vector MeasurementVector;
+  using Measurement = Point2;
+  using MeasurementVector = Point2Vector;
 
 private:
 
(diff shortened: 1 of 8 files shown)


gtsam/geometry/CalibratedCamera.h
```diff
@@ -222,7 +222,7 @@ class GTSAM_EXPORT PinholeBase {
    * exponential map parameterization
    * @return a pair of [start, end] indices into the tangent space vector
    */
-  inline static std::pair<size_t, size_t> translationInterval() {
+  static std::pair<size_t, size_t> translationInterval() {
```
dellaert: TranslationInterval

gtsam/geometry/geometry.i
```diff
@@ -965,6 +969,16 @@ virtual class Cal3_S2Stereo   : gtsam::Cal3{
   // Standard Interface
   double baseline() const;
   gtsam::Vector6 vector() const;
+
```
dellaert, line 972: try removing these. Should be inherited?

dellaert (approved):
Awesome ! I left some comments

gtsam/geometry/CalibratedCamera.h
CodeXTL: Done

gtsam/geometry/geometry.i
CodeXTL: I checked again and remembered why I did this. The inheritance doesn't work because `Cal3_S2.calibrate()` takes an OptionalJacobian Dcal with a dimension of 2 by 5, while `Cal3_S2Stereo.calibrate()` takes an OptionalJacobian Dcal with a dimension of 2 by 6. Due to this difference, the two functions are effectively different functions, so inheritance doesn't work.

I'm going to keep what I have for now because I know it works; I'll see if there are some better ways of doing this.

CodeXTL:
@dellaert I have committed the changes we discussed. Also, I don't have merge privileges, so I'm going to need some help with merging.

dellaert:
Excellent, will merge!

PR #2201, opened 2025-07-24
Add DiscreteScenario

p-zach:
DiscreteScenario for ScenarioRunner. Takes in a list of timestamped measurements, or consumes them from a CSV.


dellaert (approved):
OK, This is awesome. Two requests:
- add a Scenario.cpp file with all definitions (including older classes)
- add a unit test for each constructor in testScenario.cpp

p-zach:
@dellaert Having trouble running unit tests on windows but here are the changes I have. Will update if any of the tests fail once I can run them

dellaert (approved):
Awesome !

PR #2205, opened 2025-07-28
Fix examples using biasAccOmegaInt

varunagrawal:
Follow up from #2199, some of the examples don't compile, so this PR adds the necessary guards to get them to compile.


dellaert (approved):
Thanks!!

PR #2206, opened 2025-07-28
Add reference to wrap docs

p-zach:
> You could also point users to the documentation file at https://github.com/borglab/wrap/blob/master/DOCS.md.

Adds reference to wrap docs to DEVELOP.md. Thanks for the link @varunagrawal, very useful info.


varunagrawal (approved):
LGTM

PR #2207, opened 2025-07-29
Add SL4 to geometry

dellaert:
Add SL(4), based on fork by @LimHyungTae, here: https://github.com/MIT-SPARK/gtsam_with_sl4

I made a bunch of changes to leverage LieGroup and MatrixLieGroup, added unit test, and converted the python example to a notebook.


LimHyungTae:
Hallo Frank, thanks for taking your time. I'll take a closer look as well by the end of this week!

gtsam/geometry/SL4.cpp
```diff
@@ -0,0 +1,162 @@
...
+  Matrix44 H_inv_T = T_.inverse().transpose();
+  Matrix16x16 C_H;
+
+  // Kronecker product H ⊗ H^{-T}
+  for (int i = 0; i < 4; ++i)
+    for (int j = 0; j < 4; ++j)
+      C_H.block<4, 4>(i * 4, j * 4) = T_(i, j) * H_inv_T;
+
+  return ALG_TO_VEC * C_H * VEC_TO_ALG;
+}
```
mkielo3, line 138: You might be able to inherit AdjointMap() form MatrixLieGroup since Hat()/Vee() are defined?

gtsam/geometry/SL4.cpp
```diff
@@ -0,0 +1,162 @@
...
+bool SL4::equals(const SL4& sl4, double tol) const {
+  return T_.isApprox(sl4.T_, tol);
+}
+/* ************************************************************************* */
+SL4 SL4::ChartAtOrigin::Retract(const Vector15& v, ChartJacobian H) {
+  assert(v.size() == 15);
+  SL4 retracted(I_4x4 + Hat(v));
+  if (H) *H = I_15x15;
+  return retracted;
+}
```
mkielo3, line 96: Looks like this is an approximation for SL4::Expmap(v, H)? I see note on runtime in Expmap. If there is an efficiency reason we must use approximation here then maybe comment why we differ from Expmap? Same goes for derivative

gtsam/geometry/SL4.cpp
dellaert: This is a custom implementation by @LimHyungTae , I added a test to make sure it is equivalent. I trust @LimHyungTae is more efficient, although I did not check.

gtsam/geometry/SL4.cpp
dellaert: Retract and Local are allowed to be approximations.
I do now throw when Jacobians are requested, as they are not identity.

mkielo3 (approved):

dellaert:
@LimHyungTae any comments before I merge? **After** I merge you should in principle be able to abandon your fork and just run VGGT-SLAM after `pip install gtsam-develop`, a new package that provides bleeding-edge access.

dellaert:
I'll merge now, so @LimHyungTae can test whether `pip install gtsam-develop` does indeed work with VGGT-SLAM

LimHyungTae:
Thanks, @dellaert, everything looks good to me (and the captured figure is VGGT-SLAM with `gtsam-develop`). Next time, I'll follow your guide and refer to this PR ;). 

image

PR #2212, opened 2025-08-08
Adds a Lie-Group aware differentiable spline curve implementation

BrettRD:
This PR follows a discussion on the mailing list about clock recovery for continuous-time trajectories.
I got the ok from my team to push the abstract features upstream, and I've re-written the implementation to be more versatile and line up better with existing literature.

I'm not ready to merge just yet (the header file preambles are conspicuously missing and the comments are still a bit of a mess), but I'd like some feedback for where the files should live, name-spacing, etc.
I'd also like some guidance on how best to expose Expression based features to the python bindings.

## Basic usage:
```
  // instantiate the expression factory (uses cubic splines by default)
  TrajectoryModel<Pose3> model();

  // add some control points
  for(int i=0;i<10;i++){
    model.add_control_point(Pose3_(Key('p', i)));
  }

  size_t n; 
  std::vector<Pose3> measurements;

  // construct an Expression that samples the trajectory at the Nth timestamp
  Pose3_ sample = model.sample_trajectory(Double_(Key('t', n)));

  // construct the factor
  f = ExpressionFactor(
    /*noise model*/
    measurement[n],
    sample,
  );
 // add f to the factor graph
```

## Features:
* Differentiable cardinal splines up to 7th order (higher orders are easy to add).
* Nth derivatives of splines exposed as Expressions (necessarily including Jacobians)
* spline abstraction also covers FIR filters
* computational complexity can use lower bound of  `O(spline_order+1)`
* unknown timestamps represented with computational complexity of  `O(spline_order + timestamp_search_window)`
* Template type for `CartesianProduct`s of existing manifolds (like a std::pair but inheriting from LieGroup).
* Template type `AsVectorSpace` for extrapolating `VectorSpace` features from `Manifold` traits
* N-dimensional cardinal spline meshes (surface patches are a happy accidental feature)

### Miscellany:
* Expression template for `expmap` (probably needs a quick test)

### unused features that can be be removed from the PR:
* Irwin Hall Probability Density Function (only the CDF is required)
* Expression templates extended for operators on `Double_` types

## Implementation
Spline implementation follows [a CVPR paper](https://openaccess.thecvf.com/content_CVPR_2020/papers/Sommer_Efficient_Derivative_Computation_for_Cumulative_B-Splines_on_Lie_Groups_CVPR_2020_paper.pdf) where the weights are pre-computed, and the weighted sum happens in the Lie Algebra using the first point (of the points in the window) as the datum.
This results in an execution that can act on a minimal number of control points, considers each point once, and has a trivial derivative propagation rule.
The weights for (uniformly sampled) splines are captured by the Irwin Hall Cumulative Probability Density Function.
The (uniformly sampled) weighted sum happens to be structurally equivalent to a Finite Impulse Response (FIR) Filter.

## Questions:
* Do the navigation and IMU features set down a type convention for angular velocity? and does it differ from `traits<Rot3>::TangentVector`?
* `traits<T>::TangentVector` is a sensible type for first derivatives of `T`, but is it an adequate type for second and Nth derivatives?
* The spline curve system is implemented under a type called `TrajectoryModel`, but it's capable of defining surface-patches. what's a better name for it?
* What wrapping API would be required to use this from Python?
  * add keys to serve as control points?
  * generate factors?

## Missing Tests:
* cross-check against the [fixed-order implementation by @koide3](https://github.com/koide3/gtsam_points/blob/master/src/gtsam_points/util/bspline.cpp)
* ~Analytic derivatives of Rot3 spline curves match numerical derivatives to Nth order~
* Let me know if there's anything that you notice

## Missing Examples:
* Factor graph associating raw IMU data directly to derivatives of the trajectory in order to recover the time offset between non-monotonic, asynchronous, Pose3 and IMU data-sources [Furgale 2013](https://doi.org/10.1109/IROS.2013.6696514)

## Missing Features:
* ~`Cal3` doesn't have an expmap, so powered zoom lenses can't be built into a coordinated motion model using `TrajectoryModel<CartesianProduct<Pose3, Cal3> >`  A Riemannian embedding using the `localCoordinates` call would be (barely) enough like a logmap for this spline to function.~  See `AsVectorSpace`


dellaert:
@BrettRD this is *amazing*!

Here are my overall (Codex-assisted!) comments *without* yet trying to understand the code in detail. The first two comments are obvious/easy as they relate to style/comment currently requested for GTSAM PRs. The third one is way less obvious and might need a face-to-face meeting so I can help think through it with you:

## Code style
- the files can go straight in the basis folder, not a big fan of further cascading the folder structure 
- Please use PascalCase for class names and camelCase for methods and member variables to match the rest of the library. It’s basically Google but no snake_case
- Ensure all files pass clang-format with the repository’s configuration .clang-format (based on Google style). Running clang-format -i on changed files before committing will keep the style consistent.

## Documentation
Please add Doxygen comments to new public classes, methods, and major typedefs in header files. Briefly describe the intent, parameters, and return values; this improves discoverability and helps future contributors.

## The Basis framework
### Primer on the gtsam/basis framework
Basis.h defines a CRTP base class Basis<Derived> that supplies utility functors:
- CalculateWeights / DerivativeWeights (static in each derived class) compute the basis-function values and derivatives for a given order N at point x.
- EvaluationFunctor and VectorEvaluationFunctor evaluate a basis at runtime and return optional Jacobians.
- WeightMatrix builds matrices of basis weights for multiple evaluation points.
Specific bases (e.g., Chebyshev1Basis, Chebyshev2Basis, FourierBasis) implement these static methods and typedef Parameters to encode their coefficient structure.
BasisFactors.h wraps these functors into factor graph factors (EvaluationFactor, VectorEvaluationFactor, etc.) that enforce measurements on scalar or vector-valued parameterizations.
### Integrating your spline work
- If the spline uses a new polynomial or kernel basis, create a MySplineBasis class inheriting from Basis<MySplineBasis> and implement CalculateWeights and DerivativeWeights.
- Expose evaluation through the provided functors so you can reuse BasisFactors or FitBasis for optimization and interpolation.
- Reuse existing naming and spacing conventions (e.g., MySplineBasis::Parameters) and keep the interface consistent with current basis classes—this will allow existing tooling (factors, tests, potential Python bindings) to work with minimal adjustments.

That being said, we should have a chat about what that might look like here, and whether it fits your code.

dellaert:
PS calibrations objects satisfy the manifold traits, but not (yet) Lie group. It does not make obvious sense to “compose” 2 calibration objects, and reading the CVPR paper the cumulative spline representation seems to demand composition and inverse.

You could, in an application, “hack” this by deriving from a Calibration type, adding the composition / addition, and make that a Lie group / vector space, with corresponding traits.

BrettRD:
I'll have another read of the basis framework. On a first read, it looked like a lot of machinery to handle butterfly calculations in bounded domains and it wasn't clear if any of it could be adapted to single evaluations of FIR taps.

For Cal3, I suspect every choice of manifold embedding is going to be equally arbitrary, so I'm tempted to make a simple CRTP adaptor to hack Vector-space traits onto the side of Manifold, based on a naive Reimannian embedding.
That way, Manifold classes can remain only Manifold, and users who to need to pretend it's a more sophisticated manifold can apply a wrapping of `Reimannain<Cal3>`,  making the hackyness an explicit user choice, while still being a well-behaved and easy-to-use type.

dellaert:
> I'll have another read of the basis framework. On a first read, it looked like a lot of machinery to handle butterfly calculations in bounded domains and it wasn't clear if any of it could be adapted to single evaluations of FIR taps.

Indeed, Basis is designed to create functors (in BasisFactors.h) at single "taps" in bounded domains. But, if you have a trajectory with $t\in[a,b]$, you do have a bounded domain. I think the issue is that CalculateWeights should yield a sparse matrix in the case of splines.

> For Cal3, I suspect every choice of manifold embedding is going to be equally arbitrary, so I'm tempted to make a simple CRTP adaptor to hack Vector-space traits onto the side of Manifold, based on a naive Riemannian embedding. That way, Manifold classes can remain only Manifold, and users who to need to pretend it's a more sophisticated manifold can apply a wrapping of `Riemannian<Cal3>`, making the hackyness an explicit user choice, while still being a well-behaved and easy-to-use type.

I like it. Maybe `AsVectorSpace` ? It definitely "works" - I think we had addition at one point - but it felt wrong. They are not even groups. For the simplest $K$ calibration, they *are* all affine transforms, but already $K^{-1}$ is not a valid calibration.

BrettRD:
> I like it. Maybe `AsVectorSpace` ? It definitely "works" - I think we had addition at one point - but it felt wrong. They are not even groups. For the simplest K calibration, they _are_ all affine transforms, but already K − 1 is not a valid calibration.

Something like [that](https://github.com/borglab/gtsam/pull/2212/commits/7cab6c8da078f4ab28fd6ed2892837f259ad3f94). I really ought to split this PR

> The third one is way less obvious and might need a face-to-face meeting so I can help think through it with you

I sent an email to your gatech address, is that the best way to reach you?

dellaert:
I liked many of Copilot's suggestions. Two more comments (and I might have more):
- changes in gtsam/slam/expressions.h should probably move to gtsam/nonlinear/expressions.h
- still many snake_case variables rather than camelCase

BrettRD:
The majority of that copilot output seems to be about the example (which is only illustrative at this point and certainly won't compile)
The only remaining snake case outside of the example are stray local variables in tests, I'm holding off on any more style polishing until we can discuss some proposed changes to the Basis.h API so I can refactor everything in one go.

My last email might also be in your spam folder

dellaert:
FYI: saw this paper advertised on Linkedin https://arxiv.org/abs/2504.11580

BrettRD:
> FYI: saw this paper advertised on Linkedin https://arxiv.org/abs/2504.11580

Wonderful!
It's good to see the method being used in a Kalman filter, I've only used it in one-shot factor-graphs.

Their cubic basis function in (3) exactly matches the IrwinHall3 struct with a time-shift for each sample.
https://www.desmos.com/calculator/ucaaqvbevx
Calling out to a known distribution allows us to generalise the method over higher (and lower) order splines

Their approach to differentiate the polynomial that populates the time vector (7) is the same as in [PiecewisePolynomial.h](https://github.com/borglab/gtsam/pull/2212/files#diff-b0258cfe229a3046ae12c04ac239e40ac46e4c9f6dae3d30ed6f85aa910366b8R96-R107), giving it the connection to pseudo-spectral methods (pay quantisation errors up front, in exchange for exact derivatives thereafter)

I'm surprised to see them running 100 control points per second, that seems a bit much, though I haven't seen a (correct) analysis of how the bandwidth of the curve scales with control-point frequency and spline order.  The piecewise approximation of a Gaussian is a fairly aggressive low-pass filter, and insufficient control point density can make the SO3 component wrap when trying to describe high angular accelerations.
I'd love to see an analysis that rigorously connects this method to Grace Wahba's smoothing splines, including how gtsam factors can be used to describe smoothness priors.

I haven't forgotten about the changes we discussed, I haven't had a suitable block of time to get familiar with Eigen sparse matrices and plan out how to concatenate block-wise parameters into a contiguous model

dellaert:
codex: This adds differentiable Lie-group-aware spline trajectories, supporting mathematical types, expressions, tests, and an example. Labeled enhancement because it introduces substantial new continuous-time trajectory functionality.

dellaert:
@BrettRD, I used Codex to pick up your work from this PR, merge current develop, and address the outstanding review feedback in the new draft PR #2689. I created a separate PR rather than pushing to your fork. Are you okay with this direction? If so, we can close this PR, review the new one, and decide whether to merge it as-is or continue iterating. Thank you for the original work.

BrettRD:
@dellaert Thanks for picking this up, switching to a new PR https://github.com/borglab/gtsam/pull/2689 is fine by me

PR #2216, opened 2025-08-13
Create eval folder

p-zach:
Add eval folder for Monte Carlo evaluation of relevant files.

Currently the Accelerating scenarios pass, but the EuRoC data scenarios do not pass. The EuRoC scenarios are much more difficult of course. Maybe the tolerance should be adjusted.

My machine refuses to run the eval/test folders as test projects, so I cannot verify that the paths to the data for the Euroc tests work, and I suspect they will not. Hopefully someone with a cooperating machine can edit to the correct paths.


p-zach:
Also, the euroc data is 61mb. Can delete the unused files (MH02...V201) if we want to save on clone/download space.

dellaert:
> Also, the euroc data is 61mb. Can delete the unused files (MH02...V201) if we want to save on clone/download space.

Let’s talk tomorrow! We can definitely not add these files through a GTSAM (sefinitely not all of them). Is there another way, e.g., to have a download happen when a script is run for the first time?

gtsam/navigation/Scenario.h
```diff
@@ -115,15 +115,18 @@ class GTSAM_EXPORT DiscreteScenario : public Scenario {
    * @param angularVelocities_b Map of timestamps to ground truth angular velocities in body frame.
    * @param velocities_n Map of timestamps to ground truth linear velocities in nav frame.
    * @param accelerations_n Map of timestamps to ground truth linear accelerations in nav frame.
+   * @param t The duration of time between the first timestamp and the last.
```
dellaert: kill *if* it can be calculated.

gtsam/navigation/Scenario.h
```diff
@@ -154,6 +158,8 @@ class GTSAM_EXPORT DiscreteScenario : public Scenario {
   Vector3 velocity_n(double t) const override;
   Vector3 acceleration_n(double t) const override;
   /// @}
+  
+  double duration() const;
```
dellaert, line 178: needs a comment

eval/evalImuFactor.cpp
```diff
@@ -0,0 +1,86 @@
...
+  EXPECT(assert_equal(estimatedCov, pim.preintMeasCov(), 0.01));
+}
+
+/* ************************************************************************* */
+TEST(ImuFactor, EurocDataHard) {
+  const string data_path = ".\\data\\euroc\\euroc_V202.csv";
+
+  const DiscreteScenario scenario = DiscreteScenario::FromCSV(data_path);
+
+  ScenarioRunner runner(scenario, testing::Params(), 1.0 / 200.0);
```
dellaert: Make 200 a constant at top of file and say
`double dt = 1/kImuFreq;`

dellaert:
@p-zach let me know when I should re-review ;-)

p-zach:
@dellaert Looks like all the tests should pass, so ready for review.

The one issue is that the [euroc data](http://robotics.ethz.ch/~asl-datasets/ijrr_euroc_mav_dataset/machine_hall/MH_01_easy/) is actually stored as a zip file, so if we're doing a data download we would have to have the users download all 2.5gb of data (MH01 + V202) just to extract about 13mb of .csv data. Currently I have deleted the unused .csvs and have just kept the two used ones totaling 13mb. If neither of those options is suitable, we could self-host just the .csvs on a borglab github repo and have the users download from there. Or, we could take out a couple thousand lines from each of the datasets to cut down on size.

Let me know your thoughts.

varunagrawal:
This may not be my place but having a separate "examples"/"eval" repo instead of making it monolithic would probably be better since it decouples data and code dependencies. This leads to a lot of benefits such as not having to compile gtsam (can do an apt install) and separate issue tracking and pull requests.

It also makes contributing easier, especially for less experienced developers since they only need to focus on the relevant application.

This is seen in a lot of open source repositories (like pytorch and Jax). Jax is probably the best example right now given how it has separate repos for every component (such as optax, flax, etc).

In this scenario, a separate repo would also make finding documentation easier since it could all be linked from the readme (including data download links).

Another example is GTDynamics. We tried following the mono repo system there and now we have open PRs on subcomponents like the cable robot which will probably never get merged, leaving the repo in a semi-limbo state.

dellaert:
I like that idea! It would resolve the data bloat issue as well. I’ll discuss with @p-zach on slack

varunagrawal:
I did this for the hybrid factor graph examples, and it allowed me to also experiment more and get much better results. Currently the results look like we ran iSAM on the continuous only version.

p-zach:
Thanks for the insight @varunagrawal. Moved additions to another repo. Closing this PR as all changes made in #2228 and https://github.com/borglab/imuFactors/pull/1.

PR #2219, opened 2025-08-16
exclude windows tests

talregev:
On my local I get a linkage error when I compile vcpkg on windows as it compile on the ci.
I want to run it here, see the failing tests, and exclude them, that vcpkg windows can still run, and later when you have time, fix the tests for windows.


talregev:
I will try to run on my repo first.
You can cancel the tests for this PR.

PR #2213, opened 2025-08-11
so3::Local and Kernels, + Gal3 J_r

dellaert:
I liked Copilot's summary:

> This PR introduces a comprehensive refactoring of the SO(3) Jacobian computation system in GTSAM, replacing the old `so3::DexpFunctor` and `so3::ExpmapFunctor` classes with a new unified `so3::Local` and kernel-based architecture. The refactoring aims to provide a more generic and mathematically elegant approach to computing Jacobians for Lie groups, particularly for SO(3) and its derivatives like SE(3), SE₂(3), and Gal(3).

Key changes include:
- Introduction of `so3::Local` class with lazy evaluation of coefficients and kernel-based Jacobian computation
- New `Kernel` and `InvJKernel` structs that provide `applyJacobian` functionality with Fréchet derivatives
- Comprehensive documentation explaining the mathematical foundations through kernels and their connection to Fréchet derivatives

Update: merged in #2215 with Gal3 refactor to re-trigger CI with cleaner history.

PR #2221, opened 2025-08-16
Simplified/lazy Local functor based on Kernels

dellaert:
Attempt to pass CI for (approved) #2213 by splitting it into this PR, which allows for all deprecated functors/functions, and a second (coming) PR that no longer uses the deprecated API.

### DIFF
--- a/gtsam/geometry/Gal3.cpp
+++ b/gtsam/geometry/Gal3.cpp
@@ -22,7 +22,7 @@
 
 
 #include <gtsam/geometry/Gal3.h>
-#include <gtsam/geometry/SO3.h>
+#include <gtsam/geometry/Kernel.h>
 #include <gtsam/geometry/Event.h>
 #include <gtsam/base/numericalDerivative.h>
 #include <gtsam/base/Matrix.h>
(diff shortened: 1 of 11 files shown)


dellaert:
@ProfFan could you take a look? *something* in this PR is making CI times go through the roof. I have a bunch of optionals in `Local` now, but I'm not sure that's it: I had a pimpl before where all those were in the .cpp file: same thing.

dellaert:
Mac is fine, BTW. Seems to be an ubuntu-only issue.

dellaert:
Well, @ProfFan , it was not the constexpr :-) Basically the only Ubuntu one to finish is
- [Special Cases CI / ubuntu-clang-quaternions Release (pull_request)](https://github.com/borglab/gtsam/actions/runs/17076303647/job/48418916726?pr=2221), Successful in 26m

All other Ubuntus will time out I think. Pretty good hint that :-)

dellaert:
One [Hail Mary](https://en.wikipedia.org/wiki/Hail_Mary_pass) after another, [EIGEN_STRONG_INLINE](https://github.com/borglab/gtsam/pull/2221/commits/d0da98b48f626593e7abbebc0e6ff195c6c68523) is the latest one.

ProfFan:
I tested on latest Arch + clang without issue, so it's Ubuntu 24.04 + new kernel stuff. I'll spin a Ubuntu 24.04 Docker tomorrow to investigate

dellaert:
Closing as overtaken by #2226

PR #2223, opened 2025-08-17
Remove cxx_flags

talregev:
After #2217 
Remove cxx_flags

After above PR that fix 2 warnings, we can now remove the flags that turn off the warnings.
If the warning will happen again, we will notice, and catch it on time.
See the comment I deleted:
```
          # This is due an error in linux:
          # error: argument 2 null where non-null expected [-Werror=nonnull]
          # return __builtin_memcmp(__first1, __first2, sizeof(_Tp) * __num);
          # Remove `${{ matrix.cxxflags }}` when the compilation error solved
```

### DIFF
--- a/.github/workflows/vcpkg.yml
+++ b/.github/workflows/vcpkg.yml
@@ -33,7 +33,6 @@ jobs:
             build_type: Release
             test_target: test
             binary_cache: /home/runner/.cache/vcpkg/archives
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Assume you know what you're doing, but maybe a more informative PR comment would help future readers.

talregev:
> Assume you know what you're doing, but maybe a more informative PR comment would help future readers.

Notice. I will add a comments.

PR #2224, opened 2025-08-18
Fix single exe linking

dellaert:
The flag `GTSAM_SINGLE_TEST_EXE` is supposed to build single executables, but it gave linking errors. This PR
- adds [[maybe_unused]] static to most things in header-only tests
- fixes some printing inconsistencies
- resolves some namespace name clashes

We might be able to rescue some test programs on vcpkg CI with this, as well.


dellaert:
@talregev I tried running check_discrete_program in vcpkg but no go. Still as mysterious as ever.

talregev:
> @talregev I tried running check_discrete_program in vcpkg but no go. Still as mysterious as ever.

Thank you for keep trying.

PR #2229, opened 2025-08-21
CI Improvements for speed

dellaert:
New configuration of the workflows to speed up CI (and emit less CO2):

   1. There will be one Boost run and one Boost installation on Mac, specifically for the macos-14-xcode-15.4-boost job in build-macos.yml.
   2. There will be one Boost run and one Boost installation on Linux, specifically for the ubuntu-24.04-clang-16-boost job in build-linux.yml.

The build-special.yml workflow will not install or run Boost in any of its runs now.

Also:
- run vcpkg run on push, as suggested by @talregev 
- use mold/gold linker for Linux clang/gcc, respectively - faster linking


talregev:
another suggestion to speed up runs, is let vcpkg also run on push, like deploy workflow.
That way the cache will save for all repo, and it will not start over on every PR. vcpkg build will start over on vcpkg update, that happen once per 2 weeks.
It will save ~20min for each vcpkg run.

https://github.com/borglab/gtsam/blob/develop/.github/workflows/deploy.yml#L5-L7

dellaert:
Happy to do so, @talregev ! What change would I make in which workflow?

talregev:
> Happy to do so, @talregev ! What change would I make in which workflow?

I will make another PR and we can discuss it there.
Thank you for considering my suggestion!

.github/copilot-instructions.md
```diff
@@ -0,0 +1,9 @@
+For reviewing PRs:
+* All functions in header files should have doxygen-style API docs
+* Use /// for single-line comments rather than /** */
+* Use meaningful variable names, e.g. `measurement` not `msm`, avoid abbreviations.
```
jlblancoc, line 4: This could be even "enforced" by using clang-tidy if so wanted at some point...

jlblancoc (approved):

.github/copilot-instructions.md
dellaert: I would *love* that :-)

PR #2230, opened 2025-08-21
Add vcpkg on push, it will save the vcpkg cache for all repo

talregev:
Add vcpkg on push, it will save the vcpkg cache for all repo.

This is also mean that every merge a PR (a push) it will trigger the vcpkg ci. (3 vcpkg jobs).


dellaert:
Awesome. I canceled all CI, will include change in #2229

PR #2231, opened 2025-08-21
GalileanFactor !

dellaert:
Stubs for @giuliodelama 

@giuliodelama please continue to work on ths branch, but only `push` when you need review or want to run CI.


gtsam/navigation/tests/testCombinedImuFactor.cpp
```diff
@@ -50,9 +51,13 @@ static std::shared_ptr<PreintegratedCombinedMeasurements::Params> Params(
 }  // namespace combined
 
 /* ************************************************************************* */
-TEST_PIM(CombinedImuFactor, PreintegratedMeasurements ) {
+TEST_PIM(CombinedImuFactor, PreintegratedMeasurements) {
   // Linearization point
-  Bias bias(Vector3(0, 0, 0), Vector3(0, 0, 0)); ///< Current estimate of acceleration and angular rate biases
+  Bias bias(
```
oursland, line 56: ???

gtsam/navigation/GalileanImuFactor.cpp
```diff
@@ -0,0 +1,91 @@
...
+  wInput(9) = 1.0;  // time scale
+  Matrix10 Jl;
+  Gal3 wExp = Gal3::Expmap(-wInput * dt, Jl).inverse();
+  Matrix10 preintMatrixAdjoint = preintMatrix_.AdjointMap();
+
+  // Propagate mean - from eq. (22) in the paper
+  preintMatrix_ = preintMatrix_.compose(wExp);
+
+  // Propagate covariance - from eq. (35) in the paper
+  // IMPORTANT: the covariance propagation is based on the Right-Invariant (RI)
```
dellaert, line 57: Are these outdated now? I think you switched to LI?

dellaert:
codex: This develops a Galilean IMU factor and supporting preintegration, tests, and build updates. Labeled enhancement because the primary purpose is to add a new navigation factor capability.

actionchunkscheduling (changes requested):
edit from rohan - mistakenly made this review from the wrong account. please disregard and/or delete this 😅

rohan-bansal:
Pulled it down, it’s easy to use. Thumbs up from me! Though I defer to @giuliodelama as the paper's author.

I'm also thinking that adding Python bindings should be pretty straightforward for this: expose the new `PIM` in `navigation.i` and instantiate `ImuFactorT` for it. I can try this out in a new PR after the 4.3 release maybe!

gtsam/navigation/tests/testCombinedImuFactor.cpp
```diff
@@ -53,9 +53,13 @@ static std::shared_ptr<PreintegratedCombinedMeasurements::Params> Params(
 }  // namespace combined
 
 /* ************************************************************************* */
-TEST_PIM(CombinedImuFactor, PreintegratedMeasurements ) {
```
giuliodelama, line 56: `RUN_GALILEAN_PIM_BACKEND` pairs `PG` with `CL`, so the fourth instantiation of
this test compares a *Galilean* three-way PIM against a *LieGroup* combined PIM.
The body asserts they produce the same mean at `tol = 1e-6`, which is a real
check for the three matched backends and a non-property for this one - the
Galilean mean is deliberately different (and about 2x more accurate).

It passes only because the data here is degenerate: `measuredAcc = (0.1,0,0)` is
parallel to `measuredOmega = (pi/100,0,0)`, so `Omega*a = 0` and
`Gamma_1(a) = a`, `Gamma_2(a) = a/2` exactly, and the two coincide bit for bit.
Rotating the acceleration by 90 degrees gives max|dV| = 3.9e-4, i.e. 390x the
tolerance. For contrast the matched `<PL,CL>` pairing agrees to 0.0 either way.

Suggest excluding the Galilean backend here, as already done for
`ErrorWithBiases`:

```suggestion
TEST_PIM_WITH_COMBINED_BACKEND(CombinedImuFactor, PreintegratedMeasurements) {
```

gtsam/navigation/tests/testCombinedImuFactor.cpp
giuliodelama: Same root cause as above: with `<PG, CL>` this compares the Galilean 9x9
covariance against the LieGroup combined one's top-left block, using
`assert_equal`'s default 1e-9 tolerance.

The matched backends agree to 4.6e-27 - it is the same arithmetic. The Galilean
pairing agrees to 1.5e-11, sixteen orders of magnitude worse, and passes only
because the numbers in this test are small (dt = 0.01, |omega| = 0.04). Changing
`deltaT` to 0.1 gives 1.5e-9 and the test fails.

```suggestion
TEST_PIM_WITH_COMBINED_BACKEND(CombinedImuFactor, SameCovariance) {
```

giuliodelama (approved):
This looks really nice! I've gone through the notebook and gtsam/navigation/GalileanImuFactor.*. I checked with Claude's help, and the implementation is correct and builds on the paper's idea of using Gal(3) together with the tangent group structure. The thing I most wanted to be sure of was A_LI and B_LI, so I checked them by numerically differentiating the exact nonlinear error map, keeping the notebook's own error coordinates. Numerical tests with Claude are all ok (I can share the scripts). I've left two small fixes on the test harness inline.

There is one thing I would rather discuss than request a change on, and it is about how §5 defines the error.

The notebook takes the tangent group Gal(3) |x_Ad gal(3) from the paper, and the lift it uses is exactly the one the paper's action induces, so that part is faithful. What changes is which group error gets formed. The paper builds X Xhat^-1; the notebook builds Xhat^-1 X, because GTSAM retracts on the right. But the paper's action on the state space is a right action, and under a right action only the right group error is invariant: acting on the true and the estimated state with the same group element leaves X Xhat^-1 untouched and turns Xhat^-1 X into Y^-1 (Xhat^-1 X) Y. So the object §5 works with is a coordinate chart rather than an equivariant error, and I could not find a symmetry of this system that would make it one.

The semidirect twist in the tangent group is there to transport between two frames. The right-invariant navigation error lives in the spatial frame while the bias lives in the body frame, and Ad_Upsilonhat carries you between them. In the left-invariant convention, the navigation error is already a body-frame quantity, in the same frame as the bias, so there is nothing left to transport. Keeping the twist anyway does not add geometry; it just puts ad_bhat into every block and flips the sign of the bias coordinate.

The alternative is the direct product Gal(3) x gal(3), with the bias as an ordinary abelian factor instead of a semidirect one. The system is equivariant under its left action (Gamma, c)·(Upsilon, b) = (Gamma·Upsilon, b + c) with w̃ -> w̃ + c, and the left-invariant error is simply

Xhat^-1 X  =  ( Upsilonhat^-1 · Upsilon ,  b − bhat )

with no adjoint anywhere. Its logarithm is (e, b − bhat) exactly, and the propagation collapses to

A = [ Ad_(V^-1)     −C ;   0      I  ]        B = [ −C     0  ;   0   −h·I ]

which is the exact Ad_Upsilonhat mirror of Eqs. (36)-(37). In that chart the conditioned subspace §6 refers to is literally {d = 0} rather than the graph {d = ad_bhat·e}, and J_{k+1} = Ad_(V^-1)·J_k − C_k is just the accumulated (1,2) block. None of it touches the code: the conditioned 10x10 block and J_k come out identical either way.

dellaert:
@giuliodelama Could you please clarify your large comment? I Follow what you're saying, but If I'm correct, you're discussing multiple points at the same time. So maybe grouping them in a with some headings would just already clarify what you mean, And which points needs changes in code and which ones don't.

dellaert:
> @giuliodelama Could you please clarify your large comment? I Follow what you're saying, but If I'm correct, you're discussing multiple points at the same time. So maybe grouping them in a with some headings would just already clarify what you mean, And which points needs changes in code and which ones don't.

Never mind - I think I got it.

dellaert:
The Galilean formulation has now been rewritten around the direct-product symmetry suggested by @giuliodelama, while preserving GTSAM's right-retraction convention. We also don't need the virtual inputs anymore in this. Everything became so much simpler :-)

I also worked with the Ai to rewrite everything in terms of the equivariant filter much more clearly. The basic construction is:

- Physical state: $\xi=(\Upsilon,\beta)\in\mathrm{Gal}(3)\times\mathbb R^6$, with $\beta=(b_a,b_\omega)$.
- Symmetry group: $\mathcal G_L=\mathrm{Gal}(3)\times(\mathbb R^6,+)$.
- Left state action: $\phi_L((\Gamma,c),(\Upsilon,\beta))=(\Gamma\Upsilon,\beta+c)$.
- Left input action: $\psi_L((\Gamma,c),u)=u+c$.
- Equivariant error: $\phi_L(\hat Y^{-1},\xi)=(\hat\Upsilon^{-1}\Upsilon,\beta-\hat\beta)$, which is exactly GTSAM's left-invariant local error.

For a held IMU sample, let

$$
q(u,\beta)=(\omega-b_\omega,\ a-b_a,\ 0,\ 1),\qquad
V=\mathop{Exp}(q h),\qquad
C=J_R(qh)hE,
$$

where $E\in\mathbb R^{10\times6}$ maps GTSAM's $(a,\omega)$ ordering into Gal3's $(\omega,\nu,\rho,\alpha)$ ordering. The mean remains right-composed,

$$
\hat\Upsilon^+=\hat\Upsilon V,
$$

and the 16D left-invariant linearization is

$$
A_{\mathrm{LI}}=
\begin{bmatrix}\mathop{Ad}_{V^{-1}}&-C ; 0&I_6\end{bmatrix},
\qquad
B_{\mathrm{LI}}=
\begin{bmatrix}-C ; 0 \end{bmatrix}.
$$

The physical bias correction is consequently accumulated as

$$
J_{k+1}=\mathop{Ad}_{V_k^{-1}}J_k-C_k,
\qquad
\Upsilon(\hat\beta+\delta\beta)\simeq
\hat\Upsilon\mathop{Exp}(J\delta\beta),
$$

so the correction is applied on the right. The factor finally removes deterministic time and permutes Gal3 $(R,v,p,t)$ tangent order into `NavState` $(R,p,v)$ order. Bias evolution remains a separate factor.

The notebook now derives this construction in EqF terminology and includes an executable Python example using the new `PreintegratedImuMeasurementsG` and `GalileanImuFactor` bindings.

dellaert:
Added a reproducible four-backend NEES comparison in `gtsam/navigation/doc/GalileanImuFactorNEES.ipynb` (commit e8873828a).

The notebook compares the actual GTSAM Manifold, Tangent, Lie-group, and Galilean preintegration implementations. All four receive the same zero-order-held IMU signal and paired noise samples. The experiment uses 3,000 trials over 1 s at 20 Hz, with simultaneous specific force `[2, -1, 0.5] m/s^2` and angular velocity `[1.2, -0.8, 2.0] rad/s`. NEES is computed in the 9D factor residual using each implementation's own `residualCovariance()`, against an exact held-input Gal(3) endpoint.

For 3,000 trials, the exact aggregate 95% consistency interval for the mean 9D NEES is `[8.849, 9.152]`:

| Backend | Mean NEES | Position RMS | Velocity RMS |
|---|---:|---:|---:|
| Manifold | 14.506 | 0.0578 m | 0.1049 m/s |
| Tangent | 14.511 | 0.0578 m | 0.1049 m/s |
| Lie group | 14.506 | 0.0578 m | 0.1049 m/s |
| Galilean | **8.958** | **0.0427 m** | **0.0767 m/s** |

The deterministic timestep sweep also shows the Galilean mean update at floating-point accuracy for this held-input trajectory, while the three conventional variants have the expected finite-step discretization error. Their deterministic mean-error norm is about `0.0818` at the tested 20 Hz rate, versus `0.0011` for Galilean in the Monte Carlo sample mean.

The claim is intentionally scoped: this demonstrates the Galilean formulation's consistency and accuracy advantage for finite-rate, zero-order-held measurements with simultaneous rotation and acceleration. It is not a claim of universal superiority or lower runtime.

Supporting changes add Python-only named bindings for the three existing preintegration backends, a wrapper test covering all four, and links from the navigation documentation and Galilean factor guide.

Validation completed:

- `make -j6 python-install`
- `conda run -n py312 python -m pytest -q python/gtsam/tests/test_GalileanImuFactor.py` (`2 passed`)
- full notebook execution with all assertions passing
- HTML render with both Plotly figures present
- `git diff --check`

giuliodelama (approved):
That was exactly my point, looks good to me :)

PR #2232, opened 2025-08-22
Fix Boost and badges

dellaert:
- require boost to at least 70
- describe flags
- fix macros to omit system
- fix badges on readme
- python build after push to develop


dellaert:
@jlblancoc , @talregev the boost thing was an ordeal to get right but seems to finally work for everything.

jlblancoc:
@dellaert Awesome! All code and docs changes LGTM, and given that checks pass, what else could I add? :upside_down_face:  Thanks for relentlessly keep pushing :muscle:

jlblancoc (approved):

PR #2240, opened 2025-08-28
build vendor GeographicLib

talregev:
build vendor GeographicLib


gtsam/3rdparty/CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ if(GEOGRAPHICLIB_FOUND AND GTSAM_INSTALL_GEOGRAPHICLIB)
 endif()
 
 if(GTSAM_INSTALL_GEOGRAPHICLIB)
-  add_subdirectory(GeographicLib)
+  add_subdirectory(GeographicLib EXCLUDE_FROM_ALL)
```
talregev, line 84: `EXCLUDE_FROM_ALL` mean that the GeographicLib will not build with "all" target. also it mean it will not build everything in GeographicLib. Only when it needed (on navigation tests) and what it needed (the lib itself).
This save from the javascript compilation error that occur otherwise.

talregev:
These lines imply that GeographicLib is install when it not found the GeographicLib in system. In practice it not activate the install even if it not found the  GeographicLib in system.

I want to delete these line, and set the option to be default false. (similar the state it has today). 
If you approve, I will create another push for that. 

https://github.com/borglab/gtsam/blob/a36e89aec7b3d566d94225fe869dcadadc40ef52/gtsam/3rdparty/CMakeLists.txt#L71-L76

dellaert:
Actually, I was thinking of completely *removing* GeographicLib from GTSAM. I think it’s a pre-req for two examples and that’s it. We can exclude those examples and mention in the README you have to link them with GeographicLib…

talregev:
ok. I will close this PR.

PR #2241, opened 2025-08-29
IEKF Navstate motion propagation fix

drk944:
The current IEKF predict functions account for motion propagation in $SE(2)$ only. The current group composition equations and linearizations used do not apply to the NavState $SE_2(3)$ group.
The equations used for NavState motion propagation within the left-invariant EKF are defined in my function header and are explained in this tutorial:
```
E. R. Potokar, R. W. Beard and J. G. Mangelson, "An Introduction to the Invariant Extended Kalman Filter [Lecture Notes]," in IEEE Control Systems, vol. 44, no. 6, pp. 50-71, Dec. 2024, doi: 10.1109/MCS.2024.3466488. keywords: {Kalman filters;Mean square error methods;Stochastic processes;Jacobian matrices;Nonlinear systems;Tutorials;Error analysis;Open source software;C++ languages;Lie groups;Manifolds},
```
I also updated the IEKF_NavstateExample.cpp to use my new overloaded function. In the previous example the orientation and velocity of the vehicle would update after the predict step, but not the position.

## Evaluation:
I evaluated my fix against my own IEKF implementation that was used in this paper: [https://arxiv.org/abs/2506.10850](url)
```
D. Benham, E. R. Potokar, and J.G. Mangelson, (2025). Invariant Extended Kalman Filter for Autonomous Surface Vessels with Partial Orientation Measurements. arXiv. https://arxiv.org/abs/2506.10850
```
These graphs show the motion propagation only with no sensor updates.

This is using the original implementation. You can see that position remains constant, and the Z velocity does not account for gravity.
GTSAM InEKF Predict Only

This is what my IEKF outputs:
Python InEKF Predict Only

And now the GTSAM IEKF matches with the same results:
GTSAM InEKF Predict Only_Fix

\# _A note on the velocity estimates_
I've noticed that the velocity estimates do not match ground truth. In this instance ground truth is established via a Gazebo simulation and I assume they use a different underlying motion model which contributes to the drift as these simulations were run with no noise.

### DIFF
--- a/gtsam/navigation/InvariantEKF.h
+++ b/gtsam/navigation/InvariantEKF.h
@@ -108,6 +108,80 @@ namespace gtsam {
       predict(U, Q); // Call the group composition predict
     }
 
+        /**
+     * Predict step via tangent control vector and manual imu integration on NavState (SE_2(3)) group.
+     * Gravity vector added to allow overloaded function and account for example code 
+     * IEKF_NavstateExample.cpp that does not have gravity offset
+     *
+     * Motion propagation equations follow the formulation from:
+     * Potokar et al., "Invariant Extended Kalman Filtering for Underwater Navigation"
+     * Link to paper: https://ieeexplore.ieee.org/document/9444664
+     * Source code can be found here: https://bitbucket.org/frostlab/underwateriekf/src/main/src/system.py
+     * 
+     *   Motion Propagation:
+     *   R_new = R * Expmap(omega*dt)
+     *   v_new = v + (R * accel + g) * dt   
+     *   p_new = p + v * dt + ((R * accel + g) * dt^2) / 2
+     *   X_ = NavState(Pose3(R_new, p_new), v_new);
+     * 
+     *   Covariance Update:
+     *   A = motion linearization which is dependent on control u input
+     *   P_ = exp(A) * P_ * exp(A).transpose() + exp(A) * Q * exp(A).transpose() * dt
+     *   
+     * @param u Tangent space control vector.
+     * @param dt Time interval.
+     * @param Q Process noise covariance matrix.
+     * @param g Gravity vector
+     * 
+     * @authors Derek Benham
+     */
+    
+    void predict(const TangentVector&u, double dt, const Covariance& Q, const Vector3 g) {
+      const Rot3 R = this->X_.pose().rotation();
+      const Vector3 v = this->X_.velocity();
+      const Vector3 p = this->X_.pose().translation();
+      
+      Vector3 omega = u.template segment<3>(0);
+      Vector3 accel = u.template segment<3>(6);
+      
+      // Motion propagation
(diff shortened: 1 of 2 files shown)


dellaert:
I’ll take a look tomorrow. Our example probably does have a bug (although assuming no gravity is not a bug but a property of the example, as I think you agree). But I do see some immediate issues with the PR
- the InvariantEKF is templated, and your prediction assumes the type is NavState. That will not work :-)
- the dynamics are state dependent here, through dependence on R. But invariant should not depend on state, right? Should these dynamics instead be communicated to the LieGroupEKF?
- I’ll make a small simulation with the ScenarioRunner that should be trackable

drk944:
Thanks for the feedback! 

Regarding your state dependency question. In the invariant EKF, the system dynamics can still be state-dependent. The "invariant" property refers specifically to the error dynamics, which are linearized in a way that makes them independent of the trajectory for a specific class of functions called group affine.

Equation 21 in the InEKF tutorial paper explains this, showing that the linearized covariance update depends solely on the control input `u`, rather than the full state. [(https://ieeexplore.ieee.org/document/10752718)](url). The "Imperfect EKF" example within the paper shows that the IMU motion propagation model is group affine apart from bias in the gyroscope and accelerometer.

inekf_state_independent_error

To fix the templating issue, would it be appropriate to add an if statement inside the original templated predict function to check if `G` is of type `NavState` and run the specialized logic there? I've done a lot of basic C++ in my undergrad but I'm still learning about templates.

Let me know if that sounds like a good approach, or if you'd recommend something different.

dellaert:
@drk944 please see #2242 - we can discuss further there. I will close this PR, because I think the PR I created *based* on your code is more in line with the current structure. Also:
- I would love to have a cvs file with your data so we can reproduce the graphs above.
- I think the debate (which we can continue in new PR) is still whether the left-invariant version of IMU dynamics is in fact group invariant. At least, the Jacobians are not independent of state, and that to me makes the error dynamics also state-dependent. 
- The current "InvariantEKF" class in GTSAM demands the jacobian is just the adjoint, but perhaps that is too restrictive. If these PR's clear up that further, all the more power to us
- PS maybe @contagon could take a look at both PRS as well :-)

gtsam/navigation/InvariantEKF.h
```diff
@@ -108,6 +108,80 @@ namespace gtsam {
...
+             -omega(1),  omega(0),         0;
+
+      Matrix3 a_cross;
+      a_cross <<     0, -accel(2),  accel(1),
+              accel(2),         0, -accel(0),
+             -accel(1),  accel(0),         0;
+
+      // Left error A matrix
+      Matrix A(9, 9);
+      A << -w_cross,     zero,  zero,
```
dellaert, line 170: This was wrong, as GTSAM (maddeningly perhaps) uses (R,p,v) order in NavState.

PR #2242, opened 2025-08-30
NavState + IMU EKF

dellaert:
Create a new class that specializes LieGroupEKF to NavState driven by an IMU, and gravity.

This is based on Derek's draft PR #2241

Note I derived not from InvariantEKF but LieGroupEKF, as it seems the error dynamics are *still* dependent on the state, through v_body and g_body.

I think this means, maybe, that the left-invariant version here is *not* group-affine. But, we're pretty new to this, so let's discuss :-)

I want to also add wrapper and a notebook that mimics the new cpp example, which has an IMU "flying" in a circle.


dellaert:
Repeating comment from #2241:
I closed #2241, because I think this PR based on code by @drk944 is more in line with the current structure. Also:

- I would love to have a cvs file with your data so we can reproduce the graphs above.
- I think the debate (which we can continue in new PR) is still whether the left-invariant version of IMU dynamics is in fact group invariant. At least, the Jacobians are not independent of state, and so it *seems to me that this makes the error dynamics also state-dependent*.
- The current "InvariantEKF" class in GTSAM *demands* the jacobian is just the adjoint, but perhaps that is too restrictive. If these PR's clear up that further, all the more power to us
- PS maybe @contagon could take a look at both PRs as well :-)

gtsam/navigation/NavStateImuEKF.cpp
```diff
@@ -0,0 +1,92 @@
...
+  xi.setZero();
+  // Rotation, position, velocity in NavState
+  const Rot3& R = X.attitude();
+  const Vector3& v = X.velocity();
+
+  // Body-frame quantities needed for left-trivialized tangent
+  const Vector3 v_body = R.unrotate(v);          // R^T v
+  const Vector3 g_body = R.unrotate(n_gravity);  // R^T g
+
+  // Tangent vector components (dR,dP,dV)
```
dellaert: This might not match @drk944 code yet. I'll add a unit test and correct if needed

dellaert:
OK, the mean part of the code below now matches @drk944 code exactly (and the paper, I'd think), and it did prompt me to add a new `predictIncrement` in LieGroupEKF. Maybe should be `predictWithCompose`. *but* the (left-invariant) Jacobian is still very state-dependent, so still not InvariantEKF. 

```c++
NavState navStateImuDynamics(const NavState& X, const Vector3& gyro,
                             const Vector3& accel, double dt,
                             const Vector3& n_gravity,
                             OptionalJacobian<9, 9> H) {
  // Rotation and velocity
  const Rot3& R = X.attitude();
  Matrix3 D_vb_R, D_vb_v, D_gb_R;
  const Vector3& v_n = X.velocity();  // Has D_v_v = R !
  const Vector3 v_body =
      R.unrotate(v_n, H ? &D_vb_R : nullptr, H ? &D_vb_v : nullptr);
  const Vector3 g_body = R.unrotate(n_gravity, H ? &D_gb_R : nullptr);
  const Vector3 a_b_total = accel + g_body;

  // Construct increment directly as group element with body-frame p/v
  // increments
  const Rot3 dR = Rot3::Expmap(gyro * dt);
  const double dt2 = 0.5 * dt * dt;
  const Vector3 dp_body = v_body * dt + a_b_total * dt2;
  const Vector3 dv_body = a_b_total * dt;
  NavState U(dR, dp_body, dv_body);

  if (H) {
    Matrix3 dRt = dR.transpose();  // Jacobian of NavState::Create
    H->setZero();
    // position:
    H->template block<3, 3>(3, 0) = dRt * (D_vb_R * dt + D_gb_R * dt2);
    const Matrix3 D_v_v = R.matrix();  // Jacobian of velocity()
    H->template block<3, 3>(3, 6) = dRt * D_vb_v * D_v_v * dt;
    // velocity:
    H->template block<3, 3>(6, 0) = dRt * D_gb_R * dt;
  }

  return U;
}
```

dellaert:
Done with this for a bit :-) I added wrapping and uploaded a notebook. Prediction is perfect, but now there is an issue with update. Could be covariance that is to blame, or calculation of innovation, but not seeing it now.

dellaert:
OK, I had an error in my measurement Jacobian :-) But, silver lining, many more tests confirm correctness of update. The Notebook now works very well. e.g., position vs. ground truth:
image

gtsam/navigation/doc/NavStateImuEKF.ipynb
```diff
@@ -0,0 +1,285 @@
...
+      "]\n",
+      "p: 0 0 0\n",
+      "v: 0 0 0\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Minimal setup for NavStateImuEKF\n",
+    "params = gtsam.PreintegrationParams.MakeSharedD(9.81)  # gravity (m/s^2)\n",
```
scottiyio, line 107: Maybe add a block about what the simulation we are doing is? Or perhaps just a block mentioning that we are using NED due to the MakeSharedD(9.81) 

`We are modeling an IMU using NED coordinates as the navigation frame with sensor measurements in the body/sensor frame" ?

scottiyio (approved):

dellaert:
Note, I'm still working on this. I think the NavStateImuEKF is better done using techniques described here:

> Barrau, Axel, and Silvere Bonnabel. "Linear observed systems on groups." Systems & Control Letters 129 (2019): 36-42. Link: https://www.sciencedirect.com/science/article/pii/S0167691119300805

gtsam/navigation/tests/testNavStateImuEKF.cpp
```diff
@@ -0,0 +1,276 @@
...
+  NavState U = oldNavStateImuDynamics(X0, omega_b, f_b, dt, n_gravity, D_U_X0);
+  NavState X_old = X0.compose(U, D_X_X0, D_X_U);
+
+  // Check New dynamics function
+  Matrix9 A_ekf;
+  NavState X_ekf =
+      NavStateImuEKF::Dynamics(n_gravity, X0, omega_b, f_b, dt, A_ekf);
+
+  CHECK(assert_equal(X_old, X_ekf));
+  CHECK(assert_equal<Matrix>(D_X_X0 + D_X_U * D_U_X0, A_ekf));
```
dellaert, line 270: This line proves a hunch I had, that the old LieGroupEKF version, which is tested here, canceled the gravity pieces of the Jacobian A. PS D_X_X0 is Adjoint(inv(U)) and D_X_U = identity.

dellaert: More words: in the LeftLinearEKF, A is independent of the gravity update W, and the above proves:
```
A = Ad(inv(U)) + D_U_X0
```
where D_U_X0 is the Jacobian of the old dynamics. In fact, that is exactly what was implemented in the LieGroupEKF::predictWithIncrement.

gtsam/navigation/NavState.h
```diff
@@ -266,7 +266,25 @@ class GTSAM_EXPORT NavState : public MatrixLieGroup<NavState, 9, 5> {
   /// @name Dynamics
   /// @{
 
-  /// Integrate forward in time given angular velocity and acceleration in body frame
+  // Ψ: autonomous flow where velocity acts on position for
```
dellaert: Change to phi

gtsam/navigation/NavStateImuEKF.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+void NavStateImuEKF::predict(const Vector3& omega_b, const Vector3& f_b,
+                             double dt) {
+  if (dt <= 0.0) {
+    throw std::invalid_argument("NavStateImuEKF::predict: dt must be positive");
+  }
+
+  // Get gravity from params
+  const Vector3 n_gravity = params_->n_gravity;
```
dellaert: Just inline

gtsam/navigation/LeftLinearEKF.h
```diff
@@ -0,0 +1,93 @@
...
+            std::is_invocable_r<G, Phi, const G&>,
+            std::is_same<decltype(std::declval<Phi>().dIdentity()), Jacobian>> {
+  };
+
+  /**
+   * General left–linear dynamics using a φ functor and its differential at e.
+   * Returns W · φ(X) · U, and optional Jacobian A = Ad_{U^{-1}} Φ,  Φ := dφ|_e.
+   */
+  template <class Phi, typename = std::enable_if_t<is_automorphism<Phi>::value>>
+  static G Dynamics(const G& W, const Phi& phi, const G& X, const G& U,
```
dellaert, line 71: Maybe we want overloads for W=I

gtsam/navigation/NavStateImuEKF.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+NavStateImuEKF::NavStateImuEKF(
+    const NavState& X0, const Covariance& P0,
+    const std::shared_ptr<PreintegrationParams>& params)
+    : Base(X0, P0), params_(params) {
+  // Build process noise Q_ = block_diag(Cg, Ci, Ca)
+  const Matrix3& Cg = params_->gyroscopeCovariance;
+  const Matrix3& Ci = params_->integrationCovariance;
+  const Matrix3& Ca = params_->accelerometerCovariance;
+  Q_.setZero();
```
scottiyio, line 30: Is there not already a method somewhere that allows us to build this easier?

gtsam/navigation/NavStateImuEKF.cpp
```diff
@@ -0,0 +1,87 @@
...
+  // Get gravity from params
+  const Vector3 n_gravity = params_->n_gravity;
+
+  // Calculate W, phi, and U
+  const NavState W = Gravity(n_gravity, dt);
+  NavState::AutonomousFlow phi{dt};  // Φ: velocity acts on position
+  const NavState U = IMU(omega_b, f_b, dt);
+
+  // Scale continuous-time process noise to the discrete interval [t, t+dt]
+  Covariance Qdt = Q_ * dt;
```
scottiyio, line 64: Would this not be divides by dt? 

I would think normal way is our std deviation is divided by sqrt(dt) so then covariance is divided by dt

gtsam/navigation/LeftLinearEKF.h
```diff
@@ -0,0 +1,93 @@
...
+
+#include <type_traits>  // For std::conjunction, std::is_invocable_r, std::is_same
+
+namespace gtsam {
+
+/**
+ * @class LeftLinearEKF
+ * @brief EKF on a Lie group with a general left–linear prediction model.
+ *
+ * Discrete step: x⁺ = W · ψ(x) · U, with W,U ∈ G and ψ ∈ Aut(G).
```
dellaert: phi

gtsam/navigation/NavStateImuEKF.h
```diff
@@ -0,0 +1,102 @@
...
+   * Construct with initial state/covariance and preintegration params (for
+   * gravity and IMU covariances)
+   * @param X0 Initial NavState.
+   * @param P0 Initial covariance in tangent space at X0.
+   * @param params Preintegration parameters providing gravity and options.
+   */
+  NavStateImuEKF(const NavState& X0, const Covariance& P0,
+                 const std::shared_ptr<PreintegrationParams>& params);
+
+  // Calculate W (gravity-only left composition, world-frame increments)
```
dellaert: ///

gtsam/navigation/doc/InvariantEKF.md
```diff
@@ -13,8 +13,9 @@ The Invariant Kalman Filter operates on many Lie Groups commonly used in robotic
...
 
 - **[ManifoldEKF](https://github.com/borglab/gtsam/blob/develop/gtsam/navigation/ManifoldEKF.h)**: Implements an EKF for states that operate on a differentiable manifold.
-- **[LieGroupEKF](https://github.com/borglab/gtsam/blob/develop/gtsam/navigation/LieGroupEKF.h)**: Inherits from ```ManifoldEKF`` and implements an EKF for states that operate on a Lie group with state dependent dynamics.
+- **[LieGroupEKF](https://github.com/borglab/gtsam/blob/develop/gtsam/navigation/LieGroupEKF.h)**: Inherits from ```ManifoldEKF``` and implements an EKF for states that operate on a Lie group with state dependent dynamics.
 - **[InvariantEKF](https://github.com/borglab/gtsam/blob/develop/gtsam/navigation/InvariantEKF.h)**: Inherits from ```LieGroupEKF``` and implements EKF for states that operate on a Lie group with group composition (state independent) dynamics.
+- **[NavStateImuEKF](https://github.com/borglab/gtsam/blob/develop/gtsam/navigation/NavStateImuEKF.h)**: Specialization of ```LieGroupEKF<NavState>``` that integrates IMU to predict NavState increments; simple API for predict and generic updates. See user guide and tutorial linked below.
```
dellaert: Why triple ```?
Also, (a) add LeftLinear, (b) inherit Invariant from it. - also in navigation.i.

gtsam/navigation/NavStateImuEKF.cpp
dellaert: I don’t think so - but we can chat. It would be good to sanity-check this *and* the calculations of the Jacobian A using NEES in the ImuFactors repo.

gtsam/navigation/NavStateImuEKF.cpp
dellaert: I simplified it, easy as it gets.

dellaert:
A note on the LeftLinearEKF vs InvariantEKF. Note currently the `InvariantEKF` class does not allow giving an arbitrary Jacobian/Transition matrix, but the LeftLinearEKF should cover cases where we don't just multiply with a $U_k$:
image

varunagrawal:
I randomly came across this. Is there a reason we are not prioritizing landing #1613 and giving `NavState` the proper $SE_2(3)$ definitions? IIRC `NavState` is not a proper Lie Group at the moment and my aim with #1613 was to chunk the development of invariant filtering and smoothing in the following steps:

1. Make `NavState` $\in SE_2(3)$
2. Add the necessary machinery for the group affine property so we can generate the appropriate $A$ matrix (I believe this has been accomplished in the `InEKF` class).
3. Make some examples using both simulation and real data so we can confirm the working of it.

I won't be able to get to these until after September 25th, but just wanted to point it out so it can be considered while planning for  this.

dellaert:
That was done a while ago. The only things outstanding in #1613 were
some transpose operators that were not tested yet. Please check the
review comments. And, there is an undergrad - Warren - who volunteered
to do that.

Invariant filtering is indeed done, and now this new class, which goes
beyond it. And Jennifer is working on an equivariant template. there are
three c++ IEKF examples and one notebook, in this PR, with Left-linear.

Lets chat on Thursday.

dellaert:
Please note that this still uses a "body update" for the GPS measurement. Proposition 25 in B+B SCL paper seems to suggest we should think about how to incorporate world-frame measurements:
image

PR #2250, opened 2025-09-10
Documented NonlinearOptimizerParams in the NonlinearOptimizer doc

rafaelspring:
Just added a bit of documentation for the NonlinearOptimizerParams :)


dellaert (approved):
Many thanks!!!!

dellaert:
@talregev a strange, sudden ubuntu error on vcpkg.

talregev:
Can you retry the test?
I testing it, and it look ok for me:
https://github.com/talregev/gtsam/pull/46

for me, it now stuck on the test compilation (2 steps head). 
I think there is a slow ci machine for linux that it not have enough memory. If it will be to often, I will suggest to compile vcpkg ci linux with only one process compilation (with the flag -j1). it will be slower, but will guarantee more success.  

But I think we should try more time and see how it go in other PRs before switch to that solution. 


Edit: It actually finished the vcpkg linux ci in my test. it was just stack for couple minute that it didn't show the status.
But the same thing as above. I think there is a slow ci linux machine that can sometimes cause this error. 
and we should look how often it happen.

dellaert:
I re-ran that flow and it worked!

PR #2251, opened 2025-09-13
Gal3 IMU EKF class & header 

scottiyio:
New invariant filter that uses Gal3 for EKF pushed by IMU and gravity.

Uses the InvariantEKF instead of the LeftLinearEKF:
- there is no need for an automorphism when using Gal3
- We added a predict(W,U) method, which says `X_{t+1) = W X_t U`. The Jacobian does *not* depend on state *or* W, only on U, as we use the left-invariant error.

Behaves almost identical as NavStateImuEKF on ASV example by Derek:
image


gtsam/CMakeLists.txt
```diff
@@ -86,7 +86,9 @@ endforeach(subdir)
 
 # To add additional sources to gtsam when building the full library (static or shared)
 # append the subfolder with _srcs appended to the end to this list
-set(gtsam_srcs ${3rdparty_srcs})
+set(gtsam_srcs ${3rdparty_srcs}
```
dellaert: This is the AI flaking out :-) Not needed.

dellaert: In fact, harmful

gtsam/geometry/Gal3.h
```diff
@@ -217,6 +219,34 @@ class GTSAM_EXPORT Gal3 : public MatrixLieGroup<Gal3, 10, 5> {
   static TangentVector Vee(const LieAlgebra& X);
 
   /// @}
+  /// @name Dynamics
+  /// @{
+
+  // φ: autonomous flow where velocity acts on position for
```
dellaert: I think this is not needed for Gal3. The autonomous flow is implemented by the group product. That's what makes it so great :-)

gtsam/navigation/Gal3ImuEKF.cpp
```diff
@@ -0,0 +1,80 @@
...
+                                  const Vector3& omega_b, const Vector3& f_b,
+                                  double dt, OptionalJacobian<10, 10> A) {
+  if (dt <= 0.0) {
+    throw std::invalid_argument(
+        "Gal3ImuEKF::Dynamics: dt must be positive");
+  }
+
+  // Calculate W, phi, and U
+  const Gal3 W = Gravity(n_gravity, dt);
+  Gal3::AutonomousFlow phi{dt};  // Φ: velocity acts on position
```
dellaert: This should be a no-op automorphism, for now I'd just define that locally in this file. It is not needed in Gal3 and when we switch to InvariantEKF it'll be not needed at all.

gtsam/navigation/Gal3ImuEKF.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+void Gal3ImuEKF::predict(const Vector3& omega_b, const Vector3& f_b,
+                             double dt) {
+  if (dt <= 0.0) {
+    throw std::invalid_argument("Gal3ImuEKF::predict: dt must be positive");
+  }
+
+  // Calculate W, phi, and U
+  const Gal3 W = Gravity(params_->n_gravity, dt);
+  Gal3::AutonomousFlow phi{dt};  // Φ: velocity acts on position
```
dellaert: same

gtsam/navigation/Gal3ImuEKF.h
```diff
@@ -0,0 +1,107 @@
...
+   * Construct with initial state/covariance and preintegration params (for
+   * gravity and IMU covariances)
+   * @param X0 Initial Gal3.
+   * @param P0 Initial covariance in tangent space at X0.
+   * @param params Preintegration parameters providing gravity and options.
+   */
+  Gal3ImuEKF(const Gal3& X0, const Covariance& P0,
+                 const std::shared_ptr<PreintegrationParams>& params);
+
+  /// Calculate W (gravity-only left composition, world-frame increments)
```
dellaert: Seems like this copy/paste could be resolved with a small "mix-in" class, maybe `BaseImuEKF`

gtsam/navigation/Gal3ImuEKF.h
```diff
@@ -0,0 +1,107 @@
...
+  static Gal3 Gravity(const Vector3& n_gravity, double dt) {
+    return {Rot3(), 0.5*n_gravity*dt*dt, n_gravity*dt, dt};
+  }
+
+  /// Calculate U from raw IMU (no gravity): body-frame increments
+  static Gal3 IMU(const Vector3& omega_b, const Vector3& f_b, double dt) {
+    return {Rot3::Expmap(omega_b * dt), f_b * (0.5 * dt * dt), f_b * dt, dt};
+  }
+
+  /**
```
dellaert, line 95: same for these comments. `BaseImuEKF` could be templated and just document these *abstract* methods, dynamics and predict. Then here you just declare them as override with one liner doxygen.

gtsam/navigation/Gal3ImuEKF.h
```diff
@@ -0,0 +1,107 @@
...
+  const std::shared_ptr<PreintegrationParams>& params() const;
+  const Vector3& gravity() const;
+  const Covariance& processNoise() const;
+
+ private:
+  std::shared_ptr<PreintegrationParams> params_;
+  Covariance Q_ = Covariance::Zero();
+};
+
+}  // namespace gtsam
```
dellaert, line 143: Add newline after this line

dellaert (commented):
Cool. Some comments. Big comment: automorphism is not needed. That is my hypothesis at least, which can be unit-tested by comparing both filters.

dellaert:
I reconsidered my ‘mix-in” class comment. Maybe the way it is is perfect. There is a bit of duplication, but it’s very transparent and easy, which is probably more important at this point.

I do want to make sure it works with a “no-op” automorphism. If it does not I have completely misunderstood- and the new IMU factors is wrong, too…

dellaert (approved):
Awesome. I pushed some more commits, I think we can keep Scenario test for a next PR.

dellaert:
@giuliodelama check it out :-)

PR #2253, opened 2025-09-17
Remove GeographicLib as a 3rd party vendor

warrenyun:
Currently GeographicLib is only used in two files throughout the entire repository (in two unit test files related to navigation). It's likely better to ship GTSAM without this vendor installed by default, and instead add an optional CMake flag to build relevant unit tests if the user has it installed on their personal machine (re: discussion with @dellaert)

## Changes

* New `HandleGeographicLib.cmake` file that looks for any installed instance of GeographicLib in lib or in include directories
* Removal of GeographicLib from the 3rdparty list entirely
* CMake build flag to look for and use an installed GeographicLib instance if wanted


warrenyun:
Heads up: still making some primary changes to this PR. I just saw how GTSAM now handles Boost in regards to optional usage and I'm adopting that paradigm for GeographicLib as well

dellaert:
Ping?

warrenyun:
Almost done with this, in the middle of making sure this builds fine for other Linux distros that use FHS

dellaert:
@nebudev14 please make sure you don't reformat entire files in a PR: it makes it hard to see what is new in a PR.

dellaert:
Also, some test folders have logic to exclude tests when boost is not there, for example. You can copy that logic to make sure tests using geographiclib are included only if available.
Finally, there will need to be at least *one* CI workflow that installs Geographiclib and checks your logic.

dellaert:
@talregev Yet another persistent vcpkg failure - this time Ubuntu, something w Eigen.

talregev:
@dellaert 
There is a bug regarding eigen3 in vcpkg.
I create a PR with workaround https://github.com/borglab/gtsam/pull/2267

warrenyun:
Requesting workflow approval for this PR @dellaert

dellaert:
@nebudev14 which is the CI workflow I should look at?

warrenyun:
@dellaert Just the macos and ubuntu ones should have notable changes that build Geographiclib. I believe they should all pass now

gtsam/navigation/tests/CMakeLists.txt
```diff
@@ -1,32 +1,17 @@
...
+       testMagFactor.cpp)
 endif()
 
-# if GTSAM_ENABLE_BOOST_SERIALIZATION is OFF then exclude some tests
 if (NOT GTSAM_ENABLE_BOOST_SERIALIZATION)
   list(APPEND tests_excluded testSerializationNavigation.cpp)
 endif()
 
-gtsamAddTestsGlob(navigation "test*.cpp" "${tests_excluded}" "${test_link_libraries}")
+gtsamaddtestsglob(navigation "test*.cpp" "${tests_excluded}" "${test_link_libraries}")
```
ProfFan: no case change here

gtsam/navigation/tests/testGPSFactor.cpp
```diff
@@ -16,270 +16,270 @@
...
+ #include <GeographicLib/LocalCartesian.hpp>
+ 
+ using namespace std;
+ using namespace gtsam;
+ using namespace GeographicLib;
+ 
+ #if GEOGRAPHICLIB_VERSION_MINOR<37
+ static const auto& kWGS84 = Geocentric::WGS84();
+ #else
+ static const auto& kWGS84 = Geocentric::WGS84();
```
ProfFan, line 35: These two lines are the same in the ifdef

gtsam/navigation/tests/testGeographicLib.cpp
```diff
@@ -15,99 +15,99 @@
...
+ 
+ using namespace std;
+ //using namespace gtsam;
+ using namespace GeographicLib;
+ 
+ // Dekalb-Peachtree Airport runway 2L
+ static const double lat = 33.87071, lon = -84.30482, h = 274;
+ 
+ #if GEOGRAPHICLIB_VERSION_MINOR<37
+ static const auto& kWGS84 = Geocentric::WGS84();
```
ProfFan: ditto

ProfFan (commented):
Thank you for the PR! Some comments^

gtsam/navigation/tests/testGPSFactor.cpp
warrenyun: Didn't actually notice this initially, but I did a bit of digging and found out that this clause was meant to target the GeographicLib change in version 1.**37** which changed how the WGS84 constructor was handled. Since modern versions of Geographiclib are 2.0 and above, this ifdef becomes obsolete because it only checks the minor version and not the major version also.

dellaert (approved):
Amazing !!!

dellaert:
Will merge after CI passes. Geographiclib is very cool but too large and too little used to vendor in, so this is a great PR !

PR #2256, opened 2025-09-20
Fix tests and simplify Windows CI

Gold856:
MSBuild isn't used anymore, and Ninja isn't being used as a multi-config generator, so `--config` does nothing. `CMAKE_BUILD_TYPE` was being set as an environment variable, so that's why it's always built in the correct mode. The logic for detecting GCC/clang was removed because the Windows builds only use cl. All the tests pass on my local machine, but we'll see how it works on CI...

testDoglegOptimizer failed because `keys()` was being called multiple times, which means the iterators come from different vector objects, which trips an assertion in MSVC. The other failing test was nonlinear_unstable. For reasons I do not understand, dynamic_pointer_cast/dynamic_cast fails when building in Debug mode, but not Release mode. I added a preprocessor check to just use static_cast when the compiler is MSVC and Debug mode is being used. That way, all other builds keep the pointer check and will fail accordingly.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -70,25 +70,6 @@ jobs:
           scoop install cmake --global  # So we don't get issues with CMP0074 policy
           scoop install ninja --global
 
-          if ("${{ matrix.compiler }}".StartsWith("clang")) {
-            scoop install llvm --global
-          }
-
-          if ("${{ matrix.compiler }}" -eq "gcc") {
-            scoop install gcc --global
-            echo "CC=gcc" >> $GITHUB_ENV
-            echo "CXX=g++" >> $GITHUB_ENV
-
-          } elseif ("${{ matrix.compiler }}" -eq "clang") {
-            echo "CC=clang" >> $GITHUB_ENV
-            echo "CXX=clang++" >> $GITHUB_ENV
-
-          } else {
-            echo "CC=${{ matrix.compiler }}" >> $env:GITHUB_ENV
-            echo "CXX=${{ matrix.compiler }}" >> $env:GITHUB_ENV
-
-          }
-
           # Scoop modifies the PATH so we make the modified PATH global.
           echo "$env:PATH" >> $env:GITHUB_PATH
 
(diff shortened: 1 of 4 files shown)


dellaert:
Thanks, much simpler! CI does seem to pass :-)
I do have one question: it seems the script previously checked for gcc/clang - which surprised me a bit on windows - and now that logic is gone. Could you update the PR description to clarify what was going on what the fix was?

Gold856:
Done! Also adjusted the comment to not call out shared libraries in specific (also happens on static builds).

dellaert (approved):
Awesome! Windows FTW :-)

.github/workflows/build-windows.yml
```diff
@@ -140,47 +119,8 @@ jobs:
 
       - name: Build
         shell: bash
-        run: |
-          # Since Visual Studio is a multi-generator, we need to use --config
```
talregev, line 144: # Since Visual Studio is a multi-generator, we need to use --config
You should add --config Release to this command

.github/workflows/build-windows.yml
```diff
@@ -140,47 +119,8 @@ jobs:
...
-          # https://stackoverflow.com/a/24470998/1236990
-          cmake --build build -j4 --config ${{ matrix.build_type }} --target gtsam
-          cmake --build build -j4 --config ${{ matrix.build_type }} --target gtsam_unstable
-          cmake --build build -j4 --config ${{ matrix.build_type }} --target all.tests
+        run: cmake --build build -j4 --target all.tests
```
talregev, line 122: See the comment above you deleted:
> Since Visual Studio is a multi-generator, we need to use --config

You should add --config Release to this command

.github/workflows/build-windows.yml
```diff
@@ -140,47 +119,8 @@ jobs:
...
-          # Compile. Fail with exception
-          # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.nonlinear_unstable
-          # Compile. Fail with exception
-          # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.slam_unstable
-          # Compile. Fail with exception
-          # cmake --build build -j4 --config ${{ matrix.build_type }} --target check.partition
-
-          # Run all tests
-          # cmake --build build -j1 --config ${{ matrix.build_type }} --target check
+        run: cmake --build build -j1 --target check
```
talregev, line 126: You should add --config Release to this command

.github/workflows/build-windows.yml
Gold856: The Visual Studio generator is not the same thing as MSVC. We are using the Ninja generator with the MSVC compiler, which means --config is completely ignored.

PR #2254, opened 2025-09-18
Bug fix imu factors example

jaffarrell:
The first commit fixes minor syntax issues caught by strict compiler settings. I had to make these edits to compile and build at the "make check' stage of the install.

The second commit fixes /examples/ImuFactorsExample.cpp. The release version crashed at the end of the file. The new version runs all the way through.

PR #2260, opened 2025-09-28
Fix CI and run non-boost QP tests

dellaert:
Small PR with:
- Fix some fallout resulting from #2254 (I think)
- Fix cmake Boost exclude lists in unstable
- Run non-boost QP tests (that do not need a parser).


dellaert:
@talregev mac cvpkg seems to be failing - any ideas?

talregev:
@dellaert I create a PR with a Fix #2262

PR #2262, opened 2025-09-28
Fix vcpkg macos. Upload vcpkg logs on failure

talregev:
Fix vcpkg macos. Upload vcpkg logs on failure


dellaert (approved):
Awesome! Many thanks !!!!

PR #2267, opened 2025-10-03
temporary remove eigen3 from vcpkg

talregev:
temporary remove eigen3 from vcpkg.

There is a bug in vcpkg that someone update the eigen3 port incorrectly. This will effect all os in vcpkg.
The solution already in the latest master, but I am not advising to go there. 
For now I remove eigen3 port from installation, and when vcpkg will update in github action vm's we can bring it back.
should be around 2 weeks.

FYI @dellaert

### DIFF
--- a/.github/workflows/vcpkg.yml
+++ b/.github/workflows/vcpkg.yml
@@ -98,7 +98,6 @@ jobs:
           boost-system
           boost-thread
           boost-timer
-          eigen3
           tbb
           pybind11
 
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Thanks !

dellaert:
BTW, @talregev , I bought a spanking new windows machine - mostly for the GPU - and I built GTSAM with WSL. However, if you are willing to patch INSTALL.md with some beginners vcpkg instructions I’ll be happy to try it out. I don’t know how similar the instructions would be for Mac and windows,maybe this is a separate vcpkg section with only minor differences.

PR #2269, opened 2025-10-04
Add vcpkg compilation with gtsam in INSTALL.md doc

talregev:
Add vcpkg section in INSTALL.md doc.
Explain how to compile gtsam with vcpkg

@dellaert 
As for your request, I added a section in INSTALL.md doc that compile gtsam with vcpkg.
commands taken from CI and adapt for linux/wsl in local.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -261,3 +261,88 @@ or
 when importing GTSAM using the python wrapper.
 
 
+## Compile gtsam with vcpkg. (For linux/wsl)
+Install python dependencies + ninja + build essential in linux:   
+```
+sudo apt update
+sudo apt-get install autoconf automake autoconf-archive ninja-build build-essential -y
+```
+
+On your home dir near gtsam dir:   
+Setup vcpkg   
+```bash
+git clone https://github.com/microsoft/vcpkg
+cd vcpkg
+./bootstrap-vcpkg.sh
+```
+
+vcpkg install dependencies   
+on vcpkg dir:   
+```bash
+./vcpkg x-set-installed         \
+          boost-assign          \
+          boost-bimap           \
+          boost-chrono          \
+          boost-date-time       \
+          boost-filesystem      \
+          boost-format          \
+          boost-graph           \
+          boost-math            \
+          boost-program-options \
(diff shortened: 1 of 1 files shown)


dellaert:
Ah, maybe I miscommunicated. I do know how to install GTSAM with WSL, it’s  basically Linux. 
What I meant was maybe a section on vcpkg - which needs to be introduced, and seems to be defined for Mac/linux/windows. I will discuss WSL in the windows section, and then we can mention that as far as vcpkg is concerned, it’s the same as Linux.

dellaert:
Also, any comments on #2268 ?

talregev:
This section is explain how to install gtsam with vcpkg library on wsl / linux. Not with linux/wsl libraries.
for mac and windows there should be many small changes across this section.
You don't have to accept this PR, you can take it as a template and change it as you see fit.
Basically you have all the commands written on the CI. 

I just notice https://github.com/borglab/gtsam/pull/2268. Thank you for let me know. I will comment there also.

Gold856 (changes requested):
Remove `--config Release`. You've selected the Ninja generator, which isn't a multi-config generator, so `--config Release` is useless.

talregev:
> Remove `--config Release`. You've selected the Ninja generator, which isn't a multi-config generator, so `--config Release` is useless.

For Visual Studio (msvc) is a multi-generator, we need to use --config.
I removed the --config, but please add it when you explain how to compile in windows.

Gold856:
You're conflating the compiler with the build system. The Visual Studio generator (MSBuild) is not the same thing as MSVC. We are using the Ninja generator with the MSVC compiler, which means --config is completely ignored. If you do not pass `-G Ninja`, CMake will instead default to the `Visual Studio 17 2022` generator, which _is_ a multi-config generator, where `--config` can be used to select a build type.

talregev:
On the windows section we need to explain that we remove the --config because of the Ninja. Ninja is a single-config generator.  
If user choose to compile windows gtsam without the Ninja generator it must bring back the --config.

dellaert (approved):

PR #2273, opened 2025-10-07
Fix cephes static linking

Gold856:
Same fix I applied to METIS. Apparently the cephes header also contained declaration for C math functions? Removed those because we have those functions in the standard library.


talregev:
I like the way you handle cephes. but Remove C math function declarations should be in different PR that we can discuss and see.
What happen if you revert this commit in this PR?

Gold856:
It fails with a `inconsistent dll linkage` on those functions, because MSVC's standard libraries also export those functions: https://github.com/Gold856/gtsam/actions/runs/18264608447/job/51997148050#step:8:640

We should include `math.h` or `cmath` if we want to use those functions instead.

talregev:
> https://github.com/Gold856/gtsam/actions/runs/18264608447/job/51997148050#step:8:640

Can you show me the changes of this run? It look like static build errors.

Gold856:
The full branch is https://github.com/Gold856/gtsam/commits/windows-static, with that run being from the commit before I removed the C math function declarations. It also has the /permissive- changes so it can build up to that point.

talregev:
I think we need to solve one problem of the time. clearly you want to try to solve the static compilation. 
I think we need first to handle only the export of cephes first, that will help the direction you want also -> static compilation.
and discuss gtsam static compilation in different PR.

Gold856:
I showed you my static build branch instead of this PR is because you'll only notice the difference in static builds, which are the target of this PR. The additional changes are just so that it doesn't fail on other unrelated stuff. Yes, this includes the /permissive- changes because static linking is broken in other ways, as I showed in that PR.

I believe all the commits in this PR are both independent and small enough that they can both be reviewed independently and another PR is separate. The split would essentially be modifying the exports (to fix static linking) and removing the C math function declarations (to fix static linking.) That functionally doesn't really gain much, especially since removing C math function declarations shouldn't impact usage, as cephes is included by a header explicitly marked as internal. I think it makes more sense to keep it as part of one PR.

ProfFan:
What is cephes used for here? Only for Chi square right? If so, can we just replace the lib/inline it?

talregev:
I also have another suggestion.
Compile cephes as c++ and add a namespace

Gold856:
Yeah, it's only used for ChiSquaredInverse. This apparently isn't used anywhere except for [testGncOptimizer](https://github.com/borglab/gtsam/blob/33df753806d157f3f8d7f5977e6b1e8c2c329f9c/tests/testGncOptimizer.cpp#L634-L637)? Someone else is going to have to clarify the purpose of this function existing, because I'd vote to just outright remove cephes and this function.

As for inlining it, I tried to remove as much of cephes as possible, and got it down to 13 files:
image
We could consolidate it and put it into one big file that gets compiled into gtsam. We could compile it as C++ and provide a namespace, since there's a unit test for it, but I want to see if we can remove cephes outright.

varunagrawal:
The `GncOptimizer` class relies on `ChiSquaredInverse` to work. When we removed Boost as a dependency, GncOptimizer stopped compiling since it was using the `ChiSquaredInverse` provided by Boost.

I added Cephes since it was an easy and lightweight replacement and Luca Carlone was keen on having GncOptimizer working again (he spoke to me about this in person). There was no other implementation available and implementing the `inverse gamma` functions is far too much work.

If @ProfFan or any of you want to take it up, please feel free. I spent weeks trying to implement it without much success (especially with the full unit testing etc).

dellaert:
I'm not a  big fan of cephes either, and would welcome a local implementation of `ChiSquaredInverse`. But, heed @varunagrawal 's words :-)

Varun, is there a PR for that or a branch?

Gold856:
Ah, it's used in `setInlierCostThresholdsAtProbability`. I missed that my first time around. Then I think this PR is fine as-is. I certainly do not have the math knowledge to reimplement ChiSquaredInverse, but I can simplify down our copy of cephes to the core 13 files we need for `igami` if we want.

talregev:
better not changing the original files.
We can change only the header that export the symbols for the actual functions that gtsam is using.

varunagrawal:
> Varun, is there a PR for that or a branch?

For a local version of ChiSquaredInverse? I think I gave up and got rid of whatever hacky version I had (which was nowhere near completion). This was also almost 2 years ago. :(

dellaert (approved):
LGTM - but I know not enough. Ready to merge then?

Gold856:
Go for it!

PR #2275, opened 2025-10-07
[WIP] New FastNoiseModelFactor with more efficient linearization

ProfFan:
About 10% faster in linearization with my microbenchmarks. I think there is still a large margin for optimization, but we should focus more on the linear solver...

### DIFF
--- a/gtsam/base/OptionalJacobian.h
+++ b/gtsam/base/OptionalJacobian.h
@@ -120,7 +120,27 @@ public:
                       "Expected: ") +
           "(" + std::to_string(Rows) + ", " + std::to_string(Cols) + ")");
     }
-  } 
+  }
+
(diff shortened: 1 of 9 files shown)


dellaert:
Fan, I hope you saw that the CI fails. This could be due to unrelated changes you made. Maybe those should be a separate PR :-)

i’m wondering whether it is necessary to have a completely new class or whether we can just have a different method to overload. Then we can deprecate the old method and publicize the new method. And this PR would also have much smaller change and be easier to review.

dellaert:
@ProfFan Let's talk Monday about my comment and whether that's a good or a bad idea, and why...

ProfFan:
We should revive this so that we can be even faster

PR #2276, opened 2025-10-07
updated pybind version for GTSAM 4.2

suveergarg:


ProfFan:
Thank you for the PR! We will address this in #2277 instead :)

Issue #2264, opened 2025-09-30
"pip install gtsam" doesn't enforce working combination of requirements for python bindings


timbarfoot:
## Description

It seems that a generic "pip install gtsam" does not install compatible requirements.  In particular, "numpy" gets installed and by default is a new version (e.g., 2.0.0) that seems to not be compatible with the C++ binary that the python bindings are linking to.  Downgrading "numpy" (e.g., to 1.26.4) seems to fix the issue, but this is not obvious and should either be documents or specified in "requirements.txt" so a compatible version gets automatically installed.

## Steps to reproduce

1. pip install gtsam   (in a python 3.9.6 venv - this automatically installed numpy 2.0.0 for me)
2. run any example - fails in a segmentation fault
3. downgrade numpy to 1.26.4 and the examples work as expected 

## Expected behavior

Examples should run properly without having to figure out that the wrong numpy gets installed.

## Environment

python 3.9.6 in venv

dellaert:
Thanks! I'll take a look.

dellaert:
FWIW, I’m on python 3.12 and numpy 2.2.6, and no issues :-/

dellaert:
And Mac CI installs 2.2.3

```
Collecting numpy>=1.11.0 (from -r /Users/runner/work/gtsam/gtsam/python/requirements.txt (line 1))
   Downloading numpy-2.3.3-cp313-cp313-macosx_10_13_x86_64.whl.metadata (62 kB)
```

dellaert:
Tim responded on slack: "but you’re probably building from scratch - for my use case I need an easier install so just using “pip install gtsam” - I did try several different versions of python (3.9, 3.10, 3.11, 3.12, 3.13) and they all seg fault…I only tested my working solution to downgrade numpy on 3.9"

timbarfoot:
yeah, just to be clear...the issue is getting the python bindings to link to the pre-built binary that gets installed with "pip install gtsam"

redheli:
i cannot reproduce the segmentation fault with `develop` branch
but do reproduce the issue in version `4.2` and have a fix

in short: old gtsam's pybind11(version 2.10) too old, can not handle numpy >2.0

i am not familar with gtsam pkg release, maybe someone has proper solution to apply the patch to fix old version

## how to reproduce (build from source with debuginfo)
```
$ git checkout 4.2
$ conda create -n gtsam_issue python=3.9.6
$ conda activate gtsam_issue
[...]

PR #2277, opened 2025-10-08
Update `wrap` to latest with pybind11 3.0.1

ProfFan:
Should resolve the problem with #2264 with `numpy` 2.0.


dellaert (approved):

dellaert:
Awesome !!!

PR #2281, opened 2025-10-09
Add tests to ensure centripetal accel is dealt with properly for IMUs offset from body frame

scottiyio:
- Base test of a vehicle driving in a circle 
- 3 IMUs; one at body frame, two offset at +/- 3m in Y axis in the front right down frame as if it was on a boom 
- Ensure that the IMU properly deals with both rotation + translation of sensor frame from body frame - all 3 IMUs should produce the same results once transformed into the body frame


dellaert (approved):
Great !

PR #2284, opened 2025-10-14
Cast Eigen::Matrix dimension type to int

varunagrawal:
When compiling GTSAM with clang (particularly for GPMP2) and `-Wc++11-narrowing` enabled (which clang is more strict about), I am getting the error
```
/opt/homebrew/lib/cmake/GTSAM/../../../include/gtsam/base/ProductLieGroup.h:100:33: error: non-type template argument evaluates to 18446744073709551615, which cannot be narrowed to type 'int' [-Wc++11-narrowing]
  typedef Eigen::Matrix<double, dimension2, dimension2> Jacobian2;
                                ^
```
since `dimension2` is of type `size_t` and the template type for `Eigen::Matrix` is `int`.

ChatGPT confirmed this is a signed vs unsigned int issue and I applied a `static_cast` as it recommended.

Also replaced `typedef` with `using` for Jacobian type aliases.

### DIFF
--- a/.github/workflows/vcpkg.yml
+++ b/.github/workflows/vcpkg.yml
@@ -2,7 +2,7 @@ name: vcpkg
 on:
   pull_request:
   push:
-    # Runs on pushes targeting the develop branch 
+    # Runs on pushes targeting the develop branch
     branches: [develop]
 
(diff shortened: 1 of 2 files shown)


dellaert (approved):
LGTM, thanks!

PR #2287, opened 2025-10-17
Use doublecross instead of cross(Cross(a, b)) in NavState

scottiyio:
Change cross(cross(a, b)) to doubleCross(a, b) and add documentation

Add a no rotation coriolis test to NavState

### DIFF
--- a/gtsam/navigation/NavState.cpp
+++ b/gtsam/navigation/NavState.cpp
@@ -444,14 +444,19 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
   const double dt2 = dt * dt;
   const Vector3 omega_cross_vel = omega.cross(n_v);
 
-  // Get perturbations in nav frame
+  // Because our navigation frames are placed on a spinning Earth, we experience two apparent forces on our inertials
+  // Let Omega be the Earth's rotation rate in the navigation frame
+  // Coriolis acceleration = -2 * (omega X n_v)
+  // Centrifugal acceleration (secondOrder) = -omega X (omega X n_t)
+  // We would also experience a rotation of (omega*dt) over time - so, counteract by compensating rotation by (-omega * dt)
+  // Integrate centrifugal & coriolis accelerations to yield position, velocity perturbations
   Vector9 n_xi, xi;
   Matrix3 D_dR_R, D_dP_R, D_dV_R, D_body_nav;
   dR(n_xi) << ((-dt) * omega);
   dP(n_xi) << ((-dt2) * omega_cross_vel); // NOTE(luca): we got rid of the 2 wrt INS paper
   dV(n_xi) << ((-2.0 * dt) * omega_cross_vel);
   if (secondOrder) {
-    const Vector3 omega_cross2_t = omega.cross(omega.cross(n_t));
+    const Vector3 omega_cross2_t = doubleCross(omega, n_t);
     dP(n_xi) -= (0.5 * dt2) * omega_cross2_t;
     dV(n_xi) -= dt * omega_cross2_t;
   }
--- a/gtsam/navigation/NavState.cpp
+++ b/gtsam/navigation/NavState.cpp
@@ -451,11 +451,7 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
   // We would also experience a rotation of (omega*dt) over time - so, counteract by compensating rotation by (-omega * dt)
   // Integrate centrifugal & coriolis accelerations to yield position, velocity perturbations
 
-
-
-
   Vector9 n_xi;
-
   // Coriolis (first order) acceleration corrections
   dR(n_xi) << ((-dt) * omega);
   dP(n_xi) << ((-dt2) * omega_cross_vel); // NOTE(luca): we got rid of the 2 wrt INS paper
@@ -477,7 +473,6 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
   dV(xi) = nRb.unrotate(dV(n_xi), H ? &D_dV_R : 0);
 
   // Assemble Jacobians
-
   if (H) {
     H->setZero();
     const Matrix3 Omega = skewSymmetric(omega);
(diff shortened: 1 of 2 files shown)


gtsam/navigation/NavState.cpp
```diff
@@ -444,14 +444,19 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
...
+  // We would also experience a rotation of (omega*dt) over time - so, counteract by compensating rotation by (-omega * dt)
+  // Integrate centrifugal & coriolis accelerations to yield position, velocity perturbations
   Vector9 n_xi, xi;
   Matrix3 D_dR_R, D_dP_R, D_dV_R, D_body_nav;
   dR(n_xi) << ((-dt) * omega);
   dP(n_xi) << ((-dt2) * omega_cross_vel); // NOTE(luca): we got rid of the 2 wrt INS paper
   dV(n_xi) << ((-2.0 * dt) * omega_cross_vel);
   if (secondOrder) {
-    const Vector3 omega_cross2_t = omega.cross(omega.cross(n_t));
+    const Vector3 omega_cross2_t = doubleCross(omega, n_t);
```
dellaert: I think switching to doubleCross is most beneficial if we also grab its Jacobians and use them below to replace some complexity in the Jacobian chaining...

scottiyio: I agree. I did a bit of tests and had one term not matching (probably because I didn't check myself and just trusted copilot) so I will look again tonight 

There's also some doublecrosses elsewhere I believe

scottiyio: Some of this code was rearranged. I wanted to only create doubleCross() once which might be the wrong approach. So H2_wrt_t is created even if we don't use it. Not sure.

gtsam/navigation/NavState.cpp
```diff
@@ -444,23 +444,35 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
   const double dt2 = dt * dt;
   const Vector3 omega_cross_vel = omega.cross(n_v);
 
-  // Get perturbations in nav frame
-  Vector9 n_xi, xi;
-  Matrix3 D_dR_R, D_dP_R, D_dV_R, D_body_nav;
+  // Because our navigation frames are placed on a spinning Earth, we experience two apparent forces on our inertials
```
dellaert: Maybe move this up before sig

gtsam/navigation/NavState.cpp
```diff
@@ -444,23 +444,35 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
...
+  // Integrate centrifugal & coriolis accelerations to yield position, velocity perturbations
+
+  Vector9 n_xi;
+  // Coriolis (first order) acceleration corrections
   dR(n_xi) << ((-dt) * omega);
   dP(n_xi) << ((-dt2) * omega_cross_vel); // NOTE(luca): we got rid of the 2 wrt INS paper
   dV(n_xi) << ((-2.0 * dt) * omega_cross_vel);
+
+  // Centrifugal (second order) acceleration corrections
+  Matrix3 H2_wrt_t; // To store Jacobian (if needed/desired)
```
dellaert: D_c2_nt

gtsam/navigation/NavState.cpp
```diff
@@ -444,23 +444,35 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
...
     dV(n_xi) -= dt * omega_cross2_t;
   }
 
-  // Transform n_xi into the body frame
-  xi << nRb.unrotate(dR(n_xi), H ? &D_dR_R : 0, H ? &D_body_nav : 0), 
-        nRb.unrotate(dP(n_xi), H ? &D_dP_R : 0),
-        nRb.unrotate(dV(n_xi), H ? &D_dV_R : 0);
+  // Transform correction from navigation frame -> body frame and get Jacobians
+  Vector9 xi;
+  Matrix3 D_dR_R, D_dP_R, D_dV_R, D_body_nav;
```
dellaert: D_body_nav -> Rt, and remark that second derivative is the same for all three.

gtsam/navigation/NavState.cpp
```diff
@@ -444,23 +444,35 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
...
+  dR(xi) = nRb.unrotate(dR(n_xi), H ? &D_dR_R : 0, H ? &D_body_nav : 0);
+  dP(xi) = nRb.unrotate(dP(n_xi), H ? &D_dP_R : 0);
+  dV(xi) = nRb.unrotate(dV(n_xi), H ? &D_dV_R : 0);
 
+  // Assemble Jacobians
```
dellaert, line 475: Should be D_p_v etc..... we use p here not t

gtsam/navigation/NavState.cpp
```diff
@@ -472,9 +484,8 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
     D_v_v(H) << D_body_nav * (-2.0 * dt) * D_cross_state;
     D_v_R(H) << D_dV_R;
     if (secondOrder) {
-      const Matrix3 D_cross2_state = Omega * D_cross_state;
-      D_t_t(H) -= D_body_nav * (0.5 * dt2) * D_cross2_state;
-      D_v_t(H) -= D_body_nav * dt * D_cross2_state;
+      D_t_t(H) -= (0.5 * dt2) * D_body_nav * (H2_wrt_t * R());
```
dellaert: D_p_p(H) -= (0.5 * dt2) * D_c2_np; ???

gtsam/navigation/NavState.cpp
```diff
@@ -472,9 +484,8 @@ Vector9 NavState::coriolis(double dt, const Vector3& omega, bool secondOrder,
...
+      D_v_t(H) -= dt * D_body_nav * (H2_wrt_t * R());
```
dellaert: dt * D_c2_np; ????

I'd add a unit test with non-identity R before - make sure it fails - then apply fix, should succeed.

dellaert:
Would love to merge this!

dellaert (approved):

PR #2288, opened 2025-10-19
6 More Python Examples + Standardized Notebook Format

rohan-bansal:
Hello!

I did a deep dive into the GTSAM codebase and translated some examples from C++ to Python that haven't been done already - I also did my best to explain them in notebook format. 

Additionally, I updated the README to check off files I saw implemented already.

Converted examples:
- `CreateSFMExampleData.cpp` -> `CreateSFMExampleData.ipynb`
- `DiscreteBayesNetExample.cpp` -> `DiscreteBayesNetExample.ipynb`
- `FisheyeExample.cpp` -> `FisheyeExample.ipynb`
- `HMMExample.cpp` -> `HMMExample.ipynb`
- `Pose2SLAMwSPCG.cpp` -> `Pose2SLAMwSPCG.ipynb`
- `Pose2SLAMStressTest.cpp` -> `Pose2SLAMStressTest.ipynb`

One thing I noticed were that ExpressionFactorGraph (and maybe some of the auto differentiation framework) is not exposed yet in Python, so I wasn't able to convert that example. I noted this in the README and would be happy to tackle adding this in the future. Other than `Pose2SLAMExampleExpressions` and `Pose2SLAMExample_lago`, all the Pose2 examples are now translated.

PR #2290, opened 2025-10-20
Added discrete_sreries example tutorial

Robert-Jia00129:
Added quickstart tutorial for Python. Created a quickstart notebook that help new users quickly learn basic concepts and implementation in ~10 minutes. 

Major topics: 
- how gtsam variables are represented
- how to create variables
- how to `print`/display variable
- how to use FactorGraphs
- conditional probabilities
- sampling and assigning variables

Need helps -- TODO: 
- [x] I'm not sure how to add this to the docs pages at `https://borglab.github.io/gtsam/`
- [x] Upload to colab (but probably not under my name)


dellaert:
Nice, Robert!
- I'd not market it as a python tutorial for gtsam, per se, but as an example.
- I'd put this in the python/gtsam/examples folder, along with other examples, and call it DiscreteMotionModel.ipynb or something. It will automatically appear.
- You probably want to refer people to the online book, as that's where it's coming from
- Finally, look at #2288 - it has a handy script to add the "standard" cells at the beginning of an example. That will also install the "open in colab" button

Robert-Jia00129:
- [x] I'd not market it as a python tutorial for gtsam, per se, but as an example.
- [x] I'd put this in the python/gtsam/examples folder, along with other examples, and call it DiscreteMotionModel.ipynb or something. It will automatically appear.
- [x] You probably want to refer people to the online book, as that's where it's coming from

- Yes. I referred to it in the guide in several places. I'll refer to it in the very beginning as well. 
- [x] Finally, look at https://github.com/borglab/gtsam/pull/2288 - it has a handy script to add the "standard" cells at the beginning of an example. That will also install the "open in colab" button
- I saw the Python script. We should probably also bring it into develop branch.

Robert-Jia00129:
@dellaert Just updated the image path. It should be good now. I'll try putting everything in the examples folder so we don't waste the extra resources :)

dellaert:
Oops, CI fails on Mac. Brew issue. Not your fault but if you can fix it great :-)

dellaert:
@Robert-Jia00129 do you have a mac? Is it easy to check a fix?

Robert-Jia00129:
> @Robert-Jia00129 do you have a mac? Is it easy to check a fix?

Still looking into this. Haven't pinpointed what's causing the issue yet.

dellaert:
Any progress?

Robert-Jia00129:
Looking at the CI tests, seems like all the Mac tests have passed?

dellaert:
Ok, could you please merge in develop? I think we may have already removed macos13 from CI there. If not, at least it will run all CI once more and we can merge.

Robert-Jia00129:
Sounds great. 

The current CI without mac13 already passes. Do you want me to rerun the CI with the same files? (It shows on my end a review approval is required before I can merge it, which might be because we skipped CI on Linux and windows that have passed.)

dellaert (approved):

PR #2297, opened 2025-11-04
Added AHRSPose3Factor by extending AHRSFactor to enable integration with Pose3 variables.

junlinp:
I found that the original AHRSFactor does not accept a Pose variable for integrating gyro measurements.
I added AHRSPose3Factor to accept two Pose variables and compute the rotation error using the AHRS measurements for gyro integration.


junlinp:
@github-copilot review

junlinp:
@dellaert PTAL. I can't ask Copilot to review the pull request again.

PR #2301, opened 2025-11-17
[1/N] Split the NoiseModelFactorN to individual header

ProfFan:
Move the large class NoiseModelFactorN to its own header, which will help in future reactors. This necessitated a lot of changes wherever the older header was included.

This PR also includes the other PR #2302 

Adds output argument to NoiseModelFactorN -> NoiseModelFactorT

`NoiseModelFactorN` is now an alias to `NoiseModelFactorT<Vector, ...>`

WIP to enable us to use fixed output dimension with the NoiseModelFactor, which will save extra copy and allow other things.


dellaert:
I like this! I’ll hold off approving until CI passes, including python wrapper, of course.

dellaert:
Please do provide a PR comment even if you think the title is self-explanatory. And maybe provide a rationale even if _what_ is being done is clear.

dellaert (approved):
I'll approve, but please, if you don't outright check, *at least* add @tparam documentation to clarify that only VectorN-like things, possibly (and default) Vector, are allowed.

dellaert:
Nice that this large header is now split up !!

PR #2305, opened 2025-11-19
Fix macOS 26 Xcode compilation

ProfFan:
Fixes two errors on latest macOS Xcode.


dellaert:
So weird, I just added macos-15 with latest xcode to CI in #2304, and all checks passed.

dellaert (approved):

PR #2309, opened 2025-11-22
Fix CMake complaining about Boost

varunagrawal:
Since CMake 3.30.0, FindBoost has been deprecated in favor of `find_package(Boost)`.
This PR updates the CMake and config files so that libraries that link to GTSAM don't complain about this during build.


dellaert:
That’s so strange - I could have sworn I had already fixed this. Is this because you are using 3.30 or later on your machine?

also, do we need to check the version if we check the policy?

varunagrawal:
> That’s so strange - I could have sworn I had already fixed this. Is this because you are using 3.30 or later on your machine?
> 
> also, do we need to check the version if we check the policy?

It probably is complaining because of my machine, but homebrew only allows me to install CMake 4.2.0. :/

dellaert (approved):

Issue #2299, opened 2025-11-10
Significant runtime difference between CERES and GTSAM on same BAL problem


dvorak0:
### Description:
I noticed a large performance discrepancy between the bundle_adjuster binary and the SFMExample_bal example when running on the same BAL dataset.

### Dataset:
problem-16-22106-pre.txt from https://grail.cs.washington.edu/projects/bal/dubrovnik.html
(Problem size: 16 22106 83718)

### Version
96e53555e6ee115f4bff5d471ddef3d3c80ca935

### Build command:

cmake -DCMAKE_BUILD_TYPE=Release ..
make -j

### Timing

**ceres**
```
root@nixos:/tinynav/output/ceres-solver/build# time ./bin/bundle_adjuster --input /tinynav/problem-16-22106-pre.txt -num_threads=1 -linear_solver=dense_schur
iter      cost      cost_change  |gradient|   |step|    tr_ratio  tr_radius  ls_iter  iter_time  total_time
   0  4.185660e+06    0.00e+00    2.16e+07   0.00e+00   0.00e+00  1.00e+04        0    4.42e-02    7.22e-02
   1  1.980525e+05    3.99e+06    5.34e+06   0.00e+00   9.60e-01  3.00e+04        1    8.01e-02    1.52e-01
   2  5.086543e+04    1.47e+05    2.11e+06   1.01e+03   8.22e-01  4.09e+04        1    7.46e-02    2.27e-01
   3  1.859667e+04    3.23e+04    2.87e+05   2.64e+02   9.85e-01  1.23e+05        1    7.48e-02    3.02e-01
   4  1.803857e+04    5.58e+02    2.69e+04   8.66e+01   9.93e-01  3.69e+05        1    7.44e-02    3.76e-01
   5  1.803391e+04    4.66e+00    3.11e+02   1.02e+01   1.00e+00  1.11e+06        1    7.31e-02    4.49e-01

Solver Summary (v 2.2.0-eigen-(3.4.0)-lapack-suitesparse-(5.10.1)-eigensparse-cuda-(12020))

                                     Original                  Reduced
Parameter blocks                        22122                    22122
Parameters                              66462                    66462
Residual blocks                         83718                    83718
Residuals                              167436                   167436

Minimizer                        TRUST_REGION

Dense linear algebra library            EIGEN 
Trust region strategy     LEVENBERG_MARQUARDT
                                        Given                     Used
Linear solver                     DENSE_SCHUR              DENSE_SCHUR
[...]

PR #2313, opened 2025-12-03
add python exports for SmartStereoProjectionFactor

dvorak0:
This patch add python exports for `SmartFactors` in `gtsam_unstable`.

Considering the performance issue mentioned in https://github.com/borglab/gtsam/issues/2299, using smart factors helps to explicitly make use of the special structure of bundle adjust to speed up.


ProfFan (approved):
Thank you!

PR #2315, opened 2025-12-07
Various filter changes

dellaert:
Enhancements to EKF implementations, focusing on clarifying the process noise convention and adding support for higher-order numerical integration in the LieGroupEKF. Also standardizes how continuous-time vs. discrete-time process noise covariances are handled across the filter classes.

**Key Changes:**
- Clarified and documented process noise convention: overloads with `dt` interpret `Q` as continuous-time covariance (scaled internally by `dt`), while overloads without `dt` expect discrete covariance
- Added support for higher-order transition matrix computation via matrix exponential with configurable truncation order `K`
- Implemented Lie-algebra adjoint operations (`adjointMap` and `adjoint`) for Rot2 and Rot3 to support these higher-order methods

### DIFF
--- a/examples/GEKF_Rot3Example.cpp
+++ b/examples/GEKF_Rot3Example.cpp
@@ -57,7 +57,8 @@ int main() {
 
   // Timestep, process noise, measurement noise
   double dt = 0.1;
-  Matrix3 Q = Matrix3::Identity() * 0.01;
+  // Continuous-time process noise (scaled by dt inside predict).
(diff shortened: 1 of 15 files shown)


dellaert:
@ProfFan check this out: you think it might have to be $\sqrt{\delta t}$

dellaert:
> @ProfFan check this out: you think it might have to be δ t

Based on our discussion, multiplying $Q_c$ with $\delta t$ is correct.

scottiyio (commented):
Don't see any problems

PR #2318, opened 2025-12-13
Added BayesTree cache clearing for Marginals

travisdriver:
The cache of the underlying BayesTree of the Marginals class can get very large and lead to memory issues when computing the marginal covariance for a large number of variables in a large graph, so I added a method to allow clearing the cache as a method in the Marginals class.


dellaert (approved):
Handy. Thanks!

PR #2326, opened 2025-12-27
[test] test eigen3 v5.0.1

talregev:
This PR is only for testing.
Showing you the compilation error while compile gtsam with eigen3 version 5.0.1. to fix them in different PR. 

On vcpkg, they do this patch to compile gtsam with eigen3 version 5.0.1:
https://github.com/microsoft/vcpkg/blob/master/ports/gtsam/eigen3-fixes.patch


talregev:
@dellaert @ProfFan 
I have a compilation fix, but one test is start to failing and need to change.
Do you want that I create a PR for that and you take a look?

dellaert:
Yes, please. Happy to take a look!
PS do you see performance benefits with Eigen 5?

talregev:
I created a PR: #2331 
I don't think there is a performance difference in v5.0.1.

dellaert:
Still needed?

talregev:
No, I am close this PR.
Thank you for the review.

PR #2327, opened 2025-12-27
MultifrontalSolver

dellaert:
In order to speed up optimization, I created a fully imperative multifrontal solver that nevertheless is in the style of GTSAM. However, it creates the structure and allocates memory only once, and then can be re-used in subsequent iterations. For a simple chain with 500 nodes, with metis ordering (tree of $\log_2(500)$ deep) it is between 3.5 and 4 times as fast:

```bash
make timeMultifrontalSolver.run

Benchmark (T=500, iterations=1000):
  Standard GTSAM:     2.12248 s
  MultifrontalSolver: 0.548226 s
  Speedup:            3.87155x
```

## Update

After storing pointers into the vectorValues data structure, we are now 5-6 times as fast (on Mac):

```bash
Benchmark (T=500, iterations=1000):
  Standard GTSAM:     1.8607 s
  MultifrontalSolver: 0.314121 s
  Speedup:            5.92352x
```
Also, I realized GTSAM is using TBB, whereas new solver is single-threaded for now.

image

### DIFF
--- a/gtsam/linear/MultifrontalClique.cpp
+++ b/gtsam/linear/MultifrontalClique.cpp
@@ -29,21 +29,19 @@ namespace gtsam {
 namespace {
 
 // Build a stacked separator vector x_sep in the provided scratch buffer.
-Vector& buildSeparatorVector(const KeyVector& separatorKeys,
-                             const std::map<Key, size_t>& dims,
-                             const VectorValues& x, Vector* scratch) {
+Vector& buildSeparatorVector(const std::vector<const Vector*>& separatorPtrs,
+                             Vector* scratch) {
   size_t separatorDim = 0;
-  for (Key key : separatorKeys) {
-    separatorDim += dims.at(key);
+  for (const Vector* values : separatorPtrs) {
+    separatorDim += values->size();
   }
   if (static_cast<size_t>(scratch->size()) != separatorDim) {
     scratch->resize(separatorDim);
   }
   size_t offset = 0;
-  for (Key key : separatorKeys) {
-    const Vector& values = x.at(key);
-    scratch->segment(offset, values.size()) = values;
-    offset += values.size();
+  for (const Vector* values : separatorPtrs) {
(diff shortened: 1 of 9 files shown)


dellaert:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/borglab/gtsam/pull/2327#pullrequestreview-3614471376)

gtsam/linear/MultifrontalClique.cpp
```diff
@@ -0,0 +1,354 @@
...
+  for (Key k : frontals()) blockDims.push_back(dims.at(k));
+  for (Key k : separatorKeys_) blockDims.push_back(dims.at(k));
+  return blockDims;
+}
+
+size_t MultifrontalClique::countRows(const GaussianFactorGraph& graph) const {
+  size_t vbmRows = 0;
+  for (const auto& factor : cluster_->factors) {
+    auto indexed =
+        std::dynamic_pointer_cast<internal::IndexedSymbolicFactor>(factor);
```
ProfFan, line 147: In general we want to avoid dynamic_cast (bad perf under GCC/libstdcxx). but this can be left as a TODO

gtsam/linear/MultifrontalClique.cpp
```diff
@@ -0,0 +1,354 @@
...
+          jacobianFactor->getA(it);
+    }
+    size_t rhsBlockIdx = Ab_.nBlocks() - 1;  // RHS block is appended by VBM.
+    Ab_(rhsBlockIdx).middleRows(rowOffset, jacobianFactor->rows()) =
+        jacobianFactor->getb();
+    rowOffset += jacobianFactor->rows();
+  }
+}
+
+void MultifrontalClique::eliminate() {
```
ProfFan, line 224: `eliminateInPlace`?

ProfFan (approved):

dellaert:
I’ll implement both TODOs in follow-up PR

PR #2329, opened 2025-12-28
JacobianFactor now compute the error more efficiently by first assembling the X vector

ProfFan:
We also now directly assemble the new `GaussianConditional` instead of using the `copy-then-offset` strategy.

During the implementation I also discovered some latent bugs related to the iteration order of the `VectorValues`.

- ~~Replace `FastMap`-based `ConcurrentMap` fallback with `std::unordered_map` when TBB is disabled.~~
- Tighten `BayesTree::equals` logic and related `Values` handling.
- Silence `Point2`/`Point3` constructor unused variable warnings.

### DIFF
--- a/gtsam/base/ConcurrentMap.h
+++ b/gtsam/base/ConcurrentMap.h
@@ -19,6 +19,7 @@
 #pragma once
 
 #include <gtsam/global_includes.h>
+#include "gtsam/base/FastMap.h"
 
 // Change class depending on whether we are using TBB
 #ifdef GTSAM_USE_TBB
@@ -41,8 +42,7 @@ using ConcurrentMapBase = tbb::concurrent_unordered_map<
 
 #else
 
-// If we're not using TBB, use a FastMap for ConcurrentMap
-#include <gtsam/base/FastMap.h>
+// If we're not using TBB, use a std::map
 template <typename KEY, typename VALUE>
 using ConcurrentMapBase = gtsam::FastMap<KEY, VALUE>;
 
--- a/gtsam/base/ConcurrentMap.h
+++ b/gtsam/base/ConcurrentMap.h
@@ -19,7 +19,7 @@
 #pragma once
 
 #include <gtsam/global_includes.h>
-#include "gtsam/base/FastMap.h"
+#include <gtsam/base/FastMap.h>
 
 // Change class depending on whether we are using TBB
 #ifdef GTSAM_USE_TBB
--- a/gtsam/linear/GaussianConditional.cpp
+++ b/gtsam/linear/GaussianConditional.cpp
@@ -305,23 +305,36 @@ namespace gtsam {
     const Vector x =
         frontalValues.vector(KeyVector(beginFrontals(), endFrontals()));
 
-    // Copy the augmented Jacobian matrix:
-    auto newAb = Ab_;
-
-    // Restrict view to parent blocks
-    newAb.firstBlock() += nrFrontals_;
-
-    // Update right-hand-side (last column)
-    auto last = newAb.matrix().cols() - 1;
+    // Compute updated right-hand side: d - R * x
     const auto RR = R().triangularView<Eigen::Upper>();
-    newAb.matrix().col(last) -= RR * x;
+    const Vector rhs = d() - RR * x;
+
+    // Collect parent dimensions
+    FastVector<DenseIndex> parentDims;
+    parentDims.reserve(nrParents());
+    for (auto it = beginParents(); it != endParents(); ++it) {
+      parentDims.push_back(getDim(it));
+    }
+
+    // Build a VerticalBlockMatrix containing only parent blocks and RHS.
+    const DenseIndex m = rows();
+    VerticalBlockMatrix newAb(parentDims, m, true);
+
+    // Copy parent blocks (S matrices).
+    DenseIndex blockIndex = 0;
+    for (auto it = beginParents(); it != endParents(); ++it, ++blockIndex) {
+      newAb(blockIndex) = S(it);
+    }
+
+    // Set the RHS block.
+    const DenseIndex lastBlock = newAb.nBlocks() - 1;
+    newAb(lastBlock).col(0) = rhs;
 
     // The keys now do not include the frontal keys:
     KeyVector newKeys;
     newKeys.reserve(nrParents());
     for (auto&& key : parents()) newKeys.push_back(key);
 
-    // Hopefully second newAb copy below is optimized out...
     return std::make_shared<JacobianFactor>(newKeys, newAb, model_);
   }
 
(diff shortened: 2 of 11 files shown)


gtsam/linear/GaussianConditional.cpp
```diff
@@ -305,23 +305,36 @@ namespace gtsam {
     const Vector x =
         frontalValues.vector(KeyVector(beginFrontals(), endFrontals()));
 
-    // Copy the augmented Jacobian matrix:
-    auto newAb = Ab_;
+    // Compute updated right-hand side: d - R * x
```
dellaert, line 308: What prompted this change? `likelihood` is not used very much so it's not a performance bottleneck I think?

gtsam/base/ConcurrentMap.h
```diff
@@ -41,10 +41,10 @@ using ConcurrentMapBase = tbb::concurrent_unordered_map<
 
 #else
 
-// If we're not using TBB, use a FastMap for ConcurrentMap
-#include <gtsam/base/FastMap.h>
+// If we're not using TBB, use a std::unordered_map
```
dellaert: Seems like a lot of changes in this PR come from this change? Is it very much faster? Can you add performance numbers in the PR comment?

It's weird that that all this sorted business was not needed in TBB path which was *already* unordered...

gtsam/sfm/tests/testShonanAveraging.cpp
```diff
@@ -207,10 +207,10 @@ TEST(ShonanAveraging3, CheckWithEigen) {
 
   Matrix expected(4, 4);
 #if __APPLE__
-  expected << 0.145767, -0.938445, 0.135713, -0.282233,  //
-      0.780348, -0.0104323, 0.266238, 0.565743,          //
-      -0.383624, 0.0434887, 0.917211, 0.0983088,         //
-      -0.471849, -0.342523, -0.263482, 0.768514;
+  expected << 0.291145, -0.695082, 0.148513, -0.640343,  //
```
dellaert: Some mac CI still fails on this (or at least, on a test in testShonanAveraging).

gtsam/linear/JacobianFactor.cpp
```diff
@@ -491,10 +492,17 @@ bool JacobianFactor::equals(const GaussianFactor& f_, double tol) const {
 
 /* ************************************************************************* */
 Vector JacobianFactor::unweighted_error(const VectorValues& c) const {
-  Vector e = -getb();
-  for (size_t pos = 0; pos < size(); ++pos)
-    e += Ab_(pos) * c[keys_[pos]];
-  return e;
+  const Vector v = c.vector(keys_);
```
dellaert: We're still doing two mallocs here. Maybe inline vector(keys_) so you can avoid the v malloc? Maybe too small a win, but with 300k factors that's 300k mallocs saved.

dellaert (commented):
The JacobianFactor unweighted_error change seems like it should get its own small PR - I commented on one possible improvement.

This PR could then just be the change to unordered - which is a bit more controversial.

In both PRs it'd be nice to show the timing improvement in the PR comment, as I did in #2327.

gtsam/linear/GaussianConditional.cpp
ProfFan: I did this to avoid hitting the slow path in `unweighted_error` as this used to be using `firstBlock`

gtsam/base/ConcurrentMap.h
ProfFan: I think TBB might have latent bugs because their "unordered" container is actually not that unordered 🤷

ProfFan:
Result:

Std: standard GTSAM
MFS: new solver

## BAL Benchmarks (Dubrovnik)

_All times in seconds (Multifrontal / Standard)._

### dubrovnik‑16‑22106

| Ordering | Before (MFS) | After (MFS) | Δ MFS      | Before (Std) | After (Std) | Δ Std      |
|----------|--------------|-------------|------------|--------------|-------------|------------|
| Burn     | 0.128529     | 0.126969    | −0.00156   | 0.197182     | 0.204591    | +0.00741   |
| Metis    | 0.122680     | 0.123158    | +0.00048   | 0.193258     | 0.193156    | −0.00010   |
| Schur    | 0.124851     | 0.123400    | −0.00145   | 0.202128     | 0.190944    | −0.01118   |
| Colamd   | 0.130227     | 0.125253    | −0.00497   | 0.211059     | 0.201032    | −0.01003   |

### dubrovnik‑88‑64298

| Ordering | Before (MFS) | After (MFS) | Δ MFS      | Before (Std) | After (Std) | Δ Std      |
|----------|--------------|-------------|------------|--------------|-------------|------------|
| Burn     | 0.932189     | 0.945771    | +0.01358   | 1.01873      | 1.04878     | +0.03005   |
| Metis    | 0.951794     | 0.901886    | −0.04991   | 1.03761      | 0.986437    | −0.05117   |
| Schur    | 0.934710     | 0.926643    | −0.00807   | 0.97489      | 0.990412    | +0.01552   |
| Colamd   | 0.890421     | 0.874290    | −0.01613   | 0.989667     | 0.962819    | −0.02685   |

### dubrovnik‑135‑90642

| Ordering | Before (MFS) | After (MFS) | Δ MFS      | Before (Std) | After (Std) | Δ Std      |
|----------|--------------|-------------|------------|--------------|-------------|------------|
| Burn     | 1.39082      | 1.38696     | −0.00386   | 1.50815      | 1.54920     | +0.04105   |
| Metis    | 1.47609      | 1.47607     | −0.00002   | 1.68671      | 1.54874     | −0.13797   |
| Schur    | 1.47239      | 1.38445     | −0.08794   | 1.72968      | 1.37164     | −0.35804   |
| Colamd   | 1.39377      | 1.38017     | −0.01360   | 1.71858      | 1.51771     | −0.20087   |

---

## Synthetic Chain Benchmarks (timeMultifrontalSolver)

_All times in seconds._

| T    | Before MFS | After MFS | Δ MFS        | Before Std | After Std | Δ Std        |
|------|------------|-----------|--------------|------------|-----------|--------------|
| 10   | 0.002414   | 0.00216875| −0.00024525  | 0.0142875  | 0.0146528 | +0.00036530  |
| 50   | 0.00983542 | 0.00869446| −0.00114096  | 0.061825   | 0.060038  | −0.00178700  |
| 100  | 0.0127348  | 0.0115819 | −0.00115290  | 0.111825   | 0.112742  | +0.00091700  |
| 500  | 0.0292872  | 0.0312743 | +0.00198710  | 0.470536   | 0.46525   | −0.00528600  |
| 1000 | 0.0429256  | 0.0415565 | −0.00136910  | 0.929741   | 0.909625  | −0.02011600  |
| 5000 | 0.138232   | 0.122859  | −0.01537300  | 4.1799     | 4.17908   | −0.00082000  |

dellaert (approved):
Thanks ! PS, timing of MFS should not affected by error. Only shows up in nonlinear optimizer. There, BTW, we call it *twice*, one for delta and one for zero. I think that may be replaced with one new method errorDelta(delta) which omits the b add.

PR #2333, opened 2025-12-31
Allow for HessianFactors in MultifrontalSolver

dellaert:
- added a new method in SymmetricBlockMatrix
- used it to simplify clique and hessianFactor
- used it to allow for adding HessianFactors

### DIFF
--- a/gtsam/linear/MultifrontalClique.cpp
+++ b/gtsam/linear/MultifrontalClique.cpp
@@ -16,6 +16,7 @@
  * @date   December 2025
  */
 
+#include <gtsam/linear/HessianFactor.h>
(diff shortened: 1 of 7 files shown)


gtsam/linear/MultifrontalClique.cpp
```diff
@@ -201,6 +202,7 @@ size_t MultifrontalClique::countRows(const GaussianFactorGraph& graph) const {
 std::vector<size_t> MultifrontalClique::parentIndicesFor(
     const MultifrontalClique& parent) const {
   std::vector<size_t> indices;
```
ProfFan, line 204: comment on what `indices` are

ProfFan (approved):

PR #2340, opened 2026-01-06
Optimize Levenberg Marquardt elimination by caching JunctionTree.

tzvist:
Add JunctionIndexEliminationTree class that stores cluster roots using factor indices instead of full factor objects, enabling efficient reuse of elimination tree structure across multiple LevenbergMarquardt iterations.

Key changes:
- Add JunctionIndexEliminationTree to cache cluster roots structure
- Modify GaussianFactorGraph::optimize() to use cached elimination tree
- Cache damped system elimination tree in LevenbergMarquardtOptimizer
- Add comprehensive unit tests for JunctionIndexEliminationTree

Performance improvement: reduces runtime from ~120s to ~90s (~25% faster) in a specific test cases by avoiding repeated elimination tree construction.

### DIFF
--- a/gtsam/inference/EliminateableFactorGraph-inst.h
+++ b/gtsam/inference/EliminateableFactorGraph-inst.h
@@ -20,6 +20,10 @@
 
 #include <gtsam/inference/EliminateableFactorGraph.h>
 #include <gtsam/inference/inferenceExceptions.h>
+#include <gtsam/base/treeTraversal-inst.h>
+#include <gtsam/symbolic/IndexedJunctionTree.h>
+
+#include <unordered_set>
 
 namespace gtsam {
 
@@ -143,6 +147,112 @@ namespace gtsam {
       // Return the Bayes tree
       return bayesTree;
     }
+  }
+
+  /* ************************************************************************* */
+  template <class FACTORGRAPH>
+  IndexedJunctionTree
+  EliminateableFactorGraph<FACTORGRAPH>::buildIndexedJunctionTree(
+      const Ordering& ordering,
+      const std::unordered_set<Key>& fixedKeys) const {
+    return IndexedJunctionTree(asDerived(), ordering, fixedKeys);
+  }
+
+  /* ************************************************************************* */
+  template <class FACTORGRAPH>
+  std::shared_ptr<typename EliminateableFactorGraph<FACTORGRAPH>::BayesTreeType>
+  EliminateableFactorGraph<FACTORGRAPH>::eliminateMultifrontal(
+      const IndexedJunctionTree& indexedJunctionTree,
+      const Eliminate& function) const {
+    gttic(eliminateMultifrontal);
+
+    using BayesTreeNode = typename BayesTreeType::Node;
+    using SharedFactor = typename FactorGraphType::sharedFactor;
+
+    // Helper struct for cluster elimination traversal.
+    struct ClusterEliminationData {
+      ClusterEliminationData* const parentData;
+      size_t myIndexInParent;
+      FastVector<SharedFactor> childFactors;
+      std::shared_ptr<BayesTreeNode> bayesTreeNode;
+
+      explicit ClusterEliminationData(ClusterEliminationData* p)
+          : parentData(p), bayesTreeNode(std::make_shared<BayesTreeNode>()) {
+        if (parentData) {
+          myIndexInParent = parentData->childFactors.size();
+          parentData->childFactors.push_back(SharedFactor());
+          if (parentData->parentData)
+            bayesTreeNode->parent_ = parentData->bayesTreeNode;
+          parentData->bayesTreeNode->children.push_back(bayesTreeNode);
+        } else {
+          myIndexInParent = 0;
+        }
+      }
+    };
+
+    // Post-order visitor functor for elimination.
+    class EliminationPostOrderVisitor {
+      const FactorGraphType& graph_;
+      const Eliminate& function_;
+
+     public:
+      EliminationPostOrderVisitor(const FactorGraphType& graph,
+                                  const Eliminate& function)
+          : graph_(graph), function_(function) {}
+
+      void operator()(const SymbolicJunctionTree::sharedNode& cluster,
+                      ClusterEliminationData& myData) {
+        FactorGraphType gatheredFactors;
+        gatheredFactors.reserve(cluster->factors.size() + myData.childFactors.size());
+
+        // Gather factors from IndexedSymbolicFactor.
+        for (const auto& factor : cluster->factors) {
+          auto indexed =
+              std::static_pointer_cast<internal::IndexedSymbolicFactor>(factor);
+          gatheredFactors.push_back(graph_.at(indexed->index_));
+        }
+
+        for (const auto& childFactor : myData.childFactors)
+          if (childFactor) gatheredFactors.push_back(childFactor);
+
+        auto result = function_(gatheredFactors, cluster->orderedFrontalKeys);
+        myData.bayesTreeNode->setEliminationResult(result);
+
+        if (myData.parentData && !result.second->empty())
+          myData.parentData->childFactors[myData.myIndexInParent] = result.second;
+      }
+    };
+
+    // Run depth-first traversal.
+    ClusterEliminationData rootsContainer(nullptr);
+    EliminationPostOrderVisitor visitorPost(asDerived(), function);
+
+    auto visitorPre = [](const SymbolicJunctionTree::sharedNode& node,
+                         ClusterEliminationData& parentData) {
+      ClusterEliminationData myData(&parentData);
+      myData.bayesTreeNode->problemSize_ = node->problemSize_;
+      return myData;
+    };
+
+    treeTraversal::DepthFirstForest(indexedJunctionTree, rootsContainer, visitorPre,
+                                    visitorPost);
+
+    // If any factors are remaining, the ordering was incomplete.
+    KeySet remainingKeys;
+    for (const auto& factor : rootsContainer.childFactors) {
+      if (!factor || factor->empty()) continue;
+      remainingKeys.insert(factor->begin(), factor->end());
+    }
+    if (!remainingKeys.empty()) {
+      throw InconsistentEliminationRequested(remainingKeys);
+    }
+
+    // Create BayesTree from roots stored in the dummy BayesTree node.
+    auto bayesTree = std::make_shared<BayesTreeType>();
+    for (const auto& rootClique : rootsContainer.bayesTreeNode->children)
+      bayesTree->insertRoot(rootClique);
+
+    return bayesTree;
   }
 
   /* ************************************************************************* */
--- a/gtsam/inference/EliminateableFactorGraph-inst.h
+++ b/gtsam/inference/EliminateableFactorGraph-inst.h
@@ -23,6 +23,9 @@
 #include <gtsam/base/treeTraversal-inst.h>
 #include <gtsam/symbolic/IndexedJunctionTree.h>
 
+#ifdef GTSAM_USE_TBB
+#include <mutex>
+#endif
 #include <unordered_set>
 
 namespace gtsam {
@@ -169,73 +172,114 @@ namespace gtsam {
     using BayesTreeNode = typename BayesTreeType::Node;
     using SharedFactor = typename FactorGraphType::sharedFactor;
 
-    // Helper struct for cluster elimination traversal.
+    // Elimination traversal data - stores a pointer to the parent data and collects
+    // the factors resulting from elimination of the children.  Also sets up BayesTree
+    // cliques with parent and child pointers.
     struct ClusterEliminationData {
       ClusterEliminationData* const parentData;
       size_t myIndexInParent;
       FastVector<SharedFactor> childFactors;
       std::shared_ptr<BayesTreeNode> bayesTreeNode;
+#ifdef GTSAM_USE_TBB
+      std::shared_ptr<std::mutex> writeLock;
+#endif
 
-      explicit ClusterEliminationData(ClusterEliminationData* p)
-          : parentData(p), bayesTreeNode(std::make_shared<BayesTreeNode>()) {
+      ClusterEliminationData(ClusterEliminationData* _parentData, size_t nChildren)
+          : parentData(_parentData), bayesTreeNode(std::make_shared<BayesTreeNode>())
+#ifdef GTSAM_USE_TBB
+            , writeLock(std::make_shared<std::mutex>())
+#endif
+      {
         if (parentData) {
+#ifdef GTSAM_USE_TBB
+          parentData->writeLock->lock();
+#endif
           myIndexInParent = parentData->childFactors.size();
           parentData->childFactors.push_back(SharedFactor());
-          if (parentData->parentData)
-            bayesTreeNode->parent_ = parentData->bayesTreeNode;
-          parentData->bayesTreeNode->children.push_back(bayesTreeNode);
+#ifdef GTSAM_USE_TBB
+          parentData->writeLock->unlock();
+#endif
         } else {
           myIndexInParent = 0;
         }
+        if (parentData) {
+          if (parentData->parentData)
+            bayesTreeNode->parent_ = parentData->bayesTreeNode;
+          parentData->bayesTreeNode->children.push_back(bayesTreeNode);
+        }
+      }
+
+      static ClusterEliminationData EliminationPreOrderVisitor(
+          const SymbolicJunctionTree::sharedNode& node,
+          ClusterEliminationData& parentData) {
+        assert(node);
+        ClusterEliminationData myData(&parentData, node->nrChildren());
+        myData.bayesTreeNode->problemSize_ = node->problemSize();
+        return myData;
       }
     };
 
-    // Post-order visitor functor for elimination.
+    // Elimination post-order visitor - gather factors, eliminate, store results.
     class EliminationPostOrderVisitor {
       const FactorGraphType& graph_;
-      const Eliminate& function_;
+      const Eliminate& eliminationFunction_;
 
      public:
-      EliminationPostOrderVisitor(const FactorGraphType& graph,
-                                  const Eliminate& function)
-          : graph_(graph), function_(function) {}
+      EliminationPostOrderVisitor(
+          const FactorGraphType& graph,
+          const Eliminate& eliminationFunction)
+          : graph_(graph), eliminationFunction_(eliminationFunction) {}
 
-      void operator()(const SymbolicJunctionTree::sharedNode& cluster,
+      void operator()(const SymbolicJunctionTree::sharedNode& node,
                       ClusterEliminationData& myData) {
-        FactorGraphType gatheredFactors;
-        gatheredFactors.reserve(cluster->factors.size() + myData.childFactors.size());
+        assert(node);
 
-        // Gather factors from IndexedSymbolicFactor.
-        for (const auto& factor : cluster->factors) {
+        FactorGraphType gatheredFactors;
+        gatheredFactors.reserve(node->factors.size() + node->nrChildren());
+
+        for (const auto& factor : node->factors) {
           auto indexed =
               std::static_pointer_cast<internal::IndexedSymbolicFactor>(factor);
           gatheredFactors.push_back(graph_.at(indexed->index_));
         }
+        gatheredFactors.push_back(myData.childFactors);
 
-        for (const auto& childFactor : myData.childFactors)
-          if (childFactor) gatheredFactors.push_back(childFactor);
+        auto eliminationResult =
+            eliminationFunction_(gatheredFactors, node->orderedFrontalKeys);
 
-        auto result = function_(gatheredFactors, cluster->orderedFrontalKeys);
-        myData.bayesTreeNode->setEliminationResult(result);
+        myData.bayesTreeNode->setEliminationResult(eliminationResult);
 
-        if (myData.parentData && !result.second->empty())
-          myData.parentData->childFactors[myData.myIndexInParent] = result.second;
+        if (!eliminationResult.second->empty()) {
+#ifdef GTSAM_USE_TBB
+          myData.parentData->writeLock->lock();
+#endif
+          myData.parentData->childFactors[myData.myIndexInParent] =
+              eliminationResult.second;
+#ifdef GTSAM_USE_TBB
+          myData.parentData->writeLock->unlock();
+#endif
+        }
       }
     };
 
-    // Run depth-first traversal.
-    ClusterEliminationData rootsContainer(nullptr);
+    // Do elimination (depth-first traversal).  The rootsContainer stores a 'dummy'
+    // BayesTree node that contains all of the roots as its children.  rootsContainer
+    // also stores the remaining un-eliminated factors passed up from the roots.
+    std::shared_ptr<BayesTreeType> result = std::make_shared<BayesTreeType>();
+
+    ClusterEliminationData rootsContainer(0, indexedJunctionTree.nrRoots());
+
     EliminationPostOrderVisitor visitorPost(asDerived(), function);
+    {
+      TbbOpenMPMixedScope threadLimiter;
+      treeTraversal::DepthFirstForestParallel(
+          indexedJunctionTree, rootsContainer,
+          ClusterEliminationData::EliminationPreOrderVisitor, visitorPost, 10);
+    }
 
-    auto visitorPre = [](const SymbolicJunctionTree::sharedNode& node,
-                         ClusterEliminationData& parentData) {
-      ClusterEliminationData myData(&parentData);
-      myData.bayesTreeNode->problemSize_ = node->problemSize_;
-      return myData;
-    };
-
-    treeTraversal::DepthFirstForest(indexedJunctionTree, rootsContainer, visitorPre,
-                                    visitorPost);
+    // Create BayesTree from roots stored in the dummy BayesTree node.
+    for (const auto& rootClique : rootsContainer.bayesTreeNode->children)
+      result->insertRoot(rootClique);
 
     // If any factors are remaining, the ordering was incomplete.
     KeySet remainingKeys;
@@ -247,12 +291,7 @@ namespace gtsam {
       throw InconsistentEliminationRequested(remainingKeys);
     }
 
-    // Create BayesTree from roots stored in the dummy BayesTree node.
-    auto bayesTree = std::make_shared<BayesTreeType>();
-    for (const auto& rootClique : rootsContainer.bayesTreeNode->children)
-      bayesTree->insertRoot(rootClique);
-
-    return bayesTree;
+    return result;
   }
 
   /* ************************************************************************* */
--- a/gtsam/symbolic/IndexedJunctionTree.h
+++ b/gtsam/symbolic/IndexedJunctionTree.h
@@ -1,0 +1,106 @@
+/* ----------------------------------------------------------------------------
+ *
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+ *
+ * See LICENSE for the license information
+ *
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file    IndexedJunctionTree.h
+ * @brief   Build a symbolic junction tree that stores original factor indices.
+ * @author  Tzvi Strauss
+ */
+
+#pragma once
+
+#include <gtsam/base/types.h>
+#include <gtsam/inference/Key.h>
+#include <gtsam/inference/Ordering.h>
+#include <gtsam/symbolic/SymbolicEliminationTree.h>
+#include <gtsam/symbolic/SymbolicFactor.h>
+#include <gtsam/symbolic/SymbolicFactorGraph.h>
+#include <gtsam/symbolic/SymbolicJunctionTree.h>
+
+#include <cstddef>
+#include <unordered_set>
+
+namespace gtsam {
+
+namespace internal {
+/// A SymbolicFactor that stores the index of the originating factor.
+class IndexedSymbolicFactor : public SymbolicFactor {
+ public:
+  size_t index_;
+  IndexedSymbolicFactor(const KeyVector& keys, size_t index)
+      : SymbolicFactor(), index_(index) {
+    keys_ = keys;
+  }
+};
+}  // namespace internal
+
+/**
+ * A symbolic junction tree whose factors record the original factor indices from
+ * a corresponding (non-symbolic) factor graph. This allows the junction tree
+ * structure to be cached and reused for repeated eliminations when the factor
+ * graph structure and ordering remain unchanged, avoiding the cost of rebuilding
+ * the symbolic structure.
+ *
+ * The underlying implementation uses a `SymbolicJunctionTree` where each factor
+ * is an `IndexedSymbolicFactor` that stores the original factor index.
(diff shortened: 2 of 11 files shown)


gtsam/linear/JunctionIndexEliminationTree-inl.h
```diff
@@ -0,0 +1,68 @@
...
+#include <gtsam/inference/Ordering.h>
+#include <gtsam/linear/GaussianFactorGraph.h>
+#include <memory>
+#include <vector>
+
+namespace gtsam {
+namespace internal {
+
+/// IndexedCluster: mirrors ClusterTree::Cluster but stores factor indices
+struct IndexedCluster {
```
tzvist: In this file I I tried keeping code identical (as much as possible) to the new code and i just changed it to use indices instead of pointers to factors (this way it can be reused if the factors do not change and only values and linearization change)

dellaert:
woohoo, that's a big win! Will review Thursday.

dellaert:
@tzvist I have not read the code in detail yet; it would help me to know the philosophy a little bit better. How do the new values of the damped system enter into the equation? Am I right to assume the system is still built as usual, but because you’re using indices, the new factors are accessed?

tzvist:
> @tzvist I have not read the code in detail yet; it would help me to know the philosophy a little bit better. How do the new values of the damped system enter into the equation? Am I right to assume the system is still built as usual, but because you’re using indices, the new factors are accessed?

Yes, the damped system is built exactly the same as before (the buildDampedSystem call at line 143) but instead of recomputing the elimination structure every time, we now cache a JunctionIndexEliminationTree that only stores the indices into the factor graph.
The key insight is that across different lambda iterations, the damped system's structure doesn't change - only the damping values change. So we compute the JunctionIndexEliminationTree once (lines 147-149) from the first damped system structure, and it stores the precalculated indices.
Then in each tryLambda call, we build the new damped system with the updated lambda value, pass it to the cached elimination tree via setCachedJunctionElimination (line 151), and the elimination process accesses the new factors according to the precalculated indices. This way we can reuse the expensive elimination structure computation across all lambda search iterations.

dellaert:
I think you have the right idea here. I’ve actually used this idea as well recently in the new multifrontal solver. By the way, if you’re not aware of it, here is the link to the last PR that I just submitted: https://github.com/borglab/gtsam/pull/2343

That new solver is on track to be integrated with the nonlinear optimizers, to get hopefully much more speed-up than 75% even. But I think that your idea can be used to speed up any eliminatable factor graph, including hybrid, so I’m interested in exploring this regardless. Also only linear systems without constraints can be sped up by the new multifrontal solver.

That being said, I prefer if we could discuss the API and make it a little bit more functional. Would you be available to meet some time next week?

tzvist:
> I think you have the right idea here. I’ve actually used this idea as well recently in the new multifrontal solver. By the way, if you’re not aware of it, here is the link to the last PR that I just submitted: #2343
> 
> That new solver is on track to be integrated with the nonlinear optimizers, to get hopefully much more speed-up than 75% even. But I think that your idea can be used to speed up any eliminatable factor graph, including hybrid, so I’m interested in exploring this regardless. Also only linear systems without constraints can be sped up by the new multifrontal solver.
> 
> That being said, I prefer if we could discuss the API and make it a little bit more functional. Would you be available to meet some time next week?

Sure, I’d be happy to discuss the API.  
I’m in a different time zone, but I’m available this week (Monday–Thursday) between 9:00 and 11:00 Atlanta time.  
Let me know if any of those times work for you.

dellaert:
Awesome. Here is a link that knows about time zones and my calendar :-) https://calendly.com/dellaert/teams-meeting

tzvist:
> Awesome. Here is a link that knows about time zones and my calendar :-) https://calendly.com/dellaert/teams-meeting

I have two more perforce oriented MR:
https://github.com/borglab/gtsam/pull/2355
https://github.com/borglab/gtsam/pull/2356

Maybe it will be best if you can try taking a look at them also and then we can meet next week and discuss them all.

dellaert:
Please take a look at the junction tree I built in the multi-frontal solver. I use a symbolic factor with some extra payload. I think it might be a good strategy to template the symbolic factor with an optional payload, so we can just use the strategy in both places. 

In terms of using this junction tree (or elimination tree), which will now store the extra payload in its clique factors, I'd rather see that used in a functional interface rather than storing something in a mutable field in the Gaussian FG

tzvist:
> Please take a look at the junction tree I built in the multi-frontal solver. I use a symbolic factor with some extra payload. I think it might be a good strategy to template the symbolic factor with an optional payload, so we can just use the strategy in both places.
> 
> In terms of using this junction tree (or elimination tree), which will now store the extra payload in its clique factors, I'd rather see that used in a functional interface rather than storing something in a mutable field in the Gaussian FG

I have added [Refactor MultifrontalSolver to use JunctionIndexEliminationTree](https://github.com/borglab/gtsam/pull/2340/commits/683e2dcbeca0d481dd752332b84fa88698b543b3) i started doing the templating but this was an simpler approach, is this good in you opinion? 
i still need to do the functional interface.

also do you think we should try to add the mergeSmallClusters also to existing LM?

dellaert:
I think we're a little tripping over ourselves here :-) I would prefer to hold up on this one. I'm hesitant about adding entire junction trees and cluster trees. Basically, by only adapting the symbolic factor a tiny little bit we get all the machinery without touching those larger classes. I have another refactor in the works that splits out the SymbolicAnalysis for the new solvers. For that, I might try my hand at the templated approach first - potentially accommodating your case as well. When done, you can check that out and see whether we can take your idea  and make that work in a functional API.

tzvist:
> I think we're a little tripping over ourselves here :-) I would prefer to hold up on this one. I'm hesitant about adding entire junction trees and cluster trees. Basically, by only adapting the symbolic factor a tiny little bit we get all the machinery without touching those larger classes. I have another refactor in the works that splits out the SymbolicAnalysis for the new solvers. For that, I might try my hand at the templated approach first - potentially accommodating your case as well. When done, you can check that out and see whether we can take your idea and make that work in a functional API.

Could you review the last two commits? I think they’re close to what you had in mind.

gtsam/linear/GaussianFactorGraph.h
```diff
@@ -406,7 +410,24 @@ namespace gtsam {
...
 
+    /**
+     * Set a cached junction elimination tree to reuse for optimization.
+     * This allows avoiding recomputation of the elimination structure when
+     * optimizing the factor graph multiple times with the same structure.
+     * The cached tree is used by optimize() methods if available.
+     * @param cachedJunctionElimination Shared pointer to the pre-computed junction elimination tree
+     *                                  to cache. Can be nullptr to clear the cache.
+     */
+    void setCachedJunctionElimination(
```
dellaert: I fundamentally disagree with this :-) I have been trying to remain as functional as possible in GTSAM, avoiding state where possible. This has not always succeeded (both iSAM and the new MultifrontalSolver are egregious violations). They'd like to keep it out of `GaussianFactorGraph`, if possible. 

I think it is not needed, actually: the NonlinearOptimizer (not LM, BTW, solving the linear system is the responsibility of the base class) can just call
```
      cachedJunctionElimination_->eliminate(linearizedGraph, function)->optimize();
```

gtsam/nonlinear/LevenbergMarquardtOptimizer.h
```diff
@@ -127,6 +131,9 @@ class GTSAM_EXPORT LevenbergMarquardtOptimizer: public NonlinearOptimizer {
   const NonlinearOptimizerParams& _params() const override {
     return params_;
   }
+
+  /// Cached junction index elimination tree used to build or reuse the damped linear system.
+  std::shared_ptr<const internal::JunctionIndexEliminationTree> cachedDampedJunctionElimination_;
```
dellaert: Can we move this to the nonlinear optimizer rather than in LM? Then the two other solvers can also make use of this caching, right?

gtsam/linear/JunctionIndexEliminationTree.h
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * Builds a SymbolicJunctionTree at construction that stores original factor indices.
+ * This allows reusing the symbolic structure when factor values change but the graph
+ * structure (which variables each factor involves) remains constant.
+ *
+ * @note Internal class, used as elimination cache in GaussianFactorGraph and
+ * LevenbergMarquardtOptimizer.
+ * @see GaussianBayesTree, SymbolicJunctionTree
+ */
+class GTSAM_EXPORT JunctionIndexEliminationTree {
```
dellaert: Do we still need this class at the moment? I think the three new files in this directory could just be replaced with a single file that builds the symbolic junction tree with the new symbolic factors and also optimizes. Both of them can be functions rather than a class. 

I'm thinking that the current structure is because of the history of your PR and we can simplify it now that the building is done using these new symbolic factors ?

dellaert (commented):
Alright, I think we are starting to converge :-)

I still have structural comments and I'm hoping that you'll bear with me while we collaborate on making this a really cool addition to GTSAM :-)

gtsam/nonlinear/LevenbergMarquardtOptimizer.h
tzvist: Would you be ok that we will create the junctionTree in  NonlinearOptimizer::solve? Can we count on it to be a new object each time we have a new ordering or should we destroy it in NonlinearOptimizer::defaultOptimize, alternatively store the pointer to ordering  and destroy if ordering pointer changes? any other idea how to invalidate cache? also is it correct that for all subclass of NonlinearOptimizer we can cache the junctionTree ? (I only reviewed the logic of LM to be sure that it is ok).

gtsam/linear/JunctionIndexEliminationTree.h
tzvist: should we move this code https://github.com/tzvist/gtsam/blob/8650ec2cf8b2f415b1c749f8fc63b2676225bb5d/gtsam/linear/JunctionIndexEliminationTree.cpp#L78 into  GaussianFactorGraph.cpp, or would you prefer a separate file?

dellaert: I was thinking the three files could be combined in one .h/.cpp, something like indexedElimination.*. Because this is now starting to co-mingle with MultiFrontalSolver, maybe the new functions can even go in there.

gtsam/nonlinear/LevenbergMarquardtOptimizer.h
dellaert: Yes, this is what I do for the Multifrontal solver already - please take a look. There is  no need for cache invalidation, as the solvers are created functionally: you cannot change the graph after construction.

gtsam/linear/GaussianFactorGraph.h
tzvist: I added an additional overload of the GaussianFactorGraph::optimize method that accepts the indexed symbolic tree. This seemed like the cleanest approach.

dellaert: So, I disagree, because it breaks encasulation. A Gaussian factor graph is at a lower abstraction level than a junction tree. A junction tree organizes the factors in a factor graph. So, I don't want to mix these different abstraction levels, if possible. A junction tree solver can know about factor graphs, but not the other way around. 

Take a look at EliminateableFactorGraph, this has:
```cpp
  template<class FACTORGRAPH>
  std::shared_ptr<typename EliminateableFactorGraph<FACTORGRAPH>::BayesTreeType>
    EliminateableFactorGraph<FACTORGRAPH>::eliminateMultifrontal(
    const Ordering& ordering, const Eliminate& function,
    OptionalVariableIndex variableIndex) const
  {
    if(!variableIndex) {
      // If no VariableIndex provided, compute one and call this function again
      VariableIndex computedVariableIndex(asDerived());
      return eliminateMultifrontal(ordering, function, std::cref(computedVariableIndex));
    } else {
      gttic(eliminateMultifrontal);
      // Do elimination with given ordering
      EliminationTreeType etree(asDerived(), (*variableIndex).get(), ordering);
      JunctionTreeType junctionTree(etree);
      const auto [bayesTree, factorGraph] = junctionTree.eliminate(function);
      // If any factors are remaining, the ordering was incomplete
      if(!factorGraph->empty()) {
        throw InconsistentEliminationRequested(factorGraph->keys());
      }
      // Return the Bayes tree
      return bayesTree;
    }
  }
```
This is called from
```
  VectorValues GaussianFactorGraph::optimize(const Ordering& ordering, const Eliminate& function) const {
    gttic(GaussianFactorGraph_optimize);
    return BaseEliminateable::eliminateMultifrontal(ordering, function)->optimize();
  }
```
but the implementation, crucially, lives in a junctionTree class. That's appropriate because the FG doesn't need to know about the internals of junction trees. The FG doesn't even know that a tree is being constructed. It only provides the type to the templated version. 

But in formulating this answer, it made me think that basically _any_ factor graph can benefit from not redoing the junction tree every time. In EliminateableFactorGraph, we could have (something like):
```cpp
  template<class FACTORGRAPH>
  std::shared_ptr<IndexedJunctionTree> EliminateableFactorGraph<FACTORGRAPH>::buildIndexedJunctionTree(
    const Ordering& ordering) const {
  // build and return IndexedJunctionTree
  }

  template<class FACTORGRAPH>
  std::shared_ptr<typename EliminateableFactorGraph<FACTORGRAPH>::BayesTreeType>
    EliminateableFactorGraph<FACTORGRAPH>::eliminateMultifrontal(
    const IndexedJunctionTree& indexedJunctionTree, const Eliminate& function) const
  {
  // eliminate using IndexedJunctionTree
  }
```
This is basically what you have already, but (a) it would be in the right place, (b) it would be completely general.

There will be zero change to Gaussian factor graph in that case because you can simply do this in the optimizer:
```cpp
if (!indexedJunctionTree_) {
  indexedJunctionTree_ = linearized.buildIndexedJunctionTree(ordering);
}
auto bt = linearized.eliminateMultifrontal(*indexedJunctionTree_, eliminateFunction);
return bt->optimize();
```

tzvist: should the following code be Inside `NonlinearOptimizer.cpp` or would you prefer it being in `IndexedJunctionTreeBuilder.h`?

```

namespace {
// Helper struct for cluster elimination traversal
struct ClusterEliminationData {
  ClusterEliminationData* const parentData;
  size_t myIndexInParent;
  std::vector<GaussianFactor::shared_ptr> childFactors;
  GaussianBayesTreeClique::shared_ptr bayesTreeNode;

  ClusterEliminationData(ClusterEliminationData* p, size_t)
      : parentData(p),
        bayesTreeNode(std::make_shared<GaussianBayesTreeClique>()) {
    if (parentData) {
      myIndexInParent = parentData->childFactors.size();
      parentData->childFactors.push_back(nullptr);
      if (parentData->parentData)
        bayesTreeNode->parent_ = parentData->bayesTreeNode;
      parentData->bayesTreeNode->children.push_back(bayesTreeNode);
    } else {
      myIndexInParent = 0;
    }
  }
};

// Post-order visitor functor for elimination
class EliminationPostOrderVisitor {
  const GaussianFactorGraph& graph_;
  const GaussianFactorGraph::Eliminate& function_;

 public:
  EliminationPostOrderVisitor(const GaussianFactorGraph& graph,
                              const GaussianFactorGraph::Eliminate& function)
      : graph_(graph), function_(function) {}

  void operator()(const SymbolicJunctionTree::sharedNode& cluster,
                  ClusterEliminationData& myData) {
    GaussianFactorGraph gatheredFactors;
    // Gather factors from IndexedSymbolicFactor
    for (const auto& factor : cluster->factors) {
      auto indexed = std::static_pointer_cast<internal::IndexedSymbolicFactor>(factor);
      gatheredFactors.push_back(graph_[indexed->index_]);
    }
    for (const auto& childFactor : myData.childFactors)
      if (childFactor) gatheredFactors.push_back(childFactor);

    auto result = function_(gatheredFactors, cluster->orderedFrontalKeys);
    myData.bayesTreeNode->setEliminationResult(result);

    if (myData.parentData && !result.second->empty())
      myData.parentData->childFactors[myData.myIndexInParent] = result.second;
  }
};

std::shared_ptr<GaussianBayesTree> eliminateGaussianGraph(
    const SymbolicJunctionTree& indexedJunctionTree,
    const GaussianFactorGraph& graph,
    const GaussianFactorGraph::Eliminate& function) {
  auto bayesTree = std::make_shared<GaussianBayesTree>();
  ClusterEliminationData rootsContainer(nullptr, indexedJunctionTree.roots().size());

  // Pre-order visitor - creates the ClusterEliminationData structure
  auto visitorPre = [](const SymbolicJunctionTree::sharedNode& node,
                       ClusterEliminationData& parentData) {
    ClusterEliminationData myData(&parentData, node->nrChildren());
    myData.bayesTreeNode->problemSize_ = node->problemSize_;
    return myData;
  };

  // Use depth-first traversal with pre-order and post-order visitors
  EliminationPostOrderVisitor visitorPost(graph, function);
  treeTraversal::DepthFirstForest(indexedJunctionTree, rootsContainer, visitorPre,
                                  visitorPost);

  for (const auto& rootClique : rootsContainer.bayesTreeNode->children)
    bayesTree->insertRoot(rootClique);

  return bayesTree;
}
}  // anonymous namespace
```

dellaert: Maybe in IndexedJunctionTree.h, along with building to support `EliminateableFactorGraph<FACTORGRAPH>::buildIndexedJunctionTree`

tzvist: I think the key detail I’m missing in your explanation is that these methods also require the GaussianFactorGraph instance to be passed in.

```
  template<class FACTORGRAPH>
  std::shared_ptr<IndexedJunctionTree> EliminateableFactorGraph<FACTORGRAPH>::buildIndexedJunctionTree(
    const Ordering& ordering) const {
  // build and return IndexedJunctionTree
  }

  template<class FACTORGRAPH>
  std::shared_ptr<typename EliminateableFactorGraph<FACTORGRAPH>::BayesTreeType>
    EliminateableFactorGraph<FACTORGRAPH>::eliminateMultifrontal(
    const IndexedJunctionTree& indexedJunctionTree, const Eliminate& function) const
  {
  // eliminate using IndexedJunctionTree
  }
  ```

Comment (dellaert) [reply gtsam/linear/GaussianFactorGraph.h]: Even more, they are methods :-)
And, defined for *any* eliminateable factor graph. So, the code fragment I gave above should run in testSymbolicBayesTree.cpp and give the same BT, multiple times, e.g. in `TEST(SymbolicBayesTree, clique_structure)`

File: gtsam/symbolic/tests/testSymbolicBayesTree.cpp [hunk 5]
```diff
@@ -96,6 +96,17 @@ TEST(SymbolicBayesTree, clique_structure) {
   SymbolicBayesTree actual = *graph.eliminateMultifrontal(order);
 
   EXPECT(assert_equal(expected, actual));
+
+  // Reuse an indexed junction tree built from the same graph and ordering.
+  IndexedJunctionTree indexedJunctionTree = graph.buildIndexedJunctionTree(order);
+
+  SymbolicBayesTree actualReuse1 =
+      *graph.eliminateMultifrontal(indexedJunctionTree);
```
dellaert, line 104: Nice!!!!

gtsam/symbolic/IndexedJunctionTree.h
```diff
@@ -0,0 +1,106 @@
...
+      }
+      // Skip factors that are fully constrained away.
+      if (keys.empty()) continue;
+      symbolicGraph.emplace_shared<internal::IndexedSymbolicFactor>(keys, i);
+    }
+    return symbolicGraph;
+  }
+
+  template <typename GRAPH>
+  static SymbolicEliminationTree makeEliminationTree(
```
dellaert: Maybe `makeIndexedEliminationTree`

gtsam/inference/EliminateableFactorGraph-inst.h
```diff
@@ -145,6 +149,112 @@ namespace gtsam {
     }
   }
 
+  /* ************************************************************************* */
+  template <class FACTORGRAPH>
```
dellaert, line 156: The implementation inside eliminateMultifrontal still feels slightly out of place in this file. But I can live with it. The reason for my comment is that the other three traversals live in/near junction tree land. Maybe IndexedJunctionTree.h should move from Symbolic to inference, and the implementation can live there, in IndexedJunctionTree-inst.h ?

dellaert (commented):
Okay, this is brilliant. Can you feel the stars align? :-)

gtsam/inference/EliminateableFactorGraph-inst.h
tzvist: I’d prefer to keep it here if you’re okay with that. It feels odd for IndexedJunctionTree-inst.h to depend on Gaussian graph. In my view, IndexedJunctionTree is a very pure data structure that shouldn’t be concerned with the specific factor implementations.

gtsam/symbolic/IndexedJunctionTree.h
tzvist: done

gtsam/inference/EliminateableFactorGraph-inst.h
dellaert: The unit testing in `symbolic` proves it does not depend on Gaussian graph components. In fact, if it did, it would not be in `inference`, which should not include anything from linear. 

However, I agree `IndexedJunctionTree` seems very in the right place in `symbolic`. 

I'm fine with leaving this here.

dellaert:
Super happy with this right now. I propose we also wait for the benchmarking PR by @ProfFan so we can merge with confidence.

tzvist:
> Super happy with this right now. I propose we also wait for the benchmarking PR by @ProfFan so we can merge with confidence.

cool, maybe it will be interesting to compare the multifrontal solver with the LM after my changes.

ProfFan:
@dellaert Benchmark results look good :)

dellaert:
> @dellaert Benchmark results look good :)

They look mixed right? seems like 25% slower on Mac no?

i’m wondering whether that BAL16 dataset is too small, as well. I’ve been using the 135 cameras.

dellaert:
Also, I’m now thinking we should do with and without TBB

ProfFan:
I can add that

tzvist:
> /bench

rebased maybe this will help

ProfFan:
btw bench results will be refreshed at the original comment

dellaert:
Oops - does not look good with TBB on… @tzvist any theories?

tzvist:
> Oops - does not look good with TBB on… @tzvist any theories?


Turns out I had accidentally removed the parallel eliminate implementation. Luckily this benchmark exposed it. pushed a fix commit.

tzvist:
My example running with TBB 12 cores 
Before this MR:
```
./build/simple_gtsam_deserialize
simple_gtsam_deserialize2.cpp
Reading values file took 0.0123 seconds
Reading graph file took 0.2029 seconds
Processing factor lines took 2.8737 seconds
Processing values lines took 0.1093 seconds
Setting up optimizer took 9.1637 seconds
Initial error: 83467.7055, values: 65796
iter      cost      cost_change    lambda  success iter_time
   0          inf         0.00       0.00      0       1.99
iter      cost      cost_change    lambda  success iter_time
   0     80421.88      3045.82       0.00      1       6.30
   1          inf         0.00       0.00      0       1.21
   1     80163.07       258.81       0.00      1       3.49
   2          inf         0.00       0.00      0       1.19
   2     80113.89        49.18       0.00      1       3.16
   3          inf         0.00       0.00      0       1.20
   3     80096.39        17.50       0.00      1       3.23
   4          inf         0.00       0.00      0       1.22
   4     80089.79         6.60       0.00      1       3.28
   5          inf         0.00       0.00      0       1.16
   5     80086.41         3.38       0.00      1       3.35
   6          inf         0.00       0.00      0       1.17
   6     80084.44         1.98       0.00      1       3.24
   7          inf         0.00       0.00      0       1.16
   7     80083.05         1.39       0.00      1       3.26
   8          inf         0.00       0.00      0       1.21
   8     80081.84         1.21       0.00      1       3.27
   9          inf         0.00       0.00      0       1.19
   9     80080.52         1.32       0.00      1       3.24
  10          inf         0.00       0.00      0       1.20
  10     80078.92         1.60       0.00      1       3.52
  11          inf         0.00       0.00      0       1.18
  11     80076.99         1.93       0.00      1       3.24
  12          inf         0.00       0.00      0       1.17
  12     80074.94         2.05       0.00      1       3.29
  13          inf         0.00       0.00      0       1.21
  13     80073.22         1.72       0.00      1       3.12
  14          inf         0.00       0.00      0       1.17
  14     80071.87         1.34       0.00      1       3.21
  15          inf         0.00       0.00      0       1.21
  15     80071.04         0.83       0.00      1       3.28
  16          inf         0.00       0.00      0       1.18
  16     80070.76         0.28       0.00      1       3.22
Running gtsam optimizer took 87.3073 seconds
````

After
```
./build/simple_gtsam_deserialize
simple_gtsam_deserialize2.cpp
Reading values file took 0.0112 seconds
Reading graph file took 0.1967 seconds
Processing factor lines took 2.8551 seconds
Processing values lines took 0.1035 seconds
Setting up optimizer took 8.8668 seconds
Initial error: 83467.7055, values: 65796
iter      cost      cost_change    lambda  success iter_time
   0          inf         0.00       0.00      0       2.42
iter      cost      cost_change    lambda  success iter_time
   0     80421.88      3045.82       0.00      1       6.37
   1          inf         0.00       0.00      0       0.36
   1     80163.07       258.81       0.00      1       2.31
   2          inf         0.00       0.00      0       0.34
   2     80113.89        49.18       0.00      1       2.64
   3          inf         0.00       0.00      0       0.34
   3     80096.39        17.50       0.00      1       2.75
   4          inf         0.00       0.00      0       0.35
   4     80089.79         6.60       0.00      1       2.53
   5          inf         0.00       0.00      0       0.27
   5     80086.41         3.38       0.00      1       2.63
   6          inf         0.00       0.00      0       0.34
   6     80084.44         1.98       0.00      1       2.42
   7          inf         0.00       0.00      0       0.33
   7     80083.05         1.39       0.00      1       2.71
   8          inf         0.00       0.00      0       0.33
   8     80081.84         1.21       0.00      1       2.43
   9          inf         0.00       0.00      0       0.43
   9     80080.52         1.32       0.00      1       2.55
  10          inf         0.00       0.00      0       0.40
  10     80078.92         1.60       0.00      1       2.74
  11          inf         0.00       0.00      0       0.32
  11     80076.99         1.93       0.00      1       2.33
  12          inf         0.00       0.00      0       0.32
  12     80074.94         2.05       0.00      1       2.68
  13          inf         0.00       0.00      0       0.32
  13     80073.22         1.72       0.00      1       2.50
  14          inf         0.00       0.00      0       0.34
  14     80071.87         1.34       0.00      1       2.43
  15          inf         0.00       0.00      0       0.40
  15     80071.04         0.83       0.00      1       2.89
  16          inf         0.00       0.00      0       0.31
  16     80070.76         0.28       0.00      1       2.56
Running gtsam optimizer took 63.1366 seconds
```

**Comment by tzvist** [comment]
> /bench

please rerun again

**Comment by dellaert** [comment]
@ProfFan I cannot find the new benchmark results.

**Comment by ProfFan** [comment]
```
FAILED: [code=1] gtsam/CMakeFiles/gtsam.dir/symbolic/SymbolicFactorGraph.cpp.o 
/usr/bin/c++ -DNDEBUG -Dgtsam_EXPORTS -I/Users/runner/work/gtsam/gtsam -I/Users/runner/work/gtsam/gtsam/build -I/Users/runner/work/gtsam/gtsam/CppUnitLite -I/Users/runner/work/gtsam/gtsam/gtsam/3rdparty/metis/include -I/Users/runner/work/gtsam/gtsam/gtsam/3rdparty/metis/libmetis -I/Users/runner/work/gtsam/gtsam/gtsam/3rdparty/metis/GKlib -I/Users/runner/work/gtsam/gtsam/gtsam/3rdparty/cephes -isystem /Users/runner/work/gtsam/gtsam/gtsam/3rdparty/SuiteSparse_config -isystem /Users/runner/work/gtsam/gtsam/gtsam/3rdparty/Spectra -isystem /Users/runner/work/gtsam/gtsam/gtsam/3rdparty/CCOLAMD/Include -isystem /opt/homebrew/include -isystem /Users/runner/work/gtsam/gtsam/gtsam/3rdparty/Eigen -w -O3 -DNDEBUG -std=c++17 -arch arm64 -fPIC -Werror -Wall -Wpedantic -Wextra -Wno-unused-parameter -Wreturn-type -Wformat -Werror=format-security -Wsuggest-override -O3 -MD -MT gtsam/CMakeFiles/gtsam.dir/symbolic/SymbolicFactorGraph.cpp.o -MF gtsam/CMakeFiles/gtsam.dir/symbolic/SymbolicFactorGraph.cpp.o.d -o gtsam/CMakeFiles/gtsam.dir/symbolic/SymbolicFactorGraph.cpp.o -c /Users/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicFactorGraph.cpp
In file included from /Users/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicFactorGraph.cpp:20:
/Users/runner/work/gtsam/gtsam/gtsam/inference/EliminateableFactorGraph-inst.h:169:22: error: no matching constructor for initialization of 'JunctionTreeType' (aka 'gtsam::SymbolicJunctionTree')
  169 |     JunctionTreeType junctionTree(asDerived(), indexedJunctionTree);
      |                      ^            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicFactorGraph.cpp:31:18: note: in instantiation of member function 'gtsam::EliminateableFactorGraph<gtsam::SymbolicFactorGraph>::eliminateMultifrontal' requested here
   31 |   template class EliminateableFactorGraph<SymbolicFactorGraph>;
      |                  ^
/Users/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicJunctionTree.h:65:5: note: candidate constructor not viable: requires single argument 'eliminationTree', but 2 arguments were provided
   65 |     SymbolicJunctionTree(const SymbolicEliminationTree& eliminationTree);
      |     ^                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicJunctionTree.h:50:22: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 2 were provided
   50 |   class GTSAM_EXPORT SymbolicJunctionTree :
      |                      ^~~~~~~~~~~~~~~~~~~~
/Users/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicJunctionTree.h:50:22: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 2 were provided
   50 |   class GTSAM_EXPORT SymbolicJunctionTree :
      |                      ^~~~~~~~~~~~~~~~~~~~
1 error generated.
[147/275] Building CXX object gtsam/CMakeFiles/gtsam.dir/symbolic/SymbolicISAM.cpp.o
ninja: build stopped: subcommand failed.
Error: Process completed with exit code 1.
```

They failed

dellaert:
Oops. How can that be?

dellaert:
CI fails as well. - @tzvist please check if you can compile (and build wrapper) locally?

tzvist:
removed problematic commit

ProfFan:
@dellaert bench results is up

dellaert:
Incredible guys, this is a fantastic PR and, @ProfFan, your benchmark action is a fantastic addition. It just shows how good this contribution is. Thank you so much, @tzvist !! Will merge now.

dellaert (approved):

PR #2341, opened 2026-01-08
Jacobian implementations for sim3 transformFrom

akshay-krishnan:
Also added the jacobians for the member functions in sim3, with unit tests. 

Thanks @dellaert for lots of help!

### DIFF
--- a/gtsam/geometry/Similarity3.cpp
+++ b/gtsam/geometry/Similarity3.cpp
@@ -118,6 +118,29 @@ void Similarity3::print(const std::string& s) const {
   std::cout << "t: " << translation().transpose() << " s: " << scale() << std::endl;
 }
 
+Rot3 Similarity3::rotation(OptionalJacobian<3, 7> Hself) const { 
+  if (Hself) {
+    Hself->setZero();
+    Hself->block<3, 3>(0, 0) = I_3x3;
+  }
+  return R_;
+}
+
+Point3 Similarity3::translation(OptionalJacobian<3, 7> Hself) const {
+  if (Hself) {
+    *Hself << Z_3x3, rotation().matrix(), -t_; 
+  }
+  return t_;
+}
+
+double Similarity3::scale(OptionalJacobian<1, 7> Hself) const {
+  if (Hself) {
+    Hself->setZero();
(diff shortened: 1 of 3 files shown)


dellaert (approved):
I think the Jacobians can still be simplified by inlining the getter Jacobians, which are often simple.

gtsam/geometry/Similarity3.cpp
```diff
@@ -145,9 +168,34 @@ Point3 Similarity3::transformFrom(const Point3& p, //
...
+  // Delegate the translation jacobians to the point3 transformFrom.
+  Matrix37 Dt_dsim;
+  Matrix3 Dt_dp;
+  const Point3 t = transformFrom(bti, Hself ? &Dt_dsim : nullptr, H_bTi ? &Dt_dp : nullptr);
+
+  if (Hself) {
+    Hself->setZero();
+    Hself->block<3, 3>(0, 0) = bRi.transpose(); // DR_dsimR
+    // Chain D_result_t (ie R.T) * D_t_sim (3x7).
+    Hself->block<3, 7>(3, 0) = R.transpose() * Dt_dsim;
```
dellaert, line 190: Because Dt_sim = R_ * (some 3x7 matrix) and you multiply with R.transpose() = bRi.transpose() * R_.transpose(), you can simply even more, but it will be messier, so this is good.

gtsam/geometry/Similarity3.cpp
```diff
@@ -145,9 +168,34 @@ Point3 Similarity3::transformFrom(const Point3& p, //
...
+  }
+
+  if (H_bTi) {
+    H_bTi->setZero();
+    H_bTi->block<3, 3>(0, 0) = I_3x3; // DR_dTR = I_3x3
+    // Chain D_result_t * D_t_p * D_p_bTi (ie [Z_3x3, bRi]) .
+    H_bTi->block<3, 3>(3, 3) = R.transpose() * Dt_dp * bRi.matrix();
```
dellaert: However, this one should be simplified, as Dt_dp = s_ * R_, so I think this just becomes:
```
s_ * I_3x3
```
?

dellaert:
Merged and branch deleted

PR #2348, opened 2026-01-11
Minor fixes

varunagrawal:
Ran into some issues when compiling GTSAM so added some fixes and other improvements.
The primary one is:
```
In file included from /usr/include/c++/11/cassert:44,
                 from /tmp/gtsam_source_linux/gtsam/linear/MultifrontalClique.cpp:25:
/tmp/gtsam_source_linux/gtsam/linear/MultifrontalClique.cpp: In member function ‘void gtsam::MultifrontalClique::fillAb(const gtsam::GaussianFactorGraph&)’:
/tmp/gtsam_source_linux/gtsam/linear/MultifrontalClique.cpp:234:17: error: suggest parentheses around ‘&&’ within ‘||’ [-Werror=parentheses]
  234 |   assert((useQR && RSd_.matrix().rows() ==
      |           ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
  235 |                        static_cast<DenseIndex>(Ab_.matrix().rows()) ||
      |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

The doxygen syntax issue gives output like the below image:
Screenshot 2026-01-11 at 6 26 41 PM

PS there are a lot of commented out code blocks and tests. Perhaps we should also clean those up?


dellaert (approved):
LGTM. I don’t fully understand why the CI doesn’t get the same compilation errors, but this is good.

varunagrawal:
I was pretty surprised too. Maybe this issue is cropping up due to a newer GCC version but existing PRs haven't been rebased on the new CI?

PR #2349, opened 2026-01-12
Minimize the number of workflows

varunagrawal:
A lot of workflows which are triggered on Pull Requests are also triggered when merged into `develop`. This is redundant since we can't merge into develop unless those workflows pass.

This PR eliminates a lot of these redundant workflows to make the CI system more efficient, and take up less resources when deploying updated builds and wheels.

### DIFF
--- a/.github/workflows/build-cibw.yml
+++ b/.github/workflows/build-cibw.yml
@@ -1,7 +1,7 @@
 # This workflow builds the Python wheels using cibuildwheel and uploads them to TestPyPI.
 # It can be triggered on push to the develop branch or manually via Github Actions.
 
-name: Build Wheels for Develop
+name: Build Python Wheels for Develop
 
 on:
   push:
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -3,10 +3,9 @@ name: Linux CI
 on:
   pull_request:
     paths-ignore:
-      - "**.md"
-      - "**.ipynb"
-      - "myst.yml"
-      
+      - '**.md'
+      - '**.ipynb'
+      - 'myst.yml'
 
 # Cancels any in-progress workflow runs for the same PR when a new push is made,
 # allowing the runner to become available more quickly for the latest changes.
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -6,6 +6,7 @@ on:
       - '**.md'
       - '**.ipynb' 
       - 'myst.yml'
+
 # Cancels any in-progress workflow runs for the same PR when a new push is made,
 # allowing the runner to become available more quickly for the latest changes.
 concurrency:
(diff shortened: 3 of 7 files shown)


dellaert (approved):

talregev:
@dellaert This is mistake. 
Code that can break gtsam specailly in linux with eigen3 v5.x can inserted to develop, and no one will bother to fix it, while it there, only ignore it always. 
Currently vcpkg is the only ci that testing gtsam with eigen3 v5.x.

dellaert:
Good thinking! I definitely want to make sure that we catch regressions for Eigen 5, but would the PR CI not make sure of this?

talregev:
What we need is to create new PR that will activate vcpkg jobs on pull request. (revert 1 line in vcpg.yml ci file).
I can create such PR

varunagrawal:
> @dellaert This is mistake. 
> Code that can break gtsam specailly in linux with eigen3 v5.x can inserted to develop, and no one will bother to fix it, while it there, only ignore it always. 
> Currently vcpkg is the only ci that testing gtsam with eigen3 v5.x.
> 

Shouldn't we have a CI which tests for Eigen3 v5.x specifically instead? Similar to how we have special cases?

Packaging a branch other than develop breaks the gitflow model we have. I'm not so sure that testing and packaging should be mixed like that.

talregev:
vcpkg is not packaging. Is only testing.
If you only test only on PR it can break when you merge.
If you test only on Merge, it will break next PR.
That why you test in both, that why majority of repos do it.

Gold856:
DevOps guy is not happy. :(

Yeah, this is bad for PR caching as well. Windows CI now frequently takes an additional 30 minutes or more when it doesn't necessarily need to. I checked the Actions page and the Windows CI section and made a list of 18 times it took for the Windows CI workflow to complete on newly opened PRs (this is important for comparison because pushes to PRs will reuse the PR cache), skipped any failed jobs and starting at run 4493 (which is the run for this PR) and going back in time. These are times:
```
Workflow completion time -> Windows Release job completion time
27m 31s -> 7m 22s
1h 0m 29s -> 10m 16s
27m 5s -> 15m 16s
1h 9m 55s -> 1h 9m
1h 29m 51s -> 1h 8m
1h 32m 59s -> 1h 9m
23m 58s -> 8m 34s
48m 17s -> 16m 8s
1h 3m 56s -> 1h 3m
43m 20s -> 43m 16s
43m 59s -> 43m 53s
23m 29s -> 11m 34s
26m 51s -> 12m 8s
1h 5m 29s -> 1h 5m
1h 2m 23s -> 1h 2m
29m 48s -> 7m 35s
1h 6m 12s -> 47m 48s
50m 46s -> 10m 10s
```
Compare that to all the PRs opened since this PR was merged, which essentially all take 1 hour, 15 minutes to complete. And you actually still have caches for cephes and metis because PRs actively use them, so it'll be longer when those become stale. Just over 50% of PRs saw workflow completion time drop by over 30 minutes, which is a solid time save across all the workflows that have to run, and only 5/18 PRs had to essentially compile from scratch (actually, this is worse than I expected; seems like headers that are included everywhere get modified somewhat frequently). The fact there's no longer a cache for `develop` also means PRs consume more cache than necessary, since they can't reuse the `develop` cache, which can cause cache thrashing and cause more cache misses than necessarily (although I don't think there's enough PRs open in parallel at any given time for that to happen, but it's still wasteful.)

PR #2360, opened 2026-01-17
switch Hat/Vee to orthonormal sl(4) basis and update adjoint/tests

yluo5820:
Switched SL4’s Lie algebra coordinates from the old “raw off‑diagonal + chained diagonal” parameterization to a geometrically meaningful orthonormal basis. Hat/Vee now decompose sl(4) into 3 parts: 6 skew‑symmetric rotation generators, 6 symmetric off‑diagonal shear generators, and 3 orthonormal traceless diagonal scalings. Updated the adjoint vectorization matrices to match this basis and refreshed tests to validate trace‑zero, round‑trip Hat↔Vee, and orthonormal generator inner products.

### DIFF
--- a/gtsam/geometry/SL4.cpp
+++ b/gtsam/geometry/SL4.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <gtsam/geometry/SL4.h>
+#include <gtsam/geometry/SO4.h>
 
 // To use exp(), log()
 #include <cmath>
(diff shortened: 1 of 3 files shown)


gtsam/geometry/SL4.cpp
```diff
@@ -185,28 +194,69 @@ Matrix44 SL4::Hat(const Vector& xi) {
...
-  const double d11 = xi(12);
-  const double d22 = -xi(12) + xi(13);
-  const double d33 = -xi(13) + xi(14);
-  const double d44 = -xi(14);
-
-  A << d11, xi(0), xi(1), xi(2), xi(3), d22, xi(4), xi(5), xi(6), xi(7), d33,
-      xi(8), xi(9), xi(10), xi(11), d44;
+  Matrix44 A = Matrix44::Zero();
+
+  // Rotations (skew-symmetric), normalized by 1/sqrt(2).
```
dellaert: You could use SO4::Hat. And, similar to that, I'd at least internally define a (in the anonymous namespace above) a hatSym4. And you could then add with an eigen "diagonal" for the diagonal part.

gtsam/geometry/SL4.cpp
```diff
@@ -185,28 +194,69 @@ Matrix44 SL4::Hat(const Vector& xi) {
...
+  A(0, 1) += kInvSqrt2 * xi(0);
```
dellaert: From our chat before, I thought that only the diagonal would be rescaled somehow. But you found out we need square root of 2 for everything?

gtsam/geometry/SL4.cpp
```diff
@@ -185,28 +194,69 @@ Matrix44 SL4::Hat(const Vector& xi) {
...
-  const double x12 = X(0, 0);
-  const double x13 = X(1, 1) + x12;
-  const double x14 = -X(3, 3);
-  vec << X(0, 1), X(0, 2), X(0, 3), X(1, 0), X(1, 2), X(1, 3), X(2, 0), X(2, 1),
-      X(2, 3), X(3, 0), X(3, 1), X(3, 2), x12, x13, x14;
-  return vec;
+// used consistent notation with Hat()
+Vector SL4::Vee(const Matrix44& A) {
+  Vector xi(15);
+  xi << kInvSqrt2 * (A(0, 1) - A(1, 0)),
```
dellaert: Vee is rarely used so I'd opt for math clarity here as well. Form A-A' and A+A' and use SO4::vee, and an internal veeSym4

gtsam/geometry/SL4.h
```diff
@@ -102,6 +102,20 @@ class GTSAM_EXPORT SL4 : public MatrixLieGroup<SL4, 15, 4> {
   /// @{
 
   using LieAlgebra = Matrix44;
+  /**
+   * Lie algebra coordinates for sl(4) using an orthonormal basis.
+   *
+   * Ordering of xi (15x1):
+   *   [r12 r13 r14 r23 r24 r34 s12 s13 s14 s23 s24 s34 h1 h2 h3]
+   *
+   * Basis:
```
dellaert, line 116: Document on direct vector product of the three Lie algebras?

gtsam/geometry/SL4.cpp
yluo5820: I think the goal was mainly to decompose the lie algebra so that we can twink sampling on the subspace separately.  From the chat before I think only the diagonals were mentioned to be orthonormal, but in the test seciton of the prompt there was an item having the whole basis orthonormal, and codex probably did the scale to implement that. I guess our goal is indifferent to whether it's normalized, but if there's a preference we can do either

gtsam/geometry/SL4.h
yluo5820: added some lines for that, though I just learnt that the direct sum is not on a Lie algebra level but only on a vector space level, as symmetric off-diagonal shears are not closed under Lie bracket

dellaert (approved):
Nice !

PR #2361, opened 2026-01-17
deltaError replaces 2 error calls

dellaert:
In the nonlinear optimizer, we're calling the quite expensive `error` function twice. And essentially, we're doing double the work needed. The new `deltaError` method avoids this, and does the most optimal thing in the factors themselves. This reduces the number of work in each iteration. At least a couple of percent gain total.

The bigger gain is for the new solver - in a PR to come, which will basically do this for free.

### DIFF
--- a/gtsam/linear/GaussianFactor.cpp
+++ b/gtsam/linear/GaussianFactor.cpp
@@ -28,6 +28,20 @@ double GaussianFactor::error(const VectorValues& c) const {
   throw std::runtime_error("GaussianFactor::error is not implemented");
 }
 
+double GaussianFactor::deltaError(const VectorValues& c, double* oldError,
+                                  double* newError) const {
+  const VectorValues zero = VectorValues::Zero(c);
+  const double oldValue = error(zero);
+  const double newValue = error(c);
+  if (oldError) {
+    *oldError = oldValue;
+  }
+  if (newError) {
+    *newError = newValue;
+  }
+  return oldValue - newValue;
+}
+
 double GaussianFactor::error(const HybridValues& c) const {
   return this->error(c.continuous());
 }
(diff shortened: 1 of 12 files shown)


gtsam/linear/GaussianFactor.cpp
```diff
@@ -28,6 +28,20 @@ double GaussianFactor::error(const VectorValues& c) const {
...
+  const VectorValues zero = VectorValues::Zero(c);
+  const double oldValue = error(zero);
+  const double newValue = error(c);
+  if (oldError) {
+    *oldError = oldValue;
+  }
+  if (newError) {
+    *newError = newValue;
+  }
+  return oldValue - newValue;
```
akshay-krishnan, line 42: maybe sign does not matter here? should this not be `newValue - oldValue`?

gtsam/linear/JacobianFactor.cpp
```diff
@@ -556,6 +556,20 @@ double JacobianFactor::error(const VectorValues& c) const {
...
+                                  double* newError) const {
+  const Vector e = unweighted_error(c);
+  const Vector b = getb();
+  double oldValue =
+      model_ ? 0.5 * model_->squaredMahalanobisDistance(b) : 0.5 * b.dot(b);
+  double newValue =
+      model_ ? 0.5 * model_->squaredMahalanobisDistance(e) : 0.5 * e.dot(e);
+  if (oldError) *oldError = oldValue;
+  if (newError) *newError = newValue;
+  return oldValue - newValue;
```
akshay-krishnan, line 570: same question about the sign here.

akshay-krishnan (approved):

gtsam/linear/GaussianFactor.cpp
akshay-krishnan: or maybe this is a gtsam convention I am not aware of.

dellaert: The old value is typically bigger than the new value. We are opting for positive number.

PR #2363, opened 2026-01-18
Default noise models and TranslationRecovery patch

dellaert:
To support the new MF solver, I added default unit noise models in the unary and binary measurements. This broke translation recovery because there was a mismatch between the measurements and noise models.This PR does not address the deeper underlying issue but just patches it up.

It also removes a `using namespace std` in a header (big nono), which necessitated some changes in Shonan averaging.

### DIFF
--- a/gtsam/sfm/TranslationRecovery.cpp
+++ b/gtsam/sfm/TranslationRecovery.cpp
@@ -17,6 +17,7 @@
  */
 
 #include <gtsam/base/DSFMap.h>
+#include <gtsam/base/VectorSpace.h>
 #include <gtsam/geometry/Point3.h>
 #include <gtsam/geometry/Pose3.h>
 #include <gtsam/geometry/Unit3.h>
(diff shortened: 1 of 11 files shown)


gtsam/sfm/TranslationRecovery.cpp
```diff
@@ -229,11 +247,9 @@ Values TranslationRecovery::run(
 
 TranslationRecovery::TranslationEdges TranslationRecovery::SimulateMeasurements(
     const Values &poses, const vector<KeyPair> &edges) {
-  auto edgeNoiseModel = noiseModel::Isotropic::Sigma(3, 0.01);
+  auto edgeNoiseModel = noiseModel::Isotropic::Sigma(2, 0.01);
```
akshay-krishnan, line 250: good catch! I am surprised the TranslationFactor works fine with either 3 or 2, i.e it does not fail for 3? 🤔

akshay-krishnan (approved):

gtsam/sfm/TranslationRecovery.cpp
dellaert: It fails now :-)

PR #2365, opened 2026-01-20
Revert linear solver type to MULTIFRONTAL_CHOLESKY 

akshay-krishnan:
Revert linear solver type to MULTIFRONTAL_CHOLESKY from MULTIFRONTAL_SOLVER


dellaert (approved):
I guess the performance benchmarks I ran were too specific towards structure from motion with BAL, and we’re seeing regressions on other data sets. So, for now this is absolutely the right thing to do, until we get until we get more resolution on what is happening.

Bummer!

PR #2367, opened 2026-01-21
sccache for all os in vcpkg

talregev:
sccache for all os in vcpkg

After running second time on my repo:
https://github.com/talregev/gtsam/pull/56

Windows: 46m 8s
Ubuntu: 9m 27s
Macos: 13m 32s

### DIFF
--- a/.github/workflows/vcpkg.yml
+++ b/.github/workflows/vcpkg.yml
@@ -59,11 +59,9 @@ jobs:
           toolset: 14.40
 
       - name: Install sccache
-        if: runner.os == 'Windows'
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Woah !

Gold856:
Oooh, very nice! Love seeing CI times go down. Cache is looking kinda full but I think most of that is stale. I'd just watch the number of cache misses and see if it's unexpectedly high (is stuff that should be cached not being cached for some reason?) But otherwise, looks like a pretty solid boost! I'm gonna start some work to try and squeeze out those last few minutes to get CI really fast.

PR #2368, opened 2026-01-22
Expose EqF Python bindings (ABC), template on N, modify/add examples

rohan-bansal:
Summary:

- Add ABCEqF wrapper class (`AbcEquivariantFilter`) that templates on N
- Translated Python example for ABC EqF.
- Used ABCEqF in original C++ example to simplify a bit more

We can possibly migrate the examples to AwesomeEqF! I added them just to test the different instantiations.


gtsam_unstable/geometry/ABCEqFWrapper.h
```diff
@@ -0,0 +1,71 @@
...
+#include <gtsam/base/Vector.h>
+#include <gtsam/geometry/Rot3.h>
+#include <gtsam/geometry/Unit3.h>
+#include <gtsam/navigation/EquivariantFilter.h>
+#include <gtsam_unstable/geometry/ABC.h>
+
+namespace gtsam {
+namespace abc {
+
+class AbcEquivariantFilter1 {
```
dellaert: Not loving this. Could be not instead just as easily fix our wrapper generation?

gtsam/navigation/navigation.i
```diff
@@ -651,6 +651,27 @@ virtual class InvariantEKF : gtsam::LeftLinearEKF<G> {
   void predict(const gtsam::Vector& u, double dt, gtsam::Matrix Q);
 };
 
+// ---------------------------------------------------------------------------
+// ABC Equivariant Filter wrapper (N=1)
```
dellaert: Not just N=1 now :-) Also, let's get rid of all "wrapper" language. It's just an EqF filter specialized for ABC now

python/gtsam/examples/AbcEquivariantFilterExample.py
```diff
@@ -0,0 +1,250 @@
+"""
+Python translation of examples/AbcEquivariantFilterExample.cpp.
```
dellaert, line 2: Could we immediately make it a notebook? Would be cool to add some plotly time-series as well of the state and calibration time series.

gtsam_unstable/geometry/ABCEquivariantFilter.h
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * The filter uses the ABC Lie group structure and equivariant dynamics to
+ * provide consistent state estimation. It inherits from EquivariantFilter
+ * and provides a simplified interface for ABC-specific operations.
+ *
+ * @tparam N Number of calibrated sensors (typically 1 or more)
+ */
+template <size_t N>
+class AbcEquivariantFilter
+    : public gtsam::EquivariantFilter<State<N>, Symmetry<N>> {
```
dellaert, line 42: Nice!!!

examples/AbcEquivariantFilterExample.cpp
```diff
@@ -1,32 +1,31 @@
...
  * @file AbcEquivariantFilterExample.cpp
- * @brief Demonstration of the full Attitude-Bias-Calibration Equivariant Filter
+ * @brief Demonstration of the Attitude-Bias-Calibration Equivariant Filter
  *
  * This demo shows the Equivariant Filter (EqF) for attitude estimation
  * with both gyroscope bias and sensor extrinsic calibration, based on the
  * paper: "Overcoming Bias: Equivariant Filter Design for Biased Attitude
  * Estimation with Online Calibration" by Fornasier et al.
  *
+ * This example uses the simplified AbcEquivariantFilter wrapper class which
```
dellaert: Not wrapper anymore

dellaert (approved):
Awesome! I think it's fine to merge, but I think it would be even cooler if it was a notebook with some graphics :-) And, indeed, I think there is a case to be made to immediately move the notebook to Awesome EqF

dellaert (approved):
Nice !!!

PR #2369, opened 2026-01-24
Simple GNSS Pseudorange Factor

masoug:
Hi! This PR proposes a simple `PseudorangeFactor` that provides position (and clock timing) constraints from GNSS satellite data to GTSAM estimation problems. So far, the simple pseudorange factor focuses on just clock errors, and omits atmospheric effects. Furthermore, it is designed for code-phase measurements only, so it does not leverage high-precision carrier-phase data from the receiver. Despite that, this simple model already provides decent positioning accuracy on the scale of a city block, and my hope is to contribute several follow-up PRs that gradually improve accuracy/precision with better factor models and examples.

A basic demonstration of `PseudorangeFactor` is provided in the `SinglePointPositioningExample.ipynb` notebook. It captures [ZOA1's position](https://www.ngs.noaa.gov/CORS/ncn_station_pages/index.html?stationID=ZOA1) to ~33 meters of accuracy:
Screenshot 2026-01-23 at 8 13 26 PM

I'd like to get some feedback on this proposal while I continue working on the unit tests in the background. I think a family of pseudorange factors is interesting because it enables tightly-coupled GNSS factor graphs, which are more robust compared to `GPSFactor`s. Individual pseudoranges provide fine-grain outlier rejection while offering partial observability when < 4 satellites are visible. Complex multi-station differential-GNSS positioning graphs are also enabled by pseudorange factors.

Anyway, this is my first GTSAM contribution, so please let me know if I'm missing any convention/formatting in my PR. Otherwise, I'm looking forward to exploring precise GNSS positioning with factor graphs!

### DIFF
--- a/gtsam/navigation/PseudorangeFactor.h
+++ b/gtsam/navigation/PseudorangeFactor.h
@@ -1,0 +1,107 @@
+/**
+ *  @file   PseudorangeFactor.h
+ *  @author Sammy Guo
+ *  @brief  Header file for GNSS Pseudorange factor
+ *  @date   January 18, 2026
+ **/
+#pragma once
+
+#include <gtsam/geometry/Point3.h>
+#include <gtsam/nonlinear/NoiseModelFactorN.h>
+#include <gtsam/nonlinear/NonlinearFactor.h>
+
+#include <string>
+
+namespace gtsam {
+
+/**
+ * Simplified GNSS pseudorange model for basic positioning problems.
+ *
+ * This factor implements a simplified version of equation 5.6 [1]
+ * \rho = r + c[\delta t_u - \delta t^s] + I_{\rho} + T_{\rho} + \epsilon_{\rho}
+ * where `\rho` is measured pseudorange (in meters) from the receiver,
+ * `r` true range (in meters) between receiver antenna and satellite,
+ * `c` is speed of light in a vacuum (m/s),
+ * `\delta t_u` is receiver clock bias (seconds),
+ * `\delta t_s` is satellite clock bias (seconds),
+ * and `I_{\rho}`, `T_{\rho}`, and `\epsilon_{\rho}` are ionospheric,
+ * tropospheric, and unmodeled errors respectively.
+ *
+ * Ionospheric and tropospheric terms are omitted in this simplified factor.
+ *
+ * @ingroup navigation
+ *
+ * REFERENCES:
+ * [1] P. Misra et. al., "Global Positioning Systems: Signals, Measurements, and
+ * Performance", Second Edition, 2012.
+ */
+class GTSAM_EXPORT PseudorangeFactor
+    : public NoiseModelFactorN<Point3, double> {
+ private:
+  typedef NoiseModelFactorN<Point3, double> Base;
+
+  const double
+      pseudorange_;  ///< Receiver-reported pseudorange measurement in meters.
+  const Point3 sat_pos_;       ///< Satellite position in WGS84 ECEF meters.
+  const double sat_clk_bias_;  ///< Satellite clock bias in seconds.
+
+ public:
+  // Provide access to the Matrix& version of evaluateError:
+  using Base::evaluateError;
+
+  /// shorthand for a smart pointer to a factor
+  typedef std::shared_ptr<PseudorangeFactor> shared_ptr;
+
+  /// Typedef to this class
+  typedef PseudorangeFactor This;
+
+  /** default constructor - only use for serialization */
+  PseudorangeFactor();
+
+  virtual ~PseudorangeFactor() = default;
+
+  /**
+   * @brief TODO
+   */
+  PseudorangeFactor(Key receiver_position_key, Key receiver_clock_bias_key,
+                    double measured_pseudorange,
+                    const Point3& satellite_position,
+                    double satellite_clock_bias, const SharedNoiseModel& model);
+
+  /// @return a deep copy of this factor
+  gtsam::NonlinearFactor::shared_ptr clone() const override {
+    return std::static_pointer_cast<gtsam::NonlinearFactor>(
+        gtsam::NonlinearFactor::shared_ptr(new This(*this)));
+  }
+
+  /// print
+  //  void print(const std::string& s = "", const KeyFormatter& keyFormatter =
+  //                                            DefaultKeyFormatter) const
+  //                                            override;
+
+  /// equals
+  //  bool equals(const NonlinearFactor& expected, double tol = 1e-9) const
+  //  override;
+
+  /// vector of errors
+  Vector evaluateError(const Point3& receiver_position,
+                       const double& receiver_clock_bias,
+                       OptionalMatrixType Hreceiver_pos,
+                       OptionalMatrixType Hreceiver_clock_bias) const override;
+
+ private:
+#if GTSAM_ENABLE_BOOST_SERIALIZATION  ///
+  /// Serialization function
+  friend class boost::serialization::access;
+  template <class ARCHIVE>
+  void serialize(ARCHIVE& ar, const unsigned int /*version*/) {
+    ar& boost::serialization::make_nvp(
+        "NoiseModelFactor2", boost::serialization::base_object<Base>(*this));
+    ar& BOOST_SERIALIZATION_NVP(pseudorange_);
+    ar& BOOST_SERIALIZATION_NVP(sat_pos_);
+    ar& BOOST_SERIALIZATION_NVP(sat_clk_bias_);
+  }
+#endif
+};
+
+}  // namespace gtsam
--- a/gtsam/navigation/PseudorangeFactor.h
+++ b/gtsam/navigation/PseudorangeFactor.h
@@ -41,10 +41,10 @@ class GTSAM_EXPORT PseudorangeFactor
  private:
   typedef NoiseModelFactorN<Point3, double> Base;
 
-  const double
-      pseudorange_;  ///< Receiver-reported pseudorange measurement in meters.
-  const Point3 satPos_;      ///< Satellite position in WGS84 ECEF meters.
-  const double satClkBias_;  ///< Satellite clock bias in seconds.
+  double
+      pseudorange_;    ///< Receiver-reported pseudorange measurement in meters.
+  Point3 satPos_;      ///< Satellite position in WGS84 ECEF meters.
+  double satClkBias_;  ///< Satellite clock bias in seconds.
 
  public:
   // Provide access to the Matrix& version of evaluateError:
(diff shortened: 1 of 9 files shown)


gtsam/navigation/doc/PseudorangeFactor.ipynb
```diff
@@ -0,0 +1,121 @@
+{
```
dellaert, line 1: Please ensure GTSAM is not installed unless run on Colab. Take a look at python/gtsam/notebooks/standardize_notebooks.py

gtsam/navigation/navigation.i
```diff
@@ -485,6 +485,25 @@ virtual class GPSFactor2ArmCalib : gtsam::NonlinearFactor{
   void serialize() const;
 };
 
+#include <gtsam/navigation/PseudorangeFactor.h>
+virtual class PseudorangeFactor : gtsam::NonlinearFactor {
+  PseudorangeFactor(gtsam::Key receiver_position_key,
```
dellaert: Please use same argument names as in c++, and ensure both are camelCase. I know Python conventions are different, but if we don't make them the same, our documentation generation will fail.

gtsam/navigation/PseudorangeFactor.h
```diff
@@ -0,0 +1,115 @@
...
+   * @param receiver_clock_bias_key Receiver clock bias node.
+   * @param measured_pseudorange Receiver-measured pseudorange in meters.
+   * @param satellite_position Satellite ECEF position in meters.
+   * @param satellite_clock_bias Satellite clock bias in seconds.
+   * @param model 1-D noise model.
+   */
+  PseudorangeFactor(Key receiver_position_key, Key receiver_clock_bias_key,
+                    double measured_pseudorange,
+                    const Point3& satellite_position,
+                    double satellite_clock_bias, const SharedNoiseModel& model);
```
dellaert: Defaults for last 2 args?

gtsam/navigation/tests/testPseudorangeFactor.cpp
```diff
@@ -0,0 +1,29 @@
...
+// *************************************************************************
+TEST(TestPseudorangeFactor, Constructor) {
+  const auto factor =
+      PseudorangeFactor(Key(0), Key(1), 0.0, Point3::Zero(), 0.0,
+                        noiseModel::Isotropic::Sigma(1, 1.0));
+  const double error = factor.evaluateError(Point3::Zero(), 0.0)[0];
+  EXPECT_DOUBLES_EQUAL(0.0, error, 1e-9);
+}
+
+// *************************************************************************
```
dellaert, line 100: We absolutely need the Jacobians tested :-)

dellaert:
@masoug This is awesome !!! I made a couple of comments. We definitely need the derivatives tested. Also, GDCM uses specific naming conventions as detailed, among others, here: .github/copilot-instructions.md

I will trigger a co-pilot request and he probably will have a lot of comments about that kind of stuff. 

I especially liked your Python notebook example. I'm a bit sad we need to calculate the satellite positions by calling on a different library, but I acknowledge it's not easy to reimplement that stuff. I'm also wondering whether we could batch up range measurements at a particular time, so that if we have measurements to 12 satellites, we only create one factor with a 12x3 Jacobian and 12x1 Jacobian - If that makes sense.

33 meters seems large, but I guess it's only 4 satellites. Do you have any idea how the error decreases if you use more satellites? 

PS I was the external examiner on a tightly coupled GPS thesis, a long time ago, pre GTSAM even: [link](https://kilthub.cmu.edu/articles/thesis/Mapping_Large_Urban_Environments_with_GPS-Aided_SLAM/6720488?file=12254882). This PR is a cool first step to actually recreating that thesis in GTSAM :-).

gtsam/navigation/PseudorangeFactor.h
```diff
@@ -0,0 +1,115 @@
...
+                       OptionalMatrixType Hreceiver_clock_bias) const override;
+
+ private:
+#if GTSAM_ENABLE_BOOST_SERIALIZATION  ///
+  /// Serialization function
+  friend class boost::serialization::access;
+  template <class ARCHIVE>
+  void serialize(ARCHIVE& ar, const unsigned int /*version*/) {
+    ar& boost::serialization::make_nvp(
+        "NoiseModelFactor2", boost::serialization::base_object<Base>(*this));
```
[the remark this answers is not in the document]
dellaert: Check this in other classes. I think the 2 might be a typedef...

dellaert (commented):
By the way, we need a Python unit test as well if you add a wrapper. Otherwise there's a no guarantee that the wrapper will continue to work. After you add a Jacobian test I should be able to convert it to Python easily - although you might give some context to make sure it does so in the right style.

dellaert:
By the way, I can't stress how cool this is :-)

ProfFan:
I remember there is actually a repository on GitHub that does tightly coupled GPS with GTSAM, unfortunately I forgot where it is now...

gtsam/navigation/doc/PseudorangeFactor.ipynb
masoug: added try-catch blocks to condition %pip on google colab.

gtsam/navigation/navigation.i
masoug: Renamed!

gtsam/navigation/PseudorangeFactor.h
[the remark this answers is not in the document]
masoug: Done! `BOOST_SERIALIZATION_BASE_OBJECT_NVP(PseudorangeFactor::Base)` is very convenient.

dellaert:
Some CI failures remain, please examine.

@ProfFan A FG-based entry was one of leading contenders in Google decimeter challenge:
https://www.kaggle.com/competitions/google-smartphone-decimeter-challenge/writeups/taro-1st-place-solution

gtsam/navigation/tests/testPseudorangeFactor.cpp
dellaert: Still need this :-)

gtsam/navigation/PseudorangeFactor.h
```diff
@@ -0,0 +1,115 @@
...
+   * @param measuredPseudorange Receiver-measured pseudorange in meters.
+   * @param satellitePosition Satellite ECEF position in meters.
+   * @param satelliteClockBias Satellite clock bias in seconds.
+   * @param model 1-D noise model.
+   */
+  PseudorangeFactor(Key receiverPositionKey, Key receiverClockBiasKey,
+                    double measuredPseudorange, const Point3& satellitePosition,
+                    double satelliteClockBias,
+                    const SharedNoiseModel& model =
+                        noiseModel::Diagonal::Sigmas(Vector1(1.0)));
```
dellaert: Prefer noiseModel::Unit in this case.

gtsam/navigation/PseudorangeFactor.h
masoug: Done!

gtsam/navigation/tests/testPseudorangeFactor.cpp
masoug: Added a couple more Jacobian tests, including one for the singularity when sat position == receiver position.

gtsam/navigation/tests/testPseudorangeFactor.cpp
```diff
@@ -0,0 +1,29 @@
...
+using namespace gtsam;
+
+// *************************************************************************
+TEST(TestPseudorangeFactor, Constructor) {
+  const auto factor =
+      PseudorangeFactor(Key(0), Key(1), 0.0, Point3::Zero(), 0.0,
+                        noiseModel::Isotropic::Sigma(1, 1.0));
+  const double error = factor.evaluateError(Point3::Zero(), 0.0)[0];
+  EXPECT_DOUBLES_EQUAL(0.0, error, 1e-9);
+}
```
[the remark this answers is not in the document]
masoug: Expanded test coverage, so vast majority of lines should be covered.

gtsam/navigation/PseudorangeFactor.h
masoug: done ✅

masoug:
Thanks everyone for the comments! I'm glad folks are interested in this topic too.

> I'm a bit sad we need to calculate the satellite positions by calling on a different library, but I acknowledge it's not easy to reimplement that stuff.

Turns out, the satellite orbit algorithms are not difficult to compute, but they're also not really relevant to include within GTSAM at the moment. That said, there is potential in the future where pseudorange factors re-implement differentiable versions of orbital algorithms to refine satellite motion during optimization. But one step at a time 🦶 let's cross that bridge when we get there 🌉 

> I'm also wondering whether we could batch up range measurements at a particular time, so that if we have measurements to 12 satellites, we only create one factor with a 12x3 Jacobian and 12x1 Jacobian - If that makes sense.

That does make sense, however, I kept the factors separate to leave the door open for robust loss functions (like Huber). I'm wondering if the built-in M-estimators could function like [RAIM](https://en.wikipedia.org/wiki/Receiver_autonomous_integrity_monitoring) to reject individual faulty satellites or RF interference, improving robustness against outliers. Do batch factors support fine-grain loss functions?

> 33 meters seems large, but I guess it's only 4 satellites. Do you have any idea how the error decreases if you use more satellites?

33 feels kinda high to me too, though a large part of the error might be explained by elevation uncertainty and the lack of atmospheric corrections. That said, the next experiment (PR) I'd like to try is a primitive differential GNSS factor where I use ZOA1 as a reference to correct single-point-solution of a nearby station like [P222](https://www.ngs.noaa.gov/CORS/ncn_station_pages/index.html?stationID=P222) with a more complex factor-graph network. I think most of the error should disappear at that point, but I'm not certain yet.

re more satellites: More satellites might help if the measured pseudoranges are not all biased by the same atmospheric effects. Without correcting for those common-mode errors, however, adding more pseudorange observations likely won't improve position accuracy.

> A FG-based entry was one of leading contenders in Google decimeter challenge:

That's really cool 😎 Looks like [GTSAM is a popular choice among the GNSS community](https://chatgpt.com/s/t_69786a380194819189552c70e3c1a1d7), and I'm curious to see how far down the rabbit-hole this could go. It'd be super awesome if GTSAM can demonstrate survey-grade positioning accuracy/precision with large-scale factor graph networks fusing/estimating everything from IMU preintegration, satellite orbits, reference stations, carrier-phase measurements, doppler, SBAS, atmospheric effects, etc... Not to mention a whole other world of meteorology where GNSS factor-graph networks are used to measure the atmosphere. But I don't want to get ahead of myself--we'll see where we get through small, gradual steps 🚶

dellaert (approved):
LGTM, CI passes !

varunagrawal:
This is indeed really cool. I did a project with Sehoon Ha on GNSS based urban localization for quadrupeds using GPS+RTK back in 2024. I wish this factor existed then.

I do have a Carrier Phase factor implemented in a private repository for that project if you're interested.
These are the results I got from walking around Georgia Tech's campus:

image

PR #2373, opened 2026-01-25
EKF Reset

dellaert:
Implemented "reset" covariance transport in the EKF filter hierarchy also added a notebook showing how the NEES value is better when reset is being used. This is now the default setting for all update methods in the manifold EKF.

image


scottiyio (approved):
Interesting, never thought about resetting covariance. I don't see any issues

dellaert:
Thanks. it’s the key step in the that makes the left and right invariant filter equivalent, in that ANU/Klagenfurt paper.

dellaert:
@giuliodelama FYI

PR #2374, opened 2026-01-26
NEES Comparison

dellaert:
I exposed both versions of the pre-integrated measurements and tried to do a comparison with the EKF. This is still very much a working progress because NEES values are all over the place. So maybe we're comparing apples and oranges.


dellaert:
codex: This exposes preintegration variants and adds an experimental notebook comparing their covariance consistency against an EKF using NEES. Labeled other because it is exploratory validation and diagnostic work rather than a discrete product fix or feature.

dellaert:
Superseded by #2719

PR #2377, opened 2026-01-31
Standardize on KeySet

dellaert:
We were using `gtsam::KeySet` in some places and `std::set<Key>` in other places. This did not show up as a problem, but became a problem with new functionality in GTDynamics. This PR standardizes on `gtsam::KeySet`.


varunagrawal (approved):
LGTM

PR #2378, opened 2026-02-01
Fix mac install 

dellaert:
There was a change to this line 2 months ago that broke linking in GTDynamics. Changing this line to use @rpath fixes it.

### DIFF
--- a/cmake/HandleGlobalBuildFlags.cmake
+++ b/cmake/HandleGlobalBuildFlags.cmake
@@ -20,7 +20,7 @@ endif()
 
 if (APPLE AND GTSAM_SHARED_LIB)
     # Set the default install directory on macOS
-    set(CMAKE_INSTALL_NAME_DIR "lib")
+    set(CMAKE_INSTALL_NAME_DIR "@rpath")
 endif()
 
 ###############################################################################


varunagrawal:
That's so interesting. I've been compiling and using GTDynamics without any issue on an M1 macbook, so I'm wondering if it's something deeper?

varunagrawal (approved):

dellaert:
@varunagrawal I think it was you who changed that line. So can you check whether this works for you as well? The CI will also tell of course. 

I ran into issues with this multiple times now, myself and when helping a student set up GTDynamics. This PR seems to fix it.

varunagrawal:
Approving, but I'm legitimately concerned there is something more going on since this issue is coming up for you now.

dellaert:
Happened on my Mac, and same issue on a fresh Mac from a student. MIne is M1, student is M2

varunagrawal:
Okay here's the explanation on why this PR works and we didn't see this until now.
By setting `CMAKE_INSTALL_NAME_DIR` to `@rpath` we are forcing CMake to embed the RPATH to any executable which loads and links to gtsam. This was not happening before so unless the system and CMake variables were set properly, it would cause `Library not found` issues.

I am convinced this is the correct fix.

PR #2380, opened 2026-02-01
Fix Eigen issue in PseudorangeFactor

varunagrawal:
Compilation with GCC 14 fails due to a small issue with setting `Hreceiver_clock_bias`.

The issue is that Eigen's matrix initialization can trigger SSE-based operations on small matrices, and GCC 14's bounds checker flags this as unsafe, resulting in a warning which our CMake config treats as an error.

cc @masoug for future reference.

### DIFF
--- a/gtsam/navigation/PseudorangeFactor.cpp
+++ b/gtsam/navigation/PseudorangeFactor.cpp
@@ -70,7 +70,7 @@ Vector PseudorangeFactor::evaluateError(
   }
 
   if (Hreceiver_clock_bias) {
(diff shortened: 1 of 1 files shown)


dellaert (approved):

varunagrawal:
How did you merge that? The Windows CI was still running so it wouldn't let me merge. :/

masoug:
Thanks @varunagrawal for catching this and submitting a hotfix quickly! I’ll keep this in mind for future contributions.

dellaert:
> How did you merge that? The Windows CI was still running so it wouldn't let me merge. :/

Not required per branch protection rules

PR #2383, opened 2026-02-03
Introducing DifferentialPseudorangeFactor

masoug:
Hello again! I've experimented with applying differential corrections to the `PseudorangeFactor` and found ~98% accuracy improvement. I've attached a copy of my experiments here: [DifferentialPseudorangeExample.html](https://github.com/user-attachments/files/25033819/DifferentialPseudorangeExample.html)

Similar to before, I'm seeking feedback on:
* How to structure the PseudorangeFactors to handle commonality b/w `PseudorangeFactor` and `DifferentialPseudorangeFactor`? So far I went with simple inheritance, but if there's a more elegant way to arrange the classes, let me know!
* What are preferred conventions for notebook library modules? I placed a `gnss_utils.py` to de-duplicate a lot of repetition, so I'm wondering if that's okay for the two pseudorange examples.
  * Would `gnss_utils.py` cause issues with google colab?
* `DifferentialPseudorangeExample.ipynb`; what could be clarified, and what are some suggestions to communicate the significance of going from 30->0.4 meter accuracy? Does anyone know of good geospatial map plotting libraries that work well with Jupyter? ChatGPT suggested [folium](https://python-visualization.github.io/folium/latest/index.html)...

Anyway, I'm interested to hear what y'all think 🧠 I think I might look at carrier-phase factors next, so @varunagrawal I'd be interested in learning more about your experiences on that, if you're willing to share 🙏 

Best,
Sammy


ProfFan:
Thank you for the contributions!

> How to structure the PseudorangeFactors to handle commonality b/w PseudorangeFactor and DifferentialPseudorangeFactor?

I would recommend splitting out the common stuff as standalone utility functions instead of inheritance 

> What are preferred conventions for notebook library modules?

I think you can put things in `gtsam_unstable` but then it won't work until the next nightly build. For now a simple folded cell probably works better?

> good geospatial map plotting libraries

https://docs.kepler.gl/docs/keplergl-jupyter

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
```diff
@@ -0,0 +1,375 @@
...
+    "\n",
+    "## Overview\n",
+    "Differential GNSS (DGNSS) improves positioning accuracy by using a reference receiver at a precisely known location to correct errors affecting nearby GNSS users. The reference station receives the same satellite signals as the user and compares the measured ranges to the true ranges based on its known position. The differences represent combined errors from satellite clocks, orbits, and atmospheric delays. These corrections are transmitted to the user receiver, which applies them to its own measurements. Because both receivers experience nearly the same errors, especially when close together, many of these errors cancel out, reducing position errors from several meters to sub-meter or better accuracy.\n",
+    "\n",
+    "GNSS positioning problems can be encoded as factor graphs in GTSAM; this notebook extends the groundwork from [SinglePointPositioningExample.ipynb](https://github.com/borglab/gtsam/blob/develop/python/gtsam/examples/SinglePointPositioningExample.ipynb) to improve receiver positioning accuracy using corrections from a nearby reference receiver."
```
dellaert: Use https://borglab.github.io/gtsam/singlepointpositioningexample/

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
```diff
@@ -0,0 +1,375 @@
...
+    "error = ground_truth - receiver_position\n",
+    "print(f\"Total positioning error: {np.linalg.norm(error)} meters\")"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "30 meters is a decent starting point from single-point positioning, but let's see if we can do better.\n",
```
dellaert, line 196: On the map it doesn't look like it's 30 meters away. Is this mostly an elevation error? It might be good to also show the corrected position, which I don't think you do below. And if it's indeed an elevation error, comment on that: as I understand elevation is typically the weak point of GPS.

**If** so, it would be nice to also calculate the error in the X-Y position (and maybe even the 3x3 marginal , to show that GTSAM *knows*) to give the readers an understanding that they can be pretty good in X-Y.

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
```diff
@@ -0,0 +1,375 @@
...
+   },
+   "outputs": [],
+   "source": [
+    "try:\n",
+    "    import google.colab\n",
+    "    %pip install --quiet pyrtklib numpy gtsam-develop pyproj tabulate\n",
+    "except ImportError:\n",
+    "    pass\n",
+    "\n",
+    "import gnss_utils\n",
```
dellaert, line 52: Unfortunately, that is an issue.
We rely on the pre-computed outputs for rendering on the web, so that's not an issue.   
However, for making this work in Colab, you will need to add a `wget` (from the blob, only available after merge :-) after the pip install, I think.

dellaert (commented):
Very nice. Your factor graph illustration is also great, but maybe it would be more clear if you showed the correction and the pseudo factors for two satellites. You really have to read the text to understand that there one correction variable per satellite. 

Also wondering what happens over time and as the receiver moves.

dellaert:
Yes, I walked in them (Coyote) hills. There used to be nuclear missiles stationed there, too 🚀

dellaert:
https://en.wikipedia.org/wiki/Project_Nike

varunagrawal:
I don't think putting code in `gtsam` and the notebooks in `gtsam_unstable` is a good idea. It will make finding each part tricky.

I have used `folium` in the past and it is great! It makes really nice interactive widgets that one can play around with, and the API is really simple to get basic map and path rendering.

dellaert:
Agreed. @masoug, please put that module in the examples folder, next to the notebooks, and do the wget on colab.

masoug:
Thanks for the feedback! I'll get these comments addressed in the next couple days 👍

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
masoug: done ✅

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
masoug: added wget command for colab ✅

masoug:
> Very nice. Your factor graph illustration is also great, but maybe it would be more clear if you showed the correction and the pseudo factors for two satellites. You really have to read the text to understand that there one correction variable per satellite.

I've updated the illustration by adding an image (I couldn't get mermaid render a decent graph) showing two satellites and their pseudorange factors. Attached is an updated notebook output: [DifferentialPseudorangeExample.html](https://github.com/user-attachments/files/25120483/DifferentialPseudorangeExample.html)
 
> Also wondering what happens over time and as the receiver moves.

This is a great question! Spoiler alert: Position quality degrades over long times and great distances between the user and reference stations. I wanted to run an experiment where I plot accuracy as a function of distance-from-reference-station for a variety of test station across CORS, but I'd put that in a separate notebook not designed for documentation/tutorial.

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
masoug: At first, I thought the same--elevation likely explained most of the error. But the data shows something different! Here's the single-point solution coordinates:
```
Latitude: 37.53924499933658 degrees
Longitude: -122.08292615446172 degrees
Altitude: 56.93383957352489 meters
```
and here are the differentially-corrected results:
```
Corrected geodetic coordinates: 37.5392409437283, -122.08326490703365, 53.18318437319249 m
```
Less than 4m of vertical difference! P222's surveyed height (above ellipsoid) is 53.512m. So most of the 30m error was indeed horizontal.

I'll see if there are ways to graphically visualize the errors in the meantime...

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
```diff
@@ -0,0 +1,400 @@
...
+    "```\n",
+    "| ITRF2020 POSITION (EPOCH 2020.0)                                            |\n",
+    "| Computed in Apr 2025 using data through gpswk 2237.                         |\n",
+    "|     X =  -2684436.824 m     latitude    =  37 32 34.99617 N                 |\n",
+    "|     Y =  -4293336.957 m     longitude   = 122 00 57.42005 W                 |\n",
+    "|     Z =   3865351.638 m     ellipsoid height =   -3.960   m                 |\n",
+    "```\n",
+    "In a factor graph framework, these survey coordinates correspond to a stiff `PriorFactor` constraint on ZOA1's position node to keep it anchored in the global Earth frame. The rest of the factor graph is constructed as follows:\n",
+    "\n",
+    "![illustrative factor graph example showcasing differential correction factors](https://raw.githubusercontent.com/masoug/gtsam-gnss/refs/heads/feature/differential-pseudorange/python/gtsam/examples/images/dpr_fg.jpg)\n",
```
masoug: Note to self:
- [x] Update this URL to borglab-develop before merging!

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
dellaert: Gotcha!

dellaert (approved):
Awesome ! cool new graphic :-)

python/gtsam/examples/DifferentialPseudorangeExample.ipynb
dellaert: So, does this still need to happen or can I merge? If not, please add [skip ci] to your commit message so we don't have to run CI again.

masoug: Yes, let me make a quick edit to fix that, and then it can merge.

PR #2384, opened 2026-02-03
Fix/covariances

JzHuai0108:
1. straighten the ManifoldPreintegration cov computation, making it consistent with the Jacobians;
2. unit test covariance of PreintegratedImuMeasurements and PreintegratedCombinedMeasurements derived from ManifoldPreintegration or TangentPreintegration by comparing with ImuEKF;
3. unit test that preintegrated measurements derived from ManifoldPreintegration and TangentPreintegration have almost identical cov as the residual is defined the same;
4. largely fix the NEES computation.

### DIFF
--- a/gtsam/navigation/CombinedImuFactor.h
+++ b/gtsam/navigation/CombinedImuFactor.h
@@ -193,25 +193,6 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurementsT : public PreintegrationTyp
 // For backward compatibility:
 using PreintegratedCombinedMeasurements = PreintegratedCombinedMeasurementsT<DefaultPreintegrationType>;
 
-// Manifold combined preintegration: convert preintMeasCov() (defined in the
-// “transported” coordinate convention) into the residual covariance convention
-// used by the factor, by rotating the dp/dv blocks with dR.
-inline Eigen::Matrix<double, 15, 15> combinedImuFactorResidualCov(
-    const gtsam::PreintegratedCombinedMeasurementsT<gtsam::ManifoldPreintegration>& pim) {
-  const Eigen::Matrix<double, 15, 15> Sigma = pim.preintMeasCov();
-
-  const gtsam::Matrix3 R = pim.deltaRij().matrix();
-
-  Eigen::Matrix<double, 15, 15> J = Eigen::Matrix<double, 15, 15>::Identity();
-  J.block<3,3>(3,3) = R;  // dp
-  J.block<3,3>(6,6) = R;  // dv
-
-  Eigen::Matrix<double, 15, 15> Sigma_out;
-  Sigma_out.noalias() = J * Sigma * J.transpose();
-  return Sigma_out;
-}
-
-// Fallback for other Combined PIM types.
 template <class PIM>
 inline Eigen::Matrix<double, 15, 15> combinedImuFactorResidualCov(const PIM& pim) {
   return pim.preintMeasCov();
--- a/gtsam/navigation/CombinedImuFactor.h
+++ b/gtsam/navigation/CombinedImuFactor.h
@@ -193,6 +193,30 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurementsT : public PreintegrationTyp
 // For backward compatibility:
 using PreintegratedCombinedMeasurements = PreintegratedCombinedMeasurementsT<DefaultPreintegrationType>;
 
+// Manifold combined preintegration: convert preintMeasCov() (defined in the
+// “transported” coordinate convention) into the residual covariance convention
+// used by the factor, by rotating the dp/dv blocks with dR.
+inline Eigen::Matrix<double, 15, 15> combinedImuFactorResidualCov(
+    const gtsam::PreintegratedCombinedMeasurementsT<gtsam::ManifoldPreintegration>& pim) {
+  const Eigen::Matrix<double, 15, 15> Sigma = pim.preintMeasCov();
+
+  const gtsam::Matrix3 R = pim.deltaRij().matrix();
+
+  Eigen::Matrix<double, 15, 15> J = Eigen::Matrix<double, 15, 15>::Identity();
+  J.block<3,3>(3,3) = R;  // dp
+  J.block<3,3>(6,6) = R;  // dv
+
+  Eigen::Matrix<double, 15, 15> Sigma_out;
+  Sigma_out.noalias() = J * Sigma * J.transpose();
+  return Sigma_out;
+}
+
+// Fallback for other Combined PIM types.
+template <class PIM>
+inline Eigen::Matrix<double, 15, 15> combinedImuFactorResidualCov(const PIM& pim) {
+  return pim.preintMeasCov();
+}
+
 /**
  * CombinedImuFactor is a 6-ways factor involving previous state (pose and
  * velocity of the vehicle, as well as bias at previous time step), and current
@@ -247,7 +271,7 @@ class GTSAM_EXPORT CombinedImuFactorT
   CombinedImuFactorT(
       Key pose_i, Key vel_i, Key pose_j, Key vel_j, Key bias_i, Key bias_j,
       const PIM& preintegratedMeasurements)
-      : Base(noiseModel::Gaussian::Covariance(preintegratedMeasurements.preintMeasCov()),
+      : Base(noiseModel::Gaussian::Covariance(combinedImuFactorResidualCov(preintegratedMeasurements)),
              pose_i, vel_i, pose_j, vel_j, bias_i, bias_j),
         pim_(preintegratedMeasurements) {}
 
--- a/gtsam/navigation/ImuFactor.h
+++ b/gtsam/navigation/ImuFactor.h
@@ -173,6 +173,19 @@ public:
 // controls which class PreintegratedImuMeasurements uses):
 using PreintegratedImuMeasurements = PreintegratedImuMeasurementsT<DefaultPreintegrationType>;
 
+// For TangentPreintegration, the perturbation on deltaXij is dXt, the 3D perturbation of the rotation tangent space + 
+// 3D perturbation of the translation part + 3D perturbation of the velocity part,
+// as validated by the matrix A_k in the ImuFactor.pdf and the unit test on preintegrated_H_biasAcc()
+
+// For ManifoldPreintegration, for covariance propagation, the perturbation on deltaXij is dXn, 
+// the right perturbation of the NavState manifold, as validated by the unit tests on NavState::retract);
+// for jacobian computation, the perturbation on deltaXij is dXt defined above, as validated by the unit test on delPdelBiasAcc().
+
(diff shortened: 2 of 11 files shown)


gtsam/navigation/CombinedImuFactor.h
```diff
@@ -193,6 +193,11 @@ class GTSAM_EXPORT PreintegratedCombinedMeasurementsT : public PreintegrationTyp
 // For backward compatibility:
 using PreintegratedCombinedMeasurements = PreintegratedCombinedMeasurementsT<DefaultPreintegrationType>;
 
+template <class PIM>
```
dellaert, line 196: purpose? If the issue is the return type, it's better to change that in the pim, returing a const ref to preintMeasCov_.  This here creates *two* mallocs on every call :-)

gtsam/navigation/ImuFactor.h
```diff
@@ -173,6 +173,24 @@ class GTSAM_EXPORT PreintegratedImuMeasurementsT: public PreintegrationType {
...
+// the right perturbation of the NavState manifold, as validated by the unit tests on NavState::retract);
+// for jacobian computation, the perturbation on deltaXij is dXt defined above, as validated by the unit test on delPdelBiasAcc().
+
+// For ImuFactor or CombinedImuFactor, the residual is defined by PreintegrationBase::computeError and 
+// its perturbation is coincident to dXt (up to a right Jacobian of a small angle ~ identity).
+// So for TangentPreintegration, the covariance of the residual equals to the preintMeasCov of TangentPreintegration.
+// But for ManifoldPreintegration, we need to rotate the position and velocity blocks of the preintMeasCov to get the residual cov.
+// Otherwise, we can update the cov propagation of manifold preintegration to make it use dXt.
+template <class PIM>
+inline gtsam::Matrix9 imuFactorResidualCov(const PIM& pim) {
```
dellaert, line 189: Similar comment

gtsam/navigation/ImuFactor.h
```diff
@@ -173,6 +173,24 @@ class GTSAM_EXPORT PreintegratedImuMeasurementsT: public PreintegrationType {
 // controls which class PreintegratedImuMeasurements uses):
 using PreintegratedImuMeasurements = PreintegratedImuMeasurementsT<DefaultPreintegrationType>;
 
+// For TangentPreintegration, the perturbation on deltaXij is dXt, the 3D perturbation of the rotation tangent space +
```
dellaert: These comments might be better in the respective pre-integration classes. Also, it's hard to understand exactly what is meant. 

Maybe to clarify, I use 
- left perturbation as exp(xi)*T (which is, ironically, right-invariant)
- right perturbation as T*exp(xi) (which is left-invariant)

gtsam/navigation/ManifoldPreintegration.cpp
```diff
@@ -56,6 +56,120 @@ bool ManifoldPreintegration::equals(const ManifoldPreintegration& other,
       && equal_with_abs_tol(delVdelBiasOmega_, other.delVdelBiasOmega_, tol);
 }
 
+NavState ManifoldPreintegration::UpdatePreintegrated(
```
dellaert, line 59: Ah, so you're saying the manifold pre-integration was wrong? Not the tangent pre-integration?

And I am right in interpreting that you see the manifold pre-integration as using the *right-invariant* error instead of the left-invariant?

Please forgive my questions: I have relatively little time today to take the time I should *really* take to figure it out, so hoping the dialogue will make things easier.

gtsam/navigation/ManifoldPreintegration.cpp
```diff
@@ -78,7 +192,7 @@ void ManifoldPreintegration::update(const Vector3& measuredAcc,
 
   // Do update
   deltaTij_ += dt;
-  deltaXij_ = deltaXij_.update(acc, omega, dt, A, B, C); // functional
+  deltaXij_ = UpdatePreintegrated(acc, omega, dt, deltaXij_, A, B, C);
```
dellaert, line 165: If we stick to a right-invariant error, if I take this away from the IMU pre-integration context for a bit, could we fix it by just having a *leftUpdate in nav state?

dellaert (commented):
Many thanks for working with me on this. This is very important to me.

gtsam/navigation/CombinedImuFactor.h
JzHuai0108: Yes, indeed. This function was left here for two reasons. a. previously I used a specialization of this template function for PreintegratedCombinedMeasurements to correct its covariance. That is, the specialization was more than the plain one line return. b. This function allows supporting alternative definitions of preintegration residuals. Right now, the first 9dim of the preint factors are the same and defined such that the Jacobian of the IMU residual relative to the preint measurement is (almost) identity. For now, I will delete this func, and restore to directly using the preintMeasCov() in places calling this func.

gtsam/navigation/ImuFactor.h
JzHuai0108: Yes. Will move them to their respective class headers. I am comfortable with left perturbation exp(xi)*T notations. These comments are not easy to understand since they refer to evidence in other files. I put them here mainly for your reference, and will simplify them as I move them around.

gtsam/navigation/ImuFactor.h
JzHuai0108: Will move the comments to respective class headers.

gtsam/navigation/ManifoldPreintegration.cpp
JzHuai0108: I think the manifold preintegration's covariance/Jacobian computation is problematic, but the manifold preintegration's value is OK. The original code of manifold preintegration computes the preint measurement covariance using the error definition: dXn = [\delta \phi_ij, \delta p_ij, \delta v_ij] where R_ij = \hat{R}_ij Exp(\delta \phi_ij)
  p_ij = \hat{p}_ij + R_ij \delta p_{ij} and v_ij = \hat{v}_ij + R_ij \delta v_{ij}; but computes its bias Jacobians using the error definition: 
dXt = [\delta \phi_ij, \delta p_ij, \delta v_ij] where R_ij = \hat{R}_ij Exp(\delta \phi_ij) p_ij = \hat{p}_ij + \delta p_{ij} and v_ij = \hat{v}_ij + \delta v_{ij}. Therefore, there is an inconsistency between covariance and Jacobian of the preint measurement.

To me, the tangent preintegration is OK in value and Jacobians.
I think all error states in gtsam are left-invariant based on my observations.

gtsam/navigation/ManifoldPreintegration.cpp
JzHuai0108: I think we'd better stay with left-invariant errors, i.e., right perturbation error. This function UpdatePreintegrated just changes the Jacobian computation for the preint measurement, with still right perturbation on rotation, but different perturbations on translation and velocity.

gtsam/navigation/ManifoldPreintegration.cpp
JzHuai0108: In writing the extended abstract, I found some inconsistency in my derivations :(. I will double check it in the coming days.

dellaert:
OK, I went through this more carefully. Not fully reviewed yet, but

- The key point should be stated more explicitly in the PR description: “old manifold preintMeasCov lived in the moving $T_{\Delta}$ chart; new one lives in the increment chart consumed by biasCorrectedDelta/predict.” do you agree with that assessement?
- UpdatePreintegratedSlow is extremely useful for testing, but I would probably not leave it as public API?
- Would it be possible to add one regression test with non-identity body_P_sensor? That post-multiplication of B,C after the chart transform is an easy place for subtle mistakes.

JzHuai0108:
1. I am quite rusty about chart. :) I will put the change to ManifoldPreintegration in plain calculus terms. 

 Let's denote the right perturbation dXn on the NavState X_ij's manifold, i.e.,
 dXn = [\delta \phi_ij, \delta p_ij, \delta v_ij] where R_ij = \hat{R}_ij Exp(\delta \phi_ij)
 p_ij = \hat{p}_ij + R_ij \delta p_{ij} and v_ij = \hat{v}_ij + R_ij \delta v_{ij}.

For ManifoldPreintegration, for covariance propagation, **the perturbation on deltaXij(preint X_ij) is dXn**, 
the right perturbation of the NavState manifold, as validated by the unit tests on NavState::retract);
but for jacobian computation, the perturbation on deltaXij (ie, preint X_ij) was actually defined as
 dXt = [\delta \phi_ij, \delta p_ij, \delta v_ij] where R_ij = \hat{R}_ij Exp(\delta \phi_ij)
 p_ij = \hat{p}_ij + \delta p_{ij} and v_ij = \hat{v}_ij + \delta v_{ij},
as validated by the unit test on delPdelBiasAcc().

For ImuFactor or CombinedImuFactor, the residual is defined by PreintegrationBase::computeError and 
its perturbation is coincident to dXt (up to a right Jacobian of a small angle ~ identity).

For TangentPreintegration, the perturbation on deltaXij is dXt, the 3D perturbation of the rotation tangent space + 
3D perturbation of the translation part + 3D perturbation of the velocity part,
as validated by the matrix A_k in the ImuFactor.pdf and the unit test on preintegrated_H_biasAcc()

So for TangentPreintegration, the covariance of the residual equals to the preintMeasCov of TangentPreintegration.
But for ManifoldPreintegration, we need to rotate the position and velocity blocks of the preintMeasCov to get the residual cov.
Otherwise, we can update the cov propagation of manifold preintegration to make it use dXt. This is what we do in the pull request (see ManifoldPreintegration::update()) to keep the API change minimal.
If you are interested, expand the commits of the pull request, you will notice the former approach, but it has been overridden by the latter approach.

Overall, I think I agree with the assessment based on my understanding of charts.

2. It has been moved to the test file since it was static.
3. A test on Jacobians B and C is added with non-identity body_P_sensor. Note Jacobian A is tested by the covariance comparison with EKF (see integrateMeasurement() and RunCombinedVsEkf9D()), and the Jacobian comparison to the UpdatePreintegratedSlow.

dellaert (approved):
Many thanks for your contribution! @scottiyio and I will do one last NEES check on another branch and then merge.

JzHuai0108:
I just added the last key piece. Now the fix is complete. Also retouch the comments for clarity.

dellaert:
codex: This corrects manifold-preintegration covariance propagation and NEES computation, with cross-checks against the IMU EKF and tangent preintegration. Labeled bug because it repairs inconsistent covariance calculations.

dellaert:
@JzHuai0108 I apologize for the very long time before I had the courage to tackle this :-) On looking into this in more detail, and asking the AI to create some failing tests, including Monte Carlo, I opted for an approach that does not change the definition of the residual. All this needs then in the IMU factor is to convert the Jacobian that's being kept by the pin to this tangent space around the residual. This is all implemented in #2719, where I will also tag you. Please take a look.

dellaert (changes requested):
Possibly superseded by #2719 - so don't merge.

PR #2393, opened 2026-02-07
Fix a deadlock under Windows

### LINKED ISSUE #2391: Scheduler waitForAllTasks can hang randomly
## Description

Investigating the random hangs in Windows Debug CI led me to finding that `Scheduler`'s `waitForAllTasks` hangs in the EnqueueFireAndForget test, specifically this line:
https://github.com/borglab/gtsam/blob/4a4b06c646e0b978f6180bb6a09b5197c350b7dd/gtsam/base/Scheduler.h#L390.

## Steps to reproduce

1. Configure CMake in the same way as the Debug build in build-windows.yml
2. Build it
3. Run check_base_program.exe over and over again.
4. Observe a hang. Use WinDbg to attach to the process, switch to the main thread, and examine the stack to see that `condition_.wait` is the source of the hang.

## Expected behavior

The test should not hang.

## Environment

Windows 11 Home, MSVC version 19.44.35213, Visual Studio 2022 17.14.10.

ProfFan:
Fixes #2391 

Reason is the `condition_variable` is not guarded by mutex acquisition.

### DIFF
--- a/gtsam/base/Scheduler.h
+++ b/gtsam/base/Scheduler.h
@@ -122,6 +122,7 @@ class Scheduler {
 
       // Notify waiters only when work transitions to "done".
       if (activeTasks_.fetch_sub(1, std::memory_order_release) == 1) {
+        std::lock_guard<std::mutex> lock(waitMutex_);
         condition_.notify_all();
       }
     }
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Thanks!!!

PR #2397, opened 2026-02-08
Update SFMBAL benchmark script

ProfFan:
1. Use BAL-135 dataset
2. Add TBB ON/OFF to the benchmark matrix


ProfFan:
Note that the bench can only start to show results when this is merged, because workflow dispatch limitations.

dellaert:
Cool!! Did you finetune so it does not run on every push to develop?

ProfFan:
@dellaert I checked the logs and I think it never ran on push to `develop`

dellaert (approved):
Cool!

PR #2399, opened 2026-02-09
Clean up Boost dependencies

Gold856:
Split from #2388. Replaces Boost.Filesystem with `std::filesystem`, available since C++17 and [widely available across compilers](https://en.cppreference.com/w/cpp/compiler_support/17.html). Removes checks for Boost.Thread, Boost.Regex, and Boost.DateTime since none of those are used, and adds checks for Boost.Graph and Boost.Random. Also removes an unused Boost.Lambda include.

### DIFF
--- a/cmake/HandleBoost.cmake
+++ b/cmake/HandleBoost.cmake
@@ -39,11 +39,10 @@ endif()
 ################################################################################
 
 # Set minimum required Boost version and components.
-# NOTE: "system" is intentionally omitted. It is a transitive dependency of other
-# components (like filesystem) and will be found automatically. Explicitly
-# requesting it can cause issues with modern header-only versions of Boost.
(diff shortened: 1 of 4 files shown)


talregev:
Looks good!

talregev:
You already commit to gtsam. why the CI is needed to be approved?

Gold856:
That is an excellent question. Maybe if I rebase and push again, it'll work.

dellaert (approved):

dellaert:
CI failures

Gold856:
Seemed unrelated, but let's run the full gamut now that it's updated.

dellaert:
In which order should I merge them or does it matter?

Gold856:
They're independent, so it doesn't matter.

dellaert:
Merged. This is a great!

PR #2402, opened 2026-02-09
removed unnecessary DiscreteKeys constructor

aminoh:
Since the constructors for `std::vector<DiscreteKey>::vector are all forwarded already, this line was unnecessary.


dellaert:
Thanks !!!

PR #2403, opened 2026-02-09
Fix wrap w/o boost

dellaert:
Building the python wrapper without boost installed failed even when we turned off boost during CMake. The reason is that the templates still include an unguarded boost header. I fixed that and also added a path to the CI to exercise this combination.

### DIFF
--- a/.github/workflows/build-python.yml
+++ b/.github/workflows/build-python.yml
@@ -61,6 +61,7 @@ jobs:
         name:
           [
             ubuntu-22.04-gcc-11,
+            ubuntu-22.04-gcc-11-noboost,
             ubuntu-22.04-clang-11,
             macos-14-xcode-15.4,
             macos-15-xcode-16,
@@ -71,6 +72,11 @@ jobs:
         python_version: [3]
         include:
           - name: ubuntu-22.04-gcc-11
+            os: ubuntu-22.04
+            compiler: gcc
+            version: "11"
+
+          - name: ubuntu-22.04-gcc-11-noboost
             os: ubuntu-22.04
             compiler: gcc
             version: "11"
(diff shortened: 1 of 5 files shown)


dellaert:
PS we need this to work ASAP for a class, so I'll merge even w/o review if CI passes.

.github/workflows/build-python.yml
```diff
@@ -113,7 +119,11 @@ jobs:
           fi
 
           sudo apt-get -y update
-          sudo apt-get -y install cmake build-essential pkg-config libpython3-dev python3-numpy libboost-all-dev ninja-build
+          if [ "${{ matrix.name }}" = "ubuntu-22.04-gcc-11-noboost" ]; then
+            sudo apt-get -y install cmake build-essential pkg-config libpython3-dev python3-numpy ninja-build
```
varunagrawal, line 123: You could pull the common libraries out and just conditionally install boost.

.github/workflows/build-python.yml
```diff
@@ -220,10 +230,24 @@ jobs:
 
       - name: Build
         shell: bash
+        if: matrix.name != 'ubuntu-22.04-gcc-11-noboost'
         run: |
           bash .github/scripts/python.sh -b
 
+      - name: Build (No Boost Wrapper Check)
+        if: matrix.name == 'ubuntu-22.04-gcc-11-noboost'
```
varunagrawal, line 238: This can be massively simplified by defining a variable called boost in the build matrix.

varunagrawal (approved):
It's fine but not clean.

.github/workflows/build-python.yml
```diff
@@ -113,7 +119,11 @@ jobs:
           fi
 
           sudo apt-get -y update
-          sudo apt-get -y install cmake build-essential pkg-config libpython3-dev python3-numpy libboost-all-dev ninja-build
+          if [ "${{ matrix.name }}" = "ubuntu-22.04-gcc-11-noboost" ]; then
+            sudo apt-get -y install cmake build-essential pkg-config libpython3-dev python3-numpy ninja-build
+          else
+            sudo apt-get -y install cmake build-essential pkg-config libpython3-dev python3-numpy libboost-all-dev ninja-build
+          fi
```
Gold856, line 126: To match #2400:
```suggestion
          if [ "${{ matrix.name }}" = "ubuntu-22.04-gcc-11-noboost" ]; then
            sudo apt-get -y install build-essential pkg-config libpython3-dev python3-numpy
          else
            sudo apt-get -y install build-essential pkg-config libpython3-dev python3-numpy libboost-all-dev
          fi
```

Gold856 (changes requested):

dellaert:
Guys, did I tell you I need this fast? Which of the suggested changes are actually essential :-) ?

dellaert:
If the CI passes, we can always do another PR to clean it up. We need it in a workflow for class.

Gold856:
Fair enough. Skip my review and I'll roll it into #2400.

PR #2406, opened 2026-02-12
Remove macOS 14 from CI

Gold856:
EOL by the end of this year: see https://github.com/actions/runner-images/issues/13518. Removing these jobs also helps to unjam CI, and we have coverage for newer Mac compilers in other places.

This will require the required macOS 14 job to be updated to macOS 15 and Xcode 16; sorry for the two required job changes in a row!🥲

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -31,39 +31,33 @@ jobs:
         # Github Actions requires a single row to be added to the build matrix.
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
         name: [
-          macos-14-xcode-15.4,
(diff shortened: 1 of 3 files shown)


varunagrawal (approved):
Awesome, thank you!

dellaert:
LGTM ! @ProfFan might want to check out the benchmark-related failures. I'll assume they're orthogonal to this Pr and merge.

Gold856:
Yeah, I don't have the perms to do things like dispatching a workflow.

varunagrawal:
That's correct. Since @Gold856 is PRing from his fork, they don't have the same resource permissions.

Issue #2405, opened 2026-02-11
ISAM2::update sometimes fails when using the dogleg optimizer and adding variables


william-swarmbotics:
## Description

When calling `ISAM2::update` with new variables in the `newTheta` argument, it sometimes throws `std::invalid_argument` with the following message: `VectorValues::dot called with a VectorValues of different structure`.

This happens in `DoglegOptimizerImpl::ComputeBlend`, where the `x_n` parameter has the newly added variables and `x_u` does not. The contents of `x_u` are based on `ISAM2::gradientAtZero()`, which gets variables from the root cliques. It appears that `ISAM2::addVariables` does not add the new variables to the cliques, but it does add them to `deltaNewton_` (aka `x_n`), which is why there is a mismatch.

There are two other conditions for the error to occur, which are responsible for making it sporadic. `UpdateImpl::relinarizationNeeded` must be true in `ISAM2::update`, and the dogleg optimizer must decide that a blend between the gradient descent point and the Newton's method point is necessary.

Someone on a Google Groups thread also had this issue: https://groups.google.com/g/gtsam-users/c/BXkKCTW9nHg.

## Steps to reproduce

Rather than try to force a blend step, adding the following check makes the error visible by failing even if `ComputeBlend` isn't called:
```diff
diff --git a/gtsam/nonlinear/DoglegOptimizerImpl.cpp b/gtsam/nonlinear/DoglegOptimizerImpl.cpp
index e82a513ca..fb9fea5fd 100644
--- a/gtsam/nonlinear/DoglegOptimizerImpl.cpp
+++ b/gtsam/nonlinear/DoglegOptimizerImpl.cpp
@@ -26,6 +26,10 @@ namespace gtsam {
 VectorValues DoglegOptimizerImpl::ComputeDoglegPoint(
     double delta, const VectorValues& dx_u, const VectorValues& dx_n, const bool verbose) {
 
+  if (dx_u.size() != dx_n.size()) {
+    throw std::runtime_error("Size mismatch: ComputeBlend would fail");
+  }
+
   // Get magnitude of each update and find out which segment delta falls in
   assert(delta >= 0.0);
   double deltaSq = delta*delta;
```

The above patch causes `testDoglegOptimizer` to fail.

## Expected behavior

`ISAM2::update` does not throw.

## Environment

Ubuntu, C++

jashshah999:
[...]

PR #2408, opened 2026-02-14
iSAM2 Updates to Support riSAM Implementation

DanMcGann:
This is the first of 3 PRs that support the integration of [riSAM](https://arxiv.org/abs/2209.14359) into GTSAM. This PR is focused on updating iSAM2 with some additional functionality to support riSAM's implementation, and also fixes issue #2405 as the bug significantly affected the new DoglegLineSearch optimizer. Below are details on the changes:

## Feature 1 - One Step Look Ahead
riSAM relies on predicting the effect of an update on the bayes tree structure to identify which factors to re-convexify. To support this this PR adds `predictUpdateInfo` to the iSAM2 interface and a helper function `traverseTop` to the BayesTree interface. Tests can be found in `testGaussianISAM2` and `testSymbolicBayesTree`

## Feature 2 - DoglegLineSearch
In riSAM, factors are routinely re-convexified, causing the region of trust (a-la Dogleg) to be uncorrelated between time-steps. At the same time we require a trust region optimizer to ensure wild steps are not taken. DoglegLineSearch functions as a line-search along the dog-leg arc. The search is performed independently for each optimziation step supporting cases like riSAM where the underlying cost function is changing significantly between optimization steps. While this was motivated by riSAM, I think there are additional scenarios where this could be useful (e.g. using smart-factors to significantly change the cost function on updates).

While testing this feature I ran across the same bug identified in #2405 as DoglegLineSearch runs compute blend often. The cause is just as identified by @william-swarmbotics, adding variables in `ISAM2::update` before the bayes tree is actually updated causes a dimensionality mis-match between the the gradient and newton directions. This was fixed in `updateDelta` by ensuring the vectors match structure.

## Feature 3 - Skip redundant `updateDelta`
In `iSAM2::update`, `delta` is updated, but the check to identify when updates are needed ignores the `deltaReplaceMask` this causes redundant `updateDelta` calls especially in the case that `relinearizationSkip = 1` (which is used for riSAM). To avoid this redundent work I added a check for the `deltaReplaceMask`. Though It may be better to move this check into `relinarizationNeeded`, let me know if there is a preference on the location of this check.

## Final Notes
All functionality is tested, and I believe tests have full coverage. Notably I also added a test that is able to deterministically recreate the error in #2405 in `TEST(DogLegOptimizer, ComputeBlend)`. Additionally, the python interface file is updated (though not tested locally due to a cmake version issue. I am trusting the CI to catch any issues there)

### DIFF
--- a/gtsam/inference/BayesTree.h
+++ b/gtsam/inference/BayesTree.h
@@ -246,15 +246,16 @@ namespace gtsam {
     void addFactorsToGraph(FactorGraph<FactorType>* graph) const;
 
     /**
-     * @brief Returns the set of keys in a "contaminated" part of the tree
-     * @param keys: The keys defining a "contaminated" part of the tree
-     * @returns The keys affected by the factor graph computed by removeTop
-     * @note Allows one step look-ahead for what removeTop will modify
+     * @brief Returns the set of keys from the tree that are affected by a
+     * update to 'keys'
+     * @param keys: The keys updated and in-turn affecting the tree
+     * @returns The set of keys from the tree that are affected
+     * @note Note: Return matches contents of BayesNet from removeTop without
+     * affecting the tree
      */
-    gtsam::KeySet traverseTop(const gtsam::KeyVector& keys) const;
-  
+    gtsam::KeySet collectAffectedKeys(const gtsam::KeyVector& keys) const;
 
-  protected:
+   protected:
 
     /** private helper method for saving the Tree to a text file in GraphViz format */
     void dot(std::ostream &s, sharedClique clique, const KeyFormatter& keyFormatter,
@@ -269,12 +270,10 @@ namespace gtsam {
     /** Fill the nodes index for a subtree */
     void fillNodesIndex(const sharedClique& subtree);
 
-    /// @brief Traversal helper for a path used by traverseTop
-    void traversePath(gtsam::KeySet& traversedKeys,
-                      const sharedClique& clique) const;
-    /// @brief Traversal helper for clique used by traverseTop
-    void traverseClique(gtsam::KeySet& traversedKeys,
-                        const sharedClique& clique) const;
+    /// @brief Helper for collectAffectedKeys that recursively aggregates
+    /// affected keys from a path from 'clique' to the root of tree
+    void collectAffectedPathKeys(gtsam::KeySet& traversedKeys,
+                                 const sharedClique& clique) const;
 
     // Friend JunctionTree because it directly fills roots and nodes index.
     template<class BAYESTREE, class GRAPH> friend class EliminatableClusterTree;
--- a/gtsam/inference/BayesTree.h
+++ b/gtsam/inference/BayesTree.h
@@ -245,6 +245,15 @@ namespace gtsam {
     /** Add all cliques in this BayesTree to the specified factor graph */
     void addFactorsToGraph(FactorGraph<FactorType>* graph) const;
 
+    /**
+     * @brief Returns the set of keys in a "contaminated" part of the tree
+     * @param keys: The keys defining a "contaminated" part of the tree
+     * @returns The keys affected by the factor graph computed by removeTop
+     * @note Allows one step look-ahead for what removeTop will modify
+     */
+    gtsam::KeySet traverseTop(const gtsam::KeyVector& keys) const;
+  
+
   protected:
 
     /** private helper method for saving the Tree to a text file in GraphViz format */
@@ -259,6 +268,13 @@ namespace gtsam {
 
     /** Fill the nodes index for a subtree */
     void fillNodesIndex(const sharedClique& subtree);
+
+    /// @brief Traversal helper for a path used by traverseTop
+    void traversePath(gtsam::KeySet& traversedKeys,
+                      const sharedClique& clique) const;
+    /// @brief Traversal helper for clique used by traverseTop
+    void traverseClique(gtsam::KeySet& traversedKeys,
+                        const sharedClique& clique) const;
 
     // Friend JunctionTree because it directly fills roots and nodes index.
     template<class BAYESTREE, class GRAPH> friend class EliminatableClusterTree;
--- a/gtsam/nonlinear/DoglegOptimizerImpl.h
+++ b/gtsam/nonlinear/DoglegOptimizerImpl.h
@@ -258,14 +258,22 @@ typename DoglegOptimizerImpl::IterationResult DoglegOptimizerImpl::Iterate(
  * trust region is not correlated between algorithm iterations.
  */
 struct GTSAM_EXPORT DoglegLineSearchImpl {
+  struct Params {
+    double minDelta;  ///< Minimum allowed small delta
+    double maxDelta;  ///< Maximum allowed delta
+    double stepSize;  ///< Increase of trust region for outward steps
+    double sufficientDecreaseCoeff;  ///< Coefficient for suff. decrease check
+    bool verbose;                    ///< Whether to print debug information
+  };
+
   /**
    * Compute the update point for one iteration of the Dogleg Line Search
-   * algorithm, starting with a trust region of |N| * min_delta the algorithm
-   * searches trust regions from |N| * min_delta to max_delta where |N| is the
+   * algorithm, starting with a trust region of |N| * minDelta the algorithm
+   * searches trust regions from |N| * minDelta to maxDelta where |N| is the
    * number of variables in the system. The algorithm returns the search point
    * with minimum cost that meets the Wolfe Conditions. Evaluation points for
    * the line search are computed according to a geometric series step_{k+1} =
-   * step_size * step_k.
+   * stepSize * step_k.
    *
    *
    * @tparam M The type of the Bayes' net or tree, currently
@@ -274,12 +282,7 @@ struct GTSAM_EXPORT DoglegLineSearchImpl {
    * @tparam F For normal usage this will be NonlinearFactorGraph<VALUES>.
    * @tparam VALUES The Values or TupleValues to pass to F::error() to evaluate
    * the error function.
-   * @param min_delta The minimum trust region size (scaled by size of system)
-   * @param max_delta The maximum trust region size (scaled by size of system)
-   * @param step_size The multiplicative factor for the geometric series that
-   * defines the evaluation points for the line search
-   * @param sufficient_decrease_coeff The coefficient for sufficient decrease to
-   * ensure that the linesearch meets the Wolfe conditions
+   * @param params The parameters for dogleg line search
    * @param Rd The Bayes' net or tree as described above.
    * @param f The original nonlinear factor graph with which to evaluate the
    * accuracy of \f$ M(\delta x) \f$ to adjust \f$ \delta \f$.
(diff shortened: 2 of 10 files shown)


dellaert:
See some comments from co-pilot above.
CI fails with an error related to one of them - I think.
```
/home/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicBayesTree.cpp
In file included from /home/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicBayesTree.cpp:23:
/home/runner/work/gtsam/gtsam/gtsam/inference/BayesTree-inst.h:593:64: error: ISO C++ specifies that qualified reference to 'shared_ptr' is a constructor name rather than a type in this context, despite preceding 'typename' keyword [-Werror,-Winjected-class-name]
      this->traversePath(traversedKeys, typename sharedClique::shared_ptr(
                                                               ^
/home/runner/work/gtsam/gtsam/gtsam/symbolic/SymbolicBayesTree.cpp:30:18: note: in instantiation of member function 'gtsam::BayesTree<gtsam::SymbolicBayesTreeClique>::traversePath' requested here
  template class BayesTree<SymbolicBayesTreeClique>;
                 ^
1 error generated
```

gtsam/inference/BayesTree.h
```diff
@@ -245,6 +245,15 @@ namespace gtsam {
     /** Add all cliques in this BayesTree to the specified factor graph */
     void addFactorsToGraph(FactorGraph<FactorType>* graph) const;
 
+    /**
+     * @brief Returns the set of keys in a "contaminated" part of the tree
+     * @param keys: The keys defining a "contaminated" part of the tree
+     * @returns The keys affected by the factor graph computed by removeTop
+     * @note Allows one step look-ahead for what removeTop will modify
+     */
+    gtsam::KeySet traverseTop(const gtsam::KeyVector& keys) const;
```
dellaert: I love the intent of the function, don't love the name. It doesn't really say what it's doing. Consider `collectAffectedKeys` or something like this?   The documentation is also a bit confusing because it says the input is a set of contaminated keys and the output is as well. I understand what the function is doing, but neither the name nor the documentation reflects this accurately, which will be an issue when someone looks at the code six months from now :-)

gtsam/inference/BayesTree.h
```diff
@@ -260,6 +269,13 @@ namespace gtsam {
     /** Fill the nodes index for a subtree */
     void fillNodesIndex(const sharedClique& subtree);
 
+    /// @brief Traversal helper for a path used by traverseTop
+    void traversePath(gtsam::KeySet& traversedKeys,
```
dellaert: similar, `affectedPath` and `affectedCliqueKeys` ? These are just suggestions. You might have better intuition for the naming.

gtsam/nonlinear/DoglegOptimizerImpl.h
```diff
@@ -252,4 +252,123 @@ typename DoglegOptimizerImpl::IterationResult DoglegOptimizerImpl::Iterate(
...
+   * @param f The original nonlinear factor graph with which to evaluate the
+   * accuracy of \f$ M(\delta x) \f$ to adjust \f$ \delta \f$.
+   * @param x0 The linearization point about which \f$ \bayesNet \f$ was created
+   * @param verbose Flag to write debug information.
+   * @return A DoglegIterationResult containing the new \c delta, the linear
+   * update \c dx_d, and the resulting nonlinear error \c f_error.
+   */
+  template <class M, class F, class VALUES>
+  static DoglegOptimizerImpl::IterationResult Iterate(
+      const double min_delta, const double max_delta, const double step_size,
```
dellaert: Whoa, consider a small struct for all the non-changing parameters? And then you can have inline ///< comments for the members of the struct. Less daunting than the above doxygen monster :-)

Even nicer: make DoglegLineSearchImpl a templated function object, something like:

```cpp
template <class M, class F, class VALUES>
DoglegLineSearch {
  const double minDelta; ///< The minimum trust region size (scaled by size of system)
....
  DoglegOptimizerImpl::IterationResult operator()(/* whatever changes in each iteration */) {
  ...
  }
};
```

Note camelCase not snake_case naming convention in GTSAM. And not no longer static method, as refers to params.

gtsam/nonlinear/DoglegOptimizerImpl.h
```diff
@@ -252,4 +252,118 @@ typename DoglegOptimizerImpl::IterationResult DoglegOptimizerImpl::Iterate(
...
+  result.delta = step;
+  if (verbose)
+    std::cout << "Initial Step Error: " << result.f_error << std::endl;
+
+  // Search Increase delta
+  double eps = std::numeric_limits<double>::epsilon();
+  while (step < max_step - eps) {
+    // Compute the trust region for the evaluation point according to the
+    // Geometric Series
+    step = std::min(max_step, step * step_size);
```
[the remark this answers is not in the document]
dellaert: Is this true? did you address?

gtsam/nonlinear/DoglegOptimizerImpl.h
```diff
@@ -252,4 +252,123 @@ typename DoglegOptimizerImpl::IterationResult DoglegOptimizerImpl::Iterate(
...
+
+    // Compute the Evaluation Point and evaluate the error
+    VectorValues dx_d =
+        DoglegOptimizerImpl::ComputeDoglegPoint(step, dx_u, dx_n, verbose);
```
dellaert: We are DoglegOptimizerImpl, no? Not sure whether qualification is needed.

gtsam/nonlinear/ISAM2.cpp
```diff
@@ -777,6 +777,41 @@ void ISAM2::updateDelta(bool forceFullSolve) const {
...
+    // Compute steepest descent step
+    const VectorValues gradAtZero = this->gradientAtZero();
+    DeltaImpl::UpdateRgProd(roots_, deltaReplacedMask_, gradAtZero, &RgProd_);
+    const VectorValues dx_u =
+        VectorValues::Zero(deltaNewton_)
+            .update(DeltaImpl::ComputeGradientSearch(gradAtZero, RgProd_));
+    deltaReplacedMask_.clear();
+
+    // Do the DogLeg Line Search
+    DoglegOptimizerImpl::IterationResult doglegResult(
```
dellaert, line 805: And maybe that iterator can be instantiated once somewhere outside this function.

gtsam/nonlinear/ISAM2.cpp
```diff
@@ -832,4 +867,42 @@ VectorValues ISAM2::gradientAtZero() const {
...
+      update.findFluid(roots_, relinKeys, &result.markedKeys, result.details());
+  }
+  // Get the top of the bayes tree affected by all the involved keys
+  KeySet affectedKeys = traverseTop(
+      KeyVector(result.markedKeys.begin(), result.markedKeys.end()));
+  // Add the new keys to get the entire set of keys affected by the update
+  KeySet newKeys = newFactors.keys();
+  affectedKeys.insert(newKeys.begin(), newKeys.end());
+  // Return the affected keys and whether or not this will be a batch update
+  return std::make_pair(affectedKeys,
```
dellaert: just {} for a pair :-)

gtsam/nonlinear/ISAM2Params.h
```diff
@@ -126,6 +126,74 @@ struct GTSAM_EXPORT ISAM2DoglegParams {
...
 
+/**
+ * @ingroup isam2
+ * Parameters for ISAM2 using Dogleg Line Search optimization.  Either this
+ * class, ISAM2GaussNewtonParams, or ISAM2DoglegParams should be specified as
+ * the optimizationParams in ISAM2Params, which should in turn be passed to
+ * ISAM2(const ISAM2Params&).
+ */
+struct GTSAM_EXPORT ISAM2DoglegLineSearchParams {
+  double _min_delta;  ///< Minimum allowed small delta
```
dellaert: Naming: `minDelta` etc...

gtsam/nonlinear/ISAM2Params.h
```diff
@@ -126,6 +126,74 @@ struct GTSAM_EXPORT ISAM2DoglegParams {
...
 };
 
+/**
+ * @ingroup isam2
+ * Parameters for ISAM2 using Dogleg Line Search optimization.  Either this
+ * class, ISAM2GaussNewtonParams, or ISAM2DoglegParams should be specified as
+ * the optimizationParams in ISAM2Params, which should in turn be passed to
+ * ISAM2(const ISAM2Params&).
+ */
+struct GTSAM_EXPORT ISAM2DoglegLineSearchParams {
```
dellaert, line 136: Maybe that function object should just keep a reference to the parameters :-)

tests/testGaussianISAM2.cpp
```diff
@@ -317,6 +317,20 @@ TEST(ISAM2, slamlike_solution_dogleg)
   CHECK(isam_check(fullgraph, fullinit, isam, *this, result_));
 }
 
+/* ************************************************************************* */
+TEST(ISAM2, slamlike_solution_dogleglinesearch) {
```
dellaert: This and all other tests -> please use naming  `SlamLikeSolutionDoglegLineSearch`

dellaert (commented):
Sorry for the delay, I was busy with class assignments, etc.

gtsam/nonlinear/DoglegOptimizerImpl.h
DanMcGann: Thanks for all the review comments @dellaert! I'm working on the changes but had two questions here.

1. For camelCase conventions is this true for all variables, or only class/struct fields. I ask because the function inherits naming conventions like `dx_u` from the `DoglegOptimizerImpl`. 

Should I change all variable names that are used in this function even if it introduces inconsistencies with the other code?

2. For param/argument structuring as a function object. This functionality is only used  by `ISAM2::updateDelta` where "iterate" refers to an iSAM2 update not to a logical loop. Due to the design of iSAM2 (relying on `holds_alternative` to (at runtime) determine which optimizer to use) there is no place to store a function object such that can set non-changing parameters once. Rather if re-factor as a function object we would have to re-initialize this in `ISAM2::updateDelta` each time it is called. Thus the doxygen beast would re-appear in the constructor of the function object.  

TLDR - Due to requirements from `ISAM2::updateDelta`'s design I'm not sure that refactoring gets us any benefits do you have any thoughts?

gtsam/nonlinear/DoglegOptimizerImpl.h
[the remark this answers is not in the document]
DanMcGann: Yes this is an issue! It was addressed just above this comment with the following validation check
```
// Validate Step size will terminate search
  if (step < 1e-12 || step_size < 1.0)
    throw std::runtime_error(
        "Invalid DoglegLineSearch configuration. Would cause infinite search.");
```

gtsam/nonlinear/DoglegOptimizerImpl.h
dellaert: 1. Let's say use correct camelCase naming convention for anything new that you introduce. I realize that some of the existing code might not adhere completely. Also, underscore `_u` is actually fine because it's like a subscript, and we typically make an exception for that. 
2. I personally think a small parameter struct with the fields documented in line would still be more concise and clear. You can initialize it with brace initialization:
```
struct Foo {
double bar; ///< my bar param
...
}

Foo foo{ 60.0, ...};
callUpdate(foo, ...);
```

It;s a 10 second refactor for Codex :-)

dellaert (approved):
LGTM!

PR #2409, opened 2026-02-14
Robust Incremental Smoothing and Mapping (riSAM)

DanMcGann:
This is the second of 3 PRs that support the integration of [riSAM](https://arxiv.org/abs/2209.14359) into GTSAM. This PR adds the actual riSAM algorithm along with unit test to validate its functionality. In addition to the unit tests, I ran an independent test that confirmed that the implementation here matches exactly our internal riSAM implementation!

## Overview
This PR adds and test sthe `RISAM` class, and its helpers `RISAMGraduatedKernel` and `RISAMGraduatedFactor`. RISAM is a drop in replacement for ISAM2 with the same update interface. However, users can wrap any potential outlier in a `RISAMGraduatedFactor` and riSAM will solve updates that involve these factors using an incrementalized version of Graduated Non-Convexity. 

## TODO - Python Interface
This PR needs to be updated to include its python interface. Unfortunately, I was unable to find an effective way to implement such an interface. The template for `RISAMGraduatedFactor` allows it to wrap ANY gtsam factor type. In order to define the python interface for this it appears that we would need to explicitly enumerate every factor type with all of their possible template combinations. This seemed like an unmaintainable definition. Thus before adding this interface I wanted to check in to see if the maintainers had any suggestions for the best way to approach the implementation.

## Final Notes
This is PR 2/3 for adding RISAM. Once the python interface above is fixed the final PR will contain a jupiter notebook with a example / tutorial on using riSAM!


dellaert:
Awesome !

I'll wait to review thoroughly until other PR is merged. But, check naming convention esp. We're Google style *except* we camelCase variables and non-static methods. Also, please format everything with clang-format, Google style, if that was not already done. Finally, warnings are treated as errors, so please try to compile with that flag locally.

dellaert:
Could you merge in develop so PR diff is up to date?

gtsam/sam/RISAM.cpp
```diff
@@ -0,0 +1,339 @@
...
+  augmentMu(new_factors, new_factor_indices);
+
+  // Once we add new factors we can remove any factors and corresponding info
+  gtsam::NonlinearFactorGraph removed_factors;
+  removed_factors.reserve(update_params.removeFactorIndices.size());
+  for (const auto fidx : update_params.removeFactorIndices) {
+    removed_factors.push_back(factors_.at(fidx));
+    factors_.remove(fidx);
+    mu_.at(fidx).reset();
+    mu_inits_.at(fidx).reset();
```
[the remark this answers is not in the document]
DanMcGann: addressed line 202

gtsam/sam/RISAMGraduatedFactor.h
```diff
@@ -0,0 +1,173 @@
...
+  GaussianFactor::shared_ptr linearize(
+      const gtsam::Values& current_estimate) const override {
+    // Delegate to linearizeGraduated which is required by the GraduatedFactor
+    // Interface
+    return linearizeGraduated(current_estimate);
+  }
+
+  /// @brief Returns 0.5 \rho(r)^2
+  double error(const gtsam::Values& values) const override {
+    return 0.5 * std::pow(robustResidual(values), 2);
```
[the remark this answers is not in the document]
DanMcGann: I think co-pilot is wrong about this one (looks like it got lost in the notation?)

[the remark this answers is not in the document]
dellaert: You're the expert :-).

dellaert:
@DanMcGann, this is awesome.

First, some architectural questions: the "kernels" look very similar to the loss functions in linear. It would be good to discuss whether they are in fact identical in scope, and even line up with some of the robust loss functions we do have already. If this is indeed the case, we really ought to refactor it here so you can use any loss function and move the kernels you did define to the loss function library. 

Incidentally, a new robust loss, TLS, was added just days ago in the context of the GNC optimizer. 

The connection with GNC also warrants some language, perhaps in the documentation. Speaking of which, you probably want to add a notebook in nonlinear/doc to this PR that explains riSAM, based on and similar to [GNC's user guide entry](https://borglab.github.io/gtsam/gncoptimizer).

dellaert:
At least two CI failures, related to
```
In file included from /Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/string:608:
216
Error: /Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__memory/allocator.h:168:81: error: destructor called on non-final 'gtsam::SIGKernel' that has virtual functions but non-virtual destructor [-Werror,-Wdelete-non-abstract-non-virtual-dtor]
217
  168 |   _LIBCPP_DEPRECATED_IN_CXX17 _LIBCPP_HIDE_FROM_ABI void destroy(pointer __p) { __p->~_Tp(); }
218

FAILED: gtsam/CMakeFiles/gtsam.dir/sam/RISAMGraduatedKernel.cpp.o 
394
/usr/bin/clang++-14 -DNDEBUG -Dgtsam_EXPORTS -I/__w/gtsam/gtsam -I/__w/gtsam/gtsam/build -I/__w/gtsam/gtsam/CppUnitLite -I/__w/gtsam/gtsam/gtsam/3rdparty/metis/include -I/__w/gtsam/gtsam/gtsam/3rdparty/metis/libmetis -I/__w/gtsam/gtsam/gtsam/3rdparty/metis/GKlib -I/__w/gtsam/gtsam/gtsam/3rdparty/cephes -isystem /__w/gtsam/gtsam/gtsam/3rdparty/SuiteSparse_config -isystem /__w/gtsam/gtsam/gtsam/3rdparty/Spectra -isystem /__w/gtsam/gtsam/gtsam/3rdparty/CCOLAMD/Include -isystem /__w/gtsam/gtsam/gtsam/3rdparty/Eigen -w -O3 -DNDEBUG -fPIC -fdiagnostics-color=always -ftemplate-depth=1024 -Werror -Wall -Wpedantic -Wextra -Wno-unused-parameter -Wreturn-stack-address -Wno-weak-template-vtables -Wno-weak-vtables -Wreturn-type -Wformat -Werror=format-security -Wsuggest-override -O3 -Wno-unused-local-typedefs -std=c++17 -MD -MT gtsam/CMakeFiles/gtsam.dir/sam/RISAMGraduatedKernel.cpp.o -MF gtsam/CMakeFiles/gtsam.dir/sam/RISAMGraduatedKernel.cpp.o.d -o gtsam/CMakeFiles/gtsam.dir/sam/RISAMGraduatedKernel.cpp.o -c /__w/gtsam/gtsam/gtsam/sam/RISAMGraduatedKernel.cpp
395
In file included from /__w/gtsam/gtsam/gtsam/sam/RISAMGraduatedKernel.cpp:2:
396
/__w/gtsam/gtsam/gtsam/sam/RISAMGraduatedKernel.h:26:16: error: no template named 'shared_ptr' in namespace 'std'
397
  typedef std::shared_ptr<GraduatedKernel> shared_ptr;
398
          ~~~~~^
399
/__w/gtsam/gtsam/gtsam/sam/RISAMGraduatedKernel.h:108:16: error: no template named 'shared_ptr' in namespace 'std'
400
  typedef std::shared_ptr<SIGKernel> shared_ptr;
401
          ~~~~~^
```
Please check all CI failures if there are others...

DanMcGann:
Great point about the kernels @dellaert !

After looking into it unfortunately, refactoring is a little tricky and I would love you input.

GNC is able to re-use the existing `robustLoss` functions because it's formulation permits factoring out the convexification control parameter $\mu$ allowing `GNCOptimizer` to re-use the existing implementations in `robustLoss`. The graduated kernel proposed as part of `RISAM` does not support this resulting in an interface difference.

Interface difference:
| GNC / RobustLoss |RISAM |
|----------------------------|----------|
| $\rho(e)$                 | $\rho(e, \mu)$ |

Unifying the interface is possible but causes problems. We could support an optional $\mu$ parameter for any robust loss $\rho(e, \mu=\mathrm{default})$. However, the effect would be ill-defined due to multiple possible implementations (see below), and many kernels do not have proposed graduated variants.

Example: Both GNC and RISAM define a different graduated loss based on Geman-McClure (GM):
| GM-GNC  |GM-RISAM |
|----------------------------|----------|
| $\frac{\mu c^2 e^2}{\mu c^2 + e^2}$ | $\frac{c^2e^2}{c^2+(e^2)^\mu}$ |

Example: No graduation scheme for Cauchy loss has been proposed.

Additionally, beyond linearizing factors GNC implements additional logic specifically based on GM and TLS loss and extension of `GNCOptimizer` to arbitrary graduated losses appears to be non-trivial.

Thus, should we leave the RISAM kernel separate? Or should we fight the problems above and unify the interface?

dellaert:
Thanks for the notebook, that’s invaluable. Will look at kernel issue.

gtsam/sam/RISAMGraduatedFactor.cpp
```diff
@@ -0,0 +1,38 @@
+/* ----------------------------------------------------------------------------
```
dellaert, line 1: Thinking all these methods could be inlined in header.

gtsam/sam/RISAMGraduatedFactor.h
```diff
@@ -0,0 +1,173 @@
...
+
+namespace gtsam {
+
+/// @brief Graduated Factor for riSAM base class
+class GraduatedFactor {
+  /** TYPES **/
+ public:
+  typedef std::shared_ptr<GraduatedFactor> shared_ptr;
+
+  /** FIELDS **/
```
dellaert: Here and elsewhere: please use Doxygen regions as is common in GTSAM.

dellaert:
Re, kernels: I think we should fight the fight :-) I’d prefer adding a new virtual method in base with extra \mu parameters rather than carrying a default parameter everywhere.

The issue with \mu having a different interpretation in GNC and RiSAM requires more thought. Is there a way riSAM acan switch to the GNC version? Is there a re-parameterization that makes one mu into the other? Do they behave opposite? Since GNC was included first - and several papers use that approach (including recent one from Rene Vidal's group) I'd prefer if we can use that \mu, and *if needed* perhaps rename your mu to a different name - especially of it somehow has "opposite" meaning. That being said, I've not carefully looked into a riSAM again to try and understand deeply myself. I'm just still brainstorming with you :-)

dellaert:
@DanMcGann happy to chat about this move this along - send me an email and we can schedule.

DanMcGann:
Hey @dellaert thanks!
Before we fill up a meeting slot take a look at the most recent changes that I have been working on!

These are incomplete (see TODOs below) but should give a sense of the general approach summarized as:
- Add `graduatedLoss` and `graduatedWeight` to robust loss interface
- Move Weight + Loss compute into static methods to support code re-use
- Move graduation control parameter (mu)  logic into a "scheduler" that is used to construct `GraduatedFactors`.

It should allow riSAM to support any robust loss that has a graduated form. However GNC will still be limited to only GM and TLS due to the additional logic implemented based on those losses.

If you want to brainstorm improvements ping me on email and we can schedule time to chat!

Remaining TODOs
- [x] Add `GradScheme`s to TLS lost that are introduced in the GNC impl (Linear, and Super Linear)
- [x] Update GNC code for new Loss Methods
- [x] Add tests for any missing `graduated*` loss/weight methods to get complete

DanMcGann:
Okay following up to mark the requested changes complete!

* The former RISAMGraduatedKernel's functionality has now been split to re-use `mEstimator::RobustLoss` and a new class `RISAMGraduationScheduler`.
* All compatible robust loss functions are updated to support graduated variants of `loss` and `weight`.
* Tests are updated for to cover the new changes.
 
 Let me know if you are happy with the refactor @dellaert!

DanMcGann:
Hey @dellaert ! Just wanted to follow up to see if there were any changes you wanted to see on the implemented design!

dellaert:
@DanMcGann i’ll try to take a look today - I’ve been procrastinating on this .

gtsam/sam/RISAM.h
```diff
@@ -0,0 +1,254 @@
...
+ *  @date   October 2025
+ */
+#pragma once
+#include <gtsam/nonlinear/ISAM2.h>
+#include <gtsam/nonlinear/internal/ChiSquaredInverse.h>
+#include <gtsam/sam/RISAMGraduatedFactor.h>
+
+namespace gtsam {
+
+class RISAM {
```
dellaert, line 25: Please add a doxygen comment. Also, it is a non-template public class with out-of-line implementation, but it is not marked GTSAM_EXPORT - so that probably explains some of the CI failures.

gtsam/sam/RISAMGraduatedFactor.h
```diff
@@ -0,0 +1,199 @@
...
+ *  @author Dan McGann
+ *  @date Mar 2022
+ */
+#pragma once
+#include <gtsam/nonlinear/NonlinearFactor.h>
+#include <gtsam/sam/RISAMGraduationScheduler.h>
+
+namespace gtsam {
+
+/// @brief Graduated Factor for riSAM base class
```
dellaert, line 25: Same: GraduatedFactor has out-of-line methods in .cpp and is public API-facing, but lacks GTSAM_EXPORT.

gtsam/sam/RISAMGraduationScheduler.h
```diff
@@ -0,0 +1,175 @@
...
+
+#include <memory>
+#include <optional>
+
+namespace gtsam {
+
+/** @brief Base class for graduation scheduling for riSAM
+ * Advanced users can write their own schedulers by inheriting from this class
+ */
+class GraduationScheduler {
```
dellaert, line 21: Same: GraduationScheduler/derived scheduler classes are non-template public types with out-of-line virtual methods. Please review `Using-GTSAM-EXPORT.md` guidance.

gtsam/sam/RISAM.cpp
```diff
@@ -0,0 +1,342 @@
...
+  }
+
+  // If we have converged update mu_inits_based on status
+  if (params_.increment_outlier_mu) incrementMuInits();
+
+  return result;
+}
+
+/* ************************************************************************* */
+RISAM::UpdateResult RISAM::updateRobust(
```
dellaert, line 78: This method is too long and too complex. Consider splitting up so it's easier to read and review.

gtsam/sam/RISAM.cpp
```diff
@@ -0,0 +1,342 @@
...
+  update_result.isam2_result =
+      solver_->update(new_factors, new_theta, initial_update_params);
+  solver_->calculateEstimate();
+
+  // Run GNC iterations until all convex factors have converged w.r.t. mu_
+  gtsam::FactorIndices remaining_convex_factors(convex_factors.begin(),
+                                                convex_factors.end());
+  // Internal params force update convex factors + ensure better ordering
+  gtsam::ISAM2UpdateParams update_params_internal;
+  while (remaining_convex_factors.size() > 0) {
```
dellaert, line 110: For example, this loop body could probably be a helper function?

gtsam/sam/RISAM.cpp
```diff
@@ -0,0 +1,342 @@
...
+            grad_factor->scheduler()->updateMuInit(*(mu_inits_[fidx]), true);
+      }
+    }
+    // Reset accumulator
+    factors_to_check_status_.clear();
+  }
+}
+
+/* ************************************************************************* */
+std::set<gtsam::FactorIndex> RISAM::convexifyInvolvedFactors(
```
dellaert, line 275: Similar comment: rather complex; suggest at least key-selection and factor-selection helpers to keep logic concise and testable.

My rule: "no method longer than 4 lines" :-) excluding comments, and only approximate, but the spirit is there.

gtsam/sam/RISAM.cpp
```diff
@@ -0,0 +1,342 @@
...
+  // Fill out the Update Result
+  update_result.involved_variables = involved_keys;
+  update_result.convexified_factors = convex_factors;
+  update_result.affected_variables = affected_keys;
+
+  // Return
+  return convex_factors;
+}
+
+}  // namespace gtsam
```
dellaert, line 342: Add newline at end of all files

gtsam/linear/tests/testNoiseModel.cpp
```diff
@@ -640,6 +640,23 @@ TEST(NoiseModel, robustFunctionHuber)
   DOUBLES_EQUAL(0.5000, huber->loss(error4), 1e-8);
 }
 
+TEST(NoiseModel, robustFunctionHuberGraduated) {
```
dellaert, line 643: We should find a way to copy to double down on copy/pasta among the new (and existing) tests. Codex 5.4 is pretty good at refactoring along those lines.

gtsam/nonlinear/doc/RISAM.ipynb
```diff
@@ -0,0 +1,127 @@
+{
```
dellaert, line 1: The notebook preamble order does not match our documented standard (title/introduction, then copyright remove-cell, then Colab badge, then Colab install remove-cell, then imports/setup code). Current order has badge before copyright and no dedicated imports/setup cell.

See copilot-instructions.md

dellaert (commented):
Dan, sorry to be so late in reviewing this. I have a number of comments below, mostly about the use of the export directive, which needs to be done correctly or CI fails. 

But my main concern is still the graduated convexity convention. We now seem to have a plethora of schemes: the meaning of $\mu$ changes from noise model to noise model and in fact, _within_ a particular noise model there are different conventions based on the scheme chosen. 

I know some of these predate you, in particular the very recent linear and super linear schemes in TLS. Because it's very recent, however, I think we can still change this before we release a 4.3 version. I should have caught that inconsistency when Harneet added that. 

My preferred solution would be to re-parameterize all the graduation schemes to follow a particular scheme. Because the very first mechanism implemented in GTSAM - and several people/terams already use GNC - is the STANDARD GM-scheme from infinity (convex) to 1 (robust), that to me seems an obvious candidate.

dellaert:
A counter-argument could be that we should define the parameter$\mu$ for each of these models as they appear in the source papers. We might then want to be more circumspect about the differences, in the header file and "user guide" notebooks, and make sure the formulas and the reference to the papers are included.

dellaert:
> A counter-argument could be that we should define the parameter$\mu$ for each of these models as they appear in the source papers. We might then want to be more circumspect about the differences, in the header file and "user guide" notebooks, and make sure the formulas and the reference to the papers are included.

Ping on this, @DanMcGann  :-)

danmcgann-fieldai:
Hey Frank! Just following up that I have seen this, though time to work on it has been very limited. Hope to have more time post ICRA.

DanMcGann:
## Updates - 2026-05-25

### What Changed
* Added `GTSAM_EXPORT` where needed. 
* Broke up large `RISAM` methods into multiple helpers.
* Refactored robust loss tests to avoid code copy.
* Update ordering of `RISAM` notebook.
* General documentation + code quality improvements.

### Graduation Interface Discussion
On the topic of the interface for $\mu$, I would argue for allowing different meanings of $\mu$ for each loss.
This is both due to the points @dellaert made above, in addition to a few nuanced aspects.

1. Loss Functions have different definitions of $\infty$ 
Even if we normalize the interface to Non-Convex=1 -> Convex= $\infty$, $\infty$ will still have different meanings between loss functions. Due to their functional form different loss functions have different sensitivities to the parameter $\mu$ (from a numerical stability perspective).  This is most evident in the robust loss tests (below) where we need to provide different loss methods different values for $\infty$ to ensure numerical stability and correctness.
https://github.com/DanMcGann/gtsam/blob/e2a2ed0ba617ed78f1ad5794af70b99d322f5ab0/gtsam/linear/tests/testNoiseModel.cpp#L1007-L1027 

2. $\infty$ is challenging programmatically
Really if we instituted a standard interface (Non-Convex=1 -> Convex= $\infty$) we would really be instituting an interface of Non-Convex=1 -> Convex=`BIGNUM`. Where BIGNUM is some arbitrary large number. Given point 1 above we would then have to internally scale `BIGNUM` for each specific loss function. Given the internal scaling we can choose any value of `BIGNUM` and option is going to appear unnaturally arbitrary.   

Due to these practical challenges along with the points Frank made above I would argue that it is best to allow loss functions to define their interface independently. The interface is currently documented in the class doc strings for each loss function:
https://github.com/DanMcGann/gtsam/blob/e2a2ed0ba617ed78f1ad5794af70b99d322f5ab0/gtsam/linear/LossFunctions.h#L297-L298

Please let me know which style of interface we would like to go with and I will happily made all the needed changes! If we want to maintain different interfaces for each loss, please let me know where else we would like to document this! 

#### Side Thoughts
We could also normalize the interface through a functional interface. In this each loss class would provide a method that outputs a pair (Convex, NonConvex) which are the $\mu$ values for which the loss method is convex or non-convex. This would allow each to have arbitrary values, but would give developers a standard interface through which to access the information.

dellaert:
Thanks @DanMcGann 

Do me a favor an "resolve" all comments you addressed.

I'm looking over this and I'm still very worried. Loss functions are already complex and introducing a new parameter that has different meanings in different places and different algorithms is bound to create a lot of chaos. 

I had a conversation with ChatGPT Pro and came up with a scheme that is very close to what you propose in RISAM. I attach the proposal, and implementation notes with respect to the *current* develop. Let me know what you think.
[normalized_gnc_mu_scheme_for_gtsam.md](https://github.com/user-attachments/files/28234281/normalized_gnc_mu_scheme_for_gtsam.md)

dellaert:
PS some still CI failures as well, it seems.

DanMcGann:
Okay! Finally found the time to resolve the changes requested above! 

The key change here was to follow the suggested scheme and define a common graduation interface such that the graduation parameter $\mu$ is always in the range of $[0, 1]$ for all loss types and between both RISAM and GNC.

## What has changed:
* Rebased riSAM changes on top of the current develop.
* Reworked the interface for all graduated robust loss methods to canonically define $\mu \in [0,1]$
    * Updated `GncOptimizer` implementation to work with new interface.
        * Updated interaction with RobustLoss.
        * Updated naming conventions.
    * Updated `RISAM` implementation to work with the new interface.
        * Updated Scheduler + docs.
* Cleaned up RobustLoss function tests to reduce copied code.
* Hopefully resolved remaining CI failures.

## Important Consequences
These changes introduce a interface breaking change in naming convention to adhere to the standardized $\mu$ definition. Specifically, GNC parameters and functions that used to refer to "mu" now refer to "lambda". This changes key interface naming for the `GncOptimizer`.

## Notes and Nuance
* Most robust losses still utilize a hidden internal parameter `\lambda` that is used to actually convexify/deconvexify the loss. Each loss has different sensitivity to this param. Together this necessitates defining per-loss `LAMBDA_MAX` used to compute $\lambda$ from $\mu$ so that the graduated loss remains numerically stable.
* GNC's internal logic is heavily reliant on the parameter $\lambda$ ( defined as $\lambda$ after these changes, this was formerly defined as $\mu$ in the GNC code). I choose to leave this logic to be based on $\lambda$ rather than converting everything to use the new definition of $\mu$ as I wanted to ensure that the authors original intent was maintained.
* Documentation changes to ensure consistency occured in the GNC impl/header/notebook, RobustLoss impl/header, and in the new RISAM impl/header/notebook. I believe this is comprehensive, but it would be good to get a double check on that.

DanMcGann:
@dellaert gentle bump!

dellaert:
Sorry, @DanMcGann. I was travelling extensively and only now able to get to this.

dellaert:
### 1/8 — Proposal comparison: normalized `mu` mapping

I asked Codex to compare the latest changes directly against the attached `normalized_gnc_mu_scheme_for_gtsam.md` proposal. 

The proposal calls for exact normalized mappings and explicit endpoints: I proposed GM should use the normalized formula directly (equivalently historical `lambda = 1 / mu`), while Yang/Peng TLS should use `theta = mu / (1 - mu)`, with `mu = 0` handled as the all-inlier least-squares initialization and `mu = 1` as exact final TLS.

The current code instead maps `mu` through per-loss arbitrary constants such as `LAMBDA_MAX` and `LAMBDA_MIN`:

- [GM implementation](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.cpp#L404-L423)
- [TLS implementation](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.cpp#L500-L543)

That is the `BIGNUM` encoding the proposal specifically tried to avoid. It also gives the wrong practical Yang-TLS endpoint: the test currently expects near-zero weights/loss at `mu = 0`, rather than all-inlier least squares ([test](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/tests/testNoiseModel.cpp#L1022-L1036)).

If you agree with these points, please implement the normalized formulas directly, and keep any historical `lambda` or `theta` only as explicitly named internal scheduling state.

dellaert:
### 2/8 — General review: Python wrapper build is broken

While Codex was comparing the changes with the Markdown proposal, I also asked it to check the broader integration. The issue is caused by the C++ API changes in this PR.

Using the required `py312` environment, `conda run -n py312 cmake --build build --target gtsam_py -j 6` fails because:

1. The TLS wrapper retains `Create(c, ReweightScheme)`, while C++ now interprets the second argument as `GradScheme`: [linear.i](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/linear.i#L207-L215).
2. The Custom wrapper repeats the parameter names `weight` and `loss`, generating illegal duplicate C++ lambda parameters: [linear.i](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/linear.i#L273-L286).
3. The GNC wrapper still exposes `muStep`, `setMuStep`, and verbosity `MU`, although C++ renamed them to `lambdaStep`, `setLambdaStep`, and `LAMBDA`: [nonlinear.i](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/nonlinear/nonlinear.i#L293-L332).

Please synchronize the wrapper declarations and parameter names exactly with the headers, regenerate the bindings, and make the `gtsam_py` target pass.

dellaert:
### 3/8 — General review: graduated factor error squares an existing loss

This is a general correctness issue, introduced by the new riSAM code in this PR and is not present in the base branch.

`robustResidual()` returns `robust_loss_->graduatedLoss(...)`, i.e. an already-computed robust cost. `error()` then returns `0.5 * robustResidual(values)^2`, squaring that cost again: [RISAMGraduatedFactor.h](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/sam/RISAMGraduatedFactor.h#L180-L195).

Please either return the graduated loss directly from `error()`, or redefine/rename the intermediate API to return a genuine residual such as `sqrt(2 * loss)`. The current combination is inconsistent with GTSAM's nonlinear-factor error convention and the loss/weight linearization.

dellaert:
### 4/8 — General review: chi-square thresholds are compared with a norm

Again, a riSAM correctness issue introduced by this PR, not pre-existing in `develop`.

`GraduatedFactor::residual()` returns a whitened residual norm (`sqrt(2 * factor.error)`). However, `internal::chiSquaredQuantile()` returns a chi-square value, i.e. a threshold for the squared norm. The code compares these values directly in both:

- [`getOutliers`](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/sam/RISAM.cpp#L79-L91)
- [`incrementMuInits`](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/sam/RISAM.cpp#L333-L359)

Please compare `residual * residual` against the chi-square quantile, or take the square root of the quantile and consistently document the quantity. Please also add tests immediately below and above a known chi-square boundary; the current very large outlier test does not catch this units mismatch.

dellaert:
### 5/8 — Proposal comparison + compatibility: preserve endpoint-loss APIs

This is both a proposal-compliance issue and a public-API compatibility issue introduced by this PR.

The proposal's migration plan explicitly says to add graduated helpers without changing existing endpoint-loss behavior. The PR instead changes existing GM/TLS constructors and factories by inserting `GradScheme` before `ReweightScheme`: [GM API](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.h#L453-L483), [TLS API](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.h#L549-L580).

More seriously, `GemanMcClure::Weight` previously accepted `(distance2, c2)`; this PR silently reverses the two same-typed arguments to `(c2, distance2)` at [LossFunctions.h](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.h#L474-L477). Existing downstream code still compiles but computes a different value.

Please preserve the original constructors/factories and the original `Weight(distance2, c2)` contract, then add distinctly named graduated helpers or non-ambiguous overloads. Deprecation aliases would be preferable for any unavoidable GNC rename.

dellaert:
### 6/8 — General review: graduation mode is lost by equality and serialization

This is a general state/serialization issue introduced by this PR, as the `graduation_` fields are new in this PR.

For both Geman-McClure and TLS, `equals()` only compares the scale, and Boost serialization omits `graduation_`:

- [GM equality](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.cpp#L446-L450) and [serialization](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.h#L499-L513)
- [TLS equality](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.cpp#L568-L573) and [serialization](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/linear/LossFunctions.h#L582-L596)

As a result, a scale-invariant GM object can serialize/deserialize as `STANDARD`, while equality cannot detect the change.

Please include the graduation scheme in serialization and equality (and preferably `print()`), with round-trip tests for every scheme.

dellaert:
### 7/8 — General review: cloning resets the current graduation progress

Another riSAM correctness issue introduced by this PR:

`GraduatedFactor` has a copy constructor that preserves the current `mu` value, but `GenericGraduatedFactor` does not invoke it. It reconstructs the base from `robust_loss_` and `scheduler_`, which initializes `mu` back to `scheduler_->muInit()`: [RISAMGraduatedFactor.h](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/sam/RISAMGraduatedFactor.h#L57-L66), [derived copy constructor](https://github.com/borglab/gtsam/blob/2de39eb492e3e41726224315acbb035bdbed25b0/gtsam/sam/RISAMGraduatedFactor.h#L122-L131).

Please invoke `GraduatedFactor(other)` from the derived copy constructor and add a clone test that advances a factor to an intermediate `mu`, clones it, and verifies identical loss/weight behavior without aliasing the two `mu` states.

dellaert:
### 8/8 — Branch integration: merge current `develop`; please do not rebase

To get closer to develop and resolve conflicts, please *merge* the current `borglab/gtsam:develop` into `feature/risam` and resolve the conflicts with a normal merge commit. **Please do not rebase or force-push this branch,** as I already have `feature/risam` checked out locally. (no "license to force-push" :-))

After resolving the conflicts, please run at least:

- `make -j6 testNoiseModel.run`
- `make -j6 testGncOptimizer.run`
- `make -j6 testRISAM.run`
- the Python wrapper build/tests in the `py312` environment

Then push the merge commit to update this PR.

dellaert:
I'm sorry for the churn and the long time to integrate this. It's a very complex PR.

dellaert:
codex: This adds the riSAM robust incremental smoothing algorithm, graduated factors and scheduling, tests, and documentation. Labeled enhancement because it introduces a new robust estimation capability.

DanMcGann:
No worries about the churn Frank!
This PR has ballooned a little bit and I agree 100% that it is worth it to spend the time to ensure it is the best it can be!
I spent a bit more time on these changes than the last. 
They hopefully, address the review and generally implement our goals a bit more consistently.

I tried to address each review comment in a unique commit to make tracking the changes a little easier.
* 1 - [ac32507](https://github.com/borglab/gtsam/pull/2409/commits/ac32507e2fef86e8b5c56978133f1d01794a672d): Fixes the $\mu$ normalization.
* 2 - [ba52555](https://github.com/borglab/gtsam/pull/2409/commits/ba5255549bbd07cc20244e1a709bb2d96457db91): Updates the gtwrap definitions to fix python bindings.
* 3 - [10caa48](https://github.com/borglab/gtsam/pull/2409/commits/10caa484a751047f158384856d78f5981677d5ae): Updates the risam loss iface defs.
* 4 - [153bf56](https://github.com/borglab/gtsam/pull/2409/commits/153bf5614398d444081a26ad609e2ff056f1d9bf): Fixes Chi^2 usage.
* 5 - [0250f84](https://github.com/borglab/gtsam/pull/2409/commits/0250f844a67cd5e2f338d541676a86ecb2fd7d9b): Ensures consistent and backwards compatible loss function interface.
* 6 - [c764ce1](https://github.com/borglab/gtsam/pull/2409/commits/c764ce15ab4e7c9d9027419345a1d2af35792005): Fixes graduated factor serialization.
* 7 - [6797eec](https://github.com/borglab/gtsam/pull/2409/commits/6797eec6181776d343e7680511e51736d4143cc4): Fixes graduated factor cloning.
* 8 - [5467bd6](https://github.com/borglab/gtsam/pull/2409/commits/5467bd63be05c81cc3280bd1f11eef0d3bf08d42): Merges current develop.

The remaining commits are general cleanup + corrections from these changes, and to ensure consistency with all the changes in this PR.

Finally, I will highlight one remaining TODO which is the risam python interface. If you look up in this thread you will see I held off on this due to the apparent need to enumerate all factor types.

dellaert:
Awesome!

I'm out of copilot tokens so I'll ask codex to re-review the comments above and resolve those that are done. If all issues have been addressed I'll read through the code once more. 

About wrapper, let me think. We could also hold off and do this in a follow-up PR.

DanMcGann:
Whoops, I missed your response!

I let claude take a crack at writing the bindings. 
As I feared, its a lot of boiler plate that needs to get maintained whenever there is a new factor / geometric type.
I will remove the bindings for now, but it's at least a good reference on what is required.

dellaert:
Dan, I attached a detailed doc I created with codex.
Bottom-line: the 8 big comments above are satisfied. It found small issues with 4 and 5, but I'm good.
One issue I would like to see addressed before I do a detailed read again is the naming conventions, as spelled out in .github/copilot-instructions.md. The html document includes a detailed prompt for Claude ;-)
[pr2409-rereview-report.html](https://github.com/user-attachments/files/31093941/pr2409-rereview-report.html)

dellaert:
Looking back over it, I like `MakeGraduated` better than `MakeGraduatedFactor`

dellaert:
Could we use currying for `MakeGraduated`? The first application would select the graduation policy, while the second would contain only the original factor's constructor arguments. This makes the `()()` form visually communicate the separation:

```cpp
auto prior = RISAM::MakeGraduated<PriorFactor<Pose2>>(loss, scheduler)(x0, priorPose, priorNoise);
auto between = RISAM::MakeGraduated<BetweenFactor<Pose2>>(loss, scheduler)(x0, x1, odometry, odometryNoise);
auto bearingRange = RISAM::MakeGraduated<BearingRangeFactor<Pose2, Point2>>(loss, scheduler)(x1, l0, bearing, range, measurementNoise);
```

This is especially useful when constructing many factors of the same type with the same graduation policy:

```cpp
auto makeGraduatedBearingRange =
    RISAM::MakeGraduated<BearingRangeFactor<Pose2, Point2>>(
        loss, scheduler);

for (const auto& measurement : measurements) {
  newFactors.add(makeGraduatedBearingRange(
      measurement.poseKey,
      measurement.landmarkKey,
      measurement.bearing,
      measurement.range,
      measurementNoise));
}
```

A prototype implementation could be:

```cpp
template <class FactorType>
static auto MakeGraduated(
    RobustLoss::shared_ptr loss,
    GraduationScheduler::shared_ptr scheduler) {
  return [loss = std::move(loss),
          scheduler = std::move(scheduler)](auto&&... factorArgs) {
    return std::make_shared<GenericGraduatedFactor<FactorType>>(
        loss, scheduler,
        std::forward<decltype(factorArgs)>(factorArgs)...);
  };
}
```

This keeps the two categories of arguments structurally separate:

```text
MakeGraduated<FactorType>(graduation arguments)(original factor arguments)
```

Each invocation of the returned function still constructs a distinct graduated factor with its own per-factor graduation state, while reusing the selected loss and scheduler.

DanMcGann:
Renaming + migration to currying complete!

dellaert (approved):
Awesome! @ProfFan and I could brainstorm on how to provide a way to do the wrapping. Many thanks for this great contribution!

PR #2410, opened 2026-02-16
Allowing Non Noise Model Factors for Gnc

hkhanuja:
This PR makes sure that non NoiseModel factors do not throw errors during initialization for Gnc Optimization.
Earlier in the GncOptimizer initialization, if the factor is not a  NoiseModelFactor, then the line `graph.at<NoiseModelFactor>(i)` returns a null pointer, which later causes a segmentation fault. 
In this PR, we add fallbacks so that instead of a segmentation fault, those factors are considered as known inliers, and while they contribute to the loss, their weights are fixed.

### DIFF
--- a/gtsam/nonlinear/GncOptimizer.h
+++ b/gtsam/nonlinear/GncOptimizer.h
@@ -29,7 +29,6 @@
 #include <gtsam/nonlinear/GncParams.h>
 #include <gtsam/nonlinear/NonlinearFactorGraph.h>
 #include <gtsam/nonlinear/internal/ChiSquaredInverse.h>
-#include <set>
 
 namespace gtsam {
 /*
@@ -38,6 +37,35 @@ namespace gtsam {
  */
 static double Chi2inv(const double alpha, const size_t dofs) {
   return internal::chiSquaredQuantile(dofs, alpha);
+}
+
+/**
+ * @enum Type
+ * @brief Enum to classify factor types in GNC optimization.
+ */
+enum class Type {
+  Normal,         ///< Normal case.
+  Inlier,         ///< Factor is a known inlier.
+  Outlier,        ///< Factor is a known outlier.
+  NonNoiseModel,  ///< Factor does not have a noise model
+  NullPointer,     ///< Factor pointer is null.
+  InlierNonNoiseModel ///< Factor is an inlier and does not have a noise model
+};
+
+bool isNullType(Type type) {
+  return type == Type::NullPointer;
+}
+
+bool isNonNoiseModelType(Type type) {
+  return type == Type::NonNoiseModel;
+}
+
+bool needsWeightUpdate(Type type) {
+  return type == Type::Normal;
+}
+
+bool hasNoise(Type type) {
+  return type == Type::Normal || type == Type::Inlier || type == Type::Outlier;
 }
 
 /* ************************************************************************* */
@@ -53,7 +81,7 @@ class GncOptimizer {
   GncParameters params_; ///< GNC parameters.
   Vector weights_;  ///< Weights associated to each factor in GNC (this could be a local variable in optimize, but it is useful to make it accessible from outside).
   Vector barcSq_;  ///< Inlier thresholds. A factor is considered an inlier if factor.error() < barcSq_[i] (where i is the position of the factor in the factor graph. Note that factor.error() whitens by the covariance.
-  std::set<size_t> knownInliers_; ///< Set of known Inliers for the Gnc Optimization process
+  std::vector<Type> factorTypes_;  ///< Cached factor types for GNC.
 
  public:
   /// Constructor.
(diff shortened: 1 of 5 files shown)


gtsam/nonlinear/GncOptimizer.h
```diff
@@ -65,13 +65,25 @@ class GncOptimizer {
     for (size_t i = 0; i < graph.size(); i++) {
       if (graph[i]) {
         NoiseModelFactor::shared_ptr factor = graph.at<NoiseModelFactor>(i);
+        if (!factor) {
+          // Non-NoiseModelFactor (e.g. KarcherMeanFactorPose3).
+          nfg_[i] = graph[i];
+          params_.knownInliers.push_back(i);
```
akshay-krishnan: updating this param is a not ideal. I think users may construct this class with the params and assume it to not change. Perhaps we could create a new class member for knownInliers_, and maybe that could be a `std::set` so it avoids the sorting below.

akshay-krishnan: @dellaert  wdyt?

hkhanuja: Just a little more explanation.
If we don't add the line `params_.knownInliers.push_back(i);` then we are dependent on the user setting the non NoiseModel factors as inliers. The later sorting and unique is to make sure that the `knownInliers `are sorted and are unique. Currently, as defined in GncParams.h `knownInliers `is a `FastVector<uint64_t>`. Since we do not want a case where the users use the function `setKnownInliers` and we add duplicates to `knownInliers` with our push_back, we only keep the unique indices with `erase `and `std::unique`.

dellaert:
Based on a very quick review with Akshay, I agree that modifying the params after they've been set is not the right way to go. Something better would be a parameter setting saying, "Treat non-noise model factors as inliers." And this is then executed by the optimizer, but not by modifying the params struct, which is typically presumed constant after construction. Then if that parameter is _not_ set - and I would set it by default as false- the optimizer simply throws an exception if it encounters non-noise model factors - and the exception message would point to the existence of the flag :-).

gtsam/nonlinear/GncOptimizer.h
hkhanuja: updated to add a new class member knownInliers_

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -65,13 +65,25 @@ class GncOptimizer {
...
         // if the factor has a robust loss, we remove the robust loss
         nfg_[i] = robust ? factor-> cloneWithNewNoiseModel(robust->noise()) : factor;
       }
     }
 
+    // Ensure known inliers are sorted/unique before set operations.
+    std::sort(params_.knownInliers.begin(), params_.knownInliers.end());
+    params_.knownInliers.erase(
+        std::unique(params_.knownInliers.begin(), params_.knownInliers.end()),
+        params_.knownInliers.end());
```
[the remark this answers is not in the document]
hkhanuja: logic changed.

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -280,6 +298,10 @@ class GncOptimizer {
          */
         for (size_t k = 0; k < nfg_.size(); k++) {
           if (nfg_[k]) {
+            auto nf = std::dynamic_pointer_cast<NoiseModelFactor>(nfg_[k]);
+            if (!nf) {
+              continue;
+            }
```
akshay-krishnan: nit: this can be a single line in c++
`if (!nf) continue;` 
same below as well.

tests/testGncOptimizer.cpp
```diff
@@ -683,6 +686,49 @@ TEST(GncOptimizer, barcsq_heterogeneousFactors) {
   // std::cout <<  "fg[3]->dim() " << fg[3]->dim() << std::endl; // this segfaults?
 }
 
+/* ************************************************************************* */
```
akshay-krishnan, line 689: seems like this does not call `optimize()`?

akshay-krishnan (approved):

gtsam/nonlinear/GncOptimizer.h
hkhanuja: okay will change

tests/testGncOptimizer.cpp
hkhanuja: done

gtsam/nonlinear/GncParams.h
```diff
@@ -72,6 +72,7 @@ class GncParams {
   double relativeCostTol = 1e-5;  ///< If relative cost change is below this threshold, stop iterating
   double weightsTol = 1e-4;  ///< If the weights are within weightsTol from being binary, stop iterating (only for TLS)
   Verbosity verbosity = SILENT;  ///< Verbosity level
+  bool allow_non_noisemodel_factors = false;  ///< If true, factors without noise model are not reweighted and not not included in mu calculation
```
dellaert: naming convention is `allowNonNoiseModelFactors`

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -61,10 +63,20 @@ class GncOptimizer {
...
 
     // make sure all noiseModels are Gaussian or convert to Gaussian
+    knownInliers_.insert(params.knownInliers.begin(), params.knownInliers.end());
     nfg_.resize(graph.size());
     for (size_t i = 0; i < graph.size(); i++) {
       if (graph[i]) {
         NoiseModelFactor::shared_ptr factor = graph.at<NoiseModelFactor>(i);
+        if (!factor) {
+          if (!params.allow_non_noisemodel_factors) {
+            throw std::runtime_error("GncOptimizer::constructor: the user must set allow_non_noisemodel_factors as "
```
dellaert: fix naming here too

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -400,6 +422,11 @@ class GncOptimizer {
     for (size_t i = 0; i < nfg_.size(); i++) {
       if (nfg_[i]) {
         auto factor = nfg_.at<NoiseModelFactor>(i);
```
dellaert: Here we can use a *static* cast if we make sure all are NoiseModelFactors

dellaert (changes requested):
Unfortunately, I feel that this one is not ready to merge yet. Dynamic pointer casts in C++ are quite expensive. And now we're adding several of them in an inner loop - for a rare use case. In addition, I found that there was one in the original code that could be replaced with a static cast.

I think a workable solution is put all the non-noise model factors at the end in nfg_: then in the loops over the nfg_, which are all size_t indexed, we just have to treat the last `nrNonNoiseModelFactors_` differently. Then we also need only one new member variable: `nrNonNoiseModelFactors_`.

gtsam/nonlinear/GncParams.h
hkhanuja: changed

gtsam/nonlinear/GncOptimizer.h
hkhanuja: changed

gtsam/nonlinear/GncOptimizer.h
hkhanuja: done

dellaert (approved):
Nice !

PR #2411, opened 2026-02-16
Continuous-Time Gaussian Process Factors (Factors and Interpolation)

holmesco:
# Summary

This PR introduces the core factors that can be used for continuous-time, Gaussian-process estimation in GTSAM. Note that we restrict our focus to White-Noise-on-Acceleration (WNOA) motion priors, though we have kept the framework as general as possible for expansion to other motion priors in the future.

In particular, this PR introduces the following main features:
- A WNOA motion prior factor 
- A WNOA Interpolation class enables (on-manifold) mean and covariance interpolation based on neighboring states.
- A "wrapper class" that allows one to convert standard GTSAM factors defined on interpolated states to an equivalent factor defined on the bordering non-interpolated states. 
- A conversion function that transforms an existing factor graph into an equivalent factor graph with a set of interpolated states removed from the main optimization.
- We introduce a simple 1d class called `Point1` (used for simplified problems and testing).

# Review

Below is a brief file description. The ordering of files also represents our suggested ordering for file review.

| File | Description |
|---- | ----|
| gtsam/nonlinear/WNOAStateData.h | Lightweight container for timestamped pose/velocity state identifiers|
| gtsam/nonlinear/WNOAInterpolator.h/.cpp | Core interpolation engine for poses/velocities under WNOA prior|
| gtsam/nonlinear/WNOAFactor.h | WNOA motion prior factor implementation|
| gtsam/nonlinear/WNOAInterpFactor.h | Wrapper for interpolating factors at non-estimated states. Also contains functions for converting factor graphs|
|gtsam/geometry/Point1.h/.cpp | 1D point type for testing/examples|

# Testing

Introduced functionality has been extensively unit tested. Our unit tests were restricted to Vector space and Lie Group variables (`Point1`, `Point2`, `Point3`, `Pose2`, `Pose3`)

| File | Description |
|---- | ----|
|gtsam/nonlinear/tests/testWNOAFactor.cpp | Unit tests for WNOA motion factors|
|gtsam/nonlinear/tests/testWNOAInterpFactor.cpp | Unit tests for interpolation factors|
|gtsam/nonlinear/tests/testWNOAInterpolator.cpp | Unit tests for interpolation function |
|gtsam/geometry/tests/testPoint1.cpp | Unit tests for Point1|

# Context

This PR is the first (and largest) of two PRs that merge continuous-time gaussian process factors into GTSAM:

1. Factors and Interpolation
2. Factor Graph and LM Optimizer that reduce computation overhead by sharing interpolated states

### DIFF
--- a/gtsam/nonlinear/WNOAFactor.h
+++ b/gtsam/nonlinear/WNOAFactor.h
@@ -11,7 +11,8 @@
 
 /**
  * @file    WNOAFactor.h
- * @brief   White-Noise-On-Acceleration (WNOA) motion prior factor between two states (pose and velocity at times t_k and t_{k+1})
+ * @brief   White-Noise-On-Acceleration (WNOA) motion prior factor between two
+ * states (pose and velocity at times t_k and t_{k+1})
  * @author  Connor Holmes
  * @author  Sven Lilge
  */
@@ -25,7 +26,7 @@
 #include <gtsam/geometry/Pose2.h>
 #include <gtsam/geometry/Pose3.h>
 #include <gtsam/nonlinear/NonlinearFactor.h>
-#include <gtsam/nonlinear/StateData.h>
+#include <gtsam/nonlinear/WNOAStateData.h>
 
 #include <cassert>
 
@@ -189,7 +190,8 @@ class WNOAMotionFactor
     } else {
       xi = traits<Pose>::Logmap(traits<Pose>::Between(p1, p2), &right_jac_inv);
     }
-    // Compute error for local state vector (pose, velocity) in the tangent space
+    // Compute error for local state vector (pose, velocity) in the tangent
+    // space
     Vector2N err;
     err << xi - delta_t_ * v1, right_jac_inv * v2 - v1;
 
@@ -236,14 +238,15 @@ class WNOAMotionFactor
    *
    * Returns the 2N x 2N covariance matrix for a WNOA prior discretized over
    * `timestep` using the diagonal PSD `Q`.
-   * See (11.7) in (Barfoot, 2024) for the derivation of covariance for the WNOA prior
+   * See (11.7) in (Barfoot, 2024) for the derivation of covariance for the WNOA
+   * prior
    *
    * @param timestep Time interval over which to compute covariance
    * @param Q Diagonal PSD vector
    * @return Matrix2N Process covariance
    */
   static Matrix2N buildWNOACovariance(double timestep, const VectorN& Q) {
-    // 
+    //
     Matrix2N covariance;
     MatrixN Q_diag = Q.asDiagonal();
     covariance << (1.0 / 3.0 * pow(timestep, 3)) * Q_diag,
@@ -304,7 +307,8 @@ class WNOAMotionFactor
   }
 
   /**
-   * @brief Compute interpolation Jacobian with respect to the previous (left) state.
+   * @brief Compute interpolation Jacobian with respect to the previous (left)
+   * state.
    *
    * Computes the 2N x 2N Jacobian block that maps perturbations in the
    * previous bordering state (pose_k, vel_k) to perturbations in the
@@ -347,7 +351,8 @@ class WNOAMotionFactor
   }
 
   /**
-   * @brief Compute interpolation Jacobian with respect to the next (right) state.
+   * @brief Compute interpolation Jacobian with respect to the next (right)
+   * state.
    *
    * Computes the 2N x 2N Jacobian block that maps perturbations in the
    * next bordering state (pose_kp1, vel_kp1) to perturbations in the
--- a/gtsam/nonlinear/WNOAFactor.h
+++ b/gtsam/nonlinear/WNOAFactor.h
@@ -326,7 +326,7 @@ class WNOAMotionFactor
    * @param deltaT Time interval
    * @return Matrix2N Transition matrix
    */
-  static Matrix2N transitionFunction(double deltaT) {
+  static Matrix2N TransitionFunction(double deltaT) {
     // Construct the transition matrix for the WNOA factor
     Matrix2N F;
     F << kIdentity, deltaT * kIdentity, kZero, kIdentity;
@@ -346,7 +346,7 @@ class WNOAMotionFactor
    * @param deltaT Time interval
    * @return Matrix2N Jacobian block w.r.t. previous state
    */
-  static Matrix2N computeJacobianPrev(const std::pair<Pose, Velocity>& pv1,
+  static Matrix2N ComputeJacobianPrev(const std::pair<Pose, Velocity>& pv1,
                                       const std::pair<Pose, Velocity>& pv2,
                                       double deltaT) {
     // corresponds to F in (11.20) in SER
@@ -379,7 +379,7 @@ class WNOAMotionFactor
    * @param deltaT Time interval
    * @return Matrix2N Jacobian block w.r.t. next state
    */
-  static Matrix2N computeJacobianNext(const std::pair<Pose, Velocity>& pv1,
+  static Matrix2N ComputeJacobianNext(const std::pair<Pose, Velocity>& pv1,
                                       const std::pair<Pose, Velocity>& pv2,
                                       double deltaT) {
     // corresponds to E in (11.21) in SER
--- a/gtsam/nonlinear/WNOAFactor.h
+++ b/gtsam/nonlinear/WNOAFactor.h
@@ -84,6 +84,71 @@ class WNOAMotionFactor
   inline static const MatrixN kIdentity = MatrixN::Identity();
   inline static const MatrixN kZero = MatrixN::Zero();
 
+  /**
+   * @brief Compute relative pose/velocity terms shared by WNOA error/Jacobians.
+   *
+   * This helper computes the commonly-required relative pose and associated
+   * Jacobians: first compute `xi = Logmap(Between(p1, p2))` (with optional pose
+   * Jacobians), then optionally compute the derivative of the velocity residual
+   * with respect to `xi`.
+   *
+   * @param p1 Pose at time t_k
+   * @param p2 Pose at time t_{k+1}
+   * @param v2 Velocity at time t_{k+1}
+   * @param xi Output relative pose in tangent space
+   * @param invJr Output inverse right Jacobian from Logmap
+   * @param dxi_dT1 Optional output dxi/dT1
+   * @param dxi_dT2 Optional output dxi/dT2
+   * @param dvErr_dxi Optional output d(velocity error)/dxi
+   */
+  static void computeRelativePose(const Pose& p1, const Pose& p2,
+                                   const Velocity& v2, VectorN* xi,
+                                   MatrixN* invJr, MatrixN* dxi_dT1,
+                                   MatrixN* dxi_dT2, MatrixN* dvErr_dxi) {
+    assert(xi && "Output xi pointer must be valid");
+    assert(invJr && "Output invJr pointer must be valid");
+
+    const bool compute_dxi_dT1 = (dxi_dT1 != nullptr);
+    const bool compute_dxi_dT2 = (dxi_dT2 != nullptr);
+    const bool compute_dvErr_dxi = (dvErr_dxi != nullptr);
+
+    // Local variable for the relative pose in the tangent space
+    // Note that p1 = T(t_k), p2 = T(t_{k+1})
+    // compute xi = log(T_k^-1 T_{k+1})^check
+    if (compute_dxi_dT1 || compute_dxi_dT2) {
+      MatrixN dbetween_p1;
+      MatrixN dbetween_p2;
+      *xi = traits<Pose>::Logmap(
+          traits<Pose>::Between(p1, p2,
+                                compute_dxi_dT1 ? &dbetween_p1 : nullptr,
+                                compute_dxi_dT2 ? &dbetween_p2 : nullptr),
+          invJr);
+      if (compute_dxi_dT1) {
+        *dxi_dT1 = (*invJr) * dbetween_p1;
+      }
+      if (compute_dxi_dT2) {
+        *dxi_dT2 = (*invJr) * dbetween_p2;
+      }
+    } else {
+      *xi = traits<Pose>::Logmap(traits<Pose>::Between(p1, p2), invJr);
+    }
+
+    // Derivative of velocity error wrt xi
+    if (compute_dvErr_dxi) {
+      // Zero for vector spaces, use an approximation for Lie groups
+      if constexpr (std::is_same_v<typename traits<Pose>::structure_category,
+                                   vector_space_tag>) {
+        dvErr_dxi->setZero();
+      } else {
+        // For Lie groups
+        auto ad_v2 = Pose::adjointMap(v2);
+        auto ad_xi = Pose::adjointMap(*xi);
+        *dvErr_dxi = -ad_v2 / 2.0 -
+              (Pose::adjointMap(ad_xi * v2) + ad_xi * ad_v2) / 12.0;
+      }
+    }
+  }
+
  public:
   // Provide access to the Matrix& version of evaluateError:
   using Base::evaluateError;
@@ -93,13 +158,13 @@ class WNOAMotionFactor
    *
    * This constructor builds a factor connecting the pose and velocity keys
    * contained in `state_k` and `state_kp1`. The internal noise model is
-   * constructed from the provided diagonal PSD vector `q_psd_diag` and the timestep
-   * computed from the two state timestamps.
+   * constructed from the provided diagonal PSD vector `q_psd_diag` and the
+   * timestep computed from the two state timestamps.
    *
    * @param state_k StateData for time t_k (provides keys and timestamp).
    * @param state_kp1 StateData for time t_{k+1} (provides keys and timestamp).
-   * @param q_psd_diag Diagonal power spectral density vector used to form the process
-   * noise.
+   * @param q_psd_diag Diagonal power spectral density vector used to form the
+   * process noise.
    */
   WNOAMotionFactor(const StateData& state_k, const StateData& state_kp1,
                    const VectorN& q_psd_diag)
@@ -127,8 +192,8 @@ class WNOAMotionFactor
    */
   WNOAMotionFactor(Key poseKey0, Key velKey0, Key poseKey1, Key velKey1,
                    const double deltaT, const VectorN& q_psd_diag)
-      : Base(This::buildWNOANoiseModel(deltaT, q_psd_diag), poseKey0, velKey0, poseKey1,
-             velKey1),
+      : Base(This::buildWNOANoiseModel(deltaT, q_psd_diag), poseKey0, velKey0,
+             poseKey1, velKey1),
         deltaT_(deltaT) {}
 
   ~WNOAMotionFactor() override {}
@@ -139,10 +204,10 @@ class WNOAMotionFactor
   void print(
       const std::string& s = "",
       const KeyFormatter& keyFormatter = DefaultKeyFormatter) const override {
-    std::cout << s << "WNOAMotionFactor(" << keyFormatter(this->poseKey0())
-              << "," << keyFormatter(this->velKey0()) << ","
-              << keyFormatter(this->poseKey1()) << ","
-              << keyFormatter(this->velKey1()) << ")\n";
+    std::cout << s << "WNOAMotionFactor(" << keyFormatter(this->key1())
+              << "," << keyFormatter(this->key2()) << ","
+              << keyFormatter(this->key3()) << ","
+              << keyFormatter(this->key4()) << ")\n";
     this->noiseModel_->print("  noise model: ");
   }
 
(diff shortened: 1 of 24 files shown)


dellaert:
Awesome! Copilot had some good comments, and I started the CI. I will look at this when both of these things are green :-). I might get to this during the week, otherwise during the weekend. I have a lot of teaching stuff.

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,412 @@
...
+#include <gtsam/geometry/Point3.h>
+#include <gtsam/geometry/Pose2.h>
+#include <gtsam/geometry/Pose3.h>
+#include <gtsam/nonlinear/NonlinearFactor.h>
+#include <gtsam/nonlinear/WNOAStateData.h>
+
+#include <cassert>
+
+#pragma once
+using namespace std;
```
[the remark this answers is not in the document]
holmesco: Removed using namespace std.

gtsam/nonlinear/WNOAInterpFactor.h
```diff
@@ -0,0 +1,1035 @@
...
+        auto [left, right] = interp_to_borders.at(interp);  // get borders
+        outer_key_set.insert(left.pose);                    // add border keys
+        outer_key_set.insert(left.vel);
+        outer_key_set.insert(right.pose);
+        outer_key_set.insert(right.vel);
+      }
+    }
+    // Convert to key vector (from set)
+    keys_ = KeyVector(outer_key_set.begin(), outer_key_set.end());
+
```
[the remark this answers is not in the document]
holmesco: @dellaert, does the ordering of the keys really make a difference? We use a set here just to avoid duplicate keys when building the wrapper factor.

gtsam/nonlinear/WNOAInterpFactor.h
```diff
@@ -0,0 +1,1035 @@
...
+      std::vector<Matrix> H(8);
+      if (InterpJacobians) {
+        result = interpolator_.interpolatePoseAndVelocity(
+            state_left, state_right, interp_state.time, &H, nullptr, nullptr,
+            lambda_psi_pre_comp.at(interp_ind));
+      } else {
+        result = interpolator_.interpolatePoseAndVelocity(
+            state_left, state_right, interp_state.time, nullptr, nullptr,
+            nullptr, lambda_psi_pre_comp.at(interp_ind));
+      }
```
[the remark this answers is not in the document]
holmesco: @SvenLilge, not sure if this is a real issue or not. I remember that we switched to vectors to avoid hash maps (like the suggested fix) for speed, but I guess it's possible to get misalignment if interp_to_borders does something weird?

gtsam/nonlinear/WNOAInterpolator.cpp
```diff
@@ -0,0 +1,676 @@
...
+  // interpolate
+  if (t_tau == t_k) {
+    // t_tau equal to left boundary time
+    return interpolateBoundaryLeft(poseVel_k, H, mainSolveMarginalMatrix,
+                                   covarianceOut);
+
+  } else if (t_tau == t_kp1) {
+    // t_tau equal to right boundary time
+    return interpolateBoundaryRight(poseVel_kp1, H, mainSolveMarginalMatrix,
+                                    covarianceOut);
```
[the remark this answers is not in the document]
holmesco: These deal with pathological cases where we are intepolating at the exact time of a boundary. @qetuo1098 do we actually need these functions or can they be generalized from the extrapolation versus interpolation case?

dellaert:
I see CI fails. Please note that your code should *only* use boost for serialization, and this use should be guarded. Please inspect other headers for how to do this. For everything else we switch to c++17 std::shared etc.

gtsam/nonlinear/WNOAInterpFactor.h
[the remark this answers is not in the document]
dellaert: It depends on the context. If you're talking about elimination orderings, it matters a lot. If you're just doing deduplication, probably not.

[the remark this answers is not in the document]
holmesco: This is just to avoid key duplication when we initialize the factor. I think this comment can be safely ignored.

dellaert:
Same question: ready for my detailed review?

dellaert:
Seems CI fails on an unused variable. We run with no warnings :-) Do a `make check` before pushing - saves some CO(2) !

holmesco:
I don't think `make check` is catching whatever is causing these errors. It seems to pass even with the commit that was causing errors in the CI. The PR should be ready for your review by EOD. Just checking that our variables are named properly (shouldn't copilot review flag these?)

holmesco:
Fixed naming conventions in files. @dellaert, should be ready for your review!

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+   * @brief Construct a WNOA factor given explicit keys and timestep.
+   *
+   * @param key1 Pose key at t_k
+   * @param key2 Velocity key at t_k
+   * @param key3 Pose key at t_{k+1}
+   * @param key4 Velocity key at t_{k+1}
+   * @param delta_t Time interval t_{k+1} - t_k (must be > 0)
+   * @param Q Diagonal PSD vector used to form the process noise.
+   */
+  WNOAMotionFactor(Key key1, Key key2, Key key3, Key key4, const double delta_t,
```
dellaert: Nit: especially when these things are surfaced in Python, it would be nice to have more descriptive names for the key variables.

dellaert: Nit 2: `deltaT` ? or `dt`.

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+                   const VectorN& Q)
+      : Base(This::buildWNOANoiseModel(delta_t, Q), key1, key2, key3, key4),
+        delta_t_(delta_t) {}
+
+  ~WNOAMotionFactor() override {}
+  /** implement functions needed for Testable */
```
dellaert: Prefer Doxygen regions as in many other classes. 
```
  /// @name Testable
  /// @{
  /// @}
```

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+                       OptionalMatrixType Hv1, OptionalMatrixType Hp2,
+                       OptionalMatrixType Hv2) const override {
+    // Local variable for the relative pose in the tangent space
+    // Note that p1 = T(t_k), p2 = T(t_{k+1})
+    //  compute xi = log(T_k^-1 T_{k+1})^check
+    VectorN xi;
+    // Jacobians
+    MatrixN dxi_dT1;
+    MatrixN dxi_dT2;
+    MatrixN right_jac_inv;
```
dellaert: Rename to `invJr` ?

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+  using Velocity = typename gtsam::traits<Pose>::TangentVector;
+  using MatrixN = Eigen::Matrix<double, dim, dim>;
+  using VectorN = Eigen::Matrix<double, dim, 1>;
+  using Matrix2N = Eigen::Matrix<double, 2 * dim, 2 * dim>;
+  using Vector2N = Eigen::Matrix<double, 2 * dim, 1>;
+  using MatrixNx2N = Eigen::Matrix<double, dim, 2 * dim>;
+  typedef NoiseModelFactorN<Pose, Velocity, Pose, Velocity> Base;
+  typedef WNOAMotionFactor This;
+  // Time between the two states
+  double delta_t_;
```
dellaert: deltaT_ or dt_

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+    MatrixN dvErr_dxi;
+    if (Hp1 || Hp2) {
+      // Derivative of velocity error wrt xi
+      // Zero for vector spaces, use an approximation for Lie groups
+      if constexpr (std::is_same_v<typename traits<Pose>::structure_category,
+                                   vector_space_tag>) {
+        dvErr_dxi.setZero();
+      } else {
+        // For Lie groups
+        dvErr_dxi = -Pose::adjointMap(v2) / 2.0 -
```
dellaert: Hardcore :-). You're calling Pose::adjointMap(v2)  and Pose::adjointMap(xi)  twice though, so maybe have two temp variables ad_v2 and ad_xi ?

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+   * prior
+   *
+   * @param timestep Time interval over which to compute covariance
+   * @param Q Diagonal PSD vector
+   * @return Matrix2N Process covariance
+   */
+  static Matrix2N buildWNOACovariance(double timestep, const VectorN& Q) {
+    //
+    Matrix2N covariance;
+    MatrixN Q_diag = Q.asDiagonal();
```
dellaert: might want to `auto` so you get Eigen optimization?

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+   *
+   * @param timestep Time interval
+   * @param Q Diagonal PSD vector
+   * @return Matrix2N Inverse process covariance matrix
+   */
+  static Matrix2N buildInverseWNOACovariance(double timestep,
+                                             const VectorN& Q) {
+    // construct the inverse covariance matrix for the WNOA factor
+    Matrix2N inverse_covariance;
+    MatrixN Q_inv_diag = Q.cwiseInverse().asDiagonal();
```
dellaert: same

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+  /**
+   * @brief Build the continuous-time WNOA discretized process covariance.
+   *
+   * Returns the 2N x 2N covariance matrix for a WNOA prior discretized over
+   * `timestep` using the diagonal PSD `Q`.
+   * See (11.7) in (Barfoot, 2024) for the derivation of covariance for the WNOA
+   * prior
+   *
+   * @param timestep Time interval over which to compute covariance
+   * @param Q Diagonal PSD vector
```
dellaert: Not sure I recognize Q as a vector. Consider using a lowercase variable.

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+   * Computes the 2N x 2N Jacobian block that maps perturbations in the
+   * previous bordering state (pose_k, vel_k) to perturbations in the
+   * interpolated discrete residual.
+   *
+   * @param pv1 Pair (pose_k, vel_k)
+   * @param pv2 Pair (pose_kp1, vel_kp1)
+   * @param delta_t Time interval
+   * @return Matrix2N Jacobian block w.r.t. previous state
+   */
+  static Matrix2N computeJacobianPrev(const std::pair<Pose, Velocity>& pv1,
```
dellaert, line 372: I've seen this code before :-) Consider consolidating with evaluateError, one calling the other or both calling a third function.

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+   * Computes the 2N x 2N Jacobian block that maps perturbations in the
+   * next bordering state (pose_kp1, vel_kp1) to perturbations in the
+   * interpolated discrete residual.
+   *
+   * @param pv1 Pair (pose_k, vel_k)
+   * @param pv2 Pair (pose_kp1, vel_kp1)
+   * @param delta_t Time interval
+   * @return Matrix2N Jacobian block w.r.t. next state
+   */
+  static Matrix2N computeJacobianNext(const std::pair<Pose, Velocity>& pv1,
```
dellaert, line 406: same comment?

gtsam/nonlinear/WNOAFactor.h
```diff
@@ -0,0 +1,405 @@
...
+    return -1 * E;
+  }
+};
+
+// Make factors testable
+template <class Pose>
+struct traits<WNOAMotionFactor<Pose>>
+    : public Testable<WNOAMotionFactor<Pose>> {};
+
+}  // namespace gtsam
```
dellaert: Check EOLs at end of file(s).

dellaert (commented):
Awesome! 
First batch of comments. Many comments apply to all new factors, so do one more round then re-request?

gtsam/nonlinear/WNOAInterpolator.cpp
```diff
@@ -0,0 +1,677 @@
...
+           mainSolveMarginalMatrix->cols() == 2 * dim);
+    Matrix2N Sigma = covarianceFunction_(t_diff, Q_psd_);
+    // (11.5) in (Barfoot 2024)
+    *covarianceOut = Sigma + Psi * *mainSolveMarginalMatrix * Psi.transpose();
+  }
+  return PoseVel{T_tau, varpi_tau};
+}
+
+template <typename PoseType>
+typename Interpolator<PoseType>::PoseVel
```
dellaert, line 205: This function is way too large :-)
Also, many variables are defined that are only needed in the H-path, and the CI failure might be (so the AI tells me) a real issue. It suggests you should have
```
if (H && !localGlobalStateJacsPreComp) {
  throw std::runtime_error("Requested Jacobians but localGlobalStateJacsPreComp is null");
}
```

gtsam/nonlinear/tests/testWNOAInterpFactor.cpp
```diff
@@ -0,0 +1,838 @@
...
+Point3 p2_p3 = p0_p3 + 2 * timestep * v0_p3;
+Vector3 v2_p3 = v0_p3;
+
+/**** SE3 Test Variables *****/
+// Define First Pose  with a general velocity
+static Pose3 p0_se3 = Pose3::Expmap(Vector6(0.5, 0.0, 0.0, 0.0, 0.0, 0.0));
+static Vector6 v0_se3(1, 0.0, 0.5, 0.1, 0.0, 0.0);
+// Define Second Pose with same velocity (to get an error)
+static Pose3 p1_se3 = p0_se3.expmap(timestep * v0_se3);
+static Vector6 v1_se3 = v0_se3;
```
dellaert: Lots of unused variables in this file according to one of the CI runs.

holmesco:
@dellaert, it seems like our tests are failing when the Cayley map is used for rotations. Specifically, when we define a between factor, the `Rot3::CayleyChart::Local` function fails because it does not support Jacobians (which we need to properly map the measurement noise).

I can work around this in our tests, but it seems like a fundamental issue. Are there other error functions or factors that do not support Jacobians?

dellaert:
> @dellaert, it seems like our tests are failing when the Cayley map is used for rotations. Specifically, when we define a between factor, the `Rot3::CayleyChart::Local` function fails because it does not support Jacobians (which we need to properly map the measurement noise).
> 
> I can work around this in our tests, but it seems like a fundamental issue. Are there other error functions or factors that do not support Jacobians?

This is probably because you are calling Retract somewhere? You could explicitly call expmap or Expmap. If that’s not possible, tell me where to look :-)

dellaert:
Would love to move this along- as we are preparing to release 4.3. Was my last comment informative?

gtsam/nonlinear/WNOAFactor.h
holmesco: Consolidated common function calls to one new function `computeRelativePose`

gtsam/nonlinear/WNOAFactor.h
holmesco: auto causes eigen to not interpret the matrix correctly. copilot recommended `const MatrixN`

gtsam/nonlinear/WNOAFactor.h
holmesco: Fixed this. This code now resides in `computeRelativePose`

gtsam/nonlinear/WNOAInterpolator.cpp
holmesco: I don't think this change would do what we want: `localGlobalStateJacsPreComp` is only available when we precompute some matrices, but there are cases where we need the Jacobians and this variable is not set.
I split this function into 5 intermediate functions.

holmesco:
> > @dellaert, it seems like our tests are failing when the Cayley map is used for rotations. Specifically, when we define a between factor, the `Rot3::CayleyChart::Local` function fails because it does not support Jacobians (which we need to properly map the measurement noise).
> > I can work around this in our tests, but it seems like a fundamental issue. Are there other error functions or factors that do not support Jacobians?
> 
> This is probably because you are calling Retract somewhere? You could explicitly call expmap or Expmap. If that’s not possible, tell me where to look :-)

As far as i can tell, we are not using Retract. Our factor is calling the `evaluateError` of a `BetweenFactor` defined on a `Pose3` type. I can reproduce the CI error with the following snippet
```cpp
// Model
  const auto model = noiseModel::Diagonal::Sigmas(Vector6::Ones());
  // Set up values
  const Pose3 p0 = Pose3::Expmap(Vector6(0.0, 0.0, 0.0, 0.0, 0.0, 0.0));
  const Pose3 p1 = Pose3::Expmap(Vector6(0.5, 0.0, 0.0, 0.0, 0.0, 0.0));
  // construct between factor
  const Pose3 p01 = p0.inverse().compose(p1);
  const auto between_factor =
  std::make_shared<BetweenFactor<Pose3>>(P(0), P(1), p01, model);
  // Make values
  Values values;
  values.insert(P(0), p0);
  values.insert(P(1), p1);
  // // Test unwhitened error 
  std::vector<Matrix> jacobians(2);
  const Vector error = between_factor->unwhitenedError(values, &jacobians);
  // test evaluate error with Jacobians
  Matrix jacobianPose0;
  Matrix jacobianPose1;
  const Vector errorFromEvaluateError =
      between_factor->evaluateError(p0_se3, p1_se3, jacobianPose0, jacobianPose1);
  ```
  where both `unwhitenedError` and `evaluateError` cause the Cayley map error under these build conditions:
   ```bash
-DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS=-w -DGTSAM_BUILD_TESTS=ON -DGTSAM_BUILD_UNSTABLE=ON -DGTSAM_WITH_TBB=OFF -DGTSAM_BUILD_EXAMPLES_ALWAYS=OFF -DGTSAM_ALLOW_DEPRECATED_SINCE_V43=OFF -DGTSAM_USE_QUATERNIONS=OFF -DGTSAM_ROT3_EXPMAP=OFF -DGTSAM_POSE3_EXPMAP=OFF -DGTSAM_USE_SYSTEM_EIGEN=OFF -DGTSAM_USE_SYSTEM_METIS=OFF -DGTSAM_USE_BOOST_FEATURES=OFF -DGTSAM_ENABLE_BOOST_SERIALIZATION=OFF -DGTSAM_BUILD_WITH_MARCH_NATIVE=OFF -DGTSAM_SINGLE_TEST_EXE=OFF -DGTSAM_ENABLE_GEOGRAPHICLIB=OFF
  ```

**Comment by dellaert** [comment]
I think the best way is to maybe only allow Gaussian processes if we don't have the Cayley option turned on...

PS please merge in develop - has evolved :-)

**Comment by holmesco** [comment]
Ok, when Cayley option is enabled i can skip the tests that fail. I can also add a compile-time error if our factor is templated on SO3, but Cayley is enabled.

File: gtsam/nonlinear/WNOAFactor.h (hunk 14 above)
Comment (holmesco) [reply gtsam/nonlinear/WNOAFactor.h]: consolidated in common function

**Comment by dellaert** [comment]
@holmesco I checked out your branch and made a PR against it on your fork. It was mainly naming conventions and some tightening of the APIs, especially when many arguments were passed. So I created this new struct, StateJacobians that makes that a little bit cleaner and eliminates a lot of boilerplate code. 

Let me know what you think.

**Comment by dellaert** [comment]
@holmesco I added one more PR on your fork, eliminating some code duplication. There is two more issues to take care of on your end:
- There is still a CI issue on Windows and Ubuntu. I don't think these have to do with my changes. 
- The naming convention for the classes and the files is problematic. We largely follow Google style, and both factors and files should be `WnoaXyz` not `WNOAXyz`. See e.g., `ImuFactor`.

**Comment by dellaert** [comment]
Almost there :-)

File: gtsam/nonlinear/tests/testWnoaInterpFactor.cpp [hunk 18]
```diff
@@ -821,10 +821,12 @@ TEST(WNOAInterp, SE3InterpGraph) {
 
   // Define nominal factors
   const auto model = noiseModel::Diagonal::Sigmas(Vector6::Ones());
```
dellaert: Small nit: we have noiseModel::Isotropic and noiseModel::Unit (applicable here).maybe change that in next PR.

dellaert (approved):
Awesome. If CI passes I’ll merge.

dellaert:
Very weird that tests fail on windows.

holmesco:
Test failure was due to global variable definitions in the test files (bad practice, I know). Added test fixtures and proper scoping. Windows tests are now passing locally.

dellaert:
@holmesco I noticed you skipped C.I. Could you add a dummy commits to re-run the CI? Maybe format a file with clang-format Google's style. If that is not the case yet for that test file ;-) If you have not formatted with Clang format in general, reformat all the change files in this PR :-).

holmesco:
> @holmesco I noticed you skipped C.I. Could you add a dummy commits to re-run the CI? Maybe format a file with clang-format Google's style. If that is not the case yet for that test file ;-) If you have not formatted with Clang format in general, reformat all the change files in this PR :-).

Woops, should run now. Also we have been formatting with google's style.

holmesco:
Looks like the windows CI had a network issue when downloading boost. Can we retrigger the CI?

dellaert:
Done and merged!

PR #2415, opened 2026-02-21
Weight update in Gnc GM through static function in LossFunctions

hkhanuja:
In this branch, I add a static function to LossFunctions for the GemanMcClure loss so that GNC can use the parameter convention and naming from the loss functions. Static helper enables us to call the function without actually creating an object.
To reduce the calls to the std::sqrt function, the static function takes the squared value of the error and the scale parameter.
A scale factor squared in also added in the constructor since c^2 is used in a lot of member functions.
The weight member function now calls the static helper function, while it takes the error as input, it passes e^2 and c^2 to the static helper.
Unit tests pass locally.

### DIFF
--- a/gtsam/linear/LossFunctions.h
+++ b/gtsam/linear/LossFunctions.h
@@ -373,9 +373,21 @@ class GTSAM_EXPORT GemanMcClure : public Base {
   double loss(double distance) const override;
   void print(const std::string &s) const override;
   bool equals(const Base &expected, double tol = 1e-8) const override;
-  static double Weight(double distance2, double c2);
   static shared_ptr Create(double k, const ReweightScheme reweight = Block);
   double modelParameter() const { return c_; }
+  /** @brief A static helper function to compute the Geman-McClure robust weight.
+   * The static function takes the squared value of the residual and the scale parameter.
+   * The weight member function now calls the this function, while the member function takes the residual as input, 
+   * it passes r² and c² to the static helper.
+   * 
+   *w(d², c²) = \phi(d)/d = c⁴/(c²+d²)²
+   * 
+   * 
+   * @param distance2 Squared residual magnitude.
+   * @param c2 Squared scale parameter.
+   * @return Weight w(x) in (0, 1]
+   */
+  static double Weight(double distance2, double c2);
 
  protected:
   double c_;
--- a/gtsam/linear/LossFunctions.h
+++ b/gtsam/linear/LossFunctions.h
@@ -373,6 +373,7 @@ class GTSAM_EXPORT GemanMcClure : public Base {
   double loss(double distance) const override;
   void print(const std::string &s) const override;
   bool equals(const Base &expected, double tol = 1e-8) const override;
+  static double Weight(double distance, double c);
   static shared_ptr Create(double k, const ReweightScheme reweight = Block);
   double modelParameter() const { return c_; }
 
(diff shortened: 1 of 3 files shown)


dellaert (commented):
the noise model should now call the static function, and after that the unit tests should pass. Please edit PR comment with info on unit tests passing.

dellaert (changes requested):
Issues with sqrt then square need to be resolved.

gtsam/linear/LossFunctions.h
```diff
@@ -373,11 +373,13 @@ class GTSAM_EXPORT GemanMcClure : public Base {
   double loss(double distance) const override;
   void print(const std::string &s) const override;
   bool equals(const Base &expected, double tol = 1e-8) const override;
+  static double Weight(double distance2, double c2);
```
dellaert: This needs a doxygen block, e.g. this one generated by ChatGPT
```
/**
 * @brief Geman–McClure robust weight for a squared residual.
 *
 * Computes the M-estimator weight used to downweight outliers when the input is
 * the **squared** residual magnitude.
 *
 * Let \(d^2\) be the squared residual (e.g., \(d^2 = \|r\|^2\)) and \(c^2\) the
 * squared scale (tuning) parameter. The weight is
 * \f[
 *   w(d^2; c^2) = \frac{(c^2)^2}{(c^2 + d^2)^2} = \left(\frac{c^2}{c^2 + d^2}\right)^2 .
 * \f]
 *
 * Properties: \(w(0)=1\) and \(w(d^2)\to 0\) as \(d^2\to\infty\).
 *
 * @param distance2 Squared residual magnitude \(d^2 \ge 0\).
 * @param c2 Squared scale parameter \(c^2 > 0\).
 * @return Weight \(w \in (0, 1]\).
 */
```

gtsam/linear/LossFunctions.h
```diff
@@ -373,11 +373,13 @@ class GTSAM_EXPORT GemanMcClure : public Base {
...
   static shared_ptr Create(double k, const ReweightScheme reweight = Block);
   double modelParameter() const { return c_; }
 
  protected:
```
dellaert, line 392: put the static method here, right before protected

gtsam/nonlinear/GncOptimizer.h
```diff
@@ -489,8 +490,9 @@ class GncOptimizer {
         for (size_t k = 0; k < nfg_.size(); k++) {
           if (needsWeightUpdate(factorTypes_[k])) {
             double u2_k = nfg_[k]->error(currentEstimate);  // squared (and whitened) residual
-            weights[k] = std::pow(
-                (mu * barcSq_[k]) / (u2_k + mu * barcSq_[k]), 2);
+            const double distance2_k = 2.0 * u2_k;
+            const double c2_k = 2.0 * mu * barcSq_[k];
+            weights[k] = noiseModel::mEstimator::GemanMcClure::Weight(distance2_k, c2_k);
```
dellaert: Factors 2 cancel? Maybe just
```
weights[k] = noiseModel::mEstimator::GemanMcClure::Weight(u2_k, mu * barcSq_[k]);
```

hkhanuja: yes, removed.

gtsam/linear/LossFunctions.h
hkhanuja: added

gtsam/linear/LossFunctions.h
hkhanuja: moved

dellaert (approved):
Thanks!

PR #2416, opened 2026-02-22
Extended Pose SE_k(3)

dellaert:
A major refactor that brings in $SE_k(3)\doteq SO(3) \ltimes (\mathbb{R}^3)^k$ and eliminates a lot of duplicate code in `Pose3` and `NavState`. In addition,  $SE_k(3)$ can now be used in invariant filter implementations. A selection with $k={2, 3, 4, 6}$ is exposed in the wrapper as well.

### What changed

- Added a new generic `ExtendedPose3<K, Derived>` in `geometry`, supporting:
  - fixed `K >= 1`
  - dynamic `K = Eigen::Dynamic` (runtime `k` constructor)
- Implemented exactly the required `Manifold` / `LieGroup` / `MatrixLieGroup` surface.
- Used CRTP (`Derived` defaulted) so derived classes keep concrete return types (`compose`, `inverse`, `Expmap`, etc.) without heavy wrappers.
- Refactored:
  - `Pose3` -> `ExtendedPose3<1, Pose3>`
  - `NavState` -> `ExtendedPose3<2, NavState>`
  - Removed duplicated implementations now provided by base.
- Updated docs:
  - `ExtendedPose3` notebook with Lie-group math intro, Colab preamble tags, and Python examples
  - updated `.github/copilot-instructions.md` notebook conventions.

### Why this is better

- Single implementation for `SE_k(3)` math across `Pose3`, `NavState`, and future `k`.

#### C++
- `make -j6 testExtendedPose3.run` ✅
- `make -j6 testPose3.run` ✅
- `make -j6 testNavState.run` ✅

#### Wrapper/build
- `make -j6 install` ✅
- `make -j6 python-install` ✅
- `make -j6 python-test` ✅


dellaert:
Note: perf regression is because I had not rebased on develop before pushing. Will address review comments then re-run benchmark.

dellaert:
@ProfFan I re-requested a benchmark but I don’t know where it appears. if it does replace the earlier comment, it would be good to know when that was done with a time stamp or something. As it stands, it says three hours ago, which is the first one I think. But the actions showed that the benchmarks have finished.

dellaert:
Rather than replacing a comment, I think it would be better to _add_ a comment “Benchmarks ran at this time… here are the results:”

dellaert:
\bench

ProfFan:
image

ProfFan:
Result is cached, errrrgh, I should add a command to force re-bench

gtsam/geometry/tests/testExtendedPose3.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <gtsam/base/Lie.h>
+#include <gtsam/base/TestableAssertions.h>
+#include <gtsam/base/lieProxies.h>
+#include <gtsam/base/numericalDerivative.h>
+#include <gtsam/geometry/ExtendedPose3.h>
+
+using namespace gtsam;
+
+using ExtendedPose33 = ExtendedPose3<3>;
+using ExtendedPose3d = ExtendedPose3<Eigen::Dynamic>;
```
varunagrawal, line 29: Do we need tests for this particular case?

gtsam/geometry/ExtendedPose3-inl.h
```diff
@@ -0,0 +1,458 @@
...
+  if constexpr (dimension == Eigen::Dynamic) {
+    H->setZero(d, d);
+  } else {
+    (void)d;
+    H->setZero();
+  }
+}
+
+template <int K, class Derived>
+template <int K_, typename>
```
varunagrawal: What's going on with this template?

gtsam/geometry/ExtendedPose3-inl.h
```diff
@@ -0,0 +1,458 @@
...
+ExtendedPose3<K, Derived>::ExtendedPose3()
+    : R_(Rot3::Identity()), t_(Matrix3K::Zero()) {}
+
+template <int K, class Derived>
+template <int K_, typename>
```
varunagrawal: Same here. They don't seem to be used.

gtsam/geometry/ExtendedPose3.h
```diff
@@ -0,0 +1,423 @@
...
+ *
+ * The manifold dimension is 3+3k and the homogeneous matrix size is 3+k.
+ * Template parameter K can be fixed (K >= 1) or Eigen::Dynamic.
+ */
+template <int K, class Derived>
+class ExtendedPose3
+    : public MatrixLieGroup<std::conditional_t<std::is_void_v<Derived>,
+                                               ExtendedPose3<K, void>, Derived>,
+                            (K == Eigen::Dynamic) ? Eigen::Dynamic : 3 + 3 * K,
+                            (K == Eigen::Dynamic) ? Eigen::Dynamic : 3 + K> {
```
varunagrawal, line 53: I am questioning the need for Eigen::Dynamic since $SE_k(3)$ is currently limited to 4 I believe? Maybe I am missing something.

gtsam/geometry/ExtendedPose3.h
```diff
@@ -0,0 +1,423 @@
...
+  /// @name Constructors
+  /// @{
+
+  /**
+   * Construct a fixed-size identity element.
+   *
+   * For fixed K, this creates R=I and x_i=0 for i=1..k.
+   * The manifold dimension is 3+3k and matrix size is (3+k)x(3+k).
+   */
+  template <int K_ = K, typename = IsFixed<K_>>
```
varunagrawal, line 98: Any reason we don't do the following?
```cpp
template <typename = IsFixed<K>>
```

dellaert:
I don't trust the Mac benchmarks - probably was run on a different machine from cached, @ProfFan and I discussed a fix to benchmark flow (ps are we running that in release!? numbers below are *much* better, on an M1, with TBB). Here are my results:

### New:
| Dataset | Legacy (Cholesky) s | New (Solver) s | Speedup |
| --- | --- | --- | --- |
| dubrovnik-16-22106-pre.txt | 1.424 | 0.873 | 1.631x |
| dubrovnik-88-64298-pre.txt | 9.052 | 4.845 | 1.868x |
| dubrovnik-135-90642-pre.txt | 10.072 | 5.755 | 1.750x |

### Old:
| Dataset | Legacy (Cholesky) s | New (Solver) s | Speedup |
| --- | --- | --- | --- |
| dubrovnik-16-22106-pre.txt | 1.248 | 0.676 | 1.847x |
| dubrovnik-88-64298-pre.txt | 9.812 | 4.035 | 2.432x |
| dubrovnik-135-90642-pre.txt | 11.023 | 6.256 | 1.762x |

So I think there is no regression and this can be merged. The Pose3-specific Expmap was a bit faster than without, although we should have much stronger benchmarks to really conclude that...

gtsam/geometry/ExtendedPose3.h
```diff
@@ -0,0 +1,423 @@
...
+   */
+  ExtendedPose3(const Rot3& R, const Matrix3K& x);
+
+  /**
+   * Construct from homogeneous matrix representation.
+   *
+   * @param T Homogeneous matrix in R^((3+k)x(3+k)).
+   * Top-left 3x3 is R, top-right 3xk stores x_1..x_k.
+   */
+  explicit ExtendedPose3(const LieAlgebra& T);
```
varunagrawal: Is this the LieAlgebra? Using the analogy for SO(3), the LieAlgebra is the set of all 3x3 skew-symmetric matrices of angular velocities, which is not orthogonal.
You are doing the following in the implementation:
```cpp
R_ = Rot3(T.template block<3, 3>(0, 0));
t_ = T.block(0, 3, 3, n - 3);
```
which is clearly not correct.

We should rename this to `PoseMatrix` or something like that.

gtsam/navigation/tests/testNavState.cpp
```diff
@@ -574,14 +583,42 @@ TEST(NavState, Adjoint_compose_full) {
...
+  Vector9 x;
+  x << 0.1, 0.1, 0.1, 0.4, 0.2, 0.8, 0.4, 0.2, 0.8;
   NavState expected = T1 * NavState::Expmap(x) * T2;
   Vector y = T2.inverse().Adjoint(x);
   NavState actual = T1 * T2 * NavState::Expmap(y);
   EXPECT(assert_equal(expected, actual, 1e-6));
 }
 
+/* ************************************************************************* */
+TEST(NavState, expmaps_galore_full) {
```
varunagrawal: Can we rename the test to `ExpmapsGaloreFull`? That way CppUnitLite will display it as `NavStateExpmapsGaloreFull`.

gtsam/navigation/NavState.cpp
```diff
@@ -153,218 +118,60 @@ void NavState::print(const std::string& s) const {
...
-
-  return Ad * xi_b;
-}
+  if (H_state) {
+    //TODO(Varun)
+    // const auto w_T_hat = skewSymmetric(AdTx.head<3>()),
+    //            v_T_hat = skewSymmetric(AdTx.segment<3>(3)),
+    //            a_T_hat = skewSymmetric(AdTx.tail<3>());
+    //   *H_state << w_T_hat, v_T_hat,  //
+    //       /*  */ v_T_hat, Z_3x3;
```
varunagrawal: What is happening here? We went from an implementation to a TODO?

varunagrawal (commented):
Did a first pass and there is some weird naming that raised some flags.

I am really liking the de-duplication induced, however, I am a bit skeptical about the need to support `Eigen::Dynamic` since I haven't found any use of $SE_k(3)$ with $k > 6$.

Once I get a response on the naming comments, I can finish looking at the rest of the `gtsam/geometry/ExtendedPose3-inl.h` file and approve.

gtsam/geometry/tests/testExtendedPose3.cpp
dellaert: Yes

gtsam/geometry/ExtendedPose3.h
dellaert: No, it can be anything, and I want it to be available at run time. For example, we can model a varying number of landmarks in an egocentric, invariant filter.

gtsam/geometry/ExtendedPose3.h
dellaert: You're absolutely right. Will fix.

gtsam/navigation/tests/testNavState.cpp
dellaert: Done

varunagrawal (approved):
I believe this looks good.

PR #2446, opened 2026-03-02
Add SphericalCamera Python bindings

inuex35:
Hi, I'm not familiar with the project's contribution guidelines or conventions, so please let me know if there's anything I should change regarding code style, naming, testing, or the overall approach. Any feedback is welcome.

I have been working on enabling Structure-from-Motion (SfM) for 360-degree (omnidirectional) cameras using GTSAM and https://github.com/inuex35/gtsfm. With the SphericalCamera Python bindings added in this PR, combined with improvements we have made on the GTSFM side, it is now possible to run a full SfM pipeline for 360-degree cameras — from bearing-vector
 measurements through to bundle adjustment — entirely in Python.

  - Add full Python bindings for SphericalCamera, enabling bearing-vector-based (Unit3) projection, back-projection, and triangulation from Python
  - Generalize GeneralSFMFactor to support non-Point2 measurement types (e.g., Unit3 for SphericalCamera) by introducing MeasurementTraits.h and templating on CAMERA::Measurement
  - Add SphericalCamera support to SmartFactorBase, SmartProjectionFactor, SmartProjectionRigFactor, PriorFactor, NonlinearEquality, and Values


gtsam/geometry/MeasurementTraits.h
```diff
@@ -0,0 +1,81 @@
...
+};
+
+/**
+ * @brief Specialization for Unit3 measurements
+ *
+ * Unit3 requires errorVector instead of Local to get correct Jacobians.
+ */
+template <>
+struct MeasurementErrorHelper<Unit3> {
+  static constexpr int Dim = traits<Unit3>::dimension;
```
dellaert: This will be 2, just like for a Point2, no?

gtsam/geometry/MeasurementTraits.h
```diff
@@ -0,0 +1,81 @@
...
+  using VectorType = Eigen::Matrix<double, Dim, 1>;
+
+  static VectorType Evaluate(const Unit3& measured, const Unit3& predicted,
+                             OptionalJacobian<Dim, Dim> H_predicted) {
+    return measured.errorVector(predicted, {}, H_predicted);
```
dellaert: Why can't we use Local?

gtsam/slam/GeneralSFMFactor.h
```diff
@@ -144,17 +166,25 @@ class GeneralSFMFactor: public NoiseModelFactorN<CAMERA, LANDMARK> {
...
-      const CAMERA& camera = values.atRef<CAMERA>(key1);
-      const LANDMARK& point = values.atRef<LANDMARK>(key2);
-      b = measured() - camera.project2(point, H1, H2);
+      const CAMERA& camera = values.at<CAMERA>(key1);
+      const LANDMARK& point = values.at<LANDMARK>(key2);
+      Measurement predicted = camera.project2(point, &Dcamera, &Dlandmark);
+
+      MeasurementJacobian localJacobianStorage;
+      OptionalJacobian<ZDim, ZDim> localJac(&localJacobianStorage);
+      b = -Vector(internal::MeasurementErrorHelper<Measurement>::Evaluate(
```
dellaert: I'm sort of hoping we can just use local and it will be just as good. Then we can get rid of the traits mechanism.

dellaert (commented):
Nice !!!
I'm not sure we need the traits mechanism though. Unit3 is a 2D manifold and it has a Local method in its traits already. And the local method for Point2 is just subtraction. And because the dimension of Unit3 is also 2, there is also no need for dim Z.

inuex35:
Hi, you are right.

The updated version works fine, could you review again?

This is the result with my gtsfm.

image

dellaert (commented):
This looks more like it! Love it :-)

I’ll run CI to see if it compiles across platforms, then approve.

Thanks !!

dellaert (approved):
All required CI passed, so I’ll merge. Thank you again.

PR #2447, opened 2026-03-02
Add PseudorangeFactorArm and DifferentialPseudorangeFactorArm

inuex35:
Description:
This PR introduces new factors to the navigation module that support lever arm compensation for GNSS sensors.

Previously, PseudorangeFactor required the antenna's absolute ECEF position. The new PseudorangeFactorArm factors accept a Pose3 (body pose in the navigation frame) and a static lever arm vector in the body frame. This allows the factor to dynamically compute the antenna position based on the body's rotation.

### DIFF
--- a/gtsam/navigation/PseudorangeFactor.cpp
+++ b/gtsam/navigation/PseudorangeFactor.cpp
@@ -183,12 +183,12 @@ Vector PseudorangeFactorArm::evaluateError(
 
   // Compute associated derivatives:
   if (H_nTb) {
+    H_nTb->resize(1, 6);
     if (range < std::numeric_limits<double>::epsilon()) {
-      *H_nTb = Matrix16::Zero();
+      H_nTb->setZero();
     } else {
       // u = unit vector from satellite to antenna
-      const RowVector3d u = (position_difference / range).transpose();
-      H_nTb->resize(1, 6);
+      const Matrix u = (position_difference / range).transpose();  // 1x3
       H_nTb->block<1, 3>(0, 0) = u * (-nRb * skewSymmetric(bL_));
       H_nTb->block<1, 3>(0, 3) = u * nRb;
     }
@@ -251,12 +251,12 @@ Vector DifferentialPseudorangeFactorArm::evaluateError(
 
   // Compute associated derivatives:
   if (H_nTb) {
+    H_nTb->resize(1, 6);
     if (range < std::numeric_limits<double>::epsilon()) {
-      *H_nTb = Matrix16::Zero();
+      H_nTb->setZero();
     } else {
       // u = unit vector from satellite to antenna
-      const RowVector3d u = (position_difference / range).transpose();
-      H_nTb->resize(1, 6);
+      const Matrix u = (position_difference / range).transpose();  // 1x3
       H_nTb->block<1, 3>(0, 0) = u * (-nRb * skewSymmetric(bL_));
       H_nTb->block<1, 3>(0, 3) = u * nRb;
     }
--- a/gtsam/navigation/PseudorangeFactor.cpp
+++ b/gtsam/navigation/PseudorangeFactor.cpp
@@ -201,4 +201,76 @@ Vector PseudorangeFactorArm::evaluateError(
   return Vector1(error);
 }
 
+//***************************************************************************
+DifferentialPseudorangeFactorArm::DifferentialPseudorangeFactorArm(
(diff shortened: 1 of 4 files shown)


dellaert:
@masoug Could you take a look on whether this fits in the pseudo-range factor hierarchy cleanly?

masoug:
> @masoug Could you take a look on whether this fits in the pseudo-range factor hierarchy cleanly?

I think this is fine for the time being. Eventually we'll need to re-work the PseudorangeFactor API as more features and capability get added (e.g. different frequencies, yes/no lever arms, carrier vs code measurements, non-US GNSS satellites, etc...) in the future. So if API stability is a critical priority, we'll want to do that refactor here and now. Otherwise, I'm happy with the proposed changes in this PR, and we'll cross the PseudorangeFactor API bridge when we get there.

inuex35:
Hi — I was working on RTK support in GTSAM but ended up stalling out. I'd love to see native RTK processing land in GTSAM eventually.
I've been experimenting in this repo: https://github.com/inuex35/pyins
Happy to help with implementation if there's interest!

inuex35:
Hi, I realized the current pseudorange factor is computed in ECEF, but when combining it with IMU or other sensors, it would be more convenient to use NED or ENU due to the complexity of IMU preintegration.

dellaert:
@masoug please discuss :-) I think ECEF might be more native to GPS, and if you want NED or ENU you have the additional burden to choose that - which we might not want to have as a prerequisite to using these factors.

varunagrawal:
Minor nitpick: the naming convention usually puts `Factor` at the end, so the factor should be called `PseudorangeArmFactor`.

dellaert:
> Minor nitpick: the naming convention usually puts `Factor` at the end, so the factor should be called `PseudorangeArmFactor`.

Even better, `PseudoRangeArmFactor`

masoug:
> @masoug please discuss :-) I think ECEF might be more native to GPS, and if you want NED or ENU you have the additional burden to choose that - which we might not want to have as a prerequisite to using these factors.

Sorry for the late reply, had a bunch of work/life stuff, not to mention Claude Code is quite the rabbit hole 🐰 

I concur with @dellaert; GNSS itself is unaware of attitude because its primary focus is point positioning. Therefore, concepts like ENU/NED aren't involved in pseudorange/carrier-phase equations. Instead, users are responsible for choosing the frame convention for their particular application.

The original intention I had in my head was for users to explicitly model SE(3) body-frame and R^3 antenna-frame variables as two distinct nodes linked by a stiff pose-point-factor lever-arm constraint. This decouples GNSS point-positioning from application-specific attitude estimation, allowing flexibility for arbitrary frame conventions encoded in that pose-point factor. But I can see the convenience that comes from native lever-arm support in the pseudorange factors, so maybe the pose3 pseudorange lever-arm factor should have a "ENU" or "NED" option to specify a body-frame convention for applying the offset?

dellaert:
Well, ENU and NED are navigation frame conventions, whereas the lever-arm is a body-frame quantity, right? Let me go check the code. @inuex35 please resolve the copilot comments that you have addressed and also please do the requested name change, then we can run CI.

dellaert:
This is awesome, but before we merge, I think the frame naming should be clarified.

Using nTb / nRb suggests a generic “navigation frame”, but in both factors the satellite position is explicitly ECEF, so the pose/translation side is effectively expected to be ECEF as well. In GTSAM, n often reads like a local navigation frame such as NED or ENU, which makes this easy to misread.

Would it make sense to rename these to something more explicit, for example eTb / eRb, or even ecef_T_body / ecef_R_body? That would make it much more obvious that:

- the body pose supplied to this factor is expected in ECEF,
- the satellite position is in ECEF, and
- any conversion to or from a local navigation frame remains the caller’s responsibility.

I think that expectation should also be stated directly in the class and constructor documentation, since otherwise a reader could reasonably assume this accepts an arbitrary local navigation frame.

inuex35:
I have updated the implementation to work in ECEF coordinates for now. However, if future sensor fusion with cameras and LiDAR is planned, I believe local-frame Jacobians will be necessary.
I also aligned the factor names with the existing convention, where GpsFactor is named GpsFactorArm.

dellaert (approved):
Thanks, my comments have been addressed. I also agree with the naming.

inuex35:
Could you merge this when you get a chance? I'm planning to work on lever arm compensation and ENU/NED coordinate support next.

PR #2448, opened 2026-03-04
SOT(3) Implementation - eqVIO 1/3

rohan-bansal:
Hello!

This is PR `1/3` as pertaining to the eqVIO implementation in GTSAM!

Changes:
- Implemented SOT(3) (Scaled Orthogonal Transform) Lie Group, derived from `MatrixLieGroup`, as `SOT3.h` and `SOT3.cpp` in `gtsam/geometry`
- Implemented corresponding test suite in `gtsam/geometry/tests`

Content of this group was adapted from LiePP's (van Goor's lie library) implementation of SOT(3).

As a refresher, here is the definition:
image

Here is the PR roadmap for the eqVIO implementation:
1. (this one)
2. `VIOGroup` implementation, derived from `ProductGroup` and `PowerGroup`
3. `EqVIOFilter` implementation

### DIFF
--- a/gtsam/geometry/SOT3.h
+++ b/gtsam/geometry/SOT3.h
@@ -1,0 +1,153 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file    SOT3.h
+ * @brief   The scaled orthogonal transforms SOT(3) = SO(3) x R>0
+ *
+ * This file implements the SOT(3) group as defined by eqVIO (van Goor et al., https://arxiv.org/pdf/2205.01980)
+ *
+ * @author  Rohan Bansal
+ * @date    2026
+ */
+
+#pragma once
+
+#include <gtsam/base/MatrixLieGroup.h>
+#include <gtsam/dllexport.h>
+#include <gtsam/geometry/SO3.h>
+
+#include <string>
+
+namespace gtsam {
+
+/**
+ * @class SOT3
+ * @brief Scaled orthogonal transforms: SOT(3) = SO(3) x R>0.
+ *
+ * Represents group elements Q = (R_Q, c_Q) with R_Q in SO(3) and c_Q > 0.
+ * The 4x4 matrix representation is block-diagonal: [[R, 0], [0, c]].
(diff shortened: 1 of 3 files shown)


gtsam/geometry/SOT3.h
```diff
@@ -0,0 +1,153 @@
...
+
+  /// Identity element.
+  static SOT3 Identity() { return SOT3(); }
+
+  /// Inverse element
+  SOT3 inverse() const { return SOT3(R_.inverse(), 1.0 / c_); }
+
+  /// Group multiplication: (R1,c1)*(R2,c2) = (R1*R2, c1*c2).
+  SOT3 operator*(const SOT3& other) const {
+    return SOT3(R_ * other.R_, c_ * other.c_);
```
dellaert, line 85: If this is true, is SOT3 then not just a product group? :-)
It could be that we need things in the EqVio filter that they are not provided by product group, but if not, it's just a typedef :-/

dellaert (approved):

gtsam/geometry/SOT3.h
dellaert: Could still be the case that we need to extend it, in which case we need to convert  the product group to a CRPT, as I did for ExtendedPose3

rohan-bansal: I also noticed that it can be represented as a product group, but thought it might be beneficial to stay faithful to van Goor's SOT(3) implementation from his library just in case it is needed down the road during my port.

That being said, maybe I can club this implementation together with the VIOGroup implementation, at which point I will know for sure whether it needs to be extended?

dellaert: Well, how is Van Goor's implementation different from product group? Will read answer in the morning :-)

rohan-bansal: You are correct that the group itself is mathematically a product group. 

What I meant by staying faithful to LiePP was that it is not implemented as a composition of groups, but as a dedicated type that exposes `wedge`, `vee`, `asMatrix`, `adjoint` as all explicit 4x4 matrix operations.

In order to expose those same functions in a product group setting, correct me if I am wrong, but I assume we would have to make SOT(3) a wrapper around `ProductLieGroup<SO3, double>` and manually define them. 

However, after tracing through the code, I am also noticing now that `wedge` and `vee` are only defined for the purposes of van Goor's library, and aren't used for the rest of the eqVIO implementation. Maybe in this case it would make more sense to just define it as a product group in GTSAM.

dellaert: If we need those, 2 ways to go: CRTP+extend, or actually define ProductMatrixLieGroup - let’s chat offline. Try with ProductLieGroup for now.

dellaert (changes requested):
We will likely close after chatting more

rohan-bansal:
Closed, new revisions in separate PR.

PR #2449, opened 2026-03-05
Add BAL SmartFactor example

dvorak0:
Summary:

  - Add examples/SFMExample_SmartFactor_bal.cpp.
  - It is a BAL-input version of SFMExample_SmartFactor.cpp:
      - reads data with SfmData::FromBalFile
      - builds SmartProjectionFactor<SfmCamera> graph
      - adds two camera priors to fix gauge/scale
      - runs LM and prints final error + iterations
  - Goal: enable direct SmartFactor testing on BAL datasets (comparable to SFMExample_bal).


dvorak0:
@dellaert hi prof. frank, could you have a look at this?

dellaert (approved):
LGTM !

PR #2450, opened 2026-03-07
variable dimension product groups

dellaert:
- adds support for variable-size (dynamic dimension) ProductLieGroup compositions 
- adds support for PowerLieGroup variable N (G is still fixed-size)
- propagates GetDimension()'s return type to size_t across the library and bindings.

All tests pass, c++ and python. New c++ tests that test variable size capabilities.


dellaert:
@ProfFan timing is still useless unless caching is disabled.

rohan-bansal (approved):
PR looks good, thanks for adding this! Will build upon it.

PR #2455, opened 2026-03-10
Add EqVIO Group, Manifold Foundation

rohan-bansal:
Hello!

This PR introduces the EqVIO group/state/symmetry foundation into gtsam_unstable. Summary of changes below.

- `EqVIOGroup.h/.cpp`: VI-SLAM symmetry group using the new dynamic Product/PowerLieGroup architecture.
- Updated `ProductLieGroup-inl.h` to implement retract and localCoordinates Jacobians, required for VIOGroup to function since it has runtime dynamic size
- `EqVIOState.h/.cpp`: EqVIO manifold state
- `EqVIOSymmetry.h/.cpp`: group actions on VIOGroup/VIOState

Other:
- Documented with doxygen to the best of my ability

Next:
- Port over EqVIOEqivariantFilter and ensure it is modular/independent from the eqvio codebase


rohan-bansal:
Seems like Copilot wants stricter error checking. Should I address this?

dellaert:
> Seems like Copilot wants stricter error checking. Should I address this?

No, not all of them. Just resolve all suggestions you don’t agree with - and also resolve all the ones you address. 

In binary operations that are exposed in public API, you might want to assert the number of landmarks are equal.

gtsam_unstable/navigation/EqVIOGroup.h
```diff
@@ -0,0 +1,246 @@
...
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file    EqVIOGroup.h
+ * @brief   Dynamic VI-SLAM symmetry group from Eq. (22)
+ */
```
dellaert: Add yourself as author here and in all files

gtsam_unstable/navigation/EqVIOGroup.h
```diff
@@ -0,0 +1,246 @@
...
+#include <gtsam/geometry/SO3.h>
+#include <gtsam_unstable/dllexport.h>
+
+#include <string>
+#include <vector>
+
+namespace gtsam {
+
+/** Similarity-orthogonal group SOT(3) represented as SO(3) x R (log-scale). */
+using SOT3 = ProductLieGroup<SO3, Vector1>;
```
dellaert: Double does not work?

gtsam_unstable/navigation/EqVIOGroup.h
```diff
@@ -0,0 +1,246 @@
...
+
+/**
+ * Eq. (22) VI-SLAM symmetry group:
+ *   G = SE_2(3) x R^6 x SE(3) x SOT(3)^n.
+ *
+ * The Lie structure is delegated to nested ProductLieGroup/PowerLieGroup
+ * instances. This wrapper keeps the paper-native named accessors and the
```
dellaert: Rohan, the AI tried to do this for me as well - it’s just a bunch of extra generated code we do not need. I’d much prefer if we just use the SensorCore x LandmarCore *without* wrapping it in a class.

gtsam_unstable/navigation/EqVIOState.h
```diff
@@ -0,0 +1,149 @@
...
+   * @param lms ordered landmark list.
+   */
+  VIOState(const VIOSensorState& sensor_, const std::vector<Landmark>& lms);
+
+  /** Number of landmarks. */
+  size_t n() const;
+  int dim() const;
+  std::vector<int> ids() const;
+
+  /**
```
dellaert: This class is much leaner, so great :-) but please replace all “obvious” doxygen blocks with one liners, preceded by ///

dellaert (commented):
Very cool!

Overall request: reduce code and comment bloat. This is intended as an example of EqEKF, so it should be mean, lean, and easy to read. The wrapper class can be deleted iMHO, with some strategic free functions that provide easier to read accessors than “first.second.at()”

I also don’t fully understand the landmark id stuff. That seems like it’s book-keeping not part of the core group.

But I can see the outlines of something awesome ;-)

gtsam_unstable/navigation/EqVIOGroup.h
rohan-bansal: it does. updated!

gtsam_unstable/navigation/tests/testEqVIOSymmetry.cpp
```diff
@@ -0,0 +1,443 @@
...
+    H.col(j) =
+        (y0.localCoordinates(yPlus) - y0.localCoordinates(yMinus)) / (2.0 * h);
+  }
+  return H;
+}
+
+}  // namespace
+
+//******************************************************************************
+TEST(VIOSymmetry, RightActionLaw) {
```
dellaert, line 316: Add a one-liner // comment before each test stating what it is testing

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,201 @@
...
+
+GTSAM_UNSTABLE_EXPORT VisionMeasurement operator-(
+    const VisionMeasurement& y1, const VisionMeasurement& y2);
+
+GTSAM_UNSTABLE_EXPORT VisionMeasurement operator+(
+    const VisionMeasurement& y, const Vector& eta);
+
+/// Readable accessors for the composed ProductLieGroup VIOGroup.
+inline const VIOSE23& groupA(const VIOGroup& X) { return X.first.first; }
+inline VIOSE23& groupA(VIOGroup& X) { return X.first.first; }
```
dellaert: Do we really need write access? GTSAM philosophy is functional: things are *const* after construction. we deviate in a few cases but try to minimize.

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,201 @@
...
+#include <gtsam/geometry/SO3.h>
+
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace gtsam {
+
+using SOT3 = ProductLieGroup<SO3, double>;
```
dellaert, line 48: Not sure if I want the things in this file to be in top-level gtsam namespace. I also don't like the naming below, VIO*, Group* etc... I'd rather have
```
namespace gtsam {
namespace eqvio {
using SE23 = ExtendedPose3<2>;
using Bias = Vector6;
...
inline const SE23& A(const Group& X) { return X.first.first; }
}
}
```
The namespace makes all this nice :-)

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,201 @@
...
+
+using VIOSE23 = ExtendedPose3<2>;
+using VIOBias = Vector6;
```
dellaert: We have a type for bias in GTSAM. Can we not use that?

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,201 @@
...
+
+/// IMU reading bundle used by EqVIO.
+struct GTSAM_UNSTABLE_EXPORT IMUVelocity {
+  static constexpr int CompDim = 12;
+  using Vector12 = Eigen::Matrix<double, 12, 1>;
+
+  double stamp = -1.0;
+  Vector3 gyr = Vector3::Zero();
+  Vector3 acc = Vector3::Zero();
+  Vector3 gyrBiasVel = Vector3::Zero();
```
dellaert: I've never seen anyone use these velocity things. explain?

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,201 @@
...
+
+  /// Subtract stacked [gyr, acc, gyrBiasVel, accBiasVel].
+  IMUVelocity operator-(const Vector12& vec) const;
+
+  /// Scale all components.
+  IMUVelocity operator*(double c) const;
+};
+
+/// Camera model interface for EqVIO output/action math.
+class GTSAM_UNSTABLE_EXPORT VIOCameraModel {
```
dellaert: We have many cameras in GTSAM. Is this not equal to one of them?

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,201 @@
...
+  /// Project a camera-frame 3D point to image coordinates.
+  virtual Point2 projectPoint(const Point3& p) const = 0;
+  /// Convert image coordinates to an undistorted 3D bearing-like vector.
+  virtual Vector3 undistortPoint(const Point2& y) const = 0;
+  /// Projection Jacobian with respect to the input 3D vector.
+  virtual Matrix23 projectionJacobian(const Vector3& y) const = 0;
+};
+
+/// Vision measurement keyed by landmark id.
+struct GTSAM_UNSTABLE_EXPORT VisionMeasurement {
```
dellaert: Can we avoid having this type? Just use a std::map<Point2> ?

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,201 @@
...
+};
+
+GTSAM_UNSTABLE_EXPORT VisionMeasurement operator-(
+    const VisionMeasurement& y1, const VisionMeasurement& y2);
+
+GTSAM_UNSTABLE_EXPORT VisionMeasurement operator+(
+    const VisionMeasurement& y, const Vector& eta);
+
+/// Readable accessors for the composed ProductLieGroup VIOGroup.
+inline const VIOSE23& groupA(const VIOGroup& X) { return X.first.first; }
```
dellaert: Symbols are good for paper, but can we use descriptive names for these? `state` for this one I guess?

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,201 @@
...
+  return VIOGroup(VIOSensorCore(A, beta), VIOLandmarkCore(B, Q));
+}
+
+inline VIOGroup makeVIOGroupIdentity(size_t n = 0) {
+  return makeVIOGroup(VIOSE23::Identity(), Vector6::Zero(), Pose3::Identity(),
+                      VIOLandmarkGroup(n));
+}
+
+template <>
+struct traits<VisionMeasurement> {
```
dellaert: Why would the measurement need to be a manifold? Is that needed for the equivariant filter?

dellaert (commented):
Still a lot of comments, trying to make this even tighter and integrated with GTSAM.

CI is failing for Cayley case. Check into that, maybe disable the failing test.

gtsam_unstable/navigation/EqVIOCommon.h
rohan-bansal: I removed it in the newer commits to keep things cleaner.

gtsam_unstable/navigation/EqVIOCommon.h
rohan-bansal: I chose the Camera model with the closest similarity (PinholeCamera) and extended that instead. There isn't one single virtual base with the API needed, but I agree that now it is grounded in an existing camera.

gtsam_unstable/navigation/EqVIOCommon.h
rohan-bansal: This struct might be poorly named, its just an input/process noise bundle of values. The biases are random-walk driving inputs for bias dynamics. The current propagation/lift and process-noise mapping are formulated for a 12D input. A constant-bias model would reduce this to 6D and require corresponding model/Jacobian/noise changes.

I renamed this to IMUInput.

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,264 @@
...
+#include <cmath>
+#include <stdexcept>
+#include <vector>
+
+namespace gtsam {
+namespace eqvio {
+
+using SOT3 = ProductLieGroup<SO3, double>;
+
+using VIOSE23 = ExtendedPose3<2>;
```
dellaert: all the types starting with VIO yield the redundant naming eqvio::VIO... Please remove VIO from all those types. Maybe the only exception should be VIOGroup, as that is explicitly named in the paper. But that should then be `VioGroup`, not `VIOGroup`

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,264 @@
...
+
+  /// Construct from stacked [gyr, acc, gyrBiasVel, accBiasVel].
+  explicit IMUInput(const Vector12& vec) {
+    gyr = vec.segment<3>(0);
+    acc = vec.segment<3>(3);
+    gyrBiasVel = vec.segment<3>(6);
+    accBiasVel = vec.segment<3>(9);
+  }
+
+  /// Component-wise addition.
```
dellaert, line 108: Why do we have all these operators? Are they used?

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,264 @@
...
+    IMUInput out(*this);
+    out.gyr *= c;
+    out.acc *= c;
+    out.gyrBiasVel *= c;
+    out.accBiasVel *= c;
+    return out;
+  }
+};
+
+/// EqVIO camera built on top of GTSAM PinholeCamera<Cal3_S2>.
```
dellaert: Go full /** */ doxygen here and explain the camera model and why we can't use an existing camera model

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,264 @@
...
+  ids.reserve(measurement.size());
+  for (const auto& [id, _] : measurement) {
+    (void)_;
+    ids.push_back(id);
+  }
+  return ids;
+}
+
+/// Flatten measurements to [u0,v0,u1,v1,...] in map iteration order.
+inline Vector measurementVector(const VisionMeasurement& measurement) {
```
dellaert: these inline methods are *common* between different compilation units? Or can they be in anonymous namespace inside one .cpp file?

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,264 @@
...
+    if (itL->first != itR->first) {
+      throw std::invalid_argument("measurementDifference: id mismatch");
+    }
+    diff.segment<2>(2 * i) = itL->second - itR->second;
+    ++i;
+  }
+  return diff;
+}
+
+/// Readable accessors for the composed ProductLieGroup VIOGroup.
```
dellaert: Would be nice to be able to say
```
auto [A,Beta,B,Q] = X;
```

gtsam_unstable/navigation/EqVIOState.h
```diff
@@ -0,0 +1,128 @@
...
+#include <gtsam_unstable/navigation/EqVIOCommon.h>
+#include <gtsam_unstable/dllexport.h>
+
+#include <string>
+#include <vector>
+
+namespace gtsam {
+namespace eqvio {
+
+struct GTSAM_UNSTABLE_EXPORT Landmark {
```
dellaert, line 40: doc

gtsam_unstable/navigation/EqVIOState.h
```diff
@@ -0,0 +1,128 @@
...
+  static constexpr int CompDim = 3;
+
+  Point3 p = Point3::Zero();
+  int id = -1;
+
+  void print(const std::string& s = "") const;
+  bool equals(const Landmark& other, double tol = 1e-9) const;
+};
+
+struct GTSAM_UNSTABLE_EXPORT VIOSensorState {
```
dellaert: doc, and same comment on VIO prefix

gtsam_unstable/navigation/EqVIOState.h
```diff
@@ -0,0 +1,128 @@
...
+
+  /// Unit gravity direction expressed in the body frame.
+  Vector3 gravityDir() const;
+
+  void print(const std::string& s = "") const;
+  bool equals(const VIOSensorState& other, double tol = 1e-9) const;
+};
+
+/// Dynamic VIO state manifold with dimension 21 + 3n.
+class GTSAM_UNSTABLE_EXPORT VIOState {
```
dellaert: This could be VioState, or also just eqvio::State, unless explicitly named in paper

gtsam_unstable/navigation/EqVIOSymmetry.cpp
```diff
@@ -0,0 +1,325 @@
...
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/// @file EqVIOSymmetry.cpp
+/// @brief EqVIO symmetry actions and lift helpers.
+/// @author Rohan Bansal
```
dellaert: Need more docs in this file, at least ///...

dellaert (commented):
Sorry to be a pain, but another iteration needed :-)

gtsam_unstable/navigation/EqVIOCommon.h
rohan-bansal: Were not used. removed!

gtsam_unstable/navigation/EqVIOCommon.h
rohan-bansal: Reduced a bit more to a `using` and 2 helpers; Added the Doxygen anyway to explain.

gtsam_unstable/navigation/EqVIOCommon.h
rohan-bansal: Removed! They are used in the filter, i moved them to that PR.

gtsam_unstable/navigation/EqVIOCommon.h
rohan-bansal: I agree that this would be nice. However, I think it would add a large code footprint, as direct structured binding on VioGroup requires implementing the full `get<I>` overloads, with extra namespace plumbing (gtsam + std) because VioGroup is a type alias over ProductLieGroup rather than a dedicated wrapper type. If I do not put it in the gtsam namespace, I don't think the ADL recognizes it. 

Maybe a lighter-weight `decompose(X)`? Or can I push this change to the filter PR to see where it can be made useful concretely?

rohan-bansal:
Apologies for the day-long wait. I was implementing the filter in parallel to figure out exactly what functions are required/not; I think it is in a better place now.

gtsam_unstable/navigation/EqVIOCommon.h
dellaert: Decompose sounds good, return const refs

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,262 @@
...
+
+using Se23 = ExtendedPose3<2>;
+using Bias = imuBias::ConstantBias;
+using LandmarkGroup = PowerLieGroup<SOT3, Eigen::Dynamic>;
+using SensorCore = ProductLieGroup<Se23, Bias>;
+using LandmarkCore = ProductLieGroup<Pose3, LandmarkGroup>;
+using VioGroup = ProductLieGroup<SensorCore, LandmarkCore>;
+
+/// Approximate gravitational acceleration magnitude in m/s^2.
+constexpr double GRAVITY_CONSTANT = 9.80665;
```
dellaert, line 58: We should probably use PreintegrationParams, that has a g (which varies from place to place) as well as IMU noise params

gtsam_unstable/navigation/EqVIOCommon.h
```diff
@@ -0,0 +1,262 @@
...
+/// IMU input bundle used by EqVIO propagation.
+struct GTSAM_UNSTABLE_EXPORT IMUInput {
+  static constexpr int CompDim = 12;
+  using Vector12 = Eigen::Matrix<double, 12, 1>;
+
+  double stamp = -1.0;
+  Vector3 gyr = Vector3::Zero();
+  Vector3 acc = Vector3::Zero();
+  Vector3 gyrBiasVel = Vector3::Zero();
+  Vector3 accBiasVel = Vector3::Zero();
```
dellaert: Use Z_3x1

gtsam_unstable/navigation/EqVIOState.cpp
```diff
@@ -0,0 +1,186 @@
...
+  if (!s.empty()) std::cout << s << " ";
+  std::cout << "Landmark{id=" << id << ", p=" << p.transpose() << "}"
+            << std::endl;
+}
+
+bool Landmark::equals(const Landmark& other, double tol) const {
+  return id == other.id && equal_with_abs_tol(p, other.p, tol);
+}
+
+Vector3 SensorState::gravityDir() const {
```
dellaert, line 37: Also in params you need to choose a up or Z down.

Params should probably only appear in filter, though, and this method as well. State is a mathematical object, should be independent of gravity?

gtsam_unstable/navigation/EqVIOState.cpp
```diff
@@ -0,0 +1,186 @@
...
+      other.sensor.cameraOffset, H1 ? &Hcam1 : nullptr, H2 ? &Hcam2 : nullptr);
+
+  for (size_t i = 0; i < n(); ++i) {
+    out.segment<3>(SensorState::CompDim + 3 * static_cast<int>(i)) =
+        other.cameraLandmarks[i].p - cameraLandmarks[i].p;
+  }
+
+  if (H1) {
+    H1->setZero(d, d);
+    H1->block<6, 6>(0, 0) = -Matrix66::Identity();
```
dellaert: Use I_6x6

gtsam_unstable/navigation/EqVIOState.cpp
```diff
@@ -0,0 +1,186 @@
...
+    H1->block<6, 6>(6, 6) = Hpose1;
+    H1->block<3, 3>(12, 12) = -Matrix3::Identity();
```
dellaert: And I_3x3

gtsam_unstable/navigation/EqVIOState.h
```diff
@@ -0,0 +1,152 @@
...
+ * @brief One visual landmark state block.
+ *
+ * Stores a 3D landmark position plus its integer id. The Lie/chart dimension
+ * contribution is 3.
+ */
+struct GTSAM_UNSTABLE_EXPORT Landmark {
+  static constexpr int CompDim = 3;
+
+  /// Landmark position in world coordinates.
+  Point3 p = Point3::Zero();
```
dellaert: Z_3x1

gtsam_unstable/navigation/EqVIOSymmetry.cpp
```diff
@@ -0,0 +1,335 @@
...
+  const size_t N = state.n();
+  Vector lift = Vector::Zero(21 + 4 * static_cast<int>(N));
+
+  const SensorState& sensor = state.sensor;
+  const IMUInput v_est = velocity - sensor.inputBias;
+
+  Vector6 U_A;
+  U_A << v_est.gyr, sensor.velocity;
+  const Vector6 U_B = sensor.cameraOffset.inverse().AdjointMap() * U_A;
+  const Vector3 u_w = -v_est.acc + sensor.gravityDir() * GRAVITY_CONSTANT;
```
dellaert, line 168: Interesting. Ok, maybe leave params out of it for now and we can chat

gtsam_unstable/navigation/EqVIOSymmetry.cpp
```diff
@@ -0,0 +1,335 @@
...
+  q.resize(state.n());
+  for (size_t i = 0; i < state.n(); ++i) {
+    const Landmark& lm0 = state.cameraLandmarks[i];
+    Landmark lm1;
+    lm1.id = lm0.id;
+    lm1.p = cameraPoseChangeInv.transformFrom(lm0.p);
+
+    const Rot3 R = RotationFromTwoVectors(lm1.p, lm0.p);
+    const double a = lm0.p.norm() / lm1.p.norm();
+    q[i] = MakeSOT3(SO3(R.matrix()), a);
```
dellaert, line 226: Is that not just {R,a}?

gtsam_unstable/navigation/EqVIOSymmetry.cpp
```diff
@@ -0,0 +1,335 @@
...
+      sensor.inputBias.accelerometer() + dt * velocity.accBiasVel,
+      sensor.inputBias.gyroscope() + dt * velocity.gyrBiasVel);
+
+  const Rot3 dR = Rot3::Expmap(dt * v_est.gyr);
+  const Vector3 dXWorld =
+      dt * (sensor.pose.rotation() * sensor.velocity) +
+      0.5 * dt * dt *
+          (sensor.pose.rotation() * v_est.acc + Vector3(0, 0, -GRAVITY_CONSTANT));
+  const Point3 dXBody = sensor.pose.rotation().unrotate(dXWorld);
+  const Pose3 poseChange(dR, dXBody);
```
dellaert: Use naming convention aTb for pose of b in frame a (transforms b cords to frame a). Try to do that everywhere where named frames are involved.

dellaert (commented):
Last one I think ;-)

gtsam_unstable/navigation/EqVIOSymmetry.cpp
rohan-bansal: It is, but SOT3 is a safe constructor wrapper for it, since we need to enforce `a > 0`. It is stored as `{R, log(a)}`. Maybe there is a better way to do it?

rohan-bansal:
Updated! Will wait on the params stuff as mentioned.

dellaert (approved):
Nice! I’ll merge

PR #2456, opened 2026-03-10
Fix enforceConsistency_ in BatchFixedLagSmoother

### LINKED ISSUE #948: Bug in BatchFixLagSmoother "enforceConsistency_" not working
I am working on some monocular SFM with a sliding window and notice that the 
_enforceConsistency__ flag seems not doing anything and _linearKeys__ are never assigned with marginalized values.

This can be easily reproduced by using "FixedLagSmootherExample.cpp" by adding some assertions. In the example, one key and their factors will get marginalized out, and to make the system consistent (In my case, I want the prior factor to get carried forward), we need to fix the linearization point for the variables related to marginalized ones. I think the fix is not hard, and I am happy to put up a PR but would like to get confirmation that this is an issue. 

```cpp
          if (enforceConsistency_ && (linearKeys_.size() > 0)) {
            assert(0); // never triggered when enforceConsistency_ is on even when we doing marginalization 
            theta_.update(linearKeys_);
            for(const auto key_value: linearKeys_) {
              delta_.at(key_value.key) = newDelta.at(key_value.key);
            }
          }
```

## Steps to reproduce

1. 
2.

jashshah999:
## What this does

Fixes the `enforceConsistency_` feature in `BatchFixedLagSmoother`, which has been non-functional since `linearValues_` was never populated.

## The bug

`linearValues_` (declared in `BatchFixedLagSmoother.h:149`) is read in `optimize()` (line 268) to decide whether to restore linearization points:

```cpp
if (enforceConsistency_ && (linearValues_.size() > 0)) {
    theta_.update(linearValues_);
    ...
}
```

But `linearValues_` is never written to anywhere in the codebase. It is always empty, so the guard is always `false` and the consistency enforcement is dead code.

## The fix

After `marginalize()` inserts new `LinearContainerFactor`s (which store the linearization point at marginalization time), we now record the remaining variables and their linearization points in `linearValues_`. This allows the existing logic in `optimize()` to restore those linearization points after each LM iteration, keeping the marginal factors consistent.

The existing cleanup in `eraseKeys()` already handles removing entries from `linearValues_` when those variables are themselves eventually marginalized out.

Fixes #948

cc @dellaert

### DIFF
--- a/gtsam/nonlinear/BatchFixedLagSmoother.cpp
+++ b/gtsam/nonlinear/BatchFixedLagSmoother.cpp
@@ -342,6 +342,20 @@ void BatchFixedLagSmoother::marginalize(const KeyVector& marginalizeKeys) {
 
   // Insert the new marginal factors
   insertFactors(marginalFactors);
+
+  // Record the linearization points of variables involved in new marginal
+  // factors. When enforceConsistency_ is enabled, these variables must keep
+  // their current linearization point so the LinearContainerFactors remain
+  // valid across future optimization iterations.
+  if (enforceConsistency_) {
+    for (const auto& factor : marginalFactors) {
+      for (Key key : factor->keys()) {
+        if (!linearValues_.exists(key)) {
+          linearValues_.insert(key, theta_.at(key));
+        }
+      }
+    }
+  }
 }
 
 /* ************************************************************************* */
--- a/gtsam/nonlinear/tests/testBatchFixedLagSmoother.cpp
+++ b/gtsam/nonlinear/tests/testBatchFixedLagSmoother.cpp
@@ -18,14 +18,17 @@
 
 #include <CppUnitLite/TestHarness.h>
 #include <gtsam/nonlinear/BatchFixedLagSmoother.h>
+#include <gtsam/nonlinear/Marginals.h>
 #include <gtsam/base/debug.h>
 #include <gtsam/inference/Key.h>
 #include <gtsam/geometry/Point2.h>
+#include <gtsam/geometry/Pose2.h>
 #include <gtsam/linear/GaussianBayesNet.h>
 #include <gtsam/linear/GaussianFactorGraph.h>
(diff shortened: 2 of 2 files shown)


dellaert:
Ha ! That's a bug, thanks ! 
Did you do any experiments to show that this now improves consistency? That would be a good "cherry" on the cake.

jashshah999:
Thanks! Added a test in the latest push that exercises `enforceConsistency_` with both `true` and `false`:

- Creates two `BatchFixedLagSmoother` instances with identical data (chain of between-factors with poor initial estimates)
- Runs both through enough steps to trigger marginalization (lag=3, 8 steps)
- Verifies both produce valid estimates after marginalization

The test builds and passes with `make testBatchFixedLagSmoother`.

Before the fix, `linearValues_` was always empty so both smoothers behaved identically (the consistency guard was dead code). Now the `enforceConsistency=true` smoother actually populates `linearValues_` during `marginalize()` and restores linearization points in `optimize()`.

dellaert:
That's great ! Any NEES comparison? This flag was supposed to be implement FEJ which should make for better NEES values.

jashshah999:
Added a Monte Carlo NEES test using Pose2 (100 trials, 30 steps, lag=3, significant rotation per step).

Results:
```
enforceConsistency=true  (FEJ): avg NEES = 1.14 (expected: 3)
enforceConsistency=false       : avg NEES = 3.12 (expected: 3)
```

The key takeaway: with the fix, `enforceConsistency_` now has an observable effect (before the fix, `linearValues_` was never populated, so both modes were identical). The `false` mode gives near-ideal NEES, while the `true` mode is conservative — which makes sense for a batch solver where iterative re-linearization already handles much of the inconsistency that FEJ addresses in filtering.

dellaert (approved):
Thanks, that's fast !!! Will run CI and merge when passes!

PR #2461, opened 2026-03-12
Move free functions definitions to `GncOptimizer.cpp`

### LINKED ISSUE #2460: Including GncOptimizer causes duplicate symbol error
## Description

When I include GncOptimizer in a downstream codebase, I get duplicate symbol errors related to some GNC related methods.

I tried a version of GTSAM before all of @hkhanuja's recent changes and it compiles without any issues, so I'm guessing something was imported in a way that's causing the preprocessor to include it twice aka some kind of bad diamond inheritance of includes.

I'm tagging @hkhanuja since he's best positioned to identify the issue.

## Steps to reproduce

1. Include the GncOptimizer in a different application.
2. Link GTSAM during compilation.

## Expected behavior

Linker should not complain.

varunagrawal:
Added a new file called `GncOptimizer.cpp` so that I could move the free function definitions to it. These were originally added and defined erroneously in #2410.

This resolves the duplicate symbol error reported in #2460.

### DIFF
--- a/gtsam/nonlinear/GncOptimizer.cpp
+++ b/gtsam/nonlinear/GncOptimizer.cpp
@@ -26,17 +26,24 @@ double Chi2inv(const double alpha, const size_t dofs) {
 }
 
 /* ************************************************************************* */
-bool isNullType(Type type) { return type == Type::NullPointer; }
+bool isNullType(GncFactorType type) {
+  return type == GncFactorType::NullPointer;
+}
 
 /* ************************************************************************* */
-bool isNonNoiseModelType(Type type) { return type == Type::NonNoiseModel; }
+bool isNonNoiseModelType(GncFactorType type) {
+  return type == GncFactorType::NonNoiseModel;
+}
 
 /* ************************************************************************* */
-bool needsWeightUpdate(Type type) { return type == Type::Normal; }
+bool needsWeightUpdate(GncFactorType type) {
+  return type == GncFactorType::Normal;
+}
 
 /* ************************************************************************* */
-bool hasNoise(Type type) {
-  return type == Type::Normal || type == Type::Inlier || type == Type::Outlier;
+bool hasNoise(GncFactorType type) {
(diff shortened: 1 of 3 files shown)


gtsam/nonlinear/GncOptimizer.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * @brief   Definitions of functions declared in header file
+ * @author  Varun Agrawal
+ */
+
+#include <gtsam/nonlinear/GncOptimizer.h>
+
+namespace gtsam {
+
+/* ************************************************************************* */
+bool isNullType(Type type) { return type == Type::NullPointer; }
```
varunagrawal: @dellaert I believe `Type` is too generic an enum name, and it should instead be `GncFactorType` to better describe its purpose. Should I add that change to this PR?

dellaert: I agree. Could we make it an internal enum so it's GncOptimizer::Type?

varunagrawal: That would involve providing the template parameters, making the type name quite verbose, e.g. `GncOptimizer<GncParams<LevenbergMarquardtParams>>::FactorType`.
I am using `FactorType` since `GncOptimizer::Type` reads like it is a type variant of GncOptimizer and not the factors under consideration.

varunagrawal: Additionally, all the functions would now have to be templated. Would you prefer we move them inside the `GncOptimizer` class definition instead of keeping them as free functions?

varunagrawal: Okay making the internal namespace is a significant refactor which I would prefer to do after a design review. I'll change the name to `GncFactorType` and we can revisit this in the future.

dellaert (approved):
This solution is fine with me.

PR #2464, opened 2026-03-12
Fix tMin_ reset when an iteration has zero duration

### LINKED ISSUE #1636: tMin_ resets when an iteration has 0 us duration.
The timing class [resets](https://github.com/borglab/gtsam/blob/develop/gtsam/base/timing.cpp#L231) `tMin_` when `tMin_==0`. I guess this was set because `tMin_` [initializes](https://github.com/borglab/gtsam/blob/develop/gtsam/base/timing.cpp#L58) with `0` . However, it is well possible that during runtime a timing iteration will only last `0` micro seconds. Then `tMin_` keeps resetting.

An easy fix would probably be to initialize `tMin_(SIZE_MAX)` or whatever the maximum size for this variable is.

jashshah999:
## Summary

`TimingOutline::tMin_` is initialized to 0, and `finishedIteration()` uses `tMin_ == 0` as a proxy for "first iteration". This means any iteration that legitimately completes in 0 microseconds causes `tMin_` to keep resetting, producing incorrect minimum timing statistics.

Fix: initialize `tMin_` to `std::numeric_limits<size_t>::max()` so any real measurement is naturally smaller, and simplify the comparison in `finishedIteration()`.

Fixes #1636

### DIFF
--- a/gtsam/base/timing.cpp
+++ b/gtsam/base/timing.cpp
@@ -23,6 +23,7 @@
 #include <cassert>
 #include <cmath>
 #include <cstddef>
+#include <limits>
 #include <iomanip>
 #include <iostream>
 #include <map>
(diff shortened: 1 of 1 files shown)


dellaert (approved):
LGTM, thanks !!!

PR #2466, opened 2026-03-12
Validate rotation matrix in Rot3 constructor to prevent segfault

### LINKED ISSUE #2300: Invalid input to gtsam.Rot3 causes segfault in python wrapper.
Description

Constructing a gtsam.Rot3 object from an invalid (non-orthonormal) 3×3 numpy matrix causes a hard crash of the Python interpreter (kernel dies with no traceback), instead of raising a Python exception or a GTSAM error.

This is very very hard to debug in jupyternotebook as it does not show line of error, or exactly what cause the error at all. All it shows is: 
```
The Kernel crashed while executing code in the current cell or a previous cell. 
Please review the code in the cell(s) to identify a possible cause of the failure. 
Click [here](https://aka.ms/vscodeJupyterKernelCrash) for more info. 
View Jupyter [log](command:jupyter.viewOutput) for further details.
```

Screenshot: 
Image

Steps to reproduce

```py
import numpy as np
import gtsam

# This matrix is 3x3 but NOT a valid rotation matrix (not orthonormal, det != 1)
R_invalid = np.array([
    [0.57367956, 0.4419256,  0.87448758],
    [0.94397185, 0.06276365, 0.60304828],
    [0.92277076, 0.43834417, 0.20227668],
])

# This line immediately causes a segfault / kernel crash:
gtsam.Rot3(R_invalid)
```

Observed behavior:
	•	Python interpreter terminates instantly (segfault / abort)
	•	Jupyter kernel dies with message:
“The kernel crashed while executing code…”
	•	No Python exception is raised.
	•	No stack trace is printed.

This occurs even if this is the only line executed in a clean environment.

Expected behavior

Rot3 should:
	•	Validate the input matrix (orthonormality, determinant +1), AND
	•	Raise a Python exception (e.g., RuntimeError or ValueError) instead of terminating the interpreter.

For example:

raise ValueError("Input matrix is not a valid rotation matrix")

Failing fast with an exception is vastly preferable to a process-level abort.

Environment 
	•	OS: macOS 15 (Apple Silicon)
	•	Python: 3.9.6 (system Python), also reproducible in Python 3.11 virtual environment
	•	GTSAM: Wheel distributed for Python (via pip)
	•	Interface: Python
	•	Reproduction environments:
	•	Jupyter Notebook (VS Code)
	•	Plain Python REPL (python -X faulthandler)
	•	Minimal scripts run from terminal

The crash occurs regardless of environment.

Additional information

This issue came up during a class assignment where students construct Rot3 from numpy matrices. A single invalid 3×3 matrix instantly kills the kernel, making it difficult for students to debug their code.

A safer behavior (throwing an exception) would be extremely helpful for both users and educators relying on GTSAM in Python-heavy workflows.

jashshah999:
## Summary

Constructing a `Rot3` from a non-orthonormal 3x3 matrix causes a hard segfault that kills the Python interpreter with no traceback. This is especially painful in Jupyter notebooks.

Added validation in the `Matrix3` and template matrix constructors that checks orthonormality (`R'R ≈ I`) and determinant sign (`det > 0`), throwing `std::invalid_argument` with a helpful message suggesting `Rot3::ClosestTo()` as an alternative.

```python
# Before: segfault, kernel dies
gtsam.Rot3(invalid_matrix)

# After: clean exception
# RuntimeError: Rot3 constructor: matrix is not orthonormal.
# Use Rot3::ClosestTo() to find the closest valid rotation.
```

Fixes #2300

### DIFF
--- a/gtsam/geometry/Rot3.cpp
+++ b/gtsam/geometry/Rot3.cpp
@@ -34,7 +34,7 @@ namespace gtsam {
 /* ************************************************************************* */
 void Rot3::IsValid(const Matrix3& R) {
   // Check R'R ≈ I (orthonormality)
-  if ((R.transpose() * R - Matrix3::Identity()).norm() > 1e-5) {
+  if ((R.transpose() * R - Matrix3::Identity()).norm() > 1e-3) {
     throw std::invalid_argument(
         "Rot3 constructor: matrix is not orthonormal. "
         "Use Rot3::ClosestTo() to find the closest valid rotation.");
--- a/gtsam/geometry/Rot3.cpp
+++ b/gtsam/geometry/Rot3.cpp
@@ -25,10 +25,26 @@
 #include <cmath>
 #include <cassert>
 #include <random>
+#include <stdexcept>
 
 using namespace std;
 
 namespace gtsam {
+
+/* ************************************************************************* */
+void Rot3::IsValid(const Matrix3& R) {
+  // Check R'R ≈ I (orthonormality)
+  if ((R.transpose() * R - Matrix3::Identity()).norm() > 1e-5) {
+    throw std::invalid_argument(
+        "Rot3 constructor: matrix is not orthonormal. "
+        "Use Rot3::ClosestTo() to find the closest valid rotation.");
+  }
+  // Check det(R) ≈ +1 (proper rotation, not reflection)
+  if (R.determinant() < 0) {
+    throw std::invalid_argument(
+        "Rot3 constructor: matrix has negative determinant (reflection, not rotation).");
+  }
+}
 
 /* ************************************************************************* */
 void Rot3::print(const std::string& s) const {
--- a/gtsam/geometry/Rot3.h
+++ b/gtsam/geometry/Rot3.h
@@ -97,10 +97,13 @@ class GTSAM_EXPORT Rot3 : public MatrixLieGroup<Rot3, 3, 3> {
     template <typename Derived>
 #ifdef GTSAM_USE_QUATERNIONS
     explicit Rot3(const Eigen::MatrixBase<Derived>& R) {
-      quaternion_ = Matrix3(R);
+      const Matrix3 M(R);
+      IsValid(M);
+      quaternion_ = M;
     }
 #else
-    explicit Rot3(const Eigen::MatrixBase<Derived>& R) : rot_(R) {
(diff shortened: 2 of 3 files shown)


dellaert:
CI fails now
```
 63/309 Test  #63: testSimilarity3 ...............................***Failed    0.01 sec
/__w/gtsam/gtsam/gtsam/geometry/tests/testSimilarity3.cpp:146: Failure: "Exception: Rot3 constructor: matrix is not orthonormal. Use Rot3::ClosestTo() to find the closest valid rotation." 
/__w/gtsam/gtsam/gtsam/geometry/tests/testSimilarity3.cpp:175: Failure: "Exception: Rot3 constructor: matrix is not orthonormal. Use Rot3::ClosestTo() to find the closest valid rotation." 
There were 2 failures
```
maybe because of your stronger test :-)

jashshah999:
Good catch! The tolerance was too tight at 1e-5 -- those testSimilarity3 cases construct Rot3 from hardcoded Matlab values with ~4 decimal places, giving orthonormality errors of ~1e-4 to 2e-4.

Relaxed the tolerance to 1e-3 in 664cb92. This still catches clearly invalid matrices (the garbage matrix from issue #2300 has an error of ~1.0) while allowing for reasonable floating point imprecision. Both testRot3 and testSimilarity3 pass now.

ProfFan:
Is this correct? I don't think this patch is relevant, the original issue is for 4.2 and the reporter said there is no issue for 4.3

dellaert:
There are also two issues to discuss:
- this incurs compute at any construction - might slow down optimization 
- might it even break optimization runs where the matrix de-normalizes over time? Might be a good way to *diagnose* this issue, but it's a behavior change, and we should discuss pros and cons...

jashshah999:
Thanks for the feedback, both valid points.

@ProfFan -- you're right, I missed that the reporter confirmed the issue is specific to 4.2 and doesn't reproduce on 4.3/develop. So this is already resolved for anyone on the current version.

@dellaert -- agreed on both concerns. The per-construction overhead would add up in optimization loops, and the tolerance question (1e-3 vs 1e-5) is inherently fragile -- too tight breaks legitimate use, too loose defeats the purpose. And the risk of breaking optimization runs where matrices drift slightly out of orthonormality is real.

Given that the original issue is already fixed in 4.3, and the performance/correctness tradeoffs here are non-trivial, I think the right call is to close this. Happy to revisit if the need comes up again in a different context (e.g. a debug-only validation flag).

PR #2467, opened 2026-03-12
Fix LinearContainerFactor::rekey() segfault without linearization point

### LINKED ISSUE #1904: LinearFactorContainer rekey()
Hi,
I am getting seg faults or ValuesKeyDoesNotExist errors when using rekey on the LinearFactorContainer.

## Description

As the linearization point is optional and there are checks for `linearizationPoint_` in some methods, the rekey method is missing these checks.
https://github.com/borglab/gtsam/blob/develop/gtsam/nonlinear/LinearContainerFactor.cpp#L180

## Steps to reproduce

1. create  LinearContainerFactor with GaussianFactor but without a linearization point
2. rekey the LinearContainerFactor

```

#include <gtsam/linear/HessianFactor.h>
#include <gtsam/nonlinear/LinearContainerFactor.h>

int main(int argc, char *argv[]) {

    double mu = 1e10;
    size_t dim = 3;
    gtsam::Key key = 0;

    gtsam::HessianFactor H(key, mu * gtsam::Matrix::Identity(dim, dim), gtsam::Vector::Zero(dim), 0.0);
    gtsam::LinearContainerFactor factor(H);

    gtsam::Key new_key = 1;
    std::map<gtsam::Key, gtsam::Key> rekey_mapping = {{key, new_key}};
    factor.rekey(rekey_mapping);

    gtsam::KeyVector rekey_vector = {new_key};
    factor.rekey(rekey_vector);

    return 0;
}

```

## Expected behavior

Remap the keys of the GaussianFactor and only modify `linearizationPoint_` if it has a value.

## Environment

C++, Ubuntu 20.04, gtsam develop branch commit: https://github.com/borglab/gtsam/commit/164c35424b13057c10a261eb4c8b6343125dccc9

jashshah999:
## Description
`LinearContainerFactor::rekey()` segfaults when called on a factor constructed without a linearization point. Both overloads unconditionally dereference the optional `linearizationPoint_` member.

## Fix
Guard the linearization point rekeying with a `has_value()` check in both `rekey` overloads, so the underlying linear factor keys are still remapped correctly even when no linearization point exists.

## Test
Added `RekeyWithoutLinearizationPoint` test that exercises both `rekey` overloads on a `LinearContainerFactor` constructed from a `HessianFactor` without a linearization point.

Fixes #1904

### DIFF
--- a/gtsam/nonlinear/LinearContainerFactor.cpp
+++ b/gtsam/nonlinear/LinearContainerFactor.cpp
@@ -171,16 +171,24 @@ NonlinearFactor::shared_ptr LinearContainerFactor::rekey(
   // Update the keys to the properties as well
   // Downncast so we have access to members
   auto new_factor = std::static_pointer_cast<LinearContainerFactor>(rekeyed_base_factor);
-  // Create a new Values to assign later
-  Values newLinearizationPoint;
+  // Rekey the underlying linear factor
   for (size_t i = 0; i < factor_->size(); ++i) {
     auto mapping = rekey_mapping.find(factor_->keys()[i]);
     if (mapping != rekey_mapping.end()) {
       new_factor->factor_->keys()[i] = mapping->second;
-      newLinearizationPoint.insert(mapping->second, linearizationPoint_->at(mapping->first));
     }
   }
-  new_factor->linearizationPoint_ = newLinearizationPoint;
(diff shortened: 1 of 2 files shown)


dellaert (approved):
Thanks for the fix!

PR #2471, opened 2026-03-16
Rename Python reserved keywords in wrapper interface files

### LINKED ISSUE #2322: reserved python keywords are used as argument names in python stubs, causing mypy errors
## Description

mypy finds errors in the gtsam's generated .pyi files, because reserved python keywords (e.g. `global` and `lambda`) are used as argument names.

I observed the problem in `gtsam 4.3a0` and `gtsam 4.3a1`.

## Steps to reproduce

```bash
$ cmake .. -DGTSAM_BUILD_PYTHON=1
...
-- Build files have been written to: /home/me/Downloads/gtsam-4.3a1/build
```

```bash
$ make python-install
...
Successfully installed gtsam-4.3a1
[100%] Built target python-install
```

```bash
$ mypy --version
mypy 1.19.1 (compiled: yes)
```

```bash
$ mypy python/gtsam/gtsam/__init__.pyi
python/gtsam/gtsam/__init__.pyi:11941: error: Invalid syntax [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

```bash
$ sed -n 11941p python/gtsam/gtsam/__init__.pyi
def evaluateError(self, global: typing.Annotated[numpy.typing.ArrayLike, numpy.float64, "[3, 1]"], trans: Pose3, local: typing.Annotated[numpy.typing.ArrayLike, numpy.float64, "[3, 1]"]) -> typing.Annotated[numpy.typing.NDArray[numpy.float64], "[m, 1]"]:
```

## Expected behavior

Reserved python keywords should not be used as argument names `.pyi` files, mypy should not find errors in `.pyi` files.

jashshah999:
## Summary

The generated `.pyi` stub files use `lambda` and `global` as parameter/method names, which are Python reserved keywords. This causes mypy to report syntax errors on the stubs.

Renames in `.i` wrapper files:
- `lambda` → `lambda_` (parameter names in SmartProjectionFactor methods, LevenbergMarquardtOptimizer method name)
- `global` → `global_` (parameter name in ReferenceFrameFactor)

The wrapper tool already handles method name remapping (generating `self->lambda()` calls under a `lambda_` Python binding), so no C++ source changes are needed.

Fixes #2322

### DIFF
--- a/gtsam/slam/ReferenceFrameFactor.h
+++ b/gtsam/slam/ReferenceFrameFactor.h
@@ -97,10 +97,10 @@ public:
         NonlinearFactor::shared_ptr(new This(*this))); }
 
   /** Combined cost and derivative function using boost::optional */
-  Vector evaluateError(const Point& global, const Transform& trans, const Point& local,
+  Vector evaluateError(const Point& _global, const Transform& trans, const Point& local,
         OptionalMatrixType Dforeign, OptionalMatrixType Dtrans,
         OptionalMatrixType Dlocal) const override {
-    Point newlocal = transform_point<Transform,Point>(trans, global, Dtrans, Dforeign);
+    Point newlocal = transform_point<Transform,Point>(trans, _global, Dtrans, Dforeign);
     if (Dlocal) {
       *Dlocal = -1* Matrix::Identity(traits<Point>::dimension, traits<Point>::dimension);
     }
--- a/gtsam/slam/SmartProjectionFactor.h
+++ b/gtsam/slam/SmartProjectionFactor.h
@@ -196,7 +196,7 @@ protected:
 
   /// Create a Hessianfactor that is an approximation of error(p).
   std::shared_ptr<RegularHessianFactor<Base::Dim> > createHessianFactor(
-      const Cameras& cameras, const double lambda = 0.0,
+      const Cameras& cameras, const double _lambda = 0.0,
       bool diagonalDamping = false) const {
     size_t numKeys = this->keys_.size();
     // Create structures for Hessian Factors
@@ -230,7 +230,7 @@ protected:
 
     // build augmented hessian
     SymmetricBlockMatrix augmentedHessian =  //
-        Cameras::SchurComplement(Fs, E, b, lambda, diagonalDamping);
+        Cameras::SchurComplement(Fs, E, b, _lambda, diagonalDamping);
 
     return std::make_shared<RegularHessianFactor<Base::Dim> >(
         this->keys_, augmentedHessian);
@@ -238,9 +238,9 @@ protected:
 
   // Create RegularImplicitSchurFactor factor.
   std::shared_ptr<RegularImplicitSchurFactor<CAMERA> > createRegularImplicitSchurFactor(
-      const Cameras& cameras, double lambda) const {
+      const Cameras& cameras, double _lambda) const {
     if (triangulateForLinearize(cameras))
-      return Base::createRegularImplicitSchurFactor(cameras, *result_, lambda);
+      return Base::createRegularImplicitSchurFactor(cameras, *result_, _lambda);
     else
       // failed: return empty
       return std::shared_ptr<RegularImplicitSchurFactor<CAMERA> >();
@@ -248,9 +248,9 @@ protected:
 
   /// Create JacobianFactorQ factor.
   std::shared_ptr<JacobianFactorQ<Base::Dim, 2> > createJacobianQFactor(
-      const Cameras& cameras, double lambda) const {
+      const Cameras& cameras, double _lambda) const {
     if (triangulateForLinearize(cameras))
-      return Base::createJacobianQFactor(cameras, *result_, lambda);
+      return Base::createJacobianQFactor(cameras, *result_, _lambda);
     else
       // failed: return empty
       return std::make_shared<JacobianFactorQ<Base::Dim, 2> >(this->keys_);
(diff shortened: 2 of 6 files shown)


dellaert:
Thanks !!!
2 requests:
- parameter names should be the same in Python and c++, for doc generation to work, so please c++ names as well
- In c++, xxx_ means instance variable, so maybe rename to _xxx

jashshah999:
Thanks for the feedback!

For the parameter names (`lambda` → `_lambda`, `global` → `_global`), I can rename those in both the `.i` files and the C++ headers. However, `lambda` is used extensively as both a parameter name and local variable in `SmartFactorBase.h` (across `createHessianFactor`, `updateAugmentedHessian`, `createRegularImplicitSchurFactor`, `createJacobianQFactor`, `createJacobianSVDFactor` and their function bodies where it's passed to `Cameras::PointCov(E, lambda, ...)`). I'll do a full rename across those.

For the `lambda()` method on `LevenbergMarquardtOptimizer` -- renaming the C++ method itself would be a breaking API change. The wrapper already handles this by generating `self->lambda()` under a `_lambda` Python name. Should I still rename the C++ method, or keep it as-is with only the Python binding renamed?

Will push the updated changes shortly.

dellaert:
Please limit renaming in c++ to parameter names. C++ bodies are not visible in Python anyway.

jashshah999:
Updated! Changes:

- Switched to `_xxx` prefix (instead of `xxx_`) in both `.i` files and C++ headers
- Renamed `lambda` → `_lambda` in parameter names across `SmartProjectionFactor.h`, `SmartStereoProjectionFactor.h`, and their `.i` counterparts
- Renamed `global` → `_global` in `ReferenceFrameFactor.h` parameter name
- Renamed `LevenbergMarquardtOptimizer::lambda()` → `_lambda()` in `.h`, `.cpp`, and `.i`
- Updated body references within those functions to use the new parameter names
- Limited C++ changes to parameter names only (no local variable renames in function bodies)
- Updated expected wrapper test outputs (matlab + python)

ProfFan:
This is not right, renaming should  happen in the Python wrapper not C++ itself

jashshah999:
@ProfFan Could you clarify what you mean? @dellaert's earlier comment asked to "please [rename] c++ names as well" so that "parameter names should be the same in Python and c++, for doc generation to work", and then followed up with "Please limit renaming in c++ to parameter names."

So the current changes rename only C++ parameter names (e.g. `double lambda` → `double _lambda` in function signatures) to match the `.i` files -- not local variables or function bodies.

The one exception is `LevenbergMarquardtOptimizer::lambda()` which is a method name, not a parameter. Are you saying I should revert that method rename and let the wrapper handle the mapping instead? Happy to do that.

dellaert:
@ProfFan indeed, parameter names should match.
@jashshah999 we should not change the method name. Does that give problems in Python?

jashshah999:
Nope, no problems! The wrapper generator handles the mapping -- it will expose `lambda_` in Python while calling `self->lambda()` in C++. Reverted the method rename; only parameter names are changed now.

jashshah999:
Gentle ping @dellaert @ProfFan -- I believe the latest revision addresses all feedback (reverted the method rename, kept only parameter renames for the Python reserved keywords). Would appreciate a final review when you get a chance. Thanks!

ProfFan:
I don't think you need to change the names here, they should be handled in `wrap` only in the generated file:
```
    py::class_<gtsam::ManifoldEvaluationFactor<gtsam::Chebyshev1Basis, gtsam::Pose3>, gtsam::NoiseModelFactor, std::shared_ptr<gtsam::ManifoldEvaluationFactor<gtsam::Chebyshev1Basis, gtsam::Pose3>>>(m_, "ManifoldEvaluationFactorChebyshev1BasisPose3")
        .def(py::init<>())
        .def(py::init<gtsam::Key, const gtsam::Pose3&, const std::shared_ptr<gtsam::noiseModel::Base>, const size_t, double>(), py::arg("key"), py::arg("z"), py::arg("model"), py::arg("N"), py::arg("x"))
        .def(py::init<gtsam::Key, const gtsam::Pose3&, const std::shared_ptr<gtsam::noiseModel::Base>, const size_t, double, double, double>(), py::arg("key"), py::arg("z"), py::arg("model"), py::arg("N"), py::arg("x"), py::arg("a"), py::arg("b"));
```

ProfFan:
basically, ONLY the .i files.

dellaert:
> basically, ONLY the .i files.

No, @ProfFan , this is incorrect. Since about a year, since we added better docs generation, argument names have to match in c++ and python. So if we change them in .i we should do this in c++ as well.

dellaert (approved):

ProfFan:
Thanks Frank, I didn't realize that. However, this issue is a common task in FFI, which I don't think should be handled by renaming C++ variables. Python has a lot of reserved words, are we going to avoid all of them?

dellaert:
This was the approach taken by @p-zach - if there’s a better way, feel free to propose and implement :-)

p-zach:
Yes, IIRC the metadata on the wrapped functions is minimal, and without same-named parameters it is very difficult/impossible to guarantee a match between a given C++ function and the Python function that wraps it - which would mean you couldn't be sure the documentation for a C++ function is being assigned to the correct Python function. This is mostly an issue with overloaded methods. I worked at it for a while and that was the best I could find - probably there is a better way to do it, but it might take refactoring of the wrap library itself.

PR #2473, opened 2026-03-20
Fix link in PriorFactor documentation

nkhedekar:


dellaert:
Thanks ! Asked @rohan-bansal  to include in #2475

PR #2475, opened 2026-03-22
EqVIO Filter Implementation

rohan-bansal:
Changelog:
1. Previous PR that introduced VIO group/manifold also implemented retract and localCoordinates Jacobians in `ProductLiegroup-inl.h` for dynamic landmarks. CI was failing due to initialized Jacobians here, I patch this here.
2. Added state/output matrices according to the inverse-depth coordinate suite in the paper. I use InvDepth because the paper proves it can adequately preserve the equivariant output linearization while remaining numerically stable for distant points, and it is the path of least resistance for making the example work. The other two coordinate suites (euclid/polar) are currently not implemented; support will be added for Polar in a future PR.
3. Added `EqVIOFilter.h/cpp`, extending from `EquivariantFilter.h`. Made some changes to the base class to support syncing/re-binding filter internals once landmarks are added/removed, since this is an event where the model dimension changes. 
4. Added an `EqVIOFilterExample.cpp` that runs the filter on the first 10 seconds of the EuRoC MAV Vicon Room 1 dataset. I ran feature tracking independently and condensed input information into a CSV that is around 7MB in size. 

Working example output below!

```
CSV replay complete.
Events: 6602, IMU: 6001, vision frames: 601, vision features: 13316
Measurement noise variance (normalized): 1.77557e-05
Filter time: 1403715303.2621431
Landmarks: 32
Pose translation:   1.178667103   2.537093355 0.07318948356
GT pose translation:    1.1576   2.51599 0.0838289
Velocity: 0.0003766258116   -0.2772464545   0.05656939892
GT velocity: 0.0130831 -0.269437 0.0995449
```


rohan-bansal:
Ready for review @dellaert :)

dellaert:
I’m traveling so it might be a bit before I get to it.

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -48,10 +48,14 @@ ProductLieGroup<G, H> ProductLieGroup<G, H>::retract(const TangentVector& v,
         "ProductLieGroup::retract tangent dimension does not match product "
         "dimension");
   }
-  Jacobian1 D_g_first;
-  Jacobian1 D_g_second;
-  Jacobian2 D_h_first;
-  Jacobian2 D_h_second;
+  Jacobian1 D_g_first = Jacobian1::Zero(
```
dellaert: This is regrettable- does this only fail for certain compilers/versions?

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,193 @@
...
+  Eigen::Matrix<double, IMUInput::CompDim, IMUInput::CompDim> inputNoise =
+      Eigen::Matrix<double, IMUInput::CompDim, IMUInput::CompDim>::Identity() *
+      1e-3;
+};
+
+/**
+ * @brief Standalone EqVIO filter built on top of `EquivariantFilter`.
+ *
+ * The filter maintains an internal snapshot `(xi0, X, Sigma)`:
+ * - `xi0`: current linearization/reference state.
```
dellaert: Really? Why not xi_ref?

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,193 @@
...
+ * and state propagation (piecewise IMU holds), preserving the original EqVIO
+ * replay semantics while still using the base equivariant filter machinery.
+ */
+class GTSAM_UNSTABLE_EXPORT EqVIOFilter
+    : public EquivariantFilter<State, Symmetry> {
+ public:
+  using Base = EquivariantFilter<State, Symmetry>;
+
+  /// Internal filter state snapshot synchronized with `EquivariantFilter` base state.
+  struct Snapshot {
```
dellaert: Not very happy with this double storage. Let’s chat.

gtsam/navigation/EquivariantFilter.h
```diff
@@ -70,6 +70,45 @@ class EquivariantFilter : public ManifoldEKF<M> {
...
+  /**
+   * @brief Synchronization hooks for derived filters with dynamic runtime structure.
+   *
+   * Most fixed-dimension filters can rely on the public predict/update path only.
+   * These functions are primarily for dynamic cases (for example variable-size
+   * state/group representations) where derived classes need to resize/reset.
+   */
+
+  /// Reset reference, covariance, and group estimate; sync manifold state.
+  void resetReferenceAndGroup(const M& xi_ref, const CovarianceM& P,
```
dellaert, line 84: If we can avoid, would be better. And would we ever create a new xi_ref and a non-identity X? If we re-reference, would X not be identity?

gtsam_unstable/navigation/EqVIOFilter.h
rohan-bansal: Just a notation choice, I can change it to xi_ref.

gtsam_unstable/navigation/EqVIOFilter.h
rohan-bansal: I think it is possible to make this not an owned duplicate of data from EquivariantFilter. Will try it out.

gtsam/base/ProductLieGroup-inl.h
rohan-bansal: Yes, I noticed that it failed for the ubuntu-latest build, as a result of the `template predictwithJacobian(...)` in the EqVIOFilter. After backtracing a little bit the warning appears in compose/Logmap, which gets instantiated by the nested dynamic VioGroup during `propagateState` (`VioGroup::Logmap(...)`).

From my current understanding it is a false warning and zero-initializing fixes it. GCC can’t see that the Jacobians get assigned before they’re copied in this template path.

gtsam/navigation/EquivariantFilter.h
rohan-bansal: Some updates (landmark add/remove / dimension changes) change xi_ref for bookkeeping/structure, not to the full current estimate. In that case, X stays non-identity to preserve the same physical estimate. I think if we forced identity the estimate would jump.

rohan-bansal:
I removed the `Snapshot` struct tried to use the base `EquivariantFilter` as the single source of truth as much as possible.

Specifics:
- Removed the snapshot struct (the duplicated state/cov/group copy)
- Updated everything to read directly from:
  - `referenceState()`
  - `groupEstimate()`
  - `errorCovariance()`
  - `stateEstimate()`
- For dynamic landmark add/remove, I rebuild`(xi_ref, Sigma, X)` and apply them atomically with `resetReferenceAndGroup(...)`. I think this removes the necessity of maintaining a mirrored container struct.
- Was also able to prune some of my additions to `EquivariantFilter.h` as a result.

Let me know what you think about these changes @dellaert.

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+namespace eqvio {
+
+/**
+ * @brief Runtime parameters for the standalone EqVIO filter.
+ *
+ * These parameters mirror the process/measurement tuning knobs used by the
+ * EqVIO runtime implementation. All variances are per-axis scalar multipliers
+ * applied to identity blocks in the corresponding covariance matrices.
+ */
+struct GTSAM_UNSTABLE_EXPORT EqVIOFilterParams {
```
dellaert: Is export needed for header only?

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+ private:
+  EqVIOFilterParams params_;
+  bool initialized_ = false;
+
+ public:
+  EqVIOFilter();
+  /// Construct with explicit parameter bundle and default identity initial state.
+  explicit EqVIOFilter(const EqVIOFilterParams& params);
+  /// Construct with explicit initial reference state, covariance, and parameters.
+  EqVIOFilter(const State& xi0, const Matrix& Sigma0,
```
dellaert: Use the same parameter names as underlying filter?

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+   * assumption) and aligned with world +Z.
+   */
+  void initializeFromIMU(const IMUInput& imu);
+  
+  /**
+   * @brief Set reference state and covariance explicitly.
+   * @param xi_ref Reference state.
+   * @param Sigma0 Covariance.
+   */
+  void setReferenceState(const State& xi_ref, const Matrix& Sigma0);
```
dellaert: Why do we need this?

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+   * @brief Propagate filter state across a sequence of IMU hold intervals.
+   *
+   * Covariance is propagated once using time-weighted average IMU input, then
+   * state is propagated piecewise over each `(imuInputs[i], dts[i])` hold.
+   *
+   * @param imuInputs IMU samples defining zero-order holds.
+   * @param dts Hold durations (seconds), one per sample.
+   * @throws std::invalid_argument if input sizes differ.
+   */
+  void propagate(const std::vector<IMUInput>& imuInputs,
```
dellaert: We call this `predict` in our filter hierarchy

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * The method removes stale landmarks, rejects outliers, inserts new
+   * landmarks, performs update, and prunes numerically invalid landmarks.
+   *
+   * @param measurement Observed image points keyed by landmark id.
+   * @param camera Camera model used for projection/linearization.
+   * @param R Optional measurement covariance; if empty/wrong-sized, defaults to
+   *          `measurementNoiseVariance * I`.
+   */
+  void correct(const VisionMeasurement& measurement,
```
dellaert: And this, `update`

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+               const std::shared_ptr<const CameraModel>& camera,
+               const Matrix& R = Matrix());
+
+  /// Return the current full state estimate `phi(xi_ref, X)`.
+  State stateEstimate() const;
```
dellaert: Does that not exist in underlying filter ?

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+  /// True after IMU-based initialization.
+  bool isInitialized() const { return initialized_; }
+
+ private:
+  /// Allocate identity covariance with dimension `SensorState::CompDim + 3*nLandmarks`.
+  static Matrix defaultCovariance(size_t nLandmarks);
+
+  /// Propagate covariance through linearized error dynamics for one effective IMU segment.
+  void propagateCovariance(const IMUInput& imu, double dt);
```
dellaert: Why can EqF not do this for us?

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+  /// Propagate state estimate through lifted group increment for one IMU hold.
+  void propagateState(const IMUInput& imu, double dt);
```
dellaert: Same

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+
+  /// Build process noise matrix for current state dimension.
+  Matrix stateProcessNoise(size_t nLandmarks) const;
+
+  /// Add unseen landmarks from current measurement and expand group/covariance.
+  void addNewLandmarks(const VisionMeasurement& measurement,
+                       const std::shared_ptr<const CameraModel>& camera);
+  /// Remove landmark by contiguous index in `cameraLandmarks`.
+  void removeLandmarkByIndex(int idx);
```
dellaert: Do we need/use this?

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+  /// Remove landmark by integer id.
+  void removeLandmarkById(int id);
```
dellaert: In GTSAM we use Keys and/or Symbols

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,174 @@
...
+  /// Remove landmarks whose scale component leaves a numerically safe range.
+  void removeInvalidLandmarks();
+  /// Return 3x3 covariance block for a specific landmark id.
+  Matrix3 getLandmarkCovById(int id) const;
+  /// Return induced 2x2 output covariance for a specific landmark id.
+  Matrix2 outputCovarianceById(
+      int id, const std::shared_ptr<const CameraModel>& camera) const;
+
+  /// Run innovation update on currently matched measurements.
+  void update(const VisionMeasurement& measurement,
```
dellaert: Ah, diff with “correct”?

dellaert (commented):
It seems to me that there’s still a lot done in the derived class that should be done in the base class. One of the goals here is to demonstrate how the templates can be used with a minimum of additional code.

It’s good that the shadow state is gone!

gtsam_unstable/navigation/EqVIOFilter.h
rohan-bansal: removed. was used once in the example to init the filter, but is probably a good idea to prune it.

gtsam_unstable/navigation/EqVIOFilter.h
rohan-bansal: The paper was vague about how covariance/state was propagated, so I defaulted to the eqvio codebase implementation, which propagated covariance once and then propagated the state piecewise over each IMU hold. 

It is true that we can have the EqF do this for us. I changed the flow so that for each IMU hold we now call predict(lift, psi, etc.). This gets rid of a good chunk of code (which is nice), but also means the linearization is updated for each hold, which minorly changes the filter output; looks fine on the example, but will do broader tests later.

gtsam_unstable/navigation/EqVIOFilter.h
rohan-bansal: Yes, these are used by the update step of the filter, which along with the internal filter update, also does some landmark bookkeeping/management (removes stale landmarks, rejects outliers, inserts new landmarks as they become available)

dellaert (commented):
Cool, like where this is going. But I still have some doubts:
- I’d like to understand your comment about linearization better - does not match my EqF understanding
- it’s weird to me that the mathematical object `State` has knowledge about Keys. That seems like filter bookkeeping.

dellaert:
By the way, we also need the wrapper for the notebook in the awesome repo, right?

rohan-bansal:
I think I was able to factor out Keys from the State object, and move that bookkeeping logic into the filter and out of the math.

Sorry for ambiguity on the propagation stuff. In Section 7 of the paper, they mention:
> “Upon receiving an IMU signal, the EqF state and Ricatti equation are integrated without the correction terms.”

Looking at the actual code implementation linked [here](https://github.com/pvangoor/eqvio/blob/3bcc01a9501885d98d88c272229cf12ea7c7d3f3/src/mathematical/VIO_eqf.cpp#L62), there are two code paths. If fastRiccatti is enabled, they do one covariance/Riccati propagation over an IMU-averaged hold (frozen linearization over the window) and then propagates the observer state piecewise. If fastRiccatti is disabled, they propagate covariance and state per IMU hold. 

My earlier implementation followed the first path, but after heavily simplifying the filter in other places, I think the fastRiccatti path was unnecessary. Switching to the per-hold `predict(...)` call to the base class is cleaner, which does the same thing and recomputes local Jacobians each hold and advances mean/covariance together. The small shift in predicted output is expected I think, since it is the effect of moving from frozen-window to per-hold linearization.

rohan-bansal:
And yes, I will update with the Python wrapper bindings soon. I've been hacking on it for a little bit, but not reintroducing bloat in this regard is harder than I thought.

dellaert:
Re-request a review when ready.

rohan-bansal:
Ready for review @dellaert . The wrapper bindings are in. I was overthinking the implementation, and its simpler than I thought. If you would like to see the Python mirror of the C++ example, I put it in a gist [here](https://gist.github.com/rohan-bansal/e322f7a376fae74d50fa18f61c979bfa). I will probably move this to the AwesomeEqF repo.

gtsam_unstable/examples/EqVIOFilterExample.cpp
```diff
@@ -0,0 +1,199 @@
...
+ * @brief Run EqVIO filter replay on bundled EuRoC-derived CSV data.
+ */
+int main() {
+  const std::string csvPath =
+      findExampleDataFile("EqVIOdata_eurocmav_room1_10sec.csv");
+
+  try {
+    const replay::MetadataMap metadata = replay::readReplayMetadata(csvPath);
+    const EqVIOFilterParams params = paramsFromMetadata(metadata);
+    const State xi0 = initialReferenceState(metadata);
```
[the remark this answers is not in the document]
dellaert: @rohan-bansal ?

gtsam_unstable/examples/EqVIOFilterExample.cpp
```diff
@@ -0,0 +1,199 @@
...
+    bool gravityInitialized = false;
+    double currentTime = -1.0;
+    std::vector<IMUInput> imuBuffer;
+
+    replay::ReplayReader reader(csvPath);
+    replay::ReplayEvent event;
+    while (reader.next(event)) {
+      if (event.type == replay::ReplayEvent::Type::Imu) {
+        if (!filter) {
+          filter.emplace(params);
```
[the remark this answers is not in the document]
dellaert: @rohan-bansal again, I don't know what to do here

dellaert:
@rohan-bansal ball in your camp

gtsam_unstable/examples/EqVIOFilterExample.cpp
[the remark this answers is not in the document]
rohan-bansal: resolved.

rohan-bansal:
@dellaert I addressed the comments. I think reverting to Mahalanobis probabalistic gating is not necessary, the simplified version works well after fixing the xi0 initialization bug and Cheirality errors. 

I also verified the filter performance on the EqVIO EuRoC MAV dataset, and replaced the existing example data with a longer version that better shows how the filter performs :) Updated the top-level PR comment to display what the output looks like.

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,185 @@
...
+ * EqVIO runtime implementation. All variances are per-axis scalar multipliers
+ * applied to identity blocks in the corresponding covariance matrices.
+ */
+struct EqVIOFilterParams {
+  /// Initial landmark depth used when a new feature is first triangulated from
+  /// bearing.
+  double initialPointDepth = 10.0;
+  /// Initial 3x3 covariance scalar assigned to newly inserted landmarks.
+  double initialPointVariance = 1.0;
+  /// Default pixel-space measurement noise variance used when `correct()` is
```
dellaert: Why would we ever do that? If we don’t use this in examples or tests, cut down optionality. Is a general comment; AI tends to be too “inclusive”.

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,185 @@
...
+   *
+   * Each hold interval `(imuInputs[i], dts[i])` is propagated via the explicit
+   * base-class `predictWithJacobian(...)` path, so mean and covariance advance
+   * together.
+   *
+   * @param imuInputs IMU samples defining zero-order holds.
+   * @param dts Hold durations (seconds), one per sample.
+   * @throws std::invalid_argument if input sizes differ.
+   */
+  void predict(const std::vector<IMUInput>& imuInputs,
```
dellaert: If we just loop inside, I’d prefer just having predict(ImuInput,dt)  and have the example call multiple times.

gtsam_unstable/navigation/EqVIOFilter.h
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * @brief Apply one visual correction step with dynamic landmark management.
+   *
+   * The method removes stale landmarks, rejects outliers, inserts new
+   * landmarks, performs update, and prunes numerically invalid landmarks.
+   *
+   * @param measurement Observed image points keyed by landmark id.
+   * @param camera Camera model used for projection/linearization.
+   * @param R Optional measurement covariance; if empty/wrong-sized, defaults to
```
dellaert: Don’t allow empty, wrong size. Can we not take a MatrixN with N appropriate fixed dim?

dellaert (commented):
Some more comments and I’m also rereading eqvio paper because I want to know the relationship with current inverse depth impl in GTSAM

gtsam_unstable/navigation/EqVIOFilter.h
rohan-bansal: These are used in the examples. initialPointDepth is used to initialize newly observed landmarks along the measured bearing before they have triangulated depth, and initialPointVariance is used to seed the covariance block for those newly inserted landmarks.

General comment noted though.

rohan-bansal:
@dellaert I took a look at the InverseDepth implementations existing in GTSAM, which seem to mostly be factor-graph machinery. It represents landmarks as inverse-depth variables and computes reprojection-factor errors. In EqVIO, landmarks are still stored as `Point3` but the error coordinates inverse-depth-like. 

Replacing the EqVIO chart with GTSAM’s inverse-depth representation would mean that we need to factor out specific inverse-depth logic into geometric helpers, and also do a bigger refactor of state layout, covariance blocks, landmark initialization, and all the EqF Jacobians.


Besides this, I addressed your comments and tightened the codebase in some other areas after doing another sweep. 

Also, I am going to context-switch and work on AwesomeEqF in parallel from this current code snapshot.

dellaert (approved):
Cool! LGTM!

PR #2476, opened 2026-03-23
Steiner trees for Covariance Recovery

dellaert:
This PR contributes two main contributions:
1. a paper describing the existing covariance recovery mechanisms in the Bayes tree;
2. two improvements suggested by Codex 5.4, namely (a) generalizing multivariable queries using Steiner trees, and (b) a partial covariance block recovery rather than a full inversion of a clique's covariance matrix.


gtsam/linear/GaussianBayesTree.cpp
```diff
@@ -119,7 +119,13 @@ namespace gtsam {
...
   Matrix GaussianBayesTree::marginalCovariance(Key key) const
   {
-    return marginalFactor(key)->information().inverse();
+    const Matrix information = marginalFactor(key)->information();
+    if (!information.allFinite()) {
+      return Matrix::Zero(information.rows(), information.cols());
+    }
+    const Matrix identity =
+        Matrix::Identity(information.rows(), information.cols());
+    return information.selfadjointView<Eigen::Upper>().llt().solve(identity);
```
ProfFan: Can we use solveInPlace here? This is LAPACK's `dpotri`.

gtsam/nonlinear/Marginals.cpp
```diff
@@ -13,18 +13,171 @@
...
+
+/* ************************************************************************* */
+Matrix informationToCovariance(const Matrix& information) {
+  if (!information.allFinite()) {
+    return Matrix::Zero(information.rows(), information.cols());
+  }
+
+  const Matrix identity =
+      Matrix::Identity(information.rows(), information.cols());
+  return information.selfadjointView<Eigen::Upper>().llt().solve(identity);
```
ProfFan: ditto

gtsam/nonlinear/Marginals.cpp
```diff
@@ -13,18 +13,171 @@
...
+    const size_t begin = offsets[blockIndex];
+    const size_t dim = dims[blockIndex];
+    selectors.block(begin, selectedOffset, dim, dim).setIdentity();
+    selectedOffset += dim;
+  }
+
+  Matrix intermediate =
+      R.transpose().triangularView<Eigen::Lower>().solve(selectors);
+  return R.triangularView<Eigen::Upper>().solve(intermediate);
+}
```
ProfFan, line 148: ditto

ProfFan (approved):
Looks great to me :shipit: I think the only problem is the templated magic should have more documentation, but that can be a future improvement.

Ovardo (commented):
Ref. recent GTSAM forum thread.

I’m still not fully convinced about the treatment of the queried-clique conditional in CovarianceRecovery.pdf.

Unless I’m misunderstanding something, my reasoning is the following:

From Eqs. (17)–(18), the interface sets A_{u,b} and B_{u,b} are constructed from separator variables along the path. Under the usual Bayes tree construction, the frontal variables of the queried clique u do not appear in ancestor cliques, so it seems that F_u = F_{d_0} is not contained in A_{u,b} \cup B_{u,b}.

If that is right, then from Eq. (19), F_u \subseteq Z_{u,b}. But Eq. (20) integrates over Z_{u,b}, so it appears that the frontal variables of the queried clique are integrated out in the shortcut \chi_{u->b}.

That is where I get confused about Eq. (21): the final integral seems to treat q_1 and q_2 as variables still present in the integrand, but if they are frontal variables of the queried cliques, it looks like they would already have been integrated out by Eq. (20).

So I may be missing something in the intended role of the descendant-side interface B_{u,b} in Eq. (18), but at the moment I don’t see how the queried clique’s frontal variables are retained.

Feel free to let me know where my reasoning goes wrong.

dellaert:
@Ovardo I had another look and you're right! I updated the PDF with an example, and what I think are now the correct definitions for A and B. Let me know what you think. I'm actually on travel so I'm doing this quickly in between activities - so they could still be wrong :-/ I appreciate your help!

Ovardo:
I've had a look, and to me it now looks consistent. 

One could perhaps mention that in the edge case(s) of one or both of the queried cliques themselves being the lowest common ancestor, then one should not include the respective clique potentials in the final pairwise assembly, as they are already embedded in the ancestor clique marginal.

Thanks again for updating this — I really appreciate it.

PR #2477, opened 2026-03-23
Add optional ecef_T_nav to PseudorangeFactorArm for GNSS/IMU coupling

inuex35:
Follow-up to #2447.

As @masoug and @dellaert discussed in that PR, having ENU/NED support in the pseudorange factor would be useful for sensor fusion. I went ahead and added an optional `Pose3 ecef_T_nav` parameter to `PseudorangeFactorArm`
and `DifferentialPseudorangeFactorArm`.

The idea is simple: if you pass `ecef_T_nav`, the pose key is treated as a local nav frame pose, and the factor does `ecef_T_body = ecef_T_nav * nav_T_body` internally with proper chain-rule Jacobians. If you don't pass it, everything
works exactly as before in ECEF. This way users can pick ENU, NED, or whatever frame they want without the factor needing to know.

If this approach looks good, I'd like to work on carrier phase factors next.


dellaert (approved):

dellaert:
Thanks !!!

Issue #2489, opened 2026-04-05
MATLAB macOS R2025b: exit crash after nonlinear optimization when GTSAM is built with TBB


dellaert:
## Description

MATLAB exits with a segmentation fault after running nonlinear optimization through the MATLAB wrapper when GTSAM is built with `GTSAM_WITH_TBB=ON` on macOS.

The same installed toolbox exits cleanly if no optimization is run, and rebuilding GTSAM with `GTSAM_WITH_TBB=OFF` removes the crash.

## Steps to reproduce

1. Configure and install GTSAM with the MATLAB toolbox enabled and TBB enabled.

```bash
cmake -S . -B build \
  -DGTSAM_INSTALL_MATLAB_TOOLBOX=ON \
  -DGTSAM_WITH_TBB=ON
cmake --build build -j6
cmake --install build
```

2. Start MATLAB and add the installed toolbox root to the path.

3. Run this minimal repro:

```matlab
restoredefaultpath
addpath('/Users/dellaert/gtsam_toolbox')
import gtsam.*

priorNoise = noiseModel.Diagonal.Sigmas([0.3;0.3;0.1]);
odometryNoise = noiseModel.Diagonal.Sigmas([0.2;0.2;0.1]);

graph = NonlinearFactorGraph;
graph.add(PriorFactorPose2(1, Pose2(0,0,0), priorNoise));
graph.add(BetweenFactorPose2(1,2,Pose2(2,0,0), odometryNoise));
graph.add(BetweenFactorPose2(2,3,Pose2(2,0,0), odometryNoise));

initial = Values;
initial.insert(1, Pose2(0.5,0,0.2));
initial.insert(2, Pose2(2.3,0.1,-0.2));
initial.insert(3, Pose2(4.1,0.1,0.1));

optimizer = LevenbergMarquardtOptimizer(graph, initial);
result = optimizer.optimizeSafely();
exit
```

4. MATLAB crashes during exit with a segmentation fault.

## Expected behavior

MATLAB should exit cleanly after optimization.

## Environment

- GTSAM commit: `23147828906c2df50c4772f2863de332c7d92f67`
- MATLAB: `25.2.0.3177638 (R2025b) Update 5`
- OS: macOS `26.3.2` on Apple Silicon
- Installed toolbox path used for repro: `/Users/dellaert/gtsam_toolbox`

## Additional information

- `restoredefaultpath; addpath('/Users/dellaert/gtsam_toolbox'); exit` exits cleanly.
[...]

PR #2490, opened 2026-04-05
Fix matlab wrapper

dellaert:
This PR fixes a few MATLAB wrapper regressions and updates the MATLAB install guidance.

Main changes:
- add the missing `gtsam/geometry/cameras.i` to the MATLAB wrapper inputs, which restores the camera calibration classes to the generated toolbox
- fix the `SmartFactorBase::cameras(const Values&)` wrapper declaration in `slam.i` to use `gtsam::CameraSet<CAMERA>`, matching the actual C++ return type
- improved `wrap/cmake/MatlabWrap.cmake` MATLAB-side comments and runtime linking behavior so the installed mex can find `libgtsam` reliably at load time
- make `matlab/gtsam_tests/test_gtsam.m` only run serialization tests when the relevant serialization methods are actually present in the generated toolbox
- fix the legacy `gtsamExamples` GUI labels, which newer MATLAB versions were rendering with literal HTML tags
- rewrite `matlab/README.md` around the recommended install flow for non-developers:
  - use `cmake --install .`
  - install the toolbox outside the build tree
  - add only the toolbox root to the MATLAB path
  - document the current macOS/TBB shutdown issue for MATLAB users

I also opened [#2489](https://github.com/borglab/gtsam/issues/2489) for the separate MATLAB exit crash after optimization when GTSAM is built with TBB on macOS.

Local verification:
- rebuilt and installed the MATLAB toolbox
- exercised the MATLAB toolbox and examples manually


dellaert:
FYI @varunagrawal  - asked @ProfFan to review as he and I talked about fixing this.

varunagrawal:
> FYI @varunagrawal  - asked @ProfFan to review as he and I talked about fixing this.

Can I get some more context since I wasn't a part of those discussions?

dellaert:
> > FYI @varunagrawal  - asked @ProfFan to review as he and I talked about fixing this.
> 
> Can I get some more context since I wasn't a part of those discussions?

Was broken, somebody needs to fix it. That’s it.

ProfFan (approved):
The AI is doing a great job  🌵

Issue #2492, opened 2026-04-06
Cannot Retrieve Jacobians from Pose3 Methods in MATLAB Wrapper


dellaert:
## Description:

When calling methods like `transformFrom` or `transformPoseTo` on `gtsam.Pose3` from MATLAB and attempting to retrieve Jacobians, there seems to be no way to pass valid Jacobian output arguments that the wrapper accepts. 

## Example:
```matlab
analytic_H1 = [];
analytic_H2 = [];
[y, analytic_H1, analytic_H2] = T.transformFrom(P, analytic_H1, analytic_H2);
```
yields 
```
Error using gtsam.Pose3/transformFrom
Arguments do not match any overload of function gtsam.Pose3.transformFrom
```

## Expected behavior:
It should be possible to retrieve Jacobians from these methods in MATLAB. The documentation says:
```
help(gtsam.Pose3):
 ...
 transformPoseFrom(Pose3 aTb) : returns gtsam::Pose3
 transformPoseFrom(Pose3 pose, Ref<MatrixXd> Hself, Ref<MatrixXd> HaTb) : returns gtsam::Pose3
 ...
```

## Actual behavior:
No combination of arguments is accepted by the wrapper for Jacobian outputs.

dellaert:
@varunagrawal @ProfFan

dellaert:
Might need special treatment in wrap - as I don;t think it's fixable in GTSAM

thatdudegrantt:
I've tracked down the root cause and opened a fix in [ borglab/wrap: 189](https://github.com/borglab/wrap/pull/189)

**Root cause:** `Eigen::Ref<Eigen::MatrixXd>` is parsed by gtwrap as a `TemplatedType` with `typename.name == 'Ref'`, not a plain `Type` with `is_ref` set. The MATLAB wrapper codegen had no special handling for this pattern and emitted `isa(varargin{N}, 'Eigen.RefMatrixXd')` in the dispatch check. However, `Eigen.RefMatrixXd` is not a MATLAB class, so this always returns false, making every Jacobian overload permanently unreachable. On the C++ MEX side, the wrapper also tried to `unwrap_shared_ptr<MatrixXd>` from `in[]`, which would have segfaulted even if dispatch had somehow succeeded.

**Fix (in borglab/wrap):**
- Adds `is_eigen_ref()` to detect `Ref<MatrixXd>` TemplatedType args
[...]

PR #2493, opened 2026-04-06
numerical derivatives in MATLAB

dellaert:
To support creating custom factors, I added the MATLAB equivalent of numerical derivatives

Both matlab/README and matlab/gtsam_tests/testNumericalDerivative.m have an example, and it works. 

I created a new issue (#2492), as it seems we cannot retrieve analytical derivatives via our normal output argument mechanism - which works in Python - but it does not work in MATLAB. This should not prevent people from checking the numerical derivatives for using the custom factors, however.

### DIFF
--- a/matlab/README.md
+++ b/matlab/README.md
@@ -194,6 +194,9 @@ actual_H = factor.linearize(initial).jacobian();
 EQUALITY('custom jacobian', expected_H, actual_H, 1e-5);
 ```
 
+Wrapped overloaded methods such as `factor.unwhitenedError` still need an
+explicit lambda when passed as callbacks in MATLAB.
+
 - Optimization usually exercises both the residual and Jacobian paths.
 - Deleting the MATLAB factor removes its callback from the MATLAB-side registry, so repeated tests in one session do not accumulate stale entries.
 - Once a MATLAB `CustomFactor` has been constructed, the mex module is intentionally kept loaded for the rest of the MATLAB session to avoid unload-time crashes while callback-backed native factors are still reachable from C++.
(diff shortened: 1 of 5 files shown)


dellaert:
FYI @meyer-UCSD

ProfFan (approved):

matlab/gtsam_tests/testNumericalDerivative.m
```diff
@@ -0,0 +1,77 @@
...
+analytic_H1 = zeros(3,6);
+analytic_H1(:,1:3) = R * (-skew); % leftCols<3>()
+analytic_H1(:,4:6) = R;           % rightCols<3>()
+analytic_H2 = R;
+
+numerical_H1 = numericalDerivative.numericalDerivative21(h, T, P);
+numerical_H2 = numericalDerivative.numericalDerivative22(h, T, P);
+
+EQUALITY('test_numericalDerivative_with_pose_H1', analytic_H1, numerical_H1, 1e-5);
+EQUALITY('test_numericalDerivative_with_pose_H2', analytic_H2, numerical_H2, 1e-5);
```
varunagrawal: Add a new line please?

matlab/README.md
```diff
@@ -185,6 +185,15 @@ Rules for MATLAB `CustomFactor` callbacks:
 Practical notes:
 
 - `unwhitenedError(...)` exercises the residual-only path.
+- Jacobians should always be verified using numerical derivatives during development. The `gtsam.numericalDerivative` utility is provided for this purpose:
+
+```matlab
+% Verify Jacobians using numerical derivative
+expected_H = numericalDerivative.numericalDerivative11(@(v) factor.unwhitenedError(v), initial);
```
varunagrawal, line 192: Does `expected_H = numericalDerivative.numericalDerivative11(factor.unwhitenedError, initial);` work?
Putting `factor.unwhitenedError` inside a lambda seems a bit pointless if we can pass the function directly?

varunagrawal (approved):
Two minor comments, otherwise LGTM.

PR #2494, opened 2026-04-08
Add group action param to Product Lie Group

rohan-bansal:
I tried adding the group action parameter by way of a sibling class to ProductLieGroup, but the extra re-implementation felt unnecessary. 

The version I ran with is implemented directly into the ProductLieGroup, keeping all the branches/if-checking in the internals, and exposing a lean public template/API.


dellaert:
Awesome start! Some quick comments:
1) I remember making some changes to this class template in the other PR. I think they were only formatting changes, but please make sure. Note, we format with clang-format, Google style. 
2) I don't think we need `DirectProductAction`. The code does not actually use it now in the "direct" branches. Please check.
3) So, we can do this: if no action is given, i.e. it's a dummy argument, it's assumed to be a direct group.

jenniferoum:
Looks good to me. I was able to pull the code and run a unit test with the SD group.

gtsam/base/ProductLieGroup.h
```diff
@@ -26,16 +27,19 @@
...
 #include <vector>
 
 namespace gtsam {
 
 /**
  * @brief Template to construct the product Lie group of two other Lie groups
- * Assumes Lie group structure for G and H
+ * Assumes Lie group structure for G and H. If Action is omitted the group is
+ * the direct product G x H. If Action is provided the group is the left
+ * semidirect product G ⋉ H.
```
dellaert: Maybe a little bit more guide on what that means would be useful

tests/testActionProductLieGroup.cpp
```diff
@@ -0,0 +1,348 @@
...
+struct Rot3VectorAction : public GroupAction<Rot3VectorAction, Rot3, Vector3> {
+  static constexpr ActionType type = ActionType::Left;
+
+  Vector3 operator()(const Rot3& R, const Vector3& t,
+                     OptionalJacobian<3, 3> HR = {},
+                     OptionalJacobian<3, 3> Ht = {}) const {
+    return R.rotate(t, HR, Ht);
+  }
+
+  template <typename ProductType>
```
dellaert, line 41: Seems overly templated. All types are known?

tests/testActionProductLieGroup.cpp
```diff
@@ -0,0 +1,348 @@
...
+    Matrix6 Hpose;
+    Vector6 xi;
+    xi << w, rho;
+    const Pose3 pose = Pose3::Expmap(xi, H1 || H2 ? &Hpose : nullptr);
+    if (H1) *H1 = Hpose.leftCols<3>();
+    if (H2) *H2 = Hpose.rightCols<3>();
+    return ProductType(pose.rotation(), pose.translation());
+  }
+
+  template <typename ProductType>
```
dellaert, line 55: Same

tests/testActionProductLieGroup.cpp
```diff
@@ -0,0 +1,348 @@
...
+  static typename ProductType::TangentVector Logmap(
+      const ProductType& p, typename ProductType::ChartJacobian H = {}) {
+    Matrix6 Hpose;
+    const Vector6 xi =
+        Pose3::Logmap(Pose3(p.first, p.second), H ? &Hpose : nullptr);
+    if (H) *H = Hpose;
+    return xi;
+  }
+
+  template <typename ProductType>
```
dellaert: Same?

tests/testActionProductLieGroup.cpp
```diff
@@ -0,0 +1,348 @@
...
+}
+
+Pose3 asPose3(const Semidirect& state) {
+  return Pose3(state.first, state.second);
+}
+
+}  // namespace
+
+/* ************************************************************************* */
+// Verify the dummy-action specialization matches the original direct-product
```
dellaert: I would think that the direct product was already unit tested quite well so I am not sure if this is needed.

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -20,162 +20,221 @@
...
 
-template <typename G, typename H>
-ProductLieGroup<G, H> ProductLieGroup<G, H>::operator*(
+template <typename G, typename H, typename Action>
+ProductLieGroup<G, H, Action> ProductLieGroup<G, H, Action>::operator*(
     const ProductLieGroup& other) const {
   checkMatchingDimensions(other, "operator*");
-  return ProductLieGroup(traits<G>::Compose(this->first, other.first),
-                         traits<H>::Compose(this->second, other.second));
+  if constexpr (isDirectProduct) {
```
dellaert, line 27: I’m wondering whether about this file (and the diff!) would be cleaner if instead of branching on this constexpr we would just use partial specialization on void. Unless when there is a lot of common code in the body.

dellaert: We might not even need isDirectProduct if we do this for all methods.

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -20,162 +20,221 @@
...
-  return ProductLieGroup(traits<G>::Inverse(this->first),
-                         traits<H>::Inverse(this->second));
+template <typename G, typename H, typename Action>
+ProductLieGroup<G, H, Action> ProductLieGroup<G, H, Action>::inverse() const {
+  if constexpr (isDirectProduct) {
+    return ProductLieGroup(traits<G>::Inverse(this->first),
+                           traits<H>::Inverse(this->second));
+  } else {
+    const Action action{};
+    const G gInv = traits<G>::Inverse(this->first);
```
dellaert: These could be outside if-then-else

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -20,162 +20,221 @@
...
+  if constexpr (isDirectProduct) {
+    Jacobian1 D_g_first;
+    Jacobian1 D_g_second;
+    Jacobian2 D_h_first;
+    Jacobian2 D_h_second;
+    G g = traits<G>::Retract(
+        this->first, tangentSegment<G>(v, 0, firstDimension),
+        H1 ? &D_g_first : nullptr, H2 ? &D_g_second : nullptr);
+    H h = traits<H>::Retract(
+        this->second, tangentSegment<H>(v, firstDimension, secondDimension),
```
dellaert: Let’s just maybe use d1 and d2 everywhere- easier to read

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -20,162 +20,221 @@
...
+    }
+    if (H2) {
+      *H2 = zeroJacobian(productDimension);
+      H2->block(0, 0, firstDimension, firstDimension) = D_g_second;
+      H2->block(firstDimension, firstDimension, secondDimension,
+                secondDimension) = D_h_second;
+    }
+    return ProductLieGroup(g, h);
+  } else {
+    Jacobian D_expmap;
```
dellaert: So, semi is way simpler??

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -20,162 +20,221 @@
...
+    }
+    if (H2) {
+      *H2 = zeroJacobian(productDimension);
+      H2->block(0, 0, firstDimension, firstDimension) = D_g_second;
+      H2->block(firstDimension, firstDimension, secondDimension,
+                secondDimension) = D_h_second;
+    }
+    return makeTangentVector(v1, v2, firstDimension, secondDimension);
+  } else {
+    const ProductLieGroup relative = between(g);
```
dellaert: Same: semi is simpler?

dellaert (commented):
Left some comments. And also love to see a little bit more comment in the code on how semi product is different. And it suspicious that in many cases the semi branch seems simpler.

dellaert:
@rohan-bansal ping

rohan-bansal:
Sorry for the delay @dellaert , have been feeling a bit under the weather.

Changes:
- Used the simpler generic function implementations where possible instead of having if-else branches
- Commented the direct/semi-direct paths to explain better
- Addressed misc. comments

The one thing I was not able to do that I would appreciate some guidance on is making the Expmap/Logmap generic for the semi-direct product. 

For a semidirect product `G ⋉ H`, the exponential map is `Expmap(ξ, v) = (Expmap_G(ξ), J_left(ξ) · v)` where `J_left(ξ) = ∫₀¹ D_h φ(Expmap_G(tξ), e_H) dt`. From what I understand after talking a bit with Claude, there is no closed form solution for this integral that can be derived generically, it would need a group-specific formula or a very computationally-expensive numerical quadrature based method. I think the simplest thing to do is have the action supply Expmap/Logmap, but let me know if you think I should do it differently.

dellaert:
I’ll email - I might have just the thing.

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -233,86 +227,233 @@ ProductLieGroup<G, H> ProductLieGroup<G, H>::Expmap(const TangentVector& v,
...
-  G g = traits<G>::Expmap(v1, H1 ? &D_g_first : nullptr);
-  H h = traits<H>::Expmap(v2, H2 ? &D_h_second : nullptr);
-  if (H1) {
-    *H1 = Matrix::Zero(productDimension, firstDimension);
-    H1->block(0, 0, firstDimension, firstDimension) = D_g_first;
+  if constexpr (isDirectProduct) {
+    // Direct product: the two Lie algebras are independent, so the exponential
+    // is componentwise: exp(v₁,v₂) = (exp_G(v₁), exp_H(v₂)).
+    // The Jacobian columns split cleanly between the G and H blocks.
+    const size_t firstDimension = static_cast<size_t>(v1.size());
```
dellaert, line 240: d1, d2?

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -18,164 +18,157 @@
...
 #pragma once
 
+#include <unsupported/Eigen/MatrixFunctions>
+
 namespace gtsam {
 
-template <typename G, typename H>
-ProductLieGroup<G, H> ProductLieGroup<G, H>::operator*(
+// ---------------------------------------------------------------------------
+// phi01Kernel: compute φ₀(A)=exp(A) and φ₁(A)=Σ Aᵏ/(k+1)! in one shot.
```
dellaert, line 26: Pretty cool. But is this cheaper?
Also - maybe in a next PR - we need to think about a mechanism to special-case Rot3 and other G groups we have closed forms for.

dellaert (commented):
Good progress!!! 2 comments

gtsam/base/ProductLieGroup-inl.h
rohan-bansal: Good point, this is not intended to be cheaper than closed forms like the Rot3 kernels, and updated the comment to make that explicit. I agree that we should special-case these groups in a separate PR.

gtsam/base/ProductLieGroup-inl.h
rohan-bansal: I interpreted this as "why not call these dimensions d1, d2?" so I made that change. No behavior change overall, maybe more consistent. Let me know if you meant something else.

dellaert:
Status?

rohan-bansal:
From my end it looks ready for re-review. I did my best to be faithful to the PDFs, hopefully there are no mistakes; I would hope that the tests catch all of them. Just added another generality test with Rot2 x R2.

dellaert:
My comment on `phi01Kernel` was that it does an expm on a larger matrix. We also don't always need both. So I'd prefer two separate phi0 and phi1 kernels.

rohan-bansal:
Oh I see, I misinterpreted. Will make this change

rohan-bansal:
Updated.

rohan-bansal:
Replaced numerical code with analytical.

dellaert (approved):
LGTM !

PR #2499, opened 2026-04-11
Generalize AttitudeFactor

dellaert:
This is the first of a few PRs that generalize sub-state factors, like AttitudeFactor, to types other than Rot3 and Pose3. By templating AttitudeFactor using CRTP, we now have:
```
using Rot3AttitudeFactor = AttitudeFactor<Rot3>;
using Pose3AttitudeFactor = AttitudeFactor<Pose3>;
using NavStateAttitudeFactor = AttitudeFactor<NavState>;
using Gal3AttitudeFactor = AttitudeFactor<Gal3>;

template <int K>
using ExtendedPose3AttitudeFactor = AttitudeFactor<ExtendedPose3<K>>;

using Se23AttitudeFactor = AttitudeFactor<Se23>;
using ExtendedPose3dAttitudeFactor = AttitudeFactor<ExtendedPose3d>;
```

### DIFF
--- a/gtsam/navigation/navigation.i
+++ b/gtsam/navigation/navigation.i
@@ -399,6 +399,78 @@ virtual class Pose3AttitudeFactor : gtsam::NoiseModelFactor {
   void serialize() const;
 };
 
+virtual class NavStateAttitudeFactor : gtsam::NoiseModelFactor {
+  NavStateAttitudeFactor(gtsam::Key key, const gtsam::Unit3& nRef,
+                         const gtsam::noiseModel::Diagonal* model,
+                         const gtsam::Unit3& bMeasured);
+  NavStateAttitudeFactor(gtsam::Key key, const gtsam::Unit3& nRef,
+                         const gtsam::noiseModel::Diagonal* model);
+  NavStateAttitudeFactor();
+  void print(string s = "", const gtsam::KeyFormatter& keyFormatter =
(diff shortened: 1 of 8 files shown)


gtsam/navigation/navigation.i
```diff
@@ -399,6 +399,78 @@ virtual class Pose3AttitudeFactor : gtsam::NoiseModelFactor {
...
+  bool equals(const gtsam::NonlinearFactor& expected, double tol) const;
+  gtsam::Unit3 nRef() const;
+  gtsam::Unit3 bMeasured() const;
+  gtsam::Vector evaluateError(const gtsam::Se23& state);
+
+  // enable serialization functionality
+  void serialize() const;
+};
+
+virtual class ExtendedPose3dAttitudeFactor : gtsam::NoiseModelFactor {
```
varunagrawal, line 456: You can use template definitions in the wrapper as well and define the custom names using `typedef`s.

varunagrawal (approved):
Very cool. This makes it very clean.

Side note: I am curious about the difference between Se23 and ExtendedPose3d since they both are the same in the literature.

ProfFan (approved):
:shipit:

gtsam/navigation/navigation.i
dellaert: Thanks ! I'll try that in a follow-up PR.

PR #2502, opened 2026-04-13
feat: double difference factor for rtk gnss

inuex35:
## Add DDPseudorangeFactor and DDCarrierPhaseFactor (with and without lever-arm)

Following up on #2447, this PR adds four double-differenced GNSS factors:

- **`DDPseudorangeFactor`** – DD code pseudorange, takes a Point3 (ECEF antenna position)
- **`DDPseudorangeFactorArm`** – same, but takes a Pose3 in the navigation frame + body-frame lever arm
- **`DDCarrierPhaseFactor`** – DD carrier-phase, takes a Point3 (ECEF antenna position), with ambiguity variable `N`
- **`DDCarrierPhaseFactorArm`** – same, but takes a Pose3 in the navigation frame + body-frame lever arm, with ambiguity variable `N`

Factors without the `Arm` suffix take an ECEF antenna position directly. The `Arm` variants take a Pose3 in the navigation frame and compute the antenna position via a fixed body-frame lever arm.

I'd appreciate feedback on whether this structure fits cleanly into the existing factor hierarchy. @masoug @dellaert

Related: #2447, #2477


gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+ * @ingroup navigation
+ */
+class GTSAM_EXPORT CarrierPhaseFactor
+    : public NoiseModelFactorN<Point3, double, double>,
+      private CarrierPhaseBase {
+ private:
+  typedef NoiseModelFactorN<Point3, double, double> Base;
+
+ public:
+  using Base::evaluateError;
```
varunagrawal, line 49: No need for this since you've overridden `evaluateError` later.

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+ * @ingroup navigation
+ */
+class GTSAM_EXPORT CarrierPhaseFactorArm
+    : public NoiseModelFactorN<Pose3, double, double>,
+      private CarrierPhaseBase {
+ private:
+  typedef NoiseModelFactorN<Pose3, double, double> Base;
+
+  Point3 bL_;  ///< Lever arm from body origin to antenna in body frame.
+  std::optional<Pose3> ecef_T_nav_;  ///< Optional ECEF-from-nav transform.
```
varunagrawal, line 141: You could combine this and `CarrierPhaseFactor` by providing the `leverArm` as an optional arg in the constructor. We do a similar thing in the `ImuFactor` if you need a reference.

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+#endif
+};
+
+/// traits
+template <>
+struct traits<CarrierPhaseFactorArm>
+    : public Testable<CarrierPhaseFactorArm> {};
+
+/**
+ * DD carrier phase factor.
```
varunagrawal: Can you rewrite this as `Double Difference Carrier Phase factor`?

I would also prefer naming the factor `DoubleDiffCarrierPhaseFactor`.

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+  void print(const std::string& s = "", const KeyFormatter& keyFormatter =
+                                            DefaultKeyFormatter) const override {
+    std::cout << (s.empty() ? "" : s + " ") << "DDCarrierPhaseFactor\n";
+    std::cout << "  lam: " << lam_ << "\n";
+    Base::print("", keyFormatter);
+  }
+
+  bool equals(const NonlinearFactor& expected, double tol = 1e-9) const override {
+    const This* e = dynamic_cast<const This*>(&expected);
+    return e != nullptr && Base::equals(*e, tol) &&
```
varunagrawal: Move this to .cpp file please

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+           traits<Point3>::Equals(satTargetBase_, e->satTargetBase_, tol) &&
+           traits<Point3>::Equals(basePos_, e->basePos_, tol);
+  }
+
+  Vector evaluateError(const Point3& pos,
+                       const double& ambRef, const double& ambTarget,
+                       OptionalMatrixType Hpos,
+                       OptionalMatrixType HambRef,
+                       OptionalMatrixType HambTarget) const override {
+    const double ddObs = (cpRovRef_ - cpBaseRef_) - (cpRovTarget_ - cpBaseTarget_);
```
varunagrawal: Move this to .cpp file please

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+  double cpRovTarget_;
+  double cpBaseTarget_;
+  Point3 satRefRov_;
+  Point3 satTargetRov_;
+  Point3 satRefBase_;
+  Point3 satTargetBase_;
+  Point3 basePos_;
+  double lam_;
+  Point3 bL_;
+  std::optional<Pose3> ecef_T_nav_;
```
varunagrawal, line 338: You can similarly merge this and the previous factor.

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+
+  void print(const std::string& s = "", const KeyFormatter& keyFormatter =
+                                            DefaultKeyFormatter) const override {
+    std::cout << (s.empty() ? "" : s + " ") << "DDCarrierPhaseFactorArm\n";
+    std::cout << "  lam: " << lam_ << "\n";
+    Base::print("", keyFormatter);
+  }
+
+  bool equals(const NonlinearFactor& expected, double tol = 1e-9) const override {
+    const This* e = dynamic_cast<const This*>(&expected);
```
varunagrawal: Move this to .cpp file please

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+    if (ecef_T_nav_ && !ecef_T_nav_->equals(*e->ecef_T_nav_, tol)) return false;
+    return true;
+  }
+
+  Vector evaluateError(const Pose3& pose,
+                       const double& ambRef, const double& ambTarget,
+                       OptionalMatrixType H_pose,
+                       OptionalMatrixType HambRef,
+                       OptionalMatrixType HambTarget) const override {
+    Matrix66 H_compose;
```
varunagrawal: Move this to .cpp file please

gtsam/navigation/GnssCommon.h
```diff
@@ -0,0 +1,38 @@
...
+ * @param[in]  sat  Satellite ECEF position (m).
+ * @param[in]  rcv  Receiver ECEF position (m).
+ * @param[out] e    Unit vector from receiver to satellite.
+ * @return     Geometric range with Sagnac correction (m).
+ */
+inline double geodist(const Point3& sat, const Point3& rcv, Point3& e) {
+  const Point3 dr = sat - rcv;
+  const double r = dr.norm();
+  e = dr / r;
+  return r + OMGE * (sat.x() * rcv.y() - sat.y() * rcv.x()) / C_LIGHT;
```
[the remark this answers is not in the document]
varunagrawal: This is a valid point.

gtsam/navigation/PseudorangeFactor.h
```diff
@@ -154,63 +155,36 @@ class GTSAM_EXPORT DifferentialPseudorangeFactor
   typedef NoiseModelFactorN<Point3, double, double> Base;
 
  public:
-  // Provide access to the Matrix& version of evaluateError:
   using Base::evaluateError;
```
varunagrawal, line 158: Do you need this? You've implemented `evaluateError`.

gtsam/navigation/PseudorangeFactor.h
```diff
@@ -484,4 +415,305 @@ template <>
 struct traits<DifferentialPseudorangeFactorArm>
     : public Testable<DifferentialPseudorangeFactorArm> {};
 
+/**
+ * DD pseudorange factor.
```
varunagrawal: Again prefer writing this out in detail in the docstring and using `DoubleDiffPseudorangeFactor`.

varunagrawal (commented):
Some initial comments.

dellaert:
Thanks for initial review, @varunagrawal . @masoug - would also appreciate your take on the implemented functionality, and how it relates.

@inuex35 please run `make check` locally and report on result in the PR comment. Make sure to compile with all warnings treated as errors. After you do this and address other comments above I'll run CI.

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,416 @@
...
+namespace gtsam {
+
+/**
+ * Base class storing common members for carrier phase factors.
+ *
+ * Clock biases are in seconds (same convention as PseudorangeFactorArm).
+ * Ambiguity is in meters (= lambda * N). To recover the integer
+ * ambiguity N, divide by the wavelength: N = ambiguity_meters / lambda
+ */
+struct CarrierPhaseBase {
```
masoug: This struct shares identical members as [PseudorangeBase](https://github.com/borglab/gtsam/blob/2d3a943f96df8cb5558ab754aab0241b307941b9/gtsam/navigation/PseudorangeFactor.h#L23). I think it would be beneficial to fold `CarrierPhaseBase` and `PsuedorangeBase` into one common struct to reduce duplication.

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,416 @@
...
+
+  virtual ~CarrierPhaseFactor() = default;
+
+  /**
+   * Construct a CarrierPhaseFactor.
+   *
+   * @param receiverPositionKey Receiver gtsam::Point3 ECEF position node.
+   * @param receiverClockBiasKey Receiver clock bias node (seconds).
+   * @param ambiguityKey Ambiguity node (meters, = lambda * N).
+   * @param measuredCarrierPhase Carrier phase measurement in meters.
```
masoug, line 65: nit: "Carrier Phase" is typically specified in units of cycles (from what I understand, I may be wrong), not meters, so I'd just suggest renaming this member to something like `measuredCarrierPhaseMeters` to make it very clear (even though the documentation already says meters). That way, anyone reading the code will know to implicitly convert cycle measurements [in your formulas](https://github.com/borglab/gtsam/pull/2502/changes#diff-518f8004061e76419b596478cdb6b0838fe3826f4651336f659c53d5f47d67f1R55).

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,416 @@
...
+ * The factor forms the double-difference internally.
+ *
+ * error = [(geodist(satRefRov,pos) - geodist(satRefBase,basePos))
+ *        - (geodist(satTargetRov,pos) - geodist(satTargetBase,basePos))]
+ *       + lam * (ambRef - ambTarget)
+ *       - [(cpRovRef - cpBaseRef) - (cpRovTarget - cpBaseTarget)]
+ *
+ * @ingroup navigation
+ */
+class GTSAM_EXPORT DoubleDifferenceCarrierPhaseFactor
```
masoug, line 237: Can you help clarify the purpose of this factor? From what I understand, the double-difference algorithm is implicitly solved by constructing a factor graph of several `CarrierPhaseFactor`s connected to two triplets of position/bias/ambiguity variable nodes (one reference triplet and a second rover triplet).

Is there a use-case where this direct double-difference factor might be more appropriate than the `CarrierPhaseFactor` setup described above? Did you intend this to be a convenience factor for quick position estimation?

gtsam/navigation/GnssCommon.h
```diff
@@ -0,0 +1,61 @@
...
+ * NaNs from dividing by zero.
+ *
+ * @param[in]  sat   Satellite ECEF position (m).
+ * @param[in]  rcv   Receiver ECEF position (m).
+ * @param[out] e     Unit vector from receiver to satellite.
+ * @param[out] H_rcv Optional Jacobian of the returned range w.r.t. rcv (1x3).
+ *                   Includes the contribution from the Sagnac term.
+ * @return     Geometric range with Sagnac correction (m).
+ */
+inline double geodist(const Point3& sat, const Point3& rcv, Point3& e,
```
masoug: I know this is `inline`, but would it make sense to move this implementation into a cpp file? I think most modern C++ compilers are capable of optimizing machine code across translation units, so I'd imagine you'd still get most of `inline` performance benefits without having to maintain a header implementation?

gtsam/navigation/GnssCommon.h
```diff
@@ -0,0 +1,61 @@
...
+                      OptionalJacobian<1, 3> H_rcv = {}) {
+  const Point3 dr = sat - rcv;
+  const double r = dr.norm();
+  const double sagnac =
```
masoug: Forgive if I missed; are there unit/jacobian tests for `geodist()`?

masoug (commented):
This is great! One humble request, if I may: Would you be willing to provide an end-to-end example of these carrier phase factors using data from physical receivers ([similar to the other psuedorange factors](https://github.com/borglab/gtsam/blob/2d3a943f96df8cb5558ab754aab0241b307941b9/python/gtsam/examples/DifferentialPseudorangeExample.ipynb))? I am especially interested in getting your thoughts on cm-scale positioning by some hybrid estimation between gtsam and a integer optimizer for ambiguity resolution.

inuex35:
@masoug you can try this. make sure there is no outlier or multipath detection.
you will achieve integer fix when it's open sky.
https://github.com/inuex35/gnss-gtsam-rtk

gtsam/navigation/CarrierPhaseFactor.h
inuex35: Left as separate classes to match the existing gtsam/navigation pattern — GPSFactor / GPSFactorArm / GPSFactorArmCalib (and their GPSFactor2* NavState counterparts) are already split this way rather than folded with optional ctor args. The split is forced at the type level: the arm variant adds a Pose3 (or Point3 for calib) key, so NoiseModelFactorN<…> vs NoiseModelFactorN<…, Pose3> are genuinely different instantiations. Happy to revisit if you'd like to unify the whole GPS/pseudorange/carrier family in one pass, but I think that belongs in its own PR rather than this one.

gtsam/navigation/CarrierPhaseFactor.h
masoug: Thanks for adding the explanation 🙏

gtsam/navigation/GnssCommon.h
masoug: [thanks](https://github.com/borglab/gtsam/pull/2502/changes#diff-42930e61c5a7608772bf3fd8130ca4c55b5ce949752c3396898c0c206f82674eR21) :heavy_check_mark:

dellaert:
@inuex35 Please resolve all comments above that you have addressed or respond to them if you don't think they are relevant. I will then do a final review - maybe I'll also push some commits onto your branch; I see that you've allowed that - which is good :-)

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+      : CarrierPhaseBase{0.0, Point3(0, 0, 0), 0.0} {}
+
+  virtual ~CarrierPhaseFactor() = default;
+
+  /**
+   * Construct a CarrierPhaseFactor.
+   *
+   * @param receiverPositionKey Receiver gtsam::Point3 ECEF position node.
+   * @param receiverClockBiasKey Receiver clock bias node (seconds).
+   * @param ambiguityKey Ambiguity node (meters, = lambda * N)..
```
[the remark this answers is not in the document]
inuex35: typo fixed

gtsam/navigation/GnssCommon.h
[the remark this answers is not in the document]
inuex35: thanks, fixed

gtsam/navigation/PseudorangeFactor.h
```diff
@@ -484,4 +415,305 @@ template <>
...
+    const double rBaseRef = gnss::geodist(satRefBase_, basePos_, dummy);
+    const double rBaseTarget = gnss::geodist(satTargetBase_, basePos_, dummy);
+
+    const double ddModel = (rRovRef - rBaseRef) - (rRovTarget - rBaseTarget);
+    const double error = ddModel - ddObs;
+
+    if (H_pose) {
+      H_pose->resize(1, 6);
+      const bool ok = rRovRef > std::numeric_limits<double>::epsilon() &&
+                      rRovTarget > std::numeric_limits<double>::epsilon();
```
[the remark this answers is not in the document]
inuex35: DD implementations were moved into `PseudorangeFactor.cpp` The header no longer uses `std::numeric_limits`, and `<limits>` is already included in the .cpp.

gtsam/navigation/PseudorangeFactor.h
```diff
@@ -484,4 +415,305 @@ template <>
...
+    Point3 dummy;
+    const double rBaseRef = gnss::geodist(satRefBase_, basePos_, dummy);
+    const double rBaseTarget = gnss::geodist(satTargetBase_, basePos_, dummy);
+
+    const double ddModel = (rRovRef - rBaseRef) - (rRovTarget - rBaseTarget);
+    const double error = ddModel - ddObs;
+
+    if (H) {
+      *H = (Matrix(1, 3) << (eTarget - eRef).transpose()).finished();
+    }
```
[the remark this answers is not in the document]
inuex35: gnss::geodist(..., H_rcv)`already bundles the Sagnac-term derivative

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+    const double rBaseTarget = gnss::geodist(satTargetBase_, basePos_, dummy);
+
+    const double ddModel = (rRovRef - rBaseRef) - (rRovTarget - rBaseTarget);
+    const double error = ddModel + lam_ * (ambRef - ambTarget) - ddObs;
+
+    if (H_pose) {
+      H_pose->resize(1, 6);
+      const bool ok = rRovRef > std::numeric_limits<double>::epsilon() &&
+                      rRovTarget > std::numeric_limits<double>::epsilon();
+      if (!ok) {
```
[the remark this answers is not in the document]
inuex35: fixed

gtsam/navigation/CarrierPhaseFactor.h
```diff
@@ -0,0 +1,526 @@
...
+    Point3 eRef, eTarget, dummy;
+    const double rRovRef = gnss::geodist(satRefRov_, pos, eRef);
+    const double rRovTarget = gnss::geodist(satTargetRov_, pos, eTarget);
+    const double rBaseRef = gnss::geodist(satRefBase_, basePos_, dummy);
+    const double rBaseTarget = gnss::geodist(satTargetBase_, basePos_, dummy);
+
+    const double ddModel = (rRovRef - rBaseRef) - (rRovTarget - rBaseTarget);
+    const double error = ddModel + lam_ * (ambRef - ambTarget) - ddObs;
+
+    if (Hpos) *Hpos = (Matrix(1, 3) << (eTarget - eRef).transpose()).finished();
```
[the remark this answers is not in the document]
inuex35: `H_rRovRef - H_rRovTarget` returned by `gnss::geodist` fixed

gtsam/navigation/CarrierPhaseFactor.h
inuex35: `DD`-prefixed classes to `DoubleDifference*`

gtsam/navigation/CarrierPhaseFactor.h
inuex35: Done — the implementation is now in `gtsam/navigation/CarrierPhaseFactor.cpp`.

gtsam/navigation/CarrierPhaseFactor.h
inuex35: Done

gtsam/navigation/CarrierPhaseFactor.h
inuex35: the arm variant has a Pose3 key, the non-arm variant has a Point3 key, so `NoiseModelFactorN<Point3,…>` vs `NoiseModelFactorN<Pose3,…>` are different instantiations and
  can't share a class cleanly

gtsam/navigation/CarrierPhaseFactor.h
inuex35: done

gtsam/navigation/CarrierPhaseFactor.h
inuex35: done

dellaert:
Still seeing a bunch of comments that you did not resolve or respond to. I’m assuming you’re still looking at them :-). Please re-request a review for me when done.

gtsam/navigation/CarrierPhaseFactor.h
inuex35: same pattern as GPSFactor / GPSFactorArm. NoiseModelFactorN has multiple evaluateError overloads; overriding one hides the rest, so using Base::evaluateError; keeps them accessible.

gtsam/navigation/CarrierPhaseFactor.h
inuex35: thank you, renamed

inuex35:
@dellaert Thank you, maybe I finished the implementation. If you do not have any problem with this, please review again

dellaert:
This is looking great, thanks! My remaining concern is mostly elegance/maintainability rather than the high-level functionality, in particular:
- the DD pseudo-range and carrier-phase factors carry almost identical data and compute almost identical range/Jacobian terms, with carrier phase only adding the ambiguity term. Could we factor the shared DD observation/satellite/base-position data and the DD range/Jacobian computation into a small internal helper? 
- same goes for the arm variants: they repeat the same ecef_T_nav composition, antenna-position computation, and lever-arm Jacobian logic. Could this be pulled into a small helper, e.g. something that maps a Pose3 + lever arm + optional ecef_T_nav to antenna position and Jacobian?
- finally, some similar comments to be made about the tests. Maybe some named namespaces with quantities used in several tests would eliminate copy/paste.

All these things would make the new API feel less copy/pasted and easier to maintain.

inuex35:
Thanks @dellaert, all three addressed:

1. Added `gnss::DoubleDifferenceData` in `GnssCommon.h/.cpp` — holds the rover/base observations + 4 satellite positions + base position, and exposes `observed()` and `model(rcv, H_rcv)` (Sagnac-corrected, with Jacobian). Both DD factors store one of these and just delegate; carrier-phase only adds the `λ·(N_ref − N_target)` term and the two ambiguity Jacobians.
2. Added `gnss::LeverArm` (lever arm + optional `ecef_T_nav`) with `antennaPosition(pose, frame)` and `antennaPoseJacobian(H_antenna, frame)`. All four `*Arm` factors now share this for the Pose3 → antenna mapping and the chain rule through compose + lever arm.
3. Added `gtsam/navigation/tests/gnssTestHelpers.h` with the shared satellite/base/observation constants, lever arm, `ecef_T_nav` (Tokyo ENU origin), and `λ_L1`. Both test files use it, so each TEST body only specifies what's unique to that case.

dellaert (approved):
Awesome, LGTM!

inuex35:
Hi, I released tightly coupled gnss imu code.
You can try if you're interested in.
https://github.com/inuex35/tightly-coupled-gnss-imu-fgo

dellaert:
Nice! *If* you’re interested, one of these runs would be a nice GTSAM example illustrating both IMU factor and your DD factor, if we can isolate the dependence on other packages, e.g., by storing pre-processed data in an compact file that we can include in/examples/Data. We could then do a blog post on GTSAM.org as well :-)

inuex35:
The dataset i used is open dataset. so you can add the link to the blog!
https://github.com/taroz/PPC-Dataset
Actually rtk was not just adding double difference factor, there are many functions to achieve cm level. So I needed to use other package.

dellaert:
Right. But if you want to make a blog post and an accompanying notebook, then all these other functions would have to be done by you in advance, and the results stored in some *compact* file format - We really don't want to add large files to the GTSM distribution. 

If it really needs those functions at runtime, then indeed we will not be able to create a notebook from it. However, we might still be able to write a blog post. If you're interested in that, please open a PR on gtsam.org with the blog post and I'll give comments, or help by editing a bit.

inuex35:
Thanks @dellaert! Runtime ambiguity resolution and cycle-slip handling make a self-contained notebook tricky, so I'd like to go with the blog post route for now. I had Claude help draft it and it turned out pretty nice. I've opened a PR on gtsam.org:
https://github.com/borglab/gtsam.org/pull/80

PR #2503, opened 2026-04-14
Geometry enhancements

dellaert:
### High-level changes

- Modernized `gtsam/base/numericalDerivative.h`:
  - moved to generic callable-based derivative helpers
  - improved overload behavior for lambdas/free functions
  - added/organized Doxygen groups and explicit template-type requirements
  - removed stale/inaccurate template requirement wording

- Geometry/runtime updates:
  - `Gal3`: added analytical `LogmapDerivative(const TangentVector&)` and routed `LogmapDerivative(const Gal3&)` through it
  - `Quaternion` traits: added `GetDimension`
  - `SO3`: small cleanup (inline `ExpmapFunctor::expmap`, corrected `B` comment)
  - `Pose3`: expanded implementation comments around `Expmap`
  - wrappers (`geometry.i`): exposed `Similarity2::matrix()` / `Similarity3::matrix()` and aligned `DexpFunctor` field constness

- Test/style updates in geometry/navigation:
  - mechanically replaced `std::function<...> var = ...` with `auto var = ...` in test code
  - added `[[maybe_unused]]` only where required to satisfy `-Werror`

### Validation

- `make -j6 check.geometry` ✅
- `make -j6 check.navigation` ✅

### DIFF
--- a/gtsam/geometry/SO3.h
+++ b/gtsam/geometry/SO3.h
@@ -143,7 +143,7 @@ struct GTSAM_EXPORT ExpmapFunctor {
 
   // Ethan Eade's constants:
   double A;  // A = sin(theta) / theta
-  double B;  // B = (1 - cos(theta))
+  double B;  // B = (1 - cos(theta)) / theta^2
 
   /// Constructor with element of Lie algebra so(3)
   explicit ExpmapFunctor(const Vector3& omega);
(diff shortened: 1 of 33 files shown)


gtsam/geometry/SO3.h
```diff
@@ -155,7 +155,7 @@ struct GTSAM_EXPORT ExpmapFunctor {
   ExpmapFunctor(const Vector3& axis, double angle);
 
   /// Rodrigues formula
-  Matrix3 expmap() const;
+  Matrix3 expmap() const { return I_3x3 + A * W + B * WW; }
```
ProfFan: non inline implementation should not be in the header

dellaert: Will it not be automatically inlined? That’s what I’ve been told. All CI passes.

ProfFan: It will be a ODR violation

dellaert: OK, I will fix. The CI I guess then does not check for ODR violations?

PR #2508, opened 2026-04-27
perf(cholesky): add choleskyPartialBlocked() with panel-level Level-3 BLAS operations

jashshah999:
## What this does

Adds `choleskyPartialBlocked()` — a panel-blocked variant of `choleskyPartial()` that operates on `blockSize` columns at a time rather than one column at a time.

**Algorithm (right-looking blocked Cholesky):**

For each panel of `bs` columns:
1. Factor the `bs x bs` diagonal block column-by-column using `choleskyStep` (preserves underconstrained-pivot handling)
2. Triangular solve (TRSM): `S = R^{-T} B` — in-place on the `bs x trailing` panel row
3. Rank-`bs` Schur update (SYRK): `trailing -= S' * S` — upper triangle only

Steps 2 and 3 are Level-3 BLAS operations (O(bs) arithmetic intensity) vs. the Level-2 rank-1 updates (O(1) arithmetic intensity) done per column in `choleskyPartial`. With an optimised BLAS backend (OpenBLAS, MKL, Apple Accelerate) these map to multi-threaded `dtrsm`/`dsyrk` calls.

**Key design decisions:**
- Falls back to `choleskyPartial` when `nFrontal <= blockSize` (no overhead for small matrices)
- `assert(blockSize > 0)` guards against degenerate input
- Underconstrained-pivot handling (`zeroPivotThreshold`, `underconstrainedPrior`) is preserved exactly via the existing `choleskyStep` function; `dpotrf` could replace the panel factor in step 1 for further gain but would lose rank-deficiency handling
- API-compatible: same `(Matrix&, nFrontal, topleft)` signature with an optional `blockSize` parameter (default: 64)

## Benchmark (plain Eigen, no optimised BLAS)

On plain Eigen (no BLAS backend), `choleskyPartial` uses `Eigen::LLT` which is itself internally blocked, so the gain here is small and noisy. On a BLAS-backed build the blocked path benefits from multi-threaded `dtrsm`/`dsyrk` and the gap grows with problem size.

```
     N    nF  scalar(ms)   bs=16(ms)   bs=64(ms)
    64    48       0.015       0.012       0.013
   128    96       0.054       0.060       0.080
   256   192       0.362       0.260       0.374
   512   384       1.712       1.830       1.853
```

Results are variable at these matrix sizes because the working set fits in cache and per-call overhead dominates. The primary intended gain is at `nFrontal > 256` with a BLAS-backed build.

## Tests

Five new unit tests in `testCholesky.cpp`:
- `blockedSmall_fallback`: exact equality with `choleskyPartial` when falling back (blockSize > nFrontal)
- `blockedInvariant_medium`: verifies `R'R = A` (frontal block) for a 32x32 matrix, blockSize=8
- `blockedInvariant_large`: same for a 128x128 matrix, blockSize=16
- `blockedWithTopleft`: correctness with a non-zero `topleft` offset
- `blockedRecovery`: full invariant `[R' 0; S' I][R S; 0 L] = [A B; B' C]` on the same 7x7 matrix used by the existing `choleskyPartial` test, blockSize=2 to force multiple panels

### DIFF
--- a/gtsam/base/SymmetricBlockMatrix.cpp
+++ b/gtsam/base/SymmetricBlockMatrix.cpp
@@ -83,7 +83,8 @@ void SymmetricBlockMatrix::invertInPlace() {
 void SymmetricBlockMatrix::choleskyPartial(DenseIndex nFrontals) {
   gttic(VerticalBlockMatrix_choleskyPartial);
   DenseIndex topleft = variableColOffsets_[blockStart_];
-  if (!gtsam::choleskyPartial(matrix_, offset(nFrontals) - topleft, topleft)) {
+  const size_t nF = offset(nFrontals) - topleft;
+  if (!gtsam::choleskyPartialBlocked(matrix_, nF, topleft)) {
     throw CholeskyFailed();
   }
 }
--- a/gtsam/base/cholesky.cpp
+++ b/gtsam/base/cholesky.cpp
@@ -157,4 +157,85 @@ bool choleskyPartial(Matrix& ABC, size_t nFrontal, size_t topleft) {
     return true;
   }
 }
+/* ************************************************************************* */
+bool choleskyPartialBlocked(Matrix& ABC, size_t nFrontal, size_t topleft,
+                            size_t blockSize) {
+  if (nFrontal == 0) return true;
+
+  // For small frontal sizes fall through to the scalar path – the panel
+  // overhead is not worth it and the underconstrained-pivot logic in
+  // choleskyStep is already scalar.
+  if (nFrontal <= blockSize)
+    return choleskyPartial(ABC, nFrontal, topleft);
+
+  assert(ABC.cols() == ABC.rows());
+  assert(size_t(ABC.rows()) >= topleft);
+  const size_t n = static_cast<size_t>(ABC.rows() - topleft);
+  assert(nFrontal <= n);
+
+  // Work on the submatrix starting at (topleft, topleft).
+  // Layout (upper triangular, augmented with RHS column):
+  //
+  //   [ A   B ]    A: nFrontal x nFrontal  (frontal block)
+  //   [ B'  C ]    C: (n-nFrontal) x (n-nFrontal)  (Schur complement)
+  //
+  // Right-looking blocked Cholesky on A, accumulating Schur updates to C.
+  //
+  // At panel k (column offset into A):
+  //   bs  = min(blockSize, nFrontal - k)   actual panel width
+  //   Akk = A(k:k+bs, k:k+bs)             diagonal panel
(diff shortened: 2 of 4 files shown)


dellaert:
Cool, but I don’t see this new function being called anywhere. Are you planning to add more commits benchmarking a use case within GTSAM?

jashshah999:
Thanks for the feedback! I've now wired `choleskyPartialBlocked` into `SymmetricBlockMatrix::choleskyPartial()` so it's used by all elimination paths (HessianFactor elimination, ISAM2 clique factorization, batch solvers).

The blocked version already falls back to the original `choleskyPartial` when `nFrontal <= blockSize` (default 64), so small cliques see no change. Larger frontal blocks get the panel-level TRSM + SYRK operations.

Verified against 10 test suites — zero failures:
testCholesky, testHessianFactor, testJacobianFactor, testGaussianISAM2, testGaussianBayesNet, testGaussianFactorGraph, testSymmetricBlockMatrix, testNonlinearOptimizer, testVisualISAM2, testGaussianFactorGraphB.

dellaert:
Any perf numbers? E.g., before/after timeSfmBAL?

jashshah999:
I don't have the BAL datasets available on this machine to run timeSFMBAL. The honest benchmark I ran (standalone, plain Eigen without optimized BLAS) showed mixed results because Eigen's LLT is already internally blocked:

```
     N    nF  scalar(ms)   bs=16(ms)   bs=64(ms)
    64    48       0.015       0.012       0.013
   128    96       0.054       0.060       0.080
   256   192       0.362       0.260       0.374
   512   384       1.712       1.830       1.853
```

The blocked path uses `choleskyStep` column-by-column for the panel diagonal (to preserve underconstrained-pivot handling), then Level-3 TRSM+SYRK for the trailing update. On plain Eigen, `choleskyPartial`'s `Eigen::LLT` is already doing something similar internally, so the gain is small.

The real benefit would come with an optimized BLAS backend where `rankUpdate` and `solveInPlace` map to multi-threaded `dsyrk`/`dtrsm`. If you think the gain isn't sufficient without that, I understand — happy to close this if it's not pulling its weight.

jashshah999:
Just following up. Happy to close this if the marginal gains without an optimized BLAS backend don't justify the added code. Let me know your preference.

dellaert:
Yeah, if indeed Eigen is internally blocked, then I’d prefer having Eigen do the lifting - rather than adding mode code in GTSAM.

jashshah999:
Makes sense. Closing this since Eigen handles the blocking internally and the added code doesn't provide enough benefit to justify the maintenance. Thanks for the review.

PR #2511, opened 2026-04-27
fix(geometry): implement Cayley chart Jacobians for Rot3

### LINKED ISSUE #2504: Cayley Chart version of Rot3 does not support Jacobians
## Description

Cayley Chart version of `Rot3` does not support Jacobians. This means that the standard error functions on factors defined with `Rot3` fail when Jacobians are requested (see code below).

## Steps to reproduce

Build GTSAM with the Cayley Chart (i.e., set `GTSAM_ROT3_EXPMAP=OFF`). The following build options can be used to reproduce the issue:
```bash
-DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS=-w -DGTSAM_BUILD_TESTS=ON -DGTSAM_BUILD_UNSTABLE=ON -DGTSAM_WITH_TBB=OFF -DGTSAM_BUILD_EXAMPLES_ALWAYS=OFF -DGTSAM_ALLOW_DEPRECATED_SINCE_V43=OFF -DGTSAM_USE_QUATERNIONS=OFF -DGTSAM_ROT3_EXPMAP=OFF -DGTSAM_POSE3_EXPMAP=OFF -DGTSAM_USE_SYSTEM_EIGEN=OFF -DGTSAM_USE_SYSTEM_METIS=OFF -DGTSAM_USE_BOOST_FEATURES=OFF -DGTSAM_ENABLE_BOOST_SERIALIZATION=OFF -DGTSAM_BUILD_WITH_MARCH_NATIVE=OFF -DGTSAM_SINGLE_TEST_EXE=OFF -DGTSAM_ENABLE_GEOGRAPHICLIB=OFF
```

The following code snippet will reproduce the error if the Cayley Chart is used to represent the `Rot3` class:
```cpp
// Model
  const auto model = noiseModel::Diagonal::Sigmas(Vector6::Ones());
  // Set up values
  const Pose3 p0 = Pose3::Expmap(Vector6(0.0, 0.0, 0.0, 0.0, 0.0, 0.0));
  const Pose3 p1 = Pose3::Expmap(Vector6(0.5, 0.0, 0.0, 0.0, 0.0, 0.0));
  // construct between factor
  const Pose3 p01 = p0.inverse().compose(p1);
  const auto between_factor =
  std::make_shared<BetweenFactor<Pose3>>(P(0), P(1), p01, model);
  // Make values
  Values values;
  values.insert(P(0), p0);
  values.insert(P(1), p1);
  // // Test unwhitened error 
  std::vector<Matrix> jacobians(2);
  const Vector error = between_factor->unwhitenedError(values, &jacobians);
  // test evaluate error with Jacobians
  Matrix jacobianPose0;
  Matrix jacobianPose1;
  const Vector errorFromEvaluateError =
      between_factor->evaluateError(p0_se3, p1_se3, jacobianPose0, jacobianPose1);
```

## Expected behavior

Code snippet above should run without error.

## Environment

This error occurs when running the C++ version of GTSAM. Environment:
```
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.5 LTS
Release:        22.04
Codename:       jammy
```

jashshah999:
## Problem

`CayleyChart::Retract` and `CayleyChart::Local` throw `std::runtime_error` when Jacobians are requested:

```cpp
if (H) throw std::runtime_error("Rot3::CayleyChart::Retract Derivative");
```

This makes the Cayley chart parameterization unusable with any optimizer that needs derivatives, which is all of them. When GTSAM is built with `GTSAM_ROT3_EXPMAP=OFF`, optimization involving `Rot3` fails.

Fixes #2504.

## Fix

Derived closed-form analytical Jacobians for both maps:

**Retract:** `H = (4 / (4 + ||w||^2)) * (I - [w/2]x)`

**Local:** `H = I + [w/2]x + w*w'/4`

where `[w/2]x` is the skew-symmetric matrix of `w/2`.

Derivation: the Cayley retraction maps `w -> R(w) = (4I + 2[w]x + [w]x^2) / (4 + ||w||^2)`. The Retract Jacobian relates perturbations in Cayley coordinates to perturbations on the manifold (in the default chart, typically Expmap). The Local Jacobian is its inverse. Both satisfy `H_Retract * H_Local = I` and reduce to `I` at `w = 0`.

Both formulas verified against numerical derivatives at multiple test points (zero, small, medium, mixed signs) to ~1e-11 precision.

**Additional fix:** replaced exact float comparison `determinant() == 0.0` with tolerance-based `abs(determinant()) < 1e-10` in `CayleyChart::Local` for numerical robustness near the singularity.

## Tests

Three new tests in `testRot3.cpp`:
- `CayleyChart_Retract_Jacobian`: 4 test points vs numerical derivative
- `CayleyChart_Local_Jacobian`: 4 test points vs numerical derivative
- `CayleyChart_Retract_Local_InverseJacobians`: verifies `H_Retract * H_Local = I`


dellaert:
Please evaluate co-pilot comments and check CI, seems to have a failure.

jashshah999:
Fixed the CI failure — tests were in testRot3.cpp which compiles on all builds including quaternion, but CayleyChart only exists in the matrix representation. Moved the tests to testRot3M.cpp which is conditionally compiled.

Regarding the Copilot comments: the determinant-based singularity check is the pre-existing approach, I only changed the tolerance from exact equality to 1e-10. The test tolerances (1e-7 for numerical derivatives, 1e-9 for inverse) are consistent with the existing Expmap tests in the same file.

jashshah999:
Pushed an empty commit to re-trigger CI (checks didn't run on the last push). The CI failure was fixed in the previous commit by moving tests to testRot3M.cpp. Copilot comments were addressed in my earlier reply. Let me know if anything else is needed.

dellaert:
Still seeing (preventable) CI failures.

dellaert (approved):
LGTM! Thanks for this valuable contribution !

PR #2512, opened 2026-04-27
Prepare 4.2.1 release

ProfFan:
This prepares the 4.2.1 patch release branch on top of release/4.2.

Included here:
- bump release metadata to 4.2.1
- adapt Boost detection for current/Homebrew Boost while keeping Boost.Filesystem linked and the Boost floor at 1.65
- fix the missing Boost.Range include in testLinearEquality so check passes locally
- backport the production wheel build workflow for the 4.2 line
- pass the CMake policy floor explicitly in wheel builds
- build Boost.Filesystem in the wheel bootstrap
- split macOS wheel builds into separate arm64 and x86_64 jobs instead of labeling them universal2
- pin NumPy below 2.0.0 for wheel compatibility with the vendored pybind11 stack

Local validation:
- full check target passes in the 4.2.1 worktree

Follow-up after merge:
- tag 4.2.1 on release/4.2
- merge release/4.2 forward into develop


dellaert (approved):
Yay!!

dellaert:
I downloaded the artifacts and tested 3.11 and 3.14 on silicon Apple silicon. Everything worked fine.
I uploaded all wheels to PyPy. https://pypi.org/project/gtsam/4.2.1/

ProfFan:
Thanks Frank! We should add an action to auto-promote these to primary PyPI in the future :)

LockedThread:
I'm a bit confused on the numpy 2.0+ support. Do any published versions of gtsam support numpy  2.0+? I can see that the `develop` branch does not restrict the support to `<2.0.0`.

PR #2513, opened 2026-05-01
Continuous-Time GP Factor Graph and Optimization

holmesco:
This introduces a derived factor graph class and optimization routine that are designed to work with the Continuous-Time Gaussian Process factors introduced in #2411. In particular, the optimization scheme leverages shared memory for values and Jacobians of interpolated states that are shared between different factors.


dellaert:
Closed?

holmesco:
Making a cleaner version, sorry for the confusion

PR #2515, opened 2026-05-03
navigation: honor configured IMU bias in LeggedFixedLagSmoother

ChiyunNoh:
## Summary

This PR fixes an unintentional zero-bias initialization path in `LeggedFixedLagSmoother`.

The fixed-lag single-bias estimator now consistently uses the configured startup IMU bias from `params.imuBias`.

Specifically, this PR:
- uses `params.imuBias` when constructing the preintegration object (`pim_`)
- initializes `biasEstimate_` from `params.imuBias`
- resets `pim_` with `params_.imuBias` during full-contact reinitialization instead of using a zero bias

## Motivation

`LeggedEstimatorReplayExample` estimates an initial IMU bias and passes it through `params.imuBias`. However, `LeggedFixedLagSmoother` previously initialized some preintegration paths with a zero bias, even when a non-zero startup bias was configured.

This made the `fixed_lag_single_bias` variant inconsistent with the configured initialization and could affect replay results when a non-zero initial bias is provided.

## Validation

Built and ran the relevant legged estimator target/tests locally:

- `cmake --build build -j6 --target LeggedEstimatorReplayExample`
- `cmake --build build -j6 --target testLeggedEstimator`
- `./build/gtsam/navigation/tests/testLeggedEstimator` passes with no test failures.

### DIFF
--- a/gtsam/navigation/LeggedEstimator.cpp
+++ b/gtsam/navigation/LeggedEstimator.cpp
@@ -724,13 +724,14 @@ LeggedFixedLagSmoother::LeggedFixedLagSmoother(
       initialFootholds_(footholds0),
       baseCovariance0_(baseCovariance0),
       smoother_(lagSeconds, leggedLmParams()),
-      pim_(params.preintegrationParams, imuBias::ConstantBias()),
+      pim_(params.preintegrationParams, params.imuBias),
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Thank you !

PR #2516, opened 2026-05-06
fix(cmake): prevent -Werror from propagating to downstream projects

### LINKED ISSUE #2401: -Werror in GTSAM forces errors in downstream projects
#1959 introduced `-Werror` in the `GTSAM_COMPILE_OPTIONS_PRIVATE_COMMON` variable in `GtsamBuildTypes.cmake`. This forces the compiler to treat all warnings as errors.

Unfortunately, this also propagates this behavior to downstream projects. This is not ideal since many (smaller) organizations choose to ignore warnings in favor of fast iteration, and this flag causes them to have to fight the compiler.

The simplest fix for this would be to remove `GTSAM_COMPILE_OPTIONS_PRIVATE_COMMON` as a CACHE variable. This would involve investigating other similar variables as well.

jashshah999:
## Summary

Fixes #2401

`GTSAM_COMPILE_OPTIONS_PRIVATE_COMMON` (and the other per-config `_PRIVATE_*` variables) were declared as `CACHE STRING` variables, causing `-Werror` and other internal compile flags to leak into downstream projects that use `find_package(GTSAM)` or `add_subdirectory(gtsam)`.

Changed to regular (non-cache) variables since these options should only apply to GTSAM's own build. The aggregated `GTSAM_COMPILE_OPTIONS_PRIVATE` variable remains `CACHE INTERNAL` as needed by `list_append_cache()`, and is still applied via `target_compile_options(... PRIVATE ...)` so it does not propagate.

## Test plan

- Verified GTSAM still builds with the same compile flags
- Downstream projects no longer inherit `-Werror`


dellaert:
Nice. Please check CI issues.

Gold856:
I'm really confused at what this PR is doing. The summary is just wrong, because the CACHE variables don't impact `find_package`, at all. Only PUBLIC compile options make it to install. As I see it, this PR is a no-op.

I'm also not exactly sure what the original issue is talking about. Is it that gtsam itself is causing `-Werror` to propagate downstream, or is that others are using `GtsamBuildTypes.cmake` and `-Werror` there is causing issues?

dellaert:
> I'm really confused at what this PR is doing. The summary is just wrong, because the CACHE variables don't impact `find_package`, at all. Only PUBLIC compile options make it to install. As I see it, this PR is a no-op.
> 
> I'm also not exactly sure what the original issue is talking about. Is it that gtsam itself is causing `-Werror` to propagate downstream, or is that others are using `GtsamBuildTypes.cmake` and `-Werror` there is causing issues?

@jashshah999 I'm a bit out of my depth here - I never was a cmake expert - but both @Gold856 and @jlblancoc are. Please respond; if truly no-op, we should close.

jlblancoc:
@dellaert I think this PR should not be merged. It's true that CACHE may pollute downstream projects that embed gtsam in them using add_subdirectory(), but that's the only effect, the private flags are not exported downstream by definition, so -Werror should not be in users' projects, and it's working right (tested in dependent projects that use gtsam via find_package().

BTW: It would be a good opportunity to enable coderabit.ai in this public repo... it's **really** good and almost 99% would give you one more independent (good) recommendation.

dellaert:
@jashshah999 Please close unless you have a compelling argument :-)

ProfFan:
I'm closing this PR, thanks all for the input :)

Issue #1665, opened 2023-11-02
Significant Slowdown in Python


Dead-Hand:
## Description

I've noticed a significant slowdown (~6667% longer to run) in the Python API when using the GTSAM LM optimizer between versions 4.2 and the latest develop branch (4.3a0).

## Steps to reproduce

I modified one of the existing Python example scripts to demonstrate the problem. Please download the attached script for an example. 

1. Install GTSAM 4.2 (I've been grabbing this from pypi i.e. `pip install gtsam`)
2. Download the attached ZIP file and unzip. It contains the test script: [DogLegOptimizerExample_timing_issues.zip](https://github.com/borglab/gtsam/files/13241623/DogLegOptimizerExample_timing_issues.zip)
3. Execute the attached script which will wait until the 250th loop iteration before reporting the optimizer runtime: `python3 DogLegOptimizerExample_timing_issues.py`
4. Uninstall gtsam: `pip uninstall gtsam`
5. Build the latest GTSAM develop branch (with Python support, so -DGTSAM_BUILD_PYTHON=1). Go into build/ (or wherever you have it set to build), then python/ and then install the python library using pip: `pip install .`
6. Execute the attached script and note the optimizer runtime: `python3 DogLegOptimizerExample_timing_issues.py`

When I run the above, step (2) reports a runtime of 0.00009s, whereas the output of step (5) reports 0.006s, representing a ~6667% increase in runtime. Your timing might vary slightly, but should still be significant.

The optimizer runtime reported in step (5) is significantly longer than in step (2), with the only thing being changed is the GTSAM version.

## Expected behavior

I would expect to see approximately identical optimizer runtimes between versions 4.2 and the latest develop branch (4.3a0).

## Environment
OS: Ubuntu 20.04 LTS
API: Python (version 3.8.10)

## Additional information

I have tried compiling the latest develop branch with and without TBB support and I have not noticed any difference in the above behavior.

ProfFan:
Did you compile GTSAM in `Release`?

Dead-Hand:
> Did you compile GTSAM in `Release`?

I just now recompiled in Release mode and ran the same test a few times. It is much faster, but it's still about 50% the speed of running the same code in 4.2 release. LM optimize() takes about 0.0001s - 0.0002s with all else being equal.

ProfFan:
> > Did you compile GTSAM in `Release`?
> 
[...]

PR #2520, opened 2026-05-08
perf(python): return references from optimizer to avoid deep-copying Values

jashshah999:
## Summary

When `NonlinearOptimizer.optimize()`, `values()`, or `graph()` is called from Python, the pybind11 wrapper previously deep-copied the returned object (cloning every entry in the `Values` map) because the generated lambda deduced a by-value return type from the C++ `const T&` return.

This PR:
1. Modifies the wrapper generator (`pybind_wrapper.py`) to detect when an interface method returns `const T&` (via the `is_ref` flag on the parsed return type)
2. Generates a lambda with explicit `-> const auto&` trailing return type
3. Adds `py::return_value_policy::reference_internal` to tie the returned reference's lifetime to the parent object

Applied to `NonlinearOptimizer::optimize()`, `optimizeSafely()`, `values()`, and `graph()` in the `.i` file.

## Impact

For a typical SLAM problem with ~100 variables, this eliminates ~100 `Value::clone_()` + heap allocation calls per `optimize()` invocation from Python. The effect scales with problem size.

Partially addresses #1665 (Python slowdown between 4.2 and 4.3). The remaining gap is likely from the `boost::ptr_map` → `std::map<unique_ptr>` migration which changed allocation patterns in `Values` copy operations elsewhere.

## Test plan

- [x] All 94 wrapper generator tests pass
- [ ] CI builds (wrapper generates correct C++ that compiles on all platforms)
- [ ] Python optimizer tests still pass (returned references remain valid while optimizer is alive)

### DIFF
--- a/gtsam/nonlinear/nonlinear.i
+++ b/gtsam/nonlinear/nonlinear.i
@@ -334,12 +334,12 @@ typedef gtsam::GncParams<gtsam::LevenbergMarquardtParams> GncLMParams;
 
 #include <gtsam/nonlinear/NonlinearOptimizer.h>
 virtual class NonlinearOptimizer {
-  gtsam::Values optimize();
-  gtsam::Values optimizeSafely();
+  const gtsam::Values& optimize();
+  const gtsam::Values& optimizeSafely();
   double error() const;
   int iterations() const;
-  gtsam::Values values() const;
-  gtsam::NonlinearFactorGraph graph() const;
+  const gtsam::Values& values() const;
+  const gtsam::NonlinearFactorGraph& graph() const;
   gtsam::GaussianFactorGraph* iterate() const;
 };
 


dellaert:
I'll run CI - to gather information - but this PR cannot be merged to gtsam develop. `wrap` is a separate repo that is imported as a git subtree - so the PR needs to be done there first. Then this PR has to grab the subtree and make the change to nonlinear.i.

Adding @varunagrawal as a reviewer - he might be able to help - and is in a better position to assess the contribution. Would be amazing to avoid overhead!

varunagrawal:
The contributions needs to go to the `wrap` repo along with the tests.

jashshah999:
@varunagrawal Got it — submitted the wrapper generator change to the wrap repo: https://github.com/borglab/wrap/pull/182

Once that's merged and the subtree is updated in gtsam, this PR only needs the `nonlinear.i` change (marking `optimize()`, `values()`, and `graph()` as returning `const T&`).

LockedThread:
> @varunagrawal Got it — submitted the wrapper generator change to the wrap repo: [borglab/wrap#182](https://github.com/borglab/wrap/pull/182)
> 
> Once that's merged and the subtree is updated in gtsam, this PR only needs the `nonlinear.i` change (marking `optimize()`, `values()`, and `graph()` as returning `const T&`).

Your other PR got merged FYI.

jashshah999:
Rebased onto current develop. Since borglab/wrap#182 is now merged into the wrap subtree, this PR is down to a single one-line change: marking `optimize()`, `optimizeSafely()`, `values()`, and `graph()` as `const T&` in `nonlinear.i`.

The wrap subtree already handles `reference_internal` return policy for const-ref returns, so this should just work. Ready for CI.

dellaert (approved):
LGTM!

PR #2521, opened 2026-05-08
feat(python): TorchFactor — PyTorch autograd bridge for GTSAM factors

jashshah999:
## Summary

Adds `gtsam.utils.torch_factor.TorchFactor`, a utility class that bridges PyTorch's automatic differentiation with GTSAM's factor graph optimization.

**The problem:** Writing custom GTSAM factors requires hand-deriving Jacobians, which is error-prone and limits adoption for neural/differentiable factors (NeRF-SLAM, 3DGS-SLAM, learned odometry, photometric factors).

**The solution:** Define your residual function using PyTorch tensors, and TorchFactor computes Jacobians automatically via `torch.autograd.functional.jacobian`.

```python
from gtsam.utils.torch_factor import TorchFactor, extract_pose3

def photometric_residual(tensors):
    pose_vec = tensors[0]
    # ... render from pose, compute pixel error ...
    return error_tensor

factor = TorchFactor(
    noise_model=gtsam.noiseModel.Isotropic.Sigma(n_pixels, 1.0),
    keys=[pose_key],
    residual_fn=photometric_residual,
    value_extractors=[extract_pose3],
)
graph.add(factor.as_custom_factor())
```

## What's included

- `TorchFactor` class wrapping `CustomFactor` with autograd Jacobians
- Convenience extractors for common types: `extract_vector`, `extract_pose2`, `extract_pose3`, `extract_point2`, `extract_point3`, `extract_rot3`
- Test suite verifying:
  - Prior factor convergence
  - Between factor with two variables
  - Jacobian correctness (autograd vs analytical)
  - Nonlinear residual (sin/cos) convergence

## Motivation

This came directly from building [gtsam-splatfactors](https://github.com/jashshah999/gtsam-splatfactors), where expressing 3D Gaussian Splatting photometric errors as GTSAM factors required manual Jacobian computation through a differentiable renderer. With TorchFactor, users can leverage PyTorch's autograd for any differentiable residual while getting all the benefits of GTSAM's optimization infrastructure (iSAM2, robust kernels, multi-sensor fusion).

## Design decisions

- **Pure Python, no C++ changes** — works with existing `CustomFactor` infrastructure
- **Optional dependency** — torch is imported lazily, won't break non-torch users
- **float64 by default** — matches GTSAM's internal precision
- **Extractor pattern** — decouples value extraction from residual computation, allowing manifold-aware extraction (e.g., Pose3 → Logmap tangent vector)

## Limitations & future work

- Jacobian computation uses `torch.autograd.functional.jacobian` which recomputes the forward pass per variable — fine for small factors, could use vmap for batched factors
- No GPU dispatch yet (factors evaluated synchronously by GTSAM's optimizer)
- Extractors use Logmap which linearizes at identity — for manifold-correct optimization, a `ManifoldTorchFactor` variant with retract/localCoordinates would be the next step

## Test plan

- [x] Unit tests pass locally (pytest)
- [ ] CI (no torch in GTSAM CI, tests are skipped gracefully via `@unittest.skipUnless`)

### DIFF
--- a/python/gtsam/tests/test_torch_factor.py
+++ b/python/gtsam/tests/test_torch_factor.py
@@ -1,0 +1,168 @@
+"""Tests for TorchFactor — PyTorch autograd bridge for GTSAM."""
+
+import unittest
+
+import numpy as np
+
+try:
+    import torch
+    HAS_TORCH = True
+except ImportError:
+    HAS_TORCH = False
+
+import gtsam
+from gtsam.utils.numerical_derivative import numericalDerivative11
+
+
+@unittest.skipUnless(HAS_TORCH, "PyTorch not installed")
+class TestTorchFactor(unittest.TestCase):
+    """Test TorchFactor produces correct errors and Jacobians."""
+
+    def setUp(self):
+        from gtsam.utils.torch_factor import TorchFactor, extract_vector
+        self.TorchFactor = TorchFactor
+        self.extract_vector = extract_vector
+
+    def test_simple_prior(self):
+        """Test a simple prior factor: error = x - measurement."""
+        measurement = np.array([1.0, 2.0, 3.0])
+
+        def residual_fn(tensors):
+            x = tensors[0]
+            return x - torch.tensor(measurement, dtype=torch.float64)
+
+        noise = gtsam.noiseModel.Unit.Create(3)
+        key = 0
+        tf = self.TorchFactor(
+            noise_model=noise,
+            keys=[key],
+            residual_fn=residual_fn,
+            value_extractors=[self.extract_vector],
+        )
+
+        factor = tf.as_custom_factor()
+        graph = gtsam.NonlinearFactorGraph()
+        graph.add(factor)
+
+        # Test with initial values away from measurement
+        values = gtsam.Values()
+        values.insert(key, np.array([0.0, 0.0, 0.0]))
+
+        # Error should be non-zero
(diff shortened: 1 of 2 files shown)


dellaert:
I'll let @ProfFan review. Also, not sure what will happen in CI. We normally don't import torch and I'm not sure whether we ought to.

jashshah999:
Totally fair on the torch dependency concern. How about I maintain this as a standalone package (`gtsam-torch`) that depends on gtsam + torch, and GTSAM references it in the README/docs as the community PyTorch bridge? That way GTSAM stays lean and users who want autograd Jacobians can `pip install gtsam-torch`. Happy to add a line to the README or a docs page pointing to it.

ProfFan:
I think this is more suitable for an out-of-tree package. BTW, I think our derivatives are on-manifold, and the torch derivatives are on R^n

jashshah999:
Good point on the manifold vs R^n distinction. The current implementation uses Logmap extraction (tangent space at identity), which works for small perturbations but isn't properly manifold-aware for large deviations.

For the standalone package I'll add a `ManifoldTorchFactor` that wraps the autograd Jacobian with the appropriate retract/localCoordinates chain rule — so the user writes their residual in R^n (easier with torch) and the factor handles the manifold correction before passing Jacobians to GTSAM.

I'll close this PR and build it out-of-tree as `gtsam-torch`. Would appreciate a mention in the README or docs if it matures.

PR #2523, opened 2026-05-09
WNOA CT-GP Bindings and Example

holmesco:
# Changes

- Adjusts `nonlinear.i` for the WNOA functionality. Adds new tests for python version of interface.
- Adds a simple SE3 example as a python notebook and cpp file. Example is based on the simple example from the corresponding paper. 
- Removes assertions on GTSAM_USE_EXPMAP since #2511 fixes Cayley Map Jacobians

# Tests
- Tested that new examples perform as expected
- reran `make -j6 check.nonlinear`
- Ran python tests `pytest test_WnoaFactors.py`

# In progress

Still need to add link to arXiv version of paper when it is available (just submitted)

# Comments

New python notebook adds dependency on `pyvista`.


python/gtsam/tests/test_WnoaFactors.py
```diff
@@ -0,0 +1,405 @@
...
+            _timestep,
+            q_psd_diag,
+            keys,
+            _p0,
+            p1,
+            _p2,
+            p3,
+            _p4,
+            v0,
+            new_graph,
```
ProfFan: Formatting

dellaert:
CI failure is not an issue, but please check the co-pilot comments for validity. Fan also had a comment. Please resolve all comments that you address and provide a note if you don't address

python/gtsam/examples/GaussianProcessWnoaInterpolationSE3Helpers.py
```diff
@@ -0,0 +1,235 @@
+import numpy as np
+import numpy.linalg as npla
+from math import cos, sin, factorial
+from scipy.sparse import bmat
+from scipy.sparse.linalg import inv
+import pyvista as pv
+import time
+import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
holmesco: @dellaert we have been using pyvista for our plotting. It would be nice to keep this example consistent with the output in the paper. Do we need to add pyvista to the `requirements.txt` or is it okay if the example requires other python modules (not in requirements)?

dellaert:
I’d prefer not to add PyVista to the core GTSAM requirements, since it pulls in VTK and would make the default install significantly heavier. But inside notebooks, you can put it in the colab install block, and mention that it needs to be installed in the users env to run this example outside of colab.

dellaert:
Reviewed with codex, and agree with all issues it found:

1. `python/gtsam/examples/GaussianProcessWnoaInterpolationSE3Helpers.py:141` still fails to import without PyVista. The `try/except` sets `pv = None`, but the annotation `plotter: pv.Plotter` is evaluated at import time, causing `AttributeError: 'NoneType' object has no attribute 'Plotter'`. `from __future__ import annotations` or removing that annotation would fix it.

2. `python/gtsam/examples/GaussianProcessWnoaInterpolationSE3.ipynb:38` has a broken Colab badge URL: it points through `develop//workspace/gtsam/...` instead of `develop/python/gtsam/examples/...`. The notebook also does not follow the repo’s notebook preamble convention: copyright and Colab install cells are untagged and out of order.

3. `gtsam/nonlinear/values.i:90` looks worth challenging. Python does not expose `gtsam.Vector6`, and inserting `np.ones(6)` still stores `Eigen::Matrix<double, -1, 1>` despite the new `Vector6` overload. The tests pass with dynamic vectors, so the added `Vector6` insert/update overloads may be unnecessary or at least should be justified/reordered if typed storage is actually intended.

4. Formatting is still dirty: `git diff --check origin/develop...HEAD` fails on trailing whitespace in the new helper/test files and an extra blank line at EOF in `WnoaInterpolator.h`.

I saw (3) as well when looking through. Try if removing those Vector6 wrappings make a difference.

holmesco:
> Reviewed with codex, and agree with all issues it found:
> 
> 1. `python/gtsam/examples/GaussianProcessWnoaInterpolationSE3Helpers.py:141` still fails to import without PyVista. The `try/except` sets `pv = None`, but the annotation `plotter: pv.Plotter` is evaluated at import time, causing `AttributeError: 'NoneType' object has no attribute 'Plotter'`. `from __future__ import annotations` or removing that annotation would fix it.
> 2. `python/gtsam/examples/GaussianProcessWnoaInterpolationSE3.ipynb:38` has a broken Colab badge URL: it points through `develop//workspace/gtsam/...` instead of `develop/python/gtsam/examples/...`. The notebook also does not follow the repo’s notebook preamble convention: copyright and Colab install cells are untagged and out of order.
> 3. `gtsam/nonlinear/values.i:90` looks worth challenging. Python does not expose `gtsam.Vector6`, and inserting `np.ones(6)` still stores `Eigen::Matrix<double, -1, 1>` despite the new `Vector6` overload. The tests pass with dynamic vectors, so the added `Vector6` insert/update overloads may be unnecessary or at least should be justified/reordered if typed storage is actually intended.
> 4. Formatting is still dirty: `git diff --check origin/develop...HEAD` fails on trailing whitespace in the new helper/test files and an extra blank line at EOF in `WnoaInterpolator.h`.
> 
> I saw (3) as well when looking through. Try if removing those Vector6 wrappings make a difference.

I will address 1,2, and 4, but need to push back on 3. I originally tried to use dynamic vectors, but it caused quite a few issues internally. This is mainly because we expect the variables associated with velocities to be of type `traits<Pose>::TangentVector` with static dimension.  The python wrapper does not seem to support the use of traits to define types. If values are initialized with dynamic dimension then we get errors of the following form:
```bash
RuntimeError: Attempting to retrieve value with key "v1", type stored in Values is gtsam::GenericValue<Eigen::Matrix<double, 6, 1, 0, 6, 1> > but requested type was Eigen::Matrix<double, -1, 1, 0, -1, 1>
```
Similar errors occur if we define a prior on a the velocity variables using the dynamic `Vector` type versus the `Vector6` type. The issue may be that internally we access values as follows:
```cpp
const auto state_left = TimestampedPoseVelocity<PoseType>(
          values.at<PoseType>(left.pose),
          values.at<VelocityType>(left.velocity), left.time);
```
where `VelcocityType` is explicitly set to the tangent vector. I could try making this a dynamic, but that may result in efficiency drops.

python/gtsam/tests/test_WnoaFactors.py
```diff
@@ -0,0 +1,384 @@
...
+        self.assertEqual(stateData.velocity, velocityKey)
+        self.assertEqual(stateData.time, time)
+
+class TestWnoaMotionFactor(GtsamTestCase):
+    """Test WnoaMotionFactor class.
+    Tests are based on the more extensive C++ tests in gtsam/nonlinear/testWnoaFactor.cpp."""
+    def testConstructionAndEval(self):
+        """Test construction of WnoaMotionFactor."""
+        # First state
+        poseKey = Symbol('x', 0).key()
```
dellaert: Another comment: since people (or AI) will use this as template code, it's better to just use X(.) and V(.) everywhere, and not use intermediate variables. Everything will be much more compact and readable:
```
        stateData0 = gtsam.StateData(X(0), V(0), time = 0.0)
        stateData1 = gtsam.StateData(X(1), V(1), time = 1.0)
        qPsdDiag = np.array([1.0, 1.0, 1.0, 1.0, 1.0, 1.0])
        factor = WnoaMotionFactorPose3(stateData0, stateData1, qPsdDiag)
        self.assertIsInstance(factor, WnoaMotionFactorPose3)
```

python/gtsam/tests/test_WnoaFactors.py
```diff
@@ -0,0 +1,384 @@
...
+from dataclasses import dataclass
+
+import gtsam
+import numpy as np
+from gtsam.utils.test_case import GtsamTestCase
+
+from gtsam import Symbol
+from gtsam import WnoaInterpFactorPose3
+from gtsam import WnoaMotionFactorPose3
+from gtsam import interpolateFactorGraphPose3
```
dellaert: Only import classes, not functions. Just use gtsam.interpolateFactorGraphPose3 below.

dellaert (approved):
Good to go! Onward !

PR #2524, opened 2026-05-10
Factor for yaw in Pose2, with bias

truher:
Factor for FRC robot yaw measurement, which affects the rotation component of Pose2.  Includes bias.

My students do not know C++, but I'd like them to understand what this factor is doing, so I tried to make it *really* simple, just enough to do the job.

Our yaw sampling rate and pose estimate update rate are about the same, so there's no notion of multiple "preintegration" steps here.  In SO(2) it's easy to preintegrate separately, if anyone needs that.

See https://groups.google.com/g/gtsam-users/c/rnp9gVMm_Qc/m/FzzeafZfAwAJ


dellaert:
I think this is a good first cut, but I have 2 issues:
1. The random walk value is a parameter, and typically we do parameters as shared pointers to a parameter structure. So that would mean creating that parameter structure only once and then passing a shared pointer to it, to all factors that share that parameter setting. 
2. Once that is done the only reason for the separate measurement struct would be to have the delta R and delta T, but I think these could just be given in the constructor of the factor. I'm questioning the need for a separate struct. 

Finally,  the gyroscope gives a **rate** so my preference would be to give a rate and a delta t and calculate the delta R internally.

truher:
Thank you for getting back to me so quickly!  I'll make those changes momentarily.

One thing about deltaR is that FRC hardware performs preintegration at a higher rate than the output, so it seemed more natural to be able to use the (lower-rate) integrated output.  Most FRC hardware *also* provides a low-rate omega output, which I think uses simple decimation, so it would be (I imagine?) less accurate than a preintegrated rotation (or some better downsampling method).  Of course, it would be possible to simply compute the "average" omega represented by the gyro integrated output, to keep the factor clean and simple.   Is that what you'd recommend?

dellaert:
Yeah, good point. You could have two constructors maybe?

truher:
Well, that's what I was trying to do with the two static methods.  They're static methods because, as I understand it, if I had two constructors with identical signatures except for the type of one argument (omega: double, deltaR:Rot2), and if those arguments could be coerced to each other (as they can: Rot2(double) is not marked "explicit"), then the choice of constructor would rely on the compiler to choose wisely.  Would that be better than the static methods?

dellaert:
Right. But I’d get rid of the extra struct and add these “named constructors” to the factor. BTW, static methods are PascalCase :-)

truher:
OK I think it's ready for another look.

* removed the measurement class
* made a parameter class, containing both arw noise and bias noise
* added a convenience "between" factor for bias evolution

One question: can I delete the little conditional shared_ptr typedef, involving GNUC 4?  I cribbed it from AHRSFactor, but it seems to reference a very old gcc version?  Maybe I don't understand what it's doing.

dellaert (approved):
LGTM! Will merge when CI passes.

dellaert:
Please check CI issue?

truher:
OK, CI seems fixed.  The problem was simply including ScenarioRunner.h (which I had cribbed from AHRSFactor and forgot to remove).

I added a bunch of "workflow_dispatch" workflow triggers, so that I could test the workflows on my branch.  It seems like some of them only trigger on the "develop" branch.  I'd be happy to remove those triggers if you don't want to allow them, but they do make it easier to test the workflows on a pending PR.

dellaert:
If the triggers lead to more automatic CI runs on GitHub runners, then, yes - please remove them.

truher:
oh, no, it's not an automatic trigger, it's manual.  it's the thing that puts the little "run workflow" button on the github UI, but only for users with write access.

truher:
thank you!

PR #2526, opened 2026-05-11
Add QCQP

dellaert:
Beyond QP, this PR adds the C++ QCQP layer for constrained optimization.

The main addition is `QcqpProblem`, which lets us model quadratic-cost problems with both linear and quadratic constraints using the existing constrained optimization framework. It reuses `QpCost` for affine quadratic objectives and adds row-space quadratic cost construction for direct `Vector` and `Matrix` values.

The PR also introduces `QuadraticConstraint`, with equality, less-equal, and greater-equal senses for constraints of the form `trace(X' A X) ~ b`. These constraints are wrapped as nonlinear equality/inequality factors, so they work naturally with the augmented Lagrangian machinery and mixed linear/quadratic constraint sets.

Finally, the PR includes C++ tests and constrained-module docs for the new QCQP path, including evaluation, matrix-valued costs, quadratic constraint behavior, and a small ALM optimization example.


dellaert:
@ProfFan i want to merge this as some pre-ICRA stuff depends on it, but happy to still incorporate comments in future PRs

PR #2528, opened 2026-05-12
Fixes to Eigen's SSE vectorizer patterns from GCC/G++ 14/15

kshaji3:
1. G++/GCC 14/15 is more strict about eigen assignments to be of fixed sizes. These fixes mainly address that, and some unit tests are also written to accommodate that.
2. Weighted sampler had some issues passing unit tests due to how types were not as explicitly defined. This has been fixed in the test setups.


gtsam/base/ProductLieGroup.h
```diff
@@ -210,6 +210,23 @@ class ProductLieGroup : public std::pair<G, H> {
   /// Create an identity Jacobian with the requested runtime size.
   static Jacobian identityJacobian(size_t productDimension);
 
+  /// Assign a sub-block. Uses compile-time block dimensions when the source
+  /// has fixed-size dimensions; otherwise falls back to runtime dimensions.
+  /// The compile-time path avoids GCC -Werror=array-bounds false positives
+  /// on Eigen's SSE vectorizer when the source is a fixed-size 1x1 matrix.
+  template <typename DstType, typename SrcType>
+  static void assignBlock(DstType& dst, size_t row, size_t col,
```
dellaert: I like it, but please move to -inl.h as well. And, I'd prefer this API:

assignBlock(const SrcType& src, row, col, DstType* dst)

dellaert (commented):
Nice. Just a comment on style for output arguments. I strongly prefer the old Google style with output pointers, so the call site is
```
assignBlock(src, 4, 3, &dst)
```
which makes it clear dst is being modified.

kshaji3:
This should cover everything now.

gtsam/base/ProductLieGroup.h
```diff
@@ -210,6 +210,23 @@ class ProductLieGroup : public std::pair<G, H> {
   /// Create an identity Jacobian with the requested runtime size.
   static Jacobian identityJacobian(size_t productDimension);
 
+  /// Assign a sub-block. Uses compile-time block dimensions when the source
+  /// has fixed-size dimensions; otherwise falls back to runtime dimensions.
+  /// The compile-time path avoids GCC -Werror=array-bounds false positives
+  /// on Eigen's SSE vectorizer when the source is a fixed-size 1x1 matrix.
+  template <typename DstType, typename SrcType>
```
dellaert: Can we move to -inl.h ?

gtsam/base/ProductLieGroup-inl.h
```diff
@@ -71,15 +71,13 @@ ProductLieGroup<G, H> ProductLieGroup<G, H>::retract(const TangentVector& v,
       H1 ? &D_h_first : nullptr, H2 ? &D_h_second : nullptr);
   if (H1) {
     *H1 = zeroJacobian(productDimension);
-    H1->block(0, 0, firstDimension, firstDimension) = D_g_first;
-    H1->block(firstDimension, firstDimension, secondDimension,
-              secondDimension) = D_h_first;
+    assignBlock(D_g_first, 0, 0, &*H1);
```
dellaert, line 95: Maybe: H1->get()

dellaert:
Interesting: 26 fails with:
```
265/330 Test #265: testSubgraphSolver ............................Subprocess aborted***Exception:   0.08 sec
/usr/include/c++/15/bits/random.h:1901: std::uniform_real_distribution<_RealType>::param_type::param_type(_RealType, _RealType) [with _RealType = double]: Assertion '_M_a <= _M_b' failed.
```
Maybe enlist the help of the AI to fix this one. Seems specific to GCC15.

gtsam/base/ProductLieGroup.h
```diff
@@ -210,6 +210,14 @@ class ProductLieGroup : public std::pair<G, H> {
...
+  static void assignBlock(const SrcType& src, size_t row, size_t col,
```
dellaert: by declaring this here, you make it part of the API. But this is just an internal function, that could be defined in an anonymous namespace inside -inl.h, I think?

wrap/pybind11/.github/workflows/ci.yml
```diff
@@ -181,7 +181,7 @@ jobs:
     runs-on: ${{ matrix.runs-on }}
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
```
dellaert: Ah, `wrap` is a git subtree, and has its own repo. gtsam PR's should not make any changes in it.

dellaert (commented):
Two more issues, almost there :-)

wrap/pybind11/.github/workflows/ci.yml
kshaji3: Fixed.

kshaji3: Was not aware about that, my bad

gtsam/base/ProductLieGroup.h
kshaji3: This is fixed in the most recent commit.

gtsam/3rdparty/Eigen/Eigen/src/Core/util/DisableStupidWarnings.h
```diff
@@ -71,6 +71,15 @@
     // See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89325
     #pragma GCC diagnostic ignored "-Wattributes"
   #endif
+
```
dellaert: Sorry, once again, a “disallowed” edit: this is the vendored Eigen library - we don’t edit that.

kshaji3: Ah okay got it

kshaji3: I suspected that was what you would say, but I git diffed the 3rd-party folder and found there were some local changes that you were making (I believe, which AI did as well). What's the general rule on that?

kshaji3:
I copied over our use of anonymous namespace in ProductLieGroup.h to handle that error. I believe that should be consistent with the way you would do it?

gtsam/3rdparty/Eigen/Eigen/src/Core/util/DisableStupidWarnings.h
dellaert: The rule is we shouldn’t edit :-)

dellaert (approved):
Looks good now; and I guess a good onboarding experience :-)

PR #2529, opened 2026-05-13
Fix Continuous-Time GP Notebook

holmesco:
# Changes
- Fix the colab link
- Set up helper function imports properly
- Adjust image rendering


dellaert (approved):
LGTM!

PR #2530, opened 2026-05-16
Use sccache for Windows Python jobs

Gold856:
With a bit of finagling, we can reuse the caches from build-windows in the Windows build-python job. sccache will take into account environment variables when calculating hashes; care has been taken to ensure environment variables are _exactly_ the same between the two workflows and scripts. With that done, sccache is able to reuse the caches from build-windows, which speeds up the Windows Python build significantly: https://github.com/Gold856/gtsam/actions/runs/25951426921

Note that the run is with a nearly 100% cache hit rate, with the only misses being the actual Python object files. Actual speedups will be much lower, but still significant, and as usual, the lack of precompiled headers will penalize complete cache misses, but those are fairly rare, and on average, it will be faster.

### DIFF
--- a/.github/scripts/python.sh
+++ b/.github/scripts/python.sh
@@ -25,6 +25,7 @@ fi
 
 export PYTHON="python${PYTHON_VERSION}"
 NO_BOOST_BUILD=OFF
+SCCACHE=OFF
 
 function install_dependencies()
 {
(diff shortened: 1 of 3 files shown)


dellaert (commented):
Looks good, will merge if CI passes. One question: you seem to have disabled unstable in build-windows.yml?

Gold856:
Unstable is enabled by default, no need to explicitly enable it.

dellaert (approved):
Great!

PR #2525, opened 2026-05-11
LP and QP

dellaert:
This PR moves the LP/QP active-set work out of `gtsam_unstable` and into the stable `gtsam/constrained` module, with a modernized API around direct `Values` entries.

Main additions:

- `LinearConstraint`: shared linear equality / inequality representation around `JacobianFactor`, including `Equal`, `LessEqual`, and `GreaterEqual`.
- `LpProblem` / `LpCost`: linear objectives and constraints over direct `Vector` and `Matrix` values.
- `QpCost` / `QpProblem`: affine quadratic costs backed by `HessianFactor` / Gaussian factors, including exact linearization and row-space matrix quadratic support.
- `ActiveSetSolver`: common active-set solver for LPs and QPs.
  - LP path computes projected descent directions and detects unbounded descent.
  - QP path supports sparse working-graph solves by default.
  - Dense QP mode reuses a Hessian factorization and solves active-set Schur-complement systems.
  - No-initial-value solves use phase-I LP feasibility initialization.
  - Solver state exposes row-ordered equality/inequality multipliers and active inequality rows.
- `QpsParser`: C++17 QPS parser that builds `QpProblem` instances and reports source/line-aware parse errors.

Compatibility work:

- Legacy unstable `LPSolver` and `QPSolver` are now compatibility wrappers over the new constrained-module solver.
- Deprecated unstable LP/QP APIs are guarded by `GTSAM_ALLOW_DEPRECATED_SINCE_V43`.
- Legacy exception behavior is preserved where the wrappers translate constrained-solver failures back to the old unstable exceptions.
- The old unstable QPS parser files are removed in favor of the new stable parser.

Docs and examples:

- Added `LpProblem.ipynb` and `QpProblem.ipynb` documentation notebooks.
- Updated `gtsam/constrained/constrained.md` to point to the new LP/QP problem APIs and `ActiveSetSolver`.
- Added `timing/timeActiveSetSolverComparison.cpp` to compare sparse and dense QP subproblem strategies, including sparse-chain and quadrotor-allocation cases.

Validation performed locally:

- `make -j6 testLpProblem.run testLpSolver.run testQpProblem.run testQpSolver.run testQpsParser.run testLPSolver.run testQPSolver.run`
- `git diff --check develop...HEAD`

All local targeted tests passed.

Suggested review focus:

- Public API shape and naming for `LinearConstraint`, `LpProblem`, `QpProblem`, `QpCost`, `QpsParser`, and `ActiveSetSolver`.
[...]

PR #2532, opened 2026-05-17
Remove unused Boost dependencies

Gold856:
#2525 removed QPSParser, which was the only usage of these dependencies.


dellaert (approved):
Good catch!

PR #2535, opened 2026-05-19
refactor: route Pose2/Pose3 print() through operator<< (no trailing endl)

### LINKED ISSUE #2018: Remove std::endl from print functions
## Issue

Currently, classes append `std::endl` in their print statements, such as `Pose2`:
```c++
void Pose2::print(const string& s) const {
  std::cout << (s.empty() ? s : s + " ") << *this << std::endl;
}
```
It seems to me that this is unnecessary and goes against best practice. It means that users have less control over how they can format their print statements. For example, I would have liked the Python code 
```python
print(f"{global_point} transformed by {origin} into local space -> {transformed}")` 
```
to yield 
```
[5. 5.] transformed by (1, 1, 3.14159) into local space -> [-4. -4.]
```
but instead it gives the following:

Image

This example also demonstrates best practice as followed by Numpy, which does not automatically insert newlines.

## Solution

Remove `std::endl` from print functions. Perhaps it can be implemented so that the endline is only added if the user has not supplied a prefix string. I can create a PR if this solution is desired.

## Caveats

This is an opinionated issue, but I think that allowing the user to format their strings how they wish is strictly better than the current approach. The optional string prefix is handy for debugging and writing quick statements, but for even slightly more elaborate formatting it gets in the way.

mvanhorn:
## Summary

For each of `Pose2`, `Pose3`, `Rot2`, `Rot3`, `Point2`, `Point3`, rewrite the `print(const string& s)` body to delegate to the existing `friend std::ostream& operator<<(std::ostream&, const T&)`, replacing inline formatting duplication. Keep the print API and observable output stable: emit the optional prefix exactly as today (`s` followed by a space when non-empty), stream `*this`, then append a single newline using `"\n"` instead of `std::endl` (`std::endl` flushes the stream, which is what users complained about; `"\n"` matches the existing pattern in newer streaming-based prints elsewhere in geometry). Do not change any `print()` signature, do not touch the Python wrapper (#2018 has a separate ProfFan-led follow-up for `__format__`/streaming surfacing), and do not modify `Cal3*` / camera classes - this is the agreed first folder slice.

## Why this matters

Maintainer dellaert (in issue #2018) accepted the underlying complaint - `print()` always appends `std::endl`, which breaks any embedding inside a larger formatted string in C++ or Python - and asked for a phased fix: "phase out print (in a few releases) in favor of streaming. Some classes already support ostream, with no prefix or endl. If you wanted to make a PR with that (and where you can use it in print, to avoid duplication), that would be a very worthwhile contribution. Could be done one subfolder at a time, e.g, starting with geometry." ProfFan agreed and noted streaming is the right primitive. The scope this PR targets is exactly the maintainer-prescribed first slice: `gtsam/geometry` core types (`Pose2`, `Pose3`, `Rot2`, `Rot3`, `Point2`, `Point3`) where `friend operator<<` already exists.

## Testing

- Behavioral: existing `tests/testPose2`, `tests/testPose3`, `gtsam/geometry/tests/testRot3`, `testPoint3` all pass unchanged (output equality where the test captures stdout; otherwise just exercises the method).
- Visual diff: piping a `pose.print("p:")` through a script that asserts the string starts with `"p: "`, contains the same numeric body as before, and ends with exactly one `'\n'` (no double newline) succeeds.
- Embedding case from the issue: in C++, `std::cout << "prefix " << pose << " suffix" << std::endl;` produces the prefix, the same numeric body that `print()` would emit, and a suffix on the same line - confirming `operator<<` and `print()` are now consistent.

Fixes #2018

AI was used for assistance.


dellaert:
This issue does not address issue nor adds streaming as suggested there.

PR #2538, opened 2026-05-20
More edits for gcc 13

kshaji3:
Added fixes to EqFilter so that it works on GCC 13. Also we added Docker CI images and validated.


dellaert (approved):
Awesome. If CI passes, I'll merge.

dellaert:
gcc 13 passed :-)

PR #2539, opened 2026-05-22
Burer–Monteiro Solver for SDP Relaxations of QCQP-Representable Problems in Certifiable Estimation

zhexin1904:
# Burer-Monteiro solver for SDP relaxations (Riemannian Staircase)

## What this PR does

Adds a **Burer-Monteiro solver for the SDP relaxation** (Shor's relaxation) of factor-graph QCQPs to GTSAM. The solver runs the Riemannian Staircase outer loop over the low-rank-factorized SDP: at each level it solves the rank-K BM problem with the inner local solver, assembles the dual certificate `S = Q + Σ_m λ_m A_m`, and verifies positive semidefiniteness via a sparse minimum-eigenvalue check. A passing certificate proves the rounded estimate is the global optimum of the original QCQP. This is the missing piece needed to do **certifiable estimation** inside GTSAM.

**Worked example: rotation averaging.** The PR ships one application end-to-end — SO(d) synchronization on g2o datasets — to demonstrate the solver and exercise the full pipeline (read graph → QCQP at K = d → climb the staircase → verify → round → extract Rot2/Rot3 estimates).

To support this on top of GTSAM's existing constrained-optimization code, the PR also extends `QcqpProblem` to a column-dimension-parametric form (so the same QCQP can be re-emitted at any rank K), adds QCQP traits to `Rot2` and `Rot3`, and adds a `qcqpFactors` hook on `NonlinearFactor` / `FrobeniusBetweenFactor` that the staircase calls between levels. Adding the next certifiable application (PGO, range-aided SLAM, …) is just an additional factor override plus a trait specialization — the staircase, certificate, lift, and rounding paths don't change.

### `gtsam/constrained/`

- `RiemannianStaircaseOptimizer.{h,cpp}` — outer-loop staircase. Per level: solve the Burer-Monteiro QCQP with the inner solver, assemble the dual certificate `S = Q + Σ λ_m A_m`, verify PSD, lift to rank p+1 on failure. Includes a generic `Layout` (heterogeneous row dims, deterministic offsets, stack/unstack), `roundingSolution` (rank-d SVD plus sign-flip), and per-type extractors: `extractRotations<RotT>` for constrained variables, `extractRowVectors` for unconstrained ones.
- `internal/FastVerification.h` — LOBPCG + ILDL minimum-eigenpair test, gated by `GTSAM_USE_LOBPCG_VERIFICATION`.

### `examples/`

- `CertifiableRotationAveraging.cpp` — worked example of the solver applied to SO(d) synchronization. Single binary for 2D and 3D; takes `--data=<g2o>` and `--dim=<2|3>`; reads the g2o into a pose graph, rebuilds it as a rotation-averaging graph with isotropic noise (SE-Sync κ convention), runs the BM staircase from a random initialization, prints the rounded objective, min eig(S), final rank, and wall time.

### `gtsam/3rdparty/`

- `Optimization/` — header-only LOBPCG (vendored from david-m-rosen/Optimization @ 6d87cc2).
- `Preconditioners/` — ILDL (vendored from david-m-rosen/Preconditioners @ 68a9d24).
- `cmake/HandleLOBPCGVerification.cmake` — opt-in via `GTSAM_USE_LOBPCG_VERIFICATION`. Finds CHOLMOD + BLAS, compiles the two ILDL sources directly into `libgtsam`.

### Tests

- `tests/testQcqpProblem.cpp` (34 tests) — QpCost row-space variants, QuadraticConstraint equality / ≤ / ≥, end-to-end ALM on mixed constraints, and the D-templated trait API across D = 1, d, d+1, d+2.
- `tests/testRiemannianStaircase.cpp` (21 tests) — KKT stationarity of the certificate, lift shape and O(α²) feasibility violation, LOBPCG ↔ Spectra ↔ Dense agreement, layout invariants, `roundingSolution`, both extractors, the pMin auto-pad path, and end-to-end staircase on a 5-Rot2 ring (easy and wrong-winding init).

### QCQP framework

- `gtsam/constrained/QcqpProblem.h` — column-dimension-parametric form. New `InsertQcqpValue<T, D>`, `InsertQcqpConstraints<T, D>`, graph-driven constructor `QcqpProblem(graph, K)`, and `rebuildAt(K)` that re-emits costs and constraints at a new column count. K = d gives the natural QCQP; K > d gives the low-rank factorization the staircase climbs.
- `gtsam/constrained/QuadraticConstraint.{h,cpp}` — convention switched from `trace(X'AX) - b` to `0.5 · trace(X'AX) - b`, making `A` the Hessian of `h` directly (the staircase certificate `S = Q + Σ_m λ_m A_m` then needs no `(A + A^T)` symmetrization). Constructor now also rejects non-symmetric `A` with a clear message.
- `gtsam/constrained/QpCost.cpp` — early-out in `ExpandedBlock` when `columns == 1` (avoids an unnecessary Kronecker copy for the vec form).

### Trait + factor hooks

- `gtsam/geometry/Rot2.h`, `gtsam/geometry/Rot3.h` — added `QcqpIntrinsicDim`, `QcqpValue<D>`, `QcqpConstraints<D>` to the traits. D = 1 keeps the vec(R) form for polynomial relaxations; D ≥ d is the matrix form the staircase uses.
- `gtsam/nonlinear/NonlinearFactor.{h,cpp}` — added a virtual base hook `qcqpFactors(costs, constraints, K)` on `NonlinearFactor`. The default body throws, so `QcqpProblem` fails loudly when fed a graph containing a factor that isn't QCQP-convertible (rather than silently dropping measurements). This is the dispatch point that `FrobeniusBetweenFactor::qcqpFactors` overrides, and the extension hook any future QCQP-convertible factor uses.
- `gtsam/slam/FrobeniusFactor.h` — added the `qcqpFactors(costs, constraints, K)` override on `FrobeniusBetweenFactor<T>` plus two helpers (`qcqpFactorsForVec`, `qcqpFactorsForMatrix`). The staircase calls this once per level via `QcqpProblem::rebuildAt(K)`.

### Build system

- `CMakeLists.txt` — `include(cmake/HandleLOBPCGVerification.cmake)` so the opt-in flag and find_library calls run at configure time.
- `gtsam/CMakeLists.txt` — when `GTSAM_USE_LOBPCG_VERIFICATION` is set, appends the ILDL sources to `gtsam_srcs`, sets the per-file compile options that silence upstream warnings, defines `GTSAM_USE_LOBPCG_VERIFICATION` publicly on the `gtsam` target, and adds the Optimization / Preconditioners / CHOLMOD include dirs + link libraries.

### Docs + credits

- `gtsam/constrained/constrained.md`, `gtsam/constrained/doc/QcqpProblem.ipynb` — updated to describe the K-parametric QCQP form and the staircase entry point.
- `THANKS.md` — added the Northeastern contributors (Zhexin Xu, Hanna Jiamei Zhang, David M. Rosen).

## Benchmark

Random initialization, LOBPCG certificate, single-threaded ours vs. SE-Sync. 

| Dataset | F (ours) | F (SE-Sync) | Time (ours) | Time (SE-Sync) |
|---|---:|---:|---:|---:|
| CSAIL (2D) | 22.3131 | 22.3131 | 1.80 s | 0.006 s |
| intel (2D) | 3.63965 | 3.63965 | 5.77 s | 0.015 s |
| smallGrid3D (3D) | 484.976 | 484.976 | 0.17 s | 0.009 s |
| sphere2500 (3D) | 885.363 | 885.363 | 12.71 s | 0.245 s |

Same global optimum on every dataset; both certified.

## Followups

More efficient inner solver and more problem examples


zhexin1904:
Will split it into two PRs:

1. Changes to the QCQP support code.
2. The new Burer–Monteiro (BM) solver and examples.

dellaert:
Great/ Also, I notice you added new optimizers and preconditioners in third-party - this was not the plan we discussed, though? I thought in this first version you were simply going to use GTSAM’s ALM?

zhexin1904:
> Great/ Also, I notice you added new optimizers and preconditioners in third-party - this was not the plan we discussed, though? I thought in this first version you were simply going to use GTSAM’s ALM?

Yeah, that is only for verification, the method is Locally Optimal Block Preconditioned Conjugate Gradient (LOBPCG), for checking whether the certificate matrix is positive semidefinite. We could use Spectra (already in Shonan Rotation Averaging) as a substitute, or simply use Eigen for small problems. Both Spectra and Eigen are already included in our implementation. ALM is still used as inner local solver for the QCQP.

LOBPCG with the Optimization and Preconditioner is extremely fast and scales well to very large problems. However, it does require adding extra third-party libraries... What's your suggestion?

zhexin1904:
OK for this PRs I think we may remove LOBPCG and mainly use Spectra instead, to keep the changes as minimal as possible. So no additional third-party libraries are needed.

Still suggest adding LOBPCG later :)). We could work on a clean version with minimal external dependencies.

dellaert:
Hi Jason. Yes, please, minimal changes - use Spectra and ALM. Rather that adding LOBPCG as a vendored library, I’d love to chat about why it’s that fast, and then we can check out how to get the same performance.

zhexin1904:
Hi Frank, PR 1 has been submitted, and PR 2 has been pushed to my fork and is ready to be submitted. Of course, if anything needs to be modified, I will apply those changes before submitting it. Thank you~

PR #2541, opened 2026-05-23
Modify for QCQP support for the BM methods.

zhexin1904:
# PR 1 — Modify QCQP support for BM methods

Branch: `feature/QCQP_BM_support`.

**Summary**: generalize the existing QCQP factor traits so the column
dimension `D` can be **any** value `≥ d` (instead of only specific cases
the code previously hard-coded). The BM staircase needs this because
`D = p` varies across staircase levels.

## Each files

- **`Rot2.h`** — generalized QCQP traits to any `D ≥ 2`.
- **`Rot3.h`** — added QCQP traits (and supports any D).
- **`FrobeniusFactor.h`** — `qcqpFactors` now emits matrix-form factors
  at any column count `D ≥ d`.

### DIFF
--- a/gtsam/constrained/tests/testQcqpProblem.cpp
+++ b/gtsam/constrained/tests/testQcqpProblem.cpp
@@ -22,6 +22,7 @@
 #include <gtsam/constrained/QcqpProblem.h>
 #include <gtsam/constrained/QpCost.h>
 #include <gtsam/constrained/QuadraticConstraint.h>
+#include <gtsam/geometry/Rot3.h>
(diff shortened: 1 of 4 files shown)


dellaert:
Rot2 and Rot3 look good. As an FYI, Avinash will add D=1 cases that are *homogenized*, so we have regimes D=1 and D>=3.

I did not carefully check Frobenius factor but will merge anyway. Avinash will touch that one as well.

PS would be lovely to have SE(2|3) in a follow-up PR :-)

PR #2542, opened 2026-05-24
Initial PR2: BM solver for QCQP-Representable Factor Graphs.

zhexin1904:
# PR 2 — Burer-Monteiro Riemannian Staircase solver and RA example

Branch: `feature/QCQP_BM_solver`. Depends on PR 1.

**Summary**: add a Riemannian Staircase solver for the BM-factored SDP
relaxation of a QCQP, using ALM as the inner solver and Spectra for the
minimum-eigenvalue certificate check, with a usage example on 2D / 3D
rotation averaging.

## Benchmark vs SE-Sync (random initialziation, initiaz rank = d)

| Dataset     | F (ours) | F (SE-Sync) | Time ours / SE (s) |
|---|---:|---:|---:|
| CSAIL       | 22.3131 | 22.3131 | 0.48 / 0.02 |
| intel       | 3.63965 | 3.63965 | 2.61 / 0.04 |
| MIT         | 38.8109 | 38.8109 | 0.38 / 0.02 |
| smallGrid3D | 484.976 | 484.976 | 0.19 / 0.02 |
| sphere2500  | 885.363 | 885.363 | 3.99 / 0.38 |

### DIFF
--- a/examples/CertifiableRotationAveraging.cpp
+++ b/examples/CertifiableRotationAveraging.cpp
@@ -1,0 +1,212 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file    CertifiableRotationAveraging.cpp
+ * @brief   Certifiable SO(d) synchronization (d = 2 or 3) via the Riemannian
+ *          Staircase (Burer–Monteiro low-rank SDP + sparse-eigensolver
+ *          certificate). Minimizes Σ_ij κ_ij ‖R_i R_ij − R_j‖²_F over Rot2 / Rot3.
+ * @author  Zhexin Xu
+ * @author  David M. Rosen
+ *
+ * Usage:
+ *   ./CertifiableRotationAveraging --data=<g2o file> --dim=<2|3>
+ */
+
+#include <gtsam/constrained/QcqpProblem.h>
+#include <gtsam/constrained/RiemannianStaircaseOptimizer.h>
+#include <gtsam/geometry/Pose2.h>
+#include <gtsam/geometry/Pose3.h>
+#include <gtsam/geometry/Rot2.h>
+#include <gtsam/geometry/Rot3.h>
+#include <gtsam/nonlinear/NonlinearFactorGraph.h>
+#include <gtsam/slam/BetweenFactor.h>
+#include <gtsam/slam/FrobeniusFactor.h>
+#include <gtsam/slam/dataset.h>
+
+#include <chrono>
+#include <iostream>
+#include <random>
+#include <set>
+#include <string>
+#include <type_traits>
+
+using namespace gtsam;
+
+namespace {
+
+constexpr unsigned int kSeed = 60;
+constexpr double kPi = 3.14159265358979323846;
+
+/// If less than half of the `D × D` entries have positive determinant, negate
+/// every entry's last column. Required before per-block `Rot::ClosestTo`
+/// projects each block to SO(d) — otherwise its sign-correction fires only on
+/// the det < 0 blocks and the rounded rotations end up with inconsistent
+/// signs across keys.
+template <int D>
+void AlignBlockDetSigns(Values& qcqpValues) {
+  size_t numNeg = 0, numBlocks = 0;
+  for (const auto& [key, M] : qcqpValues.extract<Matrix>()) {
+    if (M.rows() != D) continue;
+    ++numBlocks;
+    if (M.determinant() < 0) ++numNeg;
+  }
+  if (numBlocks == 0 || numNeg <= numBlocks / 2) return;
+  for (const auto& [key, M] : qcqpValues.extract<Matrix>()) {
+    if (M.rows() != D) continue;
+    Matrix flipped = M;
+    flipped.col(M.cols() - 1) *= -1.0;
+    qcqpValues.update(key, flipped);
+  }
+}
+
+/// Random Rot2 / Rot3 init at column count `IntrinsicDim` (D ≥ ambient dim).
+template <typename RotT, int IntrinsicDim>
+Values RandomInitial(const std::set<Key>& keys) {
+  Values v;
+  std::mt19937 rng(kSeed);
+  std::uniform_real_distribution<double> uni(-kPi, kPi);
+  for (Key key : keys) {
+    if constexpr (std::is_same_v<RotT, Rot2>) {
+      InsertQcqpValue<Rot2, IntrinsicDim>(key, Rot2::fromAngle(uni(rng)), &v);
+    } else {
+      const Vector3 rpy(uni(rng), uni(rng), uni(rng));
+      InsertQcqpValue<Rot3, IntrinsicDim>(
+          key, Rot3::RzRyRx(rpy(0), rpy(1), rpy(2)), &v);
+    }
+  }
+  return v;
+}
+
+template <typename PoseT, typename RotT, int IntrinsicDim>
+int RunCertifiableRA(const std::string& dataPath) {
+  constexpr bool kIs3D = std::is_same_v<PoseT, Pose3>;
+
+  // Drop translations, wrap each BetweenFactor<PoseT>'s rotation in a
+  // FrobeniusBetweenFactor<RotT> with isotropic precision kappa (SE-Sync):
+  //   2D:  kappa = I(theta, theta)
+  //   3D:  kappa = d / (2 * trace(RotInfo^-1))
+  // g2o initial values are discarded; we use random init below.
+  auto [poseGraph, _] = readG2o(dataPath, /*is3D=*/kIs3D);
+  NonlinearFactorGraph graph;
+  std::set<Key> keys;
+  size_t skipped = 0;
+  for (const auto& f : *poseGraph) {
+    const auto bf = std::dynamic_pointer_cast<BetweenFactor<PoseT>>(f);
+    if (!bf) {
+      ++skipped;
+      continue;
+    }
+
+    const auto gaussian =
+        std::dynamic_pointer_cast<noiseModel::Gaussian>(bf->noiseModel());
+
+    double kappa;
+    if constexpr (kIs3D) {
+      const Matrix6 info =
+          gaussian ? gaussian->information() : Matrix6::Identity();
+      kappa = 3.0 / (2.0 * info.topLeftCorner<3, 3>().inverse().trace());
+    } else {
+      const Matrix3 info =
+          gaussian ? gaussian->information() : Matrix3::Identity();
+      kappa = info(2, 2);
+    }
+
+    auto rotNoise =
+        noiseModel::Isotropic::Variance(traits<RotT>::dimension, 1.0 / kappa);
+    graph.emplace_shared<FrobeniusBetweenFactor<RotT>>(
+        bf->key1(), bf->key2(), bf->measured().rotation(), rotNoise);
+    keys.insert(bf->key1());
+    keys.insert(bf->key2());
+  }
+
+  if (graph.empty()) {
+    std::cerr << "No BetweenFactor<" << (kIs3D ? "Pose3" : "Pose2")
+              << "> found in " << dataPath
+              << ". Wrong --dim, or wrong file format?\n";
+    return 1;
+  }
+  std::cout << "Loaded " << keys.size() << " rotations, " << graph.size()
+            << " edges from " << dataPath << " (d=" << IntrinsicDim << ")";
+  if (skipped > 0) std::cout << "  (" << skipped << " non-Between skipped)";
+  std::cout << "\n\n";
+
+  // Low initial mu + slow growth lets LM make cost progress before the
+  // constraints tighten. Spectra verifier + ALM solver are the defaults.
+  RiemannianStaircaseParams params;
+  params.pMin = IntrinsicDim;
+  params.verbose = true;
+  params.almParams->initialMuEq = 1.0;
+  params.almParams->muEqIncreaseRate = 2.0;
+  params.almParams->maxIterations = 50;
+  params.almParams->absoluteViolationTolerance = 1e-8;
+  params.almParams->relativeViolationTolerance = 1e-8;
+  params.almParams->absoluteCostTolerance = 1e-10;
+  params.almParams->relativeCostTolerance = 1e-10;
+  params.almParams->verbose = false;
+
+  // graph must be QCQP-representable — constructor probes once and throws.
+  RiemannianStaircaseOptimizer rso(
+      graph, RandomInitial<RotT, IntrinsicDim>(keys), params);
+
+  const auto t0 = std::chrono::steady_clock::now();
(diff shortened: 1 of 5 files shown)


gtsam/constrained/RiemannianStaircaseOptimizer.h
```diff
@@ -0,0 +1,329 @@
...
+  /// 3 for Rot3 or mixed). Initial values narrower than pMin are zero-padded
+  /// inside optimize().
+  size_t pMin = 2;
+  size_t pMax = 10;            ///< staircase cap before giving up.
+  double alpha = 1e-2;         ///< descent step size on saddle escape.
+
+  VerificationMethod verificationMethod = VerificationMethod::Spectra;
+  double eta = 1e-3;           ///< min-eigenvalue tolerance for certification.
+
+  size_t maxSpectraIters = 10000;
```
dellaert: wow?

gtsam/constrained/RiemannianStaircaseOptimizer.cpp
```diff
@@ -0,0 +1,579 @@
...
+#include <stdexcept>
+
+namespace gtsam {
+
+// ============================================================================
+// Layout
+// ============================================================================
+
+RiemannianStaircaseOptimizer::Layout
+RiemannianStaircaseOptimizer::Layout::From(const Values& values) {
```
dellaert, line 39: This Layout class is very similar I think to the existing (and powerful) VerticalBlockMatrix we have, except transposed? Pls evaluate whether we could just use that.

gtsam/constrained/RiemannianStaircaseOptimizer.h
```diff
@@ -0,0 +1,329 @@
...
+  /// override `NonlinearFactor::qcqpFactors`; the base implementation
+  /// throws).
+  ///
+  /// A callable is required because lifting changes the column count K =
+  /// p, and the K-Kronecker'd Q / A_m blocks inside the QCQP factors are
+  /// sized by K — a single `QcqpProblem` is parameterized at one K only.
+  ///
+  /// Probes `QCQP(pMin)` once at construction so a non-QCQP-representable
+  /// graph fails here, not mid-iteration.
+  RiemannianStaircaseOptimizer(std::function<QcqpProblem(size_t)> QCQP,
```
dellaert: This API places a big burden on the user: they have to know the internals of the solver, and know how to create the correct parameterized QCQP. I think maybe that’s inevitable if we start from the QCQP, but if we start from the NonlinearFactorGraph, we can supply the D parameter ourselves, right? Come to think of it, though - maybe not if variables are created using *templating*. We need a runtime D.

gtsam/constrained/RiemannianStaircaseOptimizer.h
```diff
@@ -0,0 +1,329 @@
...
+  /// can override `d`; `optimize()` uses `layout.maxRowDim()`. The result
+  /// is consumed by the per-type extractors below.
+  static RoundedSolution roundingSolution(const Values& Y, const Layout& layout,
+                                          int d);
+
+  /// Rotation extractor: projects each rowDim-d slice of `r.Yd` onto
+  /// `RotT` via `RotT::ClosestTo`. Slices of other sizes are skipped.
+  /// More constrained-variable extractors will be added as needed.
+  template <typename RotT>
+  static std::vector<std::pair<Key, RotT>> extractRotations(
```
dellaert: Ow, that’s not great. We have type-specific machinery in what is supposed to be type agnostic by now. That rounding is something that will have to go in the conversion branch along with QcQpValue, no?

gtsam/constrained/RiemannianStaircaseOptimizer.h
zhexin1904: Yeah, the reason we originally took a QCQP (more precisely, a rank-parameterized QCQP family) was to make the input requirement explicit at the type level: not every NonlinearFactorGraph can be passed to this solver, only one whose factors all override NonlinearFactor::qcqpFactors.
  
But I agree, using a NonlinearFactorGraph as the input would make the API simpler, so the new interface would look something like this?
  1. Riemannian Staircase takes a NonlinearFactorGraph directly.
  2. Internally rebuilds the rank-p QCQP at each staircase level via QcqpProblem(graph_, p).
  3. QCQP-representability is still checked at construction: we call QcqpProblem(graph_, pMin) once inside a try-catch; if any factor hits NonlinearFactor::qcqpFactors's base (which throws), we re-throw as invalid_argument with a clear "graph is not QCQP-representable" message. 

The user does pick D at compile time, but only once. When building initialValues, they call InsertQcqpValue<RotT, D>( ). Most APIs used in the Riemannian staircase are already runtime?
QcqpProblem(graph, k) and qcqpFactors(K) are both runtime.

gtsam/constrained/RiemannianStaircaseOptimizer.h
zhexin1904: OHH! Yes thanks! This is an important point.

dellaert:
Yep, I like that plan. By the way, about co-pilot comments, please evaluate each of them for merit, and resolve all of them either by fixing them or by saying not relevant. Please explicitly resolve so I know which ones you addressed and which one you didn't

gtsam/constrained/RiemannianStaircaseOptimizer.cpp
```diff
@@ -0,0 +1,579 @@
...
+        std::cout << "  [Chol] verified=1 dim=" << n << std::endl;
+      }
+      return {true, -params_.eta, Vector::Zero(n)};
+    }
+  }
+
+  // ----- Stage 2: Spectra two-pass shift trick ----------------------------
+  const int nev = 1;
+  const int ncv = std::min<int>(static_cast<int>(params_.numLanczosVectors),
+                                std::max(2 * nev + 1, n));
```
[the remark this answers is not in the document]
zhexin1904: fixed

gtsam/constrained/RiemannianStaircaseOptimizer.cpp
```diff
@@ -0,0 +1,579 @@
...
+    Matrix padded =
+        Matrix::Zero(X.rows(), static_cast<int>(params_.pMin));
+    padded.leftCols(static_cast<int>(cols)) = X;
+    Y.update(key, padded);
+  }
+
+  // ALM must record per-iteration state so we can pull lambdaEq at the end.
+  if (params_.almParams) {
+    params_.almParams->storeOptProgress = true;
+  }
```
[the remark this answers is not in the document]
zhexin1904: fixed

gtsam/constrained/RiemannianStaircaseOptimizer.cpp
```diff
@@ -0,0 +1,579 @@
...
+      continue;
+    }
+    const QuadraticConstraint& qc = qcFactor->quadraticConstraint();
+    if (m >= lambdaEq.size()) {
+      throw std::runtime_error(
+          "assembleCertificate: lambdaEq shorter than equality constraint "
+          "count.");
+    }
+    const double lambdaM = lambdaEq[m](0) / qc.sigma();
+    const Key key = qc.key();
```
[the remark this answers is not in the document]
zhexin1904: fixed

examples/CertifiableRotationAveraging.cpp
```diff
@@ -0,0 +1,192 @@
...
+constexpr unsigned int kSeed = 60;
+
+/// Random initialization of Rot2 / Rot3 stored at column count D = IntrinsicDim.
+/// For SO(d) (d = 2 for Rot2, 3 for Rot3), D must be >= d. Wider D zero-pads
+/// on the right, which preserves the row-orthonormality constraint X X' = I_d.
+template <typename RotT, int IntrinsicDim>
+Values RandomInitial(const std::set<Key>& keys) {
+  Values v;
+  std::mt19937 rng(kSeed);
+  std::uniform_real_distribution<double> uni(-M_PI, M_PI);
```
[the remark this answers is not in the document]
zhexin1904: fixed

gtsam/constrained/tests/testRiemannianStaircase.cpp
```diff
@@ -0,0 +1,514 @@
...
+                     static_cast<long>(Ynext.size()));
+  for (const auto& [key, Yi] : Ystar.extract<Matrix>()) {
+    const Matrix YiNew = Ynext.at<Matrix>(key);
+    EXPECT_LONGS_EQUAL(Yi.rows(), YiNew.rows());
+    EXPECT_LONGS_EQUAL(Yi.cols() + 1, YiNew.cols());
+    EXPECT(assert_equal(Matrix(Yi), Matrix(YiNew.leftCols(Yi.cols())), 1e-15));
+
+    const auto& slice = layout.sliceOf(key);
+    const Matrix expectedCol = alpha * v.segment(slice.offset, slice.rowDim);
+    EXPECT(assert_equal(expectedCol, Matrix(YiNew.rightCols(1)), 1e-15));
```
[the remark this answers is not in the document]
zhexin1904: Doesn't matter?

gtsam/constrained/tests/testRiemannianStaircase.cpp
```diff
@@ -0,0 +1,514 @@
...
+  const Eigen::SparseMatrix<double> S =
+      RiemannianStaircaseOptimizer::assembleCertificate(qcqp, layout, lambdaEq);
+  LONGS_EQUAL(static_cast<long>(layout.totalDim), S.rows());
+  LONGS_EQUAL(static_cast<long>(layout.totalDim), S.cols());
+
+  // (a) SDP dual cert PSD up to tolerance.
+  const Matrix Sdense(S);
+  Eigen::SelfAdjointEigenSolver<Matrix> es(Sdense);
+  const double lambdaMin = es.eigenvalues().minCoeff();
+  std::cout << "min eig(S) = " << lambdaMin << std::endl;
```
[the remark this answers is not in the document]
zhexin1904: fixed

gtsam/constrained/RiemannianStaircaseOptimizer.h
zhexin1904: set to 1000 now

gtsam/constrained/RiemannianStaircaseOptimizer.cpp
zhexin1904: Yeah I prefer to keep Layout. VerticalBlockMatrix is an int-indexed matrix container, but what we need here is a Key-indexed map, so most of Layout's methods would still be needed even if it sat on top of VerticalBlockMatrix.

gtsam/constrained/RiemannianStaircaseOptimizer.h
zhexin1904: RiemannianStaircaseOptimizer is (should be) type-agnostic now. The QCQP traits may need small additions for more complex variable / factor types beyond rotation averaging (per-type conversion methods at the variable level), but the staircase itself should stay type-agnostic. Will see how that shakes out as I add other variables/factors.

zhexin1904:
Hi Frank, the changes have been pushed and address most of the review comments. One point I am still thinking about is the output of the Riemannian Staircase solver. Currently, it returns the rounded rank-(d) solution as a stacked variable matrix(totally type-agnostic), but extracting the individual variable blocks should likely be type-specific, which we have not proposed yet. I think the appropriate design may become clearer once I add (SE(d)) and other variable types to this framework soon.
 determinant-sign check (AlignBlockDetSigns, global reflection disambiguation) is rotation-specific. I put it in the CertifiableRotationAveraging example for now, but it may move into QcqpProblem.h or become a Rot-specific QCQP trait ? once we have more variable types to compare against.

dellaert:
Hi Jason, Great. I'll take a look tomorrow morning!
The obvious return type is Values, btw, and indeed that rounding API needs to be decided upon and applied. Feel free to make a proposal. I think that should be a PR against the conversion branch.

zhexin1904:
Hi Frank, I just discussed this again with Dave. This is basically along the lines of what you discussed yesterday. Here are a few points we would like to add soon. Does this look good?

## 1. Add more usage examples / test cases

- Chordal initialization
- Chordal initialization + LM + Riemannian Staircase

## 2. Provide different interfaces

### a. Optimizer interface

We can provide both interfaces:

```cpp
RiemannianStaircaseOptimizer(NonlinearFactorGraph)
RiemannianStaircaseOptimizer(QCQP)
```

Internally, the `NonlinearFactorGraph`-based interface can convert the problem to a `QCQP` and then call the QCQP-based optimizer:

```cpp
RiemannianStaircaseOptimizer(NonlinearFactorGraph) {
  // Convert the nonlinear factor graph to a QCQP
  RiemannianStaircaseOptimizer(QCQP);
}
```

### b. Verification interface

We should also provide verification interfaces that do not require multipliers. ALM can return multipliers, but other solvers, including the one used in our paper, cannot. In those cases, we will compute the dual variables as part of the verification procedure.

Therefore, we can provide multiple interfaces with different arguments, while keeping the output the same: the certificate matrix.

```cpp
verify(QCQP problem, Solution solution, Multipliers multipliers)
verify(QCQP problem, Solution solution)
// Additional verification interfaces as needed
```

Internally, all of these interfaces should eventually construct the same certificate matrix and call the same eigenvalue-based verification routine:

```cpp
verify(Certificate certificate)
```

This essentially enables a general workflow in which a solution produced by any local solver can be passed to the dual-certification procedure.

dellaert:
Thanks ! Exciting progress !

About rotation-specific code in BM: that *is* a problem I think. Was your paper not generic on type once you had a QCQP?

New plan looks good.
About the NonlinearFactorGraph-based interface:
1) I think the thin wrapper might not be enough, right? because the conversion has to be called multiple times as we go up the staircase?
2) The verification should ideally also have a NonlinearFactorGraph-based interface + Values obv.

dellaert:
The QCQP conversion dependency has now been merged into `develop`, and this PR was accidentally closed as a result. I have recreated the work on the latest `develop` as draft PR #2608, preserving the original commit and @zhexin1904's authorship; the only follow-up change adds the direct Eigen eigenvalue includes needed by the current build.

@zhexin1904 @avinashresearch1 — please use #2608 for continued review and coordination.

zhexin1904:
> The QCQP conversion dependency has now been merged into `develop`, and this PR was accidentally closed as a result. I have recreated the work on the latest `develop` as draft PR #2608, preserving the original commit and @zhexin1904's authorship; the only follow-up change adds the direct Eigen eigenvalue includes needed by the current build.
> 
> @zhexin1904 @avinashresearch1 — please use #2608 for continued review and coordination.

Hi Frank @dellaert , sorry for the late response. I’ve been tied up (struggling 😅) with some other projects recently. I’ll get back to the staircase solver later this week. I have a few planned changes that I haven’t pushed yet, including certification for the local solver and some feasibility-restoration steps.

dellaert:
> > The QCQP conversion dependency has now been merged into `develop`, and this PR was accidentally closed as a result. I have recreated the work on the latest `develop` as draft PR #2608, preserving the original commit and @zhexin1904's authorship; the only follow-up change adds the direct Eigen eigenvalue includes needed by the current build.
> > @zhexin1904 @avinashresearch1 — please use #2608 for continued review and coordination.
> 
> Hi Frank @dellaert , sorry for the late response. I’ve been tied up (struggling 😅) with some other projects recently. I’ll get back to the staircase solver later this week. I have a few planned changes that I haven’t pushed yet, including certification for the local solver and some feasibility-restoration steps.

No worries! I will now merge to 2608 because it passes CI,  so please add any other changes as a new PR towards develop.

PR #2548, opened 2026-05-26
Add FromQcqpValue (Rot2/Rot3) and ExtractQcqpValues helpers in feature/convert_to_QCQP

zhexin1904:
## Summary

  Add an extract path from the rounded BM solution to the QCQP variable traits, so matrix-form QCQP values can be projected back into typed `Rot2` / `Rot3`. Needed by the Burer–Monteiro / Riemannian Staircase solver in the follow-up PR, which returns a rank-`d` matrix. A step toward type-agnostic extraction.

  ## Changes

  Add `FromQcqpValue` for `Rot2` and `Rot3` (via `Rot::ClosestTo`), and add `ExtractQcqpValues` in `QcqpProblem`.

  ## Notes

  - Vector form (`D = 1`) is intentionally unsupported by `FromQcqpValue` — one column of a rotation matrix doesn't determine the rotation. A `static_assert` makes this explicit at the call site.

### DIFF
--- a/gtsam/constrained/QcqpProblem.h
+++ b/gtsam/constrained/QcqpProblem.h
@@ -99,6 +99,27 @@ void InsertQcqpConstraints(Key key, NonlinearEqualityConstraints* constraints) {
 }
 
 /**
+ * Project matrix-form QCQP variables back into typed values: scan
+ * `qcqpValues` and apply `traits<T>::FromQcqpValue<D>` to each slice whose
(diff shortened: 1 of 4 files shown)


dellaert:
D=1 will be used by Avinash to support vectorized/homogenized QCQP conversions anyway :-)

avinashresearch1:
Yes, I have a couple of the homogenization fixes here: https://github.com/borglab/gtsam/pull/2549 but still WIP

PR #2553, opened 2026-05-29
Fix MATLAB 2024a MEX build: resolve size_t and uint64_t template collision

svenangerer:
Description:
I was attempting to build GTSAM with MATLAB 2024a support on Ubuntu using GCC 11.4. The build failed due to a redefinition of template specializations in matlab.h because size_t and uint64_t resolve to the same underlying type on 64-bit Linux.

Changes:

    Added <type_traits>

    Commented out wrap<uint64_t> and unwrap<size_t> to prevent the compiler error.

Note: I am aware that hard-commenting these out might affect Windows builds where the types differ. I am open to suggestions on how to conditionally compile these using #if macros or <type_traits> to ensure cross-platform compatibility.

My Configuration:

    OS: Ubuntu 22.04 (64-bit)

    Compiler: GNU 11.4.0

    MATLAB: R2024a

### DIFF
--- a/wrap/matlab.h
+++ b/wrap/matlab.h
@@ -27,6 +27,7 @@
 #include <gtsam/geometry/Point2.h>
 #include <gtsam/geometry/Point3.h>
 #include <gtsam/base/utilities.h>
+#include <type_traits>
 
 using gtsam::Vector;
 using gtsam::Matrix;
@@ -37,7 +38,6 @@ extern "C" {
 #include <mex.h>
 }
 
-#include <limits>
 #include <list>
 #include <set>
 #include <sstream>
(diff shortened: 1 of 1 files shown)


dellaert:
Thanks. Two issues:
- wrap is a subtree so it would be a PR on the borglab/wrap repo.
- How does it relate to the recent PR on that repo? One of them I think specifically touched on issues like this and also did the appropriate guarding.

svenangerer:
Thanks! I took a look at the borglab/wrap repo and saw that PR [#183](https://github.com/borglab/wrap/pull/183) recently addressed this for Windows (_WIN64).

However, because I am building on 64-bit Ubuntu, _WIN64 isn't defined, so the size_t / uint64_t template collision was still causing the MATLAB build to fail on my machine.

I have closed this PR and opened a new one on the wrap repository to extend those guards to cover 64-bit Unix (__LP64__) as well.

You can find the new pull request here: https://github.com/borglab/wrap/pull/186

PR #2558, opened 2026-06-05
Fix docs for ISAM2Params thresholds example

kriato:
Just cleaning up the docs

Upstream

image

PR

image


dellaert:
Marco, thanks ! But I'm hesitant to run full CI and generate a new wheel just for this tiny change. Would you be willing to do a broader sweep of doc issues in this PR so it's worth the CO_2 emissions ? :-) 

Note we use clang-format, Googly style for formatting.

Maybe an AI could help you track down similar issues and we can land all of them in one sweep.

dellaert:
Asked codex to address my comment. Found a few more... Thanks !

PR #2561, opened 2026-06-08
Optimize compile times

Gold856:
Let's get some stats for this, because some of these changes are pretty invasive, and need a good justification...

On my AMD Ryzen 9 7940HS, I compiled gtsam 12 times, with the first 2 samples dropped to account for laptop warmup, with 3 branches to test different optimizations. I built `all.tests` with Ninja and MSVC on develop (was fc790b0916a81a56077fc197ad8259d5efa63972) and measured the time using a PowerShell script:
```pwsh
function rebuild() {
    cmake --preset default --fresh 2>&1 > $null
    cmake --build build --target clean > $null
    Measure-Command { cmake --build build --target all.tests }
}
function cycle() {
    git switch develop
    rebuild
    git switch optimize-compile-times
    rebuild
    git switch optimize-compile-times-w-matrix-constants
    rebuild
}
cycle
cycle
...
```

I used this CMakeUserPresets.json to configure CMake:
```json
{
  "$schema": "https://cmake.org/cmake/help/latest/_downloads/3e2d73bff478d88a7de0de736ba5e361/schema.json",
  "version": 8,
  "configurePresets": [
    {
      "name": "default",
      "binaryDir": "build",
      "generator": "Ninja",
      "cacheVariables": {
        "GTSAM_BUILD_EXAMPLES_ALWAYS": false,
        "GTSAM_BUILD_WITH_PRECOMPILED_HEADERS": false,
        "GTSAM_ALLOW_DEPRECATED_SINCE_V43": false,
        "GTSAM_ENABLE_BOOST_SERIALIZATION": false,
        "CMAKE_EXPORT_COMPILE_COMMANDS": true,
        "GTSAM_USE_BOOST_FEATURES": false,
        "CMAKE_BUILD_TYPE": "Release"
      }
    }
  ]
}
```
For develop, builds took an average of 873.387 seconds to complete. For the first 5 commits of this PR (various include cleanups only), it took an average of 850.620 seconds to complete. With the addition of `Remove Matrix and Vector constants`, this dropped to an average of 824.932 seconds to complete. In total, it's a 5.55% reduction in compile time, which is less than I would've liked, but I doubt I am able to extract more out without contorting the codebase even more.

As a bonus, I had a hunch that MSVC might not like the extensive use of the comma operator and having to do codegen for all those functions, so I had another batch of runs (14, 2 samples dropped) to remove the use of comma init for matrices and vectors. It was within the noise, and was _really_ invasive, so it's not in this PR (although I personally like the initializer_list syntax more, so that can be done separately if desired).

So that means this PR does two major things: attempting to follow IWYU more closely, and removing those Matrix/Vector constants. As hinted at before, a _lot_ of things broke in the process of cleaning up headers that I had to then patch back up. The loss of `<Eigen/Dense>` meant I had to include Eigen headers in various places in order to get it to compile again, same with a bunch of other headers. The header dependencies here are quite extensive and difficult to untangle, and this could potentially impact downstream users if they are relying on a transitive header include, but that's bad practice, so I think it's fine, and it shouldn't be too bad to fix up. Hopefully downstream users try to follow IWYU.

As for the Matrix/Vector constants, turns out instantiating 21 templates across ~600 files does not make for very good compile times... the cost is that instead of a nice, clean `I_4x4`, it's now `Matrix4::Identity()`, etc, _everywhere_. So it's either accept the verbose syntax, or try to workaround it. Static constant in each TU? Macro in the Matrix/Vector headers? (ew) But I think either way, having all these constants in headers is hard to accept going forward.

<details>
<summary>Raw times in seconds for each build</summary>

Raw times for develop:
```
873.2496393
872.3945005
870.0719114
874.0157575
874.4738764
873.8862947
874.2810931
874.0959795
873.0934347
874.3111212
```

Raw times for first 5 commits:
```
848.4264968
847.1430302
850.0603893
851.6237344
850.326302
851.3154186
852.2742136
851.1269761
853.2655984
850.639124
```

Raw times for the whole PR:
```
815.5565957
825.6354879
828.3017271
826.3272054
828.9986909
824.4908644
824.4115947
824.8515002
825.531017
825.2173939
```
</details>

### DIFF
--- a/gtsam/base/MatrixConstants.h
+++ b/gtsam/base/MatrixConstants.h
@@ -1,0 +1,46 @@
+/* ----------------------------------------------------------------------------
+
+ * GTSAM Copyright 2010, Georgia Tech Research Corporation,
+ * Atlanta, Georgia 30332-0415
+ * All Rights Reserved
+ * Authors: Frank Dellaert, et al. (see THANKS for the full author list)
+
+ * See LICENSE for the license information
+
+ * -------------------------------------------------------------------------- */
+
+/**
+ * @file    MatrixConstants.h
+ * @brief   Macros for Matrix constants to avoid excessive template
+ * instantiation. Avoid using in headers to prevent name pollution.
+ * @author  Gold856
+ */
+
+#pragma once
+
+#define I_1x1 Matrix1::Identity()
+#define Z_1x1 Matrix1::Constant(0.0)
+
+#define I_2x2 Matrix2::Identity()
+#define Z_2x2 Matrix2::Constant(0.0)
+
+#define I_3x3 Matrix3::Identity()
+#define Z_3x3 Matrix3::Constant(0.0)
+
+#define I_4x4 Matrix4::Identity()
+#define Z_4x4 Matrix4::Constant(0.0)
+
+#define I_5x5 Matrix5::Identity()
+#define Z_5x5 Matrix5::Constant(0.0)
+
+#define I_6x6 Matrix6::Identity()
+#define Z_6x6 Matrix6::Constant(0.0)
+
+#define I_7x7 Matrix7::Identity()
+#define Z_7x7 Matrix7::Constant(0.0)
+
+#define I_8x8 Matrix9::Identity()
+#define Z_8x8 Matrix9::Constant(0.0)
+
+#define I_9x9 Matrix9::Identity()
+#define Z_9x9 Matrix9::Constant(0.0)
--- a/gtsam/geometry/Cal3_S2.h
+++ b/gtsam/geometry/Cal3_S2.h
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/geometry/Cal3.h>
 #include <gtsam/geometry/Point2.h>
 
--- a/gtsam/geometry/SO3.h
+++ b/gtsam/geometry/SO3.h
@@ -27,8 +27,8 @@
 #include <gtsam/geometry/Kernel.h>
 #include <gtsam/geometry/SOn.h>
 
-#include <vector>
 #include <optional>
+#include <vector>
 
 namespace gtsam {
 
--- a/gtsam/geometry/SO3.h
+++ b/gtsam/geometry/SO3.h
@@ -22,6 +22,7 @@
 
 #include <gtsam/base/Lie.h>
 #include <gtsam/base/Matrix.h>
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/dllexport.h>
 #include <gtsam/geometry/Kernel.h>
 #include <gtsam/geometry/SOn.h>
--- a/examples/AbcEquivariantFilterExample.cpp
+++ b/examples/AbcEquivariantFilterExample.cpp
@@ -18,6 +18,7 @@
  * @date 2025
  */
 
+#include <gtsam/base/VectorConstants.h>
 #include <gtsam/slam/dataset.h>
 #include <gtsam_unstable/geometry/ABC.h>
 #include <gtsam_unstable/geometry/ABCEquivariantFilter.h>
--- a/examples/CombinedImuFactorsExample.cpp
+++ b/examples/CombinedImuFactorsExample.cpp
@@ -36,6 +36,7 @@
 #include <boost/program_options.hpp>
 
 // GTSAM related includes.
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/inference/Symbol.h>
 #include <gtsam/navigation/CombinedImuFactor.h>
 #include <gtsam/navigation/GPSFactor.h>
--- a/examples/IEKF_SE2Example.cpp
+++ b/examples/IEKF_SE2Example.cpp
@@ -31,6 +31,7 @@
  * @date Apr 25, 2025
  * @authors Scott Baker, Matt Kielo, Frank Dellaert
  */
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/geometry/Pose2.h>
 #include <gtsam/navigation/InvariantEKF.h>
 
--- a/examples/IMUKittiExampleGPS.cpp
+++ b/examples/IMUKittiExampleGPS.cpp
@@ -18,6 +18,7 @@
  */
 
 // GTSAM related includes.
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/inference/Symbol.h>
 #include <gtsam/navigation/CombinedImuFactor.h>
 #include <gtsam/navigation/GPSFactor.h>
--- a/examples/ImuFactorsExample.cpp
+++ b/examples/ImuFactorsExample.cpp
@@ -38,6 +38,7 @@
 #include <boost/program_options.hpp>
 
 // GTSAM related includes.
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/inference/Symbol.h>
 #include <gtsam/navigation/CombinedImuFactor.h>
 #include <gtsam/navigation/GPSFactor.h>
--- a/examples/ImuFactorsExample2.cpp
+++ b/examples/ImuFactorsExample2.cpp
@@ -15,8 +15,9 @@
  * @author Robert Truax
  */
 
-#include <gtsam/geometry/PinholeCamera.h>
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/geometry/Cal3_S2.h>
+#include <gtsam/geometry/PinholeCamera.h>
 #include <gtsam/inference/Symbol.h>
 #include <gtsam/navigation/ImuBias.h>
 #include <gtsam/navigation/ImuFactor.h>
--- a/examples/InverseKinematicsExampleExpressions.cpp
+++ b/examples/InverseKinematicsExampleExpressions.cpp
@@ -16,6 +16,7 @@
  * @author Frank Dellaert
  */
 
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/geometry/Pose2.h>
 #include <gtsam/nonlinear/ExpressionFactorGraph.h>
 #include <gtsam/nonlinear/LevenbergMarquardtOptimizer.h>
--- a/examples/NavStateImuExample.cpp
+++ b/examples/NavStateImuExample.cpp
@@ -20,6 +20,7 @@
  */
 
 #include <gtsam/base/Matrix.h>
+#include <gtsam/base/MatrixConstants.h>
 #include <gtsam/geometry/Cal3_S2.h>
 #include <gtsam/geometry/PinholeCamera.h>
 #include <gtsam/geometry/Pose3.h>
--- a/gtsam/base/Matrix.h
+++ b/gtsam/base/Matrix.h
@@ -55,8 +55,6 @@ using Matrix6##N = Eigen::Matrix<double, 6, N>;  \
 using Matrix7##N = Eigen::Matrix<double, 7, N>;  \
 using Matrix8##N = Eigen::Matrix<double, 8, N>;  \
 using Matrix9##N = Eigen::Matrix<double, 9, N>;  \
-static const Eigen::MatrixBase<Matrix##N>::IdentityReturnType I_##N##x##N = Matrix##N::Identity(); \
-static const Eigen::MatrixBase<Matrix##N>::ConstantReturnType Z_##N##x##N = Matrix##N::Constant(0.0);
 
 GTSAM_MAKE_MATRIX_DEFS(1)
 GTSAM_MAKE_MATRIX_DEFS(2)
--- a/gtsam/base/Vector.h
+++ b/gtsam/base/Vector.h
@@ -42,26 +42,16 @@ typedef Eigen::VectorXd Vector;
 typedef Eigen::Matrix<double, 1, 1> Vector1;
 typedef Eigen::Vector2d Vector2;
 typedef Eigen::Vector3d Vector3;
-
-static const Eigen::MatrixBase<Vector2>::ConstantReturnType Z_2x1 = Vector2::Constant(0.0);
-static const Eigen::MatrixBase<Vector3>::ConstantReturnType Z_3x1 = Vector3::Constant(0.0);
-
-// Create handy typedefs and constants for vectors with N>3
-// VectorN and Z_Nx1, for N=1..9
-#define GTSAM_MAKE_VECTOR_DEFS(N)                \
-  using Vector##N = Eigen::Matrix<double, N, 1>; \
-  static const Eigen::MatrixBase<Vector##N>::ConstantReturnType Z_##N##x1 = Vector##N::Constant(0.0);
-
-GTSAM_MAKE_VECTOR_DEFS(4)
-GTSAM_MAKE_VECTOR_DEFS(5)
-GTSAM_MAKE_VECTOR_DEFS(6)
-GTSAM_MAKE_VECTOR_DEFS(7)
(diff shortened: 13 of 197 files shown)


ProfFan:
I do want to get rid of all static variables, but if we are going to do so, I'd like doing it like a macro not global replacing to `Matrixxxx::Identity()`

Gold856:
Yeah, the global replace is really verbose, although I worry about polluting the global namespace. I know when I was doing the replace, I encountered examples of similar identity and zero variables in gtsam's own code that I had to avoid replacing, so hopefully conflicts are rare.

dellaert:
Hmmm. Not in love with this. I like those constants. and this is 5% on Windows, what about Mac/Ubuntu?
I could find myself more in moving them to a small header, for example like below, and including these in .cpp files where used. In headers, we could use Zero and Identity.
This is probably a different, focused PR.
```
// gtsam/base/MatrixConstants.h
#pragma once

#include <Eigen/Core>

namespace gtsam {

#define GTSAM_MAKE_ZERO_MATRIX(M, N) \
  static const auto Z_##M##x##N = Eigen::Matrix<double, M, N>::Zero();

#define GTSAM_MAKE_ZERO_MATRIX_ROW(M) \
  GTSAM_MAKE_ZERO_MATRIX(M, 1)        \
  GTSAM_MAKE_ZERO_MATRIX(M, 2)        \
...
  GTSAM_MAKE_ZERO_MATRIX(M, 8)        \
  GTSAM_MAKE_ZERO_MATRIX(M, 9)

#define GTSAM_MAKE_IDENTITY_MATRIX(N) \
  static const auto I_##N##x##N = Eigen::Matrix<double, N, N>::Identity();

GTSAM_MAKE_ZERO_MATRIX_ROW(1)
...
GTSAM_MAKE_ZERO_MATRIX_ROW(9)

GTSAM_MAKE_ZERO_MATRIX(10, 1)
GTSAM_MAKE_ZERO_MATRIX(11, 1)
GTSAM_MAKE_ZERO_MATRIX(12, 1)
GTSAM_MAKE_ZERO_MATRIX(15, 1)

GTSAM_MAKE_IDENTITY_MATRIX(1)
GTSAM_MAKE_IDENTITY_MATRIX(2)
...
GTSAM_MAKE_IDENTITY_MATRIX(9)

#undef GTSAM_MAKE_IDENTITY_MATRIX
#undef GTSAM_MAKE_ZERO_MATRIX_ROW
#undef GTSAM_MAKE_ZERO_MATRIX

}  // namespace gtsam
```

Gold856:
I'll switch to separate constants headers + macros to only pay for the templates used.

As for Mac and Linux, I don't own a Mac, and WSL2 is the best I can do for now, so I'll take some measurements soon. When I was testing in CI though, it seem to shave about a minute or so off build times, at least for Linux.

Gold856:
gtsam doesn't compile cleanly on my WSL2 instance, but I was able to fix warnings in two spots, and that fixed it. Here's the diff:
```diff
diff --git a/gtsam/constrained/tests/testQpProblem.cpp b/gtsam/constrained/tests/testQpProblem.cpp
index aece854575..8d9ff90259 100644
--- a/gtsam/constrained/tests/testQpProblem.cpp
+++ b/gtsam/constrained/tests/testQpProblem.cpp
@@ -61,7 +61,7 @@ TEST(QpCost, HessianVectorError) {
   const Matrix G = (Matrix(2, 2) << 2.0, 0.5, 0.5, 3.0).finished();
   const Vector g = (Vector(2) << 0.2, -0.3).finished();
   const QpCost cost(HessianFactor(x0, G, g, 1.7));
-  const Vector x = (Vector(2) << 1.0, 2.0).finished();
+  const Vector2 x{1.0, 2.0};
   const double expected = 0.5 * (1.7 - 2.0 * x.dot(g) + x.dot(G * x));
 
   EXPECT_DOUBLES_EQUAL(expected, cost.error(VectorValue(x)), 1e-12);
@@ -210,7 +210,7 @@ TEST(QpProblem, Evaluate) {
 
   const Values values = VectorValue(1.0, 2.0);
   const auto [cost, eqViolation, ineqViolation] = problem.evaluate(values);
-  const Vector x = (Vector(2) << 1.0, 2.0).finished();
+  const Vector2 x{1.0, 2.0};
   const double expectedCost = 0.5 * (1.7 - 2.0 * x.dot(g) + x.dot(G * x));
 
   EXPECT_DOUBLES_EQUAL(expectedCost, cost, 1e-12);
diff --git a/gtsam/linear/JacobianFactor.cpp b/gtsam/linear/JacobianFactor.cpp
index 87b251df51..979ca12e0a 100644
--- a/gtsam/linear/JacobianFactor.cpp
+++ b/gtsam/linear/JacobianFactor.cpp
@@ -239,7 +239,7 @@ std::tuple<FastVector<DenseIndex>, DenseIndex, DenseIndex> _countDims(
   }
 #endif
 
-  return std::make_tuple(varDims, m, n);
+  return std::make_tuple(std::move(varDims), m, n);
 }
 
 /* ************************************************************************* */
 ```
With this diff, on Ubuntu 26.04 with GCC 15.2.0, `develop` (which has moved on to dfa71a30aecef46ed189822506c30d1839537687) took an average of 459.971 seconds to build, while this branch took 426.364 seconds, a 7.31% speedup(!!). I built it the same way as before, 12 samples, dropped the first two for 10 total samples.

My build script:
```bash
function rebuild() {
    cmake --preset default --fresh -B build &> /dev/null
    cmake --build build --target clean > /dev/null
    time cmake --build build --target all.tests
}
function cycle() {
    git switch develop
    rebuild
    git switch optimize-compile-times
    rebuild
    git switch optimize-compile-times-w-matrix-constants
    rebuild
}
cycle
...
```
<details>
<summary>Raw times in seconds for each build</summary>

Raw times for develop:
```
real    7m41.048s
real    7m39.640s
real    7m40.842s
real    7m40.007s
real    7m46.880s
real    7m38.309s
real    7m38.008s
real    7m38.712s
real    7m37.381s
real    7m38.891s
```

Raw times for this branch:
```
real    7m7.923s
real    7m6.429s
real    7m8.757s
real    7m7.533s
real    7m5.544s
real    7m6.553s
real    7m5.750s
real    7m4.666s
real    7m5.495s
real    7m4.994s
```
</details>

Gold856:
I've noticed that a lot of includes are out of order (in some cases in reverse alphabetical order). I've tried my best to match what clang-format would do, but I also didn't realize that uppercased file names come before lowercase file names... Since I'm touching several headers anyways, would it be a good idea to run clang-format-diff to clean up the includes that I touched anyways? (which I'd do in a separate commit for reviewability) And should I also just run clang-format-diff on the whole PR anyways, given that the Matrix/Vector constants change breaks line length in quite a few places? (I would format and amend each commit except for the Matrix/Vector constants commit if so, because I think having a separate format commit is to be avoided if possible) Unformatted code hurts my eyes...

The downside is the diff of this PR would nearly double; running clang-format-diff on just the last commit alone produced an additional patch with +417, -462 lines, which is quite large.

Either way, I finished up the Matrix/Vector constant macros. I think it does end up belonging in this PR since it plays a major role in reducing compile times. The header cleanup is a small diff though, so perhaps the macro switch should be moved to a different PR, although you could also argue that since it's so small, the fact that there's a common goal (reducing compile times) outweighs splitting up the PR. (Which I would agree with. The diff isn't _too_ bad, at least without the formatting fixes mentioned above)

dellaert:
I prefer to keep formatting/header ordering out of this PR - although it's fine to re-order headers in files you're touching already. Blanket re-formatting however makes reviewing and history very challenging.

I hit CI on this PR, please let me know what the status is and whether you think it's ready to merge (modulo above comment).

Gold856:
Everything should be ready to go now. I ran all my commits through clang-format-tidy and am mostly happy with the results (although it did explode the diff size, and testGPSFactor really took a big hit; that file can be reverted if desired), and I ran CI on my fork, and it's all green (except for some runners exploding due to unrelated reasons). One of the changes I haven't really mentioned yet is the refreshed precompiled header system, which uses the CMake built-in `target_precompile_headers` command (requires CMake 3.16), is usable across compilers, and has a few more headers that showed up as hot paths when I profiled with vcperf. That made PCH builds _significantly_ faster compared to before.

Same deal as before, 12 runs, dropped the first two on MSVC. Compiler version unfortunately isn't the same, but that's why I built `develop` for comparison as well. The last run is noticeably higher in the raw data, likely because I unlocked my laptop during that time and it used more compute/power, but it's not that much (especially because there's 9 other entries) so I'll leave this in. Although I have the extra 2 runs and can swap out the last run for the 2nd warm up run for a better fit if desired.

<details>
<summary>CMake preset file</summary>

```
{
  "$schema": "https://cmake.org/cmake/help/latest/_downloads/3e2d73bff478d88a7de0de736ba5e361/schema.json",
  "version": 8,
  "configurePresets": [
    {
      "name": "default",
      "binaryDir": "build",
      "generator": "Ninja",
      "cacheVariables": {
        "GTSAM_BUILD_EXAMPLES_ALWAYS": false,
        "GTSAM_BUILD_WITH_PRECOMPILED_HEADERS": true,
        "GTSAM_ALLOW_DEPRECATED_SINCE_V43": false,
        "GTSAM_ENABLE_BOOST_SERIALIZATION": false,
        "CMAKE_EXPORT_COMPILE_COMMANDS": true,
        "GTSAM_USE_BOOST_FEATURES": false,
        "CMAKE_BUILD_TYPE": "Release"
      }
    }
  ]
}
```
</details>

<details>
<summary>Build script</summary>

`pch-test` is the name of the branch where I reverted `precompiled_header.h`.
```pwsh
function rebuild() {
    cmake --preset default --fresh 2>&1 > $null
    cmake --build build --target clean > $null
    Measure-Command { cmake --build build --target all.tests }
}
function rebuild2() {
    cmake --preset default --fresh -DGTSAM_BUILD_WITH_PRECOMPILED_HEADERS=OFF 2>&1 > $null
    cmake --build build --target clean > $null
    Measure-Command { cmake --build build --target all.tests }
}
function cycle() {
    git switch develop
    rebuild
    rebuild2
    git switch optimize-compile-times
    rebuild
    git switch pch-test
    rebuild
}
cycle
cycle
cycle
cycle
cycle
cycle

cycle
cycle
cycle
cycle
cycle
cycle
```
</details>

|                                                     | Average Time (s) | % time saved vs. develop |
| --------------------------------------------------- | ------------ | ------------------------ |
| `develop`, PCH enabled                              | 907.71056558 | N/a                      |
| `develop`, PCH disabled                             | 954.37365963 | -5.14%                   |
| This PR, PCH enabled                                | 775.56556894 | 14.56%                   |
| This PR, precompiled_header.h reverted, PCH enabled | 880.65893091 | 2.98%                    |

I was actually surprised it dropped that much, but I suppose I did add almost all the headers that WPA showed with high Wall Clock Time Responsibility. I also enabled it everywhere in CI, except for CIBW (for now) and any job that uses sccache, although I'm not too sure if GCC and Clang get much of a speedup from PCH. Certainly doesn't seem to hurt.

I also kicked out a bunch of code from types.h in the hopes that it improves parse time. It _is_ included in ~600 TUs after all, and MSVC's preprocessor is slow, especially with `/permissive-`. Side effect is that `_USE_MATH_DEFINES` has to be defined globally and needs to be a usage requirement, otherwise a stray `<cmath>` include can prevent pi constants from being defined (if you prefer the in-header approach, I can roll it back). If/when you upgrade to C++20, you'll hopefully be able to use `<numeric>` for the pi constant instead. I can actually remove a bit more, since the Eigen alignment stuff isn't needed for C++17, but I think that deserves its own PR.

Also I've run vcperf a ton of times now, so if you want, I can re-profile after this is merged and try and export a subset of the data for everyone to poke at. The resulting .etl file is... 18 GB so it needs to be pared down and curated first but that's doable (and it takes like 48 GB of RAM to load, but I have 96 so I can load it all just fine).
<details>
<summary>Raw times</summary>

Raw times for develop, PCH enabled:
```
905.5236008
906.6030145
906.2655522
905.3155451
904.7513883
905.6578477
905.6361925
906.5629976
912.3453032
918.4442139
```

Raw times for develop, PCH disabled:
```
951.23317
951.1641204
951.682251
954.1619203
953.4820664
951.3776549
951.8315943
954.1610627
958.6928439
965.9499124
```

Raw times for this PR, PCH enabled:
```
772.3918545
772.4276521
773.2841412
772.3230228
773.90059
774.5425797
772.9011432
776.4598346
781.5070456
785.9178257
```

Raw times for this PR, but with precompiled_header.h reverted to develop, PCH enabled:
```
875.2175647
877.8504669
876.363438
876.2229112
877.8145588
878.8414424
879.1445303
883.1285998
885.7777469
896.2280501
```

gtsam/base/MatrixConstants.h
```diff
@@ -0,0 +1,43 @@
...
+#define I_5x5 Matrix5::Identity()
+#define Z_5x5 Matrix5::Constant(0.0)
+
+#define I_6x6 Matrix6::Identity()
+#define Z_6x6 Matrix6::Constant(0.0)
+
+#define I_7x7 Matrix7::Identity()
+#define Z_7x7 Matrix7::Constant(0.0)
+
+#define I_9x9 Matrix9::Identity()
```
dellaert, line 45: 8?

gtsam/geometry/Cal3_S2.h
```diff
@@ -104,8 +104,8 @@ class GTSAM_EXPORT Cal3_S2 : public Cal3 {
   Cal3_S2 between(const Cal3_S2& q,
                          OptionalJacobian<5, 5> H1 = {},
                          OptionalJacobian<5, 5> H2 = {}) const {
-    if (H1) *H1 = -I_5x5;
-    if (H2) *H2 = I_5x5;
+    if (H1) *H1 = -Matrix5::Identity();
```
dellaert: Sorry, I agree with putting the definitions in the header, but not with the replacements. I think the macros are more clear.

gtsam/geometry/SO3.h
```diff
@@ -155,9 +155,9 @@ struct GTSAM_EXPORT ExpmapFunctor {
   ExpmapFunctor(const Vector3& axis, double angle);
 
   /// Rodrigues formula
-  inline Matrix3 expmap() const { return I_3x3 + A * W + B * WW; }
+  inline Matrix3 expmap() const { return Matrix3::Identity() + A * W + B * WW; }
```
dellaert: Especially but not limited to cases like this.

dellaert (commented):
Awesome wins! but I still have two main comments:
- now there many separate things in this PR, and the size of the PR makes it really hard to see their footprint.
- i really like the macros, and see no reason to replace them. I’d rather do the inverse: use them everywhere they are not yet used.

My suggestion: 
1) take out all non macro/constant related changes and create a separate PR, would be nice if those make a difference at all. Easy to do with smartgit or gitkraken…
2) undo the macro replacements
3) optionally: do the reverse :-)
4) then re-target this PR to that branch

gtsam/base/MatrixConstants.h
Gold856: Never was used apparently. But for completeness I can put it back.

dellaert: yes, please.

dellaert, line 45: Still need the 8-versions :-)

dellaert (approved):
LGTM, modulo 8

gtsam/base/MatrixConstants.h
Gold856: Weird. I wrote the change, but it got lost somewhere. Fixed now.

dellaert:
After latest merge we have conflicts :-( Sorry !

dellaert:
Merged !!!

PR #2565, opened 2026-06-12
Add Alpine Linux (musl libc) images to the Linux CI matrix

LockedThread:
### Summary

Adds musl libc coverage to Linux CI using the new Alpine images from
borglab/docker-images (companion PR: https://github.com/borglab/docker-images/pull/12).

**`build-linux.yml`** — four new matrix entries, each running Debug and Release
(8 new jobs). They run on `ubuntu-24.04` runners and select the Alpine container
via the existing `image_os` override (same pattern as `ubuntu-26.04-gcc-15`):

- `alpine-3.22-gcc-14`
- `alpine-3.22-clang-16`
- `alpine-3.23-gcc-15`
- `alpine-3.24-clang-22`

**`build-special.yml`** — new `alpine-gcc-tbb` job (Alpine 3.22, GCC 14, Release)
alongside `ubuntu-clang-tbb`, exercising GTSAM's TBB allocator/threading path
against oneTBB on musl. The container image expression gained the same
`image_os || os` fallback used by build-linux.yml; existing jobs are unaffected.

**Test fixes for musl** — a full `unix.sh -t` run (Debug, GCC 14, Alpine 3.22)
passed 329/331; the 2 failures were tests that hardcode the exact elimination
ordering METIS produces, which differs (validly) per platform and already has
macOS/QNX and Windows special cases:

- `gtsam/inference/tests/testOrdering.cpp`
- `gtsam/symbolic/tests/testSymbolicBayesTree.cpp`

These now have musl branches (`__linux__ && !__GLIBC__`; musl intentionally
defines no identifying macro) with the orderings/Bayes tree METIS produces on
musl. Both tests pass on Alpine after the fix; glibc branches are untouched.

### Why

- musl differs from glibc in ways that surface real portability bugs (default
  thread stack size, no `execinfo.h`, stricter libc headers).
- The Alpine images also bracket newer toolchains than the Ubuntu lineup,
  including Clang 22 and CMake 4.x.

### Why merge this

- **It catches real bugs Ubuntu CI structurally cannot.** Every current Linux
  CI job links glibc, so glibc-specific assumptions are invisible: GNU-only
  headers (`execinfo.h`), transitive includes that strict musl headers don't
  provide, 8MB default thread stacks (musl: 128KB) hiding stack overflows, and
  allocator-dependent behavior. Users deploying GTSAM on Alpine/embedded-musl
  systems currently find these issues in production; this moves them into PR
  review.
- **The test suite already passes on musl.** A full local Debug build + test
  run on `alpine-3.22-gcc-14` is green (331/331 with the two METIS ordering
  tests taught about musl, exactly as they were for macOS and Windows). The
  matrix entries are validating known-good configurations, not aspirational
  ones.
- **It widens toolchain coverage cheaply.** Alpine ships compilers ahead of
  Ubuntu LTS, so the matrix gains Clang 22 and CMake 4.x, catching
  forward-compatibility breakage early. The images are also about a third the
  size of the Ubuntu ones, so the added jobs pull faster than existing ones.
- **Zero impact on existing jobs.** All Ubuntu entries, their images, and the
  build script are untouched; the `image_os` mechanism is already used by
  `ubuntu-26.04-gcc-15`.

### Validation

- [x] Full Debug build + test suite green on `alpine-3.22-gcc-14` locally
      (331/331 with the test fixes)
- [x] TBB-enabled configure verified on the Alpine image (oneTBB 2022.1 detected)
- [x] Workflow YAML validated

### Requirements

- [ ] `borglab/gtsam-ci:alpine-*` tags published to Docker Hub (from the
      companion docker-images PR)

### Notes

GitHub Actions supports musl containers natively — the runner detects Alpine and
injects its musl Node build for `actions/checkout` etc.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -40,6 +40,11 @@ jobs:
             ubuntu-24.04-clang-16,
             ubuntu-24.04-clang-16-boost,
             ubuntu-24.04-clang-16-geographiclib,
+            # Alpine (musl libc) images; they run on ubuntu runners via image_os
+            alpine-3.22-gcc-14,
+            alpine-3.22-clang-16,
+            alpine-3.23-gcc-15,
+            alpine-3.24-clang-22,
           ]
 
         build_type: [Debug, Release]
@@ -85,6 +90,30 @@ jobs:
             os: ubuntu-24.04
             compiler: clang
             version: "16"
+
+          - name: alpine-3.22-gcc-14
+            os: ubuntu-24.04
+            image_os: alpine-3.22
+            compiler: gcc
+            version: "14"
+
+          - name: alpine-3.22-clang-16
+            os: ubuntu-24.04
+            image_os: alpine-3.22
+            compiler: clang
+            version: "16"
+
+          - name: alpine-3.23-gcc-15
+            os: ubuntu-24.04
+            image_os: alpine-3.23
+            compiler: gcc
+            version: "15"
+
+          - name: alpine-3.24-clang-22
+            os: ubuntu-24.04
+            image_os: alpine-3.24
+            compiler: clang
+            version: "22"
 
         exclude:
           - name: ubuntu-24.04-clang-16-boost
--- a/.github/workflows/build-special.yml
+++ b/.github/workflows/build-special.yml
@@ -19,7 +19,7 @@ jobs:
     name: ${{ matrix.name }} ${{ matrix.build_type }}
     runs-on: ${{ matrix.os }}
     container:
-      image: borglab/gtsam-ci:${{ matrix.os }}-${{ matrix.compiler }}-${{ matrix.version }}
+      image: borglab/gtsam-ci:${{ matrix.image_os || matrix.os }}-${{ matrix.compiler }}-${{ matrix.version }}
       env:
         CMAKE_BUILD_TYPE: ${{ matrix.build_type }}
       volumes:
@@ -35,6 +35,7 @@ jobs:
             ubuntu-clang-deprecated,
             ubuntu-clang-quaternions,
             ubuntu-clang-tbb,
+            alpine-gcc-tbb,
             ubuntu-clang-cayleymap,
             ubuntu-clang-system-libs,
             ubuntu-no-unstable,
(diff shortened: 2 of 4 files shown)


LockedThread:
This PR is dependent on https://github.com/borglab/docker-images/pull/12. I will remove the "Draft" status if https://github.com/borglab/docker-images/pull/12 is merged.

dellaert:
Thanks for the suggestion. I’m generally supportive of changes that improve GTSAM’s portability and compatibility with Alpine Linux. However, I’m hesitant to add Alpine to the official CI matrix.

Our CI resources are focused on the platforms that represent the vast majority of GTSAM users (Ubuntu/Debian-based Linux, macOS, and Windows). Alpine’s musl-based environment introduces a distinct compatibility target with its own maintenance burden, and each additional CI platform increases the long-term cost of development and release management.

I’m happy to review and accept PRs that improve Alpine compatibility, provided they do not negatively affect existing supported platforms. However, I would prefer that Alpine compatibility be validated through CI in a fork, downstream project, or community-maintained repository rather than becoming part of GTSAM’s required CI infrastructure.

If there is substantial evidence of demand from the GTSAM user community in the future, we can revisit whether official Alpine support is warranted.

LockedThread:
> Thanks for the suggestion. I’m generally supportive of changes that improve GTSAM’s portability and compatibility with Alpine Linux. However, I’m hesitant to add Alpine to the official CI matrix.
> 
> Our CI resources are focused on the platforms that represent the vast majority of GTSAM users (Ubuntu/Debian-based Linux, macOS, and Windows). Alpine’s musl-based environment introduces a distinct compatibility target with its own maintenance burden, and each additional CI platform increases the long-term cost of development and release management.
> 
> I’m happy to review and accept PRs that improve Alpine compatibility, provided they do not negatively affect existing supported platforms. However, I would prefer that Alpine compatibility be validated through CI in a fork, downstream project, or community-maintained repository rather than becoming part of GTSAM’s required CI infrastructure.
> 
> If there is substantial evidence of demand from the GTSAM user community in the future, we can revisit whether official Alpine support is warranted.

Okay. I understand the cost of running the CI for Alpine and I will run it within my corporate environment. As for the tests which do not pass in a musl environment, I will make a new PR to fix those.

PR #2573, opened 2026-06-19
[Python] Add batch triangulateSafe supporting incomplete tracks

### LINKED ISSUE #919: Batch triangulation API for faster triangulation
## Feature
In the python wrapper, we have functions `triangulateNonLinear` and `triangulatePoint3`. These functions operate on a single track (i.e. they take input a collection of 2D measurements all belonging to the same 3D point and output one single 3D point.

The request is to have a batch triangulation API, which takes input a set of cameras (2+), collection of 2D tracks (with each track having measurement in all the cameras), and outputs a collection of 3D point, one point for each camera. This new API is equivalent to calling the existing APIs in a loop.

## Motivation
We expect the batch API to perform speed-up when performing triangulation for a 2-camera setup. In GTSFM, we plan to add 2-view bundle adjustment for each image pair.  The first step in this process is to perform triangulation for all the correspondences between the 2 images, which usually run into thousands. 

The triangulation for all the points is achieved by a loop in Python, and it consumes much more time than the final bundle adjustment. By having a batch API, we might be able to save some compute time.

## Pitch
I request for a batch triangulation function, which will also be available in the Python wrapper, which performs the triangulation for a set of 2D tracks.

## Alternatives
N/A

## Additional context
N/A

zcjhao:
## Summary

 Closes #919.
  
Adds a new overload of triangulateSafe that accepts a vector of tracks, where each track is a map from camera index to 2D measurement. This lets callers triangulate many 3D points in a single C++ call instead of looping from Python (which is more expensive since it's crossing the Python and C++ boundary continuously), and naturally supports incomplete tracks where some cameras do not observe a given point.

  ## Changes

  - **`gtsam/geometry/triangulation.h`** — new `triangulateSafe` template function that iterates over tracks, builds a per-track sub-`CameraSet` from the map keys, and delegates to the existing
  single-track `triangulateSafe`. Tracks with fewer than 2 cameras return `Degenerate()` via the existing check.
  - **`gtsam/geometry/cameras.i`** — expose the new overload for `Cal3_S2`, `Cal3DS2`, `Cal3Bundler`, `Cal3Fisheye`, `Cal3Unified`, and `SphericalCamera`.
  - **`python/gtsam/tests/test_Triangulation.py`** — two new tests: complete tracks and incomplete tracks (including the single-camera degenerate case).
  
  ## Test Ran

  - [x] `test_batch_complete_tracks` — 2 cameras, 2 landmarks, all valid
  - [x] `test_batch_incomplete_tracks` — 3 cameras: complete / 2-of-3 / 1-of-3 (degenerate)
  - [x] Full Python test suite: 391 passed, 4 skipped, 0 failures

### DIFF
--- a/gtsam/geometry/cameras.i
+++ b/gtsam/geometry/cameras.i
@@ -659,6 +659,10 @@ gtsam::TriangulationResult triangulateSafe(
     const gtsam::CameraSetCal3_S2& cameras,
     const gtsam::Point2Vector& measurements,
     const gtsam::TriangulationParameters& params);
+std::vector<gtsam::TriangulationResult> triangulateSafe(
+    const gtsam::CameraSetCal3_S2& cameras,
+    const std::vector<std::map<size_t, gtsam::Point2>>& tracks,
+    const gtsam::TriangulationParameters& params);
 
 // Cal3DS2 versions
 gtsam::Point3 triangulatePoint3(const gtsam::Pose3Vector& poses,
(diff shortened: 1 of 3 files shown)


dellaert (approved):
Awesome!

dellaert:
FYI @akshay-krishnan @kathirgounder

PR #2575, opened 2026-06-20
navigation: add undifferenced (PPP) pseudorange and carrier-phase factors

inuex35:
Adds undifferenced (PPP) GNSS factors. In PPP the receiver clock, tropospheric zenith wet delay and slant ionosphere do not cancel by differencing, so they are carried as state variables:

- **`UndifferencedPseudorangeFactor` (+`Arm`)** — keys `[pos/pose, clock, ZTD, slant-iono]`; `error = geodist(sat, rcv) + c*(dt_u - dt_s) + m_w*ZTD + mu_f*I - measured`.
- **`UndifferencedCarrierPhaseFactor` (+`Arm`)** — additionally carries the float ambiguity (`+ lam*N`); the ionosphere enters with the opposite sign (carrier advance).

`m_w` (tropospheric wet mapping) and `mu_f` (first-order iono coefficient) are constants per factor; ZTD and slant iono are estimated nodes. The `Arm` variants take a body-frame lever arm and optional `ecef_T_nav`, matching the existing factors. Analytical Jacobians, Python bindings, and unit tests with numerical-derivative checks (`EXPECT_CORRECT_FACTOR_JACOBIANS`) are included.

Geometry is routed through `gnss::geodist` (Sagnac-corrected), consistent with the double-difference factors. Note: on `develop` the existing `PseudorangeFactor`/`CarrierPhaseFactor` still use a plain Euclidean range; #2574 routes them through `gnss::geodist`, after which all factors compute geometry consistently.


dellaert:
Would it be possible to add a notebook example before I review?

inuex35:
Sure — I can add a short open-sky notebook example. A clean open-sky case avoids the multipath/cycle-slip complications, so it should make the factors easy to follow. I'll push it shortly.

dellaert:
The notebook helps a lot in showing the RTK vs PPP-RTK story.

One documentation clarification: I initially wondered whether the ambiguity state `N` is optimized as an integer inside GTSAM. From the implementation and notebook, it looks like GTSAM estimates `N` as a continuous float ambiguity in cycles, and integer fixing is then performed separately via LAMBDA using `resolve_ar(...)`.

Could we make that explicit in the notebook and the carrier-factor docs? For example:

> The ambiguity state `N` is represented as a continuous `double` in the GTSAM factor graph, in cycles. This gives the float ambiguity estimate. Integer ambiguity fixing is performed separately, e.g. with LAMBDA, using the float estimate and covariance.

That would make the modeling boundary very clear for readers who are newer to carrier-phase GNSS.

I'd also shy away from acronyms like `resolve_ar`. Prefer verbose and clear.

inuex35:
Thanks — both done:
- Documented `N` as a float ambiguity (cycles), with integer fixing as a separate LAMBDA step, in the notebook and carrier-factor docstrings.
- Renamed `resolve_ar` → `resolve_integer_ambiguities`.

dellaert (approved):
Thanks ! LGTM!

PR #2577, opened 2026-06-21
Add compiler hash to vcpkg cache key

Gold856:
After poking around, I realized that the reason why vcpkg caches kept getting invalidated was because the underlying compiler was getting updated. So in order to reliably bust the cache, the cache key must contain the compiler hash. After doing a bit more research, it turns out vcpkg will record compiler hashes in a JSON file, which can be generated by doing a dry run of an install (which avoids unnecessary building). Extracting that hash (or hashes for other platforms) and then incorporating it into the cache key (and not restoring the cache if it's different) is easy enough, and now I believe that the Windows vcpkg cache should stop accumulating dead binaries, which should reduce the amount of cache thrashing.

Test runs:
https://github.com/Gold856/gtsam/actions/runs/27895197804 (first run)
https://github.com/Gold856/gtsam/actions/runs/27917852605 (second run, restored from cache successfully)

### DIFF
--- a/.github/workflows/vcpkg.yml
+++ b/.github/workflows/vcpkg.yml
@@ -78,13 +78,21 @@ jobs:
         env:
           VCPKG_BINARY_SOURCES: default,rw
 
+      - name: Get compiler hash
+        id: hash-compiler
+        shell: bash
+        run: |
+          ./vcpkg/vcpkg install --dry-run
+          cat vcpkg_installed/vcpkg/compiler-file-hash-cache.json
+          echo hash=$(jq '.[].["hash"]' < vcpkg_installed/vcpkg/compiler-file-hash-cache.json) >> $GITHUB_OUTPUT
+
       - name: Restore cache dependencies
         id: cache-restore
         uses: actions/cache/restore@v3
         with:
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Good catch!

dellaert:
@Gold856 seems Ubuntu still errors out almost every time

Gold856:
Yeah, I've been seeing a lot of my own Linux runners die similarly, with long hangs and runners getting "interrupted". Not sure if it's a RAM thing or if GitHub is just being weird.

Assuming it _is_ a RAM thing (because I know this has been a persistent issue), we could try using mold (which we don't use in vcpkg right now). I think it uses less RAM overall, although that may incidentally cause heavier jobs to run at the same the time instead. More likely though we may just need to build the Python targets individually, or with no parallelism. The compilation units might just be too heavy to do 4 at a time. I prefer having parallelism leading up to those targets though.

Also, do we know how much RAM it takes to compile those Python TUs? Would be good to now if we should split them up...

dellaert:
I have no idea. I know python wrappers takes a long time even on my fast Mac. Esp. linking.

PR #2578, opened 2026-06-21
Clean up headers/includes and refresh precompiled headers

Gold856:
Split from #2561. As mentioned there, precompiled headers have been updated to use `target_precompile_headers` (CMake 3.16 feature) which allows use across compilers and is enabled for all non-cached jobs except for cibw. A few hot headers have also been added to the precompiled header, which drops build time by ~12% on MSVC compared to before (this number includes the additional template optimizations, so it's not exactly accurate, but should be fairly close). I also cleaned up types.h and Matrix.h in the hopes that it makes them faster to parse, although I'm not too sure if it actually makes a significant difference and am open to reverting those changes (especially because some of them are breaking changes, although GitHub search would seem to indicate a lack of consumers to break for what I removed). In total, the header/include optimizations made up about half of the original speed up in #2561, so the deeper include optimizations are definitely doing something.


dellaert (approved):
Incredible!

Gold856:
For posterity, I benched PCH on Ubuntu 26.04 in WSL2 with GCC 15.2.0. It made no significant difference: 
PCH enabled:
```
7m9.401s
7m10.482s
7m9.926s
7m10.596s
7m12.310s
7m11.495s
7m10.435s
7m9.665s
7m11.910s
7m11.598s
7m9.729s
7m15.932s
```

PCH disabled:
```
7m9.171s
7m10.056s
7m12.088s
7m12.706s
7m9.946s
7m11.136s
7m11.659s
7m9.490s
7m10.098s
7m10.441s
7m13.977s
7m13.996s
```

So it's fine to leave on in CI I think, although it doesn't actually help, and I think it can hide missing includes, but there's also enough jobs that don't have PCH enabled that it likely makes no difference. I'll leave it up to you if you want to leave PCH enabled for all the Linux/Mac CI jobs.

dellaert:
Thanks for the timings. Is it enabled by default? If it only makes a difference on windows, I think we should guard that with a windows flag, then.

Gold856:
It's enabled by default for Windows, and disabled by default for everything else. I'll put the option under just Windows then.

PR #2579, opened 2026-06-21
Remove aligned operator new macro

Gold856:
It's not needed in C++17. This also aligns with #2578 in removing more code from globally included headers.


dellaert:
That would be awesome. Do you have a source saying that this is not needed for C+ +17 and with the current vendored Eigen version?

Gold856:
https://libeigen.gitlab.io/eigen/docs-nightly/group__TopicUnalignedArrayAssert.html
> If you can target [c++17] only with a recent compiler (e.g., GCC>=7, clang>=5, MSVC>=19.12), then you're lucky: enabling c++17 should be enough

And a similar doc from the 3.4.0 tag on gitlab: https://gitlab.com/libeigen/eigen/-/blob/3.4.0/doc/StructHavingEigenMembers.dox#L11

dellaert:
Cool!!! I think that covers all our CI configurations, so in that case I'm good.

dellaert (approved):
LGTM, modulo CI passing

dellaert:
🥇

PR #2580, opened 2026-06-21
Fix GCC 15 warnings

Gold856:
Broader question: should we more broadly refactor to trying to use initializer lists instead of comma init? I already did all the work as part of making #2561, and I personally think it's much cleaner. It seems like GCC's analyzers can't see past the comma init and throws up its hands. Note that I changed to the fixed size types, but that's unnecessary to fix the warning. I just think it looks cleaner without the extra pair of braces.


dellaert:
I’m all for it, but you sure brace initialization works for Eigen version currently distributed with GTSAM?

Gold856:
Yes, it built locally for me with MSVC using the bundled Eigen.

dellaert:
If you have the diffs for a more broad brace init refactor, go for it. I think it's much better than the alternative. But note it only works for fixed size vectors/ matrices.

dellaert:
Let me know if you are going to undertake the broader brace init refactor here or you want me to merge

Gold856:
Merge this, and I’ll send the brace unit refactor separately (so the branch names in the merge commits make sense).

dellaert (approved):
LGTM, Thanks!!

PR #2584, opened 2026-06-25
Fix(python/windows): fix DLL load failure on Windows for Python wrapper

### LINKED ISSUE #2390: python error
I install python gtsam on windows platform according the INSTALL.md document, when finished, I run the command "import gtsam" in powershell terminal, I encounter the problem "ImportError: cannot import name 'gtsam' from partially initialized module 'gtsam' (most likely due to a circular import) (F:\anaconda3\envs\da3_py310\lib\site-packages\gtsam\__init__.py)", what's the reason of this.

thatdudegrantt:
**Fixes #2390**
**Problem**
On Windows with Python 3.8+, `import gtsam` from a source build fails in two ways:
1. Build-time (MSB3073): The `POST_BUILD` step that renames `gtsam_py.pyd` to `gtsam.pyd `uses a hardcoded path that assumes a single-config generator layout. On MSVC (a multi-config generator), CMake appends the build configuration to `LIBRARY_OUTPUT_DIRECTORY`, so the `.pyd` actually lands in `python/gtsam/Release/gtsam_py.pyd`. The hardcoded source path `python/gtsam/gtsam_py.pyd` does not exist, causing MSB3073 and aborting the entire POST_BUILD chain before the DLL copy step runs.

2. Runtime (misleading circular import error): Python 3.8 changed how Windows searches for DLLs when loading compiled extensions. Before 3.8, Windows automatically searched the directory containing the `.pyd` file. From 3.8 onwards it no longer does, so `gtsam.pyd` cannot find `gtsam.dll` even when co-located, and Python reports a confusing circular import error instead of a clear DLL not found message.

**Fix**
`python/CMakeLists.txt`: Replace the hardcoded `.pyd` source path with `$<TARGET_FILE:gtsam_py>`, a generator expression that resolves to the actual build artifact location regardless of configuration or generator.
`python/gtsam/__init__.py`: Add `os.add_dll_directory()` pointing at the package directory, guarded by `sys.platform == "win32"` and `hasattr(os, "add_dll_directory")` (Python 3.8+). This is the same fix used by NumPy, OpenCV, and Pillow for the identical issue.

**Testing**
Tested on Windows 11 x64, MSVC 19.51 (Visual Studio 18 2026 Community), Python 3.12.10, CMake 4.2.3, GTSAM commit bbe9177469313328345476434393e29676c239be (develop).

Before fix: `cmake --build ` fails with MSB3073 due to an error copying file (if different) from `python/gtsam/gtsam_py.pyd` to `python/gtsam/gtsam.pyd` as the hardcoded source path does not account for the MSVC Release/ subdirectory created by the multi-config generator.
After fix: Build completes cleanly. `gtsam.dll`, `metis-gtsam.dll`, and `cephes-gtsam.dll` are correctly copied into `python/gtsam/` alongside `gtsam.pyd`. End to end verified as `import gtsam`, `NonlinearFactorGraph()`, and `Values()` all succeed on Windows 11 x64 with Python 3.12.10


dellaert:
What's our currently minimum supported version of python? Anything before 3.10 is EOL anyway.

dellaert:
If anything, we should be removing any conditional stuff that looks at versions before 3.10 and double check the documentation that we only support 3.10 and upward.

thatdudegrantt:
GTSAM's wheel builds target 3.11-3.14 (per `.github/workflows/build-cibw.yml`), so the `hasattr `guard for `os.add_dll_directory` (added in 3.8) is unnecessary. I'll simplify to just the sys.platform == "win32" check. Will update the PR.

thatdudegrantt:
So the minimum supported version is python 3.11

dellaert (approved):

PR #2585, opened 2026-06-26
Fix Gal3 maybe-uninitialized workaround by initializing Rt unconditionally

### LINKED ISSUE #2576: Gal3.cpp incorrectly suppresses maybe-uninitialized warning
## Description

Looking at Gal3.cpp, there's a pragma to suppress `maybe-uninitialized` warnings. Seems sane, but I removed it for fun and then looked at the rest of the file and found that it's actually not spurious. The `Rt` local matrix in several functions is actually left uninitialized.

## Expected behavior

I'd expect `Rt` to be set to _something_. All zero? The identity matrix? I'm not a math person, so not sure what a sane default is. And there's also no unit tests that catch this?

aryanputta:
Fixes #2576.

## Analysis
In `Gal3.cpp`, `Create`, `FromPoseVelocityTime`, and `Expmap` each declared

```cpp
Matrix3 Rt;
if (H2 || H3 || H4) Rt = R.transpose();   // Create; analogous guards elsewhere
```

and then used `Rt` only inside the nested `if (H2)` / `if (H3)` / `if (H4)` (or `if (Hxi)`) blocks. So `Rt` is always assigned before any read, there is no real uninitialized use. GCC 15 cannot prove the correlation between the outer `||` guard and the inner per-handle guards, so it raises a `-Wmaybe-uninitialized` false positive. That warning was silenced by a version-gated pragma:

```cpp
#if defined(__GNUC__) && __GNUC__ == 15
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
#endif
```

whose `push` had no matching `pop` anywhere in the file, so the suppression leaked to the rest of the translation unit.

## Fix
`R` is always available in all three functions and `R.transpose()` is a trivial 3x3 op, so `Rt` is now computed unconditionally as a `const`:

```cpp
const Matrix3 Rt = R.transpose();
```

This removes the false positive at its source and lets the version-gated pragma block be deleted entirely. No behavior change: the Jacobian blocks are still written only when their handle is non-null. The only difference is the cheap transpose now runs even when no Jacobian is requested.

## Tests
No functional behavior changed, so existing `testGal3` coverage applies. As a follow-up, the analytic Jacobians of `Create` and `FromPoseVelocityTime` (raised in the issue) could get dedicated `numericalDerivative` checks like the existing `RetractJacobians`/`Expmap` tests; happy to add those in a separate PR.

### DIFF
--- a/gtsam/geometry/Gal3.cpp
+++ b/gtsam/geometry/Gal3.cpp
@@ -20,12 +20,6 @@
  * refer to the aforementioned paper.
  */
 
-// GCC bug workaround
-#if  defined(__GNUC__) && __GNUC__ == 15
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#endif
-
 #include <gtsam/base/Matrix.h>
 #include <gtsam/base/numericalDerivative.h>
 #include <gtsam/geometry/Event.h>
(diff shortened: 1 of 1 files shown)


dellaert:
Thanks, but I think it might have to be solved in a different way. If we don't request a Jacobian, we should not do *any* work. Note if you compile with quaternions it's not so cheap as you might think.

dellaert:
Oh, you killed the pragma. No, let's leave the pragma and the code as is :-)

PR #2589, opened 2026-07-03
navigation: add GNSS Doppler (range-rate) factors

inuex35:
Adds GNSS Doppler (range-rate) factors, complementing `PseudorangeFactor` /
`CarrierPhaseFactor`.

- **`DopplerFactor`** — keys `[velocity, clock bias(k-1), clock bias(k)]`:

      error = e·(v_s - v_r)
            + c·((dt_r(k) - dt_r(k-1))/dt - ddt_s)
            + sagnac_rate
            - (-λ·Doppler)

- **`DopplerFactorArm`** — keys `[pose, velocity, clock bias(k-1), clock bias(k)]`:
  lever-arm variant using `v_antenna = v_body + ecef_R_body·(omega × leverArm)`,
  with an optional `ecef_T_nav` overload; reduces to `DopplerFactor` when omega = 0.

Both include the earth-rotation (Sagnac) rate correction and reuse the
Sagnac-corrected line-of-sight from `gnss::geodist`. Analytical Jacobians
(checked with `EXPECT_CORRECT_FACTOR_JACOBIANS`), Python bindings, and
obj/XML/binary serialization tests are included.


ProfFan (approved):
Thanks! Could you also add a Python test?

inuex35:
Two updates:
Added python/gtsam/tests/test_DopplerFactor.py (Python bindings for both factors).
Aligned DopplerFactorArm with the other lever-arm factors: with ecef_T_nav the velocity is now nav-frame too (rotated to ECEF by ecef_R_nav), matching the pose. Double-checking against real data now — will follow up.

inuex35:
Verified on real data (PPC tokyo/run2, 500 epochs, via the Python bindings): DopplerFactor recovers horizontal velocity to 0.19 m/s RMS (5–12 cm/s open-sky), and DopplerFactorArm's lever term + ecef_T_nav nav-frame velocity check out. No sign/frame issues. I think this is good to merge.

inuex35:
Addressed the Copilot findings: `kBase` for the receiver / `ecef_T_nav` origin in the serialization tests, `velSagnac_` relabeled unitless in `print()`, and the missing `///` doc-comments added. Tests pass.

dellaert:
Thanks for this !! Could you please re-base on develop and possibly add a Python notebook example? We also now typically add notebooks in the doc folder with a "user guide".

dellaert:
@inuex35 I had codex add a documentation notebook just to check the wrapper.  An example notebook from your own application would still be nice, and additional validation of the changes.

inuex35:
Hello, sorry for the late reply. I've added a notebook for Doppler.
I tried to reuse the data from the pseudorange example, but it has no Doppler observations, so I used a same dataset with rtk. It would probably be better to move the pseudorange example over to the same data as well at some point.

dellaert:
I downloaded and ran the example myself but I had to install CSSR lib. So I pushed another commit that puts a pre-processed numpy file in data. 

By the way, I got different results when I used CSSRlib vs CSSRlib-numba.

inuex35:
Could you try again?
Maybe wheel did not work well, so I modifed the package.
And I realized cssrlib's original satelite velocity had a small bug and fixed in cssrlib-numba.

dellaert:
I propose you figure out which one is correct :-) Then update the NPZ file in this PR.

inuex35:
I checked against RTKLIB and cssrlib-numba is the correct one (matches satpos to 2.6e-4 m/s; upstream is off by up to 0.21 m/s). Your NPZ already used cssrlib-numba, so nothing to update — I'll open an issue on hirokawa/cssrlib.

dellaert:
Awesome, merged !!!

PR #2597, opened 2026-07-14
Fix PartialPriorFactor instantiation for vector-space types like Point3

### LINKED ISSUE #2413: PartialPriorFactor is failing with Point3
In C++ on GTSAM 4.3a, when using PartialPriorFactor, on Point3(which is a landmark who's Down axis is having a prior applied) template instantiation fails, the error is 

"dimension is not a member of Eigen::Matrix<double,3,1>".  

It appears GTSAM Point3 typedefs to Eigen Vector3 and is missing some functionality it's supposed to have.

Any suggestions/fixes? I can try adding a really big uncertainty to the other two axis but this is not ideal.

jashshah999:
## Summary

`PartialPriorFactor<Point3>` fails to compile: `evaluateError` accesses `T::dimension`, `T::LocalCoordinates`, and `T::Logmap` as direct static members, which don't exist when `T` is an Eigen typedef like `Point3` (`Eigen::Vector3d`):

```
PartialPriorFactor.h:116:57: error: 'dimension' is not a member of 'Eigen::Matrix<double, 3, 1>'
PartialPriorFactor.h:121:54: error: 'LocalCoordinates' is not a member of ... 'Eigen::Matrix<double, 3, 1>'
```

Fixes #2413

## Changes

- `evaluateError` now goes through `traits<T>` (`traits<T>::dimension`, `traits<T>::Logmap`, and `traits<T>::Local(Identity, p)` for the `GTSAM_ROT3_EXPMAP` branch), matching the GTSAM convention for generic code. For Lie-group types these delegate to exactly the same implementations as before, so behavior is unchanged.
- Added a `PartialPriorFactor<Point3>` unit test (zero error at the linearization point + Jacobian vs numerical derivative).

## Testing

`cmake --build build --target testPartialPriorFactor && ./build/gtsam_unstable/slam/tests/testPartialPriorFactor` — no test failures (all existing Pose2/Pose3 Jacobian tests still pass, confirming identical behavior for Lie types).

### DIFF
--- a/gtsam_unstable/slam/PartialPriorFactor.h
+++ b/gtsam_unstable/slam/PartialPriorFactor.h
@@ -113,18 +113,22 @@ namespace gtsam {
 
     /** Returns a vector of errors for the measured tangent parameters.  */
     Vector evaluateError(const T& p, OptionalMatrixType H) const override {
-      Eigen::Matrix<double, T::dimension, T::dimension> H_local;
+      // Use traits so this also works for types without static members,
+      // e.g. Point3, which is just an alias for Eigen::Vector3d (see #2413).
+      constexpr int Dim = traits<T>::dimension;
+      Eigen::Matrix<double, Dim, Dim> H_local;
 
       // If the Rot3 Cayley map is used, Rot3::LocalCoordinates will throw a runtime error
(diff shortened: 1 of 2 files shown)


dellaert (approved):
Nice! Thanks for the contribution!

PR #2598, opened 2026-07-15
fix: change to direct formulation of FrobeniusPrior to QCQP lift

avinashresearch1:
some discussion with fable attached to show the motivation for this change. section 6 concerns the D = 2 case which I do not fully understand. 

[negative_rotations.pdf](https://github.com/user-attachments/files/30058063/negative_rotations.pdf)


dellaert (approved):
LGTM!

PR #2599, opened 2026-07-16
Fixed lag and python

bboyack:
Added return option of fixed lag smoother to include:
- indices for new factors added during marginalization
- indices for removed factors during marginalization
- keys of nodes deleted during marginalization

Added several functions to the python bindings that hadn't been there before. Also added functions marginalizeLeavesWithIndices to return the marginal and deleted factor indices directly for use in python.
Also added a specific setSmootherLag function to be used in python as well.


dellaert (commented):
Nice!
I agree with most of co-pilots' comments and we should have some unit tests :-)

bboyack:
I fixed the variable naming and added a unit test. If there's anything else, let me know.

dellaert:
CI failure due to missing print body:
Root cause: the Windows Python wrapper link step is missing a definition for  
`gtsam::FixedLagSmoother::Result::print()`, causing:

- `nonlinear.cpp.obj : error LNK2019 unresolved external symbol ... FixedLagSmoother::Result::print(void)const`
- `python/gtsam/gtsam_py.pyd : fatal error LNK1120: 1 unresolved externals`

This happens while building `gtsam_py.pyd` in the Python CI workflow (`.github/workflows/build-python.yml`) on the Windows job.

## Solution

Ensure `FixedLagSmoother::Result::print()` has an out-of-line implementation compiled into `gtsam` (or make it inline in the header if intended).

### Option A (preferred): add/restore `.cpp` definition

In the corresponding source file for `FixedLagSmoother::Result` (typically in `gtsam/nonlinear/`), add:

```cpp
void FixedLagSmoother::Result::print() const {
  // existing/expected print body
}
```

If this method was recently introduced in headers/wrappers but not implemented, this will resolve the linker error immediately.

### Option B: inline in header (if small utility)

If this is meant to be header-only, define it directly in the class declaration:

```cpp
struct Result {
  ...
  void print() const {
    // body
  }
};
```

## Why this is the right fix

- The failure is at **link** time (not compile time), which means declaration is visible, but definition is missing from linked objects/libs.
- The unresolved symbol is referenced from wrapped nonlinear bindings (`nonlinear.cpp.obj`), so Python binding generation is exposing this method and requiring a concrete symbol.
- The workflow and build config are otherwise proceeding to near completion (`367/369`) before failing, consistent with a single missing symbol regression.

## Suggested guardrail

Add a small unit test (or wrapper compile check) that instantiates and calls `FixedLagSmoother::Result::print()` from C++ to catch missing symbol regressions before wrapper link on Windows.

dellaert:
Could also be a missing GTSAM_EXPORT…

dellaert:
(Please point AI to the markdown doc about GTSAM_EXPORT)

dellaert (approved):
Nice addition, thanks!!!

PR #2606, opened 2026-07-24
Implements monolithic and chordal SDP from QCQP intermediate

avinashresearch1:


dellaert:
I’m still reviewing the PR, but one thing I noticed: testMosekSDP.cpp should be moved out of gtsam/inference/tests and into gtsam/certifiable/tests, and also guarded for a MOSEK install.

dellaert:
The CMake organization of certifiable should follow the standard GTSAM module structure. This directory will also contain solver-independent functionality such as Burer–Monteiro, so the module itself should not be treated as an optional MOSEK component.

Please add certifiable to gtsam_subdirs, install its public headers, and use the usual add_subdirectory(tests) and example/test helper conventions. Solver-independent sources should become part of the normal GTSAM module.
Only the MOSEK-specific implementation should be guarded by GTSAM_WITH_MOSEK.

The current arrangement does the reverse: the entire certifiable module is placed in a special uninstalled static library, while HandleMOSEK.cmake injects MOSEK libraries and Fusion sources into libgtsam. That will not scale cleanly once solver-independent certifiable methods are added.

dellaert:
I'll ask Codex to post specific instructions consistent with gtsam.

dellaert:
Codex-generated follow-up with concrete implementation suggestions (reviewed and posted by Frank):

To make the requested structure explicit, I suggest the following:

1. Add `certifiable` to `gtsam_subdirs` in `gtsam/CMakeLists.txt`.
2. Make `gtsam/certifiable/CMakeLists.txt` follow the normal module pattern:
   - install the public headers;
   - call `add_subdirectory(tests)`;
   - do not create a separate `gtsam_certifiable` library.
3. Remove the special top-level `add_subdirectory(gtsam/certifiable)`.
4. Compile solver-independent certifiable sources normally into `gtsam`. Keep MOSEK-specific declarations and definitions behind `GTSAM_USE_MOSEK`, preferably in clearly named backend files such as `MosekSDPSolver.{h,cpp}`.
5. Encapsulate MOSEK/Fusion in a CMake target, e.g. `MOSEK::Fusion`, containing its include directories, library and—in installations that require it—the Fusion sources. Link that target only where the MOSEK backend is compiled. Avoid the global `GTSAM_ADDITIONAL_SOURCES`, `GTSAM_ADDITIONAL_LIBRARIES`, and `GTSAM_ADDITIONAL_INCLUDE_DIRS` mechanism.
6. Move `testMosekSDP.cpp` from `inference/tests` to `certifiable/tests`, and register it conditionally there. Consolidate it with `testMOSEKInstall.cpp`, since both are currently MOSEK smoke checks.
7. Put the production-path SDP tests in `certifiable/tests` and link them against `gtsam`; gate only the tests that actually require MOSEK and a license.
8. Register the Rot2 program through the normal example machinery, rather than manually recreating the `examples` and `EXCLUDE_FROM_ALL` behavior.
9. Since the implementation becomes part of `gtsam`, add `GTSAM_EXPORT` to the public MOSEK-backed specializations and follow the normal installed-header/export conventions.

This gives us a regular, extensible `certifiable` module for Burer–Monteiro and other solver-independent work, with MOSEK isolated as one optional backend rather than defining the module’s build structure.

dellaert:
Also, please add a PR comment.

gtsam/certifiable/LiftedSDPProblem.h
```diff
@@ -0,0 +1,189 @@
+#pragma once
```
dellaert, line 18: In all source files, add the standard GTSAM header, and a doxygen block. Please check other files in GTSAM for examples.

gtsam/certifiable/LiftedSDPProblem.h
```diff
@@ -0,0 +1,189 @@
...
+
+template <typename SDPFormulation, typename SDPSolver>
+class LiftedSDPProblem;
+
+#ifdef GTSAM_USE_MOSEK
+template <>
+class LiftedSDPProblem<MonolithicSDP, MosekSDPSolver> {
+ public:
+
+  explicit LiftedSDPProblem(const QcqpProblem& problem);
```
dellaert, line 67: For all API methods, please add doxygen documentation

gtsam/certifiable/tests/testMOSEKInstall.cpp
```diff
@@ -0,0 +1,70 @@
...
+                   Expr::add(X->index(0, 0), X->index(1, 1)));
+  model->solve();
+
+  auto level = X->level();
+  return {(*level)[0], (*level)[1], (*level)[2], (*level)[3]};
+}
+
+}  // namespace
+
+int main() {
```
dellaert: Is this a test or an example? Unit tests in GTSAM all follow the same pattern. If this is not a test, move it to the examples folder, please.

avinashresearch1:
@dellaert I think most comments are now addressed.

dellaert:
@avinashresearch1 I have updated draft PR #2608 to establish `certifiable` as the normal, solver-independent GTSAM module described in the review comments. It now:

- adds `certifiable` to `gtsam_subdirs`;
- installs public headers under `include/gtsam/certifiable`;
- uses the standard `add_subdirectory(tests)`, `gtsamAddTestsGlob`, and `check.certifiable` machinery;
- builds the generic Burer–Monteiro Riemannian Staircase implementation directly into `gtsam`;
- leaves optional backends such as MOSEK to layer on top.

Please retarget #2606 from `develop` to the `codex/recover-pr-2542` branch (PR #2608). I compared the current heads: the expected overlap is `gtsam/certifiable/CMakeLists.txt`. When resolving it, keep #2608's normal module scaffold and integrate only the MOSEK-specific sources/tests conditionally—do not restore the separate `gtsam_certifiable` library, special top-level `add_subdirectory(gtsam/certifiable)`, or global backend source/library injection.

Once #2608 merges, please retarget #2606 back to `develop` before the base branch is deleted, so we do not repeat the accidental closure that happened to #2542.

dellaert (approved):
Awesome!

PR #2609, opened 2026-07-28
Lifting Rot3, Pose2 and Pose3 to QCQP (for D = 1) and monolithic/chordal SDP including recovery and examples.

avinashresearch1:
## Summary

This PR extends the homogenized D=1 QCQP and lifted-SDP pipeline across SO(2), SO(3), SE(2), and SE(3).

## Main changes

- Adds D=1 QCQP representations for all four Lie groups:
  - Rot2: 5-dimensional homogeneous vector.
  - Rot3: 10-dimensional homogeneous vector.
  - Pose2: 7-dimensional homogeneous vector.
  - Pose3: 13-dimensional homogeneous vector.

- Adds the corresponding quadratic manifold constraints \(x^\top A x=b\) to the Lie-group traits.

- Adds `FromQcqpValue<1>()` recovery for converting optimized QCQP vectors back into Rot2, Rot3, Pose2, and Pose3 values.

- Introduces `TruncatedVecRows` to represent the number of matrix rows retained by the D=1 vectorization:
  - Rotation groups retain the complete rotation matrix.
  - Pose groups omit the final homogeneous matrix row.

- Extends `FrobeniusPriorFactor` and `FrobeniusBetweenFactor` QCQP conversion to the new D=1 representations.

Implements the between-factor residual using

$$
A = M^\top \otimes I_K,
\qquad
B = \begin{bmatrix} -A & I \end{bmatrix},
$$

with a small internal Kronecker-product implementation that avoids depending on Eigen’s unsupported module.

Adds the hard-prior lifted constraints

$$
Ax=b
\quad\Longrightarrow\quad
Axx^\top=bx^\top
\quad\Longrightarrow\quad
AX=bx^\top.
$$

 to prevent unconstrained PSD slack in anchored variable blocks.

- Refactors SDP recovery so monolithic and chordal formulations share the same QCQP-vector recovery logic.

- Moves type-specific manifold recovery out of `LiftedSDPProblem` and into the constrained/geometry layer:
  - The SDP layer returns keyed D=1 QCQP vectors.
  - `ExtractQcqpValues<T,1>()` converts them into typed manifold values.

- Preserves the existing Rot2 and Rot3 matrix-form behavior for \(D>1\). Pose2 and Pose3 explicitly report that recovery is only implemented for \(D=1\).

- Updates the MOSEK wrapper interface to expose the new keyed QCQP recovery API.

## Examples and tests

- Adds N=20 monolithic and chordal ring-SLAM examples for both Pose2 and Pose3.
  - examples/Pose2RingSLAM_LiftedSDP_Monolithic.cpp
  - examples/Pose2RingSLAM_LiftedSDP_Chordal.cpp
  - examples/Pose3RingSLAM_LiftedSDP_Monolithic.cpp
  - examples/Pose3RingSLAM_LiftedSDP_Chordal.cpp

- Adds tests for:
  - QCQP manifold constraints.
  - Frobenius prior and between-factor objective lifting.
  - D=1 lifting and recovery for all four groups.
  - Homogeneous scaling invariance.
  - Invalid and unsupported recovery inputs.
  - Monolithic and chordal SDP objective consistency.
  - Rank-one recovery of individual SDP variable blocks.

- Verified:
  - `testQcqpProblem.run` passes.
  - `testLiftedSDPs.run` passes.
  - All Pose2 and Pose3 N=20 examples recover rank-one variable blocks.


dellaert:
#2608 is now merged, so please rebase on develop while carefully resolving the conflicts. Conflicts are probably formatting (which is correct on develop, pease use clang-format !) and comments (some slop was still there).

dellaert:
Please rebase

avinashresearch1:
done, but I need to fix comments.

dellaert:
Ok, please update the PR description? And let me know when it needs to be reviewed.

gtsam/certifiable/LiftedSDPProblem.h
```diff
@@ -127,8 +128,21 @@ class GTSAM_EXPORT LiftedSDPProblem<MonolithicSDP, MosekSDPSolver> {
         // QcqpValue<Rot2, 1> stores vec(R) in column-major order.
         Rraw << xCanonical(1), xCanonical(3), xCanonical(2), xCanonical(4);
         recoveredPoses.push_back(Rot2::ClosestTo(Rraw));
+      } else if constexpr (std::is_same_v<T, Pose3>) {
```
dellaert: It seems inconsistent to “lift” in the QCQP context, but then do recovery here. I think the whole QCQP conversion and recovery should be in constrained, no?

gtsam/slam/FrobeniusFactor.h
```diff
@@ -146,15 +149,30 @@ class FrobeniusPrior : public NoiseModelFactorN<T> {
...
+          "Rot3, Pose2, and Pose3.");
     } else {
       (void)costs;
       if (this->noiseModel_->isConstrained()) {
         InsertQcqpConstraints<T, 1>(this->key(), constraints);
 
-        constexpr int LiftedDim = Dim + 1;
+        constexpr int K =
+            std::is_same_v<T, Pose2> || std::is_same_v<T, Pose3> ? N - 1 : N;
+        constexpr int ActiveDim = K * N;
```
dellaert: Once again introducing a new constexpr. Are we sure ActiveDim is not simply Dim?

avinashresearch1:
Yes, I need to fix a couple of things still. I finished reviewing about half of this PR on the weekend. 

There is still one constraint (on the SDP side) that appears quite hacky. I am trying to work through the implications, should be done in a couple of hours.

gtsam/slam/FrobeniusFactor.h
avinashresearch1: This is because for `Pose2` and `Pose3`, it is not simply `N*N` because we don't want to have the last row. This is different from `Rot2`, `Rot3`.

avinashresearch1: So `  inline constexpr static auto Dim = N * N` is no longer the right pre-homogenization lifted dimension.

avinashresearch1: But I agree `ActiveDim` is not a great name. I'll rename it to `TruncatedVecDim`

gtsam/certifiable/LiftedSDPProblem.h
avinashresearch1: This is recovery from the slice of the SDP matrix to the QCQP vector.

avinashresearch1: so there are two pieces: 

Xii -> x_i -> T

The first should be in the SDP part, but yes I'll move the second to the QCQP parts, and use: `FromQcqpValue` there.

avinashresearch1: Okay, that is addressed in last commit. The SDP extracts the QCQP vector, FromQqqpVector recovers the poses.

dellaert (approved):
LGTM. I will do a follow-up PR.

Issue #1627, opened 2023-09-05
Shonan rotation averaging converging but relative rotation error is high!!


anilesec:
Dear @dellaert,

I tried shonan rotation averaging for a toy example.(Let say I take 3 random rotation matrices and form a full connected graph. i.e 3 nodes and 3 edges). 

I input the correct(no noise added) relative rotations between the nodes. I expect that cost should be close to zero and relative rotation errors between the estimated absolute rotations and know relative rotations should be zero? But the relative rotations error is not zero?

So, can shonan rotation averaging provide perfect rotations given correct(without any noise) relative rotations as input?

I have spent too much time to get the zero relative rotation errors but failed. Is adding noise to input relative rotations a necessary condition to get zero errors? Any help here is much appreciated

Thanks in advance!

anilesec:
Also do you have an example where shonan rotation is tried on toy example to achieve zero relative rotation errors? I would like to understand how you compute relative rotation errors? Do you align obtained absolute rotations and known rotations  before computing the error?

anilesec:
I have tried with graphs created synthetically and listed relative rotation error from shonan, below is the table:

![image](https://github.com/borglab/gtsam/assets/23581858/ec02e8c6-b8a7-49d1-b0be-086cafc2436d)

dellaert:
I could try to take a look this weekend if you have a simple python script that shows a failure case

anilesec:
Thank you very much. I have **attached the simple Python script**:
Basically, it generates `NUM_NODES` random rotations(as ground truth rotations) -->  and creates a fully connected graph with relative rotations computed from generated random rotations to Shonan method  --> and runs Shonan method --> estimates relative rotation from obtained absolute rotations ---> and computes error with relative rotations obtained from generated rotations --> prints Error

```
import numpy as np
import gtsam
from gtsam import BinaryMeasurementRot3, Rot3, ShonanAveraging3, BinaryMeasurementsRot3
from scipy.stats import special_ortho_group

def mat2rotvec(rotmat, in_degrees=True, axis='xyz'):
    from scipy.spatial.transform import Rotation as ROT
    r = ROT.from_matrix(rotmat)
    if in_degrees:
        return r.as_euler(axis, in_degrees)
    else:
        return r.as_rotvec()

def compute_angular_error_batch(rotation1, rotation2):
    R_rel = np.einsum("Bij,Bjk ->Bik", rotation1.transpose(0, 2, 1), rotation2)
[...]

PR #2629, opened 2026-08-06
Fix Shonan certificate eigenvalue tolerance

ProfFan:
## Summary

- Tighten the sparse Lanczos nonnegativity tolerance from 10e-4 (1e-3) to the documented 1e-4.
- Add deterministic regressions for randomized cases 26 and 54.
- Compare each reported staircase certificate against a dense self-adjoint eigensolve through ranks 3-8.
- Use a specified polar transform so the randomized fixtures are reproducible across standard-library implementations.
- Replace the Eigen std::rand-based Lanczos perturbation with the fixed-seed Spectra generator, making certificate evaluation deterministic and thread-safe across platforms.

## Motivation

Follow-up investigation related to #1627 found two false optimality certificates with the previous tolerance. Both occurred at rank 7:

- Case 26: reported minimum eigenvalue +5.44e-5; reference minimum -2.61e-3.
- Case 54: reported minimum eigenvalue +4.94e-4; reference minimum -4.80e-3.

The implementation comment documents a 10^-4 nonnegativity tolerance, but the prior 10e-4 literal evaluates to 10^-3. That error allowance was large enough for Lanczos to stop on inaccurate eigenvalue estimates and accept both indefinite certificate matrices.

The initial regression used std::srand to control the Lanczos perturbation. Eigen random vectors use the platform C RNG, so Windows followed a different staircase path and still false-certified case 26. Using the deterministic generator bundled with Spectra removes that platform dependence while retaining the small perturbation needed to escape the certificate null space.

With the corrected 1e-4 tolerance and deterministic initialization, neither case is falsely certified through rank 8.

## Validation

- Negative control: with 10e-4, RandomizedCase26 and RandomizedCase54 both fail.
- With 1e-4, testShonanAveraging.run reports no test failures.
- The aggregate check_sfm_program target used by Windows passes locally.
- git diff --check passes.

Related to #1627.


dellaert:
Nice !!! But CI fails :-)

ProfFan:
@dellaert CI is passing

dellaert (approved):
Nice!!

dellaert:
Hold off on merge? Want to first merge #2591

PR #2630, opened 2026-08-06
Avoid redundant system METIS linking

ilkecan:
## Summary

Link system METIS exactly once, using the result from the successful discovery path.

When a METIS CMake config package is found, GTSAM continues to link its `metis` target. When the fallback `find_library()` path is used, GTSAM now links only the absolute path stored in `METIS_LIBRARY`.

Previously, the fallback added both the absolute library path and a bare `metis` item. This produced link lines containing both `/path/to/libmetis.so` and `-lmetis`. The latter required the system library directory to be discoverable independently, defeating the purpose of the absolute path returned by `find_library()`.

## Verification

Verified with nixpkgs METIS 5.2.1 through the fallback discovery path:

- Configured without temporary executable or shared library linker search flags.
- Confirmed that generated link rules contain the absolute `libmetis.so` path and no bare `-lmetis`.
- Built and ran `testOrdering.run` successfully.
- Built and linked `gtsam_unstable` successfully.

### DIFF
--- a/cmake/HandleMetis.cmake
+++ b/cmake/HandleMetis.cmake
@@ -13,11 +13,14 @@ option(GTSAM_USE_SYSTEM_METIS "Find and use system-installed libmetis. If 'off',
 if(GTSAM_USE_SYSTEM_METIS)
   # Debian package: libmetis-dev
   find_package(METIS CONFIG NAMES metis)
-  if(NOT METIS_FOUND)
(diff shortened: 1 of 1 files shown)


dellaert:
@ilkecan Thanks ! Could you combine the two PRs? They seem very related.

ilkecan:
Sure.

PR #2631, opened 2026-08-06
METIS: Avoid redundant system linking & use public API for separator partitioning

ilkecan:
### Summary

Link system METIS exactly once, using the result from the successful discovery path.

When a METIS CMake config package is found, GTSAM continues to link its `metis` target. When the fallback `find_library()` path is used, GTSAM now links only the absolute path stored in `METIS_LIBRARY`.

Previously, the fallback added both the absolute library path and a bare `metis` item. This produced link lines containing both `/path/to/libmetis.so` and `-lmetis`. The latter required the system library directory to be discoverable independently, defeating the purpose of the absolute path returned by `find_library()`.

### Verification

Verified with nixpkgs METIS 5.2.1 through the fallback discovery path:

- Configured without temporary executable or shared library linker search flags.
- Confirmed that generated link rules contain the absolute `libmetis.so` path and no bare `-lmetis`.
- Built and ran `testOrdering.run` successfully.
- Built and linked `gtsam_unstable` successfully.

### Summary

Replace the separator implementation's use of private METIS internals with the public `METIS_PartGraphRecursive` API.

This removes a private ABI dependency that causes memory corruption when GTSAM is compiled with its bundled METIS 5.1.0 headers but linked against system METIS 5.2.1.

### Background

System METIS support was introduced in [#570](https://github.com/borglab/gtsam/pull/570) as part of the broader system-library work tracked by [#292](https://github.com/borglab/gtsam/issues/292). [#1117](https://github.com/borglab/gtsam/pull/1117) later enabled the unstable separator with system METIS by compiling against GTSAM's bundled private `metislib.h` headers while linking against the system library.

The description of #1117 explicitly identified a possible desynchronization between the bundled private headers and the version used to compile the system library. That risk was accepted because METIS was considered old and stable, with updating GTSAM's bundled copy proposed as the remedy if it occurred.

That risk has now materialized. METIS development resumed and version [5.2.1](https://github.com/KarypisLab/METIS/tree/v5.2.1) was tagged in December 2022. Code compiled from the bundled METIS 5.1.0 private structure declarations is not ABI-compatible with objects created by METIS 5.2.1, resulting in memory corruption and a `testFindSeparator` segmentation fault.

Updating the bundled private headers would only move the problem. It would make the code compatible with one particular system version while retaining the same unsupported private ABI dependency.

### Changes

- Remove the private `metislib.h` include and `modefied_EdgeComputeSeparator`.
- Use the public `METIS_PartGraphRecursive` API with two partitions.
- Select the public controls corresponding to the old behavior where possible:
  - `METIS_IPTYPE_GROW`
  - one cut
  - `METIS_OPTION_UFACTOR = 200`
- Continue passing `nullptr` for edge weights because the previous private implementation also ignored `adjwgt`.
- Check the METIS return code and throw an actionable exception on failure.
- Remove the bundled `libmetis` and `GKlib` directories from GTSAM consumer include paths.
- Retain only the public `metis.h` include path and the existing library-selection behavior.
- This is an internal implementation change and introduces no new public API or build option.

### Partition compatibility

The public recursive API does not expose the private `CoarsenTo = 100` setting used by the old implementation. The remaining public controls have been selected to approximate the previous behavior.

Existing test expectations remain unchanged and pass with both tested METIS versions. Larger graphs may nevertheless receive different, equally valid partitions because the old private coarsening threshold cannot be requested through the public API.

### Verification

Testing was performed from GTSAM commit bc82a2f8a2364a494d5a59ba69d822108eb32406.

The METIS versions were:

- Bundled METIS: 5.1.0
- nixpkgs system METIS: 5.2.1

Before this change:

| Configuration | Result |
| --- | --- |
| Bundled METIS 5.1.0 | 355/355 tests passed |
| System METIS 5.2.1 | 354/355 tests passed |

With system METIS 5.2.1, `testFindSeparator.run` reached the third reduced-camera graph and then exited with a segmentation fault before printing the bisection result. CTest reported `testFindSeparator` as the only failure.

After this change:

| Configuration | Result |
| --- | --- |
| Bundled METIS 5.1.0 | 355/355 tests passed |
| System METIS 5.2.1 | 355/355 tests passed |

Additional checks:

- `testFindSeparator.run` passes with bundled and system METIS.
- `testOrdering.run` passes with bundled and system METIS, covering core `Ordering::Metis`.
- The focused system-METIS `testFindSeparator.run` passes under AddressSanitizer.
- No non-vendored source includes `metislib.h`.
- GTSAM consumer compile rules do not expose the bundled `libmetis` or `GKlib` directories.
- Generated consumer configurations do not export those private directories.

### Deliberately out of scope

- Edge weights remain ignored. This is pre-existing behavior: the private implementation passed `nullptr` to METIS instead of `adjwgt`.

### DIFF
--- a/cmake/HandleMetis.cmake
+++ b/cmake/HandleMetis.cmake
@@ -13,11 +13,14 @@ option(GTSAM_USE_SYSTEM_METIS "Find and use system-installed libmetis. If 'off',
 if(GTSAM_USE_SYSTEM_METIS)
   # Debian package: libmetis-dev
   find_package(METIS CONFIG NAMES metis)
-  if(NOT METIS_FOUND)
+  if(METIS_FOUND)
+    set(GTSAM_METIS_LINK_TARGET metis)
+  else()
     find_path(METIS_INCLUDE_DIR metis.h REQUIRED)
     find_library(METIS_LIBRARY metis REQUIRED)
     if(METIS_INCLUDE_DIR AND METIS_LIBRARY)
       set(METIS_FOUND ON)
+      set(GTSAM_METIS_LINK_TARGET ${METIS_LIBRARY})
     endif()
   endif()
 
@@ -31,7 +34,7 @@ if(GTSAM_USE_SYSTEM_METIS)
       $<BUILD_INTERFACE:${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis/libmetis>
       $<BUILD_INTERFACE:${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis/GKlib>
     )
-    target_link_libraries(metis-gtsam-if INTERFACE ${METIS_LIBRARY} metis)
+    target_link_libraries(metis-gtsam-if INTERFACE ${GTSAM_METIS_LINK_TARGET})
   endif()
 else()
   # Bundled version:
--- a/cmake/HandleMetis.cmake
+++ b/cmake/HandleMetis.cmake
@@ -28,12 +28,7 @@ if(GTSAM_USE_SYSTEM_METIS)
     mark_as_advanced(METIS_INCLUDE_DIR)
     mark_as_advanced(METIS_LIBRARY)
     add_library(metis-gtsam-if INTERFACE)
-    target_include_directories(metis-gtsam-if BEFORE INTERFACE ${METIS_INCLUDE_DIR}
-      # gtsam_unstable/partition/FindSeparator-inl.h uses internal metislib.h API
-      # via extern "C"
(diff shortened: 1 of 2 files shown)


ProfFan (commented):
METIS is 5.1.0 for almost all mainstream Linux distributions though. Does this work with 5.1 as well?

ilkecan:
What do you mean? The test suite passes with both 5.1.0 (gtsam bundled) and 5.2.1 (on my system) after these changes.

ProfFan:
Cool, thanks!

ProfFan (approved):

PR #2632, opened 2026-08-06
Allow using system-installed Spectra

ilkecan:
## Summary

Add an opt-in `GTSAM_USE_SYSTEM_SPECTRA` CMake option while preserving the bundled Spectra default.

Also use standard `Spectra/...` include paths so the same source works with both GTSAM's vendored copy and normal Spectra installations.

This advances the system-library work tracked by [#292](https://github.com/borglab/gtsam/issues/292) and follows the existing `GTSAM_USE_SYSTEM_EIGEN` and `GTSAM_USE_SYSTEM_METIS` patterns.

A separate documentation only commit also refreshes the stale bundled Eigen and Spectra versions listed in `LICENSE`, noticed while working on the PR.

## Background

GTSAM currently adds `gtsam/3rdparty/Spectra` itself to the include path and includes headers such as `<SymEigsSolver.h>` and `<MatOp/SparseSymMatProd.h>`. Standard Spectra installations place those files below an `include/Spectra` directory, so they cannot satisfy the unprefixed includes.

Adding a system installation's `include/Spectra` directory globally would expose Spectra's internal directory layout and generic names such as `MatOp/...` to every GTSAM source. Using the upstream `Spectra/...` spelling avoids that problem and works unchanged with the bundled copy after moving its include root up to `gtsam/3rdparty`.

Unlike Eigen and METIS, Spectra also had no GTSAM_USE_SYSTEM_* option. GTSAM always added the bundled Spectra headers to the include path, so there was no supported way to build against a system installed Spectra.

## Changes

The changes are split into three commits so the "include normalization" can be reviewed independently:

1. Use prefixed Spectra includes.
   - Change the three include sites to `<Spectra/SymEigsSolver.h>` and `<Spectra/MatOp/SparseSymMatProd.h>`.
   - Change the bundled build include root from `gtsam/3rdparty/Spectra` to `gtsam/3rdparty`.

2. Allow using system Spectra.
   - Add `GTSAM_USE_SYSTEM_SPECTRA`, defaulting to `OFF`.
   - When enabled, require `find_package(Spectra 1.0 CONFIG REQUIRED)` and use the imported `Spectra::Spectra` target.
   - When disabled, continue using the bundled Spectra headers.
   - Report the option in GTSAM's configuration summary.

3. Update bundled Eigen and Spectra versions in LICENSE.
   - The third-party list still named Eigen 3.3.7 and Spectra v0.9.0, while `gtsam/3rdparty` vendors Eigen 3.4.0 and Spectra 1.1.0. Both entries were written when those copies were first added and never updated on the subsequent bumps.

## Version and interface compatibility

GTSAM's vendored Spectra is version 1.1.0. The minimum required system version is set to 1.0 because it provides the APIs used by GTSAM, including the exported Spectra::Spectra CMake target.

The system build was tested with nixpkgs Spectra 1.2.0.

## Verification

Testing was performed from GTSAM commit bc82a2f8a2364a494d5a59ba69d822108eb32406.

| Configuration | Result |
| --- | --- |
| Bundled Spectra baseline on `develop` | 355/355 tests passed |
| Bundled Spectra after this change | 355/355 tests passed |
| nixpkgs system Spectra 1.2.0 after this change | 355/355 tests passed |

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,7 @@ endif()
 include(cmake/HandleCCache.cmake)           # ccache
 include(cmake/HandleCPack.cmake)            # CPack
 include(cmake/HandleEigen.cmake)            # Eigen3
+include(cmake/HandleSpectra.cmake)          # Spectra
 include(cmake/HandleMetis.cmake)            # metis
 include(cmake/HandleCephes.cmake)           # cephes
 include(cmake/HandleMKL.cmake)              # MKL
--- a/LICENSE
+++ b/LICENSE
@@ -13,12 +13,12 @@ ordering library
 - ceres: Google's nonlinear least-squares optimization library
     - Includes only auto-diff/jet code, with minor modifications to includes
     - http://ceres-solver.org/license.html
-- Eigen 3.3.7:  General C++ matrix and linear algebra library
+- Eigen 3.4.0:  General C++ matrix and linear algebra library
     - Licenced under MPL2, provided in gtsam/3rdparty/Eigen/COPYING.README
     - Some code that is 3rd-party to Eigen is BSD and LGPL
 - METIS 5.1.0: Graph partitioning and fill-reducing matrix ordering library
     - Included unmodified in gtsam/3rdparty/metis
     - Licenced under Apache License v 2.0, provided in
       gtsam/3rdparty/metis/LICENSE.txt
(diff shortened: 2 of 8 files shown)


dellaert (approved):
Nice !! will merge when CI passes.

dellaert:
Same here, let's first merge #2591

PR #2633, opened 2026-08-07
Speed up PR CI with nightly compatibility matrices

dellaert:
## Summary

- add compile-only smoke builds on the latest Linux, Windows, and macOS runners before Python CI
- preserve the existing required Python check names and make them report fast-gate failures without running expensive setup
- move Linux, macOS, Windows, vcpkg, and special-case compatibility matrices to a parallel nightly workflow
- create, update, and close a single assigned incident issue for scheduled nightly failure episodes
- make timeSFMBAL benchmarking opt-in through manual dispatch or `/benchmark` (with `/bench` retained as an alias)

## Impact

Code PRs now get the fastest cross-platform compilation signal first, followed by the existing five-lane Python matrix. Documentation-only PRs retain their current exclusions. The broad compatibility matrices run nightly or by manual dispatch, reducing routine runner load while preserving coverage.

Branch-protection contexts and repository settings are unchanged.

## Validation

- built the exact Release/Ninja smoke profile locally on macOS with `--target gtsam --parallel 4`
- parsed all 15 workflow YAML files
- passed actionlint syntax and expression checks for every changed workflow
- verified CI topology, required context names, matrix counts, and nightly issue lifecycle guards
- passed `git diff --check`

The repository-wide default actionlint invocation still reports pre-existing ShellCheck warnings and two obsolete-action warnings in unchanged workflows.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -1,17 +1,11 @@
 name: Linux CI
 
 on:
-  workflow_dispatch: 
-  pull_request:
-    paths-ignore:
-      - '**.md'
-      - '**.ipynb'
-      - 'myst.yml'
+  workflow_call:
+  workflow_dispatch:
 
-# Cancels any in-progress workflow runs for the same PR when a new push is made,
-# allowing the runner to become available more quickly for the latest changes.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  group: linux-ci-${{ github.run_id }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -1,17 +1,11 @@
 name: macOS CI
 
 on:
-  workflow_dispatch: 
-  pull_request:
-    paths-ignore:
-      - '**.md'
-      - '**.ipynb' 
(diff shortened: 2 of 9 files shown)


dellaert:
@ProfFan whaddaya think - this also makes timeSfmBAL opt-in for now

dellaert:
Windows is still pretty slow - don't know what cache would do here @Gold856 ?

Gold856:
MSVC is just slow. Caching would likely explode worst-case build times due to the lack of PCH (as it did before) and decently speed up Windows on average. You might be better off just compiling in Debug mode. That'll probably speed things up more.

A while back, I got a recommendation by word-of-mouth for https://www.blacksmith.sh/, which has much faster runners and an OSS tier you can apply for. I opted to not forward that recommendation here out of fear GitHub may resume [their original plan of charging per minute for use of their control plane](https://github.com/resources/insights/2026-pricing-changes-for-github-actions), which for a project that (ostensibly) can't use paid resources, would have meant Blacksmith would start costing money. Since GitHub hasn't yet resumed their original plans, I'll leave it up to you on if that's an acceptable risk to take. It would significantly help your CI build time issues; experimental testing months back showed a 2.75x speed boost. It's likely nothing else will allow you to go much faster than just a bigger runner for the same amount of effort (C++ modules may be helpful here, but that's likely years out).

PR #2636, opened 2026-08-08
Preserve joint marginal query key order

### LINKED ISSUE #2622: JointMarginal::fullMatrix() ordering
"marginals.jointMarginalCovariance(keys).fullMatrix()" appears to return a matrix whose blocks are ordered in increasing-key-order, NOT in the order of the keys which were provided.  This is perhaps a surprising result, and the documentation is silent about what order to expect.  Suggest to document the expected behaviour.

E.g. to reproduce:
```
Eigen::MatrixXd getFullMatrix( const gtsam::Marginals &marginals, const gtsam::KeyVector &keys )
{
    return marginals.jointMarginalCovariance(keys).fullMatrix();
}
```
If 'keys' is {2,1}, I expected this to return a matrix corresponding to keys {2,1}, but instead it returns a matrix corresponding to keys {1,2}.

This is with 4.3a2.

dellaert:
## Summary

- preserve first-occurrence key order in `BayesTree::jointBayesNet(KeyVector)`
- have joint information and covariance queries consume the ordered Bayes net directly
- document the ordering contract across `Marginals`, `GaussianBayesTree`, `ISAM2`, `JointMarginal`, and `BayesTree`
- document every helper in `GaussianBayesTreeQueries.h` and attribute the file to Frank Dellaert
- add regression coverage for unsorted and duplicate key queries

## Root cause

Both the joint-marginal query helper and `jointBayesNet(KeyVector)` sorted their input keys before constructing the result. Consequently, a query such as `{2, 1}` returned matrix blocks in `{1, 2}` order.

`jointBayesNet` now stably deduplicates its input, so `{7, 2, 1, 2}` produces elimination order `{7, 2, 1}`. The existing ordered marginalization path then creates the Bayes net in that order, and the Gaussian query helpers use its ordering directly. No dense post-processing permutation is required.

## Validation

- `make -j6 testGaussianBayesTreeB.run`
- `make -j6 testMarginals.run`
- `make -j6 testGaussianISAM2.run`
- `make -j6 testSymbolicBayesTree.run`
- `make -j6 testDiscreteBayesTree.run`
- `git diff --check origin/develop...HEAD`
- changed-lines `git clang-format` check

Fixes #2622

### DIFF
--- a/gtsam/inference/BayesTree-inst.h
+++ b/gtsam/inference/BayesTree-inst.h
@@ -433,11 +433,17 @@ namespace gtsam {
   }
 
   /* ************************************************************************* */
+  /** Return unique keys in order of first occurrence. */
   template <class CLIQUE>
   static KeyVector uniqueKeys(const KeyVector& keys) {
-    KeyVector unique = keys;
-    std::sort(unique.begin(), unique.end());
-    unique.erase(std::unique(unique.begin(), unique.end()), unique.end());
+    KeyVector unique;
+    unique.reserve(keys.size());
+    KeySet seen;
+    for (Key key : keys) {
+      if (seen.insert(key).second) {
+        unique.push_back(key);
+      }
+    }
     return unique;
   }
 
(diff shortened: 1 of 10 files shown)


dellaert:
An independent codex "de novo" review found no actionable issues. Will merge.

PR #2646, opened 2026-08-09
Fix GenericValue RTTI across shared libraries

### LINKED ISSUE #254: ValuesIncorrectType is thrown even when type_info names for value are same
## Description

As originally tracked in https://github.com/CogRob/omnimapper/issues/26 , it seems that we're encountering segfaults when a `ValuesIncorrectType` error is thrown even when `name` valudes for `type_info`s are the same.

```
[omnimapper_ros_node-1] BoundedPlanePlugin: Creating new plane 8070450532247928832: 0.769868 -0.097762 -0.630671 0.699390
[omnimapper_ros_node-1] Adding new symbol: p0
[omnimapper_ros_node-1] terminate called after throwing an instance of 'gtsam::ValuesIncorrectType'
[omnimapper_ros_node-1]   what():  Attempting to retrieve value with key "p0", type stored in Values is N10omnimapper13BoundedPlane3IN3pcl12PointXYZRGBAEEE but requested type was N10omnimapper13BoundedPlane3IN3pcl12PointXYZRGBAEEE
```

## Steps to reproduce

See run instructions from docker demo described in the project readme:
https://github.com/CogRob/omnimapper_ros/blob/plugin/planes/README.md

A minimal ros2bag that can induce the segfault can be found here:
https://drive.google.com/file/d/1cVxsfCp2LSuRGkaUUA-bOC2fhhts1Cyo

```
$ ros2 bag play ./rosbag2_2020_...
$ docker run -it --rm \
  --net=host --ipc=host --pid=host" \
  cogrob/omnimapper_ros \
  ros2 launch omnimapper_ros realsense.launch.xml \
    use_rosbag:=true
```

Stacktrace:

<details>

```
__GI_raise 0x00007fb4c71eae97
__GI_abort 0x00007fb4c71ec801
<unknown> 0x00007fb4c7841957
<unknown> 0x00007fb4c7847ae6
std::terminate() 0x00007fb4c7847b21
std::rethrow_exception(std::__exception_ptr::exception_ptr) 0x00007fb4c7847ada
<unknown> 0x00007fb4c7b5fd0c
<unknown> 0x00007fb4c7b64e9b
<unknown> 0x00007fb4c7b61790
gtsam::NonlinearFactorGraph::linearize(gtsam::Values const&) const 0x00007fb4c84c1f8c
gtsam::ISAM2::update(gtsam::NonlinearFactorGraph const&, gtsam::Values const&, gtsam::ISAM2UpdateParams const&) 0x00007fb4c848dbc4
gtsam::ISAM2::update(gtsam::NonlinearFactorGraph const&, gtsam::Values const&, std::vector<unsigned long, tbb::tbb_allocator<unsigned long> > const&, boost::optional<gtsam::FastMap<unsigned long, int> > const&, boost::optional<gtsam::FastList<unsigned long> > const&, boost::optional<gtsam::FastList<unsigned long> > const&, bool) 0x00007fb4c848a67c
omnimapper::OmniMapperBase::optimize omnimapper_base.cpp:467
omnimapper::OmniMapperBase::spinOnce omnimapper_base.cpp:657
omnimapper::OmniMapperBase::spin omnimapper_base.cpp:640
boost::_mfi::mf0<void, omnimapper::OmniMapperBase>::operator() mem_fn_template.hpp:49
boost::_bi::list1<boost::_bi::value<omnimapper::OmniMapperBase*> >::operator()<boost::_mfi::mf0<void, omnimapper::OmniMapperBase>, boost::_bi::list0> bind.hpp:259
boost::_bi::bind_t<void, boost::_mfi::mf0<void, omnimapper::OmniMapperBase>, boost::_bi::list1<boost::_bi::value<omnimapper::OmniMapperBase*> > >::operator() bind.hpp:1294
boost::detail::thread_data<boost::_bi::bind_t<void, boost::_mfi::mf0<void, omnimapper::OmniMapperBase>, boost::_bi::list1<boost::_bi::value<omnimapper::OmniMapperBase*> > > >::run thread.hpp:116
<unknown> 0x00007fb4c7d8bbcd
start_thread 0x00007fb4c99806db
clone 0x00007fb4c72cd88f
```

</details>

Relevant code path:

https://github.com/borglab/gtsam/blob/f538d1dc7bdd7126cb683f3e961c985f76a872b0/gtsam/nonlinear/Values.cpp#L160-L163

https://github.com/borglab/gtsam/blob/16dbf27375fdefb83ce2355beb0c147fa9c07600/gtsam/nonlinear/Values.cpp#L233-L240

## Expected behavior

When calling update for new factors with the same type, a `ValuesIncorrectType` error is not expected.

``` c++
gtsam::ISAM2Result result = isam2.update(new_factors, new_values);
```

## Environment

OS: Ubuntu 18.04
Lang: C++ 14

Exact checkout of gtsam:
https://github.com/CogRob/omnimapper_ros/blob/a17c4017dfde4cb022384e8d8f521e2ab05197fb/install/underlay/underlay.repos#L2-L5

Exact build setup
https://github.com/CogRob/omnimapper_ros/blob/a17c4017dfde4cb022384e8d8f521e2ab05197fb/Dockerfile#L55-L66

## Additional information

I'd like to get a better stack trace, but there may be a separate issue in using TBB downstream when compiling from gtsam debug builds. Perhaps gtsam skips finding TBB when debug flags are set?
See: https://github.com/CogRob/omnimapper/issues/25

dellaert:
## Summary

- give `GenericValue<T>` RTTI default visibility so the C++ ABI can identify matching specializations across shared-library boundaries
- document the visibility requirement for user-defined value types crossing DSOs
- add a hidden-by-default helper library regression covering RTTI equality, `Values::at<T>`, `Values::update<T>`, and genuine type mismatch rejection

## Root cause

A `GenericValue<T>` instantiated in one shared library could have a different RTTI identity from the same specialization instantiated in another library. The resulting `type_info::name()` strings were identical, but ABI type equality and `dynamic_cast` failed, causing `ValuesIncorrectType`.

The fix preserves ABI type checks and does not treat matching RTTI names as proof of type identity.

Fixes #254.

## Validation

- C++17 Release shared build with Clang 22.1.7/libc++ and hidden inline visibility
- `make -j6 testValuesCrossLibrary.run`
- `make -j6 testValues.run`
- `ctest -R '^testValuesCrossLibrary$' --output-on-failure`
- `git diff --check`

The new regression was also verified to fail before the visibility fix with the equal-name `ValuesIncorrectType` reported in #254.

### DIFF
--- a/gtsam/base/GenericValue.h
+++ b/gtsam/base/GenericValue.h
@@ -26,8 +26,12 @@
 #include <iostream>
 #include <typeinfo> // operator typeid
 
-#ifdef _WIN32
+// GenericValue's RTTI must remain visible so the C++ ABI can identify a
+// specialization consistently when a Value crosses a shared-library boundary.
+#if defined(_WIN32)
 #define GENERICVALUE_VISIBILITY
+#elif defined(__GNUC__)
+#define GENERICVALUE_VISIBILITY __attribute__((visibility("default")))
 #else
 // This will trigger a LNKxxxx on MSVC, so disable for MSVC build
 // Please refer to https://github.com/borglab/gtsam/blob/develop/Using-GTSAM-EXPORT.md
@@ -37,10 +41,13 @@
 namespace gtsam {
 
 /**
- * Wraps any type T so it can play as a Value
+ * Wraps any type T so it can play as a Value.
+ *
+ * A user-defined T that crosses a shared-library boundary must also have
+ * default visibility so the C++ ABI recognizes GenericValue<T> as one type.
  */
 template<class T>
-class GenericValue: public Value {
+class GENERICVALUE_VISIBILITY GenericValue: public Value {
 
 public:
 
--- a/gtsam/base/GenericValue.h
+++ b/gtsam/base/GenericValue.h
@@ -29,12 +29,10 @@
 // GenericValue's RTTI must remain visible so the C++ ABI can identify a
 // specialization consistently when a Value crosses a shared-library boundary.
 #if defined(_WIN32)
-#define GENERICVALUE_VISIBILITY
-#elif defined(__GNUC__)
-#define GENERICVALUE_VISIBILITY __attribute__((visibility("default")))
-#else
-// This will trigger a LNKxxxx on MSVC, so disable for MSVC build
+// Exporting this header-only template triggers linker errors on MSVC.
 // Please refer to https://github.com/borglab/gtsam/blob/develop/Using-GTSAM-EXPORT.md
+#define GENERICVALUE_VISIBILITY
+#else
 #define GENERICVALUE_VISIBILITY GTSAM_EXPORT
 #endif
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -19,7 +19,7 @@ endif()
 gtsamAddTestsGlob(tests "test*.cpp" "${excluded_tests}" "gtsam")
 
 # Exercise Values RTTI across a hidden-by-default shared-library boundary.
-if(GTSAM_SHARED_LIB)
+if(GTSAM_BUILD_TESTS AND GTSAM_SHARED_LIB)
   add_library(testValuesCrossLibraryHelper SHARED
               values_cross_library/ValuesCrossLibrary.cpp)
   target_compile_definitions(testValuesCrossLibraryHelper
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -18,6 +18,43 @@ endif()
 # Build tests
 gtsamAddTestsGlob(tests "test*.cpp" "${excluded_tests}" "gtsam")
 
+# Exercise Values RTTI across a hidden-by-default shared-library boundary.
+if(GTSAM_SHARED_LIB)
+  add_library(testValuesCrossLibraryHelper SHARED
+              values_cross_library/ValuesCrossLibrary.cpp)
+  target_compile_definitions(testValuesCrossLibraryHelper
+                             PRIVATE VALUES_CROSS_LIBRARY_EXPORTS)
+  target_include_directories(testValuesCrossLibraryHelper
+                             PUBLIC values_cross_library)
+  target_link_libraries(testValuesCrossLibraryHelper PRIVATE gtsam)
+  set_target_properties(testValuesCrossLibraryHelper PROPERTIES
+                        CXX_VISIBILITY_PRESET hidden
+                        VISIBILITY_INLINES_HIDDEN YES
+                        EXCLUDE_FROM_ALL ON)
+  gtsam_apply_build_flags(testValuesCrossLibraryHelper)
+
+  add_executable(testValuesCrossLibrary
+                 values_cross_library/testValuesCrossLibrary.cpp)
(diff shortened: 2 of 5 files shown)


dellaert:
@ProfFan I am cleaning up issues and ask Codex to try it and fix this. Might be a dud because the CI immediately fails.

dellaert:
@ProfFan CI issue was easily fixed. What do you think about Codex' solution?

ProfFan:
Wait a sec I thought this was fixed years ago?

ProfFan:
I think this issue should be directly closed, no fix needed

dellaert:
@ProfFan You remembered correctly that this visibility problem was addressed in 2020, but the current 4.3/develop line has regressed.

History:
- `f48f5f0dc` added `GTSAM_EXPORT` to `GenericValue`.
- `a18a74778` / develop PR #259 changed that to `GENERICVALUE_VISIBILITY` and still applied it to the class.
- The 4.3 ancestry (blame points to backport `d0724a77b`) retains the macro and its explanatory comment, but `GenericValue` itself is again declared without the macro. That is the state at this PR's base, `1dc3a189e`.

I reproduced from a pristine archive of `1dc3a189e`, applying only the regression test (not the header fix), with C++17, shared GTSAM, and a hidden-by-default helper DSO plus executable. The helper constructs a conforming, traits-based `GenericValue<TestValue>`; the executable inserts it into `Values` and updates the same key with the same `TestValue`.

The unpatched output is:

```
foreign name: N5gtsam12GenericValueIN20values_cross_library9TestValueEEE
local name:   N5gtsam12GenericValueIN20values_cross_library9TestValueEEE
names equal: 1
typeids equal: 0
Exception: WARNING: Detected types with same name but different `typeid`.
```

The exception is thrown by `Values::update` before retrieval. With this PR, the same focused target completes with “There were no test failures.”

The original OmniMapper reproducer in #254 was indeed separately diagnosed as directly inheriting `Value`, so that particular application was non-conforming. But the titled same-name/different-`typeid` failure is demonstrably present today for a conforming traits-based value across a DSO boundary. This PR restores the previously intended visibility and now tests the user-visible `Values::update` failure directly.

dellaert:
@ProfFan see above - that was answer by Sol (High). WDYT?

ProfFan:
Ah I guess the answer is YES, somehow that change is lost in merge? btw:

```
#define GENERICVALUE_VISIBILITY
#elif defined(__GNUC__)
#define GENERICVALUE_VISIBILITY __attribute__((visibility("default")))
#else
// This will trigger a LNKxxxx on MSVC, so disable for MSVC build
// Please refer to https://github.com/borglab/gtsam/blob/develop/Using-GTSAM-EXPORT.md
#define GENERICVALUE_VISIBILITY GTSAM_EXPORT
```

I think this special case (`elif defined(__GNUC__)`) is not needed, since GTSAM_EXPORT should handle this already.

dellaert:
Good point—updated the non-Windows path to use `GTSAM_EXPORT` and kept only the required MSVC exception. The focused cross-library regression still passes. Thanks! (`a34a23c6f`)

dellaert:
Let's continue discussion or approve if this is now ready to merge. Please do look over the solution, because I am (blindly) trusting codex here.

ProfFan (approved):
LGTM

PR #2652, opened 2026-08-10
Automatically linearize fixed-size binary and ternary factors

dellaert:
## Summary

This PR makes fixed-size linearization a general rule in GTSAM:

> A two- or three-key `NoiseModelFactorT` whose error and argument dimensions are all known at compile time linearizes to `BinaryJacobianFactor` or `TernaryJacobianFactor`. Any dynamic dimension retains the generic `JacobianFactor` path.

It also gives those fixed-size linear factors specialized downstream kernels for Hessian diagonals, delta error, whole/ranged Hessian assembly, and the scalar-block case.

This changes a central linearization policy and audits a broad set of core and unstable factors. The implementation is mathematically conservative, but the scope and ABI implications deserve careful review.

## Why

Previously, `NoiseModelFactorN` erased the error dimension by fixing the output type to dynamic `Vector`. Even factors with completely static residual and variable dimensions therefore produced generic dynamic `JacobianFactor` objects unless each factor supplied a handwritten `linearize()`.

That left two kinds of performance on the table:

- repeated dynamic matrix handling during linearization;
- generic downstream operations even when every block shape was known at compile time.

It also made the optimization opt-in and inconsistent across factor families.

### Automatic selection

`NoiseModelFactorT<OutputVec, ValueTypes...>` carries the residual type explicitly. Compile-time dimension traits select:

- `BinaryJacobianFactor<M, N1, N2>` for fixed two-key factors;
- `TernaryJacobianFactor<M, N1, N2, N3>` for fixed three-key factors;
- qualified generic `NoiseModelFactor::linearize()` for all other arities or any dynamic dimension.

The selection and construction machinery lives in the public `gtsam/linear/FixedJacobianFactor.h` header, with its implementation types kept in `gtsam::internal`. Linearization itself is variadic over the argument-dimension pack: binary and ternary factors share one implementation for evaluation, dimension validation, vector-based joint whitening, and constrained-model handling. Only two small internal factory specializations adapt final construction to the existing `BinaryJacobianFactor` and `TernaryJacobianFactor` APIs. This keeps the nonlinear design ready for a future variadic fixed Jacobian factor without introducing that larger refactor here.

The fixed paths preserve the generic algorithm:

1. honor inactive factors;
2. call `unwhitenedError()` once to obtain the residual and every Jacobian;
3. perform the same dimension validation;
4. whiten all Jacobians and the RHS together, including robust weighting;
5. preserve the constrained unit model exactly;
6. construct only the final concrete linear factor differently.

Null, Unit, Isotropic, Diagonal, full Gaussian, Robust, and Constrained models are covered.

### Fixed linear-factor kernels

`BinaryJacobianFactor` and the new `TernaryJacobianFactor` specialize:

- `hessianDiagonalAdd()`;
- `deltaError()`;
- whole `updateHessian()`;
- ranged `updateHessian(beginCol, endCol)`.

Ranged assembly follows physical upper-triangular ownership, including reversed key/slot ordering. Non-unit/constrained behavior falls back to the existing generic implementation where required.

For one-column blocks, `BinaryJacobianFactor` uses the explicit scalar outer product instead of Eigen's substantially more expensive `rankUpdate`.

### Factor audit

The audit migrates fixed-output binary and ternary factors across SAM, SLAM, SFM, navigation, dynamics, and `gtsam_unstable`. This includes Between, mono/stereo projection, bearing/range/bearing-range, variable-calibration SFM, `ImuFactor2`, carrier/pseudorange/Doppler factors, planar gyro, reference-frame factors, rolling-shutter projection, and the fixed-size dynamics factors.

Factors with a genuinely runtime-sized output or argument remain on `NoiseModelFactorN` and generic linearization.

`GeneralSFMFactor` deliberately retains its direct fixed-size `project2()` linearization, which avoids the dynamic intermediate Jacobians used by the shared nonlinear path. Its custom path now uses joint `WhitenSystem(A1, A2, b)` so robust weighting and constrained-model behavior match generic linearization; the noise-model equivalence test includes a Robust model.

## Correctness

The tests compare the specialized result against an explicitly qualified generic linearization, including keys, every whitened matrix block, RHS, error, concrete runtime type, robust weighting, constrained models, inactive factors, malformed dimensions, missing values, and cheirality behavior.

Notably, the `ImuFactor2` test and timing harness require all three Jacobians and the RHS to be **bitwise identical** to generic linearization before timing either path.

Whole and per-column ranged Hessian assembly are compared against generic assembly for normal and reversed slot orderings. Scalar blocks and accumulation into pre-populated `VectorValues` are also covered.

## Performance

All measurements below are from a Release build on arm64 macOS. Comparative harnesses rotate/alternate trial order and verify equal outputs. Times are medians; percentages are lower-is-better.

### Targeted binary kernels: Pose3 BetweenFactor with Unit noise

Nine alternating trials:

| operation | generic | fixed binary | change |
|---|---:|---:|---:|
| linearize | 0.437 us | 0.361 us | -17.4% |
| deltaError | 0.0943 us | 0.0159 us | -83.1% |
| hessianDiagonalAdd | 0.0840 us | 0.0693 us | -17.5% |
| whole updateHessian | 0.371 us | 0.277 us | -25.4% |
| ranged updateHessian | 0.429 us | 0.274 us | -36.0% |
| projection-factor linearize | 0.288 us | 0.229 us | -20.6% |

For the BAL-shaped `BinaryJacobianFactor<2,9,3>`, alternating fixed-shape trials measured `deltaError()` at 6.07x faster and `hessianDiagonalAdd()` at 1.34x faster than generic.

The isolated one-column Hessian update benchmark measured explicit outer products versus Eigen `rankUpdate`:

| residual rows M | outer-product speedup |
|---:|---:|
| 1 | 3.36x |
| 2 | 3.76x |
| 3 | 4.43x |
| 9 | 21.65x |

### W10000 Pose2 graph

The graph has 10,000 poses and 64,311 between factors. Generic and binary runs use the same graph, values, COLAMD ordering, and fixed LM iteration count, and produce identical final errors and poses.

A warmed linearization-focused run (`warmup=2`, 9 trials, 10 linearizations/trial) measured:

| | generic | fixed binary | paired median change |
|---|---:|---:|---:|
| linearize | 7.732 ms | 6.106 ms | -22.5% |
| one LM iteration | 69.980 ms | 56.601 ms | -18.6% |

The full ten-iteration run measured 1268.877 ms generic versus 869.937 ms binary, a **-30.9% paired median** change, with zero pose difference.

Linearization-only timing is allocator/warmup-sensitive: the full run with only one warmup produced a noisy +24.0% paired linearization result before the heavier-warmup confirmation above. The end-to-end optimization improvement was repeatable; the PR does not rely on a guaranteed standalone linearization percentage.

### Variable-calibration BAL: ternary SFM

`dubrovnik/problem-16-22106-pre.txt`: 16 cameras, 22,106 points, 83,718 projection factors, five alternating trials and five LM iterations.

| representation | linearize | optimize |
|---|---:|---:|
| generic split Pose3/Point3/Cal3Bundler | 18.707 ms | 1109.337 ms |
| ternary split | 9.244 ms | 969.314 ms |
| existing packed-camera binary reference | 5.578 ms | 704.097 ms |

Relative to the mathematically identical generic split graph, ternary linearization is **50.6% faster** and five-iteration optimization is **12.6% faster**, with zero local-value difference.

The packed binary row is a useful architectural reference, not an apples-to-apples claim: packing calibration into the camera changes the variable structure.

### Exact ImuFactor2 chain

1,000 IMU intervals / 2,002 variables, seven alternating trials:

| | generic | ternary | change |
|---|---:|---:|---:|
| linearize | 0.771 ms | 0.716 ms | -7.2% |
| one Gauss-Newton step | 7.474 ms | 7.190 ms | -3.8% |

The harness first verifies bitwise-identical keys, Jacobian blocks, RHS, and noise model for every factor. Final values are also identical.

### Plaza2 range regression workload

The robust Plaza2 workload completed 4,090 odometry entries, 1,816 range factors, and 65 incremental updates. Across five measured runs:

- median solve: 0.159341 s;
- median total: 0.159781 s.

This is a regression workload, not a controlled before/after claim.

### Shared fixed-size kernel follow-up

Binary and ternary factors now delegate residual, diagonal, and Hessian block work to reusable fixed-size helpers. Common per-Jacobian Hessian kernels (`AᵀA` plus `Aᵀb`) are instantiated once inside `libgtsam` for 22 recurring `(M,N)` shapes. Cross-pair dimensions are canonicalized while preserving the caller's original Eigen multiplication order. The instantiated symbols are hidden; downstream template users retain the header-defined inline fallback and acquire no library ABI dependency.

A fully out-of-line version of every self/cross product was rejected: isolated kernels regressed by 6–14% because the shared-library call cost outweighed the small matrix products. Grouping one Jacobian's self and RHS work amortized that cost. Eleven alternating microbenchmark trials measured +0.9%, +1.4%, +1.5%, and +0.5% for representative `<3,3,3>`, `<2,9,3>`, `<2,6,3,3>`, and `<9,9,9,6>` kernels—each within the 2% acceptance gate.

Release Mach-O sizes decreased:

| Artifact | Before | After | Change |
|---|---:|---:|---:|
| `libgtsam.4.3a2.dylib` | 9,110,352 B | 9,072,480 B | -37,872 B (-0.416%) |
| Pose-graph timing executable | 213,768 B | 213,192 B | -576 B (-0.269%) |
| IMU timing executable | 301,976 B | 301,528 B | -448 B (-0.148%) |
| BAL timing executable | 479,472 B | 477,968 B | -1,504 B (-0.314%) |

Symbol inspection finds exactly 22 local `updateJacobianHessian` instantiations in `libgtsam`, no exported helper symbols, and no helper references in the timing executables. Uncommon `BinaryJacobianFactor<4,2,4>` and `TernaryJacobianFactor<4,2,4,5>` tests exercise the header-defined fallback.

Regression-control reruns still show the expected overall specialization gains with numerically identical outputs:

| Workload | Linearization | One optimization |
|---|---:|---:|
| W10000 unit-noise Pose2 graph | -18.0% | -14.8% |
| 1000-step ternary IMU | -8.0% | -3.1% |
| Dubrovnik BAL ternary split | -49.4% | -13.5% |

These final runs are reported as regression controls, not as a claimed before/after speedup from this helper-only refactor; allocator and run-to-run noise dominate that smaller comparison.

## Validation

Release validation:

- `make -j6 gtsam gtsam_unstable`
- `testBinaryJacobianFactor.run`
- `testTernaryJacobianFactor.run`
- `testJacobianFactor.run`
- `testBetweenFactor.run`
- `testProjectionFactor.run`
- `testStereoFactor.run`
- `testBearingFactor.run`
- `testRangeFactor.run`
- `testBearingRangeFactor.run`
- `testGeneralSFMFactor.run`
- `testGeneralSFMFactor_Cal3Bundler.run`
- `testImuFactor.run`
- `testConstantVelocityFactor.run`
- `testPlanarGyroFactor.run`
- `testDopplerFactor.run`
- `testSimpleHelicopter.run`
- `testVelocityConstraint.run`
- `testProjectionFactorPPP.run`
- `testSerializationSam.run`
- `testSerializationLinear.run`
- `git diff --check`

All passed.

The checked-in timing executables are:

- `timeBinaryFactorPoseGraph`;
- `timeTernarySfmBAL`;
- `timeTernaryImuFactor`.

## Compatibility and review notes

- The concrete runtime type returned by fixed binary/ternary nonlinear factors changes, but it remains a `JacobianFactor`.
- Runtime-sized factor instantiations retain generic behavior.
- Several factor declarations now expose fixed-size error-vector return types and use a different `NoiseModelFactorT` template base. Normal source usage is compatible, but exported template instantiations deserve ABI review.
- No nonlinear factor storage or serialization fields are added.
- The experimental `BetweenJacobianFactor` structural Hessian specialization is intentionally **not included**. It lives on `codex/between-jacobian-experiment` pending stronger evidence that its extra complexity is worthwhile.

Suggested review order:

1. `FixedJacobianFactor.h` and `NoiseModelFactorN.h` variadic selection, construction, and whitening semantics;
2. `BinaryJacobianFactor.h` and `TernaryJacobianFactor.h` kernels/tests;
3. `ImuFactor2` bitwise-equivalence path;
4. the mostly mechanical fixed-output factor audit;
5. timing harnesses and interpretation.


dellaert:
@ProfFan these are really great speedups across the board!

ProfFan:
Nice! Though I think this should be doable without special (Binary/Ternary) factors, by variadic templates

PR #2653, opened 2026-08-10
Use normalized all-pivot Cholesky checks

### LINKED ISSUE #600: Sequential Cholesky incorrect for large problems
## Description

This was initially noticed by @ProfFan in #111.  Linear solving/elimination using Sequential Cholesky and Multifrontal Cholesky don't give the same results, and it appears actually sequential is the erroneous one.

For the large dataset from BAL "Final" with 394 images [(link)](https://grail.cs.washington.edu/projects/bal/data/final/problem-394-100368-pre.txt.bz2), running [timing/timeSFMBAL](https://github.com/borglab/gtsam/blob/develop/timing/timeSFMBAL.h) with this dataset using Sequential Cholesky gives a different result than Multifrontal versions.

## Steps to reproduce

1. Download [BAL Final 394](https://grail.cs.washington.edu/projects/bal/data/final/problem-394-100368-pre.txt.bz2)
2. Run `timeSFMBAL` with `params.setLinearSolverType("SEQUENTIAL_CHOLESKY");` and `params.setLinearSolverType("MULTIFRONTAL_CHOLESKY");`

MULTIFRONTAL_CHOLESKY:
```
Processing ../examples/Data/BALfinal-394-100368-pre.txt with solver MULTIFRONTAL_CHOLESKY
Initial error: 4.54573e+06, values: 100762
iter      cost      cost_change    lambda  success iter_time
   0  3.395533e+05    4.21e+06    1.00e-04     1    2.87e+01
   1  3.010002e+05    3.86e+04    3.33e-05     1    1.37e+01
   2  2.977927e+05    3.21e+03    1.11e-05     1    1.22e+01
...
```

SEQUENTIAL_CHOLESKY:
```
Processing ../examples/Data/BALfinal-394-100368-pre.txt with solver SEQUENTIAL_CHOLESKY
Initial error: 4.54573e+06, values: 100762
iter      cost      cost_change    lambda  success iter_time
   0      inf        2.12e-314    1.00e-04     0    1.96e+00
iter      cost      cost_change    lambda  success iter_time
   0      inf        2.12e-314    2.00e-04     0    1.97e+00
iter      cost      cost_change    lambda  success iter_time
   0      inf        2.12e-314    8.00e-04     0    1.97e+00
iter      cost      cost_change    lambda  success iter_time
   0      inf        2.12e-314    6.40e-03     0    1.99e+00
iter      cost      cost_change    lambda  success iter_time
   0  7.062675e+05    3.84e+06    1.02e-01     1    8.21e+00
   1  5.298729e+05    1.76e+05    3.41e-02     1    7.25e+00
   2  4.336619e+05    9.62e+04    1.14e-02     1    7.19e+00
...
```
Unfortunately I couldn't test MULTIFRONTAL_QR due to memory overflow, nor could I test SEQUENTIAL_QR because it just took too long.

## Expected behavior

SEQUENTIAL_CHOLESKY should give the _exact_ same result as MULTIFRONTAL_CHOLESKY.  It does when we use a smaller dataset, eg [Dubrovnik 16](https://grail.cs.washington.edu/projects/bal/data/dubrovnik/problem-16-22106-pre.txt.bz2):
Multifrontal:
```
Processing ../examples/Data/dubrovnik-16-22106-pre.txt with solver MULTIFRONTAL_CHOLESKY
Initial error: 4.18566e+06, values: 22122
iter      cost      cost_change    lambda  success iter_time
   0  1.016823e+05    4.08e+06    1.00e-04     1    5.51e-01
   1  5.290685e+04    4.88e+04    3.33e-05     1    3.84e-01
   2  1.925219e+04    3.37e+04    3.31e-05     1    3.45e-01
```
Sequential:
```
Processing ../examples/Data/dubrovnik-16-22106-pre.txt with solver SEQUENTIAL_CHOLESKY
Initial error: 4.18566e+06, values: 22122
iter      cost      cost_change    lambda  success iter_time
   0  1.016823e+05    4.08e+06    1.00e-04     1    4.68e-01
   1  5.290685e+04    4.88e+04    3.33e-05     1    4.31e-01
   2  1.925219e+04    3.37e+04    3.31e-05     1    4.25e-01
```

## Environment

Ubuntu 18.04, C++
I only tested on #111 (branch feature/fan/Eigen) so far, but I'm 95% sure it applies to develop as well.  I'll test shortly and edit this issue accordingly.

### LINKED ISSUE #1229: Why do we add this additional condition to judge the positive qualitative of the matrix in choleskyPartial function?
## Description

I just found that when I adjust some noise params in my application, it may throw **Indeterminant linear system**  error. And then i  log the info matrix and its eigen values:
2.40599e+07,2.40596e+07,2.40599e+07,     140441,     141696,     141701,     416447,     416324,     416323,    **803.371**,     286506,     286541,     286541,    4623.53,    4622.57,    4632.01,
We can see that even the minimum eigen value after normalised(803/2.40599e+07=3.3e-5) is still larger than 0. And then i check the code and found that we have additional condition for positive qualitative.
```
bool choleskyPartial(Matrix& ABC, size_t nFrontal, size_t topleft) {
  gttic(choleskyPartial);
  if (nFrontal == 0)
    return true;

  assert(ABC.cols() == ABC.rows());
  assert(size_t(ABC.rows()) >= topleft);
  const size_t n = static_cast<size_t>(ABC.rows() - topleft);
  assert(nFrontal <= size_t(n));

  // Create views on blocks
  auto A = ABC.block(topleft, topleft, nFrontal, nFrontal);
  auto B = ABC.block(topleft, topleft + nFrontal, nFrontal, n - nFrontal);
  auto C = ABC.block(topleft + nFrontal, topleft + nFrontal, n - nFrontal, n - nFrontal);

  // Compute Cholesky factorization A = R'*R, overwrites A.
  gttic(LLT);
  Eigen::LLT<Matrix, Eigen::Upper> llt(A);
  Eigen::ComputationInfo lltResult = llt.info();
  if (lltResult != Eigen::Success)
    return false;
  auto R = A.triangularView<Eigen::Upper>();
  R = llt.matrixU();
  gttoc(LLT);

  // Compute S = inv(R') * B
  gttic(compute_S);
  if (nFrontal < n)
    R.transpose().solveInPlace(B);
  gttoc(compute_S);

  // Compute L = C - S' * S
  gttic(compute_L);
  if (nFrontal < n)
    C.selfadjointView<Eigen::Upper>().rankUpdate(B.transpose(), -1.0);
  gttoc(compute_L);

  // Check last diagonal element - Eigen does not check it
  if (nFrontal >= 2) {
    int exp2, exp1;
    // NOTE(gareth): R is already the size of A, so we don't need to add topleft here.
    (void)frexp(R(nFrontal - 2, nFrontal - 2), &exp2);
    (void)frexp(R(nFrontal - 1, nFrontal - 1), &exp1);
    return (exp2 - exp1 < underconstrainedExponentDifference);
  } else if (nFrontal == 1) {
    int exp1;
    (void)frexp(R(0, 0), &exp1);
    return (exp1 > -underconstrainedExponentDifference);
  } else {
    return true;
  }
}
```

I am confused why we add this additional condition. As far as I know, other optimize_lib like g2o and ceres don't have similar operations like this. And why the underconstrainedExponentDifference is set to 12, does it has any theoretical basis? I found many other people have meet the same issue and wondering if we can refine this part. In fact, i removed the addition condition in my local code and the result looks good.
other similar issues:
https://github.com/TixiaoShan/LIO-SAM/issues/165
https://github.com/MIT-SPARK/Kimera-VIO/issues/155

 @richardroberts @dellaert  Thank you in advance.

dellaert:
## Summary

This replaces the legacy Cholesky safeguard with an all-pivot test normalized by each variable's original diagonal:

```
min_i |R_ii| / sqrt(A_ii) >= 2^-12
```

The implementation evaluates the equivalent squared comparison,

```
R_ii^2 >= 2^-24 A_ii
```

for every diagonal pivot of the original frontal block's Eigen LLT factor. It factors the block once, adds one O(n) scan, and does not allocate another dense matrix, refactor the system, or run a condition-estimation solve. Accepted factors continue through the existing triangular solve and Schur update unchanged.

## Motivation

The previous safeguard inspected only the final two pivots and compared their binary exponents. It could therefore miss a weak pivot earlier in the factor; for a one-dimensional frontal, it also applied an absolute-scale cutoff. Those are the concrete blind spot and false-positive mechanisms behind #1229.

A raw reciprocal condition estimate is not an appropriate replacement. It changes under diagonal coordinate rescaling, so mixed variable units can look rank deficient even when the system is numerically usable after unit normalization. That is the failure in #600: at `lambda=1e-4`, the first camera frontal block is positive definite, its smallest normalized Cholesky pivot is about `0.154`, but raw `rcond(A)` is only about `8.4e-11`.

For a positive diagonal change of variable units, `R_ii^2` and `A_ii` receive the same squared scale factor, so this test is invariant to that change. The pivots of an unpivoted Cholesky factor still depend on elimination ordering. This is therefore a scale-invariant elimination safeguard, not a permutation-invariant condition-number estimate.

## Implementation

- Factor the original frontal block directly with `Eigen::LLT`.
- Reject non-finite or non-positive input diagonals and LLT failures.
- Inspect every factor diagonal before overwriting the frontal block.
- Reject non-finite pivots or normalized pivots below the exact `1 / 4096` threshold.
- Use the squared comparison to avoid a square root and division per pivot.
- Preserve the public `choleskyPartial` API and the existing solve/update path.
- Update `IndeterminateSystemException` documentation to explain nearly indeterminate systems, the normalized-pivot criterion, and its ordering dependence.
- Document a very strong finite prior combined with loose measurement noise as an explicit example, and recommend a hard constraint when the intent is exact gauge fixing.

Several optimizer tests used extremely strong finite priors as exact gauge anchors. They now express that intent with constrained noise models. The BAL unit test fixes exactly the seven-dimensional similarity gauge—first-camera pose plus one inter-camera range—while leaving camera calibration free.

The HS52 QP objective tolerance is adjusted from `1e-7` to `1e-5`. On the tested macOS toolchain, the valid sparse and dense solutions differ from the reference objective by approximately `4e-9` and `2e-6`, respectively, so the former tolerance was narrower than the solver's numerical variation.

## Validation

Focused Cholesky coverage includes:

- a three-variable near-dependent system whose weak pivot is not last
- a representative permutation of that system
- a diagonally scaled valid SPD matrix that a raw `rcond` cutoff rejects
- reconstruction of that scaled matrix from the returned factor
- very small and very large one-dimensional SPD matrices
- the existing rank-deficiency and underconstrained cases

The dependent linear-solver tests and Cholesky-sensitive optimizer tests pass. A full local Release/TBB-enabled `make -j6 check` passes all 294 tests.

## Issue #600 reproduction

On the full 394-camera, 100,368-point, 534,408-observation BAL problem, sequential and multifrontal Cholesky both accept the first `lambda=1e-4` step and reach the same one-iteration error:

- sequential: `339553.27716250974`
- multifrontal: `339553.27716251981`

## Performance

`timeSFMBAL --cholesky-only` was benchmarked on the official Dubrovnik-16 BAL problem (16 cameras, 22,106 points, 83,718 observations), comparing the pre-PR base `6d80b01f9` with `a6640cea5`. Both builds used the same Release/TBB-enabled configuration. Results are medians of 20 measured runs per build using balanced alternating order; warmups were excluded.

| Measurement | Before | After | Change |
|---|---:|---:|---:|
| Regular Cholesky | 0.5490 s | 0.5475 s | -0.27% |
| Batch Cholesky | 0.5280 s | 0.5250 s | -0.57% |
| Process wall time | 1.1555 s | 1.1473 s | -0.71% |

The differences are within run-to-run noise; no performance regression was measured.

Closes #1229  
Closes #600


ProfFan (approved):

dellaert:
Performance follow-up for the final squared-pivot implementation (`a6640cea5`):

The normalized test now compares `R_ii^2 < 2^-24 A_ii`, preserving the `2^-12` cutoff while avoiding a square root and division per pivot. It still uses the original LLT factorization and adds no allocation, factorization, or solve.

Benchmark setup:

- Release build with TBB enabled and matching compiler/configuration
- Before: pre-PR merge base `6d80b01f9`
- After: `a6640cea5`
- Official Dubrovnik-16 BAL problem: 16 cameras, 22,106 points, 83,718 observations
- `timeSFMBAL --cholesky-only`
- 20 measured runs per build in balanced alternating order; warm-ups excluded; medians reported

| Median timing | Before | After | Change |
| --- | ---: | ---: | ---: |
| Regular Cholesky | 0.5490 s | 0.5475 s | -0.27% |
| Batch Cholesky | 0.5280 s | 0.5250 s | -0.57% |
| Process wall time | 1.1555 s | 1.1473 s | -0.71% |

The differences are within run-to-run noise, so the final check shows no measurable performance regression.

On the full issue #600 BAL problem (394 cameras, 100,368 points, 534,408 observations), multifrontal Cholesky still accepts the first `lambda=1e-4` step and reaches approximately `339553.27716` after one LM iteration, with the same four-iteration convergence in `timeSFMBAL`.

Validation: `make -j6 check` passes all 294 local tests.

PR #2165, opened 2025-06-07
Fix 32 bit build of GTSAM 4.2 Python wrapper

ProfFan:
Cherry-picked my other PR to `develop` and also a fix for a GCC extension `#warning`

However it seems other parts, more than just the `gtsam::Key` issue, pops up. So marked as WIP for now.

#2089

PR #2654, opened 2026-08-10
Fix 32-bit Python bindings for 64-bit GTSAM keys

### LINKED ISSUE #2089: Build fails on 32-bit architectures

ProfFan:
Fixes #2089.

Several wrapper declarations used size_t where the corresponding C++ APIs use gtsam::Key or other fixed-width types. On i386, size_t is 32-bit while gtsam::Key is always 64-bit, so generated Python bindings rejected or narrowed symbolic keys. This was hidden on 64-bit platforms.

Changes:
- Align wrapper declarations with gtsam::Key, FactorIndex, DenseIndex, and uint64_t.
- Preserve full-width LabeledSymbol indices.
- Add Python and C++ regressions for symbolic keys and indices above 2^32.

Validation:
- Native Python suite: 419 passed, 6 skipped.
- Focused i386 regressions: 5 passed.
- Full i386 suite: 392 passed, 20 skipped; the remaining 13 failures are unrelated minimal-build or i386 numerical baselines.
- No key-conversion TypeErrors remain.

### DIFF
--- a/gtsam/discrete/discrete.i
+++ b/gtsam/discrete/discrete.i
@@ -207,7 +207,7 @@ virtual class TableDistribution : gtsam::DiscreteConditional {
 
   gtsam::TableFactor table() const;
   double evaluate(const gtsam::DiscreteValues& values) const;
-  size_t nrValues() const;
+  uint64_t nrValues() const;
 };
 
 #include <gtsam/discrete/DiscreteBayesNet.h>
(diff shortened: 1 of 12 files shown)


dellaert (approved):
Great, thanks ! 
Will you include these in #2165 or kill that and make a new PR to be included in 4.2.3 ?

Issue #803, opened 2021-06-23
aarch64 Nonlinear Optimizer Issues


jaelrod:
## Description

I have some unit tests that chain together some BetweenFactor using an odometry mock, inject some slightly contradictory supplemental evidence in the form of other factors, such as a unary factor akin to the "GPS-like" factor from LocalizationExample.cpp, and finally use LevenbergMarquardtOptimizer to `optimize()` the factor graph and assert the result.

For example, a test might start with some rather noisy mocked odometry moving forward at a consistent speed with a heading of pi/4, qualify the noisy estimates with some very not-noisy GPS-like factors that suggest movement is actually happening in the pi/2 direction, rather than the pi/4 direction, then optimize the graph and make sure that the optimized poses were along the trajectory suggested by the more reliable GPS-like factors.

Such a test and several others work fine on my desktop (x86_64, Ubuntu 18.04) but fail on my Jetson Xavier NX (aarch64, Ubuntu 18.04).

Checking out borglab gtsam source from github and running `make check` to invoke unit tests on the aarch64 device results in many failures, which I expect are related to my own, the following example of which I think is similar to one of the more basic included examples in gtsam :
```
gtsam/tests/testNonlinearOptimizer.cpp:484: Failure: "assert_equal(expected, dl_result, tol)" 
Not equal:
expected:
Values with 3 values:
Value x1: (gtsam::Pose2)
(0, 0, 0)
Value x2: (gtsam::Pose2)
(1.5, 0, 0)
Value x3: (gtsam::Pose2)
(3, 0, 0)

actual:
Values with 3 values:
Value x1: (gtsam::Pose2)
(-3.74995e-06, 0, 0)
Value x2: (gtsam::Pose2)
(1.49999, 0, 0)
Value x3: (gtsam::Pose2)
(3, 0, 0)
```

I am in the process of debugging gtsam source through verbose optimizer output and will provide updates as I make progress, but I am suspicious of Eigen anti-aliasing or use of `auto` keyword and type resolution on aarch64.

One thing I have observed so far only in aarch64 is:
```
Partial Cholesky on HessianFactor failed.

Frontal keys Position 0: 0
HessianFactor:
 keys: 0(3) 1(3) 
Augmented information matrix: [
	1.00453889587e+81, -8.863578493e+80, 1.35333966459e+82, -1.22992576855e+42, -7.68703605343e+40, 0, -1.78961791309e+41;
[...]

PR #2659, opened 2026-08-11
Avoid Rot2 reference-to-value returns on ARM64

dellaert:
## Summary

- normalize the cosine/sine scalars before constructing the returned `Rot2`
- return direct `Rot2` prvalues from `fromCosSin` and `atan2`
- add regressions for normalized factory returns, `BetweenFactor<Pose2>` error evaluation, and LM optimization with conflicting odometry and GPS-like position evidence

## Rationale

Issue #803 localized the ARM64 failures to functions returning `Rot2` by value. The 2021 workaround in #806 added a user-provided copy constructor, which made the failing tests pass but also changed `Rot2`'s triviality and potentially its calling convention. That constructor was changed back to `= default` in 846c29fa2.

The current factories normalized a local `Rot2` and returned `R.normalize()`. Because `normalize()` returned `Rot2&`, returning it from a function returning `Rot2` required the reference-to-value copy path implicated by the original report.

This change preserves the existing normalization threshold and public behavior while returning `Rot2(c, s)` directly. C++17 guarantees copy elision for that prvalue return, so the fragile copy path is removed without making the copy constructor user-provided or changing the class ABI. `atan2` delegates to the same safe construction path.

Codex believes this will address #803 because it removes the architecture-sensitive return path at the point where the corrupt cosine and sine values were observed, and the regressions exercise the behavior from raw `Rot2` construction through `BetweenFactor<Pose2>` and nonlinear optimization.

I added a separate Ubuntu/ARM workflow to test, fixed one unrelated issue, and it ran in this PR successfully. I moved the workflow to nightly to catch regressions in the future rather than adding another PR run.

## Testing

- `make -j6 testRot2.run`
- `make -j6 testNonlinearOptimizer.run`
- `git diff --check`

### DIFF
--- a/.github/workflows/build-linux-arm64.yml
+++ b/.github/workflows/build-linux-arm64.yml
@@ -3,7 +3,6 @@ name: Linux ARM64 CI
 on:
   workflow_call:
   workflow_dispatch:
-  pull_request:
 
 concurrency:
   group: linux-arm64-ci-${{ github.head_ref || github.run_id }}
--- a/.github/workflows/build-linux-arm64.yml
+++ b/.github/workflows/build-linux-arm64.yml
@@ -1,0 +1,42 @@
+name: Linux ARM64 CI
+
+on:
+  workflow_call:
+  workflow_dispatch:
+  pull_request:
+
+concurrency:
+  group: linux-arm64-ci-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: ubuntu-24.04-arm Release
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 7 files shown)


ProfFan (changes requested):
I think this is not necessary, what if you only push the test? I think it will pass even only with the test.

GCC has had AArch64 ABI bugs/changes concerning C++17 classes with empty bases and otherwise homogeneous floating-point members. GCC's own fix notes describe empty bases erroneously affecting whether such objects are passed in vector registers. https://gcc.gnu.org/gcc-10/changes.html

dellaert:
I did a number of other things in this PR (add ARM workflow in nightly, fix tiny regression) and the Rot2 changes are tiny and actually an improvement. So I'd like to merge this as is. If you don't think it will close the issue I will remove that line from the PR comment before merging.

dellaert:
But - it would be nice if "someone" assessed whether or not it actually fixes it.

ProfFan (approved):

PR #2660, opened 2026-08-11
Fix manifold preintegration Python wrapper

### LINKED ISSUE #2145: Build fails with `GTSAM_TANGENT_PREINTEGRATION` off

dellaert:
## Summary

- add a meaningful `preintegrated()` accessor to `ManifoldPreintegration`
- verify the accessor through C++ and Python regression tests

## Root cause

`navigation.i` exposes `PreintegratedImuMeasurements.preintegrated()` for every build, but only `TangentPreintegration` implemented that accessor. When `GTSAM_TANGENT_PREINTEGRATION=OFF`, the generated wrapper therefore called a nonexistent member on `PreintegratedImuMeasurementsT<ManifoldPreintegration>`.

The manifold implementation now returns the stored `NavState` as tangent coordinates, matching `biasCorrectedDelta(biasHat())` and preserving the existing Python API.

## Validation

Configured a clean Python build with `GTSAM_TANGENT_PREINTEGRATION=OFF` and ran:

- `make -j6 testManifoldPreintegration.run`
- full `gtsam_py` wrapper build
- `python -m unittest -v gtsam.tests.test_PreintegratedImuMeasurements`
- `python -m unittest -v gtsam.tests.test_ImuFactorWithGravity`
- `git diff --check`

Closes #2145


ProfFan (approved):
LGTM

PR #2661, opened 2026-08-11
Enable correct Lie-group Jacobians in Prior/Between by default

dellaert:
## Summary

This is an important correctness and convergence change for GTSAM's Lie-group factors.

`BetweenFactor` has historically returned the exact nonlinear residual while, by default, omitting the Jacobian of `Local`. `PriorFactor` similarly hard-coded an identity Jacobian. Those approximations are exact at zero residual, but they give the optimizer an inconsistent local model away from the linearization point.

This PR:

- makes Lie-group derivative support explicit in the traits API;
- removes unsupported Jacobian overloads and aligns the generated wrappers with the actual C++ API;
- uses the `Local` Jacobian in `BetweenFactor` and `PriorFactor` whenever the traits provide it;
- preserves the existing fallback for groups that do not provide `Local` Jacobians;
- makes `GTSAM_SLOW_BUT_CORRECT_BETWEENFACTOR` default to `ON`; and
- checks in reproducible micro- and end-to-end benchmarks using the shared timing infrastructure.

The name of the flag is now misleading: this is not simply "slow but correct." The correct Jacobians are more expensive in isolation, but they can make the complete optimizer substantially faster by reducing rejected LM steps and nonlinear iterations.

## Why this matters

For a between measurement, the residual is

```text
Local(measured, Between(x1, x2)).
```

The old default differentiated `Between(x1, x2)` but not the surrounding `Local`. The resulting Jacobian error is small near zero residual and can be significant elsewhere. The same issue appears in the prior factor, where identity is only the correct derivative at zero residual in general.

This affects more than derivative checking. It changes the local quadratic model used by Gauss-Newton and Levenberg-Marquardt, and therefore changes step acceptance, damping, convergence speed, and the basin of convergence.

## Isolated factor cost and accuracy

`timing/timeBetweenFactor.cpp` migrates the experiment from #88 to the current timing infrastructure. It requests both `Rot3` Jacobians at a nonzero residual and compares them with numerical derivatives.

| Path | Median | H1 max error | H2 max error |
| --- | ---: | ---: | ---: |
| Legacy approximation | 47.85 ns/call | 5.89e-3 | 5.01e-3 |
| Correct `Local` Jacobian | 114.10 ns/call | 2.08e-12 | 2.83e-12 |

The corrected path costs about 66 ns more in this isolated benchmark and recovers numerical precision.

## End-to-end `w10000` refinement after FAST-Sync

`timing/timeBetweenFactorPoseGraph.cpp` loads the 10,000-pose, 64,311-factor `w10000` graph. It computes one shared FAST-Sync initialization, outside the timed region, and feeds those identical initial values to two interleaved LM refinements: the current factors and benchmark-only copies of the legacy Jacobian paths.

FAST-Sync took 0.638 s and reduced the graph error from 24,720,119.96 to 23,454.32.

| Refinement | Correct Jacobians | Legacy Jacobians |
| --- | ---: | ---: |
| Median time | 1.288 s | 3.041 s |
| Outer LM iterations | 10 | 23 |
| Inner LM attempts | 17 | 42 |
| Final error | 144.910 | 144.877 |

Both variants reach essentially the same objective. The correct Jacobians are **2.36x faster end to end after initialization**, with 56.5% fewer outer iterations and 59.5% fewer inner attempts. Including FAST-Sync, total time is about 1.93 s versus 3.68 s.

This reverses the microbenchmark result: paying tens of nanoseconds for the correct factor Jacobian saves more than half of the nonlinear refinement time because LM sees the right local model.

## Raw-initialization convergence result

The benchmark also retains `--raw-initial` because it exposes a more dramatic difference in the convergence basin.

| Raw `w10000` start | Correct Jacobians | Legacy Jacobians |
| --- | ---: | ---: |
| Outer LM iterations | 9 | 9 |
| Inner LM attempts | 9 | 28 |
| Final error | 144.863 | 4,642,335.23 |
| Final lambda | approximately 0 | 100,000 ceiling |

Here the legacy optimizer returns normally but does not reach a useful solution. I validated this separately by rebuilding the production library with `GTSAM_SLOW_BUT_CORRECT_BETWEENFACTOR=OFF`: the real production factors and the copied legacy factors produced identical iteration counts, final error, and final damping. This is therefore not a benchmark artifact.

I do not interpret this raw-start result as a timing comparison. It shows that the correct derivatives materially enlarge the useful convergence basin.

## Compatibility

The new traits detector only requests `Local` Jacobians from types that actually provide them. Types such as `Similarity3`, `Similarity2`, `SO4`, and `SL4` retain the derivative-free fallback and continue to compile. Frobenius factors do not use `Local` and are unchanged.

The CMake flag remains available for comparison and compatibility, but now defaults to `ON`.

## Validation

The focused C++ suites pass with the flag enabled:

```text
testBetweenFactor
testPriorFactor
testExtendedPriorFactor
testLieGroupDerivatives
testSimilarity3
testFrobeniusFactor
testFastSync
testTimingUtils
```

The Python suite passed with 421 tests and 4 skips during the Lie API and wrapper cleanup.

This PR supersedes the experiment in #88 with the implementation, compatibility path, and reproducible timing evidence needed to make the correct behavior the default.


dellaert:
@ProfFan this is a big deal. Speedup is much more dramatic than I suspected.
As part of this PR, we now no longer "pretend" to have Jacobians at compile time, and throwing at runtime. This is why this PR touches that many files - although most is rather mechanical.

ProfFan:
Why did we remove all the Jacobians for Logmap/Expmaps?

dellaert:
> Why did we remove all the Jacobians for Logmap/Expmaps?

Did we? My intent was only to remove the ones that threw an exception. So if not implemented is it inspectable at compile time.

ProfFan:
Ah I see, then it's a good

ProfFan (approved):
Might be good to check the CI matrix as well, for the new default CMake option

dellaert:
Fixing CI now

ProfFan (approved):

varunagrawal:
Glad to see this fixed and the benchmarks providing valuable information about correctness and speed.

PR #2671, opened 2026-08-12
Add Frobenius factor support to FastSync

dellaert:
## Summary

- teach `FastSync` to consume `FrobeniusBetweenFactor` and `FrobeniusPrior` directly
- preserve support for the existing `BetweenFactor` and `PriorFactor` inputs
- expose the small measurement/prior accessors needed by the generic implementation
- add Rot3 coverage for a Frobenius graph with a prior

## Why

Rotation-averaging clients that already construct a Frobenius QCQP graph should not need to build a duplicate graph of conventional between/prior factors solely for FAST-Sync initialization. This keeps FAST-Sync usable as an application-level initializer while allowing the downstream Burer--Monteiro API to remain purely QCQP-based.

## Validation

- `make -j6 testFastSync.run`
- `make -j6 testFrobeniusFactor.run`
- `git diff --check`

### DIFF
--- a/gtsam/slam/FastSync-inl.h
+++ b/gtsam/slam/FastSync-inl.h
@@ -21,6 +21,7 @@
 #include <gtsam/linear/GaussianFactorGraph.h>
 #include <gtsam/linear/JacobianFactor.h>
 #include <gtsam/linear/NoiseModel.h>
+#include <gtsam/slam/FrobeniusFactor.h>
 
 #include <cmath>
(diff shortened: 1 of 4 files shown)


dellaert:
@ProfFan Adding you because Ian is probably still out for the summer and I can't merge it without a review. It's a very simple change allowing different types of factors to also create a fast sync init.

ProfFan (approved):

PR #2675, opened 2026-08-12
Optimize QCQP construction for generic BM

dellaert:
## Summary

- make nonlinear-factor-graph to QCQP lowering linear by merging quadratic equality constraints through a per-key index while preserving source cost order, distinct same-key quadratics, and all nonquadratic constraints
- cache the constructor-built rank-`pMin` QCQP so fail-fast validation does not force a second lowering at the first staircase level
- separate QCQP build, local ALM solve, and certificate timings in C++, Python, the examples, benchmark CSV, and benchmark-action JSON

This is the third PR in the stack and targets #2672. It keeps BM generic and QCQP-based; it does not change ALM or add rotation-specific behavior.

## Benchmark

Release build with AppleClang 21.0.0 (`-O3 -DNDEBUG`). The parent commit and this branch were built separately with identical compiler/build flags and run consecutively under the same machine load.

```sh
./timing/timeShonanInitialization \
  --g2o-directory /Users/dellaert/git/shonan-github/datasets/yfcc2 \
  --largest 2 \
  --methods shonan-pcg,bm-alm-default,bm-alm-tuned \
  --warmup 1 \
  --repeats 3 \
  --csv /tmp/bm-qcqp-{base,new}-top2-sameload.csv
```

The ALM parameters are unchanged: tuned `initialMuEq=1`, `muEqIncreaseRate=2`, `maxIterations=50`, absolute/relative violation tolerance `1e-8`, and absolute/relative cost tolerance `1e-10`; default uses the existing `AugmentedLagrangianParams` defaults.

Mean end-to-end time (initialization + BM solve), with every attempt included:

| Dataset | Method | Before | After | Speedup |
|---|---|---:|---:|---:|
| `taj_mahal` | BM/ALM default | 14.723 s | 2.982 s | 4.94x |
| `taj_mahal` | BM/ALM tuned | 13.631 s | 2.910 s | 4.68x |
| `brandenburg_gate` | BM/ALM default | 18.621 s | 7.964 s | 2.34x |
| `brandenburg_gate` | BM/ALM tuned | 19.277 s | 8.410 s | 2.29x |

The Shonan/PCG controls were 0.706 -> 0.760 s and 0.696 -> 0.808 s respectively, showing that the after run was not benefiting from a lighter machine load.

### Timing accounting

The phase measurements are nested rather than additional:

```text
End-to-end time
├── Initialization: FAST-Sync + conversion to QCQP values
└── BM solve wall time
    ├── QCQP construction
    ├── Local ALM optimization
[...]

PR #2676, opened 2026-08-12
Prefer system SuiteSparse with a namespaced CCOLAMD fallback

### LINKED ISSUE #2656: Deal w CCOLAMD
Keep vendored but allow to find a system, rename like cephes.

ProfFan:
## Summary

- Add `GTSAM_USE_SYSTEM_SUITESPARSE`, defaulting to `ON`.
- When unset, try the official `CCOLAMD` config package and fall back to the bundled implementation if unavailable.
- Fail configuration when explicitly enabled but `SuiteSparse::CCOLAMD` is unavailable.
- Prefix bundled CCOLAMD and SuiteSparse_config symbols with `gtsam_` to prevent collisions.
- Keep CCOLAMD includes source-local and out of GTSAM's public include interface.
- Use the official config package without adding `FindCCOLAMD.cmake`.

## Validation

- `testOrdering` passes for system and bundled SuiteSparse in shared and static builds.
- Installed-package consumers pass in all four configurations.
- Bundled GTSAM can coexist with system CCOLAMD in the same executable.
- Symbol checks confirm the bundled library exposes no unprefixed CCOLAMD or SuiteSparse_config definitions.

## Note

This intentionally relies on SuiteSparse's official CMake package. On systems where that package requests OpenMP's C component, downstream system-mode consumers must enable the C language.

Resolves #2656


dellaert:
Could we clarify or reconsider the name GTSAM_USE_SYSTEM_SUITESPARSE? This option selects system CCOLAMD plus SuiteSparse_config, not SuiteSparse generally. GTSAM_USE_SYSTEM_CCOLAMD would be more precise, or the option description/docs could make the narrower scope explicit.

dellaert:
The PR removes CCOLAMD and SuiteSparse_config from GTSAM's public include interface, but gtsam/3rdparty/CMakeLists.txt still installs the original unprefixed headers. In bundled mode, GTSAM now provides only the gtsam_* symbols, so an installed consumer can find a header that declares functions it cannot link against. Should we stop installing these private headers, or explicitly document that they are not part of GTSAM's API and are not linkable against the bundled implementation?

dellaert:
Two comments above were two comments offered by codex. Others were not as relevant/needed.

dellaert:
Approving anyway.

dellaert (approved):
Cool!

dellaert:
Caches are cool :-)
image

dellaert:
Please only merge after I merged #2675

PR #2679, opened 2026-08-13
Upgrade release/4.2 CI actions and stabilize Linux wheels

dellaert:
## Summary

- upgrade supported GitHub Actions used by the `release/4.2` packaging and dispatch workflows to Node 24-compatible releases
- replace the unmaintained Node 12 `ProfFan/repository-dispatch` action with `peter-evans/repository-dispatch@v4`
- serialize the memory-heavy GTSAM/Python wrapper install in Linux wheel containers while retaining two-job Boost and macOS builds

## Motivation

GitHub-hosted runners are moving JavaScript actions from deprecated Node runtimes to Node 24. The release branch still used several older first-party action releases and one unmaintained Node 12 dispatch action.

The Linux wheel wrapper build also used `-j2` for large generated translation units. That can exceed the memory available in the manylinux container, particularly in the x86_64 Python 3.11 job. Running the wrapper install serially avoids that OOM failure without slowing the separate Boost build.

`ilammy/msvc-dev-cmd@v1` remains unchanged because upstream has no safe Node 24 release; it is the only retained Node 20 action.

## Validation

- `bash -n` passed for all Python wheel shell scripts
- `actionlint -shellcheck= .github/workflows/*.yml` passed
- full `actionlint` reported only pre-existing ShellCheck findings in untouched build workflows
- `git diff --check` passed
- audited every `uses:` declaration: no Node 12 or Node 16 actions remain, and Node 20 remains only for `ilammy/msvc-dev-cmd@v1`
- verified the artifact upload/download and repository-dispatch inputs remain supported by the upgraded actions

No C++ or Python behavior changed; GitHub Actions provides the platform-specific wheel acceptance test.

### DIFF
--- a/.github/scripts/python_wheels/cibw_before_all.sh
+++ b/.github/scripts/python_wheels/cibw_before_all.sh
@@ -5,7 +5,16 @@ set -x
 
 PYTHON_VERSION="$1"
 PROJECT_DIR="$2"
-WHEEL_BUILD_JOBS=2
+BOOST_BUILD_JOBS=2
+GTSAM_BUILD_JOBS=2
+
+# The generated Python wrapper translation units require substantial memory.
+# Compile them serially in Linux wheel containers to avoid the OOM killer.
+if [ "$(uname)" == "Linux" ]; then
+    GTSAM_BUILD_JOBS=1
+fi
+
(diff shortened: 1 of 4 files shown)


dellaert:
@ProfFan upgrading our 4.2 workflows to keep up w Github infra

ProfFan (approved):

PR #2681, opened 2026-08-13
Remove vendored Ceres and use system Ceres for timing targets

### LINKED ISSUE #2655: Remove ceres
and conditionally compile the two timing scripts only if ceres is there.

ProfFan:
## Summary

- remove the vendored Ceres autodiff subset and its build and license metadata
- update AdaptAutoDiff and its tests to use the public Ceres AutoDiffCostFunction API
- look for system Ceres only while configuring timing targets
- disable the Ceres-dependent test and timing executables when Ceres is unavailable
- preserve correct system Eigen version reporting for modern Eigen CMake packages

## Testing

- configured with bundled Eigen 3.4.0 and system Ceres 2.2.0
- configured and built with system Eigen 5.0.1 and system Ceres 2.2.0
- built and ran testAdaptAutoDiff.run successfully
- built timeAdaptAutoDiff and timeSFMBALautodiff successfully
- git diff --check

Resolves #2655


gtsam/nonlinear/tests/CMakeLists.txt
```diff
@@ -10,4 +10,7 @@ if(NOT GTSAM_ENABLE_CUDA AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/testCudaDeviceV
   list(APPEND EXCLUDE_TESTS "testCudaDeviceValues.cpp")
 endif()
 
+# This optional Ceres test is configured alongside the Ceres timing targets.
+list(APPEND EXCLUDE_TESTS "testAdaptAutoDiff.cpp")
```
dellaert: How do I run the test?

timing/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+find_package(Ceres CONFIG QUIET)
+if(Ceres_FOUND)
+  message(STATUS "Found Ceres ${CERES_VERSION}")
+else()
+  message(STATUS "Ceres not found; Ceres-dependent tests and timing scripts disabled")
```
dellaert: This is in the timing directory, so I'm assuming it's only timing.

dellaert (approved):
LGTM, but a better design would be to detect Ceres centrally, then conditionally include the test from the nonlinear test directory.

ProfFan:
I moved the test to the independent tests folder, and conditioned on finding Ceres

PR #2685, opened 2026-08-13
Suppress GCC 16 Spectra false positive

ProfFan:
## Summary

- Suppress the GCC 16 false-positive -Wmaybe-uninitialized diagnostic emitted by Spectra SortEigenvalue.
- Scope the diagnostic override to the Spectra includes in RiemannianStaircaseOptimizer.cpp.

## Testing

- cmake --build build -j6
- cmake --build build --target testRiemannianStaircase.run -j6
- git diff --check


dellaert (approved):
LGTM!

PR #2686, opened 2026-08-13
Avoid recursive BayesTree serialization stack overflow

### LINKED ISSUE #1434: ISAM large graph serialization crash
## Description

Serialization of an ISAM2 solver object reliably fails after the graph reaches a certain size.

## Steps to reproduce

Download data file here:  https://drive.google.com/file/d/1bVyEQc3ogHmfuGTL9fHVptIqakV_15Vs/view?usp=sharing

Compile and run the new gtest added here:

https://github.com/ScottMcMichael/gtsam/blob/isam_serialization_bug/gtsam/nonlinear/tests/testSerializationNonlinear.cpp

Either put the data file in the execution directory or manually set the path to the data file in the serialization test.

(executable is testSerializationNonlinear)

## Expected behavior

The test will produce the following output:

Write: .//gt0.bin
Segmentation fault (core dumped)

## Environment

Ubuntu 20.04.5, default CMake build configuration

## Additional information

The crash appears to be caused by a stack overflow, either from an infinite recursion or a very expansive recursion that needs to be replaced with another kind of loop.  According to the stack trace the two lines of GTSAM code that the program keeps bouncing between are:

#542768 gtsam::ISAM2Clique::serialize<boost::archive::binary_oarchive> (ar=..., this=0x7ffa7c001dd0) at .../install/gtsam/include/gtsam/nonlinear/ISAM2Clique.h:154

and

#542739 gtsam::BayesTreeCliqueBase<gtsam::ISAM2Clique, gtsam::GaussianFactorGraph>::serialize<boost::archive::binary_oarchive> (ar=..., this=0x7ffa7c001
dd0) at .../install/gtsam/include/gtsam/inference/BayesTreeCliqueBase.h:212

dellaert:
## Summary

- serialize version-1 Bayes trees as a breadth-first list of detached shallow clique copies plus ordered child IDs
- reconstruct parent/child links, roots, and the frontal-key index iteratively on load
- retain the version-0 `nodes_`/`roots_` path for legacy archives and unversioned external derived trees
- version the built-in Gaussian and discrete trees that directly inherit the base serializer

Boost preserves pointer identity, but the existing clique serializer still recursively follows child links and can exhaust the stack for deep trees. This keeps the flat-ID insight while reusing each clique's existing serialization unchanged: links are cleared only on temporary copies and restored from the child table after loading.

The approach was inspired by @ScottMcMichael's [proposed flat-ID fix](https://github.com/ScottMcMichael/gtsam/tree/fix_large_graph_serialization_stack_overflow) for #1434, while preserving legacy loads and minimizing changes outside `BayesTree`.

## Tests

- `make -j6 testSerializationLinear.run`
- `make -j6 testSerializationNonlinear.run`
- `make -j6 testSerializationHybrid.run`
- `make -j6 testSerializationDiscrete.run`
- `git diff --check`

Closes #1434.

### DIFF
--- a/gtsam/discrete/DiscreteBayesTree.h
+++ b/gtsam/discrete/DiscreteBayesTree.h
@@ -119,3 +119,7 @@ template <>
 struct traits<DiscreteBayesTree> : public Testable<DiscreteBayesTree> {};
 
 }  // namespace gtsam
+
+#if GTSAM_ENABLE_BOOST_SERIALIZATION
+BOOST_CLASS_VERSION(gtsam::DiscreteBayesTree, 1)
+#endif
--- a/gtsam/inference/BayesTree.h
+++ b/gtsam/inference/BayesTree.h
@@ -19,13 +19,23 @@
 
 #pragma once
 
-#include <memory>
-
 #include <gtsam/inference/Key.h>
 #include <gtsam/base/FastList.h>
 #include <gtsam/base/ConcurrentMap.h>
 #include <gtsam/base/FastVector.h>
 
+#if GTSAM_ENABLE_BOOST_SERIALIZATION
+#include <boost/serialization/shared_ptr.hpp>
+#include <boost/serialization/split_member.hpp>
+#include <boost/serialization/vector.hpp>
+#include <boost/serialization/version.hpp>
+
+#include <stdexcept>
+#include <unordered_map>
+#include <vector>
+#endif
+
+#include <memory>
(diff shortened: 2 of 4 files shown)


dellaert:
I really had codex do this so @ScottMcMichael , If you want to take a look, I would be obliged.

dellaert:
Merging for now, but I can always include comments in a later PR.

PR #2691, opened 2026-08-14
Improve batch-factor and LM optimization hot paths

dellaert:
## Summary

- ignore null Gaussian factors during key-dimension discovery
- evaluate compact batch-factor error changes and Hessian diagonals without materializing dense Jacobians
- expose fixed compact rows for batched damping
- add overridable LM damping and linear-model evaluation hooks while preserving the default behavior

These are classical GTSAM optimizer improvements and do not select a new linear solver backend.

## Validation

- `make -j6 testPCGSolver.run`
- `make -j6 testBatchJacobianFactor.run`
- `make -j6 testBatchFactor.run`
- `make -j6 testNonlinearOptimizer.run`

A separate BAL-88 before/after timing comment will follow.

### DIFF
--- a/gtsam/linear/BatchJacobianFactor.h
+++ b/gtsam/linear/BatchJacobianFactor.h
@@ -513,8 +513,9 @@ class BatchJacobianFactor : public BatchJacobianFactorBase {
       const auto& A = std::get<I>(blocks_)[rowIndex];
       const RhsVector& b = rhs_[rowIndex];
       if (weights) {
-        updateOffDiagonalNormalized(
-            targetI, targetJ, A.transpose() * weights->asDiagonal() * b, info);
+        const RhsVector weightedRhs = weights->asDiagonal() * b;
+        updateOffDiagonalNormalized(targetI, targetJ,
+                                    A.transpose() * weightedRhs, info);
       } else {
         updateOffDiagonalNormalized(targetI, targetJ, A.transpose() * b, info);
       }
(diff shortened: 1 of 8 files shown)


dellaert:
BAL-88 timing (Release, identical build directory/configuration, three runs per revision)

Command:

```
./timing/timeSFMBAL --cholesky-only /private/tmp/problem-88-64298-pre.txt
```

| Revision | Regular samples (s) | Regular median (s) | Batch samples (s) | Batch median (s) |
|---|---|---:|---|---:|
| parent `f214e7bc4d` | 3.077, 2.633, 2.723 | 2.723 | 2.494, 2.414, 2.209 | 2.414 |
| PR tip `ca155d5512` | 2.988, 2.195, 2.436 | 2.436 | 2.012, 2.021, 1.936 | 2.012 |

The batch-factor median is 16.7% lower (1.20× faster) at the PR tip. That is the path this PR intentionally optimizes.

The regular path also moved by 10.5% even though the null-factor fix and LM hooks are behavior-neutral and should not accelerate it. Together with the visibly wide sample spread, that indicates substantial machine/run-order noise. I therefore treat the batch result as encouraging evidence, not a precise attribution of the full 16.7% to the compact kernels. There is no claimed performance effect from the null handling or default virtual LM hooks.

dellaert:
@ProfFan I reviewed all changes by Codex. I am not in love with state of gtsam/linear/BatchJacobianFactor.h - might refactor later - but given this PR is the bottom of the stack I'd prefer to do that in a follow up.

PR #2693, opened 2026-08-14
Add flat Gaussian-factor kernels and BAL smart/PCG benchmarks

dellaert:
## Summary

- add an exported `FlatGaussianFactor` interface for applying Gaussian-factor kernels with solver-compiled scalar offsets and block slots
- route compact batch factors and implicit-Schur factors through one reusable flat execution plan in `PCGSolver`
- precompute key mappings, assemble block-Jacobi diagonals without keyed maps, and parallelize flat-factor products with thread-local reduction
- implement scratchless preindexed kernels for `RegularImplicitSchurFactor`
- share smart-factor BAL graph construction, camera-only values, and ordering
- expand `timeSFMBALsmart` with direct, Hessian-PCG, and implicit-Schur-PCG modes plus machine-readable output

### Why `BatchPlan` became `FlatFactorPlan`

The PCG execution plan was previously named `BatchPlan` because it handled only `BatchJacobianFactorBase`. Dispatch now uses the `FlatGaussianFactor` interface, so the same plan serves both compact batch factors and `RegularImplicitSchurFactor`. `FlatFactorPlan` names the interface the plan executes and avoids incorrectly implying that implicit-Schur factors are batch factors. The rename has no algorithmic effect.

Smart factors provide a camera-only benchmark for the flat Gaussian-factor interface; this does not imply that structureless bundle adjustment is intrinsically faster.

This draft is based on #2691.

## Validation

- `make -j6 testPCGSolver.run`
- `make -j6 testBatchJacobianFactor.run`
- `make -j6 testBatchFactor.run`
- `make -j6 testSfmBalBenchmark.run`
- Release build of `timeSFMBALsmart`

BAL-16 runtime validation on the PR tip:

| Linearization | Solver | Final cost | Linear solves | Avg. CG iterations | Non-converged |
|---|---|---:|---:|---:|---:|
| Hessian | multifrontal Cholesky | 18124.659576 | n/a | n/a | n/a |
| Hessian | parallel block-Jacobi PCG | 18124.683266 | 4 | 28 | 0 |
| implicit Schur | parallel block-Jacobi PCG | 18124.683266 | 4 | 28 | 0 |

The two PCG representations reach the same final smart-factor objective, and both closely agree with the direct Hessian path.

## BAL-88 timing

Release optimizer timings for the retained flat implicit-Schur mode:

```bash
timeSFMBALsmart --implicit-schur-pcg-only <BAL-88>
```

| Run | Optimizer time |
|---|---:|
[...]

PR #2694, opened 2026-08-14
Assemble compact point-batch sparse Schur systems

dellaert:
## Summary

This PR adds the backend-neutral half of an explicit-point Schur solver for the BAL benchmark. It takes the compact camera–point Jacobian batches produced by the batch-factor path, eliminates each 3-DOF landmark, and returns a camera-only linear system in fixed 9×9 blocks. A later backend can solve that camera system without knowing how BAL factors, damping, or landmark recovery are represented.

The implementation lives under `timing/internal`; it is benchmark infrastructure, not a new public GTSAM sparse-solver API.

## Motivation

A generic Schur-complement path first expands the linearized graph into general-purpose matrix structures and then discovers the camera/landmark block structure during elimination. BAL already has a fixed structure:

- each residual row is 2-dimensional;
- each camera delta is 9-dimensional;
- each landmark delta is 3-dimensional; and
- measurements belonging to one landmark are already grouped in a `BatchJacobianFactor<2, 9, 3>`.

This PR preserves that structure from linearization through landmark elimination. It avoids forming the full camera-plus-point Hessian and produces data that can be handed directly to a sparse camera-system backend.

### Compact camera system

`CompactCameraSystem` contains:

- the number of cameras;
- packed upper-triangular 9×9 camera blocks;
- a mask identifying blocks that are structurally present;
- the reduced camera right-hand side; and
- the per-landmark state needed for back-substitution.

`upperCameraBlockIndex()` gives a backend-independent mapping from a camera pair to its packed block slot. Unused blocks remain zero and are marked absent, so a backend can emit only the actual sparse block pattern.

### Direct landmark elimination

`buildPointBatchCameraSystemParallel()` processes one compact batch per landmark. For each landmark it forms and factors only the 3×3 point block `H_pp`, then accumulates the usual Schur terms

```text
H_reduced = H_cc - H_cp H_pp^-1 H_pc
b_reduced = b_c  - H_cp H_pp^-1 b_p
```

directly into fixed-size camera blocks and the camera RHS. The retained landmark state is limited to the original compact batch, the landmark key, `H_pp^-1`, and the point information RHS.

Assembly is parallelized over landmarks. Each worker uses a private camera-block accumulator and RHS, followed by a deterministic reduction into the returned system; workers do not contend on shared matrix blocks.

### Landmark recovery

`backSubstitutePointBatchLandmarksParallel()` accepts a camera solution from any backend and recovers all 3-vector landmark deltas in parallel:

```text
delta_p = H_pp^-1 (b_p - H_pc delta_c)
```

Landmarks that have damping but no measurement batch are retained with a zero delta.

### LM integration

`PointBatchSchurLevenbergMarquardtOptimizer` provides the common LM behavior needed by a future camera-system solver:

- `buildDampedSystem()` emits camera and point damping as two compact unary batches instead of one factor per variable; and
- `linearDeltaError()` evaluates the linearized model change in parallel.

A derived optimizer therefore only needs to implement the camera-system solve. These overrides use the virtual hooks introduced by #2691.

## Accepted input and failure modes

The compact path deliberately has a narrow BAL contract:

- camera keys are contiguous `c0 ... cN` variables of dimension 9;
- landmark keys are `p*` variables of dimension 3;
- there is at most one compact measurement batch per landmark;
- each measurement row couples one camera and one landmark;
- compact measurement factors are already whitened (unit noise); and
- additional linear factors are unary camera or landmark terms, including LM damping.

The implementation reports clear runtime errors for unsupported factor shapes, non-contiguous camera indices, duplicate or empty landmark batches, non-unit compact noise models, and singular 3×3 landmark blocks.

## Scope and dependencies

This PR is based on #2691, which supplies the compact `BatchJacobianFactor` operations and virtual LM hooks.

It intentionally contains no CHOLMOD discovery, headers, linkage, solver calls, or command-line mode. The output is backend-neutral so the sparse backend can be reviewed independently. It is also independent of the flat/smart-factor/PCG work in draft #2693.

## Validation

```sh
make -j6 testSfmPointBatchSchur.run
```

The focused test covers two systems:

1. a hand-constructed two-camera, one-landmark system; and
2. the repository's tiny BAL fixture using the production batch linearization.

For both, it compares the assembled reduced Hessian and RHS with GTSAM's generic dense Schur complement, solves the reduced system, performs landmark back-substitution, and compares every camera and landmark delta with the full reference solution.

Dense matrices are used only to construct the tiny test oracle. The implementation and BAL measurement path never form a dense camera-plus-point Hessian.


dellaert:
Follow-up validation now uses the repository's requested tiny BAL fixture, `dubrovnik-3-7-pre`, in addition to the synthetic fixed-block system. The test linearizes the real point-batch graph, adds small unary damping, and compares the compact reduced Hessian, reduced RHS, camera deltas, and recovered landmark deltas with the generic dense reference. `make -j6 testSfmPointBatchSchur.run` passes on the updated tip `4878fe7c85`.

ProfFan (approved):

dellaert:
Pushed an update that implements partial multifrontal elimination as a two-phase ordering and optimizes the large-leaf Schur path.

The main performance change is that sufficiently large Cholesky leaf cliques no longer materialize the dense separator block (C). They form only the frontal rows ([A\ B\ a]), accumulate the original separator contribution directly into the parent, and subtract ([S\ d]^T[S\ d]). Small separators keep the existing dense path, and load + partial elimination can now run in one fused bottom-up traversal. Because this is implemented in `MultifrontalClique`, complete multifrontal solves benefit as well.

Reduced-system timings (`lambda=0.04`, leaf merge 256, general merge 32; medians, symbolic setup excluded):

| BAL dataset | Compact SFM system | Fused multifrontal | Ratio |
|---|---:|---:|---:|
| 16 cameras / 22,106 points | 14.8 ms | 19.7 ms | 1.33x |
| 88 cameras / 64,298 points | 71.8 ms | 158.5 ms | 2.21x |
| 135 cameras / 90,642 points | 154.1 ms | 430.7 ms | 2.79x |

BAL-135 was scheduler-sensitive (best fused sample 297.5 ms), but this is still substantially below the previous generic path at roughly 1.5–1.6 s. The reduced Hessian and RHS agree with the compact SFM implementation to about (10^{-15}). Focused batch-factor, partial/full multifrontal, and nonlinear multifrontal tests pass.

dellaert:
Follow-up on the partial multifrontal work in `8b4528a251`:

### What changed

- Replaced algebraic merging of independent sibling leaves with bounded scheduling aggregation.
- Added an optional Cholesky-only `SameSeparator` mode. Leaves with identical separator scatter maps are accumulated into separator-local information matrices and scattered into the parent once per batch.
- Identical-separator groups are divided into nearly equal batches whose conservative total problem size is bounded by `leafAggregationProblemSize` (currently 2048).
- QR leaves are excluded from separator accumulation.
- Removed the experimental one-batch-per-parent mode because it eliminated too much point-level parallelism.
- Applied bounded scheduling to partial, complete, and damped nonlinear bottom-up elimination.

### Reduced-camera assembly timing

Global damping 0.04, general merge disabled, 2 warmups, 7 repetitions; entries are fused median times.

| Dataset | Compact | Bounded scheduling | Same separator |
| --- | ---: | ---: | ---: |
| BAL16 | 11.714 ms | **17.197 ms** | 19.542 ms |
| BAL88 | 72.261 ms | **143.884 ms** | 146.195 ms |
| BAL135 | 150.445 ms | 467.559 ms | **398.140 ms** |

BAL135 remained noisy: bounded samples ranged from 359–1035 ms and same-separator samples from 306–700 ms.

### Full SFM elimination and solve

I also reran `timeMultifrontalSolver` three times. Each reported sample contains two complete eliminate-and-solve iterations after a warmup. The table reports mean and observed range across the three independent runs.

| Dataset | New multifrontal mean (range) | Standard GTSAM mean | New/standard time |
| --- | ---: | ---: | ---: |
| BAL16 | 0.585 s (0.567–0.608) | 0.381 s | 1.53x |
| BAL88 | 4.798 s (4.570–4.955) | 3.306 s | 1.45x |
| BAL135 | 8.848 s (8.535–9.250) | 5.097 s | 1.74x |

The full-solver spread is about 7–8%, so the new scheduling has not introduced extreme timing variance. However, full BAL elimination/solve is still slower than standard GTSAM, particularly on BAL135; that remains an optimization target.

### Validation

- `testForestTraversal.run`
- `testMultifrontalSolver.run`
- Reduced systems agree within approximately `4e-15`.
- Aggregation modes preserve the symbolic clique count.

dellaert:
Post-merge benchmark update (`229a8e0d8a`, Release build, local arm64 machine):

```console
./timing/timeSfmPartialElimination \
  --dataset dubrovnik-88-64298-pre.txt \
  --warmup 1 --repeats 7
```

Dataset: 88 cameras, 64,298 points. Parameters: damping `0.04`, bounded leaf aggregation `2048`, merge cap `32`.

| Pipeline | Median ms | Mean ms | Min ms | Max ms |
| --- | ---: | ---: | ---: | ---: |
| Compact complete assembly | 174.761 | 186.814 | 168.193 | 247.201 |
| Partial fused end-to-end | 455.349 | 454.263 | 407.650 | 552.817 |
| Partial elimination only | 453.901 | 453.085 | 406.783 | 551.087 |
| Retained-factor export only | 0.972 | 1.178 | 0.867 | 1.730 |
| Separate load + eliminate (diagnostic) | 492.945 | 527.435 | 418.077 | 745.879 |
| Separate load | 33.925 | 41.726 | 27.005 | 64.571 |
| Separate eliminate | 450.753 | 485.708 | 391.071 | 681.308 |

Primary partial end-to-end / compact median ratio: **2.606x**. Elimination-only / compact diagnostic ratio: **2.597x**.

The exported retained system still matches the compact reference: maximum relative Hessian difference `1.006e-15`; maximum relative RHS difference `4.327e-16`. Symbolic setup was `718.144 ms` and is reported separately from the measured pipelines.

dellaert:
Pushed two performance follow-ups:

- 30b6fdb800 — make the one-thread multifrontal path genuinely sequential and give both benchmark pipelines the same explicit thread count
- a3184c1d3a — cache Hessian block scalar destinations and use fixed-size BatchJacobianFactor products in the direct Cholesky path

Latest BAL 88-camera results (problem-88-64298-pre.txt; 64,298 points; lambda 0.04; bounded leaf mode; leaf aggregation 2048; merge cap 32; 1 warmup + 7 measured repetitions):

| Threads | Compact complete assembly, median | Partial fused end-to-end, median | Partial / Compact |
| --- | ---: | ---: | ---: |
| 1 | 255.8 ms | 462.2 ms | 1.807x |
| 10 | 78.3 ms | 133.5 ms | 1.704x |

| Pipeline | 1-to-10-thread speedup |
| --- | ---: |
| Compact | 3.27x |
| Partial | 3.46x |

Retained-factor export was 0.46 ms single-threaded and 0.53 ms with 10 threads, so export is not the remaining gap. Numerical agreement remained near machine precision: maximum relative Hessian/RHS differences were 1.64e-15 / 2.81e-15 for one thread and 1.13e-15 / 3.25e-16 for ten threads.

The machine was busy during these runs, so the absolute numbers and ratios should be repeated on an otherwise quiet host. Directionally, both pipelines benefit similarly from threading; the remaining partial-versus-compact gap is primarily in non-threading assembly/elimination work.

dellaert:
## Updated `timeSFMBAL` benchmark: narrow single-batch QR override

- Head: `0fd5ab72ef`
- Runner: local macOS arm64, Release
- Scope: four configurations of interest only: regular vs `BatchFactor`, `MULTIFRONTAL_CHOLESKY` and `MULTIFRONTAL_SOLVER`, Schur ordering.
- Policy: `QRMode::Allow` still selects QR for ordinary eligible leaves, but a fully eliminated non-root leaf with one frontal block and exactly one directly updatable batch factor uses fused-star Cholesky.
- Caveat: single runs on a busy machine; absolute timings remain noisy.

### TBB=OFF

| Dataset | Regular Cholesky (s) | Regular Solver (s) | Batch Cholesky (s) | Batch Solver (s) | Cholesky Speedup | Solver Speedup |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `dubrovnik-16-22106-pre.txt` | 0.697 | 1.193 | 0.593 | **0.633** | 1.175x | **1.885x** |
| `dubrovnik-88-64298-pre.txt` | 3.958 | 6.504 | 2.440 | **1.935** | 1.622x | **3.360x** |
| `dubrovnik-135-90642-pre.txt` | 5.241 | 11.976 | 3.051 | **2.202** | 1.718x | **5.439x** |

### TBB=ON

| Dataset | Regular Cholesky (s) | Regular Solver (s) | Batch Cholesky (s) | Batch Solver (s) | Cholesky Speedup | Solver Speedup |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `dubrovnik-16-22106-pre.txt` | 0.588 | 0.902 | 0.499 | **0.247** | 1.178x | **3.656x** |
| `dubrovnik-88-64298-pre.txt` | 3.101 | 5.981 | 1.773 | **0.975** | 1.749x | **6.135x** |
| `dubrovnik-135-90642-pre.txt` | 3.415 | 7.495 | 2.327 | **1.349** | 1.468x | **5.558x** |

Validation passed:

- `make -j6 testMultifrontalSolver.run`
- `make -j6 testNonlinearMultifrontalSolver.run`
- `make -j6 testSfmPointBatchSchur.run`
- `git diff --check`

dellaert:
## Latest Dubrovnik-88 sanity timing

Pushed through `445289563f`, including restored bounded algebraic leaf merging and the internal batch-factor elimination adapter.

The adapter now owns the solver-only scatter/mapped-Hessian interface, and `FactorLoadPlan` stores typed local, parent, and separator mappings instead of six parallel slot/offset vectors. The fixed-size kernels and virtual declaration order are unchanged; the adapter calls are inline.

One Release sanity run on local arm64 (`dubrovnik-88-64298-pre.txt`, 88 cameras / 64,298 points, lambda 0.04, leaf merge 256, bounded leaf aggregation 2048, merge cap 32, 10 threads, 1 warmup + 7 alternating repetitions):

| Pipeline | Median ms | Mean ms | Min ms | Max ms |
| --- | ---: | ---: | ---: | ---: |
| Compact complete assembly | 74.621 | 75.319 | 73.068 | 82.480 |
| Partial fused end-to-end | **69.677** | 69.958 | 68.870 | 71.704 |
| Partial elimination only | 69.350 | 69.626 | 68.536 | 71.365 |
| Retained-factor export only | 0.334 | 0.332 | 0.322 | 0.339 |
| Separate load + eliminate (diagnostic) | 81.272 | 80.271 | 77.739 | 82.762 |
| Separate load | 9.535 | 9.545 | 9.358 | 9.697 |
| Separate eliminate | 71.577 | 70.726 | 68.269 | 73.318 |

- Primary partial end-to-end / compact median ratio: **0.934x**
- Elimination-only / compact diagnostic ratio: **0.929x**
- Maximum relative Hessian difference: `1.509e-15`
- Maximum relative RHS difference: `2.704e-16`
- Symbolic setup, reported separately: 319.199 ms

Machine load averages were 1.93 / 2.41 / 2.46 at launch. This is a sanity run rather than a controlled isolated-host comparison, but it shows no performance regression from the internal API cleanup.

Validation:

- `make -j6 testBatchJacobianFactor.run`
- `make -j6 testMultifrontalSolver.run`
- `make -j6 testNonlinearMultifrontalSolver.run`
- `make -j6 testSfmPointBatchSchur.run`
- `git diff --check`

PR #2658, opened 2026-08-10
Backport 32-bit Python key fixes to GTSAM 4.2

ProfFan:
## Summary

- Backport the Key/FactorIndex wrapper corrections from #2654 and complete the older release/4.2 `size_t` audit.
- Preserve 64-bit keys in `JacobianFactor` when `Eigen::Index` is 32-bit.
- Make the 4.2 split-wrapper `KeyVector` and `DiscreteValues` casters consistent.
- Add C++ and Python regressions using keys and indices above 32 bits.

Fixes #2089.

## Testing

- Debian bookworm i386 (GCC 12, Python 3.11), Release: built `gtsam_py` with a minimal Python-enabled configuration; verified the extension is ELF 32-bit.
- i386: `testLabeledSymbol` passed; 6 focused key-width checks passed; 218 applicable Python tests passed (4 skipped).
- Native x86_64 (GCC 16, Python 3.14), Release: rebuilt `gtsam_py`; `testLabeledSymbol`, large-key caster smoke checks, and the Metis-only sum-product test passed.

## Known unrelated 4.2/i386 baseline issue

- Excluded `test_ShonanAveraging.test_initializeWithDescent`: on i386, the 4.2 eigensolver returns exactly zero and its line search starts at infinity, causing an existing infinite loop. The other nine Shonan tests pass.

PR #2699, opened 2026-08-16
Align Python wrapper declarations with C++ APIs

dellaert:
## Summary

This aligns the wrapper declarations with the public C++ APIs and exposes those signatures directly in Python:

- Restore complete signatures, trailing defaults, exact return types, references, constness, and parameter types across the geometry, graph, Bayes tree/net, nonlinear, navigation, SFM, and unstable interfaces.
- Model the C++ inheritance hierarchy in `.i` files so inherited graph, elimination, marginal, joint, and container methods are declared once on their actual base classes.
- Add pybind casters for `OptionalJacobian` and mutable Eigen matrix/vector arguments. Python callers can pass writable NumPy arrays for optional Jacobians and output parameters, while existing calls continue to use `None` defaults.
- Register supporting types and concrete specializations needed by the exact APIs, including assignment maps, ISAM2 optimization parameter alternatives, Bayes-tree specializations, constrained optimizer parameters, and smart-factor linearization results.
- Add small C++ aliases and overloads where a named or concrete type is required by the wrapper, and add regression coverage for the newly exposed signatures and return types.
- Remove deprecated PoseRTV and PartialPriorFactor wrapper surfaces and dependent examples/tests while retaining the C++ implementations and native tests.

The `wrap/` dependency is unchanged from `develop`.

## Python API impact

Existing reduced-arity calls remain valid through C++ defaults. Callers additionally gain access to the complete optional arguments, writable Jacobians/output matrices, inherited base methods, and authoritative C++ return types.

## Validation

- Stable Python: **448 passed, 4 skipped, 6 subtests passed**
- Unstable Python: **4 passed, 6 subtests passed**
- Stable and unstable MATLAB MEX targets both build successfully.

## Outstanding MATLAB generator issues

Both MATLAB MEX targets compile, but runtime testing exposes several generator-level gaps that likely require changes in `wrap`:

- Templated bases are emitted literally in MATLAB `classdef` inheritance, for example `GaussianFactorGraph < gtsam.EliminateableFactorGraph<...>`, which MATLAB cannot parse.
- Fixed-size Eigen aliases can be emitted as wrapper objects such as `gtsam.Matrix3` instead of numeric matrices.
- Some generated overload dispatch does not match valid calls such as `insertProjectionFactors`.
- Internal C++ smart-pointer spellings can leak into MATLAB as names such as `std.shared_ptrNonlinearFactor`.

Seventeen stable MATLAB test scripts pass. The remaining failures are `testSFMExample`, `testUtilities`, and `testNumericalDerivative`; the unstable `testTSAMFactors` smoke test reaches the leaked smart-pointer-name issue above.


gtsam/base/base.i
```diff
@@ -86,7 +86,8 @@ class IndexPairSetMap {
 
 #include <gtsam/base/Matrix.h>
 #include <gtsam/base/MatrixSerialization.h>
-bool linear_independent(gtsam::Matrix A, gtsam::Matrix B, double tol);
+bool linear_independent(const gtsam::Matrix& A, const gtsam::Matrix& B,
+                        double tol = 1e-9);
```
ProfFan, line 90: Check that all the defaults are the same with the corresponding header files

gtsam/inference/inference.i
```diff
@@ -27,61 +27,61 @@ void PrintKeySet(
...
 };
 
-gtsam::Key symbol(char chr, uint64_t index);
+gtsam::Key symbol(char chr, size_t index);
 char symbolChr(gtsam::Key key);
-uint64_t symbolIndex(gtsam::Key key);
+size_t symbolIndex(gtsam::Key key);
 
 namespace symbol_shorthand {
-gtsam::Key A(uint64_t j);
```
ProfFan, line 47: Check if all the keys are using `gtsam::Key` as the type, check PR #2658 and #2654

gtsam/inference/inference.i
```diff
@@ -27,61 +27,61 @@ void PrintKeySet(
...
   LabeledSymbol(gtsam::Key full_key);
   LabeledSymbol(const gtsam::LabeledSymbol& key);
-  LabeledSymbol(unsigned char valType, unsigned char label, uint64_t j);
+  LabeledSymbol(unsigned char valType, unsigned char label, size_t j);
 
   gtsam::Key key() const;
   unsigned char label() const;
   unsigned char chr() const;
-  uint64_t index() const;
+  size_t index() const;
```
ProfFan: this can exceed 32 bits

gtsam/linear/GaussianConditional.h
```diff
@@ -243,7 +243,7 @@ namespace gtsam {
     constABlock S(const_iterator it) const { return BaseFactor::getA(it); }
 
     /** Get a view of the r.h.s. vector d */
-    const constBVector d() const { return BaseFactor::getb(); }
+    Vector d() const { return BaseFactor::getb(); }
```
ProfFan: double check if this is a valid change

gtsam/inference/Factor.h
```diff
@@ -29,6 +29,7 @@
 #include <gtsam/inference/Key.h>
 
 #include <algorithm>
+#include <functional>
```
ProfFan: It's better that we don't have <functional> in a top-level header

gtsam/linear/GaussianFactorGraph.cpp
```diff
@@ -206,7 +206,7 @@ namespace gtsam {
           }
       }
 
-      JacobianFactor::constBVector whitenedb(whitened.getb());
+      const Vector whitenedb = whitened.getb();
```
ProfFan: double check this

gtsam/linear/HessianFactor.h
```diff
@@ -264,12 +264,12 @@ namespace gtsam {
 
     /** Return the complete linear term \f$ g \f$ as described above.
      * @return The linear term \f$ g \f$ */
-    SymmetricBlockMatrix::constBlock linearTerm() const {
+    Matrix linearTerm() const {
```
ProfFan: this seems wrong

gtsam/linear/JacobianFactor.h
```diff
@@ -341,13 +341,13 @@ namespace gtsam {
     SharedDiagonal& get_model() { return model_;  }
 
     /** Get a view of the r.h.s. vector b, not weighted by noise */
-    const constBVector getb() const { return Ab_(size()).col(0); }
+    Vector getb() const { return Vector(Ab_(size()).col(0)); }
```
ProfFan: check similar stuff

gtsam/linear/JacobianFactor.h
```diff
@@ -341,13 +341,13 @@ namespace gtsam {
...
 
     /** Get a view of the A matrix for the variable pointed to by the given key iterator */
     constABlock getA(const_iterator variable) const { return Ab_(variable - begin()); }
 
     /** Get a view of the A matrix, not weighted by noise */
-    constABlock getA() const { return Ab_.range(0, size()); }
+    Matrix getA() const { return Matrix(Ab_.range(0, size())); }
```
ProfFan: We need a mechanism in Python and MATLAB to access these without changing the header itself, potentially `Eigen::Ref`

dellaert:
@ProfFan is the 32-bit issue addressed in my latest commits? I also addressed the "matrix view" issue in `linear`, although to do it without a lambda it will probably need change in wrap.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
```diff
@@ -242,6 +242,17 @@ Values RiemannianStaircaseOptimizer::padInitialValues(const Values& Y,
   return padded;
 }
 
+/* ************************************************************************* */
+Values RiemannianStaircaseOptimizer::padInitialValues(const Values& Y,
```
ProfFan: what is this one for?

gtsam/nonlinear/nonlinear.i
```diff
@@ -831,11 +849,13 @@ virtual class ExtendedPriorFactor : gtsam::NoiseModelFactor {
...
                       const gtsam::Matrix& covariance);
-  T origin() const;
+  const T& origin() const;
   // Optional tangent space mean (may be empty / None)
-  std::optional<gtsam::Vector> mean() const;
-  std::optional<gtsam::Matrix> covariance() const;
-  std::optional<std::pair<gtsam::Matrix, gtsam::Matrix>> gaussian() const;
+  const std::optional<gtsam::Vector>& mean() const;
+  std::optional<gtsam::Matrix> covariance(
+      const string& method = "<unknown>", bool throwOnFailure = false) const;
```
ProfFan, line 856: double check this

ProfFan (approved):

PR #2702, opened 2026-08-17
Enhance EquivariantFilter with error coordinate handling and matrix output

arihantb2:
### Files changed

**gtsam/navigation/EquivariantFilter.h**: Fixes the congruence direction in `covariance()`; factors the action differential out as a public `actionDifferential()` accessor; documents which tangent space the filter's stored quantities live in.

**gtsam/navigation/tests/testEquivariantFilter.cpp**: Replaces a self-fulfilling covariance test with three independent tests; fixes a clobbered Jacobian in the test fixture's `MeasurementFunctor`.

### Issues identified and fixed

1. **`covariance()` applied the congruence in the wrong direction.** It returned `Jᵀ P J`, the pullback, where the pushforward `J P Jᵀ` is required. The returned matrix was the covariance transported away from the current state rather than to it.

2. **The defect was not detectable by the existing test**, which recomputed the implementation's own expression.

3. **The test fixture's `MeasurementFunctor` returned a Jacobian missing a factor `c_m`**, because `Unit3::point3(H)` overwrites the Jacobian that had just been written.

### Background: which tangent space the filter lives in

The EqF keeps a reference (origin) state `xi_ref`, a group estimate `g`, and recovers the state estimate as `xi_hat = phi_g(xi_ref)`. The true state is expressed through the error `e`:

```
e   = Retract(xi_ref, epsilon)      # error, near the origin
xi  = phi_g(e)                      # true state
J   = D phi_g |_{xi_ref}            # T_{xi_ref}M  ->  T_{xi_hat}M
```

The covariance the filter propagates is in **error coordinates**, i.e. the tangent space at `xi_ref`, not at `xi_hat`. This is not an assumption; it is forced by the code:

- `InnovationLift_ = pinv(Dphi0_)` where `Dphi0_` is the action differential **at the reference state**. It maps `T_{xi_ref}M -> g`, so the `delta_xi` fed into it must live at the reference.
- `computeErrorDynamicsMatrix()` returns `Dphi0_ * D_lift` with `D_lift` evaluated at `xi_ref_`, again the origin frame.

So `P_` is the covariance of `epsilon`. A perturbation `epsilon` at the reference appears at the current state as `J * epsilon`, and the covariance pushes forward as `J P Jᵀ`.

### Issue 1 — covariance transported the wrong way

**Before**: `return J.transpose() * this->P_ * J;` (pullback)

**After** (`EquivariantFilter.h:130-159`):
```cpp
/// J = D phi_g|_{xi_ref}, the bridge between error coordinates and the
/// tangent space at the current state.
MatrixM actionDifferential() const { ... }

CovarianceM covariance() const {
  const MatrixM J = actionDifferential();
  return J * this->P_ * J.transpose();
}
```

`actionDifferential()` is factored out as a public accessor so callers (and the tests) can move between the two frames without recomputing the action differential by hand.

The class documentation now states the frame contract for the quantities the filter propagates — `P`, the error dynamics matrix `A` and the process noise `Qc` — and points at `covariance()` / `actionDifferential()` for the map to the current state.

### Issue 2 — the test could not have caught the defect

The deleted `CovarianceRotation` test recomputed the implementation's expression verbatim, so it passed for whichever congruence direction the implementation happened to use. Its covariance was isotropic anyway (`Sigma0 = 0.01 * I` and `Qc * dt = 0.002 * I` give `P = 0.012 * I`) and with an orthogonal `J`, `J P Jᵀ = Jᵀ P J` identically.

**Replacement tests** — new fixture namespace `covariance_transport` (`testEquivariantFilter.cpp:356-421`). All three use an anisotropic, correlated prior and a non-identity group estimate, and derive their expectation only from the state relation `eta = phi_g(Retract(eta_ref, epsilon))`, never from the filter's own Jacobian:

- `CovarianceAtIdentity` — with `g = e` the error coordinates already sit at the current state, so `covariance()` must return `P` unchanged.

- `CovariancePushforward` — takes `D` by `numericalDerivative11` of `statePerturbation` and expects `D * Sigma * Dᵀ`. It carries an anti-vacuity guard so the test cannot silently lose its discriminating power if the fixture constants are ever edited.

- `CovarianceMonteCarlo` — 200 000 samples `epsilon ~ N(0, Sigma)` pushed through `statePerturbation`; the sample covariance must match `covariance()` to `1e-5` (sampling error ≈ 1.3e-6).

### Issue 3 — clobbered Jacobian in the test fixture

**Before**:
```cpp
if (H) {
  *H = c_m_ * eta_hat.basis();   // dead store
}
return c_m_ * eta_hat.point3(H); // Unit3::point3 does *H = basis(), overwriting it
```

`Unit3::point3` writes its own Jacobian unconditionally, so the functor returned the correct value with a Jacobian short by `c_m`.

**After**:
```cpp
const Point3 direction = eta_hat.point3(H);
if (H) *H *= c_m_;
return c_m_ * direction;
```

### Compatibility

- `covariance()` now returns a different matrix. The previous value was incorrect; `errorCovariance()`, which returns `P_` in error coordinates, is unchanged. There are no in-tree callers of `EquivariantFilter::covariance()` outside the tests.
- `EquivariantFilter` has no `.i` wrapper declaration, so no wrapper or Python changes are needed.

### Testing

Build and tests pass.


dellaert:
Thanks for addressing the covariance transpose and the `Point3` Jacobian issue—those are real fixes.

I’m less convinced by the conclusion that `update()` is supposed to require an EqF `C*` matrix. When I designed this API, I deliberately wanted it to look and behave like the ordinary GTSAM/`ManifoldEKF` API: users provide the Jacobian they naturally think about, namely the measurement Jacobian with respect to tangent coordinates at the current state estimate.

This was intentional in [`a091e61b94`](https://github.com/borglab/gtsam/commit/a091e61b94539389557fd67a05904a50b5e51e04), whose title is “Bring API in line with ManifoldEKF.” In particular, the model-function overload calls:

```cpp
Z prediction = h(this->X_, H);
```

so `H` is naturally the ordinary GTSAM Jacobian at the current state. The API is deliberately aligned with `ManifoldEKF`, rather than exposing the internal EqF error-coordinate representation.

Internally, the filter can convert this Jacobian into the matrix needed by the EqF equations.

arihantb2:
Hi Frank, thank you for the comments and the context for the update API. I will revert the update interface changes and the PR shortly.

arihantb2:
Hi Frank, I had claude revert the update() API changes. The changes are now strictly fixing the covariance function. PR is ready for review.

dellaert (approved):
Nice, thanks ! @rohan-bansal we should probably re-run some experiments :-)

dellaert:
@rohan-bansal Were you able to rerun those experiments?

rohan-bansal:
@dellaert i was not able to do it yet, caught up in first week class chaos. 🥲 will put it on top of my priority list for tonight if urgent.

PR #2707, opened 2026-08-18
wrap: declare navigation factors under NoiseModelFactor

DLuminary:
Thirty factors in navigation.i are declared with gtsam::NonlinearFactor as their parent, but every one of them derives from NoiseModelFactorN or NoiseModelFactorT in C++. Because the wrapper registers the shallower parent, the NoiseModelFactor interface -- noiseModel(), unwhitenedError() and whitenedError() -- is unreachable from Python on these factors.

This affects GPSFactor and its five variants, BarometricFactor, ImuFactor, ImuFactor2, ImuFactorWithGravityT, AHRSFactor, ConstantVelocityFactor, MagFactor, MagFactor1, and the sixteen Pseudorange/CarrierPhase/Doppler factors. Only CombinedImuFactor, CombinedImuFactorWithGravityT, AttitudeFactor and MagPoseFactor were declared correctly.

The practical consequence is that no chi-square style gate or residual diagnostic can be computed from a GNSS, barometer or IMU factor in Python, because the noise model cannot be reached. slam.i already uses NoiseModelFactor throughout (25 declarations to 3); this brings navigation.i in line.

The Pseudorange and CarrierPhase factors additionally inherit privately from PseudorangeBase / CarrierPhaseBase, which is not a registered type, so NoiseModelFactor remains the single correct wrapper parent.

Generated the pybind sources before and after: every changed line is a py::class_ parent declaration, with no other delta.

### DIFF
--- a/gtsam/navigation/navigation.i
+++ b/gtsam/navigation/navigation.i
@@ -270,7 +270,7 @@ class PreintegratedImuMeasurements {
   void serialize() const;
 };
 
-virtual class ImuFactor: gtsam::NonlinearFactor {
+virtual class ImuFactor: gtsam::NoiseModelFactor {
   ImuFactor(gtsam::Key pose_i, gtsam::Key vel_i, gtsam::Key pose_j, gtsam::Key vel_j,
       gtsam::Key bias,
       const gtsam::PreintegratedImuMeasurements& preintegratedMeasurements);
@@ -286,7 +286,7 @@ virtual class ImuFactor: gtsam::NonlinearFactor {
   void serialize() const;
 };
 
-virtual class ImuFactor2: gtsam::NonlinearFactor {
+virtual class ImuFactor2: gtsam::NoiseModelFactor {
   ImuFactor2();
   ImuFactor2(gtsam::Key state_i, gtsam::Key state_j,
       gtsam::Key bias,
@@ -307,7 +307,7 @@ virtual class ImuFactor2: gtsam::NonlinearFactor {
 
 #include <gtsam/navigation/ImuFactorWithGravity.h>
 template <PIM, GRAVITY>
-virtual class ImuFactorWithGravityT : gtsam::NonlinearFactor {
+virtual class ImuFactorWithGravityT : gtsam::NoiseModelFactor {
   ImuFactorWithGravityT(gtsam::Key pose_i, gtsam::Key vel_i,
       gtsam::Key pose_j, gtsam::Key vel_j, gtsam::Key bias, gtsam::Key gravity,
       const PIM& preintegratedMeasurements);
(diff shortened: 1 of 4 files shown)


dellaert (approved):
Nice ! Does it capture all such cases? Would it be worth expanding this PR to be comprehensive?

DLuminary:
Good question — it captured all of navigation.i, but not the whole repo. I enumerated all 23 .i files in the repo and pushed a second commit. 48 classes declare gtsam::NonlinearFactor as parent; 36 are wrong. navigation.i has 30 (this PR's first commit) and gtsam_unstable.i has 6 (second commit). The second commit is covering six more in gtsam_unstable.i: TOAFactor (via ExpressionFactor<double>) and VelocityConstraint3, PendulumFactor1, PendulumFactor2, PendulumFactorPk, PendulumFactorPk1 (all NoiseModelFactorT). That's 36 in total.

The other 12 are correct and I've left them: SmartFactorBase, SmartProjectionPoseFactor, SmartStereoProjectionFactor, KarcherMeanFactor, LinearContainerFactor, QpCost, BetweenFactorEM, TransformBtwRobotsUnaryFactor{,EM}, DummyFactor — all derive from NonlinearFactor directly — plus NoiseModelFactor itself and ExpressionFactorGraph, which isn't a factor.

dellaert:
Please check CI failure

DLuminary:
CI caught something pre-existing rather than from this PR — pushed a commit switching these tests to whitenedError.

whitenedError(values) works fine on the reparented GPSFactor, so the base class change is doing its job. unwhitenedError(values) raises TypeError: incompatible function arguments, listing H: Sequence[...] = None as the only overload. The difference is the gtsam::OptionalMatrixVecType H = nullptr parameter.

Root cause: gtsam.tpl gives each .i its own translation unit including only its own preamble, and PYBIND11_MAKE_OPAQUE(std::vector<gtsam::Matrix>) is declared only in preamble/custom.h. So in nonlinear.cpp the type falls back to the stl.h list caster, which can't load None — the nullptr default won't round-trip, and the argument can't be omitted. It's also an ODR inconsistency, since the type is opaque in custom.cpp and transparent elsewhere in the same module.

The same shape appears on three whitenedError declarations in gtsam_unstable.i. No Python test called unwhitenedError or whitenedError before this one — I checked back through the history and the last references are from the pre-pybind wrapper — which is presumably why it went unnoticed.

The fix looks like one line in each of preamble/nonlinear.h and python/gtsam_unstable/preamble.h; py::bind_vector already registers JacobianVector and the registry is module-wide. Happy to open that as a separate PR if useful.

These four declarations (one in nonlinear.i, three in gtsam_unstable.i) gained the H parameter in #2699 — OptionalMatrixVecType doesn't appear in any .i file before it. The preamble gap predates that; #2699 is just what made it reachable, since nothing outside the custom module referenced the type until then.

dellaert:
Yeah, maybe exclude the problematic classes and add a separate PR for those? Or an issue - if you're not ready to create a PR just yet.

DLuminary:
Added a potential fix in #2712

PR #2708, opened 2026-08-18
wrap: expose adaptive reorder settings and Bayes tree fill-in

DLuminary:
The adaptive reorder heuristic added in #2509 is unreachable from Python. ISAM2Params::enableAdaptiveReorder defaults to false and neither it nor adaptiveReorderThreshold is declared in the wrapper, so the feature cannot be switched on at all from Python.

Meanwhile ISAM2::update() computes result.treeNnz unconditionally on every update -- it is not gated by enableDetailedResults or anything else -- so Python callers already pay for the full nnz_internal walk of the Bayes tree on every update while having no way to read the result.

Declares, all of which already exist in C++:
  - ISAM2Params::enableAdaptiveReorder, ISAM2Params::adaptiveReorderThreshold
  - ISAM2Result::getTreeNnz(), ISAM2Result::getBatchReorderTriggered()
  - ISAM2::treeNnz(), which is public but was never wrapped

Together these make fill-in growth observable over long incremental sessions, which is what the treeNnz doc comment describes, and let the adaptive reorder heuristic be enabled and its effect confirmed.

### DIFF
--- a/gtsam/nonlinear/nonlinear.i
+++ b/gtsam/nonlinear/nonlinear.i
@@ -492,6 +492,8 @@ class ISAM2Params {
   bool enableDetailedResults;
   bool enablePartialRelinearizationCheck;
   bool findUnusedFactorSlots;
+  bool enableAdaptiveReorder;
+  double adaptiveReorderThreshold;
 
   enum Factorization { CHOLESKY, QR };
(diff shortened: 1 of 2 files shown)


dellaert (approved):
Nice, Thanks !

PR #2709, opened 2026-08-18
wrap: complete targeted single-key calculateEstimate coverage

DLuminary:
IncrementalFixedLagSmoother::calculateEstimate<VALUE>(Key) exists in C++ and its own doc comment notes it is faster than the no-argument version, but it was never declared in the wrapper. The only way to read a single key from an IncrementalFixedLagSmoother from Python was to reach through getISAM2() and call the accessor on the ISAM2 instance. This declares the templated accessor directly, mirroring the list of the ISAM2 method it forwards to.

Also fills in three missing instantiations:

  - `double` on ISAM2 and BatchFixedLagSmoother. Systems that estimate genuinely scalar states (barometer bias, clock bias) otherwise have to fall back to calculateEstimate() and read the key out of the full Values, which is an O(window) recompute for one scalar. double is already wrappable here: PriorFactor's list starts with it and Values exposes atDouble.

  - `gtsam::NavState` on both. Values already carries NavState in its at<T>, insert and update lists, and ExtendedKalmanFilter already instantiates on it, so no targeted NavState read was available on either smoother despite the type being fully supported elsewhere.

  - `gtsam::imuBias::ConstantBias` on BatchFixedLagSmoother, which ISAM2 already had. No IMU-based system using the batch smoother could pull a single bias by key.

### DIFF
--- a/gtsam/nonlinear/nonlinear.i
+++ b/gtsam/nonlinear/nonlinear.i
@@ -589,13 +589,15 @@ class ISAM2 {
   const gtsam::Values& getLinearizationPoint() const;
   bool valueExists(gtsam::Key key) const;
   gtsam::Values calculateEstimate() const;
-  template <VALUE = {gtsam::Point2,
+  template <VALUE = {double,
+                     gtsam::Point2,
                      gtsam::Rot2,
                      gtsam::Pose2,
                      gtsam::Point3,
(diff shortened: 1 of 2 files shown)


dellaert (approved):
Nice !

dellaert:
Please check CI failure

PR #2710, opened 2026-08-18
Generalize fixed-size Jacobian factors to arbitrary arity

### LINKED ISSUE #2667: Variadic templated Jacobian factors
Extend #2652 to any arity for fixed types/error, in FixedJacobianFactor.h, replacing bespoke binary and ternary templated versions. Will need to assess trade-off between compile time and run-time win.

dellaert:
## Summary

- generalize `FixedJacobianFactor` to `template<int M, int... Ns>` so fixed-size factors support any positive arity
- remove the obsolete `BinaryJacobianFactor` and `TernaryJacobianFactor` types and headers, and migrate all in-tree consumers to `FixedJacobianFactor`
- select the fixed-size linear factor automatically for any all-fixed `NoiseModelFactorN`
- document every public `FixedJacobianFactor` API following the repository's Doxygen guidance
- add unary/quaternary correctness coverage, retain focused binary/ternary-arity coverage under the fixed-factor name, and add a reusable timing benchmark

## Why

#2652 introduced fixed-size Jacobian factors through bespoke binary and ternary implementations. Extending the implementation variadically removes the duplicated arity-specific code while retaining fixed-size residual, diagonal-Hessian, whole-Hessian, and ranged-Hessian paths.

The legacy names and headers are intentionally removed rather than retained as compatibility aliases. Downstream code should include `FixedJacobianFactor.h` and use `FixedJacobianFactor<M, Ns...>`.

## Performance and compile-time trade-off

On the included four-variable benchmark (9 alternating trials, Release build), the fixed-size path improved median time versus `JacobianFactor` by:

- linearization: 23.8%
- residual update: 79.5%
- diagonal Hessian update: 11.5%
- whole Hessian update: 68.1%
- ranged Hessian update: 67.6%

Two order-balanced clean library build pairs measured a 3.8% wall-time / 3.6% user-time compile cost. The resulting shared library was effectively unchanged (-0.045%). A synthetic translation unit showed the expected template scaling: median compile time increased from 1.40 s at arity 1 to 1.86 s at arity 8.

## CI fix

The first PR run exposed a GCC 11 `-Werror=unused-but-set-parameter` failure from instantiating the pairwise cross-Hessian helpers with an empty pack for the final variable. The variadic dispatch now visits only the first `arity - 1` blocks, preserving all pairwise products without creating the empty instantiation. Unary whole/ranged Hessian coverage exercises the zero-pair case.

## Validation

- clean warning-as-error Release build of `gtsam` and `gtsam_unstable`
- `make -j6 testFixedJacobianFactor.run`
- `make -j6 testFixedJacobianFactorBinary.run`
- `make -j6 testFixedJacobianFactorTernary.run`
- affected navigation tests: IMU, constant velocity, planar gyro, and Doppler
- affected SAM tests: range, bearing, and bearing-range
- affected SLAM tests: between, projection, stereo, general SFM, and essential-matrix constraint
- affected unstable tests: simple helicopter and PPP projection
- all four affected timing targets compile
- `make -j6 doc` completes; no warnings originate from `FixedJacobianFactor.h`
- `git diff --check`

Fixes #2667


dellaert:
@ProfFan happy to address comments ina follow-up PR, but you and I already discussed this - and timing shows it's worth it.

PR #2711, opened 2026-08-18
wrap: expose noise model introspection and Robust accessors

DLuminary:
On a stock build the entire Python surface of a robust noise model is:

```python
>>> [x for x in dir(robust) if not x.startswith('_')]
['Create', 'deserialize', 'print', 'serialize']
```

and a `Diagonal` has `sigmas`, `covariance` and `whiten` but no `dim`,
`isUnit` or `isConstrained`. `NoiseModelFactor::noiseModel()` is wrapped, so
from Python you can reach the `Base*` and then go no further.

### The commented-out block

`linear.i` says these methods can't be added "because wrap (incorrectly)
thinks robust classes derive from this Base as well." That comment dates from
`1d6392dc` (Aug 2020, the initial pybind migration) and moved to `linear.i` in
`9bafebb5` without being revisited. It no longer reproduces.

Generated the wrappers with the vendored `wrap`, before and after:

- **pybind**: the only delta is the new `.def` lines on
  `gtsam::noiseModel::Base`. All thirteen `mEstimator` classes still emit as
  `py::class_<gtsam::noiseModel::mEstimator::X, gtsam::noiseModel::mEstimator::Base, ...>`.
- **MATLAB**: `mEstimator` classes still emit
  `classdef X < gtsam.noiseModel.mEstimator.Base`, and the new methods appear
  as `gtsamnoiseModelBase_dim_25` etc. against the correct class.

No contamination on either path. I couldn't find the commit that fixed it —
the comment predates most of `wrap`'s history.

### Why more than the four commented-out methods

`squaredMahalanobisDistance` and `unweightedWhiten` are the two methods
`Robust` overrides specifically to bypass the m-estimator re-weighting.
Exposing them on `Base` means a chi-square style gate can be computed
polymorphically off `NoiseModelFactor::noiseModel()` without the caller
knowing whether a factor is robust-wrapped — including for factors it didn't
construct. `whiten()` on a robust model is silently down-weighted, so a gate
built on it understates a gross fault.

`robust()` and `noise()` are included for introspection. They generate
identically to the existing `Sampler::model()` precedent in the same file:
`-> const auto&` with `return_value_policy::reference_internal`.

Parameter names match the headers exactly, per DEVELOP.md.

### Second commit

`Base::sigmas()` used `throw("...")`, which throws a `const char*` rather than
a `std::exception`. `Robust` doesn't override `sigmas()` — it has none of its
own — so with the method now reachable, the fallback surfaced as
`RuntimeError: Caught an unknown exception!` with the message discarded.

It's the only such throw in the file; the other six use
`throw invalid_argument(...)` and `NoiseModel.h` uses `std::runtime_error` and
`std::invalid_argument`. Split into its own commit so it can be dropped
independently.

### DIFF
--- a/gtsam/linear/NoiseModel.cpp
+++ b/gtsam/linear/NoiseModel.cpp
@@ -71,7 +71,7 @@ std::optional<Vector> checkIfDiagonal(const Matrix& M) {
 
 /* ************************************************************************* */
 Vector Base::sigmas() const {
-  throw("Base::sigmas: sigmas() not implemented for this noise model");
+  throw runtime_error("Base::sigmas: sigmas() not implemented for this noise model");
 }
 
 /* ************************************************************************* */
--- a/gtsam/linear/linear.i
+++ b/gtsam/linear/linear.i
@@ -33,12 +33,27 @@ namespace noiseModel {
 #include <gtsam/linear/NoiseModel.h>
 virtual class Base {
   void print(string s = "") const;
-  // Methods below are available for all noise models. However, can't add them
-  // because wrap (incorrectly) thinks robust classes derive from this Base as well.
-  // bool isConstrained() const;
-  // bool isUnit() const;
-  // size_t dim() const;
-  // gtsam::Vector sigmas() const;
+
+  // Introspection, available on every noise model.
+  bool isConstrained() const;
+  bool isUnit() const;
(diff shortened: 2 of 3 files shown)


dellaert (approved):
Awesome !

PR #2712, opened 2026-08-18
Make JacobianVector opaque for every module fragment

DLuminary:
Confirmed with a minimal pybind11 module built twice from identical source
with only `PYBIND11_MAKE_OPAQUE` toggled:

    ===== transparent =====
      f.unwhitenedError(1)        -> TypeError: incompatible function arguments
      f.unwhitenedError(1, None)  -> TypeError: incompatible function arguments
      f.whitenedError(1)          -> ok          (no optional arg)

    ===== opaque =====
      f.unwhitenedError(1)        -> no H
      f.unwhitenedError(1, None)  -> no H
      f.whitenedError(1)          -> ok

The opaque build also populates `H` when one is passed. I also compile-tested
the macro in this exact position — global scope, ahead of the
`pybind11::detail` block that follows it in this header.

This moves the declaration rather than adding it to two more preambles,
mirroring how the `OptionalJacobian` caster from #2699 is already handled —
that one lives in this header for the same reason. Happy to do the narrower
version instead (add `PYBIND11_MAKE_OPAQUE` to `preamble/nonlinear.h` and
`gtsam_unstable/preamble.h`, leave `custom.h` alone) if you'd rather not move
a declaration `custom.cpp` currently depends on.

Surfaced by the CI run on #2707.

### DIFF
--- a/python/gtsam/optional_jacobian_pybind.h
+++ b/python/gtsam/optional_jacobian_pybind.h
@@ -6,6 +6,22 @@
 #include <gtsam/base/OptionalJacobian.h>
 
 #include <memory>
+#include <vector>
+
+/* gtsam::OptionalMatrixVecType is std::vector<gtsam::Matrix>*, an input/output
+ * argument used by NoiseModelFactor::unwhitenedError and by CustomFactor's
+ * error callback. It must be opaque so that mutations on the Python side are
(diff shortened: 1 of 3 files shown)


dellaert (approved):
Thanks!!

Issue #2683, opened 2026-08-13
Compile a rank-independent QCQP representation for BM staircases


dellaert:
## Motivation

The Riemannian staircase currently constructs a new `QcqpProblem(graph, p)` at every visited rank. PR #2675 makes each construction linear and reuses the constructor-built `pMin` problem, but advancing the staircase still calls every source factor's `qcqpFactors(..., p)` again and rebuilds equivalent rank-independent coefficients and constraints.

For matrix-valued BM problems, most of the mathematical representation is independent of the staircase rank `p`:

- variable keys and row dimensions;
- quadratic cost block coefficients;
- quadratic constraint `(A, b, sense, sigma)` data;
- constraint identity/order and source-factor diagnostics.

Only the runtime matrix values and their column dimension change with `p`.

## Proposal

Introduce a compiled, rank-independent QCQP representation between the source `NonlinearFactorGraph` and rank-specific optimizer adapters.

Conceptually:

```text
NonlinearFactorGraph
    -> compile once
CompiledQcqp
    -> instantiate/view at p
QcqpProblem(p)
```

The compiled form should own or reference the row-space quadratic data once. Creating the rank-`p` problem should be cheap and should not call the original source factors again or copy rank-independent dense coefficient matrices.

This should remain a general QCQP/BM facility. It must not encode rotation-averaging or Shonan behavior.

## Design questions

- Should this be a new `CompiledQcqp` type, or rank-independent storage directly owned by `QcqpProblem`?
- Should source factors expose a new rank-independent lowering API while `qcqpFactors(..., columnDimension)` remains as a compatibility path?
- Can `QpCost` and quadratic constraint factors operate as rank-parameterized views over shared row-space data?
- How should genuinely rank-dependent or nonquadratic constraints declare themselves and fall back without weakening fail-fast diagnostics?
- Which metadata should be retained to report the source factor responsible for an unsupported lowering?

## Acceptance criteria

- Lower every compatible source factor exactly once for an entire staircase run.
- Instantiate every visited rank without rescanning the source graph or duplicating rank-independent coefficient storage.
- Preserve cost and constraint order, exact quadratic deduplication semantics, objective values, violations, certificates, and existing unsupported-factor diagnostics.
- Demonstrate numerical equivalence across multiple ranks and at least two QCQP-lifted problem types.
- Add counting/storage tests proving compilation occurs once and rank instantiation does not relower source factors.
[...]

Issue #39, opened 2019-06-01
Travis fails but shows success


dellaert:
As an example, [this run]( https://travis-ci.com/borglab/gtsam/jobs/204686142) did not run the tests but failed (why?, was just a warning), but still shows up as a success.

dellaert:
Travis is awesome ! We'll get it right :-)

jlblancoc:
Hmm... I think the point is that Travis build #5, [here](https://travis-ci.com/borglab/gtsam/builds/113966855), spans into the matrix of jobs we defined and only one of jobs failed, [this one, 5.2](https://travis-ci.com/borglab/gtsam/jobs/204686143)... in this case due to a timeout (50 minutes) :man_facepalming: 

Any of the admins, could try hitting the "retry" button of [job 5.2](https://travis-ci.com/borglab/gtsam/jobs/204686143), since sometimes build times differ depending on the load of the server. 

Anyhow, it seems we'll have to use Travis CI's "stages" to overcome the time limit (sigh)

dellaert:
Fixed issues in pr  #45

PR #2713, opened 2026-08-18
Adding certifiable estimation applications

zhexin1904:
## Summary

Adds certifiable pose-graph optimization, landmark-SLAM and range-aided SLAM.
Also a few fixes in using ALM with Riemannian staircase optimizer.

## What changed

- Three new examples, `CertifiablePoseGraphOptimization`,
  `CertifiableLandmarkSLAM` and `CertifiableRangeAidedSLAM`.
- Two new factors, `RelativeTranslationFactor` for the SE(d) translation term and landmark measurement, 
  and `QuadraticRangeFactor` for range measurements with an auxiliary
  direction. Planar directions are `Rot2` and spatial ones `Unit3`.
- Riemannian staircase verification is reworked. 
- `leastSquaresMultipliers` recovers multipliers from the current point, so a
  certificate is no longer built from the all-zero multipliers ALM sometimes
  returns.
- `escapeSaddleAndLift` is renamed `liftWithDescent`, and an optional
  merit-guarded line search is added for the escape itself.
- Python wrappers and notebook examples for all three applications, plus
  reduced g2o and pyfg datasets under `examples/Data`.

## Results

Rotation averaging,

| dataset | init | rank | objective | solve | verify | round | other | total |
|---|---|---|---|---|---|---|---|---|
| MIT | fast-sync | 2->2 | 19.4055 | 0.101 | 0.0010 | 0.0003 | 0.011 | 0.104 |
| MIT | random | 2->4 | 19.4055 | 1.24 | 0.0358 | 0.0003 | 0.015 | 1.24 |
| intel | fast-sync | 2->2 | 1.81982 | 0.249 | 0.0023 | 0.0009 | 0.047 | 0.259 |
| intel | random | 2->4 | 1.81982 | 2.35 | 0.0912 | 0.0007 | 0.069 | 2.35 |
| CSAIL | fast-sync | 2->2 | 11.1566 | 0.111 | 0.0010 | 0.0004 | 0.019 | 0.115 |
| CSAIL | random | 2->2 | 11.1566 | 0.153 | 0.0010 | 0.0004 | 0.017 | 0.153 |
| sphere2500 | fast-sync | 3->3 | 442.681 | 1.71 | 0.0197 | 0.0021 | 0.258 | 1.74 |
| sphere2500 | random | 3->3 | 442.681 | 3.37 | 0.0201 | 0.0026 | 0.254 | 3.37 |

Pose-graph optimization:

| dataset | init | rank | objective | solve | verify | round | other | total |
|---|---|---|---|---|---|---|---|---|
| MIT | fast-sync | 2->2 | 30.5771 | 0.224 | 0.0016 | 0.0009 | 0.025 | 0.227 |
| MIT | random | 2->2 | 30.5771 | 0.286 | 0.0023 | 0.0007 | 0.025 | 0.287 |
| intel | fast-sync | 2->2 | 26.1742 | 0.497 | 0.0057 | 0.0018 | 0.111 | 0.507 |
| intel | random | 2->4 | 26.1744 | 5.36 | 0.315 | 0.0014 | 0.167 | 5.36 |
| CSAIL | fast-sync | 2->2 | 15.8519 | 0.205 | 0.0022 | 0.0012 | 0.039 | 0.209 |
| CSAIL | random | 2->2 | 15.8519 | 0.293 | 0.0022 | 0.0009 | 0.040 | 0.294 |
| sphere2500 | fast-sync | 3->3 | 843.508 | 3.58 | 0.0339 | 0.0043 | 0.475 | 3.62 |
| sphere2500 | random | 3->3 | 843.503 | 15.8 | 0.0326 | 0.0049 | 0.464 | 15.8 |

Landmark SLAM:

| dataset | init | rank | objective | solve | verify | round | QCQP build | total |
|---|---|---|---|---|---|---|---|---|
| cityTrees_1k | fast-sync | 2->2 | 23.7559 | 0.2568 | 0.0024 | 0.0006 | 0.0035 | 0.3025 |
| cityTrees_1k | random | 2->4 | 23.7562 | 14.62 | 0.0584 | 0.0009 | 0.0091 | 14.75 |
| victoriaPark_1k | fast-sync | 2->2 | 46.8794 | 0.2290 | 0.0026 | 0.0006 | 0.0039 | 0.2770 |
| victoriaPark_1k | random | 2->3 | 46.9118 | 4.431 | 0.0071 | 0.0006 | 0.0063 | 4.491 |

Range-aided SLAM (objective reported as the locally refined one):

| dataset | init | rank | objective | solve | verify | round | QCQP build | total |
|---|---|---|---|---|---|---|---|---|
| goats_15 | fast-sync | 2->3 | 18204.6 | 4.467 | 0.0255 | 0.0006 | 0.0051 | 4.542 |
| goats_15 | random | 2->3 | 18204.6 | 4.166 | 0.0245 | 0.0005 | 0.0055 | 4.243 |
| goats_16 | fast-sync | 2->3 | 3894.12 | 1.738 | 0.1104 | 0.0003 | 0.0030 | 1.868 |
| goats_16 | random | 2->3 | 3894.12 | 1.107 | 0.1153 | 0.0005 | 0.0031 | 1.242 |

## To do

- Performance enhancement, as in the issue #2683.
- More scalable inner local solver.
- Computational enhancement. Rotation-only/constrained-only formulation (like Smart-factor). LOBPCG-based eigenvalue solver for certification.


dellaert (approved):
NIce! Would be great if there was also an example notebook (and python wrapper). I already showed the way, I think it’s just mechanical.

dellaert:
And please check CI failures :-)

zhexin1904:
Hi Frank @dellaert , I have the landmark SLAM and range-aided SLAM application almost ready now. I plan to convert this PR back to a draft, add those commits, and then mark it as ready for review again.

dellaert:
FYI @avinashresearch1 - BM work in progress.

gtsam/geometry/Unit2.h
```diff
@@ -0,0 +1,163 @@
+/* ----------------------------------------------------------------------------
```
dellaert: I think this the same as Rot2 :-/

zhexin1904: Ahh conceptually I think so, just Stiefel manifold in the QCQP setting. I still have a few commits that I haven’t pushed yet.

dellaert: I think it's exactly the same manifold, so if you *can* make it work with Rot2 I'd prefer that. The alternative is to generalize Unit3 to Unit<n> -> spheres in R^n, with dim=n-1.

zhexin1904: Actually, we already have a complete implementation of the Stiefel manifold type, and the unit sphere is a special case: UnitSphere = StiefelManifold<1, P>,  just needs some review and cleanup before I commit it.

dellaert: Ah, excellent. I'd need to be convinced it does not trade performance for generality, esp. in the often used Unit3 case, but I think that's a great development :-)

zhexin1904:
Rot2 is now used, and Unit2 has been removed, but I still have a few optimizer changes to commit before completing this PR.

gtsam/slam/RelativeTranslationFactor.h
```diff
@@ -0,0 +1,155 @@
...
+          "RelativeTranslationFactor::qcqpFactors: costs is null.");
+    }
+
+    const double sw = std::sqrt(weight_);
+    Matrix B = Matrix::Zero(1, d + 2);
+    B.block(0, 0, 1, d) = -sw * measured_.transpose();  // rotation row
+    B(0, d) = -sw;                                      // pose i translation
+    B(0, d + 1) = sw;                                   // pose j translation
+
+    const Matrix Q = B.transpose() * B;
```
dellaert, line 144: @zhexin1904 This is probably quite expensive. The QP cost takes a symmetric block matrix. This indeed has a constructor from a regular `Matrix`, but it might not be the most efficient way to create it. A little bit of a timing script might show whether it matters, But we can also keep this as a follow up PR.

dellaert:
@zhexin1904 When talking to @avinashresearch1 we discovered that some "certifiable" papers use conventions that do not agree with GTSAM's pose convention. Lest we confuse the hell out of all GTSAM users we should make sure we use `Pose3` variables $T^w_k$, written in code as `wTk`, where $k$ indexes the $k^{th}$ pose and $w$ is the world frame. Often, especially in the `navigation` folder, we actually use $n$ instead of $w$, for the "navigation frame".

A pose $T^w_k$ is a pair $(R^w_k, t^w)$, where $R^w_k\in SO(3)$. The SO(3) group action transforms points $P^k$ **from** the pose frame $k$ *to* points in the world frame $w$:

$$
P^w =  R^w_k P^k
$$

or, in code (please use this convention wherever possible !!!):
```python
wP = wRk * kP; # the k indices abut and "cancel"!
```

For $SE(3)$/`Pose3` the group action also transforms points **from** $k$ **to** $w$:

$$
P^w =  R^w_k P^k + t^w
$$

This is what the method `Pose3.transformFrom` does: **from** the pose frame to the world frame. To go the other way, from $w$ to $k$  you need to use `Pose3.transformTo`.

Please make sure that all your code honors these conventions, and that the examples also use and reinforce this. The idea is as is that certifiable optimization is something that regular GTSAM users can just add very easily. If we switch convention somewhere in the middle, it will be a nightmare. **One way to get this right: whenever you have a `Pose3`, please make sure it is written as `wTk`, and never use a `kTw` or `kRw`. Instead, you'd write `wTk.inverse()` and `wRk.transpose()` if you need to convert from $w$ to $k$. And when you can, just use `transformTo`.**

gtsam/slam/RelativeTranslationFactor.h
zhexin1904: Yeah, I may need some check like this for an enhancement PR, but the target PR will focus on the complete framework.

zhexin1904:
> @zhexin1904 When talking to @avinashresearch1 we discovered that some "certifiable" papers use conventions that do not agree with GTSAM's pose convention. Lest we confuse the hell out of all GTSAM users we should make sure we use `Pose3` variables T k w , written in code as `wTk`, where k indexes the k t h pose and w is the world frame. Often, especially in the `navigation` folder, we actually use n instead of w , for the "navigation frame".
> 
> A pose T k w is a pair ( R k w , t w ) , where R k w ∈ S O ( 3 ) . The SO(3) group action transforms points P k **from** the pose frame k _to_ points in the world frame w :
> 
> P w = R k w P k
> 
> or, in code (please use this convention wherever possible !!!):
> 
> ```python
> wP = wRk * kP; # the k indices abut and "cancel"!
> ```
> 
> For S E ( 3 ) /`Pose3` the group action also transforms points **from** k **to** w :
> 
> P w = R k w P k + t w
> 
> This is what the method `Pose3.transformFrom` does: **from** the pose frame to the world frame. To go the other way, from w to k you need to use `Pose3.transformTo`.
> 
> Please make sure that all your code honors these conventions, and that the examples also use and reinforce this. The idea is as is that certifiable optimization is something that regular GTSAM users can just add very easily. If we switch convention somewhere in the middle, it will be a nightmare. **One way to get this right: whenever you have a `Pose3`, please make sure it is written as `wTk`, and never use a `kTw` or `kRw`. Instead, you'd write `wTk.inverse()` and `wRk.transpose()` if you need to convert from w to k . And when you can, just use `transformTo`.**

Hi Frank @dellaert, I think we follow GTSAM’s coordinate convention. (I think I know papers you mean using world-to-body.) I can add some comments to clarify this if needed.

zhexin1904:
The local solver + certificate usage hasn’t been added yet. But I think I can submit it as the next PR over the weekend (or after the ICRA deadline:)).

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
```diff
@@ -555,39 +829,66 @@ RiemannianStaircaseResult RiemannianStaircaseOptimizer::optimize() const {
...
     }
+    /// An ALM run that stops before its first dual update returns lambda = 0.
+    /// Then S = Q + A*(0) = Q, which is positive semidefinite by construction
+    /// and would certify any point, so recover the multipliers from Y instead.
+    std::vector<Vector> lambdaEq = inner.lambdaEq;
+    double maxAbsLambda = 0.0;
+    for (const auto& value : lambdaEq) {
+      maxAbsLambda = std::max(maxAbsLambda, value.cwiseAbs().maxCoeff());
+    }
+    if (maxAbsLambda == 0.0) {
```
dellaert, line 895: **Correctness — require feasibility and KKT stationarity before certifying.** `runLocalSolver()` may stop because the inner solve did not converge or ALM hit its iteration cap, but this path still proceeds to the PSD check and can set `certified=true`. Recovering least-squares multipliers only when the ALM multipliers are bitwise zero does not close that gap, and `ls.totalResidual` is only printed. PSD of `S` alone is insufficient: the candidate must also be primal-feasible and satisfy complementary slackness (`S Y ≈ 0`). The new `LeastSquaresMultipliersFlagNonStationaryPoint` test demonstrates the underlying failure mode: `S = Q` is PSD and `verify()` passes regardless of a non-stationary point. Please gate certification on the QCQP constraint violation and a normalized residual such as `||S Y|| / max(1, ||Y||)`, ideally recovering/checking the least-squares multipliers for every candidate rather than only exact-zero multipliers. If ALM did not converge, continue/refine or return an uncertified result rather than treating PSD as a certificate.

dellaert (commented):
## AI review — human review pending

This review was generated by an AI coding agent on Frank's behalf. Frank has **not yet completed a personal read-through of the current PR**; he will do that after the comments below have either been addressed or answered with a rationale for why they are not relevant.

The comments are grouped into three areas:

- **Correctness:** certification prerequisites, stationarity diagnostics, and saddle-merit state.
- **Performance:** repeated sparse-matrix assembly and per-measurement QCQP construction.
- **Style/maintainability:** the current CI failure, certificate-convention documentation, and implementation-comment style.

@zhexin1904, please reply to any comment you believe is not relevant so Frank can evaluate that judgment during his review. **Please resolve a conversation only after the underlying comment has actually been addressed.** If you are rejecting a comment as not relevant, leave that conversation unresolved after replying with the rationale. Once all threads are either addressed or have a clear response, Frank will read through the PR and make the human review decision.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
```diff
@@ -260,7 +262,8 @@ RiemannianStaircaseOptimizer::runLocalSolver(
         "RiemannianStaircaseOptimizer::runLocalSolver: ALM produced no "
         "progress.");
   }
-  return {Y, alm.progress().back().lambdaEq};
+  const auto& last = alm.progress().back();
+  return {Y, last.lambdaEq, last.augmentedLagrangianStationarity};
```
dellaert: **Correctness — populate the stationarity history.** This now returns the last ALM stationarity, and the PR exposes `stationarityPerLevel` plus a wrapper-facing getter, but `optimize()` never appends `inner.stationarity`. Consequently `getStationarityPerLevel()` always returns an empty vector even when ranks were visited. Please push this value beside the other per-level diagnostics and extend the diagnostics-size test to cover it.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
```diff
@@ -606,16 +907,22 @@ RiemannianStaircaseResult RiemannianStaircaseOptimizer::optimize() const {
...
-    // O(alpha^2) violation is absorbed by the next inner solve.
+    /// Alg. 1 lines 10-12: S not PSD, lift to rank p+1 along v_min. The
+    /// O(alpha^2) violation is absorbed by the next inner solve.
     if (p < params_.pMax) {
-      Y = escapeSaddleAndLift(Y, vMin, layout, params_.alpha);
+      const auto buildStart = Clock::now();
+      carriedQcqp = std::make_shared<QcqpProblem>(graph_, p + 1);
+      carriedQcqpBuildTime = ElapsedSeconds(buildStart);
+      Y = saddleEscapeWithLineSearch(*carriedQcqp, layout, Y, vMin, lambdaEq,
+                                     params_.almParams->bclInitialPenalty,
```
dellaert: **Correctness — use the final ALM state for the saddle merit.** The line search says it guards “ALM's own augmented Lagrangian,” but this passes `bclInitialPenalty`. BCL may increase the penalty during the local solve, so an enabled line search can accept a step against a different merit function from the final primal-dual state. Please return/pass the relevant final ALM state (including its penalty) from `runLocalSolver()`, or narrow the guarantee in the API. This dereference also conflicts with `runLocalSolver()` explicitly accepting a null `almParams` and creating defaults; normalize or reject null once rather than crashing here.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
```diff
@@ -369,28 +371,216 @@ Eigen::SparseMatrix<double> RiemannianStaircaseOptimizer::buildMultiplierMatrix(
...
+      throw std::runtime_error(
+          "leastSquaresMultipliers: constraint A size does not match key "
+          "rowDim.");
+    }
+    groups[key].push_back({m, 0.5 * (A + A.transpose()), constraint.sigma()});
+    ++m;
+  }
+
+  const Matrix Ystack = layout.stack(Y);
+  const Matrix G = buildDataMatrix(qcqp, layout) * Ystack;
```
dellaert: **Performance — avoid assembling the same sparse data matrix twice.** Multiplier recovery builds `Q` here to compute `QY`, after which `buildCertificate()` immediately rebuilds the identical `Q`. On the large pose/range graphs introduced by this PR, that repeats factor casts, expanded-block copies, triplet allocation/sorting, and sparse compression. Please assemble `Q` once per staircase level and pass it to both routines, or have multiplier recovery return the completed certificate. Reuse would also make an always-on KKT residual gate affordable.

gtsam/slam/RelativeTranslationFactor.h
dellaert, line 144: **Performance — follow up with a measured cheaper QCQP construction.** Each measurement forms dynamic dense `B`, then dense `Q = BᵀB`, then copies it into `SymmetricBlockMatrix`; `QuadraticRangeFactor` follows the same pattern. With thousands of factors this means many small allocations. I agree this need not block the framework PR if benchmarked as a follow-up, but please track and measure a fixed-size/direct-block or rank-one construction on the included 1k datasets so QCQP build cost does not become the scaling bottleneck.

gtsam/certifiable/tests/testRiemannianStaircase.cpp
```diff
@@ -837,3 +839,473 @@ int main() {
...
+
+  const auto ls = RiemannianStaircaseOptimizer::leastSquaresMultipliers(
+      qcqp, layout, Y);
+  const auto S = RiemannianStaircaseOptimizer::buildCertificate(qcqp, layout,
+                                                                ls.lambdaEq);
+  EXPECT(ls.totalResidual > 1e-6);
+
+  // d/d lambda_m of ||S Y||^2 vanishes: <A_m Y_n, (S Y)_n> = 0 for every m.
+  const Matrix SY = Matrix(S) * layout.stack(Y);
+  size_t m = 0;
```
dellaert: **Style/CI — remove the unused counter.** The full macOS job fails under `-Werror` because `m` is incremented but never read; the Linux and Windows full jobs were then cancelled. Please delete `m`/`++m` (or use the index if an indexed assertion was intended) and rerun the complete matrix.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
```diff
@@ -369,28 +371,216 @@ Eigen::SparseMatrix<double> RiemannianStaircaseOptimizer::buildMultiplierMatrix(
...
     const QcqpProblem& qcqp, const Layout& layout,
     const std::vector<Vector>& lambdaEq) {
   Eigen::SparseMatrix<double> S = buildDataMatrix(qcqp, layout);
-  S += buildMultiplierMatrix(qcqp, layout, lambdaEq);
+  /// The factor of two reconciles the cost and constraint conventions:
+  /// QpCost is 0.5 * tr(X' Q X) so its gradient is Q Y, while a
+  /// QuadraticConstraint h = tr(X' A X) - b has gradient 2 A X. Stationarity of
+  /// L = cost + sum_m lambda_m h_m is therefore (Q + 2 A*(lambda)) Y = 0.
+  /// So a factor of 2 is added.
+  S += 2.0 * buildMultiplierMatrix(qcqp, layout, lambdaEq);
```
dellaert, line 498: **Style/maintainability — align the public certificate convention with the implementation.** The class and method documentation define `S = Q + A*(lambda)` and describe `buildCertificate()` as data plus multiplier matrices, while the implementation now forms `Q + 2 A*(lambda)`. The implementation may be correct given GTSAM's 0.5-cost and constraint-gradient conventions, but the public API currently communicates a different multiplier convention. Please state consistently whether `lambdaEq` multiplies `h`, `h/sigma`, or the already doubled matrix term, and update the theorem summary and both builder method docs accordingly.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
```diff
@@ -369,28 +371,216 @@ Eigen::SparseMatrix<double> RiemannianStaircaseOptimizer::buildMultiplierMatrix(
...
+/* ************************************************************************* */
+RiemannianStaircaseOptimizer::LeastSquaresMultipliers
+RiemannianStaircaseOptimizer::leastSquaresMultipliers(const QcqpProblem& qcqp,
+                                                     const Layout& layout,
+                                                     const Values& Y) {
+  const size_t M = qcqp.eConstraints().size();
+  LeastSquaresMultipliers result;
+  result.lambdaEq.assign(M, Vector1(0.0));
+
+  /// Constraints are unary, so grouping by key partitions them and the
```
dellaert: **Style/maintainability — reserve `///` for documentation comments.** This PR changes many ordinary implementation notes in the `.cpp` file from `//` to Doxygen `///`, including local algorithm steps and comments inside functions. These are not API documentation and make the documentation signal noisy. Please keep `///` on declarations/public members and use regular `//` comments within function bodies.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
zhexin1904: I think this is a limitation of the current ALM solver. Stationarity is satisfied
throughout, and feasibility is satisfactory on pose-graph and rotation
averaging, less so on the harder RA-SLAM problems.

**Gradient.** Using multipliers recovered by least squares,
`||SY||/max(1,||Y||)` at the certifying level is `6.5e-13` to `1.7e-04`, below
`1e-7` on 17 of 24 runs, so the stationarity condition is met at every point we
certify.

The better long-term fix is gradient-based termination in the inner solver
rather than a residual criterion. The ALM currently reports `converged = false`
on all 24 runs, and the outer loop exits on inner non-convergence before its
feasibility test applies — which is why the examples request
`absoluteViolationTolerance = 1e-8` and get `2e-4`.

**Violation.** Small for pose-graph and rotation averaging (`5.5e-08` to
`1.2e-03`), moderate for landmark (`2.9e-03` to `4.0e-02`), and large for
range-aided SLAM (`0.19` to `0.59`). Dave's WAFR paper proposes a
feasibility-restoration strategy
([[link](https://link.springer.com/chapter/10.1007/978-3-030-66723-8_33)](https://link.springer.com/chapter/10.1007/978-3-030-66723-8_33)); we
could adopt it, but it needs implementation and verification and I'm not
certain of the benefit here. This is again an ALM issue, and since we intend to
propose a better solver, I'd leave it for now given the timeline.

**On the multipliers.** Yes — saddle escape needs KKT stationarity, since the
lifting direction and the merit function both use `lambda`, and that is exactly
what the least-squares multipliers are for: they are defined as
`argmin_lambda ||[Q + A*(lambda)] Y||_F`, so they are the multipliers that drive
the KKT stationarity residual to its minimum at `Y`. This is the usual
local-solver-plus-certificate arrangement — the local solver supplies the primal
point and the certificate machinery recovers the dual variable itself rather
than trusting whatever the solver happened to end on — and under LICQ they are
unique, so there is no ambiguity about which multipliers the certificate refers
to. Across all 37 levels in 24 runs the fallback fires exactly once, on landmark
cityTrees_1k / random at rank 2, where ALM returns `lambda = 0`: `||SY||` goes
from `1088` to `0.0139` (normalized `1.4e-05`). Stationarity is recovered;
feasibility at that point is `1e-1`, so it is not a KKT point and it does not
certify. The warm start therefore doesn't strictly meet the theorem's
hypotheses, but the next level re-optimizes in a better landscape — that run
goes on to certify at rank 4 with an objective matching the fast-sync run.

If ALM did not converge, re-optimizing at the current level would not fix it. The appropriate approach is to certify that it is a local solution and lift it to a higher level for further optimization. I know this does not **strictly** satisfy the requirements of the theorem, but it is one practical approach given the limitations of ALM on **certain difficult** datasets.

**In conclusion**, ALM seems work well on pose-graph optimization: KKT
stationarity is met and the violation stays between `5.5e-08` and `1.2e-03`. It
still works on the small RA-SLAM data. All the results reported here give the
same objective as the ones in our draft. The better fix is the improved inner
solver we have planned after ICRA.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
zhexin1904: BUGS! Fixed.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
zhexin1904: Fixed, the ALM’s final penalty is now passed to the saddle merit function. In most runs, the benefit of this line search is marginal, with almost no difference in runtime or correctness compared with only using the original returned eigenvector. Therefore, I disabled the line search by default.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
zhexin1904: Good point! Fixed, now only assemble data matrix once and reuse.

gtsam/certifiable/tests/testRiemannianStaircase.cpp
zhexin1904: Fixed.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
zhexin1904: Fixed.

gtsam/certifiable/RiemannianStaircaseOptimizer.cpp
zhexin1904: Update comments in a few other places related to this conversion.

dellaert (approved):
Approving again :-) Will merge after CI succeeds. 
Congratulations on this awesome contribution ! After the ICRA deadline we can make it super fast. :-).

dellaert:
@zhexin1904 a CI failure :-(
```
 41/362 Test  #39: testRiemannianStaircase .......................***Failed    0.02 sec
/Users/runner/work/gtsam/gtsam/gtsam/certifiable/tests/testRiemannianStaircase.cpp:1247: Failure: "lambdaMin >= -params.eta" 
There were 1 failures
```

zhexin1904:
> @zhexin1904 a CI failure :-(
> 
> ```
>  41/362 Test  #39: testRiemannianStaircase .......................***Failed    0.02 sec
> /Users/runner/work/gtsam/gtsam/gtsam/certifiable/tests/testRiemannianStaircase.cpp:1247: Failure: "lambdaMin >= -params.eta" 
> There were 1 failures
> ```

I see. I’ll do a global check of the CI tomorrow.

dellaert:
Another failure :-(

zhexin1904:
> Another failure :-(

OK it seems due to a mismatch between the wrapper in my fork and the upstream version. I am fixing it now...

dellaert:
If CI fails again, you should probably rebase on or merge in develop. The latter is more appropriate since you have already published this branch.

dellaert:
Done! Congrats :-)

zhexin1904:
> Done! Congrats :-)

Thanks a lot for the valuable feedback! and looking forward to the next solver work :))

dellaert:
@zhexin1904 Hey Jason, when rereading this PR, we had a couple of questions like:
- in the pose graph examples, do you add a prior, or do you rely on LM to to deal with singular matrices? 
- You could do SE(3) pose graph with `FrobeniusBetweenFactorPose3`? Did you ever try that?

zhexin1904:
Hi @dellaert ,
Q1: We have not added the prior yet. So, yes, if ill-conditioned matrices arise, they will be handled by LM. All our implementations of BM-based certifiable estimators are **gauge-invariant**. We can add a prior, but its implementation would be slightly different because a standard prior such as (||data-state||^2) is not a homogeneous QCQP. It would need to be homogenized, similar to what is done in the Chordal-SDP branch by appending a constant scale of (1) to the vector. In short, we have not added it yet, but we can add it later (I remember Shonan-RA does provide prior right? I haven't checked).

Q2: I have not tried it, but I do not think we can use FrobeniusBetweenFactorPose3 in the same way as we do for rotation averaging, at least through a direct compact lifting. The most immediate reason is that BM methods need to distinguish between constrained and unconstrained variables. A pose combines a constrained rotation with an unconstrained translation, so it does not have a single compact lift. Rotation averaging is the simpler case.
This is also where BM differs slightly from SDP methods. An SDP can vectorize the entire pose and let the constraint list encode its structure. In our BM framework, however, the constrained part must be handled explicitly because retraction, rounding, and multiplier computation all depend on which variables are constrained. Therefore, this partition must be explicit in the variable layout.
So,  I feel the best way is to provide an interface that makes it look like pose averaging externally, while internally it still splits the rotation and translation.

avinashresearch1:
Thanks. I appreciate the detailed response.

zhexin1904:
> Thanks. I appreciate the detailed response.

Of course! Please let me know if needs such a prior factor maybe for a benchmark? (and also any other questions!). Or I’ll send some PRs after the ICRA deadline.

PR #2715, opened 2026-08-20
fix: prevent BoundingConstraint1 iSAM2 update segfault

### LINKED ISSUE #1905: BoundingConstraint1  segfaults on optimization
Hi，
I've been using **BoundingConstraint1** from the gtsam library lately 。
Here's my code ，code is from [here](https://github.com/MIT-SPARK/Hydra/blob/0dd9653136d2f1f31a8a205d0e69a78a738d4403/eval/tools/gt_trajectory_optimizer.cpp#L76)（thanks to the author）：

```
class Pose3Bounds : public gtsam::BoundingConstraint1<Eigen::Vector3d> {
 public:
  using shared_ptr = gtsam::NonlinearFactor::shared_ptr;

  Pose3Bounds(
      gtsam::Key key, size_t idx, double threshold, bool is_greater, double mu = 1000.0)
      : gtsam::BoundingConstraint1<Eigen::Vector3d>(key, threshold, is_greater, mu),
        idx_(idx),
        threshold_(threshold),
        is_greater_(is_greater),
        mu_(mu) {}

  double value(const Eigen::Vector3d& x,
                GtsamJacobianType H = JACOBIAN_DEFAULT) const override {
    if (H) {
        gtsam::Matrix D =
          gtsam::Matrix::Zero(1, 3);
      D(0, idx_) = 1.0;
      *H = D;
    }
    return x(idx_);
  }

  shared_ptr clone() const override {
    return shared_ptr(new Pose3Bounds(front(), idx_, threshold_, is_greater_, mu_));
  }

 private:
  size_t idx_;
  double threshold_;
  bool is_greater_;
  double mu_;
};

//This part is pseudocode
	std::shared_ptr<gtsam::NonlinearFactorGraph> graph_factor_;
	std::shared_ptr<gtsam::Values> initial_values_;
	std::shared_ptr<gtsam::ISAM2> optimizer_imu_gps_;

graph_factor_->emplace_shared<Pose3Bounds>(V(*correction_count), 1,-0.4, true);
optimizer_->update(*graph_factor_, *initial_values_);
```

I have a three-dimensional vector and Key is V。  
`graph_factor_->emplace_shared<Pose3Bounds>(V(*correction_count), 1,-0.4, true);
optimizer_->update(*graph_factor_, *initial_values_);` 
Within a loop, there is constantly new data and then optimized。 

 if don't use **BoundingConstraint1** , this **optimizer_->update(）**  can run smoothly and optimize successfully。 Once **BoundingConstraint1**  is used, the program will be segfaults on the second execution of the **optimizer_->update()** session。

 this issue like at line [268] of file [testBoundingConstraint.cpp](https://docs.ros.org/en/melodic/api/gtsam/html/testBoundingConstraint_8cpp_source.html) 。 
 
 
 OS [e.g. Ubuntu 18.04, Ubuntu 20.04]

mvanhorn:
## Summary

Update `ISAM2::relinearizeAffectedFactors` so cached inactive factors are handled as a supported state: do not dereference a null cached or freshly linearized factor, re-evaluate an affected nonlinear factor when its cache cannot safely be reused, and keep the cache synchronized when the factor transitions between active and inactive. Preserve the existing factor-slot and variable-index bookkeeping so active factors still participate in elimination while inactive factors are omitted without introducing a test-only production abstraction. Add a focused regression to `testBoundingConstraint.cpp` using the existing simulated 2D bounding-constraint types and two sequential iSAM2 updates, mirroring the reported incremental workflow rather than changing `BoundingConstraint1`'s public API or its established inactive-linearization semantics.

## Validation

- Run two sequential `ISAM2::update` calls on a graph with an existing bounded variable and newly added incremental data; verify the second update completes and produces a finite estimate instead of crashing.
- Exercise a bounding constraint whose linearization is inactive during an update, then cause its key to be reconsidered by a later update; verify iSAM2 safely skips or refreshes the null cached factor.
- Exercise the complementary active constraint path across the two updates; verify the constraint's Gaussian factor remains in elimination and the estimate respects the configured bound.
- Retain the existing direct tests that inactive bounding constraints linearize to null and active constraints produce the expected constrained Jacobian factor, ensuring the fix does not alter factor-level semantics.

## Motivation

`BoundingConstraint1` can crash incremental optimization on the second `ISAM2::update`, even though the same update sequence works without the constraint. Inequality constraints legitimately become inactive, and `NoiseModelFactor::linearize` represents an inactive factor with a null Gaussian factor. iSAM2 retains linearized factors between updates, so a cached null entry or a factor that changes active state can reach affected-factor relinearization paths that assume a usable Gaussian factor. The issue is open, unassigned, confirmed as a bug by a maintainer, and has no competing or prior closed pull request in the supplied repository evidence.

Fixes #1905

### DIFF
--- a/gtsam/nonlinear/ISAM2.cpp
+++ b/gtsam/nonlinear/ISAM2.cpp
@@ -101,22 +101,23 @@ GaussianFactorGraph ISAM2::relinearizeAffectedFactors(
         useCachedLinear = false;
     }
     if (inside) {
-      if (useCachedLinear) {
+      GaussianFactor::shared_ptr linearFactor;
+      if (useCachedLinear && linearFactors_[idx]) {
 #ifdef GTSAM_EXTRA_CONSISTENCY_CHECKS
-        assert(linearFactors_[idx]);
         assert(linearFactors_[idx]->keys() == nonlinearFactors_[idx]->keys());
 #endif
-        linearized.push_back(linearFactors_[idx]);
+        linearFactor = linearFactors_[idx];
       } else {
-        auto linearFactor = nonlinearFactors_[idx]->linearize(theta_);
-        linearized.push_back(linearFactor);
+        linearFactor = nonlinearFactors_[idx]->linearize(theta_);
         if (params_.cacheLinearizedFactors) {
 #ifdef GTSAM_EXTRA_CONSISTENCY_CHECKS
-          assert(linearFactors_[idx]->keys() == linearFactor->keys());
+          assert(!linearFactor ||
+                 linearFactor->keys() == nonlinearFactors_[idx]->keys());
 #endif
           linearFactors_[idx] = linearFactor;
         }
       }
+      if (linearFactor) linearized.push_back(linearFactor);
     }
   }
   gttoc(check_candidates_and_linearize);
--- a/tests/testBoundingConstraint.cpp
+++ b/tests/testBoundingConstraint.cpp
@@ -18,6 +18,7 @@
 #include <CppUnitLite/TestHarness.h>
 #include <gtsam/base/MatrixConstants.h>
 #include <gtsam/inference/Symbol.h>
+#include <gtsam/nonlinear/ISAM2.h>
 #include <gtsam/nonlinear/LevenbergMarquardtOptimizer.h>
 #include <gtsam/nonlinear/LevenbergMarquardtParams.h>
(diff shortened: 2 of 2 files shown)


dellaert (approved):
Nice, thank you !! Steady progress towards GTSAM 4.3 !

dellaert:
Thanks @mvanhorn !!!

PR #2716, opened 2026-08-20
Expose unregistered GaussianFactor subclasses as their registered base

DLuminary:
NonlinearFactor::linearize and GaussianFactorGraph::at are declared to return gtsam::GaussianFactor*. pybind11 resolves a polymorphic return by looking up typeid(*src); when the dynamic type is not registered it does not walk up the hierarchy, it falls back to the declared type. So an unregistered concrete factor arrives in Python as a bare GaussianFactor, whose wrapped surface is only equals, error, clone, negate, augmentedInformation, information, augmentedJacobian and jacobian. getA, getb, rows, jacobianUnweighted and get_model are declared on JacobianFactor and are simply absent.

Since #2710, every NoiseModelFactorT whose error and argument dimensions are all fixed linearizes to FixedJacobianFactor<M, Ns...>. That type is templated on its dimensions, so it cannot practically be registered. BetweenFactor names a fixed-size TangentVector as its error type and is affected; PriorFactor goes through NoiseModelFactorN, whose error type is a dynamic Vector, so it keeps the generic path and a plain JacobianFactor. The declared return type has been GaussianFactor since long before that change, so the behaviour difference comes entirely from the concrete type.

RegularJacobianFactor and RegularHessianFactor are unregistered for the same reason and are covered by the same hook.

Note the object itself is unchanged: FixedJacobianFactor overrides hessianDiagonalAdd, multiplyHessianAdd and updateHessian, and those overrides still run. Only the Python-side wrapper type changes.

The hook consults the registry first, so registered subclasses keep their own identity: GaussianConditional derives from JacobianFactor and must still arrive in Python as a GaussianConditional.

It lives in a header included by both module templates rather than in a per-module preamble, because linearize is declared in nonlinear.i while the factor types are registered from linear.i, and pybind11 needs the hook visible in every translation unit that casts a GaussianFactor -- the same constraint that applies to PYBIND11_MAKE_OPAQUE.

Two limitations, both noted in the header: the list of fallback bases is hardcoded, since C++ RTTI cannot enumerate base classes; and detail::get_type_info is pybind11 internal API. A general fix -- having pybind11 fall back to the nearest registered base rather than the declared type -- would belong upstream in pybind11.

### DIFF
--- a/python/gtsam/gaussian_factor_pybind.h
+++ b/python/gtsam/gaussian_factor_pybind.h
@@ -1,0 +1,83 @@
+#pragma once
+
+/* Surface unregistered GaussianFactor subclasses as their nearest registered
+ * base.
+ *
+ * pybind11 resolves a polymorphic return value by looking up typeid(*src). If
+ * the dynamic type is not registered it does NOT walk up the hierarchy -- it
+ * falls back to the *declared* return type. NonlinearFactor::linearize is
+ * declared to return gtsam::GaussianFactor*, so any concrete factor type that
+ * is not registered surfaces in Python as a bare GaussianFactor, losing the
+ * JacobianFactor interface (getA, getb, jacobian, augmentedJacobian, ...).
+ *
+ * That is what happens to FixedJacobianFactor<M, Ns...>, which is templated on
+ * its dimensions and so cannot practically be registered: every fixed-size
+ * NoiseModelFactorT linearizes to one. gtsam::BetweenFactor names a fixed-size
+ * TangentVector as its error type and is affected; gtsam::PriorFactor goes
+ * through NoiseModelFactorN, whose error type is a dynamic Vector, so it keeps
+ * the generic path and a plain JacobianFactor.
+ *
+ * The hook checks the registry first, so registered subclasses keep their own
+ * identity -- gtsam::GaussianConditional derives from JacobianFactor and must
+ * still arrive in Python as a GaussianConditional.
+ *
+ * This lives in a header included by both module templates rather than in a
+ * per-module preamble: linearize is declared in nonlinear.i while the factor
+ * types are registered from linear.i, and pybind11 needs the hook visible in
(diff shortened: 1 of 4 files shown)


dellaert:
@ProfFan Would you be willing to take a look at this?

dellaert:
I must say I don't quite understand which problem this PR is trying to solve. Yes, a polymorphic base class is returned and that loses some of the methods in the derived classes. That's entirely the same as in C++, though, where if you want to dynamic cast to a derived class you can but it might also fail. Of course, dynamic casting is not available in Python and maybe that's the problem you're trying to solve?

ProfFan:
Ah, I think I know, @DLuminary wants to get JacobianFactor after calling `linearize()` which at this moment they will get a `GaussianFactor` instead which does not allow them to get the matrices

dellaert:
Right. In C++ we solve this for by dynamic casting. I'm not sure that the mechanism proposed here is a good way of providing a capability. Maybe something akin to dynamic casting is more appropriate.

DLuminary:
You're right, and I'd missed that JacobianFactor(const GaussianFactor&) is already wrapped — gtsam.JacobianFactor(gf) does exactly the dynamic_cast-based conversion you're describing. Closing this.

For the record on what prompted it: factor.linearize(values) used to return something I could call getA()/getb() on directly, and now returns a bare GaussianFactor. The change came in with #2652, where the binary/ternary fast paths began returning BinaryJacobianFactor<M,N1,N2>; #2710 generalised that to FixedJacobianFactor<M,Ns...> for all arities. Those types are templated on their dimensions so they aren't registered, and pybind falls back to the declared return type rather than the nearest registered base.

So before the breaking change (python-side) it was:
jf = factor.linearize(values)
And now I call it via:
jf = gtsam.JacobianFactor(factor.linearize(values))

Since the conversion already exists this is a one-line fix on my side. Might be worth a note in the Python docs for anyone else whose code changed behavior, but no code change needed.

PR #2720, opened 2026-08-21
wrap: complete the fixed-lag smoother state accessors

DLuminary:
Both fixed-lag smoothers keep public accessors for their optimizer state that are not declared in the wrapper, so from Python the only way to reach them is indirectly, or not at all.

BatchFixedLagSmoother gains getLinearizationPoint, getOrdering, getDelta and marginalCovariance. The last is a plain asymmetry: IncrementalFixedLagSmoother has had marginalCovariance wrapped for some time and the batch smoother has the same method, so of the five classes that offer it -- Marginals, ISAM2, NonlinearISAM and both smoothers -- the batch one was the only place it was unreachable.

IncrementalFixedLagSmoother gains getLinearizationPoint and getDelta. Both forward directly to the contained ISAM2 and return the same object by reference, so this is purely a matter of not having to reach through getISAM2() to get at them.

FixedLagSmootherResult gains getIntermediateSteps, the one getter on that struct that was left out. Under Levenberg-Marquardt it counts damped linear solves: each outer iteration adds at least one, and more whenever lambda has to be increased before the error drops, which is what explains an expensive batch update alongside the iteration count already reported.

Worth noting that IncrementalFixedLagSmoother does not populate the Result statistics at all -- it assigns iterations = 1 and leaves error, linearVariables, nonlinearVariables and now intermediateSteps at their constructed defaults. That is pre-existing and true of the four getters already wrapped; getISAM2Result() is where its live numbers are. The new getter is added for consistency of the struct's Python surface, and the test exercises it against the batch smoother accordingly.

All seven already exist in C++ and are public. The container accessors are declared const& to match their C++ signatures, so the wrapper emits reference_internal for five of them; getIntermediateSteps returns a size_t and marginalCovariance computes a Matrix, so neither takes a policy.

FixedLagSmoother itself needs nothing: its remaining unwrapped members -- getCurrentTimestamp, findKeysBefore, findKeysAfter, updateKeyTimestampMap and eraseKeyTimestampMap -- are protected, as are eraseKeysBefore and createOrderingConstraints on the incremental smoother. Exposing those would require changing their access in C++ and is left alone here.

Generated the pybind sources before and after: the delta is exactly these seven declarations and nothing else.

### DIFF
--- a/gtsam/nonlinear/nonlinear.i
+++ b/gtsam/nonlinear/nonlinear.i
@@ -1007,6 +1007,7 @@ class FixedLagSmootherKeyTimestampMap {
 
 class FixedLagSmootherResult {
   size_t getIterations() const;
+  size_t getIntermediateSteps() const;
   size_t getNonlinearVariables() const;
   size_t getLinearVariables() const;
   double getError() const;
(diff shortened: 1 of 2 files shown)


DLuminary:
CI caught that BatchFixedLagSmoother::marginalCovariance is declared but its definition throws not implemented. Dropped it and its tests; the remaining six all have real definitions.

dellaert (approved):
Nice!

PR #2721, opened 2026-08-21
Expose the remaining per-update key sets on ISAM2Result

DLuminary:
ISAM2Result carries five public members that describe what an update actually touched -- factorsRecalculated, unusedKeys, observedKeys, keysWithRemovedFactors and markedKeys -- but unlike the rest of the class they have no getters, so they were not reachable from the wrappers.

Adds inline getters in the existing style and declares them in nonlinear.i. The container getters return const& to match the convention established in #2699; the wrapper emits reference_internal for them.

These are the fields that explain why a given update was expensive, which matters for long-running incremental sessions where an occasional relinearization storm is otherwise invisible from Python.

Note this is the one change in this series that touches a C++ header rather than only a .i file, hence keeping it separate from the adaptive reorder commit.

The same result is achievable with no header change by declaring the five members as fields in the .i, the way ISAM2Params already does — I've verified that generates cleanly (def_readwrite for each). The trade-off is that fields are writable, which is odd for a result object, whereas the getters are read-only. Happy to switch if you'd prefer to keep it wrapper-only.

### DIFF
--- a/gtsam/nonlinear/ISAM2Result.h
+++ b/gtsam/nonlinear/ISAM2Result.h
@@ -203,6 +203,13 @@ struct ISAM2Result {
   bool getBatchReorderTriggered() const { return batchReorderTriggered; }
   double getErrorBefore() const { return errorBefore ? *errorBefore : std::nan(""); }
   double getErrorAfter() const { return errorAfter ? *errorAfter : std::nan(""); }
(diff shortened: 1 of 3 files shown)


dellaert (approved):
Thanks !

Issue #886, opened 2021-10-05
Buggy axis-angle computation in GTSAM


johnwlambert:
GTSAM returns angle's of well over 180 degrees when transforming a Rot3() object with `.axisAngle()`.

Here's an example below, where the angle returned is 204 degrees. Scipy's Rotation package returns the same axis vector as GTSAM, but returns the angle as 179 degrees, which seems legitimate.

I believe the angle of the angle-axis representation should be confined to [0, pi] radians, per [Duke](http://motion.pratt.duke.edu/RoboticSystems/3DRotations.html#:~:text=Axis%2Dangle%20representation%20uses%20a,3%2DD%20space%20of%20rotations.) and the notion of being smallest rotation to take us there (we would go the opposite way, instead, otherwise):

Given the following 3x3 matrix:
```python
(Pdb) p relative_rot
R: [
	-0.0309726, -0.789768, -0.612625;
	-0.787144, -0.358463, 0.501909;
	-0.615993, 0.497768, -0.610558
]
```
GTSAM provides an axis-angle decomposition:
```python
(Pdb) p relative_rot.axisAngle()
(:-0.696212
 0.566258
 0.441181
, 3.576080425974451)
```
Angle is 204 degrees:
```python
(Pdb) np.rad2deg( 3.576080425974451)
204.89431560768165
```

Scipy states the angle is 179.93 degrees:
```python
(Pdb) from scipy.spatial.transform import Rotation
(Pdb) Rotation.from_matrix(relative_rot.matrix()).as_rotvec()
array([-2.18469647,  1.77759913,  1.38497925])
(Pdb) axisangle = Rotation.from_matrix(relative_rot.matrix()).as_rotvec()
(Pdb) np.linalg.norm(axisangle)
3.138618297175063
(Pdb) np.rad2deg(3.138618297175063)
179.8295819306683
```

Scipy's axis of rotation is identical:
```python
(Pdb) axisangle /= np.linalg.norm(axisangle)
(Pdb) axisangle
array([-0.6960695 ,  0.56636359,  0.44127037])
```

johnwlambert:
The angle can be 300 - 800 degrees in other cases:
```python
>>> import numpy as np
>>> import gtsam
>>> R = np.array([[ -0.999957, 0.00922903, 0.00203116],[ 0.00926964, 0.999739, 0.0208927],[ -0.0018374, 0.0209105, -0.999781 ]])
>>> R.T @ R
[...]

PR #2723, opened 2026-08-21
Fix SO3::Logmap accuracy near theta = pi (#1233)

DLuminary:
The near-pi branch of SO3::Logmap pivoted on 2 + 2*R(i,i) rather than 1 + 2*R(i,i) - tr. Those differ by exactly 1 + tr = 4*cos^2(theta/2), so the branch was exact only at theta == pi and lost up to ~1e-4 relative accuracy across the window it covered. The angle was also taken from the first-order expansion mag = pi - 2w rather than computed exactly.

Retuning the threshold does not help: it relocates the error rather than removing it, and lowering it re-exposes the failure from #886, where the generic acos/sin formula amplifies the orthogonality defect of real-world input by several orders of magnitude. That is why the two issues appeared to be in conflict.

Instead, convert R to an un-normalized quaternion using the largest-pivot method (Shepperd 1978; Shoemake 1987, as used by Eigen) and take the atan-based quaternion log (C. Hertzberg et al., Information Fusion, 2011), which is what Sophus does. Both theta = 2*atan2(|v|, w) and v/|v| are invariant to a positive rescaling of (w, v), so no normalization is needed. The branch point moves to tr > 0, i.e. theta = 120 degrees, far from both singularities, and no Taylor series is required at either end.

Max relative error over (0, pi] drops from 2.3e-4 to 4.2e-16, and becomes insensitive to where the branch point is placed: any switch between 174 and 60 degrees gives the same result to machine precision.

Tests:
  - testSO3: adds a LogmapNearPi regression test (fails on develop at 8.0e-4, passes at 4.4e-16). The near-pi exemption in LogmapDerivative is no longer needed, since the numerical derivative now matches J_r^{-1} to 2.3e-7, within the existing 1e-6 tolerance.
  - testRot3: AxisAngle2 tolerance tightened from 1e-5 to 1e-7, so that one test guards both #886 and #1233.
  - python test_Rot3: assert_almost_equal was being passed 1e-7 positionally as `decimal`, which amounted to a ~1.5 radian tolerance.

### DIFF
--- a/gtsam/geometry/Quaternion.h
+++ b/gtsam/geometry/Quaternion.h
@@ -97,49 +97,41 @@ struct traits<QUATERNION_TYPE> {
 
   /// We use our own Logmap, as there is a slight bug in Eigen
   static TangentVector Logmap(const Q& q, ChartJacobian H = {}) {
-    using std::acos;
-    using std::sqrt;
+    using std::atan;
 
-    // define these compile time constants to avoid std::abs:
-    static const double twoPi = 2.0 * M_PI, NearlyOne = 1.0 - 1e-10,
-    NearlyNegativeOne = -1.0 + 1e-10;
+    // theta = 2 * atan2(|v|, w),  omega = theta * v / |v|
+    // (C. Hertzberg et al., Information Fusion, 2011), written with atan()
+    // because it is measurably cheaper than atan2() and w >= 0 is arranged
+    // below anyway. Both theta and v/|v| are invariant to a positive rescaling
+    // of (w, v), so this stays correct when q is not of unit norm -- which
+    // matters because Rot3Q builds its quaternion straight from a rotation
+    // matrix supplied by the caller and never renormalizes.
+    //
+    // This also removes the need for the two Taylor branches the previous
+    // acos-based form required: atan is accurate as |v| -> 0, so the only
+    // degenerate case left is |v| == 0 exactly, i.e. the identity.
+    const _Scalar qw = q.w();
+    const _Scalar n = q.vec().norm();
 
     TangentVector omega;
-
-    const _Scalar qw = q.w();
-    // See Quaternion-Logmap.nb in doc for Taylor expansions
-    if (qw > NearlyOne) {
-      // Taylor expansion of (angle / s) at 1
-      // (2 + 2 * (1-qw) / 3) * q.vec();
-      omega = ( 8. / 3. - 2. / 3. * qw) * q.vec();
-    } else if (qw < NearlyNegativeOne) {
-      // Taylor expansion of (angle / s) at -1
-      // (-2 - 2 * (1 + qw) / 3) * q.vec();
-      omega = (-8. / 3. - 2. / 3. * qw) * q.vec();
+    if (n == _Scalar(0)) {
+      omega = TangentVector::Zero();
     } else {
-      // Normal, away from zero case
-      if (qw > 0) {
-        _Scalar angle = 2 * acos(qw), s = sqrt(1 - qw * qw);
-        // Important:  convert to [-pi,pi] to keep error continuous
-        if (angle > M_PI)
-          angle -= twoPi;
-        else if (angle < -M_PI)
-          angle += twoPi;
-        omega = (angle / s) * q.vec();
-      } else {
-        // Make sure that we are using a canonical quaternion with w > 0
-        _Scalar angle = 2 * acos(-qw), s = sqrt(1 - qw * qw);
-        if (angle > M_PI)
-          angle -= twoPi;
-        else if (angle < -M_PI)
-          angle += twoPi;
-        omega = (angle / s) * -q.vec();
-      }
+      // Branch exactly as the previous implementation did, so that the sign
+      // chosen at qw == 0 (an exact pi rotation, where both signs are valid
+      // logs) is unchanged. Note !(qw > 0) rather than qw < 0.
+      const bool negate = !(qw > _Scalar(0));
+      const _Scalar w = negate ? -qw : qw;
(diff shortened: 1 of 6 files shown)


dellaert:
I'll look at it in detail. Is it possible for you to create a graph as in #2106? That PR seems to be doing something similar, so please point your AI to it and compare/contrast?

DLuminary:
image

Thanks — graphs attached, generated with the same experiment as #2106.

Left: exact rotation matrices, the #1233 setup. Right: the same rotations with the matrix rounded to 6 decimals first, i.e. the #886 setup. Below 3.11 develop and #2106 share a code path, so orange covers blue there.

Compare/contrast: #2106 and this PR agree on the diagnosis — the near-π branch is a quaternion log and should be computed as one. #2106 fixes one of its two defects — the angle, (mag = π − 2w → Quaternion::Logmap) but keeps the pivot Q1 = 2 + 2*R_aa, which differs from the correct 1 + 2*R_aa − tr by exactly 1 + tr = 4cos²(θ/2) — exact only at θ = π. That residual is why the threshold had to drop to 1e-6, and the shrunken window is what puts the generic acos/sin branch back in charge of angles where it can't cope with non-orthogonal input — hence #886 returning (143/165 stress-test pairs ≥ 180°, test_axisangle → 6.0751853181, the original reported value).

That the pivot is the only thing left is testable — patch that one term in #2106 and change nothing else. At δ = 1e-3 the max error goes from 3.4e-7 to 2.1e-16. But the pivot alone isn't sufficient.

With the threshold still at 1e-6, acos/sin remains responsible for everything below 179.94° — which is where #886 lives. The pivot fix is what permits widening the window; widening it is what fixes #886. Both are needed.

So this PR is close to "#2106 + correct pivot + branch at tr > 0", and if you prefer that as a smaller diff against your branch, it's a legitimate fix and I'd be glad to send it instead. What the fuller version adds: one code path rather than four (the acos/sin branch, the #746 Taylor branch, and the two Taylor branches inside Quaternion::Logmap all go away), no normalization (atan2 is scale-invariant so the quaternion is used raw), and no small-angle series. It's also cheaper on the common path, where acos + sin is replaced by a single atan2.

image


Second graph, if useful: both formulas swept over the same axis, varying where the branch point sits. The wrong pivot bottoms out at 7.2e-10 and only in a narrow window — push the branch toward π and acos/sin breaks on real data (#886), push it away and the wrong pivot covers more ground (#1233). The two issues are the two walls of one valley. With the pivot correct the curve is flat at ~3e-16 for any branch point from 174° down to ~88°, so it stops being a tuning parameter.


Reproduction script attached (numpy + mpmath only), including a #2106 + pivot variant so the isolation above can be re-run.
[logmap_study.py](https://github.com/user-attachments/files/31328016/logmap_study.py)

dellaert (approved):
Amazing !

dellaert:
CI failure :-/

DLuminary:
The Lund expectation in TEST(Rot3, log) was the old branch's output, not the true Logmap — it's 1.2e-7 off, which is exactly the #1233 error at 179.9887°. Replaced with the exact value (verified to 1.1e-16 against a 60-digit reference; agrees with the log of the nearest true rotation to 1.3e-9). Tolerance unchanged at 1e-8. The test's own comment blamed non-orthogonality; the cause was the pivot.

Separately, and not fixed here: the GTSAM_USE_QUATERNIONS expectation differs by 2.5e-7 because Rot3Q passes a quaternion of norm 1.000000082 into traits<Quaternion>::Logmap, which assumes unit norm. Normalize first and it reproduces the SO3 value to 1.1e-16. Happy to open a separate issue.

dellaert:
> Separately, and not fixed here: the GTSAM_USE_QUATERNIONS expectation differs by 2.5e-7 because Rot3Q passes a quaternion of norm 1.000000082 into traits::Logmap, which assumes unit norm. Normalize first and it reproduces the SO3 value to 1.1e-16. Happy to open a separate issue.

Let's roll it in this PR if you're willing to fix it.

ProfFan:
Do you have benchmarks on the performance of the methods?

dellaert:
Thanks again !

DLuminary:
Benchmarks below — real Eigen types, -O2, 20k rotations sized to stay in L2, median of 9 runs, out-of-line calls to match how SO3::Logmap is actually exported.

```
                    before #2723   merged   +atan+unroll   +no-sqrt
small (1e-4..0.1)      22.5 ns    18.1 ns      13.7 ns      12.8 ns
uniform (0..pi)        39.1 ns    39.9 ns      30.5 ns      28.0 ns
near pi                27.0 ns    54.6 ns      34.9 ns      29.9 ns
```

The near-π regression is real and I looked into where it goes. It isn't atan2 — it's two things. The pivot branch writes v(i), v(j), v(k) with runtime indices, which keeps v out of registers (the old code sidestepped this by unrolling the three cases). And the branch divides w and all of v by r, which cancels: the formula is invariant to positive rescaling and r > 0, so multiplying through leaves only r², which we already have before the sqrt. Eigen has to do that division because it must return a properly scaled quaternion; fused into Logmap it's dead work.

Fixing both plus atan2 → atan gives the last column: faster than the pre-#2723 code on small angles and uniform, with near-π down to ~11% over. Output stays within 1.3e-15 of current over 900k rotations, and the existing expectations clear their tolerances by ~5 orders.

Caveats: one machine, -O2, GCC. The register-pressure effect is compiler-specific — dropping the division lets GCC hoist all nine matrix loads above the tr > 0 branch, so the pivot arms need to sit in their own function or small angles get 3× worse. Untested on Clang/MSVC. Happy to send it as a follow-up if you think it's worth the complexity.

dellaert:
If you have it anyway, send it :-)

PR #2725, opened 2026-08-21
Fix shared pointer handling in MATLAB bindings

ProfFan:
## Summary

- normalize shared-ownership declarations to the wrapper T* DSL so MATLAB and Python generate the intended shared_ptr ownership
- add named StereoPoint2Vector and Cal3_S2StereoVector bindings for MATLAB and unstable Python
- preserve existing Python iterable inputs for the new bound vectors
- add shared-pointer lifetime and batch-container regression coverage

## Dependency

Depends on borglab/wrap#200 for the MATLAB shared-pointer property-setter generator fix. This PR intentionally contains no direct wrap subtree changes.

## Testing

- cmake --build build --target python-test: 497 passed, 6 skipped, 6 subtests passed
- cmake --build build --target python-test-unstable: 11 passed, 6 subtests passed
- cmake --build build --target testSmartStereoProjectionPoseFactor.run -j6
- with borglab/wrap#200 applied: generated, compiled, and linked the stable and unstable wrappers with Octave
- with borglab/wrap#200 applied: Octave testSharedPointers and testUnstableSharedPointers
- git diff --check

### DIFF
--- a/gtsam/basis/basis.i
+++ b/gtsam/basis/basis.i
@@ -193,10 +193,10 @@ class FitBasis {
 
   static gtsam::NonlinearFactorGraph NonlinearGraph(
       const std::map<double, double>& sequence,
-      const std::shared_ptr<gtsam::noiseModel::Base>& model, size_t N);
-  static gtsam::GaussianFactorGraph::shared_ptr LinearGraph(
+      const gtsam::noiseModel::Base* model, size_t N);
+  static gtsam::GaussianFactorGraph* LinearGraph(
       const std::map<double, double>& sequence,
-      const std::shared_ptr<gtsam::noiseModel::Base>& model, size_t N);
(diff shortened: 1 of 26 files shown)


dellaert:
OK, hold up. The reason the signatures changed was that we could then use pointers instead of lambdas. That only works if the signatures are identical, so I had a PR in wrap (that can still be merged) that takes these identical signatures.

I think you wanted to look into an alternative PR in wrap, but if you want to use method pointers, I think you will need those identical signatures unless you have an other mechanism that doesn’t care about this.

discuss :-)

dellaert:
Would you also be good to say what prompted this PR, for history’s sake

ProfFan:
This is not needed because we have another path which is let wrap directly support `std::shared_ptr` etc

PR #2726, opened 2026-08-21
Add system libraries CI workflow

### LINKED ISSUE #292: [Feature Request] Allow use of system libraries for 3rdparty codebases

dellaert:
## Summary

- add a standalone workflow that can be dispatched manually or called by a future nightly workflow
- install Ubuntu system dependencies and require system Eigen, METIS, Spectra, CCOLAMD, and pybind11 providers
- exercise system Boost, TBB, and GeographicLib while explicitly excluding CUDA, cuDSS, CHOLMOD, and MKL
- build the stable and unstable C++ libraries and run the complete C++ test suite

## Validation

- actionlint .github/workflows/build-system-libraries.yml
- Ubuntu 24.04 disposable container build
- 380/380 C++ tests passed

Fixes #292

### DIFF
--- a/.github/workflows/build-system-libraries.yml
+++ b/.github/workflows/build-system-libraries.yml
@@ -1,0 +1,86 @@
+name: System Libraries CI
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: system-libraries-ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Ubuntu 24.04 system libraries
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            build-essential \
+            cmake \
+            libboost-chrono-dev \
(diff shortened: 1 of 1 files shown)


dellaert:
I stopped CI because this just adds a workflow

ProfFan (approved):

dellaert:
image

PR #2727, opened 2026-08-22
Add public CPU and CUDA SFM solver paths

dellaert:
## Summary

- add a public CPU SFM Levenberg-Marquardt optimizer with Full and Schur elimination modes, typed solver selection, and point-first ordering
- add reusable CHOLMOD support for ordinary and compact batch factors and expose the new solver choices through C++ and Python APIs
- expose CUDA SFM elimination mode, reject the reserved Full mode explicitly, and refactor CPU/CUDA timings around the public optimizers
- add CPU and CUDA tests, user guides, complete benchmark records, and concise fast-path guidance

## Performance snapshot

On an Intel Core i7-14700F with 20 physical cores and an RTX 5060 Ti, the public multifrontal CPU path completed BAL-16, BAL-88, and BAL-135 in 0.252 s, 1.001 s, and 1.423 s. CUDA Schur with dense Cholesky completed them in 0.055 s, 0.218 s, and 0.290 s. Full and Schur CPU multifrontal results were effectively tied.

The complete solver matrices, environment, commands, unsupported cases, and numerical checks are recorded in timing/README.md.

## Validation

- built timeSfmPartialElimination and timeCudaSFMBAL in Release with CUDA, cuDSS, TBB, and CHOLMOD enabled
- passed testSfmLevenbergMarquardt.run
- passed testCudaSfm.run
- passed testCholmodSolver.run
- passed testMultifrontalSolver.run
- completed three capped runs for every CPU and CUDA matrix cell
- rendered gtsam/sfm/sfm.md successfully with MyST 1.10.1 and visually checked the page
- git diff --check passes

The full strict MyST site build was also run; it remains nonzero because of unrelated existing notebook image, DOI, and cross-reference diagnostics. The affected SFM page itself builds without diagnostics.


dellaert (commented):
I looked over all the file changes and I approve this message :-)

dellaert:
@leolrg Could you possibly check this out and do the timing results on the A100? My graphics card is not too powerful, so we can change the or add the timing results, especially if they look better. 

Any other comments on this PR would be highly appreciated as well.

leolrg:
> @leolrg Could you possibly check this out and do the timing results on the A100? My graphics card is not too powerful, so we can change the or add the timing results, especially if they look better.
> 
> Any other comments on this PR would be highly appreciated as well.

Ok I will do this right now

leolrg:
<p>
  @dellaert Unfortunately, the A100 server I have access to is down at the moment.
  In the meantime, I reran the benchmark matrix on my own machine
  (RTX 5090 + Ryzen 9 5900X, 12C/24T), using the median of three separate
  invocations per cell on <code>dubrovnik-16/88/135-pre</code>.
</p>

<p><strong>CUDA <code>schur-dense</code> versus each CPU variant</strong></p>

<p>
  Times are median seconds; parentheses show speedup relative to CUDA
  <code>schur-dense</code>.
</p>

<table>
  <thead>
    <tr>
      <th>Configuration</th>
      <th>BAL-16</th>
      <th>BAL-88</th>
      <th>BAL-135</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>CUDA <code>schur-dense</code></td>
      <td>0.070</td>
      <td>0.200</td>
      <td>0.267</td>
    </tr>
    <tr>
      <td>Schur/MultifrontalSolver</td>
      <td>0.412 (5.9×)</td>
      <td>1.505 (7.5×)</td>
      <td>2.115 (7.9×)</td>
    </tr>
    <tr>
      <td>Full/MultifrontalSolver</td>
      <td>0.437 (6.2×)</td>
      <td>1.511 (7.6×)</td>
      <td>2.107 (7.9×)</td>
    </tr>
    <tr>
      <td>Schur/MultifrontalCholesky</td>
      <td>0.643 (9.2×)</td>
      <td>2.232 (11.2×)</td>
      <td>3.012 (11.3×)</td>
    </tr>
    <tr>
      <td>Schur/PCG</td>
      <td>0.678 (9.7×)</td>
      <td>2.349 (11.7×)</td>
      <td>3.029 (11.3×)</td>
    </tr>
  </tbody>
</table>

<p>
  I will rerun the same matrix on the A100 once that server is available again.
</p>

<p><strong>I also have an API/scope question.</strong></p>

<p>
  My understanding is that the CPU implementation is relatively general at the
  linear-solver level, but its automatic SFM partition specifically identifies
  values of type <code>SfmCamera</code>. 
</p>

<p>
  The CUDA implementation is currently narrower: its graph-conversion path accepts
  <code>GeneralSFMFactor&lt;SfmCamera, Point3&gt;</code> and the corresponding
  batched factor, and the CUDA kernels assume a 9-dimensional camera variable and
  a 3-dimensional landmark variable.
</p>

<p>
  This covers BAL/Bundler-style bundle adjustment well, but
  <code>GeneralSFMFactor&lt;SfmCamera, Point3&gt;</code> is only one subset of the
  SFM problems we may care about in practice. For example:
</p>

<ul>
  <li>
    With fixed intrinsics, a graph would normally use
    <code>GenericProjectionFactor&lt;Pose3, Point3, CALIBRATION&gt;</code> and
    optimize a 6-dimensional pose rather than a 9-dimensional
    <code>SfmCamera</code>.
  </li>
  <li>
    Shared or separately optimized calibration can be represented using
    <code>GeneralSFMFactor2</code>.
  </li>
  <li>
    Other SFM graphs may use different camera models, priors, constraints, or
    smart/structureless projection factors.
  </li>
</ul>

<p>
  Because of this, I am not sure whether the CPU and CUDA implementations should
  be presented as two interchangeable implementations of the same general SFM
  optimizer. They solve the same mathematical problem for the overlapping
  BAL-style case, but they currently accept different graph representations and
  have different capabilities.
</p>

  If broader CUDA support is desired, I can quickly make it more general.
</p>

dellaert:
Good point, @leolrg ! Actually, I think the CPU implementation could be made more generally immediately :-) we can do a Schur compliment on Point3 and Unit3, instead of doing the split based on a single camera type. That then allows for a global calibration as well! 

I can add that quite easily to this PR I think.

CUDA will need a different kernel for each, so then maybe you could do that as a follow up PR?

dellaert:
As for timing, let me know when you have your A100 matrix.

gtsam/sfm/doc/images/cuda-optimizers.png
ProfFan, line 1: Maybe compress/resize this file a bit? It's 1.32MB

ProfFan (approved):

PR #2728, opened 2026-08-22
Add sparse AllDiff support for hybrid inference

dellaert:
## Summary

- add a gentle hybrid data-association tutorial with a six-Gaussian Plotly visualization and FGZ diagrams for the decision tree, factor graph, Bayes net, and Bayes tree
- preserve sparse TableFactor multiplication through reverse dispatch and scope-aware allocation bounds
- add a native sparse AllDiff path for hybrid sequential and multifrontal inference
- document the established version-gated GTSAM API deprecation procedure

## Validation

- executed HybridDataAssociationTutorial.ipynb completely in the py312 environment with no cell errors
- make -j6 testTableFactor.run testHybridAllDiff.run
- verified the branch contains exactly four commits and excludes MultiObjectDataAssociation.ipynb and the unrelated .gitignore additions

PR #2729, opened 2026-08-23
Speed up discrete multiplication with generic TableFactor conversion

dellaert:
## Core discrete-factor change

This PR speeds up generic multiplication in `gtsam/discrete` by removing an unnecessary copy of the input `TableFactor`.

Previously, when a `TableFactor` encountered another `DiscreteFactor` subclass, core `TableFactor::multiply` made an owning copy of the entire input table solely to obtain a second virtual dispatch:

```cpp
const auto table = std::make_shared<TableFactor>(*this);
return factor->multiply(table);
```

The new core API asks the other factor for its most efficient table representation and multiplies it against the existing input by reference:

```cpp
return std::make_shared<TableFactor>(
    this->operator*(factor->toTableFactor()));
```

The dynamic cast and dereference borrow the existing table. Only the other factor's table representation and the unavoidable product are materialized, and the product is moved into the returned shared allocation.

To support that path generically, this PR adds virtual `DiscreteFactor::toTableFactor()` with direct core overrides for `DecisionTreeFactor`, `TableFactor`, and `TableDistribution`. `DiscreteFactor::multiply` is now a concrete table-aware fallback, while the existing specialized table/table and table/decision-tree paths remain intact.

## Core discrete performance

`AllDiff` is a sparse non-`TableFactor` implementation that exercises the generic core dispatch. The benchmark therefore uses `AllDiff` as the second operand while measuring multiplication and input-table ownership in `gtsam::TableFactor`.

`timeHybridAllDiff` compares three numerically identical routes:

1. conversion through `DecisionTreeFactor`;
2. sparse multiplication using the former core dispatch, including its input-table copy;
3. the new core direct dispatch, borrowing the input table.

Representative Release medians:

| Scope | Objects | Input nonzeros | Decision tree | Copied core table | Direct core table | Tree/direct | Copy/direct |
|---|---:|---:|---:|---:|---:|---:|---:|
| Full | 7 | 823,543 | 3,613,712 us | 68,337 us | 58,587 us | 61.68x | 1.17x |
| Partial | 7 | 2,401 | 2,615,171 us | 3,214 us | 2,997 us | 872.62x | 1.07x |

For the large full-scope core table, removing the copy reduces the median by about 14%. Small tables are close to timing noise, as expected. PR #2728 supplies the much larger decision-tree-to-sparse improvement; this follow-up generalizes that representation through the core API and removes the remaining copy.

## Hybrid inference impact

Hybrid inference consumes the new core API directly:

- discrete-factor collection calls virtual `toTableFactor()` instead of cast-or-convert blocks;
- full-clique sum-product obtains its `TableDistribution` through the same conversion;
- Bayes-tree max-product uses the same sparse conversion path;
- the always-enabled `GTSAM_HYBRID_WITH_TABLEFACTOR` macro and inactive branches are removed.

Consequently, hybrid elimination preserves sparse representations through the discrete product without bespoke knowledge of `AllDiff` or future discrete-factor subclasses. The copy-removal measurements above apply to this discrete multiplication phase when used in hybrid inference.

The core `timeHybridInference` target tracks a complete 2D one-to-one association problem using only `gtsam` APIs. Representative three-object medians were:

| Hybrid operation | Median |
|---|---:|
| Sequential Bayes-net creation (sum-product) | 78.12 us |
| Bayes-net optimization (max-product plus continuous solve) | 54.20 us |
| Multifrontal Bayes-tree creation (sum-product) | 83.72 us |
| Bayes-tree optimization (max-product plus continuous solve) | 16.29 us |

This core benchmark is checked in for continuing hybrid performance regression tracking.

## Stacked PR

This PR is stacked on and depends upon #2728. Its base is `codex/hybrid-all-diff-table`; only the two follow-up commits should be reviewed here.

## Validation

Passed in the existing Release build and in a Release build with `GTSAM_ALLOW_DEPRECATED_SINCE_V43=OFF`:

- `testTableFactor.run`
- `testDiscreteFactorGraph.run`
- `testCSP.run`
- `testHybridAllDiff.run`
- `testHybridGaussianFactorGraph.run`
- `testHybridBayesTree.run`
- `testHybridGaussianISAM.run`
- `testHybridNonlinearISAM.run`

Both timing targets compile and run in both configurations. Benchmark Action JSON output was also validated, and `git diff --check` passes.

### DIFF
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -69,10 +69,9 @@ namespace gtsam {
       const DiscreteFactor::shared_ptr& f) const {
     DiscreteFactor::shared_ptr result;
     if (auto tf = std::dynamic_pointer_cast<TableFactor>(f)) {
-      // If f is a TableFactor, we convert `this` to a TableFactor since this
-      // conversion is cheaper than converting `f` to a DecisionTreeFactor. We
-      // then return a TableFactor.
-      result = std::make_shared<TableFactor>((*tf) * TableFactor(*this));
+      // If f is a TableFactor, use the virtual conversion so derived decision
+      // tree factors can provide their actual sparse representation.
+      result = std::make_shared<TableFactor>((*tf) * toTableFactor());
 
     } else if (auto dtf = std::dynamic_pointer_cast<DecisionTreeFactor>(f)) {
       // If `f` is a DecisionTreeFactor, simply call operator*.
--- a/gtsam/discrete/DecisionTreeFactor.cpp
+++ b/gtsam/discrete/DecisionTreeFactor.cpp
@@ -89,6 +89,11 @@ namespace gtsam {
   }
 
   /* ************************************************************************ */
+  TableFactor DecisionTreeFactor::toTableFactor() const {
+    return TableFactor(*this);
+  }
+
+  /* ************************************************************************ */
   DiscreteFactor::shared_ptr DecisionTreeFactor::operator/(
       const DiscreteFactor::shared_ptr& f) const {
     if (auto tf = std::dynamic_pointer_cast<TableFactor>(f)) {
--- a/gtsam/hybrid/HybridBayesTree.cpp
+++ b/gtsam/hybrid/HybridBayesTree.cpp
@@ -47,13 +47,7 @@ DiscreteValues HybridBayesTree::discreteMaxProduct(
     const DiscreteFactorGraph& dfg) const {
   DiscreteFactor::shared_ptr product = dfg.scaledProduct();
 
-  // Check type of product, and get as TableFactor for efficiency.
-  TableFactor p;
-  if (auto tf = std::dynamic_pointer_cast<TableFactor>(product)) {
-    p = *tf;
-  } else {
-    p = TableFactor(product->toDecisionTreeFactor());
-  }
+  TableFactor p = product->toTableFactor();
   DiscreteValues assignment = TableDistribution(p).argmax();
   return assignment;
 }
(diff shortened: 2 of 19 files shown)


gtsam/discrete/DecisionTreeFactor.cpp
```diff
@@ -88,6 +88,11 @@ namespace gtsam {
     return result;
   }
 
+  /* ************************************************************************ */
+  TableFactor DecisionTreeFactor::toTableFactor() const {
+    return TableFactor(*this);
```
varunagrawal, line 92: I was under the impression that this operation was really expensive and time consuming? Is that still the case?

gtsam/hybrid/HybridBayesTree.cpp
```diff
@@ -47,13 +47,7 @@ DiscreteValues HybridBayesTree::discreteMaxProduct(
...
   DiscreteFactor::shared_ptr product = dfg.scaledProduct();
 
-  // Check type of product, and get as TableFactor for efficiency.
-  TableFactor p;
-  if (auto tf = std::dynamic_pointer_cast<TableFactor>(product)) {
-    p = *tf;
-  } else {
-    p = TableFactor(product->toDecisionTreeFactor());
-  }
+  TableFactor p = product->toTableFactor();
```
varunagrawal, line 50: Awesome!

varunagrawal (approved):
This partially implements an idea I had about making the DiscreteFactor operations more opaque. 👍🏾

gtsam/discrete/DecisionTreeFactor.cpp
dellaert: Yes, converting a general DecisionTreeFactor can still be expensive. The benefit is that specialized types such as TableFactor, TableDistribution, and AllDiff override it and avoid decision-tree expansion entirely.

PR #2706, opened 2026-08-18
Add CUDA nonlinear optimization with cuDSS and PCG backends

leolrg:
## Summary

This PR adds opt-in CUDA acceleration for nonlinear factor-graph optimization.

It introduces two complementary paths:

1. A **general sparse Levenberg-Marquardt optimizer** for factor graphs whose
   factors linearize to unconstrained Jacobian factors.
2. A **fully GPU-resident SFM optimizer** specialized for BAL-style bundle
   adjustment.

Both paths use a shared CUDA linear-solver layer supporting:

- cuDSS sparse direct solves
- matrix-free preconditioned conjugate gradient (PCG)
- dense Schur Cholesky for the specialized SFM path

CUDA remains experimental and disabled by default, so existing CPU builds and
APIs are unchanged.

### Shared CUDA infrastructure

- RAII wrappers for device arrays, pinned host memory, streams, and events
- reusable dense, sparse, and operator-based linear-system interfaces
- a shared LinearSolverSession lifecycle
- cuDSS symbolic-analysis reuse
- device-resident PCG with warm starts
- Jacobi and block-Jacobi preconditioners
- backend-independent solver statistics and profiling
- consistent naming under the gtsam::cuda namespace

### General sparse CUDA LM

gtsam::cuda::SparseLevenbergMarquardtOptimizer accepts a normal
NonlinearFactorGraph and Values.

The graph topology is compiled once into a reusable sparse-Jacobian plan.
During optimization:

- factor linearization and sparse-Jacobian packing run in parallel on the CPU;
- Jacobian numerics are transferred into a persistent GPU allocation;
- CUDA updates the transpose and applies the normal-equation operator;
- cuDSS forms and solves the sparse normal matrix, or PCG applies the
  matrix-free (J^T J + lambda D) operator;
- retraction and nonlinear trial-error evaluation remain on the CPU;
- final Values are exposed through the normal optimizer interface.

Unsupported graphs fall back to the standard CPU LM optimizer by default.
Applications that require CUDA execution can set fallbackOnUnsupported to
false.

### GPU-resident SFM LM

gtsam::cuda::SfmLevenbergMarquardtOptimizer provides a specialized path for
BAL-style bundle adjustment.

Projection linearization, Schur/full-normal system construction, damping,
linear solve, retraction, and trial-error evaluation remain on the GPU. The
host only drives the LM lambda search, and final values are downloaded once.

Supported graph factors are GeneralSFMFactor<PinholeCamera<Cal3Bundler>,
[...]

PR #2730, opened 2026-08-23
Deduplicate variable layout metadata around KeyInfo

leolrg:
## Summary
- extract `KeyInfo` into a standalone linear component and add validated construction from key/dimension pairs
- remove the duplicate `VariableBlock`, `BlockLayout`, and sparse-column layout abstractions
- migrate sparse CUDA and SFM planning to use `KeyInfo`, with checked offsets, permutations, fingerprints, and conversion back to `VectorValues`
- expand coverage for invalid orderings, overflow, stale metadata, vector-size mismatches, and SFM camera layout

This is the code-deduplication follow-up to the review discussion in #2706. It is independent of #2727; the two changes touch a pair of SFM files in separate hunks, and a synthetic merge was conflict-free.

## Test plan
- [x] GCC + CUDA, cuDSS disabled: focused 6/6 tests passed
- [x] GCC + CUDA, cuDSS enabled: focused 6/6 tests passed
- [x] Clang 18 + CUDA with `GTSAM_BUILD_WITH_WERROR=ON`: focused 6/6 tests passed
- [x] `git diff --check upstream/develop..HEAD`


dellaert (approved):
Awesome!

PR #2731, opened 2026-08-23
Revamp MATLAB bindings and wrapper interfaces

ProfFan:
## Summary

- synchronize the merged wrap generator updates for direct pybind adapters, MATLAB fixed-size Eigen handling, safer MEX error paths, and idempotent native handle deletion
- align GTSAM interface declarations with their C++ signatures and annotate APIs that require generated pybind lambdas
- make MATLAB CustomFactor callback registration, cleanup, and exception handling work safely in both MATLAB and Octave
- remove import gtsam.* from MATLAB tests, fully qualify package APIs, and externalize callbacks that Octave must resolve by name

## Testing

- fresh Release build of GTSAM, the Python extension, and the Octave MEX binding
- Python: 494 passed, 6 skipped, 1 warning, 6 subtests passed
- Octave 11.3: all 20 tests in test_gtsam passed
- wrap standalone tests: 128 passed
- git diff --check


dellaert (commented):
Looks good! I'll try to run matlab tests locally first though

dellaert:
Wrap follow-up summary

My commit removes the five reasonable overload-family lambda candidates from the interface files (six annotation lines in total, because the two Custom::Create declarations were consolidated into one defaulted declaration). The Python wrapper now builds successfully with direct function-pointer bindings for GemanMcClure, TruncatedLeastSquares, and Custom.

There are currently 38 explicit @pybind_lambda annotations remaining. The remaining wrap work should be handled in a separate PR:

- Add dedicated std::string and const std::string& handling in the MATLAB generator. This is already causing a MATLAB failure: testThinTreeBayesNet.m crashes in GaussianBayesNet.saveGraph because the generated wrapper treats std::string as a shared pointer and calls unwrap_shared_ptr<std::string>. testUtilities.m and testThinBayesTree.m pass.
- Improve direct pointer binding for const std::shared_ptr<T>& and shared-pointer aliases; insertProjectionFactors still requires a lambda.
- Add generator-side conversions for Eigen views and STL iterator/container results.
- Improve template/inherited method resolution for discrete add and Bayes-tree push_back overloads.
- Add cleaner support for optional Jacobian arguments and char conversions.

dellaert (approved):
I'll approve and merge, but see "my" comment about a follow-up PR in wrap. Does this close one of the two remaining wrapper issues?

PR #2735, opened 2026-08-23
Speed up SO3::Logmap

DLuminary:
Follow-up to #2723, which fixed the accuracy of Logmap near pi but left it about 2x slower than before on a near-pi-only workload. Two changes, neither of which alters the result by more than 8.9e-16:

1. The pivot branch wrote v(i), v(j), v(k) with runtime indices, which keeps v out of registers -- every write goes to memory and v.norm() reads it back. Writing the three cases out, as the pre-#2723 code did, avoids that.

2. theta = 2*atan2(n, w) is spelled with atan() instead. w >= 0 is arranged explicitly, so atan2's quadrant handling is redundant work.

Measured with real Eigen types at -O2, 20k rotations sized to stay in L2, median of 9 runs, out-of-line calls (SO3::Logmap is an exported symbol):

```
                       before #2723   after #2723   this commit
  small (1e-4..0.1)       22.5 ns        18.1 ns       13.6 ns
  uniform (0..pi)         39.1 ns        39.9 ns       30.6 ns
  near pi (pi-0.1..pi)    27.0 ns        54.6 ns       34.4 ns
```

so faster than the pre-#2723 code on the first two, with the near-pi case down from ~100% slower to ~27%.

Output is unchanged to within 8.9e-16 over 1.2M rotations spanning (0, pi], within 1e-6 of pi, and within 1e-6 of identity. TEST(Rot3, log) and TEST(SO3, LogmapNearPi) clear their tolerances by five orders of magnitude.

There is a third opportunity not taken here. The pivot branch divides w and all three components of v by r, and that division cancels: the formula is invariant to a positive rescaling of (w, v) and r > 0, so multiplying through leaves only r^2, which is already in hand before the sqrt. That removes a sqrt and a division and is worth a further ~2-5 ns. It is left out because with the divisions gone GCC hoists all nine matrix loads above the tr > 0 test, which costs more than it saves on small angles unless the pivot arms are forced into a separate non-inlined function -- and that needs a compiler-specific attribute GTSAM does not currently have a portable spelling for.

### DIFF
--- a/gtsam/geometry/SO3.cpp
+++ b/gtsam/geometry/SO3.cpp
@@ -335,7 +335,7 @@ Vector3 SO3::Logmap(const SO3& Q) {
 template <>
 GTSAM_EXPORT
 Vector3 SO3::Logmap(const SO3& Q, ChartJacobian H) {
-  using std::atan2;
+  using std::atan;
   using std::sqrt;
 
   const Matrix3& R = Q.matrix();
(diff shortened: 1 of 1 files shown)


dellaert (approved):
Nice!

PR #2737, opened 2026-08-24
Declare Python wrapper types before binding methods

### LINKED ISSUE #1824: Declare-then-bind for the Python wrapper for better type generation
Currently the Python type generation has a problem: many classes are not registered to pybind11 when they got used as function parameters, resulting in errors like:
```
pybind11_stubgen - [  ERROR] In gtsam.gtsam.TranslationRecovery.addPrior : Invalid expression 'gtsam::BinaryMeasurement<gtsam::Unit3>'
```

This is not an issue for `pybind11-stubgen` or `mypy`'s stub-gen. This is actually caused by this: https://pybind11.readthedocs.io/en/latest/advanced/misc.html#avoiding-c-types-in-docstrings

The solution is "declare-then-bind", which works by first only do the declaration and then after all declarations start binding the functions.

So instead of this:
```cxx
py::class_<gtsam::Value, std::shared_ptr<gtsam::Value>>(m_, "Value")
        .def("print",[](gtsam::Value* self, string s){ py::scoped_ostream_redirect output; self->print(s);}, py::arg("s") = "")
        .def("__repr__",
                    [](const gtsam::Value& self, string s){
                        gtsam::RedirectCout redirect;
                        self.print(s);
                        return redirect.str();
                    }, py::arg("s") = "")
        .def("dim",[](gtsam::Value* self){return self->dim();});
```

we should do:
```cxx
py::class_<gtsam::Value, std::shared_ptr<gtsam::Value>>(m_, "Value") gtsam_value;
py::class_<XXX>(m_, "Xxx") gtsam_xxx;
...

gtsam_value
        .def("print",[](gtsam::Value* self, string s){ py::scoped_ostream_redirect output; self->print(s);}, py::arg("s") = "")
        .def("__repr__",
                    [](const gtsam::Value& self, string s){
                        gtsam::RedirectCout redirect;
                        self.print(s);
                        return redirect.str();
                    }, py::arg("s") = "")
        .def("dim",[](gtsam::Value* self){return self->dim();});
gtsam_xxx
        .def(...);
```

Opening this here as a future TODO.

ProfFan:
## Motivation

The Python wrapper currently registers each class and immediately binds its
constructors, methods, and properties. A method can therefore reference another
wrapped class before pybind11 has registered that class.

When pybind11 constructs the Python-visible signature for such a method, it
cannot resolve the referenced type to its Python name. The resulting docstring
contains a raw C++ expression, such as:

```text
gtsam::BinaryMeasurement<gtsam::Unit3>
```

That expression is not a valid Python type annotation. Tools such as
`pybind11-stubgen` consequently report invalid expressions or generate
incomplete type information.

This is fundamentally a wrapper declaration-order problem rather than a
stub-generator parsing problem. Every wrapped class must be registered with
pybind11 before any constructors or methods using those classes are bound.

## Approach

This PR adopts a declaration-then-bind generation model:

1. Declare every pybind11 class and enum object.
2. Bind constructors, methods, properties, and operators only after all
   declarations exist.
3. Initialize submodules and handwritten specializations after the generated
   declarations are available.

This ensures pybind11 knows the Python name of every wrapped type before it
constructs method signatures and docstrings.

### Wrapper generator integration

- Sync the updated `wrap` subtree containing declaration-first pybind
  generation.
- Split generated wrapper output into declaration, binding, and module
  initialization phases.
- Update both the stable and unstable GTSAM Python templates to use these
  phases.
- Preserve the ordering required by handwritten specialization headers.

### Missing type registrations

Several types already appeared in public wrapper signatures but were not
independently registered. This PR adds the required wrapper declarations and
template instantiations for:

- Additional `CameraSet` camera families.
- `PinholePoseCal3f` and `EmptyCal`.
- `VariableSlots` and `VerticalBlockMatrix`.
- Hybrid Gaussian conditional and nonlinear factor decision trees.
- `ISAM2UpdateParams`.

These declarations allow pybind11 to use proper Python type names in signatures
and make the corresponding APIs callable from Python.

### Cross-module and container handling

- Register `JacobianVector` before binding methods whose signatures use
  `OptionalMatrixVecType`.
- Make that registration global so the stable and unstable GTSAM modules share
  the same Python type.
- Add a Python-facing conversion for `ISAM2UpdateParams::newAffectedKeys`,
  bridging GTSAM's internal `FastMap<FactorIndex, KeySet>` representation to
  standard Python-compatible map and set containers.
- Preserve optional semantics, including assigning and retrieving `None`.

### Signature consistency

- Align wrapper parameter names with their C++ declarations for carrier-phase
  factor constructors.
- Rename the wavelength parameter to `lambda_` to avoid Python's reserved
  `lambda` keyword.
- Match wrapper default arguments for unstable `print()` methods with their C++
  declarations.

## Regression coverage

New and extended tests verify that:

- Every newly exposed wrapper type can be constructed and used.
- All supported `CameraSet` instantiations accept their corresponding camera
  types.
- `PinholePoseCal3f` round-trips through `Values`.
- `VariableSlots` and `VerticalBlockMatrix` work with their dependent
  constructors.
- `ISAM2UpdateParams` scalar, container, and optional properties round-trip
  correctly.
- Hybrid decision-tree types construct multi-key hybrid conditionals and
  factors.
- `JacobianVector` receives populated Jacobians from stable factors and works
  across the stable/unstable module boundary.

## Validation

- Built `gtsam_py` and `gtsam_unstable_py` successfully.
- Stable Python suite: **510 passed, 6 skipped**.
- Unstable Python suite: **6 passed**.
- `python-stubs` and `python-unstable-stubs` completed successfully with
  `pybind11-stubgen` 2.5.5.
- Generated `.pyi` files contain no unresolved `gtsam::`, `std::`, or `boost::`
  type spellings.
- Generated and compiled the MATLAB MEX bindings with Octave 11.3.
- MATLAB/Octave suite: **20/20 passed**.
- `git diff --check` passed.

Fixes #1824.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,38 @@ pybind_wrap(${PROJECT_NAME}_py # target
 
 For more information, please follow our [tutorial](https://github.com/borglab/gtsam-project-python).
 
+### Pybind module templates
+
+Pybind templates use two generated phases so every Python type is registered
+before pybind11 constructs callable signatures. Custom templates must define
+both phase functions and invoke the generated module initializer:
+
+```cpp
+{submodules}
+
+{declaration_module_def} {{
+// Include declaration-only specializations here.
+{wrapped_declarations}
+}}
+
+{binding_module_def} {{
+// Include methods, functions, and property specializations here.
+{wrapped_bindings}
+}}
+
+{module_def} {{
+{module_init}
+}}
+```
+
+Declaration specializations may create modules, classes, or enums but must not
+bind callables. Binding specializations can recover those objects and add their
+constructors, methods, functions, and properties. See
+`templates/pybind_wrapper.tpl.example` for the complete template. Declaration
+calls follow the interface-file list passed to `pybind_wrap`, so an interface
+which declares a base class must still precede interfaces declaring its derived
+classes.
+
 ## Documentation
 
 Documentation for wrapping C++ code can be found [here](https://github.com/borglab/wrap/blob/master/DOCS.md), including the [pybind callable-adapter annotation](https://github.com/borglab/wrap/blob/master/DOCS.md#pybind-callable-adapters).
--- a/gtsam/geometry/cameras.i
+++ b/gtsam/geometry/cameras.i
@@ -500,9 +500,14 @@ typedef gtsam::PinholePose<gtsam::Cal3_S2> PinholePoseCal3_S2;
 typedef gtsam::PinholePose<gtsam::Cal3DS2> PinholePoseCal3DS2;
 typedef gtsam::PinholePose<gtsam::Cal3Unified> PinholePoseCal3Unified;
 typedef gtsam::PinholePose<gtsam::Cal3Bundler> PinholePoseCal3Bundler;
+typedef gtsam::PinholePose<gtsam::Cal3f> PinholePoseCal3f;
 typedef gtsam::PinholePose<gtsam::Cal3Fisheye> PinholePoseCal3Fisheye;
 
 #include <gtsam/geometry/SphericalCamera.h>
+class EmptyCal {
+  EmptyCal();
+};
+
 class SphericalCamera {
   // Standard Constructors
   SphericalCamera();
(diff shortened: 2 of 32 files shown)


dellaert (approved):
Nice!

PR #2742, opened 2026-08-24
Remove using namespace std from header

B-Leslie:
Removes the `using namespace std` that could pollute other code on inclusion, and make the `#include` directives self contained for the header.

The only side effects I can see would be 3rd party code that relies on the `using namespace std` transitively. I hope that's not a serious worry. I've run the `check_navigation_program` test locally and it passes.


dellaert (approved):
Fully agree! Thanks!

PR #2745, opened 2026-08-24
Use hybrid TBB gathering for multifrontal parent updates

dellaert:
## Summary

Reduce TBB parent-update overhead in `MultifrontalSolver` without regressing compact point-batched Full or Schur SFM workloads.

The gather path now selects execution by update representation:

- materialized Cholesky updates use disjoint destination-column ownership;
- QR residual updates use destination-column ownership with bounded narrow scratch;
- compact and fused updates retain child-wise compute-once accumulation.

The stages run additively rather than writing concurrently. Same-separator aggregation, small-clique sequential execution, and the non-TBB implementation are unchanged. There are no public API, solver parameter, or default changes.

## Implementation

- Add a private reusable parent-gather plan after children resolve their final representations.
- Precompute materialized and QR children contributing to each destination column.
- Assign RHS cross-blocks to variable-column owners and reduce only the scalar RHS diagonal for column-owned updates.
- Keep compact and fused children on the existing compute-once reduction path.
- Add stress coverage for ordinary Cholesky, forced QR, compact/fused leaves, mixed representations, reloads, partial elimination, mappings, RHS terms, and more than 1,024 overlapping children.
- Extend `timeMultifrontalSolver` with sampled BAL dataset, iteration, sample, and thread selection while preserving its no-argument behavior.
- Record the full acceptance procedure and update the current CPU SFM guidance.

## Correctness

Full and Schur point-batched runs matched across the original baseline, cleanup reference, and candidate:

| Dataset | Final error | LM iterations | Inner iterations |
| --- | ---: | ---: | ---: |
| BAL-16 | 18,033.914832 | 5 | 5 |
| BAL-88 | 291,581.106404 | 4 | 4 |
| BAL-135 | 377,782.056028 | 2 | 2 |

## Performance

Measured on an Intel Core i7-14700F with CPU affinity fixed to logical CPUs 0-20, a 21-thread budget, GCC 13.3, Release `-O3 -DNDEBUG`, and TBB enabled.

The formal regular-factor BAL-135 gate used seven paired alternating samples, two eliminate-and-solve iterations per sample, and an untimed warmup per invocation:

| Reference | Median s |
| --- | ---: |
| Original parent gather | 4.955640 |
| Hybrid parent gather | 2.199900 |

This is a **55.6% reduction**, exceeding the required 5% improvement. Peak full-process RSS fell from 4,065,540 KiB to 3,957,944 KiB (105.1 MiB, 2.65%).

Point-batched SFM used seven separately invoked BAL-135 samples with cyclic run ordering. The candidate remained within the 3% no-regression limit against both references:

| Mode | Candidate median s | Vs. original | Vs. cleanup |
| --- | ---: | ---: | ---: |
| Full | 1.436844 | +0.3% | +0.5% |
| Schur | 1.426606 | +0.4% | +0.6% |

BAL-16 and BAL-88 regular, Full, and Schur diagnostic medians were all faster than both references. The complete commands, commits, medians, and memory results are recorded in `timing/README.md`.

## Validation

- `testMultifrontalSolver.run`
- `timeMultifrontalSolver` build
- `timeSfmPartialElimination` build
- BAL-16 and BAL-88 diagnostic timing/correctness runs
- BAL-135 seven-sample regular, Full, and Schur acceptance protocol
- `git range-diff` confirmed the multifrontal patches were unchanged by the final rebase onto `develop`

### DIFF
--- a/gtsam/linear/MultifrontalClique.cpp
+++ b/gtsam/linear/MultifrontalClique.cpp
@@ -32,6 +32,9 @@
 #include <tbb/blocked_range.h>
 #include <tbb/enumerable_thread_specific.h>
 #include <tbb/parallel_for.h>
+#include <tbb/parallel_reduce.h>
+
+#include <functional>
 #endif
 
 #include <algorithm>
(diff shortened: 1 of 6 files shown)


gtsam/linear/MultifrontalClique.cpp
```diff
@@ -1160,34 +1273,206 @@ void MultifrontalClique::gatherUpdatesSequential() {
...
-      [&](const tbb::blocked_range<size_t>& range) {
-        auto& local = locals.local();  // Thread-local info matrix.
-        for (size_t i = range.begin(); i < range.end(); ++i) {
-          const auto& child = children[i];
-          if (!childInSameSeparatorGroup_.empty() &&
-              childInSameSeparatorGroup_[i]) {
-            continue;
+void MultifrontalClique::ParentGatherPlan::build(
+    const MultifrontalClique& parent) {
+  assert(!built);
```
ProfFan, line 1279: Avoid C-style asserts

gtsam/linear/MultifrontalClique.cpp
```diff
@@ -1160,34 +1273,206 @@ void MultifrontalClique::gatherUpdatesSequential() {
...
+    }
+
+    const bool qr = child->useQR();
+    auto& childIndices = qr ? qrChildIndices : materializedChildIndices;
+    auto& updatesByColumn = qr ? qrByColumn : materializedByColumn;
+    childIndices.push_back(childIndex);
+    for (size_t sourceBlock = 0; sourceBlock + 1 < child->parentIndices_.size();
+         ++sourceBlock) {
+      const DenseIndex column = child->parentIndices_[sourceBlock];
+      assert(column >= 0 &&
```
ProfFan, line 1306: Is column possible to be <0 in this loop? if not we should not have this

ProfFan (approved):

dellaert:
Cool. I'll resolve both comments in the follow up PR #2746.
Attached is an AI overview of this PR for history's sake.
[pr-2745-overview.html](https://github.com/user-attachments/files/31400207/pr-2745-overview.html)

PR #2746, opened 2026-08-25
Consolidate multifrontal block update plumbing

dellaert:
## Summary

- centralize physical block-column ownership and explicit-source mapped updates in the block-matrix layer
- remove temporary active-view mutation from multifrontal parent and separator updates
- simplify and hide the constructor-built TBB parent-gather plan
- consolidate Cholesky representation dispatch while retaining the measured-fast fused scalar Schur kernel

## Design

Materialized Cholesky updates now use mapped block-column primitives, QR reads explicit row and block ranges, and ordinary compact updates use explicit mapped outer products. Compact and fused children remain on the compute-once path.

The fused star path deliberately retains its scalar dot-product scatter: replacing it with general dynamic block products materially regressed BAL-88 Full point-batch timing.

Per-thread compute-once accumulators retain construction from the full symbolic block dimensions. Partial-elimination active views can be smaller than cached child destinations.

## Correctness

Passed with TBB:

- `testSymmetricBlockMatrix`
- `testHessianFactor`
- `testJacobianFactor`
- `testFixedJacobianFactor` (dynamic, binary, and ternary)
- `testBatchJacobianFactor`
- `testMultifrontalSolver`

The multifrontal suite covers ordinary, forced QR, compact and fused, mixed representations, reload, partial elimination, same-separator aggregation, and more than 1,024 children.

A TBB-disabled `testMultifrontalSolver` build and run also passes.

## BAL-88 performance

Release build with TBB, CPUs 0-20, 21 threads for regular timing, and seven alternating paired baseline/candidate samples.

| Path | Baseline median | Candidate median | Change |
| --- | ---: | ---: | ---: |
| Regular factors, 2 iterations | 0.921464 s | 0.930401 s | +0.97% |
| Full point batch | 1.018249 s | 1.011170 s | -0.70% |
| Schur point batch | 0.984561 s | 0.985988 s | +0.15% |

All paths satisfy the 3% no-regression gate. Full and Schur preserve final error 291581.106404 and 4 LM / 4 inner iterations.

Peak RSS on the regular path was effectively flat: 2,454,380 KB baseline versus 2,462,320 KB candidate.

### DIFF
--- a/gtsam/base/SymmetricBlockMatrix.cpp
+++ b/gtsam/base/SymmetricBlockMatrix.cpp
@@ -11,7 +11,8 @@
 
 /**
  * @file    SymmetricBlockMatrix.cpp
- * @brief   Access to matrices via blocks of pre-defined sizes.  Used in GaussianFactor and GaussianConditional.
+ * @brief   Access to matrices via blocks of pre-defined sizes.  Used in
+ * GaussianFactor and GaussianConditional.
  * @author  Richard Roberts
  * @date    Sep 18, 2010
  */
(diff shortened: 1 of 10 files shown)


dellaert:
AI generated overview (several prompts):
[pr-2746-overview.html](https://github.com/user-attachments/files/31399376/pr-2746-overview.html)

PR #2747, opened 2026-08-25
Expose METIS random seed on Ordering::Metis

SamanyuB910:
Ordering::Metis is deterministic for a given factor graph: MetisIndex numbers vertices in first-encounter order over the factors, and METIS_NodeND runs with default options, so every program plays one frozen draw of METIS's randomized nested dissection. The only way to redraw today is to permute factor insertion order (shuffling the intel pose graph's factors spans largest elimination clique 17-27), which is implicit and fragile.

This adds Ordering::Metis(graph, int seed), which writes METIS_OPTION_SEED into the options passed to METIS_NodeND, and a new-style Ordering::Metis(graph, std::mt19937_64& rng) overload that draws the seed from a caller-supplied generator, following the GaussianBayesNet::sample / geometry Random() RNG convention. A negative seed (and the existing 1-arg overload, unchanged) keeps the historical nullptr-options call, so default behavior is bit-identical. Keeping the best of k seeds is a cheap ordering-quality boost (best-of-30 on intel improves the largest elimination clique 27 -> 17 in under a second).

### DIFF
--- a/gtsam/inference/Ordering.cpp
+++ b/gtsam/inference/Ordering.cpp
@@ -253,7 +253,7 @@ Ordering Ordering::ColamdConstrained(const VariableIndex& variableIndex,
 }
 
 /* ************************************************************************* */
-Ordering Ordering::Metis(const MetisIndex& met) {
+Ordering Ordering::Metis(const MetisIndex& met, int seed) {
 #ifdef GTSAM_SUPPORT_NESTED_DISSECTION
   gttic(Ordering_METIS);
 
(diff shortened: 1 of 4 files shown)


dellaert (commented):
Thanks!
Correct intent, but I don't see how the rng gets used. It seems Metis has its own rng built-in? Explain how Metis handles this? Then i can suggest a better way, perhaps just a default arg.

SamanyuB910:
So METIS has its own RNG built in. Every METIS_NodeND call seeds it fresh from METIS_OPTION_SEED, and with no options it falls back to a hardcoded 4321, so every call replays the same stream and the ordering never varies.

This means GTSAM adds no randomness itself. The mt19937_64 overload just draws one integer from the caller's generator to use as the seed; I added it to match the sample(rng) convention, but let me know if there is a better way to implement

dellaert:
Then I think the only thing we needed is an additional argument with a default of 4321. The rng is not needed.

dellaert (approved):
LGTM!

PR #2748, opened 2026-08-25
Aggregate non-TBB leaf traversal tasks

dellaert:
## Summary

- honor `leafAggregationProblemSize` in the non-TBB bottom-up traversal
- group adjacent sibling leaves into size-bounded jobs on the existing persistent `TaskScheduler`
- preserve the original one-continuation-per-child path when aggregation is disabled
- propagate exceptions raised within aggregated leaf jobs through the traversal state

No public API, parameter, or default changes are introduced. The TBB implementation is unchanged.

## Design

The non-TBB traversal previously accepted `leafAggregationProblemSize` but ignored it, creating one scheduler task per eligible leaf. This change scans adjacent sibling leaves using the same problem-size accounting as the TBB path and visits each batch in one scheduler job.

A dispatch sentinel prevents inline completions from running a parent before all sibling units have been submitted. Singleton leaves continue through the ordinary threshold-controlled path.

## Correctness

Passed without TBB:

- `testTaskScheduler`
- `testForestTraversal`
- `testMultifrontalSolver`

Passed with TBB:

- `testForestTraversal`
- `testMultifrontalSolver`

Focused tests verify that aggregated leaves run on scheduler workers, every node preserves bottom-up ordering, and exceptions propagate to the caller.

BAL-88 Full and Schur preserve final error `291581.106404` and 4 LM / 4 inner iterations.

## BAL-88 performance

Release build without TBB, CPUs 0-20, 21 workers, and seven paired samples alternating baseline/candidate execution order.

| Path | Baseline median | Candidate median | Change |
| --- | ---: | ---: | ---: |
| Regular factors, 2 iterations | 3.046130 s | 3.018570 s | -0.9% |
| Full point batch | 2.174888 s | 1.795552 s | -17.4% |
| Schur point batch | 2.151441 s | 1.781247 s | -17.2% |

### DIFF
--- a/gtsam/base/tests/testForestTraversal.cpp
+++ b/gtsam/base/tests/testForestTraversal.cpp
@@ -17,6 +17,8 @@
 #include <gtsam/base/ForestTraversal.h>
 
 #include <memory>
+#include <stdexcept>
+#include <thread>
 #include <vector>
 
 using namespace gtsam;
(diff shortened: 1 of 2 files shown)


gtsam/base/tests/testForestTraversal.cpp
```diff
@@ -109,6 +111,76 @@ TEST(ForestTraversal, BottomUpLeafAggregation) {
   EXPECT_LONGS_EQUAL(7, root->subtreeCount);
 }
 
+#ifndef GTSAM_USE_TBB
+namespace {
+
+struct AggregationNode {
+  int problemSize_ = 1;
```
ProfFan, line 118: Is this used anywhere?

ProfFan (approved):

gtsam/base/tests/testForestTraversal.cpp
ProfFan: `problemSize_`

dellaert: I thought it was. If not, I'll remove it in the follow-up PR.

PR #2749, opened 2026-08-25
Fix unreferenced values in IncrementalFixedLagSmoother

### LINKED ISSUE #2741: IncrementalFixedLagSmoother: bare `map::at` when a key is in Values but absent from keyTimestampMap_ — is this in scope for #2474?
### Summary

On `IncrementalFixedLagSmoother`, we hit `std::out_of_range` with the message `map::at` when a key is present in `Values` but has no entry in `keyTimestampMap_`. We believe **we** create that state, so this is not filed as "GTSAM is broken". Two narrower questions:

1. **Is it a supported caller state to pass a value in `newTheta` that no factor in `newFactors` references, and that is never subsequently connected?** `ISAM2` appears to accept this silently. See "Why #2474 cannot reach our case" below — we have now traced this in the 4.3a2 source and it is a concrete question about a specific code path, answerable without our data.
2. **Can the failure be made diagnosable?** Today it is a bare `map::at` with no indication of which key or which map, which cost us several days to localise.

Happy to be told this is entirely our bug — a documented precondition would be a perfectly good outcome.

### Environment

| | |
|---|---|
| GTSAM | `4.3a1` and the `4.3a2` tag (`835f791f`), built side-by-side |
| Build | Release / RelWithDebInfo, `GTSAM_USE_TBB=ON`, both `GTSAM_SLOW_BUT_CORRECT_*=OFF` |
| Platform | aarch64 (Jetson), Ubuntu 22.04, GCC 11, C++17 |
| Smoother | `IncrementalFixedLagSmoother`, 1.0 s lag |

### What we observe

`std::out_of_range("map::at")`, thrown from inside `update()`, on the same `update()` call as a marginalisation.

libstdc++ distinguishes the containers by message — `std::map::at` gives `"map::at"`, `std::unordered_map::at` gives `"_Map_base::at"` — so with `GTSAM_USE_TBB=ON` the Bayes-tree `nodes_` map (`tbb::concurrent_unordered_map`) is excluded. That leaves two `std::map`-backed candidates on this path, and notably **both are touched by #2474**:

- `FixedLagSmoother.h:41/:114` — `typedef std::map<Key,double> KeyTimestampMap; KeyTimestampMap keyTimestampMap_;`
- `ISAM2UpdateParams.h:40` — `std::optional<FastMap<Key,int>> constrainedKeys`, and `FastMap : public std::map` (`FastMap.h:39`), which is a `std::map` regardless of TBB. This feeds the `ColamdConstrained` call that #2474's second hunk rewrote.

We have not distinguished which of the two it is.

### The state we believe triggers it (and we appear to create it)

Our estimator commits `Values` **and** timestamps for every key referenced by a newly built factor, and *afterwards* drops any factor that still references a missing key. A dropped factor's keys keep their value and timestamp but lose their only factor. With no factor, the key gets no clique and nothing ever marginalises it — while marginalisation does erase timestamp entries as the lag window advances. Values and timestamps then diverge permanently for that key.

Our own admission test is values-only (`estimate.exists(key)`), so a key that is retrievable from `Values` while absent from `keyTimestampMap_` passes it, the factor is admitted, and the smoother is then asked for a timestamp that is not there.

**That is our defect to fix, and we are fixing it.** We are reporting it because the resulting *state* — a variable with a timestamp and no clique — is the state #2474 addresses, even though (as traced below) the route by which we reach it puts us outside the path that PR patches.

### Why we think #2474 is relevant, and what we measured

#2474 ("Fix IncrementalFixedLagSmoother with factor removal") describes a caller that adds values frequently, retains only some, and removes the rest **by deleting all the factors connecting them to the graph**. At the level of *caller intent* that is our pattern — but the mechanism differs in a way that turns out to matter, and that difference is the point of this report. We verified the fix is physically present in our 4.3a2 build rather than inferring it from a version string — `IncrementalFixedLagSmoother.cpp:118` `eraseKeyTimestampMap(unusedKeys)`, `:127` the `marginalizableKeys` filter, and 13 `unusedKeys` references in `ISAM2.cpp`.

A/B on one deterministic input, same session, serial on an idle machine, **version as the only variable** (flag sets differ only by `GTSAM_USE_MOSEK`, which does not exist in 4.3a1 and is OFF). Both arms self-attributed from `/proc/<pid>/maps`:

| metric | 4.3a1 | 4.3a2 |
|---|---|---|
| `map::at` | 20 | 16 |
| `Index out of requested size range` | **8** | **8** |
| total GTSAM exceptions | 28 | 24 |
| keyframes | 3953 | 3925 |
| runtime library | `libgtsamRelWithDebInfo.so.4.3a1` | `libgtsam.so.4.3a2` |

**We are explicitly not claiming 20 → 16 is a reduction.** Run-to-run scatter on this exact input and binary is a factor of two (an earlier session measured 40 where this one measured 20), so a 20 % move is well inside noise. n=1 per arm.

The row we would highlight is the second one: there is a **second exception class**, `Index out of requested size range`, **identical at 8 in both arms**. Whatever is going wrong on this path, #2474 moves neither count materially.

### On rarity

#2474's own test appears to expose the bugs only when factor removal makes keys unused *simultaneously with* marginalisation **and** a large fraction of keys is affected, triggering a batch recalculation that rebuilds the ordering through `constrainedKeys`. That matches what we see: the throw needs two coincidences, not one. One of our runs manufactured the unused-key condition 220 times and threw zero, which is why it looked luck-dependent for a long time.

Consistent with that, we can vary the throw rate ~20× **at constant factor count** by changing only *which* measurements are dropped — decimating scattered frames (maximising track churn) throws far more often than dropping contiguous blocks of the same total. That is an indirect way of tuning "fraction of keys affected per update", and it is how we made the failure reproducible at all.

### Why #2474 cannot reach our case (traced in the 4.3a2 source)

`ISAM2-impl.h:175-189`:

```cpp
void computeUnusedKeys(const NonlinearFactorGraph& newFactors,
                       const VariableIndex& variableIndex,
                       const KeySet& keysWithRemovedFactors,
                       KeySet* unusedKeys) const {
  KeySet removedAndEmpty;
  for (Key key : keysWithRemovedFactors) {          // <-- the gate
    if (variableIndex.empty(key))
      removedAndEmpty.insert(removedAndEmpty.end(), key);
  }
  ...
}
```

and `keysWithRemovedFactors` is populated only from explicitly removed factors — `ISAM2-impl.h:158-170`, `updateParams_.removeFactorIndices` → `removedFactors` → `*keysWithRemovedFactors = removedFactors.keys()`.

So a key can enter `unusedKeys` **only if a factor referencing it was removed in the same `update()` call**. Our orphan cannot: the value arrives in `newTheta`, and the factor that would have referenced it is dropped by our own guard *before* the graph reaches `ISAM2`. Nothing is removed, so `keysWithRemovedFactors` never contains it, `unusedKeys` never contains it, and `eraseKeyTimestampMap(unusedKeys)` never fires. The same set drives #2474's `marginalizableKeys` and `constrainedKeys` filters, so all three of its fixes miss this for one reason.

Put briefly:

> **#2474 handles a value orphaned by *removing* its factors. We produce a value orphaned by *never adding* its factor.** Both end as a variable with a timestamp and no clique; only the first is visible to the path #2474 patches.

That is consistent with the 16 throws we still measure on the tag.

### What we cannot provide, and a suggested way in

We do not have a minimal reproducer. Ours is a large ROS 2 visual-inertial system on a proprietary dataset, which is not something you can run.

The most economical path we can see is to extend the test added in #2474 (`testIncrementalFixedLagSmoother.cpp`) with a value that is orphaned **by absence of a referencing factor rather than by removal of one**: pass a key in `newTheta` that no factor in `newFactors` references, never connect it, and let marginalisation advance past it. Our reading is that this leaves `Values` and `keyTimestampMap_` divergent for that key, with nothing on the `unusedKeys` path able to notice.

### What would help most

1. Confirmation of whether a key in `Values` with no `keyTimestampMap_` entry is an **unsupported caller state** (in which case: documenting the precondition would be plenty), or something the smoother is meant to tolerate.
2. A diagnosable failure instead of a bare `map::at` — even just the key and the map name in the exception text. That single change would have saved us most of a week.

**Scope of what is verified here:** `computeUnusedKeys`, the provenance of `keysWithRemovedFactors`, and the conclusion that our orphan cannot enter `unusedKeys` are all read from the 4.3a2 source. That this *specific* orphaned key is the one the smoother asks for at the moment of the throw is **not** instrumented — it is inference from the state we know we create.

We are happy to run patches against our reproducer and report back; we already have the 4.3a1/4.3a2 harness standing and it turns around in about eight minutes.

tandede:
## Summary

Allow `IncrementalFixedLagSmoother` to accept values before their factors arrive, while ensuring that values which remain unconnected are removed once they leave the fixed-lag window.

Fixes #2741.

## Problem

`ISAM2` accepts a value in `newTheta` even when no factor in the same update references it. Such a value remains in the linearization point, but it has no `VariableIndex` entry or Bayes-tree clique until a later factor connects it.

The smoother previously sent every expired timestamped key through ordering and Bayes-tree marginalization. An expired unconnected value therefore reached code that requires a clique and failed instead of being reaped.

## Implementation

The update path separates expired keys into two groups:

- connected keys continue through the existing constrained-ordering and `marginalizeLeaves` path;
- pending values with no factor or clique are removed directly from ISAM2 state once they age out.

A pending value remains available while it is inside the lag window, including when a factor connects it in a later update. Keys referenced by factors in the current update are treated as active, so a newly connected value follows the normal update and marginalization path.

The internal variable cleanup retains the existing unused check for indexed variables while also supporting pending values that never had a `VariableIndex` entry.

The separate timestamp-only `map::at` bookkeeping issue is intentionally left out of this PR.

## Regression coverage

The focused tests cover:

1. adding a value without a factor and connecting it before it ages out;
2. allowing an unconnected value to age out and verifying that both its ISAM2 value and timestamp are removed.

## Validation

- `testIncrementalFixedLagSmoother` — passed
- `testBatchFixedLagSmoother` — passed
- `testGaussianISAM2` — passed
- `testVisualISAM2` — passed
- clang-format check over changed lines — passed
- `git diff --check` — passed

### DIFF
--- a/gtsam/nonlinear/FixedLagSmoother.cpp
+++ b/gtsam/nonlinear/FixedLagSmoother.cpp
@@ -75,8 +75,13 @@ void FixedLagSmoother::updateKeyTimestampMap(const KeyTimestampMap& timestamps) 
 /* ************************************************************************* */
 void FixedLagSmoother::eraseKeyTimestampMap(const KeyVector& keys) {
   for(Key key: keys) {
+    const KeyTimestampMap::iterator keyIter = keyTimestampMap_.find(key);
+    if (keyIter == keyTimestampMap_.end()) {
+      continue;
+    }
+
     // Erase the key from the Timestamp->Key map
-    double timestamp = keyTimestampMap_.at(key);
+    double timestamp = keyIter->second;
 
     TimestampKeyMap::iterator iter = timestampKeyMap_.lower_bound(timestamp);
     while(iter != timestampKeyMap_.end() && iter->first == timestamp) {
@@ -87,7 +92,7 @@ void FixedLagSmoother::eraseKeyTimestampMap(const KeyVector& keys) {
       }
     }
     // Erase the key from the Key->Timestamp map
-    keyTimestampMap_.erase(key);
+    keyTimestampMap_.erase(keyIter);
   }
 }
 
--- a/gtsam/nonlinear/FixedLagSmoother.cpp
+++ b/gtsam/nonlinear/FixedLagSmoother.cpp
@@ -75,13 +75,8 @@ void FixedLagSmoother::updateKeyTimestampMap(const KeyTimestampMap& timestamps) 
 /* ************************************************************************* */
 void FixedLagSmoother::eraseKeyTimestampMap(const KeyVector& keys) {
   for(Key key: keys) {
-    const KeyTimestampMap::iterator keyIter = keyTimestampMap_.find(key);
-    if (keyIter == keyTimestampMap_.end()) {
-      continue;
-    }
-
     // Erase the key from the Timestamp->Key map
-    double timestamp = keyIter->second;
+    double timestamp = keyTimestampMap_.at(key);
 
     TimestampKeyMap::iterator iter = timestampKeyMap_.lower_bound(timestamp);
     while(iter != timestampKeyMap_.end() && iter->first == timestamp) {
@@ -92,7 +87,7 @@ void FixedLagSmoother::eraseKeyTimestampMap(const KeyVector& keys) {
       }
     }
     // Erase the key from the Key->Timestamp map
-    keyTimestampMap_.erase(keyIter);
+    keyTimestampMap_.erase(key);
   }
 }
 
--- a/gtsam/nonlinear/ISAM2.cpp
+++ b/gtsam/nonlinear/ISAM2.cpp
@@ -413,7 +413,16 @@ void ISAM2::addVariables(const Values& newTheta,
 void ISAM2::removeVariables(const KeySet& unusedKeys) {
   gttic(removeVariables);
 
-  variableIndex_.removeUnusedVariables(unusedKeys.begin(), unusedKeys.end());
+  // Values added before any factor references them have no VariableIndex
+  // entry, but still need the remaining ISAM2 state removed.
+  KeySet indexedUnusedKeys;
+  for (Key key : unusedKeys) {
+    if (variableIndex_.find(key) != variableIndex_.end()) {
+      indexedUnusedKeys.insert(key);
+    }
+  }
+  variableIndex_.removeUnusedVariables(indexedUnusedKeys.begin(),
+                                       indexedUnusedKeys.end());
   for (Key key : unusedKeys) {
     delta_.erase(key);
     deltaNewton_.erase(key);
--- a/gtsam/nonlinear/ISAM2.h
+++ b/gtsam/nonlinear/ISAM2.h
@@ -374,6 +374,8 @@ class GTSAM_EXPORT ISAM2 : public BayesTree<ISAM2Clique> {
    */
   void removeVariables(const KeySet& unusedKeys);
 
+  friend class IncrementalFixedLagSmoother;
+
   void updateDelta(bool forceFullSolve = false) const;
 
  private:
--- a/gtsam/nonlinear/IncrementalFixedLagSmoother.cpp
+++ b/gtsam/nonlinear/IncrementalFixedLagSmoother.cpp
@@ -57,18 +57,14 @@ FixedLagSmoother::Result IncrementalFixedLagSmoother::update(
   FastVector<size_t> removedFactors;
   std::optional<FastMap<Key, int> > constrainedKeys = {};
 
-  // ISAM2 requires every new value to occur in a new factor. Treat values
-  // outside that contract as unused instead of leaving them without a
-  // VariableIndex entry, which would make later constrained ordering fail.
   const KeySet newFactorKeys = newFactors.keys();
-  KeySet unreferencedNewKeys;
-  Values referencedNewTheta;
-  for (const auto& keyValue : newTheta) {
-    if (newFactorKeys.find(keyValue.key) != newFactorKeys.end() ||
-        isam_.valueExists(keyValue.key)) {
-      referencedNewTheta.insert(keyValue.key, keyValue.value);
-    } else {
-      unreferencedNewKeys.insert(keyValue.key);
+  for (const auto& keyTimestamp : timestamps) {
+    const Key key = keyTimestamp.first;
+    if (!isam_.valueExists(key) && !newTheta.exists(key)) {
+      throw std::invalid_argument(
+          "IncrementalFixedLagSmoother::update: timestamp provided for key " +
+          DefaultKeyFormatter(key) +
+          ", but no value for that key exists in ISAM2 or newTheta");
     }
   }
 
@@ -85,25 +81,18 @@ FixedLagSmoother::Result IncrementalFixedLagSmoother::update(
   KeyVector marginalizableKeys = findKeysBefore(
       current_timestamp - smootherLag_);
 
-  if (not unreferencedNewKeys.empty()) {
-    // The unreferenced values are not forwarded to ISAM2, so remove their
-    // timestamps and keep them out of ordering and marginalization as well.
-    marginalizableKeys.erase(
-        std::remove_if(marginalizableKeys.begin(), marginalizableKeys.end(),
-                       [&](const auto& key) {
-                         return unreferencedNewKeys.find(key) !=
-                                unreferencedNewKeys.end();
-                       }),
-        marginalizableKeys.end());
-
-    KeyVector unreferencedTimestampKeys;
-    for (Key key : unreferencedNewKeys) {
-      if (keyTimestampMap_.find(key) != keyTimestampMap_.end()) {
-        unreferencedTimestampKeys.push_back(key);
-      }
-    }
-    eraseKeyTimestampMap(unreferencedTimestampKeys);
+  // A value can be added before any factor references it. Keep such pending
+  // values and their timestamps, but do not include them in constrained
+  // ordering or marginalization until a factor connects them.
+  KeySet activeKeys = newFactorKeys;
+  const VariableIndex& variableIndex = isam_.getVariableIndex();
+  for (const auto& keyFactors : variableIndex) {
+    activeKeys.insert(keyFactors.first);
   }
+  marginalizableKeys.erase(
+      std::remove_if(marginalizableKeys.begin(), marginalizableKeys.end(),
+                     [&](Key key) { return !activeKeys.exists(key); }),
+      marginalizableKeys.end());
 
   if (debug) {
     std::cout << "Marginalizable Keys: ";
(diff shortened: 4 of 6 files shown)


DLuminary:
I've been looking at the same issue independently and ran a couple of things against current develop that seem worth having here.

**I don't think this closes #2741.** The filter iterates `newTheta`, but a
timestamp can be supplied for a key that has no value *and* no factor. Such a
key never appears in `newTheta`, so it isn't in `unreferencedNewKeys` — yet
`updateKeyTimestampMap` still records it and `findKeysBefore` still returns it,
so it reaches `nodes_.at` with no clique and throws exactly as before:

```
    ts[X(0)] = 0.0
    ts[Symbol('p', 0)] = 0.0        # no value, no factor
    smoother.update(g, v, ts)
    # advance past the lag -> IndexError: map::at
```

That's a second route into the same state, and it doesn't pass through
`newTheta` at all. (Read from the diff — I haven't built your branch.)

**It also regresses a pattern that works today** — supplying a value in one
update and connecting it in a later one:

```
    # update 1: value for X(1), no factor references it
    s.update(g_with_prior_on_X0, v_with_X0_and_X1, ts)
    # update 2: now connect it
    s.update(graph_with_Between(X(0), X(1)), Values(), KTM())   # no exception today
```

As I read it the filter drops X(1) at update 1, so update 2 fails with
`Attempting to at the key "x1", which does not exist in the Values` — naming the
factor's key rather than the value discarded three updates earlier. Worth a
regression test for that pattern either way.

How about taking a different approach —
partitioning `marginalizableKeys` on clique existence rather than filtering
`newTheta` — which covers both routes and keeps the deferred case working.

tandede:
> I've been looking at the same issue independently and ran a couple of things against current develop that seem worth having here.
> 
> **I don't think this closes #2741.** The filter iterates `newTheta`, but a timestamp can be supplied for a key that has no value _and_ no factor. Such a key never appears in `newTheta`, so it isn't in `unreferencedNewKeys` — yet `updateKeyTimestampMap` still records it and `findKeysBefore` still returns it, so it reaches `nodes_.at` with no clique and throws exactly as before:
> 
> ```
>     ts[X(0)] = 0.0
>     ts[Symbol('p', 0)] = 0.0        # no value, no factor
>     smoother.update(g, v, ts)
>     # advance past the lag -> IndexError: map::at
> ```
> 
> That's a second route into the same state, and it doesn't pass through `newTheta` at all. (Read from the diff — I haven't built your branch.)
> 
> **It also regresses a pattern that works today** — supplying a value in one update and connecting it in a later one:
> 
> ```
>     # update 1: value for X(1), no factor references it
>     s.update(g_with_prior_on_X0, v_with_X0_and_X1, ts)
>     # update 2: now connect it
>     s.update(graph_with_Between(X(0), X(1)), Values(), KTM())   # no exception today
> ```
> 
> As I read it the filter drops X(1) at update 1, so update 2 fails with `Attempting to at the key "x1", which does not exist in the Values` — naming the factor's key rather than the value discarded three updates earlier. Worth a regression test for that pattern either way.
> 
> How about taking a different approach — partitioning `marginalizableKeys` on clique existence rather than filtering `newTheta` — which covers both routes and keeps the deferred case working.

Thanks for the careful review. Both points are correct, and the suggested approach of partitioning the marginalization path instead of filtering `newTheta` is the right one.

I replaced the original filtering approach. Values without a referencing factor now remain in ISAM2, so they can still be connected by a later update. While pending, they are excluded from constrained ordering, marginalization, and Bayes-tree re-elimination until they have an existing VariableIndex entry or are referenced by a factor in the current update.

I also added an early, key-specific diagnostic for timestamps that have no corresponding value, and made timestamp cleanup a no-op when an unused variable was legitimately added without a timestamp.

The regression coverage now includes immediate delayed connection, remaining pending while other variables are marginalized, connection after exceeding the lag, timestamp-only input, and removal of an untimestamped value.

The updated implementation is in `24ffb82ef`. The IncrementalFixedLagSmoother, BatchFixedLagSmoother, GaussianISAM2, and VisualISAM2 tests all pass locally.

dellaert:
@DLuminary I'll let you review further :-) Take a look at the test - it seems rather verbose. I ran CI.

dellaert:
PS @tandede - I appreciate the PR but please know we prefer PRs from GTSAM users that have encountered real issues. This PR seems primarily AI generated, perhaps from an open issue - and that is **not** what we want. We want at least *some* understanding of the GTSAM code base and some taste on how to fix things. If you are indeed *not* a GTSAM user, please work with @DLuminary to land this PR. Note that many open source maintainers might feel the same way.

dellaert:
Thanks, this clarifies the choices. I think “accept it, reap it when it ages out” has the right semantics.

A legitimate use case is to insert values at a known frequency while factors arrive asynchronously or conditionally. A value may therefore remain unconnected for some time, but if it never becomes connected it should simply disappear once it leaves the fixed-lag window. I would prefer preserving that behavior over rejecting such values at insertion.

Also agreed that the map::at issue you identified is separate; let’s not widen this PR to address it. Please keep the PR focused on correctly supporting and eventually reaping these temporarily unconnected values.

tandede:
> Thanks, this clarifies the choices. I think “accept it, reap it when it ages out” has the right semantics.
> 
> A legitimate use case is to insert values at a known frequency while factors arrive asynchronously or conditionally. A value may therefore remain unconnected for some time, but if it never becomes connected it should simply disappear once it leaves the fixed-lag window. I would prefer preserving that behavior over rejecting such values at insertion.
> 
> Also agreed that the map::at issue you identified is separate; let’s not widen this PR to address it. Please keep the PR focused on correctly supporting and eventually reaping these temporarily unconnected values.

Thanks, that distinction makes sense. I updated the PR to follow the “accept it, reap it when it ages out” semantics.

Values without factors now remain available while they are inside the fixed-lag window, so a later asynchronous or conditional factor can still connect them. If they remain unconnected after aging out, they are removed directly from the ISAM2 state without entering the Bayes-tree marginalization path. Connected keys continue through the normal constrained-ordering and `marginalizeLeaves` path.

I also removed the timestamp-only `map::at` handling and its tests from this PR to keep that issue separate. The regression coverage is now focused on connecting a pending value before expiry and reaping one after expiry.

The update is in `e5a2ed7`. The IncrementalFixedLagSmoother, BatchFixedLagSmoother, GaussianISAM2, and VisualISAM2 tests all pass locally.

dellaert:
@DLuminary ready to go?

dellaert (approved):
I looked over it and LGTM. @DLuminary if you see anything you don't like we can have a follow-up PR.

PR #2751, opened 2026-08-25
Avoid full parent scratch matrices without TBB

### LINKED ISSUE #2734: Multifrontal solver: remaining performance work after #2350
Follow-up to #2350 after re-evaluating the current `develop` implementation. Most of the original TODOs have either been implemented, superseded by newer machinery, or no longer look likely to produce a measurable end-to-end speedup.

The remaining performance work that still looks actionable from code inspection is:

- [x] Reduce the cost of parallel parent updates, especially the full per-worker `SymmetricBlockMatrix` scratch/merge strategy in `MultifrontalClique::gatherUpdatesParallel()`.
- [x] Bring the non-TBB `TaskScheduler` path to parity with the TBB path for leaf aggregation / intra-parent parallel work, reusing the persistent scheduler rather than introducing separate nested parallel mechanisms.
- [x] Benchmark and tune thread count, `problemSize` thresholds, and aggregation parameters across representative graph families, and use the results to select robust defaults.

The goal for all three items is measurable wall-clock improvement on representative workloads, not just local microbenchmark wins. See the follow-up comment for a concrete plan for each TODO.

dellaert:
## Summary

This brings the shared multifrontal parent-gather plan to non-TBB builds. Ordinary Cholesky and QR children now update disjoint physical parent columns instead of allocating and reducing one full parent-sized matrix per worker.

Compact and fused Schur children deliberately remain on the existing compute-once reduction path so their expensive contributions are evaluated only once. The non-TBB backend retains a bounded short-lived thread team because the traversal scheduler does not provide a safe cooperative wait for nested work.

No public API, solver parameter, threshold, or default changes.

## BAL-88 performance

Compared with `9adbeecb5` using seven paired alternating samples, fixed CPU affinity, the same thread budget, and TBB disabled:

| Configuration | Median runtime change | Result |
|---|---:|---|
| Regular | **-66.7%** | **About 3x throughput** |
| Full point-batch | +2.61% | Within 3% gate |
| Schur point-batch | -0.23% | No regression |

The Regular run also reduced peak RSS by **196 MiB**, confirming that ordinary and QR parent gathering no longer creates full parent-sized worker matrices.

A smaller TBB regression check remained stable:

| Configuration | Median runtime change |
|---|---:|
| Regular | -2.23% |
| Full point-batch | +0.97% |
| Schur point-batch | +0.52% |

## Implementation

- Reuse `ParentGatherPlan` and the materialized/QR range operations in both backends.
- Statically partition materialized columns, QR scratch preparation, and RHS reduction over a bounded non-TBB worker batch.
- Count the calling traversal worker toward the requested thread count.
- Join before the QR scratch-combination phase.
- Allocate full parent-sized local matrices only for compact/fused compute-once children.
- Preserve same-separator exclusion, lazy plan reuse across numerical reloads, sequential small-clique behavior, and all numerical representations.
- Extend forced-QR stress coverage to numerical reload and partial elimination.

## Validation

- TBB `testMultifrontalSolver`: passed
- Non-TBB `testMultifrontalSolver`: passed
- BAL-88 paired benchmarks: passed the 3% non-regression gate for Regular, Full, and Schur
- `clang-format` applied to edited C++ files
- `git diff --check`: clean

Fixes #2734 (because tuning, according to report I attached there, yielded no change in defaults).

### DIFF
--- a/gtsam/base/tests/testForestTraversal.cpp
+++ b/gtsam/base/tests/testForestTraversal.cpp
@@ -115,13 +115,12 @@ TEST(ForestTraversal, BottomUpLeafAggregation) {
 namespace {
 
 struct AggregationNode {
-  int problemSize_ = 1;
   bool throwOnVisit = false;
   bool visited = false;
   std::thread::id visitThread;
   std::vector<std::shared_ptr<AggregationNode>> children;
 
-  int problemSize() const { return problemSize_; }
+  int problemSize() const { return 1; }
 
   void visit() {
     visited = true;
--- a/gtsam/linear/MultifrontalClique.cpp
+++ b/gtsam/linear/MultifrontalClique.cpp
@@ -40,8 +40,10 @@
 #include <algorithm>
 #include <cassert>
 #include <cmath>
+#include <exception>
 #include <iostream>
 #include <thread>
+#include <utility>
 
 namespace gtsam {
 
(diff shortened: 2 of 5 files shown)


dellaert:
@ProfFan asked Sol to add a lot more comments for review and readability. `problemSize_` in test is gone

ProfFan (approved):

PR #2616, opened 2026-08-01
Certifiable Wahba Problem and Landmark Localization

avinashresearch1:
## Summary

Adds certifiable Wahba and known-landmark localization workflows, with frame-explicit factors, exact QCQP conversions, and C++/Python tutorials.

Compared with `develop`, this PR changes 27 files (`+6686/-77`).

## Core factors and wrappers

- Adds `KnownLandmarkFactor<Pose2/Pose3>` for conventional GTSAM `wTk` states:
  `wTk.transformTo(wL) - measured_kP`.
- Adds `KnownLandmarkFactor2<Pose2/Pose3>` for inverse `kTw` states, with an exact D=1 QCQP conversion for certifiable localization.
- Adds `WahbaFactor` for `Rot3` and `Unit3` direction correspondences using the three-dimensional chordal residual
  `aRb * bDirection - measured_aDirection`, also with an exact D=1 QCQP conversion.
- Adds `FrobeniusLeftBetweenFactor` for left-composed relative measurements satisfying
  `iTw = iTj * jTw`, including exact D=1 QCQP support.
- Exposes Pose2/Pose3 wrappers for both known-landmark conventions and a direct `gtsam.WahbaFactor` wrapper. `RotateDirectionsFactor` remains the two-dimensional `Unit3` tangent-error alternative.

## Dataset support

3D g2o landmark parsing and writing, fixtures, deterministic dataset generation, committed datasets, and their tests were split out and merged in #2754.

## Examples and tutorials

- Adds a header-only C++ data-preparation helper that reads a 3D g2o file, converts track factors into conventional and certifiable factor graphs, selects a bounded pose/landmark subset, remaps sparse landmark keys to dense indices, and preserves frame-bearing names.
- Adds one C++ certifiable-localization example comparing Gauss-Newton with monolithic and chordal SDP solves when MOSEK is enabled.
- Adds executed Plotly notebooks for:
  - conventional known-landmark localization;
  - certifiable inverse-state landmark localization; and
  - a certifiable Wahba problem comparing local optimization with a monolithic SDP certificate.

## Compatibility cleanup

- Moves legacy Coriolis compatibility APIs behind `GTSAM_ALLOW_DEPRECATED_SINCE_V43`, keeps their Doxygen deprecation notices, and removes their Python wrapper declarations so generated bindings no longer invoke deprecated APIs.

## Test coverage

[...]

PR #2754, opened 2026-08-27
Add g2o 3D landmark dataset support

dellaert:
## Summary
- add `EDGE_SE3_TRACKXYZ` and `PARAMS_SE3OFFSET` support to the 3D g2o reader and writer
- expose Cartesian landmark observations as `BearingRangeFactor<Pose3, Point3>` with validated information conversion
- add deterministic landmark-localization dataset generation, committed datasets, and regression fixtures

## Behavior
- require each track edge to reference an existing identity offset while accepting unreferenced nonidentity offsets
- reject nonfinite or non-positive-range measurements and invalid information matrices
- serialize compatible track records using an identity sensor offset

## Validation
- `make -j6 testDataset.run`
- rebuilt `gtsam_py`
- `conda run -n py312 python -m pytest ../../python/gtsam/tests/test_dataset.py ../../python/gtsam/tests/test_LandmarkLocalizationDataset.py` (7 passed)
- generated dataset files reproduced byte-for-byte
- `git diff --check`

## Context
This dataset support is split from #2616. That PR branch has intentionally not been modified.


varunagrawal (approved):
Went through this in detail and approving in posterity.

PR #2757, opened 2026-08-28
Wrap PIM backends through template interface

### LINKED ISSUE #2756: MATLAB wrapper fails to build after Galilean IMU wrapper additions
## Description

The MATLAB wrapper no longer builds on current `develop` when the MATLAB toolbox is enabled. Building or installing reaches the generated `gtsam_wrapper.cpp` and fails because it refers to three types that do not exist in the public `gtsam` C++ namespace:

- `gtsam::PreintegratedImuMeasurementsManifold`
- `gtsam::PreintegratedImuMeasurementsTangent`
- `gtsam::PreintegratedImuMeasurementsLieGroup`

Representative compiler output:

```text
error: no template named 'PreintegratedImuMeasurementsManifold' in namespace 'gtsam'
error: no template named 'PreintegratedImuMeasurementsTangent' in namespace 'gtsam'
error: no template named 'PreintegratedImuMeasurementsLieGroup' in namespace 'gtsam'
```

The subsequent collector errors in the generated wrapper are consequences of these missing types.

## Reproduction

Configure GTSAM with the MATLAB toolbox enabled, then build or install:

```bash
cmake -S . -B build -DGTSAM_INSTALL_MATLAB_TOOLBOX=ON
cmake --build build -j6
cmake --install build
```

The failure occurs while compiling the generated MATLAB MEX wrapper:

```text
Building CXX object matlab/CMakeFiles/gtsam_matlab_wrapper.dir/__/wrap/gtsam/gtsam_wrapper.cpp.o
```

## Why it fails

The three class declarations were added to the shared wrapper interface in `gtsam/navigation/navigation.i`, which is consumed by both the Python and MATLAB wrapper generators.

Their corresponding C++ shim definitions were added only to `python/gtsam/preamble/navigation.h`. The Python wrapper includes that preamble, but the MATLAB wrapper does not. Consequently, MATLAB wrapper generation emits code for the declared class names without any C++ definitions for them.

## Introducing pull request

This was introduced by #2231, **Galilean IMU Factor**, specifically commit `e8873828a291952d7c3c46890bd7b800b912db90` (“Add Galilean IMU NEES comparison”), which added both the shared interface declarations and the Python-only preamble definitions.

## Environment observed

- macOS on Apple Silicon
- MATLAB MEX target: `mexmaca64`
- GTSAM `develop`

dellaert:
## Summary

- wrap all three `PreintegratedImuMeasurementsT<Backend>` specializations through the normal template/typedef interface
- expose the complete real PIM API on every named backend, including covariance, deltas, bias access, deskewing, equality, prediction, and `preintegrated()`
- keep `PreintegratedImuMeasurements` as the canonical serialized wrapper and alias the configured backend-specific Python name to it, avoiding duplicate pybind registration
- use backend-neutral generated lambdas for the backend-dependent `preintegrated()` return type and the historical Python `deltaT=` keyword
- remove the Python-only artificial subclasses and expose all four MATLAB classdefs using real C++ template specializations
- remove the remaining hard-coded Tangent types from the legacy PIM and combined-PIM equality signatures

The canonical declaration remains separate because `serialize()` is wrapper metadata rather than a C++ member. Applying that marker to all four typedefs makes MATLAB emit duplicate Boost export registrations for the configured specialization. All real PIM methods are shared by the backend template interface; historical canonical serialization remains unchanged.

## Testing

- `cmake --build build --target gtsam_py -j6`
- focused Python PIM, serialization, and Galilean suite: 9 passed in `py312`
- `cmake --build build --target python-stubs -j6`; all four names and the complete APIs are exported
- configured and built Manifold, Tangent, and LieGroup defaults; import, identity, construction, equality, integration, and `preintegrated()` smoke checks pass
- generated MATLAB output contains all four full classdefs and references the real template specializations
- `gtsam_matlab_wrapper` links after correcting an unrelated existing `RelativeTranslationFactor` generated-code qualification only in the temporary build output
- `git diff --check`

Closes #2756


ProfFan:
> keep PreintegratedImuMeasurements as the canonical Python class

Probably should do this in Python directly (`*.py` files) not in Pybind?

PR #2761, opened 2026-08-29
Add Python bindings for general CUDA Levenberg-Marquardt

leolrg:
## Summary

- Expose shared CUDA linear-solver configuration and the general sparse Levenberg-Marquardt optimizer under `gtsam.cuda`.
- Reuse the shared `LinearSolverType` with the existing CUDA SFM bindings and keep CUDA-only tests out of CPU-only packages.
- Add GPU-backed Python coverage for solver options, status/results, zero-iteration behavior, PCG optimization, and SFM enum compatibility.
- Document CUDA Python builds, General LM and SFM usage, fallback diagnostics, and a runnable Pose2 notebook example.

## Python API

The binding includes the practical configuration and outcome surface: `LinearSolverOptions`, `PcgOptions`, `SparseLevenbergMarquardtParams`, optimizer accessors, backend/fallback/termination enums, and core result counters and errors. Detailed profiling structures remain C++-only.

`gtsam.cuda` remains conditional on `GTSAM_ENABLE_CUDA`; non-CUDA builds and packages are unchanged.

## Test plan

- [x] Full CUDA-enabled Python target: 536 passed, 6 skipped, 13 subtests passed
- [x] Wrapper-generator suite: 82 passed
- [x] General CUDA LM and CUDA SFM tests run successfully on the GPU
- [x] CPU-only module imports without `gtsam.cuda` and excludes both CUDA test modules
- [x] General CUDA LM notebook example reduces the Pose2 objective and reports the device backend
- [x] Notebook structure, local documentation links, and `git diff --check` validated

Testing used Python 3.13.11 with CUDA 13.0 and compute capability 8.0.


dellaert (approved):
Awesome! One question: can we currently do GNC from python? If lots of code change, maybe a question for a separate PR.

leolrg:
> Awesome! One question: can we currently do GNC from python? If lots of code change, maybe a question for a separate PR.

@dellaert Not yet, I will do a PR to add python bindings

PR #2762, opened 2026-08-29
Remove duplication: Consolidate gravity magnitude resolution and reject PIMs without params

nkhedekar:
The gravity-aware IMU factors repeated the same constructor logic for the stored gravity magnitude: take the given value or default to the params' gravity norm, require it to be positive for the Unit3 parametrization, and reject it for Point3. It now lives once in
internal::resolveGravityMagnitude next to GravityParametrization, which also throws std::invalid_argument instead of dereferencing a null params() when a default magnitude is needed from preintegrated measurements that have no params.


dellaert (approved):
LGTM!

PR #2759, opened 2026-08-28
Bring Galilean PIM in line with templated IMU preintegration

dellaert:
## Summary

- refactor the Galilean backend into `GalileanPreintegration` and define `PreintegratedImuMeasurementsG` as `PreintegratedImuMeasurementsT<GalileanPreintegration>`
- replace the separately implemented Galilean PIM with the common templated PIM API, including Boost, Python pickle, and MATLAB string serialization
- expose the Galilean PIM through the normal template/typedef wrapper machinery, using the alias-serialization support from borglab/wrap#208
- optimize Galilean integration, bias correction, residual, and Jacobian hot paths

## Performance

The Galilean hot-path microbenchmarks improved approximately as follows:

- integration: `0.99 us` to `0.56 us` per sample
- factor residual: `158 ns` to `115 ns`
- factor residual with Jacobians: `1235 ns` to `954 ns`

Analytical outputs and Jacobians are unchanged.

## Wrapper behavior

Python and MATLAB expose `PreintegratedImuMeasurementsG` through the same template/typedef machinery as the other preintegration backends. The generated wrapper includes the standard PIM methods and serialization without registering a duplicate native type.

## Compatibility

- existing `PreintegratedImuMeasurements`, `ImuFactor`, and `ImuFactor2` construction, serialization, naming, and APIs remain supported
- the NEES notebook was rerun and produced unchanged results, so its output is not included

## Testing

- `testGalileanImuFactor.run`
- `testSerializationNavigation.run`
- targeted Python IMU, Galilean, gravity, and serialization tests in the `py312` environment
- Python extension and stub targets built
- `gtsam_matlab_wrapper` generated and compiled
- wrap test suite: 134 passed
- `git diff --check`

Closes #2756

PR #2764, opened 2026-08-29
Optimize shared IMU prediction and factor linearization

dellaert:
## Summary

- express shared `NavState` prediction as the exact inertial or rotating-frame transition `X_j = W * phi_dt(X_i) * U`
- specialize the common non-Coriolis path using its sparse block structure, while preserving exact Coriolis and centrifugal dynamics in the rotating path
- optimize the shared PIM bias-Jacobian chain using its inertial 3x3 block structure
- give the five-key `ImuFactor` its fixed `Vector9` output type and fixed-size `FixedJacobianFactor<9, 6, 3, 6, 3, 6>` linearization path, matching `ImuFactor2`
- deprecate the legacy `NavState::coriolis` and `NavState::correctPIM` helpers in favor of concrete PIM `predict()`

## Performance

Measured on top of #2759, the Galilean factor benchmarks improve further:

- factor residual: `115 ns` to `95 ns`
- factor residual with Jacobians: `954 ns` to `650 ns`

A focused benchmark of the shared non-Coriolis prediction path improves from `60.4 ns` to `39.0 ns` for values, from `365.9 ns` to `107.8 ns` with the bias Jacobian, and from `576.4 ns` to `118.6 ns` with both state and bias Jacobians. Rotating-Earth prediction also benefits from the shared refactor. Fixed-size five-key factor linearization is another 2-6% faster than the generic dynamic factor path.

Analytical outputs and Jacobians are unchanged.

## Compatibility

- existing PIM, `ImuFactor`, and `ImuFactor2` construction, serialization, naming, and APIs remain supported
- legacy `NavState::coriolis` and `NavState::correctPIM` remain available behind `GTSAM_ALLOW_DEPRECATED_SINCE_V43`; concrete PIM `predict()` is the supported replacement
- `use2ndOrderCoriolis` remains serialized for compatibility but exact rotating-frame prediction ignores it

## Testing

- `testNavState.run`
- `testImuFactor.run`
- `testImuFactorCovariance.run`
- `testCombinedImuFactor.run`
- `testCombinedImuFactorWithGravity.run`
- `testImuFactorWithGravity.run`
- `testGalileanImuFactor.run`
- `testTangentPreintegration.run`
- `testLieGroupPreintegration.run`
- `testManifoldPreintegration.run`
- focused navigation tests with `GTSAM_ALLOW_DEPRECATED_SINCE_V43` disabled
- numerical prediction Jacobians checked for all four PIM backends, including rotating-Earth gravity derivatives
- `git diff --check`

This PR is intentionally stacked on #2759. After #2759 merges, it can be rebased and retargeted to `develop` without changing its scope.

### DIFF
--- a/gtsam/navigation/ImuFactor.cpp
+++ b/gtsam/navigation/ImuFactor.cpp
@@ -123,10 +123,10 @@ bool ImuFactorT<PIM>::equals(const NonlinearFactor& other, double tol) const {
 
 //------------------------------------------------------------------------------
 template <class PIM>
-Vector ImuFactorT<PIM>::evaluateError(const Pose3& pose_i, const Vector3& vel_i,
-    const Pose3& pose_j, const Vector3& vel_j,
-    const imuBias::ConstantBias& bias_i, OptionalMatrixType H1,
-    OptionalMatrixType H2, OptionalMatrixType H3,
+Vector9 ImuFactorT<PIM>::evaluateError(
+    const Pose3& pose_i, const Vector3& vel_i, const Pose3& pose_j,
+    const Vector3& vel_j, const imuBias::ConstantBias& bias_i,
+    OptionalMatrixType H1, OptionalMatrixType H2, OptionalMatrixType H3,
     OptionalMatrixType H4, OptionalMatrixType H5) const {
   return internal::preintegrationErrorAndJacobians(
       pim_, pose_i, vel_i, pose_j, vel_j, bias_i, H1, H2, H3, H4, H5);
--- a/gtsam/navigation/ImuFactor.h
+++ b/gtsam/navigation/ImuFactor.h
@@ -215,13 +215,15 @@ using PreintegratedImuMeasurements = PreintegratedImuMeasurementsT<DefaultPreint
  * @ingroup navigation
  */
 template <class PIM = PreintegratedImuMeasurements>
-class GTSAM_EXPORT ImuFactorT: public NoiseModelFactorN<Pose3, Vector3, Pose3, Vector3,
-    imuBias::ConstantBias> {
+class GTSAM_EXPORT ImuFactorT
+    : public NoiseModelFactorT<Vector9, Pose3, Vector3, Pose3, Vector3,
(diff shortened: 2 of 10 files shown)


fallenmi (changes requested):
Changing `ImuFactorT::evaluateError` from the dynamic `Vector` return to `Vector9` breaks existing `libgtsam.4` consumers while preserving the same exported symbol and library compatibility identity. I compiled a small consumer against exact-base `245014884` headers and linked it to the base dylib; it returns the expected nine-element residual and exits 0. Running that same untouched binary with exact-head `5daeb8dd` `libgtsam.4.dylib` substituted is accepted by `dyld` and then segfaults (exit 139; the assertion-enabled build aborts in Eigen `Redux.h`). `dyld` confirms the head library is loaded.

`nm -C` shows the `ImuFactorT<...>::evaluateError(...)` symbol is identical in both libraries because the C++ ABI does not encode the return type here, while the caller and callee now disagree on the return-object layout. Both exact libraries still advertise `@rpath/libgtsam.4.dylib`, compatibility version 4.0.0, so the loader has no way to reject the incompatible replacement.

Please preserve the existing exported dynamic-return ABI (for example, keep `ImuFactorT` on its current public base and implement the fixed linearization behind an override or a new type), or deliberately change the library ABI/SOVERSION and document the break. Please also add an old-header/new-library consumer test so this cannot pass same-build CI unnoticed.

The numerical refactor itself checks out: `testNavState` and `testImuFactor` pass on exact base and head, and an independent 16-case oracle across Manifold, Tangent, LieGroup, and Galilean backends found base/head state and Jacobian differences no larger than 1.8e-15 while head analytic Jacobians agree with numerical derivatives within 1.5e-9. The blocker is specifically the mixed-version public ABI.

Disclosure: I used Codex to inspect and build the exact revisions and run the numerical and mixed-library oracles; I verified the result.

ProfFan:
We don't promise any ABI compatibility across minor revisions.

dellaert:
Agreed, @ProfFan

PR #2765, opened 2026-08-29
Optimize IMU measurement integration

dellaert:
## Summary

- add a focused Release benchmark for one-sample standard and Combined IMU preintegration across every supported backend
- replace dense 9x9 and 15x15 covariance products with their exact sparse 3x3 block equations, evaluating one triangle and mirroring the off-diagonal blocks
- specialize the default Tangent path for its exact identity and `dt * I` transition blocks and its ordinary no-sensor-pose gyro-noise structure
- propagate Tangent and LieGroup bias Jacobians through the same nonzero transition blocks instead of dense products
- replace the LieGroup backend's redundant `X * Expmap(Logmap(U))` update with the exactly equivalent direct composition `X * U`, canceling the corresponding 9x9 Expmap and Logmap Jacobians
- add dense-equation regressions for every standard and Combined backend, including correlated sensor noise and displaced/rotated IMU poses

## Performance

Powered Mac Release-build medians, using `10,000` measurements per trial, `20` warmups, and `41` repetitions:

| Standard PIM backend | Before (ns/sample) | After (ns/sample) | Improvement |
|---|---:|---:|---:|
| Tangent | 502.9 | 151.7 | 69.8% |
| Manifold | 467.4 | 317.4 | 32.1% |
| LieGroup | 1227.5 | 387.9 | 68.4% |
| Galilean | 555.3 | 406.7 | 26.8% |

| Combined PIM backend | Before (ns/sample) | After (ns/sample) | Improvement |
|---|---:|---:|---:|
| Tangent | 882.2 | 444.3 | 49.6% |
| Manifold | 842.8 | 570.5 | 32.3% |
| LieGroup | 1617.1 | 635.8 | 60.7% |

The optimized LieGroup mean-and-bias update takes 240.9 ns/sample, down from approximately 852 ns/sample immediately before the direct-composition change.

## Compatibility and correctness

- no public API or serialization changes
- covariance and bias-Jacobian equations are unchanged apart from floating-point evaluation order
- `Expmap(Logmap(U)) = U`, and its right-local Jacobian product is the identity, so the direct LieGroup composition preserves both the value and analytical derivatives
- dense-reference tests compare 25 sequential updates for all four standard and all three Combined backends, both with and without a nontrivial sensor pose
- an additional 20,000-case randomized comparison against the previous LieGroup implementation found worst errors of `2.0e-14` in state coordinates, `3.6e-14` in `F`, `1.1e-16` in `G1`, and `1.3e-15` in `G2`

## Testing

- `testTangentPreintegration.run`
- `testManifoldPreintegration.run`
- `testLieGroupPreintegration.run`
- `testGalileanImuFactor.run`
- `testImuFactorCovariance.run`
- `testImuFactor.run`
- `testCombinedImuFactor.run`
- `testCombinedImuFactorWithGravity.run`
- `testImuFactorWithGravity.run`
- `testCentripetal.run`
- `git diff --check`

This PR is intentionally stacked on #2764. After #2764 merges, it can be rebased and retargeted without changing its scope.

### DIFF
--- a/gtsam/navigation/CombinedImuFactor.cpp
+++ b/gtsam/navigation/CombinedImuFactor.cpp
@@ -79,19 +79,130 @@ void PreintegratedCombinedMeasurementsT<PreintegrationType>::resetIntegration() 
   preintMeasCov_.setZero();
 }
 
-//------------------------------------------------------------------------------
-// sugar for derivative blocks
-#define D_R_R(H) (H)->block<3,3>(0,0)
-#define D_R_t(H) (H)->block<3,3>(0,3)
-#define D_R_v(H) (H)->block<3,3>(0,6)
-#define D_t_R(H) (H)->block<3,3>(3,0)
-#define D_t_t(H) (H)->block<3,3>(3,3)
-#define D_t_v(H) (H)->block<3,3>(3,6)
-#define D_v_R(H) (H)->block<3,3>(6,0)
-#define D_v_t(H) (H)->block<3,3>(6,3)
-#define D_v_v(H) (H)->block<3,3>(6,6)
-#define D_a_a(H) (H)->block<3,3>(9,9)
-#define D_g_g(H) (H)->block<3,3>(12,12)
+namespace {
+
+using Matrix15 = Eigen::Matrix<double, 15, 15>;
+
+/** Propagate Combined PIM covariance using its sparse 5-by-5 block form. */
+template <bool kTangentStructure>
+void propagateCombinedCovariance(Matrix15* covariance, const Matrix9& A,
+                                 const Matrix93& B, const Matrix93& C,
+                                 const PreintegrationCombinedParams& params,
+                                 double dt) {
+  // State blocks are ordered R,p,v,b_a,b_g. The Combined transition is
+  //
+  //       [ A00   0    0    0   C0 ]
+  //       [ A10  A11  A12  B1    0 ]
+  //   F = [ A20   0   A22  B2    0 ] .
+  //       [  0    0    0    I    0 ]
+  //       [  0    0    0    0    I ]
+  const auto A00 = A.block<3, 3>(0, 0);
+  const auto A10 = A.block<3, 3>(3, 0);
+  const auto A11 = A.block<3, 3>(3, 3);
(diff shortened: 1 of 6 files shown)


gtsam/navigation/CombinedImuFactor.cpp
```diff
@@ -79,19 +79,130 @@ void PreintegratedCombinedMeasurementsT<PreintegrationType>::resetIntegration()
...
-#define D_v_v(H) (H)->block<3,3>(6,6)
-#define D_a_a(H) (H)->block<3,3>(9,9)
-#define D_g_g(H) (H)->block<3,3>(12,12)
+namespace {
+
+using Matrix15 = Eigen::Matrix<double, 15, 15>;
+
+/** Propagate Combined PIM covariance using its sparse 5-by-5 block form. */
+template <bool kTangentStructure>
+void propagateCombinedCovariance(Matrix15* covariance, const Matrix9& A,
```
ProfFan, line 88: Won't `Matrix15&` work here?

gtsam/navigation/ImuFactor.cpp
```diff
@@ -33,6 +33,133 @@ namespace gtsam {
...
+                                      const Matrix3& integrationCovariance,
+                                      double dt, bool hasSensorPose) {
+  // Every PIM backend has the same transition sparsity in NavState ordering:
+  //
+  //                       [ A00   0    0  ]
+  //                   A = [ A10  A11  A12 ] .
+  //                       [ A20   0   A22 ]
+  //
+  // Form A*P by 3x3 blocks, skipping those structural zeros.
+  Matrix9 AP;
```
ProfFan, line 53: I don't think Eigen matrices are default initialized to zero, in the 2nd codepath below not all of AP's blocks are assigned?

gtsam/navigation/CombinedImuFactor.cpp
dellaert: I prefer * for in/out args.

gtsam/navigation/ImuFactor.cpp
dellaert: The concern is understandable, but AP is fully assigned.

ProfFan (approved):

PR #2768, opened 2026-08-30
fix cmake ceres cudss dependency issue

jmackay2:
When building with ceres with a cuda dependency, the following error is seen:
```
 Found Ceres 2.3.0; enabling testAdaptAutoDiff
 CMake Error at /usr/lib/x86_64-linux-gnu/cmake/cudss/cudss-config.cmake:176 (set_target_properties):
   Attempt to promote imported target "cudss" to global scope (by setting
   IMPORTED_GLOBAL) which is not built in this directory.
 Call Stack (most recent call first):
   /usr/share/cmake-3.28/Modules/CMakeFindDependencyMacro.cmake:76 (find_package)
   /usr/local/lib/cmake/Ceres/CeresConfig.cmake:129 (find_dependency)
   timing/CMakeLists.txt:1 (find_package)

 CMake Error at /usr/lib/x86_64-linux-gnu/cmake/cudss/cudss-config.cmake:178 (set_target_properties):
   Attempt to promote imported target "cudss_static" to global scope (by
   setting IMPORTED_GLOBAL) which is not built in this directory.
 Call Stack (most recent call first):
   /usr/share/cmake-3.28/Modules/CMakeFindDependencyMacro.cmake:76 (find_package)
   /usr/local/lib/cmake/Ceres/CeresConfig.cmake:129 (find_dependency)
   timing/CMakeLists.txt:1 (find_package)

 Found Ceres 2.3.0
```

This fixes the issue by finding ceres in cmake a single time. 

This was tested on Ubuntu 24.04 with ceres built with cuda and cudss.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,6 +153,9 @@ add_subdirectory(CppUnitLite)
 # Build GTSAM library
 add_subdirectory(gtsam)
 
+# Find ceres for tests and timing dependencies
+find_package(Ceres CONFIG QUIET)
+
 # Build Tests
 add_subdirectory(tests)
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -26,7 +26,6 @@ endif()
 gtsamAddTestsGlob(tests "test*.cpp" "${excluded_tests}" "gtsam")
 
 if(GTSAM_BUILD_TESTS)
-  find_package(Ceres CONFIG QUIET)
   if(Ceres_FOUND)
     message(STATUS "Found Ceres ${CERES_VERSION}; enabling testAdaptAutoDiff")
     gtsamAddTestsGlob(ceres_autodiff "testAdaptAutoDiff.cpp" ""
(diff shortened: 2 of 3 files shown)


fallenmi (approved):
Codex-assisted review.

Moving [`find_package(Ceres)` into the common parent scope](https://github.com/borglab/gtsam/blob/ba737b46f4f7450d44317414904470411a6bf4ec/CMakeLists.txt#L150-L166) is correct: `Ceres_FOUND`, `CERES_VERSION`, and `Ceres::ceres` remain visible to both the [test consumer](https://github.com/borglab/gtsam/blob/ba737b46f4f7450d44317414904470411a6bf4ec/tests/CMakeLists.txt#L26-L35) and the [timing consumer](https://github.com/borglab/gtsam/blob/ba737b46f4f7450d44317414904470411a6bf4ec/timing/CMakeLists.txt#L1-L4), while the package config is loaded only once.

I validated this with a deterministic CMake package-scope oracle that models a promotion-sensitive cuDSS dependency by rejecting a second Ceres config load. The exact PR merge-base and current `develop` each loaded Ceres first from `tests` and then from `timing`, failing on the second load. Exact head and the clean current merge loaded it once from the repository root, configured and generated successfully, and retained both `testAdaptAutoDiff.run` and `timeSFMBALautodiff`. A complementary Ceres-disabled matrix also configured successfully on all four refs and omitted both optional targets as expected. `git diff --check` is clean.

dellaert (approved):
Thanks!!!

PR #2769, opened 2026-08-30
Handle a timestamp supplied for a key with no value and no factor

DLuminary:
Follow-up to #2749. The pending-value partition added there treats a marginalizable key as pending only when a value exists for it:

    !activeKeys.exists(key) && (isam_.valueExists(key) || newTheta.exists(key))

A caller can supply a timestamp for a key that has neither a value nor a factor. It is not active and has no value, so it is not pending and stays in marginalizableKeys. The variableIndex filter added in #2749 keeps it out of the copy passed to BayesTreeMarginalizationHelper, but marginalizableKeys itself is unfiltered and goes on to marginalizeLeaves, which does nodes_[j] and has no node for the key.

Before #2749 the same input threw from getCliquesContainingKeys, which runs before isam_.update(). It now reaches further and the process segfaults. Reproduced with published wheels: gtsam-develop 4.3a2.dev202608252015 raises "map::at", 4.3a2.dev202608300315 dies.

Partition on activity first, then on whether a value exists. Keys with a value are reaped exactly as before; keys without one have nothing in ISAM2 to remove, so only the timestamp is dropped. Both leave marginalizableKeys.

Also names the key in BayesTreeMarginalizationHelper rather than failing inside the Bayes tree, where the message identifies neither the key nor the caller and varies by build -- "map::at" without TBB, "Index out of requested size range" with it. This was one of the asks in #2741.

Tests cover both graph shapes, a stale key alongside a pending one to pin the partition split, and the new diagnostic.

### DIFF
--- a/gtsam/nonlinear/BayesTreeMarginalizationHelper.h
+++ b/gtsam/nonlinear/BayesTreeMarginalizationHelper.h
@@ -27,6 +27,8 @@
 #include <gtsam/inference/BayesTreeCliqueBase.h>
 #include <gtsam/base/debug.h>
 #include "gtsam/dllexport.h"
+#include <stdexcept>
+#include <string>
 
 namespace gtsam {
 
@@ -136,7 +138,17 @@ public:
       const BayesTree& bayesTree,
       const std::unordered_set<Key>& keysOfInterest) {
     std::unordered_set<const Clique*> cliques;
+    const auto& nodes = bayesTree.nodes();
     for (const Key& key : keysOfInterest) {
+      if (!nodes.exists(key)) {
+        // Otherwise the lookup fails inside bayesTree[key], naming neither the
+        // key nor this helper, with text that varies by build: "map::at"
+        // without TBB, "Index out of requested size range" with it. A variable
+        // that no factor references is never eliminated, so it has no clique.
+        throw std::out_of_range(
+            "BayesTreeMarginalizationHelper: key '" + DefaultKeyFormatter(key) +
+            "' has no clique in the Bayes tree.");
+      }
       cliques.insert(bayesTree[key].get());
     }
     return cliques;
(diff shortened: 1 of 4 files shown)


gtsam/nonlinear/BayesTreeMarginalizationHelper.h
```diff
@@ -136,7 +138,17 @@ class GTSAM_EXPORT BayesTreeMarginalizationHelper {
...
+      if (!nodes.exists(key)) {
+        // Otherwise the lookup fails inside bayesTree[key], naming neither the
+        // key nor this helper, with text that varies by build: "map::at"
+        // without TBB, "Index out of requested size range" with it. A variable
+        // that no factor references is never eliminated, so it has no clique.
+        throw std::out_of_range(
+            "BayesTreeMarginalizationHelper: key '" + DefaultKeyFormatter(key) +
+            "' has no clique in the Bayes tree.");
+      }
       cliques.insert(bayesTree[key].get());
```
dellaert: Avoid looking up every Bayes-tree node twice here. `nodes.exists(key)` performs a lookup, and `bayesTree[key]` immediately performs the same lookup again through `nodes_.at(key)`. This keeps the same asymptotic complexity, but doubles the lookup portion for every marginalizable key; that is avoidable for very large trees or large marginalization batches.

Could we use the iterator from a single lookup instead?

```cpp
const auto node = nodes.find(key);
if (node == nodes.end()) {
  throw std::out_of_range(...);
}
cliques.insert(node->second.get());
```

This preserves the improved diagnostic without adding a normal-path lookup.

DLuminary: Thanks for the feedback, it seems correct. Done in 6c8456b.

dellaert (approved):
Cool!

DLuminary:
While checking my own patch for the same problem — I found one in update(). Pushed as a separate last commit, c1c9dd0 — drop it if you'd rather it went in its own PR; it's only here because it touches the same lines as the partition in b1b1963.

PR #2770, opened 2026-08-30
Tolerate a missing timestamp in FixedLagSmoother::eraseKeyTimestampMap

DLuminary:
eraseKeyTimestampMap does a bare keyTimestampMap_.at(key), but not every caller can guarantee the key has a timestamp.

IncrementalFixedLagSmoother::update() passes ISAM2Result::unusedKeys straight to it. ISAM2 derives those from the variable index, which says nothing about timestamps, so a key given a value and a factor but no timestamp, whose factor is later removed, arrives here with no entry. std::map::at then throws "map::at", naming neither the key nor the map.

Erasing a key that is not present is a no-op rather than an error, as it is for std::map::erase, so skip it. The other caller in IncrementalFixedLagSmoother and the one in BatchFixedLagSmoother pass keys obtained from findKeysBefore, which are in the map by construction, so their behaviour is unchanged. A second test covers that path.

This is a second, independent cause of the "map::at" reported in #2741, which was closed by #2749. That fix addressed a value with no referencing factor reaching Bayes tree marginalization without a clique; this one needs factorsToRemove instead and involves no clique at all, so it is still reachable. Also independent of #2769.

Since the lookup now yields an iterator, the trailing erase goes through it rather than repeating the search by key. The intervening loop touches only timestampKeyMap_, so the iterator remains valid.

### DIFF
--- a/gtsam/nonlinear/FixedLagSmoother.cpp
+++ b/gtsam/nonlinear/FixedLagSmoother.cpp
@@ -76,7 +76,17 @@ void FixedLagSmoother::updateKeyTimestampMap(const KeyTimestampMap& timestamps) 
 void FixedLagSmoother::eraseKeyTimestampMap(const KeyVector& keys) {
   for(Key key: keys) {
     // Erase the key from the Timestamp->Key map
-    double timestamp = keyTimestampMap_.at(key);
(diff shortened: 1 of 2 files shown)


dellaert (approved):
LGTM!

PR #2773, opened 2026-08-31
fix sagnac term in doppler factor

scottiyio:
Old code was not consistent; used $r_s \times r_r$ but defined derivative w respect to $r_r \times r_s$. This was pulled directly from RTKLIB() which uses a different definition of Sagnac term. 
Existing code would end up returning 2x the error instead of compensating for the Sagnac term correctly

New code is defined consistently with geodist() in the rest of GTSAM.

### DIFF
--- a/gtsam/navigation/DopplerFactor.cpp
+++ b/gtsam/navigation/DopplerFactor.cpp
@@ -48,12 +48,12 @@ DopplerFactor::DopplerFactor(const Key velocityKey, const Key clockBiasPrevKey,
   los_ = e;
 
   // Earth-rotation (Sagnac) rate term:
-  //   (OMGE/c) * (v_s.y*r_r.x + r_s.y*v_r.x - v_s.x*r_r.y - r_s.x*v_r.y)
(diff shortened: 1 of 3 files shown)


dellaert:
Thanks @scottiyio !
In the PR description, please elaborate on how the bug was discovered and how you are convinced it is right now.

@inuex35 could you take a look at this as well?

dellaert (approved):
Running locally, I had derivative error was \(3.3\times10^{-8}\) m/s; the old expression differed by \(6.17\times10^{-4}\) m/s, so I'll approve but I'd still like PR comment, if possible, and an acknowledgement from @inuex35

inuex35:
Checked myself — the sign was indeed flipped. I originally took it from RTKLIB's `resdop()`, but RTKLIB itself has it flipped (see https://destevez.net/2020/04/earth-rotation-corrections-for-range-and-range-rate-in-gnss/). This PR matches the derivative of `geodist()`'s Sagnac term, which is correct. LGTM.

dellaert:
Thanks @inuex35 ! @scottiyio I'll merge but if you can please add something in PR description for traceability.

PR #2775, opened 2026-08-31
Avoid full covariance construction in Gaussian::sigmas()

owenbales:
## Summary

- compute marginal sigmas for an exact `Gaussian` from the row norms of the inverse upper-triangular square-root information matrix, avoiding the full covariance product
- preserve virtual `covariance()` dispatch for derived `Gaussian` types
- add regressions for correlated square-root information and a derived covariance-only implementation

This optimization was found using Zynoptes, the AI Performance Engineer for Robotics.

## Performance

Linux GCC 12.2 Release, with two warm-up pairs and 20 measured baseline/candidate pairs in balanced alternating order. The timing columns are marginal medians; improvement and its 95% paired-bootstrap confidence interval are computed from paired samples.

| Workload | Before | After | Paired improvement |
|---|---:|---:|---:|
| 8 dimensions x 4,096 calls | 1.5714 ms | 0.9534 ms | 38.90% [38.24%, 40.53%] |
| 256 dimensions x 64 calls | 276.7816 ms | 113.4085 ms | 59.09% [58.72%, 59.40%] |
| 512 dimensions x 16 calls | 484.8402 ms | 192.9990 ms | 60.29% [59.97%, 60.56%] |

All paired outputs produced identical checksums. These timings cover only `Gaussian::sigmas()` and do not claim a whole-GTSAM or application speedup.

## Correctness and compatibility

- GCC 12.2 and Clang 16 Release full `check`: 367/367 tests passed with examples built
- `testNoiseModel` passed under GCC AddressSanitizer and UndefinedBehaviorSanitizer builds
- randomized differential coverage matched 436 values within a `2e-15` relative bound
- empty, singular, non-finite, and varied-diagonal cases retained baseline behavior
- a covariance-only derived class verifies that inherited `sigmas()` still dispatches through its `covariance()` override

## Testing

- GCC 12.2 Release full `check` and examples
- Clang 16 Release full `check` and examples
- GCC Release and Clang Release `testNoiseModel.run`
- GCC AddressSanitizer and UndefinedBehaviorSanitizer `testNoiseModel.run`
- randomized and edge-case differential harnesses

### DIFF
--- a/gtsam/linear/NoiseModel.cpp
+++ b/gtsam/linear/NoiseModel.cpp
@@ -158,7 +158,10 @@ Matrix Gaussian::covariance() const {
 
 /* ************************************************************************* */
 Vector Gaussian::sigmas() const {
-  return Vector(covariance().diagonal()).cwiseSqrt();
+  const Matrix& R = this->R();
+  Matrix Rinv = Matrix::Identity(R.rows(), R.cols());
+  R.triangularView<Eigen::Upper>().solveInPlace(Rinv);
+  return Rinv.rowwise().squaredNorm().cwiseSqrt();
 }
 
 /* ************************************************************************* */
--- a/gtsam/linear/tests/testNoiseModel.cpp
+++ b/gtsam/linear/tests/testNoiseModel.cpp
@@ -41,6 +41,16 @@ static const Matrix kCovariance = I_3x3 * kVariance;
 static const Vector3 kSigmas(kSigma, kSigma, kSigma);
 
 /* ************************************************************************* */
+// Verifies sigmas for a correlated Gaussian square-root information matrix.
+TEST(NoiseModel, GaussianSigmasFromUpperTriangularInformation) {
+  const Matrix22 sqrtInformation{{2.0, 1.0}, {0.0, 4.0}};
+  const Vector2 expected{std::sqrt(17.0) / 8.0, 0.25};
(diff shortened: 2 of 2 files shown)


gtsam/linear/NoiseModel.cpp
```diff
@@ -158,7 +158,14 @@ Matrix Gaussian::covariance() const {
 
 /* ************************************************************************* */
 Vector Gaussian::sigmas() const {
-  return Vector(covariance().diagonal()).cwiseSqrt();
+  if (typeid(*this) != typeid(Gaussian)) {
```
dellaert: When would this happen?

owenbales: Only for custom Gaussian subclasses that override covariance() but not sigma(s). The built-in subclasses already in gtsam won't use this path. I could remove it if you don't think the compatibility is needed, but I think it would be nice to preserve compatibility for potential external subclasses

ProfFan: I don't think this is correct

owenbales: I could just remove the typeid fallback; the performance optimization remains w/o it. It was an attempt to ensure no possible external subclass has any change in behavior; but ultimately there is no case for it in the repo today, so it shouldn't matter.

dellaert (approved):
LGTM, thanks!

PR #2777, opened 2026-09-01
python: ship a PEP 561 py.typed marker

DLuminary:
The wheels ship ~1.1 MB of generated .pyi stubs, but no py.typed marker. Under PEP 561 a type checker must ignore inline stubs in an installed package unless that marker is present, so the stubs are dead weight for anyone running mypy:

    error: Skipping analyzing "gtsam": module is installed, but missing
    library stubs or py.typed marker  [import-untyped]

Every gtsam symbol then resolves to Any. pyright only reads the stubs because useLibraryCodeForTypes defaults to true, and still reports reportMissingTypeStubs under strict mode.

Add the marker to both python packages and list it in package_data. python/gtsam and python/gtsam_unstable are staged into the build tree with copy_directory, so no CMake change is needed.

### DIFF
--- a/python/setup.py.in
+++ b/python/setup.py.in
@@ -14,6 +14,7 @@ package_data = {
         "./*.dll",
         "./*.pyd",
         "*.pyi", "**/*.pyi",  # Add the type hints
+        "py.typed",  # PEP 561 marker, without which the type hints are ignored
     ]
 }
 


dellaert (approved):
Nice! Does this mean we can omit a lot of # ignore warnings in our code? Should this be in the same PR? (Assuming make Python-install creates the typing stubs)

DLuminary:
A few, but fewer than it looks — only 3 come from this PR.

py.typed removes one specific error: mypy's import-untyped, which fires on the import line itself, and the # type: ignore there is what silences it. So the removable ones are import gtsam, from gtsam.utils import plot, from gtsam.symbol_shorthand import P.

The rest fall into two groups. Some are unrelated to gtsam entirely — import google.colab, dataset[...] on a Dict[str, object], the unbound TypeVar at numerical_derivative.py:40. The others, in the two ASV notebooks, are already stale: they were added the same day NavStateImuEKF/Gal3ImuEKF were wrapped, when the installed stubs didn't have those classes yet. I checked NavStateImuASVExample.ipynb with the ignores stripped — pyright reports nothing on any of those lines, with or without this PR.

Also numerical_derivative.py:13's # mypy: disable-error-code="import-untyped" is dead: that file imports only typing and numpy and per git never imported gtsam.

I'd do the cleanup as a follow-up. Nothing type-checks the Python in CI right now (no mypy/pyright job, neither in dev_requirements.txt), so removals would be unverified. Order: land this → add a mypy job → drop the ignores with the checker proving each. Happy to take that on.

And yes on the parenthetical — python/CMakeLists.txt:401 appends python-stubs to GTSAM_PYTHON_INSTALL_EXTRA, which python-install lists in DEPENDS (plus python-unstable-stubs at :376 when unstable is on). If pybind11-stubgen isn't installed the target fails and make aborts before pip runs, so a local install can't end up with the marker but no stubs.

PR #2778, opened 2026-09-01
CustomFactor: make the jacobian argument optional in its own type

DLuminary:
The Python callback receives None for its jacobians whenever only the error is requested -- python/gtsam/tests/test_custom_factor.py asserts exactly that -- but the generated stub says otherwise:

    errorFunction: Callable[[CustomFactor, Values, JacobianVector], ...]

pybind11 renders pointer arguments without Optional, so the nullability that CustomErrorFunction expresses with 'const JacobianVector *' is lost on the way to Python. Callers who follow the stub omit the None check and get a TypeError at runtime; callers who annotate H as JacobianVector have their correct 'if H is not None' guard reported as unreachable.

Express the nullability in a form pybind11 understands. A std::optional<std::reference_wrapper<JacobianVector>> keeps the reference semantics the raw pointer was chosen for -- JacobianVector is opaque, so pybind11 preserves the std::vector memory layout and Python's writes are still visible in C++ -- and pybind11 renders it as 'JacobianVector | None'.

Verified against pybind11 3.0.4 with an isolated module: passing std::ref(v) delivers the bound vector and in-place writes propagate back to C++, passing std::nullopt delivers None, and the generated signature is Callable[[VecInt | None], ...].

As a side effect the const_cast in each native C++ callback goes away, since the reference is already non-const.

Updates the two C++ tests, the MATLAB wrapper callback, and the CustomFactor notebook.

### DIFF
--- a/gtsam/nonlinear/CustomFactor.cpp
+++ b/gtsam/nonlinear/CustomFactor.cpp
@@ -27,7 +27,7 @@ Vector CustomFactor::unwhitenedError(const Values& x, OptionalMatrixVecType H) c
 
     if(H) {
       /*
-       * In this case, we pass the raw pointer to the `std::vector<Matrix>` object directly to pybind.
+       * In this case, we pass a reference to the `std::vector<Matrix>` object directly to pybind.
        * As the type `std::vector<Matrix>` has been marked as opaque in `preamble/custom.h`, any changes in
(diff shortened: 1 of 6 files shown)


dellaert (approved):
LGTM!

PR #2779, opened 2026-09-01
python: remove the Rot3 validation shim and fix wrapper matrix shape checks

DLuminary:
f8469b12 of #2621 added matrix validation by deriving a Python subclass:

    class Rot3(_Rot3_orig):
        def __init__(self, *args, **kwargs): ...

That makes gtsam.Rot3 a *different* type from gtsam.gtsam.Rot3, while every inherited factory and every C++ method returning a rotation keeps returning the extension base:

    gtsam.Rot3 is gtsam.gtsam.Rot3            -> False
    isinstance(Rot3.Ypr(.1,.2,.3), gtsam.Rot3) -> False
    isinstance(Pose3(...).rotation(), gtsam.Rot3) -> False

So any user code doing isinstance(x, gtsam.Rot3) silently stops matching rotations produced by Ypr/Rz/ClosestTo/inverse, Pose3.rotation(), Values.atRot3(), and optimizer results.

pybind11-stubgen faithfully records the split. The re-export

    from gtsam.gtsam import Rot3

disappears from gtsam/__init__.pyi and is replaced by a second class, so type checkers reject ordinary code:

    error: Type "gtsam.gtsam.Rot3" is not assignable to declared type
    "gtsam.Rot3" (reportAssignmentType)
    error: Argument of type "Rot3" cannot be assigned to parameter "r"
    of type "Rot3" (reportArgumentType)

**Review of #2621 established that the validation itself is unnecessary, so remove it rather than repair it.** #2300 reproduces, but not for the reason given: on gtsam 4.2 (cp38–cp311 wheels only, which is why the reporter mentions switching interpreters) every matrix constructor segfaults, including gtsam.Rot3(np.eye(3)) — a valid rotation. It's a numpy 2.x ABI break; 4.2 was built in 2023, before numpy 2.0. With numpy<2 the exact matrix from the issue constructs cleanly, and gtsam.Pose3(np.eye(4)) crashes identically under numpy 2. 4.3a0 onward, built against numpy 2, don't crash on any of it.

Rot3::IsValid, its wrapper exposure and its C++ test are kept — useful independent of the shim, along with Rot3.ClosestTo. A regression test pins gtsam.Rot3 is gtsam.gtsam.Rot3 so the split cannot return.

**The second commit fixes the crash that is real.** It is shape, not values:

```
gtsam.Rot3(np.eye(2))        -> SIGSEGV   on 4.2.2, 4.3a0/a1/a2 and develop
gtsam.Pose3(np.eye(2))       -> a garbage 4x4 transform
gtsam.Rot3(np.zeros((3,4)))  -> an all-zero rotation
```

geometry.i:514 declared Rot3(gtsam::Matrix R) while the C++ ctor is explicit Rot3(const Matrix3&); pybind11 accepts any shape and the dynamic-to-fixed Eigen conversion is UB under NDEBUG. SO3 at :315 is already declared Matrix3 and correctly raises TypeError. Eight declarations fixed across geometry.i, cameras.i, slam.i; genuinely variable-length parameters left alone.

**Behaviour changes.** Wrong-shaped arrays now raise TypeError instead of returning garbage. A correctly-shaped but invalid 3×3 constructs silently again, as before #2621 and as in C++ — the shim only ever covered one entry point, since gtsam.gtsam.Rot3(M) and gtsam.SO3(M) accept it even with the shim in place.

### DIFF
--- a/gtsam/geometry/Pose2.h
+++ b/gtsam/geometry/Pose2.h
@@ -100,11 +100,6 @@ public:
   /// @name Advanced Constructors
   /// @{
 
-  /** Construct from canonical coordinates \f$ [T_x,T_y,\theta] \f$ (Lie algebra) */
-  Pose2(const Vector& v) : Pose2() {
-    *this = Expmap(v);
-  }
-
   /**
    *  Create Pose2 by aligning two point pairs
    *  A pose aTb is estimated between pairs (a_point, b_point) such that 
@@ -340,6 +335,12 @@ public:
   /// @deprecated: use Hat
   static inline Matrix3 wedge(double vx, double vy, double w) {
     return Hat(TangentVector(vx, vy, w));
+  }
+
+  /// @deprecated: use Expmap. This is the exponential map, not componentwise
+  /// construction: use Pose2(x, y, theta) for that.
+  Pose2(const Vector& v) : Pose2() {
+    *this = Expmap(v);
   }
 #endif
   /// @}
--- a/gtsam/geometry/cameras.i
+++ b/gtsam/geometry/cameras.i
@@ -326,7 +326,7 @@ class CalibratedCamera {
   // Standard Constructors and Named Constructors
   CalibratedCamera();
   CalibratedCamera(const gtsam::Pose3& pose);
-  CalibratedCamera(const gtsam::Vector& v);
+  CalibratedCamera(const gtsam::Vector6& v);
   static gtsam::CalibratedCamera Level(const gtsam::Pose2& pose2,
                                        double height);
 
@@ -514,7 +514,7 @@ class SphericalCamera {
   SphericalCamera(const gtsam::Pose3& pose);
   SphericalCamera(const gtsam::Pose3& pose,
                   const gtsam::EmptyCal::shared_ptr& cal);
-  SphericalCamera(const gtsam::Vector& v);
+  SphericalCamera(const gtsam::Vector6& v);
 
   // Testable
   bool equals(const gtsam::SphericalCamera& camera, double tol = 1e-9) const;
--- a/gtsam/geometry/geometry.i
+++ b/gtsam/geometry/geometry.i
@@ -511,7 +511,7 @@ class Quaternion {
 class Rot3 {
   // Standard Constructors and Named Constructors
   Rot3();
-  Rot3(gtsam::Matrix R);
+  Rot3(gtsam::Matrix3 R);
   Rot3(const gtsam::Point3& col1, const gtsam::Point3& col2,
        const gtsam::Point3& col3);
   Rot3(double R11, double R12, double R13, double R21, double R22, double R23,
@@ -651,7 +651,7 @@ class Pose2 {
   Pose2(double x, double y, double theta);
   Pose2(double theta, const gtsam::Point2& t);
   Pose2(const gtsam::Rot2& r, const gtsam::Point2& t);
-  Pose2(gtsam::Vector v);
+  Pose2(gtsam::Vector3 v);
 
   static std::optional<gtsam::Pose2> Align(const gtsam::Point2Pairs& abPointPairs);
   static std::optional<gtsam::Pose2> Align(gtsam::ConstMatrixView a, gtsam::ConstMatrixView b);
@@ -770,7 +770,7 @@ class Pose3 {
   Pose3(const gtsam::Pose3& other);
   Pose3(const gtsam::Rot3& r, const gtsam::Point3& t);
   Pose3(const gtsam::Pose2& pose2);
-  Pose3(gtsam::Matrix mat);
+  Pose3(gtsam::Matrix4 mat);
 
   static std::optional<gtsam::Pose3> Align(const gtsam::Point3Pairs& abPointPairs);
   static std::optional<gtsam::Pose3> Align(gtsam::ConstMatrixView a, gtsam::ConstMatrixView b);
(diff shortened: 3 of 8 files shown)


dellaert (approved):
Cool! Will merge if CI’s ok

ProfFan:
I think the original issue, and this fix, are both not valid - gtsam.Rot3 constructor in 4.3 never crashes. I don't think the original #2466 author reproduced the crash described in #2300 at all. I recommend to kill the entire Python-side workaround by reverting #2621

DLuminary:
Reworked after @ProfFan's review — thanks, you were right that the premise was wrong.

I could reproduce #2300, but only on gtsam 4.2 with Python 3.11 (4.2 has cp38–cp311 wheels only, which is why the reporter mentions switching interpreters). On that build everything segfaults, including gtsam.Rot3(np.eye(3)) — a perfectly valid rotation. It's a numpy 2.x ABI break: 4.2 was built in 2023, before numpy 2.0. Pin numpy<2 and the exact matrix from the issue constructs fine; gtsam.Pose3(np.eye(4)) crashes the same way under numpy 2. So it was never about Rot3 or about invalid input. 4.3a0 onward, built against numpy 2, don't crash on any of it.

So the Python validation guards against something that doesn't happen — and it costs a type split, since class Rot3(_Rot3_orig) makes gtsam.Rot3 a different type from gtsam.gtsam.Rot3 (which every factory and every C++ return still produces). isinstance(Rot3.Ypr(...), gtsam.Rot3) is False, and type checkers reject ordinary code with "Rot3 is not assignable to Rot3".

I have added 2 commits:

5b059c7 removes the shim, keeping Rot3::IsValid, its wrapper exposure and its C++ test — those are useful independent of the shim. The shim tests are replaced by IsValid tests plus a regression test pinning the type identity.

652e838 fixes the crash that is real. It's shape, not values:

```
gtsam.Rot3(np.eye(2))        -> SIGSEGV   on 4.2.2, 4.3a0/a1/a2 and develop
gtsam.Pose3(np.eye(2))       -> a garbage 4x4 transform
gtsam.Rot3(np.zeros((3,4)))  -> an all-zero rotation
```
Cause: geometry.i:514 declared Rot3(gtsam::Matrix R) while the C++ ctor is explicit Rot3(const Matrix3&). pybind11 accepts any shape and the dynamic→fixed Eigen conversion is UB under NDEBUG. SO3 at :315 is already declared Matrix3 and correctly raises TypeError. Pose2::Pose2(const Matrix&) even has a shape check, but behind #ifndef NDEBUG, so it's compiled out of release wheels. Eight declarations fixed across geometry.i, cameras.i and slam.i; genuinely variable-length parameters (noise sigmas, Values::insert, JacobianFactor blocks, SOn, retract deltas) left alone.

This is a behavior change — code passing a wrong-shaped array previously got garbage and now gets TypeError. And removing the shim means an invalid-but-correctly-shaped 3×3 constructs silently again, same as before #2621 and same as C++; the shim only ever covered one entry point anyway (gtsam.gtsam.Rot3(M) and gtsam.SO3(M) accept it even with the shim in place). Rot3.IsValid and Rot3.ClosestTo remain for callers who want a check.

dellaert:
@ProfFan I'll let you approve and merge or ask for more changes

DLuminary:
macOS CI caught a real problem with one of the eight declarations:

```
error: call to constructor of 'gtsam::Pose2' is ambiguous
  note: candidate constructor  Pose2(const Matrix &T)   Pose2.h:90
  note: candidate constructor  Pose2(const Vector& v)   Pose2.h:104
```

`Pose2` is the only one of the eight with two competing single-argument Eigen constructors, and an `Eigen::Vector3d` converts to both `MatrixXd` and `VectorXd`, so neither overload wins. The dynamic `gtsam::Vector` spelling compiles because it matches `Pose2(const Vector&)` exactly, which beats the conversion.

Pushed a commit reverting `Pose2` to `gtsam::Vector` and dropping its test. The other seven keep their fixed sizes.

So `Pose2(np.array([1., 2.]))` stays unchecked for now. Fixing it needs a disambiguating adapter, and `wrap_ctors` always emits `py::init<...>` with no `@pybind_lambda` support, so it's out of scope here.

Worth flagging separately: `Pose2(gtsam::Vector v)` is a 2011 shim from 847e0cb3e ("Wrapped Expmap, logmap, Expmap constructor") — it's `Expmap`, not `(x, y, theta)`. `Pose2([1,2,3])` gives `x=-1.280 y=0.757 theta=3.000`, and it only looks componentwise when `theta=0`. `Pose2.Expmap` is separately wrapped, already declared `Vector3`, and already rejects wrong lengths. `Pose3` has no vector constructor at all. Deprecating the alias would remove the ambiguity at the source rather than working around it — might be worth a separate PR.

dellaert:
I'm happy do deprecate that Pose2 constructor right here :-) With the deprecation mechanism outlined in copilot-instructions.

DLuminary:
Done.

dellaert:
@ProfFan ?

gtsam/geometry/Pose2.h
```diff
@@ -341,6 +336,12 @@ class GTSAM_EXPORT Pose2: public MatrixLieGroup<Pose2, 3, 3> {
   static inline Matrix3 wedge(double vx, double vy, double w) {
     return Hat(TangentVector(vx, vy, w));
   }
+
+  /// @deprecated: use Expmap. This is the exponential map, not componentwise
+  /// construction: use Pose2(x, y, theta) for that.
+  Pose2(const Vector& v) : Pose2() {
+    *this = Expmap(v);
+  }
```
ProfFan, line 344: missing macro and deprecation annotation

ProfFan (changes requested):

gtsam/geometry/Pose2.h
dellaert: macro is a few lines up...

ProfFan (approved):
LGTM, thanks!

PR #2781, opened 2026-09-01
Reject timestamps without corresponding values in IncrementalFixedLagSmoother

talfeiner-mc:
## Summary

This PR validates timestamp keys passed to `IncrementalFixedLagSmoother::update()`.

Every key in `timestamps` must either already have a value in the smoother or have a value supplied through `newTheta` in the same update:

```cpp
isam_.valueExists(key) || newTheta.exists(key)
```

Otherwise, the update throws `std::invalid_argument` with an error identifying the key. Validation occurs before modifying the timestamp map or ISAM2 state, so a rejected update leaves the smoother unchanged.

An earlier revision derived the clock only from active keys in response to [this counterexample](https://github.com/borglab/gtsam/pull/2781#issuecomment-5497700365), as described in [this follow-up](https://github.com/borglab/gtsam/pull/2781#issuecomment-5510246504). That approach has been removed. This revision preserves the existing clock behavior and narrows the change to input validation.

## Problem

The smoother clock is derived from the largest stored timestamp. A timestamp for a key with no corresponding value can therefore move the fixed-lag cutoff forward and silently marginalize established states, even though the timestamped key does not exist in ISAM2.

For example, with a lag of `5.0`:

```text
update 1: X0, value and prior factor, t = 10.0
update 2: X1, value and between factor, t = 11.0
update 3: no values or factors, only P0 timestamped at 16.1
```

The third update moves the cutoff to `11.1`, marginalizes the established states, and leaves the nonexistent `P0` as the newest timestamp.

In our application this input came from a caller-side bookkeeping bug. Rejecting it immediately provides a useful diagnostic instead of silently changing the estimator state.

## Contract

* A timestamp for an existing value is accepted, even if no factor currently references it, and may advance the smoother's time.
* A timestamp and its value may be supplied together in the same update.
* A timestamp for a key with neither an existing nor a same-update value is rejected before any state mutation.
* Timestamp-before-value is not supported.
* A far-future timestamp attached to an existing or same-update value remains invalid caller input and is outside this PR's protection.
* Existing clock and marginalization semantics are unchanged.
* `BatchFixedLagSmoother` is unchanged.

## Tests

Regression tests verify that:

* A timestamp without a corresponding value is rejected and the error identifies the key.
* A rejected update leaves the smoother unchanged.
* Timestamps for existing and same-update values remain accepted, including values not yet referenced by a factor.
* Existing #2769 cleanup tests seed their otherwise unreachable timestamp state through a test fixture, ensuring those tests remain non-vacuous.

Follow-up to #2769. Related: #2741, #2749, and #2770.

### DIFF
--- a/gtsam/nonlinear/IncrementalFixedLagSmoother.cpp
+++ b/gtsam/nonlinear/IncrementalFixedLagSmoother.cpp
@@ -23,6 +23,8 @@
 #include <gtsam/nonlinear/BayesTreeMarginalizationHelper.h>
 #include <gtsam/base/debug.h>
 
+#include <stdexcept>
+
 namespace gtsam {
 
 /* ************************************************************************* */
@@ -58,6 +60,23 @@ FixedLagSmoother::Result IncrementalFixedLagSmoother::update(
   std::optional<FastMap<Key, int> > constrainedKeys = {};
 
   const KeySet newFactorKeys = newFactors.keys();
+
+  // Every supplied timestamp must name a value the smoother already holds or
+  // one arriving in this update. A timestamp for any other key describes
+  // nothing the smoother estimates, yet it would enter the map that
+  // getCurrentTimestamp() -- the maximum over all entries -- derives the clock
+  // from, and could silently expire every established state. Validate before
+  // updateKeyTimestampMap(), the first state mutation, so a rejected update
+  // leaves the smoother unchanged.
+  for (const auto& keyTimestamp : timestamps) {
+    const Key key = keyTimestamp.first;
+    if (!isam_.valueExists(key) && !newTheta.exists(key)) {
+      throw std::invalid_argument(
+          "IncrementalFixedLagSmoother::update: timestamp supplied for key '" +
+          DefaultKeyFormatter(key) +
+          "', but no value exists in the smoother or newTheta.");
+    }
+  }
 
   // Update the Timestamps associated with the factor keys
   updateKeyTimestampMap(timestamps);
--- a/gtsam/nonlinear/IncrementalFixedLagSmoother.h
+++ b/gtsam/nonlinear/IncrementalFixedLagSmoother.h
@@ -63,6 +63,22 @@ public:
    * @param newTheta new values for new variables only
    * @param timestamps an (optional) map from keys to real time stamps
    * @param factorsToRemove an (optional) list of factors to remove.
+   *
+   * Every key in @p timestamps must name a value the smoother already holds or
+   * one supplied in @p newTheta in this update; otherwise the update throws
+   * std::invalid_argument identifying the key, before any state is mutated, so
+   * a rejected update leaves the smoother unchanged. Timestamp-before-value is
+   * not supported. A timestamp for an existing value is accepted whether or
+   * not a factor references the key yet, and participates in the smoother
+   * clock normally.
+   *
+   * If this validation fails, no timestamp is retained and no part of the
+   * update is applied. A caller that later supplies the value must supply its
+   * timestamp again; until timestamped, that value is not eligible for
+   * fixed-lag expiration.
+   *
+   * @throws std::invalid_argument if a key in @p timestamps has no value in
+   * the smoother or in @p newTheta.
    */
   Result update(const NonlinearFactorGraph& newFactors = NonlinearFactorGraph(),
                 const Values& newTheta = Values(), //
--- a/gtsam/nonlinear/tests/testIncrementalFixedLagSmoother.cpp
+++ b/gtsam/nonlinear/tests/testIncrementalFixedLagSmoother.cpp
@@ -36,6 +36,7 @@
 
 // #include <algorithm>
 #include <iostream>
+#include <stdexcept>
 #include <string>
 
 using namespace std;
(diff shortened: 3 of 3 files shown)


dellaert:
@talfeiner-mc Please add a human-written PR comment at the top.
@DLuminary take a look?

DLuminary:
The analysis is right, but I think the fix is one layer too shallow — raising it now rather than after merge.

Reproduced your scenario verbatim on develop with #2769 in: ['x0','x1'] → [], no exception, phantom the sole surviving timestamp. Batch identical. getCurrentTimestamp() is the max while findKeysBefore() only selects entries older than the cutoff, so a newest-timestamp entry is structurally invisible to #2769's partition. My tests never covered it — every phantom in them sits at t = 0.0.

The filter closes one route, not the failure. The predicate is isam_.valueExists(key) || newTheta.exists(key), so attaching a value to the far-future key passes it — and that input is #2749's supported pending value, so it can't be filtered without breaking #2749. Window size after five further legitimate updates, lag 5.0:

```
no phantom                : [2, 3, 4, 5, 6]
phantom @ 1e6, no value   : [0, 0, 0, 0, 0]   <- blocked by this PR
phantom @ 1e6, with value : [1, 1, 1, 1, 1]   <- passes the predicate, still collapses
```

Root cause is that the clock is the max over every entry, so any one of them can drive it arbitrarily far forward.

Suggestion: derive the clock from active keys. A key with no factor isn't being estimated, so it shouldn't define the window — the same isActive predicate #2769 uses (newFactorKeys.exists(k) || variableIndex.find(k) != end()). Modelling the window logic:

```
                       max(all)            max(active)
phantom, no value      [0, 0, 0]           [2, 3, 4]
phantom, with value    [1, 1, 1]           [3, 4, 5]
#2749 pipeline         [2,3,4,5,6,6,6,6]   [2,3,4,5,6,6,6,6]
```

Both routes closed, #2749 unchanged — the clock just runs one update behind while a value waits for its factor. It also avoids two costs this PR carries: nothing is dropped at admission, so timestamp-before-value keeps working, and there's no new path to a value with no timestamp, which never expires since findKeysBefore can't select it (confirmed on develop: such a value is still in the estimate after the clock advances to t=35 with a 5.0 lag).

talfeiner-mc:
First of all, thank you for the quick reply and for maintaining this project.

I am working on a visual-inertial odometry/SLAM pipeline, and I use GTSAM as the backend.
I noticed that a timestamp belonging to a key that is not part of the factor graph can move the smoother's clock far forward. This can cause valid states to be marginalized even though that key is not actually being estimated.

After the review discussion, I changed the approach so that only active keys can define the fixed-lag clock. Inactive/pending timestamps are still kept, so the existing pending-value behavior is preserved.

You were right about the counterexample. I've reproduced it and pushed a revision that takes your suggestion.

**Confirming the counterexample.** On the filter as filed, the predicate is
`isam_.valueExists(key) || newTheta.exists(key)`, so attaching a value to the
far-future key passes it — and that input is #2749's supported pending value,
which is exactly why it can't be filtered.
 Measured here, window size after
five further legitimate updates, lag 5.0:

```
                            filed PR        this revision
no phantom                  [2,3,4,5,6]     [2,3,4,5,6]
phantom @1e6, no value      [2,3,4,5,6]     [2,3,4,5,6]
phantom @1e6, with value    [1,1,1,1,1]     [3,4,5,6,7]
```

The `+1` in the last cell is the pending value itself, correctly retained.
Separately from the window collapse: a continuation where a later factor
references one of the states the phantom clock had already marginalized throws
`ValuesKeyDoesNotExist`, in both smoothers — the same root cause surfacing
wherever the collapsed window is next touched.

Same root cause you named: the clock is the max over every entry, so any single
entry can drive it.

**The revision.** The clock is now derived from active keys, using the same
notion the expiry partition already uses — with one refinement, because
activity is read from the *pre-update* `VariableIndex`: a key whose only factor
is removed in this very update is active by that measure while being inactive
once the update is applied, and would otherwise set a clock no surviving
variable supports. So the clock asks which keys will **remain** active:

```
t_now = max{ t(k) : k has an arriving factor,
                    or an existing factor not in factorsToRemove }
```

The #2769 predicate is left exactly as it is for the partition and the ordering
constraints, which run against the pre-update index by design. (Batch never had
this variant of the problem, since it builds its `VariableIndex` after the
`factorsToRemove` reset.)

The map is walked newest-first and stops at the first qualifying key, so the
ordinary case is one evaluation. Iterating the active keys and looking each
timestamp up instead would cost a pass over every variable in the window on
every update, and this predicate is the more expensive one to evaluate — happy
to switch if you would rather bound the pathological case, where many retained
far-future entries sit ahead of the clock.

Nothing is dropped at admission, so timestamp-before-value keeps working and
this revision adds no new admission-time path to a value with no timestamp --
both costs you flagged. (An inactive timestamp can still expire before its
value arrives; that is pre-existing and not something this changes.) Expiry still
inspects every entry; only the clock is restricted. The clock is
`std::optional`, so when no active key carries a timestamp there is no window
and nothing is expired, rather than a sentinel standing in for one.

**One additional Batch case.** While testing the clock fix, I found a related pre-existing edge case in `BatchFixedLagSmoother`: a pending value with a timestamp but no factor can age out, at which point Batch tries to marginalize a key that has no `VariableIndex` entry and throws.

Incremental already handles this through the partition added in #2769. I added the equivalent handling to Batch: active keys are marginalized normally, values without factors are removed directly, and timestamp-only entries have only their timestamp erased.


**Coverage**, in both smoothers: far-future timestamp with no value; far-future
timestamp **with** a value (your counterexample); a pending value that ages out;
timestamp-before-value retained and reconnecting without being resent; no
active timestamp expiring nothing; and an active far-future key that *does*
move the window, as a positive control against over-blocking. Your existing
#2769 tests are unmodified — with admission untouched, they cover what their
names say again.

Every phantom in the previous version's tests sat inside the window, which is
why none of them caught this; the new ones are all far-future.

DLuminary:
LGTM.

One efficiency point on Batch. VariableIndex(factors_) at the top of update() is a full pass over the graph on every update, and marginalize() builds an identical one, so an update that marginalizes builds it twice.

factorIndex_ can serve after all, if filtered rather than trusted: factorIndex_.find(key), then factors_[slot] non-null. That handles both objections in your comment — reset-in-place slots are listed but null, and empty KeySets fall through. It's also safe against recycling: reset-in-place never pushes to availableSlots_, and removeFactors erases factorIndex_[key] before pushing, so the map holds exactly live plus reset-in-place slots.

dellaert:
I think we are overcomplicating this. I don’t think whether a key currently appears in a factor should decide whether its timestamp advances the smoother’s time. A timestamp of `1e6` followed by `2...6` is simply bad input.

Could we split the Batch fix into a separate PR and keep the existing clock behavior here? Wouldn’t the simplest fix be to reject timestamps that have no corresponding value? Do we need to support timestamp-before-value?

Could you also update the title and description to match the current code?

DLuminary:
> Wouldn’t the simplest fix be to reject timestamps that have no corresponding value? Do we need to support timestamp-before-value?

I think that was @talfeiner-mc original PR version.

dellaert:
Thanks, I see now that this was changed in response to @DLuminary’s earlier comment. Before choosing an approach, could @talfeiner-mc and @DLuminary each briefly state the desired behavior and the real use case motivating it?

Specifically:

1. Should a timestamped value with no factor advance the smoother’s time?
2. What should happen when a timestamp is supplied with no corresponding value?

A short concrete example would help. Please keep it to a few sentences—the detailed discussion is making it difficult for me to identify the intended contract.

DLuminary:
1. Should a timestamped value with no factor advance the clock? Yes. A value inserted before the factor that constrains it is a real state at a real time — that's the ordering #2749 supports, and I'd keep it working. Accepting that does mean a wrong timestamp on such a value still moves the clock, but that's a wrong timestamp rather than a wrong ordering.

2. What should happen to a timestamp with no value? Reject it. Until a value exists nothing can select that entry, so its only possible effect is the clock, and the smoother has no key to attach it to. Throw or silently drop — I don't have a preference.

The case I care about keeping: a caller inserts X10's value one update before the factor connecting it, both stamped t=10. That must keep working. I have no case needing a timestamp for a key with no value.

One consequence worth a line in the docs either way: under rejection, a caller who stamps ahead and doesn't re-stamp when the value arrives ends up with a value carrying no timestamp, which then never expires because findKeysBefore can't select it.

For context, I came to this from #2741/#2769 rather than from a pipeline of my own — @talfeiner-mc has the actual VIO use case.

talfeiner-mc:
Thanks. I’ll keep this short.

**Use case:** a visual inertial odometry backend. Visual updates can arrive after IMU dead reckoning updates. Because of a caller side bookkeeping bug, we occasionally supplied a timestamp before its corresponding value and factor. We do not intentionally rely on that ordering.

**1. Should a timestamped value with no factor advance the smoother’s time?**

Yes. If the value exists, its timestamp may advance the smoother’s time even when no factor currently references it. This preserves the existing pending value behavior. A far future timestamp inconsistent with the rest of the stream remains invalid caller input and is outside this PR’s protection.

**2. What should happen when a timestamp is supplied with no corresponding value?**

`IncrementalFixedLagSmoother` should reject the entire update before mutating any state, with an error identifying the key. We do not need timestamp before value support.

I prefer throwing rather than silently dropping the timestamp because it makes both the invalid input and the required recovery explicit. Because the rejected timestamp is not retained, a caller that later supplies the value must supply its timestamp again. otherwise that value remains untimestamped and cannot be selected for fixed lag expiration until it is timestamped.

Although the caller bug is ours, it can currently cause the established smoother window to be silently marginalized. My preference is therefore to drop the active-backed-watermark revision, preserve the existing clock semantics, and narrow this PR to pre-mutation validation with focused regression tests for `IncrementalFixedLagSmoother`. The Batch changes have been removed and are outside this PR’s scope.

If this contract sounds right, I’ll update the branch, title, and description accordingly.

If you prefer this validation to remain entirely the caller’s responsibility, I’m happy to close the PR.

Thank you for your time.

dellaert:
Excellent! Unless I misunderstood I think you’re in perfect agreement. Throwing seems like the right thing rather than silently doing something incorrectly. So, yes, @talfeiner-mc , please update as you suggest, and @DLuminary - if you’re willing - check the final version, and I’ll approve and merge after CI passes.

talfeiner-mc:
I’ve updated the branch and PR description to the agreed contract:

* `IncrementalFixedLagSmoother` rejects a timestamp whose key has neither an existing value nor a value in the same update, before any state mutation, with `std::invalid_argument` identifying the key.
* Timestamped values remain accepted without a factor and may advance the clock, preserving the pending-value behavior from #2749.
* Existing clock and marginalization semantics are unchanged.
* `BatchFixedLagSmoother` is untouched.
* The regression tests cover atomic rejection and retry, accepted valued keys, pending-value behavior, and keep the #2769 cleanup tests non-vacuous.

`testIncrementalFixedLagSmoother` passes locally.

dellaert (approved):
LGTM. Thanks for narrowing this to the agreed validation contract.

PR #2783, opened 2026-09-03
Support left group actions in EquivariantFilter

### LINKED ISSUE #2753: Support left group actions in EquivariantFilter automatic prediction
### Problem

`EquivariantFilter::computeErrorDynamicsMatrix` currently computes

```math
A = D\phi_0\,D\Lambda
```

and prediction always right-composes the lifted increment,

```math
\hat Y^+ = \hat Y\mathop{Exp}(\Lambda h).
```

The implementation does not inspect `Symmetry::type`, and the existing automatic-prediction tests exercise right actions only.

For a left group action, a left-invariant error, and a right-composed prediction, the error dynamics also contain transport through the lifted velocity. A minimal counterexample is the left-regular action of a noncommutative group on itself with a state-independent lift. The automatic path returns `A = 0`, while the actual left-invariant error satisfies

```math
E^+ = \mathop{Exp}(-uh)E\mathop{Exp}(uh),
```

so the continuous generator is `-ad_u` and the discrete transition is `Ad_{Exp(-uh)}`.

This arose while deriving the left-action Galilean IMU model in #2231. For its augmented state, the expected continuous and discrete matrices are

```math
A_c = \begin{bmatrix}-\mathop{ad}_{q^\circ}&-E\\0&0\end{bmatrix},
\qquad
\Phi = \begin{bmatrix}\mathop{Ad}_{V^{-1}}&-J_R(q^\circ h)hE\\0&I\end{bmatrix}.
```

### Expected behavior

- Derive the automatic error dynamics consistently for both `ActionType::Right` and `ActionType::Left`.
- Add a regression test using a left-regular action on a noncommutative group and compare against the numerical error flow.
- Consider an explicit overload accepting an already-discretized transition `Phi` and process covariance `Qd`; `predictWithJacobian` currently expects continuous-time `A` and discretizes it internally.

### Temporary safeguard

Until automatic left-action dynamics are implemented, reject the automatic path for `ActionType::Left` and direct callers to `predictWithJacobian` with an explicit continuous-time `A`.

DLuminary:
computeErrorDynamicsMatrix() rejected ActionType::Left (a safeguard added in #2231) and directed callers to predictWithJacobian() with an explicit A. Remove the guard and make the left-action paths correct:

- A = Dphi0 * D_lift holds for both action types provided the lift is equivariant, Lambda(phi_X(xi), psi_X(u)) = Ad_X Lambda(xi, u); D_lift then carries any transport term. The A = 0 observed in #2753 comes from a constant lift paired with a trivial input action, which is not an equivariant lift for the body-velocity system xi+ = xi Exp(u h). The equivariant lift Ad_xi u gives D_lift = -ad_u, the generator the issue expects and the same term LieGroupEKF::transitionMatrix() derives from Baker-Campbell-Hausdorff. Document the precondition and the invariant-lift alternative, for which the formula would omit -ad_Lambda.

- Prediction always right-composes, X <- X Exp(Lambda dt). For a left action that places the increment at the reference state, so predict() evaluates the lift there, Lambda(xi_ref, u_origin), instead of at the estimate. The lift value and its Jacobian come from a single evaluation.

- A measurement correction lives in error coordinates at the reference, so for a left action it must compose on the right, X Exp(dx). update() and updateWithVector() composed on the left unconditionally, which rotates the correction by Ad_{X^-1} whenever X is not the identity. Both now go through applyCorrection().

- Add predictWithTransition() for callers that already have a discretized Phi and Qd. Add the protected helpers liftAtOrigin(), propagate(), applyCorrection() and incrementIsAtOrigin().

Tests: the former guard fixture becomes a constant-lift fixture with the adjoint input action that makes it equivariant. New left-regular Rot3 and SO(3)-on-S^2 fixtures check A against the numerical Jacobian of the exact error flow, the mean, predictWithTransition() and update(), all with a non-identity group estimate and, for Rot3, a non-identity reference state.

Update GalileanImuFactor.ipynb, which described the removed guard and the update side. Existing right-action users (ABC, EqVIO) are unaffected.

Fixes #2753


dellaert:
Wow! :-)

dellaert:
@rohan-bansal i’m going to be on planes today, so could you take a look at this?

gtsam/navigation/EquivariantFilter.h
```diff
@@ -103,6 +107,58 @@ class EquivariantFilter : public ManifoldEKF<M> {
...
+    return lift_u_origin(xi_ref_, D_lift);
+  }
+
+  /// Advance the group estimate by a lifted increment and propagate the
+  /// covariance with an already-discretized transition and process noise.
+  void propagate(const TangentG& increment, const MatrixM& Phi,
+                 const CovarianceM& Qd) {
+    g_ = traits<G>::Compose(g_, traits<G>::Expmap(increment));
+    Base::predict(act_on_ref_(g_), Phi, Qd);
+  }
```
rohan-bansal, line 153: Propagate seems to change the group estimate `g_` before Base::predict() validates dynamic sized Phi/Qd, so if an error is thrown then the group estimate is not synced with state/cov. We should validate the matrices before making any changes here.

gtsam/navigation/EquivariantFilter.h
```diff
@@ -272,20 +341,34 @@ class EquivariantFilter : public ManifoldEKF<M> {
...
+   * @param Qd Discrete process noise over dt, in error coordinates.
+   * @param dt Time step, used for the mean only.
+   */
+  template <typename Lift, typename InputOrbit>
+  void predictWithTransition(const Lift& lift_u, const InputOrbit& psi_u,
+                             const MatrixM& Phi, const CovarianceM& Qd,
+                             double dt) {
+    const TangentG Lambda = incrementIsAtOrigin() ? liftAtOrigin<Lift>(psi_u)
+                                                  : lift_u(this->state());
+    propagate(Lambda * dt, Phi, Qd);
```
rohan-bansal: Might be better to use `if constexpr` here? The conditional expression statically compiles both branches, so for a right action, I believe this still requires the Lift to be able to be constructed with [psi_u](https://github.com/borglab/gtsam/pull/2783/changes#diff-0f89cde367a4c44596e9f4a21c57615c6fb33b28258e9c588c2d3eeff1584000R140) from `liftAtOrigin`.

rohan-bansal (changes requested):
Cool! Some minor requested changes. 

I'll rerun ABC/EqVIO experiments with these changes and #2702 to verify nothing changed, and post status in this channel.

gtsam/navigation/EquivariantFilter.h
DLuminary: Agreed — `ManifoldEKF::predict` validates before mutating, so a throw there left `g_` one step ahead of `X_/P_`. Fixed in c4284f1: `g_next` is computed locally, the base class validates and updates, then `g_` is committed. (Same order existed in the old `predictWithJacobian`, so this also fixes the pre-existing path.)

gtsam/navigation/EquivariantFilter.h
DLuminary: Correct — I confirmed it: a right-action Lift that's only constructible from an opaque input type compiles with `predictWithJacobian` but failed with `predictWithTransition`, from `liftAtOrigin`. Switched both `predict()` and `predictWithTransition()` to `if constexpr` in c4284f1 so the discarded branch isn't instantiated, and documented the per-action-type requirement.

DLuminary:
Pushed c4284f1 addressing both comments. Thanks for the review!

On the ABC/EqVIO rerun: both use ActionType::Right, and every left/right branch this PR adds is compile-time selected, so the right-action code path is unchanged — the existing right-action tests pass unmodified. #2702 was already in the develop this branches from; the only delta between its merged state and my base was the #2231 guard this PR removes.

rohan-bansal (approved):
Changes approved, thanks! Will defer to @dellaert for the final merge. 

I sanity checked ABC/EqVIO, and both perform as expected. There is a small bug in ABC Python binding (unrelated), I will open a separate PR for this.

dellaert:
Was out this weekend - let me take another look...

AlessandroFornasier:
I’ve got some time tonight, let me also have a look :)

dellaert:
@AlessandroFornasier @DLuminary I pushed two follow-up commits: f91bfa61d and d4e8b8da3.

- Simplified prediction so all paths evaluate the lift at the current estimate and choose the composition side from the state action: left composition for left actions, right composition for right actions. `predictWithJacobian(lift, A, Qc, dt)` retains its signature; `predictWithTransition(lift, Phi, Qd, dt)` no longer needs an input orbit. Automatic linearization still requires the input orbit and an equivariant lift.
- Preserved the right-action prediction/update conventions used by ABC and the validation-before-commit fix for the group estimate. Added regression coverage for constant spatial lifts, non-identity references/estimates, plain callable lifts, and agreement between explicit and automatic prediction.
- Added `gtsam/navigation/doc/EquivariantFilter.ipynb` as the dedicated user guide, covering lift versus equivariance requirements, action conventions, error coordinates, prediction APIs, and corrections. `EKF-variants.md` retains the short right-action overview and links to the guide.

AlessandroFornasier:
LGTM. I validated the math with some old notes on the left EqF I had since my PhD and all is sound.

DLuminary:
@dellaert Changes LGTM. 
One minor note: the automatic path now evaluates the lift twice (origin for `D_lift`, estimate for the increment); they're related by `Ad_X` if that's ever worth reusing. Thanks for the notebook.

giuliodelama (approved):
Math looks solid to me!

PR #2784, opened 2026-09-04
Add D=1 MOSEK support to certifiable applications

dellaert:
## Summary

- add D=1 QCQP representations for Euclidean vectors, relative translation, and quadratic range factors
- use the compact `[h, cos(theta), sin(theta)]` lift for `Rot2` while preserving the existing D>=2 row-Stiefel path
- add monolithic and chordal MOSEK solver modes to the three certifiable applications, including recovery diagnostics and clean failure reporting
- add a loop-rich 30-pose `w100` slice for nondegenerate planar PGO certification benchmarks

## Results

- all three PR #2713 applications compile and run through both MOSEK SDP backends
- loop-rich PGO/30 certifies with BM, monolithic MOSEK, and chordal MOSEK at the same positive objective
- noisy GOATS/30 certifies monolithically with all 151 variable blocks rank one
- MOSEK remains optional; the modified applications also compile with `GTSAM_WITH_MOSEK=OFF`

## Testing

- `make -j6 testRot2.run`
- `make -j6 testRelativeTranslationFactor.run`
- `make -j6 testQuadraticRangeFactor.run`
- `make -j6 testQcqpProblem.run`
- `make -j6 testFrobeniusFactor.run`
- `make -j6 testLiftedSDPs.run`
- exercised BM, monolithic MOSEK, and chordal MOSEK in all three certifiable applications

A detailed HTML report with formulation notes, tightness measurements, BM results, ordering experiments, and remaining work is attached:

[mosek_d1_pr2713_report.html](https://github.com/user-attachments/files/31821420/mosek_d1_pr2713_report.html)


dellaert:
@avinashresearch1 @david-m-rosen @zhexin1904

dellaert:
Pushed two separately reviewable follow-ups: COLAMD (`1dd18fb26`) and experimental null elimination (`9d0a1ba66`), along with the previously committed opt-in MOSEK diagnostics (`c3c660376`).

### 1. COLAMD for chordal examples

The MOSEK chordal examples now select COLAMD for **clique construction**, and the shared application helper prints that choice. MOSEK's separate interior-point ordering remains automatic: forcing its approximate-minimum-local-fill option increased factor storage and arithmetic in our ordering-only probes. This is not an expanded GTSAM permutation passed into MOSEK.

On the full, unchanged `goats_16.pyfg` graph (201 poses, 4 landmarks, 572 ranges), COLAMD reduces submitted equalities from 87,127 to 55,900 and the largest PSD block from order 69 to 24. Vanilla chordal returned after **124.128 seconds** of MOSEK time; the historical METIS run hit its **900-second process timeout**. This is an improvement in time to a returned iterate, **not a speedup to a certified solution**: COLAMD still terminates with `Unknown / trm_stall`.

### 2. Experimental forced-null elimination, opt-in

`SDPNullspaceReduction.{h,cpp}` implements a small, separate post-processing pass, called at the clearly marked `POST-PROCESSING PASS` in the chordal constructor. It substitutes `X = B Z Bᵀ` to remove duplicate homogeneous coordinates and individually fixed gauge coordinates. It leaves the nonlinear graph, QCQP, clique membership, and original-coordinate constraints intact; recovery maps back to the original coordinates. Coupled anchor equations are not eliminated by this first pass.

The constructor option `eliminateKnownNullDirections` defaults to **false**. The shared application examples enable it with `GTSAM_MOSEK_ELIMINATE_NULLS=1`. Tests cover substitution/objective parity, conservative handling of unsupported constraints, and nonzero-optimum recovery with both METIS and COLAMD.

For full goats16 with COLAMD, the pass lowers the largest cone from order 24 to 17 and scalarized PSD entries from 88,017 to 48,528. Submitted equalities remain 55,900; we have **not** integrated duplicate-row cleanup.

| Full goats16, unscaled | MOSEK time | Status | Primal objective | Refined feasible cost | Rank-one blocks |
|---|---:|---|---:|---:|---:|
| COLAMD vanilla | 124.128 s | Unknown / stall | 6,217.73 | 3,894.12 | 0/978 |
| COLAMD + null elimination | 104.680 s | Unknown / stall | 6,610.60 | 3,894.12 | 2/978 |

These were sequential single runs under the same 15-minute external limits, with unchanged tolerances and no new scaling. Null elimination is useful experimental infrastructure, **not a demonstrated certification fix**. Both primal values exceed the known feasible cost, and independent audits find non-PSD reconstructed dual certificates. Neither run establishes relaxation tightness or supplies a valid certificate.

Fresh validation before pushing: `testSDPNullspaceReduction.run` and `testLiftedSDPs.run` passed, and all six affected MOSEK examples rebuilt successfully. Diagnostics are opt-in via `GTSAM_MOSEK_DIAGNOSTICS`; the committed inspector recomputes residuals and certificate matrices. Scaling experiments, the reverted homogeneous-consistency forest, HTML, and solver dumps are not included in these commits.

examples/CertifiableMosek.h
```diff
@@ -0,0 +1,203 @@
...
+  if (name == "staircase") return CertifiableSolver::Staircase;
+  if (name == "mosek-monolithic") {
+    return CertifiableSolver::MosekMonolithic;
+  }
+  if (name == "mosek-chordal") return CertifiableSolver::MosekChordal;
+  throw std::invalid_argument("Unknown solver: " + name);
+}
+
+/** Fix one rotation and point to remove the global SE(d) gauge from an SDP. */
+template <typename Rotation, typename Point>
```
avinashresearch1, line 65: Is this meant to be in an example? I was using `FrobeniusPrior` for this.

examples/CertifiableMosek.h
```diff
@@ -0,0 +1,203 @@
...
+            << "Rank-one blocks:   " << rankOneBlocks << "\n"
+            << "Minimum EVR:       " << minimumEVR << "\n"
+            << "Median EVR:        " << medianEVR << "\n"
+            << "Maximum EVR:       " << maximumEVR << "\n"
+            << "MOSEK solve time:  " << result.solveTimeSeconds << " s\n"
+            << "SDP wall time:     " << result.wallTimeSeconds << " s\n";
+}
+
+#ifdef GTSAM_USE_MOSEK
+template <typename Solver>
```
avinashresearch1, line 130: Ditto, I think should not be in the `examples` folder?

avinashresearch1: This functionality already exists here: https://github.com/borglab/gtsam/blob/dde7b70d1a3055dec24e0cb21bb6360b0fc52cf4/gtsam/certifiable/LiftedSDPProblem.cpp#L48 

This is repetition.

examples/CertifiableMosek.h
```diff
@@ -0,0 +1,203 @@
...
+MosekExampleResult solveMosek(Solver* solver) {
```
avinashresearch1, line 131: Duplication: Functionality already exists: in gtsam/gtsam/certifiable/LiftedSDPProblem.cpp

gtsam/certifiable/LiftedSDPProblem.cpp
```diff
@@ -40,9 +43,26 @@ namespace gtsam {
...
+  mf::Expression::t asExpr() const { return expression; }
+  mf::Expression::t index(int row, int col) const {
+    return expression->index(monty::new_array_ptr<int, 1>({row, col}));
+  }
+  Matrix matrix() const;
+};
+
+// Maps each pair of QCQP keys to its original-coordinate SDP block.
 using LiftedVariableXijToSDPVariableViewMap =
-    std::map<std::pair<Key, Key>, mf::Variable::t>;
```
avinashresearch1, line 45: Oh no :( 

I wish we could have more discussion before changing the core data structures of the SDP implementation. Its now getting hard to retain a mental model.

dellaert:
### Null-space reduction moved to a follow-up

This PR now ends at `1dd18fb26`. The final null-reduction commit, `9d0a1ba66`, is preserved on the published [`codex/chordal-nullspace-reduction` branch](https://github.com/borglab/gtsam/tree/codex/chordal-nullspace-reduction), rather than being included in this PR's diff. The published history was updated with an explicit force-with-lease against the previous tip.

The D=1 factor extensions, compact Rot2 lift, MOSEK application modes, loop-rich dataset, opt-in diagnostics, and COLAMD example defaults remain here. The null-reduction implementation, constructor/wrapper flag, example environment switch, and associated tests are deferred together as one eight-file change.

The separate null-reduction PR is now open as **#2797**, targeting this PR's branch (`codex/new-mosek-applications`) so its diff contains only the one-commit, eight-file reduction change. **Merge #2784 first; do not merge #2797 into its current feature-branch base.** After #2784 merges, replay/rebase only the saved reduction commit onto the actual merged `develop` (including after a squash merge), retarget #2797 to `develop`, and rerun its focused tests and rotation-ring comparison. Reduction remains disabled by default. [Avinash's concern about changes to the SDP's core representation](https://github.com/borglab/gtsam/pull/2784#discussion_r3962866658) is an explicit design-review topic in #2797. The separate comments about example helpers remain relevant to this PR and have not been resolved by this split.

The existing HTML attachment is retained unchanged. Any null-reduction experiments in the reports or earlier discussion describe deferred work, not functionality in this PR's current diff. The successful noisy 100-rotation-ring experiment found all three formulations numerically tight at the known optimum; null reduction lowered median MOSEK solver time from 60.34 to 44.69 ms (26%), but increased construction-through-recovery time from 87.76 to 94.15 ms (7%). Full goats16 still did not certify. This is evidence for a separate numerical/implementation discussion, not a claim that null reduction fixes chordal convergence.

Validation on the trimmed branch:

- Its tracked tree exactly matches `1dd18fb26`, with no null-reduction implementation or opt-in API remaining.
- Passed `make -j6 testRot2.run testQcqpProblem.run testRelativeTranslationFactor.run testQuadraticRangeFactor.run testFrobeniusFactor.run testLiftedSDPs.run` in the MOSEK-enabled build.
- Rebuilt `CertifiableLandmarkSLAM`, `CertifiablePoseGraphOptimization`, `CertifiableRangeAidedSLAM`, `CertifiableLocalizationExample`, `Pose2RingSLAM_LiftedSDP`, and `Pose3RingSLAM_LiftedSDP`.
- Regenerated CMake first to remove cached references to the extracted source file; `git diff --check` is clean.

examples/CertifiableMosek.h
dellaert: Addressed in cc0d50a12. The example now adds a hard FrobeniusPrior to a copy of the nonlinear graph and lets the existing D=1 lowering construct the rotation constraints; the handwritten Rot2/Rot3 lift selectors are gone. The point anchor remains an explicit hard linear constraint. The original graph is unchanged for BM and physical-cost evaluation. The QCQP, Frobenius, and lifted-SDP test suites passed, all three applications rebuilt, and standalone 2D/3D checks verified hard anchoring, unchanged source/BM graphs, and both MOSEK backends.

examples/CertifiableMosek.h
dellaert: Addressed the duplicated solver policy in cc0d50a12: removed the example-level relative-gap tolerance, so all accuracy settings come from the existing backend defaults. The example overrides only optimizerMaxTime. Its remaining helper delegates to the existing solver API and collects application output; it does not implement a second solve path. A parameter-probe check confirms that only the time limit is supplied. PGO/30 smoke tests retained the prior chordal output (including Unknown status), while monolithic matched the recovered objective.

examples/CertifiableMosek.h
dellaert: The concrete duplication is removed in cc0d50a12: hard rotation anchoring reuses FrobeniusPrior, and MOSEK accuracy defaults are no longer repeated. I have kept example-specific solver selection, runtime limit, timing, result collection, and printing in examples; these delegate to the existing backend. Leaving this thread open for the broader helper-location question rather than treating that design discussion as settled by the cleanup.

gtsam/certifiable/LiftedSDPProblem.cpp
dellaert: Moved to separate PR #2797

PR #2785, opened 2026-09-05
Fix nonlinear symbolic cache invalidation

ProfFan:
A factor can become active after an optimizer updates its estimate, but the cached indexed junction tree still reflects the first linearization. The nonlinear error then includes a factor that the linear solve omits. The regression starts at `x = 0` with a prior at `2` and a hinge penalty above `0.5`: the original Gauss-Newton solve stops at `2`, while the corrected solve reaches the minimum at `1.25`.

Cache the ordering and keys in each Gaussian factor slot alongside the tree, and rebuild when those symbolic inputs change. Exact comparisons also handle deactivation, graph resizing, in-place key changes, and ordering changes. Numerical coefficients, noise models, and variable dimensions can change while reusing the tree. The private cache keeps replacement exception-safe without adding public methods.

Validation used Clang 22.1.8 in Release mode with deprecated APIs and TBB disabled:

- Five added regression tests fail against the original library and pass with the fix. Coverage includes automatic activation in GN and LM with multifrontal Cholesky and QR, factor-slot and key changes, ordering validation and failed-rebuild recovery, numerical changes, and an empty graph.
- `testNonlinearOptimizer.run`, `testGaussianFactorGraph.run`, `testDoglegOptimizer.run`, `testNonlinearConjugateGradientOptimizer.run`, and `testNonlinearMultifrontalSolver.run` all pass.
- The new `timeNonlinearOptimizerCache` benchmark builds and runs. `git diff --check` passes.

The documentation directory cleanup is tracked separately in #2788.

### DIFF
--- a/gtsam/inference/EliminationTree-inst.h
+++ b/gtsam/inference/EliminationTree-inst.h
@@ -98,7 +98,7 @@ namespace gtsam {
       for (size_t j = 0; j < n; j++)
       {
         // Retrieve the factors involving this variable and create the current node
-        const FactorIndices& factors = structure[order[j]];
+        const FactorIndices& factors = structure.at(order[j]);
         const sharedNode node = std::make_shared<Node>();
         node->key = order[j];
 
@@ -135,7 +135,7 @@ namespace gtsam {
         nodes[j] = node;
       }
     } catch(std::invalid_argument& e) {
-      // If this is thrown from structure[order[j]] above, it means that it was requested to
+      // If this is thrown from structure.at(order[j]) above, it means that it was requested to
       // eliminate a variable not present in the graph, so throw a more informative error message.
       (void)e; // Prevent unused variable warning
       throw std::invalid_argument("EliminationTree: given ordering contains variables that are not involved in the factor graph");
--- a/gtsam/nonlinear/NonlinearOptimizer.cpp
+++ b/gtsam/nonlinear/NonlinearOptimizer.cpp
@@ -34,10 +34,44 @@
 #include <stdexcept>
 #include <iostream>
 #include <iomanip>
+#include <vector>
 
 using namespace std;
 
 namespace gtsam {
+
+/// Cache only the symbolic inputs; numerical factors are read at each solve.
+struct NonlinearOptimizer::IndexedJunctionTreeCache {
+  Ordering ordering;
+  std::vector<KeyVector> factorKeys;
+  IndexedJunctionTree tree;
+
+  /// Snapshot the factor slots and ordering used to build the tree.
+  IndexedJunctionTreeCache(const GaussianFactorGraph& graph,
+                           const Ordering& ordering)
+      : ordering(ordering), tree(graph.buildIndexedJunctionTree(ordering)) {
+    factorKeys.reserve(graph.size());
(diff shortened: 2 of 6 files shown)


dellaert:
Good find !!

Review: There is a CI error. Also, could you quantify the effect on timing? It seems that with the cache mechanism we now have to check potentially large things.

dellaert:
I merged #2788 and took it out of this PR. Sorry for the force-push. Still needs the CI fix.

dellaert:
Powered timing results for 1737f67ed5448d11c132842eacc9485205dc8f55: I recommend merging once the remaining Windows CI job passes. Performance should not block this correctness fix.

Median overhead compared with the previous unchecked cache policy:

| Variables | Scalar linear solve | 6D linear solve | Full Gauss–Newton iteration |
|---:|---:|---:|---:|
| 100 | −0.4% | +1.7% | +3.0% |
| 1,000 | +2.1% | +4.2% | +3.0% |
| 10,000 | +5.7% | +0.2% | +3.4% |

The roughly 3% full-iteration overhead is an acceptable cost for preventing silent wrong solutions when factors activate or change. Cached linear solves remain 1.24–1.43× faster than rebuilding the symbolic tree. Cache hits allocate nothing; the stored ordering and factor-key snapshot add approximately 0.80 MB for the 10,000-variable chain, excluding allocator metadata.

Method: Apple M5/arm64 on confirmed AC power, low-power mode off; Clang 22.1.7, Release (-O3), bundled Eigen, TBB and deprecated APIs disabled. Three independent executions per size, each reporting the median of seven samples of 100 operations after warmup. Alternatives are interleaved with their order reversed between samples. Reproduce with `build/timing/timeNonlinearOptimizerCache SIZE 100` three times for each SIZE in 100, 1000, 10000. Percentages above are medians of per-execution checked/unchecked ratios.

The graphs contain N priors and N−1 adjacent-variable factors with COLAMD ordering and multifrontal Cholesky. Full iterations use 3D vector factors and include linearization, retraction, and error evaluation. The unchecked baseline is the benchmark-only implementation of the previous cache policy using the same library, rather than a separate pre-PR build.

Timing noise remains on AC power: the 10,000-variable full-iteration trials ranged from +1.3% to +15.5% overhead, with a +3.4% median. Negative measurements indicate noise, not an intrinsic speedup. These synthetic warm-cache results do not cover initial construction, changing topology, QR, TBB, or a complete SLAM application.

A future optimization could detect structural changes during linearization, while retaining exact validation for direct `solve()` calls. A trustworthy O(1) structural revision counter would require all structural mutations to be tracked. That redesign can be a separate follow-up.

The missing-key CI failure is fixed, six relevant local test targets pass, and full macOS/Linux CI plus all three fast compile checks are green. Windows full CI is still running: https://github.com/borglab/gtsam/actions/runs/34280082427

dellaert (approved):
Approved. The cache validation fixes a demonstrated correctness bug, and the powered benchmarks show an acceptable roughly 3% median full-iteration overhead on the tested chains. The missing-ordering-key CI failure is addressed with checked access and regression coverage. Recommend merging once the remaining Windows CI job passes; further cache optimization can follow separately.

PR #2786, opened 2026-09-06
Reset adaptive LM rejection multiplier after accepted steps

ProfFan:
In adaptive LM (`useFixedLambdaFactor=false`), accepted steps were doubling `currentFactor`, the multiplier used to increase damping after a rejection. Starting at `2`, two accepted steps leave it at `8`, so the next rejection increases lambda eightfold. This accumulation can cause premature termination at `lambdaUpperBound` and eventually overflow the multiplier.

The fix resets `currentFactor` to `2.0` after acceptance and doubles it only after rejection. This follows **Nielsen's update rule**, given in [Madsen, Nielsen & Tingleff (2004), Eq. (2.21), page 14](https://www2.imm.dtu.dk/pubdb/edoc/imm3215.pdf#page=8). [Ceres implements the same reset](https://ceres-solver.googlesource.com/ceres-solver/+/master/internal/ceres/levenberg_marquardt_strategy.cc) in `StepAccepted()`. A custom `lambdaFactor` still controls the initial rejection sequence.

The correction lives in the shared CPU/CUDA policy. The BAL profiling loop now uses that helper too, removing a duplicate implementation. The update remains constant-time and allocation-free.

Validation:

- Five regression cases fail with the original policy and pass with the fix, covering acceptance/rejection sequences, custom initial factors, the damping floor, overflow, and actual LM steps with Cholesky and QR.
- Nonlinear optimizer, multifrontal solver, and SFM optimizer suites pass.
- CUDA sparse PCG and SFM PCG/dense Cholesky validation passed on an RTX 5090. cuDSS was not tested because it is not installed.

### DIFF
--- a/gtsam/nonlinear/internal/LevenbergMarquardtPolicy.h
+++ b/gtsam/nonlinear/internal/LevenbergMarquardtPolicy.h
@@ -35,7 +35,10 @@ inline void increaseLevenbergMarquardtLambda(
   }
 }
 
-/** Decrease LM damping after accepting a step with the given fidelity. */
+/**
+ * Update LM damping after accepting a step with the given fidelity.
+ * The adaptive Nielsen policy resets the rejection multiplier to two.
+ */
 inline void decreaseLevenbergMarquardtLambda(
     const LevenbergMarquardtParams& params, double modelFidelity,
     double* lambda, double* currentFactor) {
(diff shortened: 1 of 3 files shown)


dellaert (approved):
Awesome! Did you see any positive effects in timing?

ProfFan:
Nope, need to test on diverse datasets for this

PR #2787, opened 2026-09-07
Add Python bindings for CUDA GNC optimizers

leolrg:
Python GNC currently exposes only the ordinary CPU LM and Gauss–Newton instantiations. This adds GNC with the general CUDA sparse LM and specialized CUDA SfM LM inner solvers, preserving each backend's parameter type and settings.

The new APIs are `gtsam.cuda.GncSparseLMParams` / `GncSparseLMOptimizer` and `gtsam.cuda.GncSfmLMParams` / `GncSfmLMOptimizer`. They instantiate the existing C++ templates through a CUDA-only wrapper interface and use Python aliases without registering the native types twice. The GNC algorithm is unchanged: weighting and graph reconstruction remain on the CPU, while the selected inner solver uses CUDA.

The documentation covers backend selection, factor-index semantics, supported SfM projection factors, and the requirement to insert specialized SfM landmarks with `Values.insertPoint3`.

Validation:

- CUDA-enabled Python 3.13 build succeeded.
- All 8 new tests passed on an RTX 5090, including general TLS/GM outlier rejection with CPU fallback disabled, specialized SfM weight updates, retained backend settings, and unsupported-factor rejection.
- Full `python-test` target: 542 passed, 21 skipped, 37 subtests passed.
- `make -j6 testGncOptimizer.run` passed.
- CUDA-disabled configuration and wrapper generation exclude the new interface and tests.
- `git diff --check` passed.


dellaert (approved):
Cool! I pushed a commit because I didn't think we need another dot i file and we also should not have a dependence in nonlinear on SFM quantities. The latter adds (even more) duplication in the .i files but can't be helped without changes in wrap.

dellaert:
Thanks !!!!

PR #2791, opened 2026-09-08
Report expired pending keys in FixedLagSmoother::Result

DLuminary:
`IncrementalFixedLagSmoother::update()` removes values that leave the lag window
before any factor references them (via `removeVariables()`, erasing their
timestamps) but does not report which keys it reaped. `Result::keysOfDeletedNodes`
only lists the keys marginalized through `marginalizeLeaves()`. A front-end holding
queued measurements for a pending state can only discover its removal by diffing
the linearization point or timestamps, or by catching the exception from the next
update that references it.

Add `Result::expiredPendingKeys` (KeySet) with `getExpiredPendingKeys()`, populated
from the vector `update()` already builds, exposed in the Python wrapper, and
counted in `Result::print()`. It is deliberately a separate field: the documented
meaning of `keysOfDeletedNodes` is unchanged. `BatchFixedLagSmoother` has no
pending-value path and leaves it empty.

Tests: extends ReapsPendingValueAfterLag to check the result (and its disjointness
from keysOfDeletedNodes) and UpdateResultMarginalizationMetadata to check it is
empty when every expired key had a factor.

### DIFF
--- a/gtsam/nonlinear/FixedLagSmoother.cpp
+++ b/gtsam/nonlinear/FixedLagSmoother.cpp
@@ -27,7 +27,10 @@ void FixedLagSmoother::Result::print() const {
             << "Nr intermediateSteps: " << intermediateSteps << '\n'
             << "Nr nonlinear variables: " << nonlinearVariables << '\n'
             << "Nr linear variables: " << linearVariables << '\n'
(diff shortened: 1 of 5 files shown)


dellaert (approved):
LGTM

PR #2793, opened 2026-09-09
Fix nightly Cayley-map smart PCG test failure

### LINKED ISSUE #2792: [CI] Nightly matrix failure
A scheduled compatibility nightly failed.

- Develop SHA: [8be5fe667ccf6b2c37593eeb8965917fd560db38](https://github.com/borglab/gtsam/commit/8be5fe667ccf6b2c37593eeb8965917fd560db38)
- Run: [34211328467](https://github.com/borglab/gtsam/actions/runs/34211328467)
- Timestamp: 2026-09-08T10:57:02.174Z

### Matrix results

- Linux: **success**
- Linux ARM64: **success**
- macOS: **success**
- Windows: **success**
- vcpkg: **success**
- Special cases: **failure**
- Python extras: **success**

dellaert:
The Cayley-map nightly fails `SmartPcgLinearizationsAgree` because the tiny BAL fixture is underconstrained beyond the similarity gauge. The explicit-Hessian and implicit-Schur optimizations consequently follow different nonlinear trajectories, producing final errors around 0.02146 and 0.03611.

Add soft priors on each camera's pose and calibration within the test fixture, and tighten the final-error comparison from 3e-3 to 1e-6. Also check that both modes start at the same error. Production solvers and benchmark graph builders are unchanged.

Validation:
- Reproduced the original failure locally with `GTSAM_ROT3_EXPMAP=OFF` and `GTSAM_POSE3_EXPMAP=OFF`.
- `make -j6 testSfmBalBenchmark.run` passes with the fix under both Cayley-map and default exponential-map configurations.
- `git diff --check` passes.

Fixes #2792

### DIFF
--- a/timing/tests/testSfmBalBenchmark.cpp
+++ b/timing/tests/testSfmBalBenchmark.cpp
@@ -63,8 +63,16 @@ namespace smart_pcg_tests {
 bal::PcgOptimizationResult optimize(const SfmData& data,
                                     LinearizationMode mode) {
   bal::BalBenchmarkConfig config;
-  const NonlinearFactorGraph graph = bal::buildSmartSfmGraph(
+  NonlinearFactorGraph graph = bal::buildSmartSfmGraph(
       data, config, SmartProjectionParams(mode));
+  // This tiny dataset is underconstrained even beyond the similarity gauge.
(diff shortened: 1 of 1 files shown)


dellaert:
Fixed in better way in #2795

PR #2796, opened 2026-09-10
(do not merge, not ready for review): direct SDP method for comparison, several case runs

avinashresearch1:


dellaert:
Do we still need this now that you made a separate repo?

dellaert:
Please review #2802 so I can merge, and in this PR you can rebase on develop. Tests and timing scripts can be checked into GTSEM. Don't check in any data or results. That is more appropriate in a different repo.

avinashresearch1:
> Do we still need this now that you made a separate repo?


I think having this "direct SDP" method is still useful for GTSAM, even if the default method is to go through a QCQP. 

`explicit LiftedSDPProblem(const NonlinearFactorGraph& graph);`

gtsam/certifiable/internal/DirectSDP.h
```diff
@@ -0,0 +1,50 @@
...
+  std::map<Key, Vector> anchors;
+};
+
+/**
+ * Transcribe homogeneous Pose2 or Pose3 Frobenius graphs without QCQP lowering.
+ * Costs use 0.5 * <Q, X>. Manifold and exact-prior constraints are independent
+ * of the QCQP constraint generator; the coordinate convention is shared.
+ * Unsupported factors, robust noise, and partially hard priors are rejected.
+ */
+GTSAM_EXPORT DirectSDPData buildDirectSDP(const NonlinearFactorGraph& graph);
```
dellaert, line 47: I don't see an implementation of this. Also, I think it's useful for tests, but because it's limited to a particular type of graph, I would move it to the tests folder.

dellaert:
Please re-open when ready

PR #2797, opened 2026-09-11
Add opt-in chordal SDP null-space reduction

dellaert:
## Summary

Stacked on #2784: this PR targets `codex/new-mosek-applications`, so its diff contains only the eight-file null-space-reduction commit, `9d0a1ba66`.

- Add an explicit post-processing pass that removes known forced-null directions from chordal SDP clique matrices: duplicate homogeneous coordinates and individually fixed coordinates.
- Transcribe the objective, original constraints, and full overlap equations through the reconstruction bases, then reconstruct original-coordinate blocks for recovery and rank diagnostics.
- Add an opt-in constructor argument, matching wrapper declaration, and the example switch `GTSAM_MOSEK_ELIMINATE_NULLS=1`. Reduction remains disabled by default.
- Add focused substitution and solver-equivalence tests. Monolithic reduction is not implemented here.

## Review and merge order

Merge #2784 first; **do not merge this PR into its current feature-branch base**. After #2784 merges, rebase/replay only the null-reduction commit onto the actual merged `develop` (including after a squash merge), retarget this PR to `develop`, and rerun the focused tests and rotation-ring comparison.

[Avinash's concern about changes to the SDP's core representation](https://github.com/borglab/gtsam/pull/2784#discussion_r3962866658) is an explicit design-review topic here. In particular, review the boundary between the small basis-construction pass and the original-coordinate expression/recovery adapter. The separate example-helper review comments remain with #2784.

## Experimental evidence

The noisy 100-rotation ring was numerically tight with monolithic, vanilla chordal, and null-reduced chordal formulations, matching the independently known optimum of approximately `0.0099999166669`.

Across six runs, null reduction lowered median MOSEK solver time from **60.34 to 44.69 ms (26%)**, but construction-through-recovery time increased from **87.76 to 94.15 ms (7%)**. This is a solver-only saving, not an end-to-end speedup over vanilla chordal. Full goats16 still failed to certify; smaller cones do not guarantee convergence or tighten the mathematical relaxation.

The existing HTML report remains attached to #2784 and has not been modified as part of this split. Earlier null-reduction results describe the work now isolated here.

## Validation

- Previously passed `make -j6 testSDPNullspaceReduction.run` and `make -j6 testLiftedSDPs.run` on the preserved implementation; six affected certifiable examples were also built.
- The rotation-ring experiment checked objective agreement, reconstructed rank-one blocks, and independent primal/dual diagnostics.
- Verified that the stacked diff contains exactly one commit and eight files. No implementation changes or fresh null-reduction test runs were made while creating this PR.
- The trimmed #2784 baseline separately passed all six agreed test suites and rebuilt all six affected examples after the split.

### DIFF
--- a/examples/CertifiableMosek.h
+++ b/examples/CertifiableMosek.h
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <chrono>
 #include <cmath>
+#include <cstdlib>
 #include <iostream>
 #include <map>
 #include <stdexcept>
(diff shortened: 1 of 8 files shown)


dellaert:
Obsoleted by #2802

PR #2798, opened 2026-09-12
Fix CMake configuration, exports, and installation paths

dellaert:
Several optional build configurations and downstream package consumers fail because CMake exports, dependency discovery, and target properties disagree with the selected options. This cleanup fixes those failures and removes the stale exclusion for the nonexistent `testSerializationConstraint.cpp`.

Changes are split into nine focused commits:

- Skip private SuiteSparse dependency discovery for shared GTSAM consumers, allowing C++-only projects to configure.
- Export the MOSEK backend for static builds and guard MOSEK test properties when tests are disabled.
- Select the legacy Boost timer with the correct macro and the explicit opt-in required by recent Boost versions.
- Discover MKL when loading GTSAM's exported vcpkg MKL interface.
- Remove the MinSizeRel postfix from both Python extension modules.
- Honor `CMAKE_INSTALL_INCLUDEDIR` in installed headers, exported include paths, and the package include-directory variable.
- Link all six bundled METIS executables to `metis-gtsam`.
- Remove the nonexistent constrained serialization test exclusion.

Validation:

- `make -j6 testOrdering.run` passed in `build/`.
- Configuration/generation checks passed for no-Boost builds, static MOSEK and its downstream consumer, MOSEK with tests disabled, and C++-only shared-library consumers.
- MKL dependency discovery passed a downstream configuration check using a minimal MKL package fixture; no real MKL runtime build was performed.
- The legacy timer implementation compiled without unresolved Boost timer symbols.
- Generated MinSizeRel link commands use the correct Python module names; checked with the py312 environment.
- Generated installation rules and exports agree for a custom include directory.
- All six bundled METIS executables built, and `gpmetis` successfully partitioned a sample graph.
- `git diff --check` passed.


dellaert:
@jlblancoc @ProfFan asked Astra to do a review of any remaining cmake issues. Did not find any P1 but found 8 P2 issues.

cmake/HandleMetis.cmake
```diff
@@ -36,8 +36,15 @@ else()
   option(GTSAM_BUILD_METIS_EXECUTABLES "Build metis library executables" OFF)
   add_subdirectory(${GTSAM_SOURCE_DIR}/gtsam/3rdparty/metis)
 
+  # Upstream programs still link "metis", but GTSAM renames the bundled target.
```
ProfFan: This does not seem right

gtsam/constrained/tests/CMakeLists.txt
```diff
@@ -1,9 +1 @@
-# if GTSAM_ENABLE_BOOST_SERIALIZATION is OFF then exclude some tests
-if (NOT GTSAM_ENABLE_BOOST_SERIALIZATION)
-  # create a semicolon seperated list of files to exclude
-  set(EXCLUDE_TESTS "testSerializationConstraint.cpp")
-else()
-  set(EXCLUDE_TESTS "")
-endif()
-
-gtsamAddTestsGlob(constrained "test*.cpp" "${EXCLUDE_TESTS}" "gtsam")
+gtsamAddTestsGlob(constrained "test*.cpp" "" "gtsam")
```
ProfFan, line 1: Did we already remove this test?

ProfFan (changes requested):

cmake/HandleMetis.cmake
dellaert: Tell me more

ProfFan: I don't know what the agent is trying to do. The stuff listed below are not what we've used - they seem to be extensions to metis, and we renamed metis so we should not be supplying these stuff anyways

dellaert: Okay, I reverted this.

dellaert:
@ProfFan Take another look and approve if ok?

ProfFan (approved):

PR #2799, opened 2026-09-12
Fix robust objective and linearization inconsistencies

dellaert:
This PR fixes disagreements between robust objectives and the linearizations used to optimize them, including Scalar robust models on CPU and CUDA, batched factors, spherical-camera residuals, and DCS.

Nonlinear `error()` evaluates the robust objective. A linearized Gaussian factor evaluates a quadratic approximation with weights frozen at the linearization point. LM and Dogleg use the former for actual decrease and the latter for predicted decrease. Their gradients must agree at the expansion point; their absolute values need not agree.

### Changes

- **Scalar robust objectives:** `loss(const Vector&)` accepts an unwhitened residual and sums component losses after whitening in Scalar mode. Previously, objective evaluation always applied the loss to the Mahalanobis norm, even when linearization used independent component weights. With unit noise, Huber threshold 1, and residual `(2, 2)`, the componentwise objective is 3; the old norm-based value was approximately 2.328. Nonlinear factors, direct extended-prior evaluation, wrappers, and CUDA SFM use the corresponding objective.
- **Signed batch linearization:** both compact and dense/Hessian BatchFactor paths negate the residual before robust reweighting, matching ordinary factors. For asymmetric Scalar losses, the vector-loss implementation evaluates `rho(-r)` to agree with this existing RHS convention. The sign conversion stays centralized in `Robust::loss()`; factor callers pass ordinary residuals.
- **Spherical-camera residuals:** TriangulationFactor uses the same residual chart for evaluation and linearization. Both GeneralSFMFactor Jacobian overloads and TriangulationFactor apply the derivative of the measurement's Local chart when available. Unit3 gains analytic Local Jacobians, including a stable coincident-direction limit. Reversing Unit3 Local arguments changes tangent bases and cannot replace componentwise negation. Requesting a derivative at the antipode now explicitly throws, while value-only evaluation retains its existing convention.
- **DCS:** preserve the existing clipped IRLS weights and integrate them to obtain a matching loss: `0.5*r²` for `r² <= c`, otherwise `1.5*c - 2*c²/(c+r²)`. The loss and derivative are continuous at the threshold. Graduated DCS now has the standard `0.5*r²` least-squares endpoint. This intentionally changes DCS objective values in both Scalar and Block modes.
- Clarify the robust-noise documentation's distinction between the nonlinear objective and the frozen weighted quadratic approximation.

The default reweighting scheme remains Block. The Scalar aggregation change leaves ordinary Block losses unchanged; the separate DCS correction and spherical-camera Jacobian corrections also apply to Block configurations. The squared-distance loss overload remains available.

### Validation

Targeted C++ tests on Cyber with deprecated APIs disabled, CUDA/cuDSS enabled, and one compilation job: BatchFactor, Unit3, TriangulationFactor, GeneralSFMFactor, NoiseModel, Robust, SphericalCamera, Triangulation, and CudaSfm. Regressions compare numerical objective derivatives with linearized gradients across both residual signs, correlated whitening, compact/dense/Hessian batches, and Scalar/Block kernels. DCS tests additionally cover graduation and threshold continuity. Python wrapper generation was checked for the new Unit3 overload.

### DIFF
--- a/gtsam/linear/NoiseModel.cpp
+++ b/gtsam/linear/NoiseModel.cpp
@@ -801,6 +801,20 @@ bool Robust::equals(const Base& expected, double tol) const {
   return noise_->equals(*p->noise_,tol) && robust_->equals(*p->robust_,tol);
 }
 
+double Robust::loss(const Vector& v) const {
+  if (robust_->reweightScheme() == RobustModel::Block) {
+    return Base::loss(v);
+  }
+  const Vector whitened = noise_->whiten(v);
+  double result = 0.0;
+  // Linearization reweights the right-hand side -v. Match that convention
+  // here so asymmetric losses agree with the existing linearized gradient.
+  for (Eigen::Index i = 0; i < whitened.size(); ++i) {
+    result += robust_->loss(-whitened[i]);
+  }
+  return result;
+}
+
 void Robust::WhitenSystem(Vector& b) const {
   noise_->whitenInPlace(b);
   robust_->reweight(b);
--- a/gtsam/linear/NoiseModel.h
+++ b/gtsam/linear/NoiseModel.h
@@ -105,9 +105,19 @@ namespace gtsam {
         return std::sqrt(squaredMahalanobisDistance(v));
       }
 
-      /// loss function, input is Mahalanobis distance
+      /// Loss function, input is squared Mahalanobis distance.
       virtual double loss(const double squared_distance) const {
         return 0.5 * squared_distance;
+      }
+
+      /**
+       * Evaluate the loss of an unwhitened residual v. Unlike the
+       * squared-distance overload, this retains the individual components
+       * and their signs for scalar robust losses.
+       * The default delegates to the squared-distance loss.
+       */
+      virtual double loss(const Vector& v) const {
+        return loss(squaredMahalanobisDistance(v));
       }
 
       virtual void WhitenSystem(std::vector<Matrix>& A, Vector& b) const = 0;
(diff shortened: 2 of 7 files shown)


dellaert:
@leolrg I asked Codex to align CUDA behavior with my change above, after copilot flagged a discrepancy now. Could you take a look?

Codex says:

> The CUDA SFM path had the same objective/linearization mismatch as the CPU path: `robustRowScales` applied componentwise weights in Scalar mode, but `computeSfmProjectionErrorKernel` always evaluated the loss of the residual norm. Consequently, the device objective used for initial/trial errors and step acceptance did not match scalar reweighting.

> The kernel now sums the losses of the two whitened components in Scalar mode, using the same signed convention as the CPU scalar objective. Block mode keeps its existing norm-based loss. CUDA currently supports Huber and Tukey, both symmetric, so the sign convention does not change their numerical values.

> The existing mixed-noise CUDA test also expected a block loss for scalar observations; that expectation is corrected. Added regressions check a known scalar Huber objective for both residual signs and compare CUDA initial/final optimization objectives with CPU graph evaluation for Huber and Tukey, in both Scalar and Block modes, with correlated whitening.

I validated on Cyber's NVIDIA GeForce RTX 5060 Ti using a Release build with CUDA 13.0, cuDSS enabled, and deprecated APIs disabled. Against the unchanged kernel, the there were 3 test failures, , but after the fix, both `testCudaSfm.run` and `testRobust.run` passed.

dellaert:
@leolrg ?

ProfFan:
Should `error()` return the robust loss value? From what I remember IRLS should only use the robust objective at linearization time. I could be wrong though.

leolrg:
The CUDA change looks good to me.

One potential CPU issue that codex found: both `BatchFactor` paths apply robust weights before negating the residual, unlike ordinary factors. With scalar AsymmetricCauchy and residual +2, the objective derivative is 2 but the batch gradient is 0.4.

Negating before `WhitenSystem` fixes it. I tested that locally; the new regression and relevant CPU/CUDA tests pass. Could we include this before merging?

dellaert:
@ProfFan I checked this against both the implementation and the IRLS formulation. Nonlinear `error()` should evaluate the robust objective; the linearized Gaussian factor evaluates a quadratic surrogate with weights frozen at the linearization point. LM and Dogleg use the nonlinear graph's error for actual decrease and the linearized graph's error for predicted decrease, so the distinction matters for step acceptance.

The pre-PR implementation already returned robust loss. The defect is that it always applied that loss to the residual norm, even when linearization used independent Scalar weights. This PR makes Scalar objective evaluation sum the component losses, while preserving Block behavior.

For asymmetric Scalar losses, ordinary factor linearization has always reweighted the right-hand side `-r`. Evaluating `sum rho(-r_i)` makes the objective gradient agree with that existing convention, assuming `rho'(t) = t*w(t)`. The frozen quadratic model need not have the same absolute value as the objective: its constant cancels in predicted decrease. For example, Huber with threshold 1 at residual 2 has robust loss 1.5, while the weighted quadratic value is 1.0.

The class-level Robust documentation still describes weighted squared residuals as “error” without making the nonlinear/linearized distinction clear; that wording deserves clarification.

The broader audit also confirmed Ruogu's BatchFactor finding and found separate spherical-camera residual/Jacobian and DCS loss/weight inconsistencies. I am addressing those in separate commits with regression tests.

akshay-krishnan (approved):

PR #2801, opened 2026-09-12
Fix fixed-lag smoother state cleanup and input validation

dellaert:
Removing the last factor referencing a state could leave `BatchFixedLagSmoother` with stale values and timestamps, causing ordering or marginalization failures. This change removes newly unused states before optimization, keeps the factor index and reusable slots synchronized, and erases and reports expired factorless values before constructing the ordering. A replacement factor in the same update preserves its state.

Batch removal requests are deduplicated against live factors at update entry, so an already empty slot cannot delete a newly inserted factor that reuses it. The marginalization cutoff is captured before unused states are erased, so removing the newest state cannot move the cutoff backward.

Batch now also rejects timestamps whose keys have no value in the smoother or incoming `newTheta`, matching Incremental's contract from #2781. The `std::invalid_argument` diagnostic identifies the key. This prevents a timestamp for a nonexistent state from advancing the clock and silently emptying the window. Timestamps for existing or incoming values, including pending factorless values, remain accepted and participate in the clock normally.

Both smoothers validate removal indices against the factor graph at update entry, before timestamp validation or state mutation. Incremental now reports `std::out_of_range` with its class name and the offending index instead of failing inside ISAM2 after partial changes. Adding factors cannot make an invalid removal index valid. Both `update()` methods document the validation order and guarantee that these validation failures leave the smoother unchanged.

Regression coverage includes unused-state cleanup, removal during expiration, empty-slot reuse, cutoff preservation, expired pending values, factor replacement, invalid timestamps mixed with valid additions and removals, corrected retries, accepted pending timestamps, and removal-index diagnostics in both smoothers.

Validation:
- `make -j6 testBatchFixedLagSmoother.run testIncrementalFixedLagSmoother.run` from `build/`: both suites pass.
- `git diff --check` passes.
- The empty-slot, cutoff, timestamp-admission, and Incremental removal-validation regressions were confirmed to fail before their respective fixes.

cc @DLuminary for the related fixed-lag timestamp and state-lifecycle work. The timestamp and removal-validation changes address both observations in the review comment.

### DIFF
--- a/gtsam/nonlinear/BatchFixedLagSmoother.cpp
+++ b/gtsam/nonlinear/BatchFixedLagSmoother.cpp
@@ -23,6 +23,8 @@
 #include <gtsam/linear/GaussianFactorGraph.h>
 #include <gtsam/linear/GaussianFactor.h>
 
+#include <stdexcept>
+
 using namespace std;
 
 namespace gtsam {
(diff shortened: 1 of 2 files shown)


DLuminary:
Two asymmetries with Incremental while you're here (that may also be addressed via some follow-up to this PR). 
* The timestamp-with-no-value check merged in #2781 isn't mirrored — Batch still lets such an entry drive the clock and silently empty the window. 
* The new `out_of_range` for factor indices gives Batch a better message than Incremental's raw `vector::_M_range_check`.

dellaert:
@DLuminary I asked Astra to address both in this PR. Please take a quick look?

DLuminary:
LGTM

PR #2804, opened 2026-09-16
Prepare documentation and ROS colcon defaults for 4.3

dellaert:
## Summary

- refresh the top-level Markdown documentation and correct outdated installation, usage, contributor, and export guidance
- document the tested Clang 14 toolchain and clarify Windows symbol-export guidance
- default `GTSAM_USE_SYSTEM_EIGEN` to `ON` for colcon builds while preserving the existing plain-CMake default and explicit user overrides

## Why

ROS packages declare Eigen as a dependency, so GTSAM and downstream ROS packages should use the same Eigen installation. Official ROS packaging already supplies this setting explicitly; applying the same default to source builds under colcon avoids mixed Eigen versions without changing ordinary CMake builds.

## Validation

- plain CMake configure retains bundled Eigen by default
- colcon-mode configure selects system Eigen by default
- ROS-equivalent core build and install completed successfully
- installed-package C++ consumer configured, built, linked, and ran successfully
- XML parsing and patch whitespace checks passed


dellaert:
## ROS 2 release-readiness summary

The audit found one small source-level alignment worth including before the final 4.3 release: use system Eigen by default when CMake is invoked by colcon. ROS already declares Eigen as a dependency, and the official ROS release repository has supplied `-DGTSAM_USE_SYSTEM_EIGEN=ON` as a packaging override since 2023. Making this the colcon default brings source builds into line with released binary packages and avoids mixing GTSAM's bundled Eigen with the system Eigen used by downstream ROS packages.

The change is deliberately narrow:

- ordinary CMake builds keep the bundled-Eigen default
- explicit `GTSAM_USE_SYSTEM_EIGEN` values still take precedence
- the existing colcon defaults that disable Boost serialization and Boost features are unchanged
- no API, target, package-layout, dependency-model, or directory refactor is involved

The current upstream source has successful ROS development builds for Lyrical, Rolling, Kilted, and Jazzy, each reporting 369 tests with no failures. The ROS manifest's Eigen and TBB dependencies remain appropriate, Boost is correctly absent, and no manifest change is needed. Existing release-repository flags become redundant but remain harmless.

This also addresses the kind of packaging follow-up encountered after the first ROS 4.3 package: `4.3.0-1` was followed by `4.3.0-2` with explicit Boost-disable flags. Those Boost defaults are already handled in source; the Eigen adjustment applies the same principle to the remaining long-lived ROS packaging override.

Items intentionally left outside this PR include expanding SuiteSparse/CHOLMOD packaging, ROS 1 Noetic support, and removing defensive flags from the ROS release repository. For the final stable release, `GTSAM_PRERELEASE_VERSION` still needs to be cleared as part of the normal release-version update, and the final tag will need to be bloomed for the supported ROS 2 distributions.

dellaert:
FYI @matlabbe

jlblancoc:
Having system Eigen enabled in the ROS build farm is finally great to have, thanks! IIRC, I have been overriding that for years with obscure patches in the ros2 gbp repos directly, which was a bit of a pain to maintain.

Having all boost off by default (and not failing to build if boost are not installed at all) would be the other half to have 100% happy, minimal dependencies as a ROS package (just mentioning, maybe was already done?).

jlblancoc (approved):

dellaert:
In ROS/colcon builds, both GTSAM_ENABLE_BOOST_SERIALIZATION and GTSAM_USE_BOOST_FEATURES default to OFF, and this works without a problem.

Thanks for the quick review! I'm having copilot do a final check and then I'll merge.

dellaert:
I pushed another commit which addresses the (minor) concerns flagged by co-pilot.

PR #2810, opened 2026-09-21
Make testFindSeparator independent of METIS' partition labelling

Tobias-Fischer:
### Problem

`testFindSeparator` fails on macOS, with five assertions of the form `expected 1 but was: 2`, even though the partitions it computes are correct:

```
gtsam_unstable/partition/tests/testFindSeparator.cpp:146: Failure: "expected 1 but was: 2"
gtsam_unstable/partition/tests/testFindSeparator.cpp:173: Failure: "expected 1 but was: 2"
gtsam_unstable/partition/tests/testFindSeparator.cpp:220: Failure: "expected 1 but was: 2"
...
There were 5 failures
```

### Cause

The input GTSAM hands METIS is deterministic — there are no unordered containers in `FindSeparator-inl.h`, so the CSR arrays are identical across platforms. The nondeterminism is inside METIS: `METIS_PartGraphRecursive` with `METIS_IPTYPE_GROW` picks its growth seeds from GKlib's RNG, which falls back to the platform libc generator. glibc and Darwin therefore produce different — but equally valid — bisections. The separator is identical; only the two islands' labels (`1` vs `2`), or their `A`/`B` roles, are swapped.

So the labelling is arbitrary, and the test hard-codes one platform's choice.

The test already acknowledges exactly this ambiguity for `edgePartitionByMetis`:

```cpp
CHECK(A_expected == actual->A || A_expected == actual->B);
CHECK(B_expected == actual->B || B_expected == actual->A);
```

### Fix

Apply that same reasoning to the remaining assertions: canonicalize the labelling before comparing, instead of hard-coding one platform's arbitrary result. Only the test's expectations change — no library code is touched.

### Verification

- Fails before, passes after, on macOS arm64.
- Verified against both METIS **5.1.0** and **5.1.1** (the failure count actually differs between them — 1 vs 5 — which is further evidence that the labelling is incidental rather than meaningful).
- Also green in conda-forge CI on `osx_arm64`, where `testFindSeparator` now passes as part of the full 385-test suite.
- Should be a no-op on Linux, where the labels already come out in the expected order.

### Provenance

Found while packaging GTSAM 4.3.0 for conda-forge, where the full `ctest` suite is now run on macOS as well: https://github.com/conda-forge/gtsam-feedstock/pull/44

This patch was written by Claude (Anthropic's AI assistant) working in Claude Code, and reviewed by me before submitting. Happy to adjust the approach if you would prefer the canonicalization done differently — e.g. comparing partitions as sets rather than normalizing labels.


dellaert (approved):

Tobias-Fischer:

PR #2811, opened 2026-09-21
Let the caller override Boost_USE_STATIC_LIBS on MSVC

Tobias-Fischer:
### Problem

`cmake/HandleBoost.cmake` unconditionally forces static Boost under MSVC:

```cmake
if(MSVC)
    # By default, Boost pre-compiled binaries for Windows are static libraries.
    set(Boost_USE_STATIC_LIBS ON)
```

This makes it impossible to build against a **shared** Boost on Windows. Configuring with `-DBoost_USE_STATIC_LIBS=OFF` has no effect, because the value is overwritten before `find_package(Boost)` runs.

There is a second, smaller consequence. The branch immediately below is currently dead code:

```cmake
    if(NOT Boost_USE_STATIC_LIBS)
        list_append_cache(GTSAM_COMPILE_DEFINITIONS_PUBLIC BOOST_ALL_DYN_LINK)
    endif()
```

`Boost_USE_STATIC_LIBS` was just set to `ON` two lines above, so `BOOST_ALL_DYN_LINK` can never be defined — even though the surrounding code clearly intends to support the shared case.

### Fix

Only apply the default when the caller has not already made a choice:

```cmake
    if(NOT DEFINED Boost_USE_STATIC_LIBS)
        set(Boost_USE_STATIC_LIBS ON)
    endif()
```

Behaviour is unchanged for anyone who does not set the variable — vcpkg and the pre-compiled-binary case still get static Boost by default. It also makes the `BOOST_ALL_DYN_LINK` branch reachable, as intended.

### Motivation

conda-forge ships Boost on Windows as shared libraries. The feedstock already passes `-DBoost_USE_STATIC_LIBS:BOOL=OFF`, but has to carry a patch flipping this line because the flag is ignored. With this change the patch can be dropped and the standard CMake option is enough.

### Provenance

Found while packaging GTSAM 4.3.0 for conda-forge: https://github.com/conda-forge/gtsam-feedstock/pull/44

This patch was written by Claude (Anthropic's AI assistant) working in Claude Code, and reviewed by me before submitting.

Note on testing: I have verified the equivalent change in the conda-forge Windows build (which is what the existing downstream patch does); I have not built this branch standalone on Windows with static Boost, so a CI run on your side would be worthwhile to confirm the default path is untouched.


dellaert (approved):

ProfFan:
I recall this was due to some issues with a dynamically linked Boost on Windows. Might be already resolved though.

Tobias-Fischer:

PR #2812, opened 2026-09-21
Link Threads::Threads so std::thread resolves on older glibc

Tobias-Fischer:
### Problem

Linking `testPriorityScheduler` fails on a toolchain targeting glibc < 2.34:

```
x86_64-conda-linux-gnu-ld: gtsam/base/tests/CMakeFiles/testPriorityScheduler.dir/testPriorityScheduler.cpp.o:
  in function `std::thread::_M_thread_deps_never_run()':
testPriorityScheduler.cpp:(...): undefined reference to `pthread_create'
testPriorityScheduler.cpp:(...): undefined reference to `pthread_join'
collect2: error: ld returned 1 exit status
```

### Cause

`gtsam/base/PriorityScheduler.h` uses `std::thread`, and `testPriorityScheduler` exercises it, but GTSAM never calls `find_package(Threads)` or links `Threads::Threads` anywhere in the build:

```console
$ grep -rn "Threads::Threads\|find_package(Threads\|CMAKE_THREAD_LIBS\|-pthread" cmake/ CMakeLists.txt gtsam/CMakeLists.txt
$   # no matches
```

Since glibc 2.34 the pthread symbols are folded into libc, so toolchains targeting a recent glibc link successfully by accident. Anything targeting an older glibc — which is normal for redistributable Linux binaries — needs `-pthread` explicitly.

This does not appear to be a regression so much as newly exposed: `PriorityScheduler` is recent, and it is the first thing to pull `std::thread` into a test.

### Fix

```cmake
find_package(Threads REQUIRED)
target_link_libraries(gtsam PUBLIC Threads::Threads)
```

`PUBLIC` so tests and downstream consumers inherit it. On platforms where pthread already lives in libc this is a no-op: CMake reports `CMAKE_HAVE_LIBC_PTHREAD=1` and adds no flags, which I confirmed on macOS — the generated `build.ninja` contains no `pthread` and nothing relinks.

### Verification

- Reproduced on conda-forge's linux-64 and linux-aarch64 builders (all 8 jobs).
- Confirmed the CMake is valid and a genuine no-op on macOS arm64.
- I have **not** yet seen a green Linux CI run with this applied — it is building now at https://github.com/conda-forge/gtsam-feedstock/pull/44, and I will follow up here with the result.

### Provenance

Found while packaging GTSAM 4.3.0 for conda-forge, which runs the full `ctest` suite: https://github.com/conda-forge/gtsam-feedstock/pull/44

This patch was written by Claude (Anthropic's AI assistant) working in Claude Code, and reviewed by me before submitting.


dellaert (approved):

dellaert:
Thanks for these three PRs, @Tobias-Fischer ! Note to self: "try to build for Conda before releasing."  :-)

Tobias-Fischer:
No problem @dellaert. Would you be interested in a conda GitHub workflow? Happy to create one if you like.

PS: see you at IROS?

dellaert:
> No problem @dellaert. Would you be interested in a conda GitHub workflow? Happy to create one if you like.

Yes, that would be nice. If all these PRs are related to conda then maybe you could stack them on top of each other? Maybe we can merge in one go, Including the workflow. 

> PS: see you at IROS?

Unfortunately not :-( Too much travel.

Tobias-Fischer:

PR #2814, opened 2026-09-21
Fix DoglegLineSearch throwing on a zero Gauss-Newton step

### LINKED ISSUE #2813: ISAM2 with `ISAM2DoglegLineSearchParams` throws "Would cause infinite search" when the Gauss-Newton step is zero
This was encountered while working on my robot vacuum system and integrating riSAM.

### Description

With `ISAM2DoglegLineSearchParams`, `ISAM2::calculateEstimate()` throws

```
Invalid DoglegLineSearch configuration. Would cause infinite search.
```

whenever the Gauss-Newton step is exactly zero, i.e. when the current linearization point already satisfies the factors. That's a valid state, not a configuration error, and the default parameters hit it.

`RISAM` hits it much more often, because `RISAM::updateRobust` calls `solver_->calculateEstimate()` after every internal `ISAM2::update` ("force back-substitution"). So with the Dogleg line-search parameters (the ones `testRISAM`'s integration test uses), riSAM throws on its first update whenever the initial value is exact. A common case is seeding the first pose at its own prior. `testRISAM` doesn't catch this because it perturbs every initial guess.

### Minimal reproducer (GTSAM 4.3.0, `71a25ca`)

```cpp
#include <gtsam/geometry/Pose2.h>
#include <gtsam/nonlinear/ISAM2.h>

#include <iostream>

using namespace gtsam;

int main() {
  const auto noise = noiseModel::Isotropic::Sigma(3, 0.1);
  NonlinearFactorGraph graph;
  graph.addPrior(0, Pose2(0, 0, 0), noise);

  for (const bool exact : {false, true}) {
    Values init;
    init.insert(0, exact ? Pose2(0, 0, 0) : Pose2(0.01, 0.01, 0.01));
    ISAM2 isam(ISAM2Params{ISAM2DoglegLineSearchParams{}});
    isam.update(graph, init);
    try {
      isam.calculateEstimate();
      std::cout << (exact ? "exact    " : "perturbed") << " initial value: ok\n";
    } catch (const std::exception& e) {
      std::cout << (exact ? "exact    " : "perturbed") << " initial value: threw: "
                << e.what() << "\n";
    }
  }
}
```

Output:

```
perturbed initial value: ok
exact     initial value: threw: Invalid DoglegLineSearch configuration. Would cause infinite search.
```

Replacing `ISAM2` with `RISAM` (with `rp.isam2Params` set to the same parameters) fails the same way, because riSAM calls `calculateEstimate()` internally. Plain `ISAM2` without that call doesn't throw, because the delta is only updated lazily.

### Cause

In `DoglegLineSearchImpl::Iterate` (`gtsam/nonlinear/DoglegOptimizerImpl.h`, added in #2408):

```cpp
const double gnStep = dx_n.norm();
...
maxStep = std::min(params.maxDelta * numVars, gnStep);
step = std::min(params.minDelta * numVars, gnStep);
...
if (step < 1e-12 || params.stepSize < 1.0)
  throw std::runtime_error(
      "Invalid DoglegLineSearch configuration. Would cause infinite search.");
```

`step` is capped by `gnStep`, so a zero Gauss-Newton step gives `step == 0`, and the guard treats that as bad configuration. Only the `stepSize < 1.0` half of the guard is a configuration error. `step == 0` just means x0 already minimizes the linear model, so there's nothing to search.

### Proposed fix

Keep the configuration check, and return the zero step instead of throwing. At `step == 0`, `ComputeDoglegPoint(step, dx_u, dx_n)` has already returned `dx_n` (== 0), so `result` is already correct:

```diff
--- a/gtsam/nonlinear/DoglegOptimizerImpl.h
+++ b/gtsam/nonlinear/DoglegOptimizerImpl.h
@@ -331,10 +331,16 @@ typename DoglegOptimizerImpl::IterationResult DoglegLineSearchImpl::Iterate(
     std::cout << "Initial Step Error: " << result.f_error << std::endl;
 
   // Validate Step size will terminate search
-  if (step < 1e-12 || params.stepSize < 1.0)
+  if (params.stepSize < 1.0)
     throw std::runtime_error(
         "Invalid DoglegLineSearch configuration. Would cause infinite search.");
 
+  // A zero Gauss-Newton step means x0 already minimizes the linear model, so
+  // there is nothing to search. This is a valid state (e.g. the initial values
+  // already satisfy the factors), not a configuration error. With step == 0
+  // ComputeDoglegPoint returned dx_n (== 0), so result is already the answer.
+  if (step < 1e-12) return result;
+
   // Search Increase delta
   double eps = std::numeric_limits<double>::epsilon();
   while (step < maxStep - eps) {
```

Tested on 4.3.0 (aarch64, GCC, Release, TBB on):
- with the patch, the reproducer and its `RISAM` variant both succeed with exact and perturbed initial values;
- `testDoglegOptimizer`, `testGaussianISAM2` and `testRISAM` all pass.

A regression test could add the exact-initial-value case above to `testGaussianISAM2` (Dogleg line-search parameters, then `calculateEstimate()`), or to `testRISAM`.

### Workaround

Use the default Gauss-Newton `ISAM2Params` with `RISAM`, or perturb initial values. We hit this integrating riSAM as a pose-graph backend, where the first pose is seeded exactly at its prior.

oursland:
## Summary

`DoglegLineSearchImpl::Iterate` throws `Invalid DoglegLineSearch configuration. Would cause infinite search.` whenever the Gauss-Newton step is exactly zero, i.e. when the linearization point already minimizes the linear model. That's a valid state, not a configuration error, and the default parameters hit it.

Consequences, with `ISAM2DoglegLineSearchParams`:
- `ISAM2::calculateEstimate()` throws after any update whose initial values already satisfy their factors. A common case is seeding the first pose exactly at its prior.
- `RISAM` throws on such a first update, because `RISAM::updateRobust` calls `calculateEstimate()` after every internal update. The riSAM integration test uses these parameters but perturbs every initial guess, so it doesn't catch this.

## Cause

The first step is capped by the Gauss-Newton step norm, and the guard added in #2408 checks it together with the configuration:

```cpp
step = std::min(params.minDelta * numVars, gnStep);   // 0 when gnStep == 0
...
if (step < 1e-12 || params.stepSize < 1.0)
  throw std::runtime_error("Invalid DoglegLineSearch configuration. ...");
```

Only `stepSize < 1.0` makes the search non-terminating. `step == 0` just means there's nothing to search.

## Fix

Keep the `stepSize` check, and return the zero step instead of throwing. At `step == 0`, `ComputeDoglegPoint` has already returned `dx_n` (== 0), and `f_error` is the initial error, so `result` is already correct.

## Tests

- `DoglegOptimizer.IterateLineSearchZeroGaussNewtonStep`: `Iterate` at the minimum returns a zero step with an unchanged error, and `stepSize < 1` still throws.
- `ISAM2.DoglegLineSearchZeroGaussNewtonStep`: an exact prior, then an exact odometry update; `calculateEstimate()` returns the values in both cases.

Both new tests fail without the fix, with the exception above, and pass with it. `testDoglegOptimizer`, `testGaussianISAM2`, `testRISAM`, `testNonlinearISAM` and `testIncrementalFixedLagSmoother` all pass. Built on `develop` (`6e63185`) on aarch64, GCC, Release, with TBB.

We found this while integrating riSAM as a pose-graph backend in a robot-vacuum SLAM stack, where the first pose is seeded at its prior.

Closes #2813


dellaert (approved):
LGTM. Thanks !!!

